PR #1125, opened 2020-10-21
Revert: Define Propagation-only Span to simplify active Span logic in Context

tsloughter:
This reverts commit d14c9b7a74328c22e5d2223c4ddcaca5630aceae.

Fixes #1124

PR #1131, opened 2020-10-22
Make clear the propagated/non-recording span concept needs no type.

### LINKED ISSUE #1087: Clarify purpose of `PropagatedSpan`
From the discussion on #1086 and as discussed in the SIG spec meeting:
We should clarify the purpose of `PropagatedSpan` since the current description in the API seems to be confusing.

### LINKED ISSUE #1124: Revert use of PropagatedSpan back to using a special Context key for extracted Span Context

In discussion last night in the Spec SIG meeting we talked about the confusion added with PropagatedSpan and whether it didn't cause more confusion than the issue it was trying to resolve (the code required to check if an extracted Span Context exists if there isn't a parent in the current Context or the Context manually passed as the parent to `StartSpan`.

I'm opening a revert of the PropagatedSpan PR to go along with this issue.

Oberon00:
Fixes #1087, closes #1124, closes #1125

## Changes

Makes clear that the non-recording span / propagated Span has no semantic meaning and should not be a new public type.

### DIFF
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -645,7 +645,7 @@ as a `Span` in operations such as in-process `Span` propagation.
 
 If a new type is required for supporting this operation, it SHOULD not be exposed
 publicly if possible (e.g. by only exposing a function that returns something
-with the Span interface type). If a new type is required to be publicly exposed
+with the Span interface type). If a new type is required to be publicly exposed,
 it SHOULD be named `NonRecordingSpan`.
 
 The behavior is defined as follows:
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -32,7 +32,7 @@ Table of Contents
     * [End](#end)
     * [Record Exception](#record-exception)
   * [Span lifetime](#span-lifetime)
-  * [Propagated Span creation](#propagated-span-creation)
+  * [Wrapping a SpanReference in a Span](#wrapping-a-spanreference-in-a-span)
 * [SpanKind](#spankind)
 * [Concurrency](#concurrency)
 * [Included Propagators](#included-propagators)
@@ -377,8 +377,8 @@ When a new `Span` is created from a `Context`, the `Context` may contain a `Span
 representing the currently active instance, and will be used as parent.
 If there is no `Span` in the `Context`, the newly created `Span` will be a root span.
 
-A `SpanReference` cannot be set as active in a `Context` directly, but through the use
-of a [Propagated Span](#propagated-span-creation) wrapping it.
+A `SpanReference` cannot be set as active in a `Context` directly, but by
+[wrapping it into a Span](#wrapping-a-spanreference-in-a-span).
 For example, a `Propagator` performing context extraction may need this.
 
 #### Specifying links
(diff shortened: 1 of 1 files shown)


specification/trace/api.md
```diff
@@ -637,13 +637,16 @@ timestamps to the Span object:
...
 
-### Propagated Span creation
+### Wrapping a SpanReference in a Span
 
 The API MUST provide an operation for wrapping a `SpanReference` with an object
 implementing the `Span` interface. This is done in order to expose a `SpanReference`
 as a `Span` in operations such as in-process `Span` propagation.
 
-If a new type is required for supporting this operation, it SHOULD be named `PropagatedSpan`.
+If a new type is required for supporting this operation, it SHOULD not be exposed
```
bogdandrutu, line 646: Maybe also clarify what @tsloughter suggested, the returned wrapped Span has `is_recording` return always false

bogdandrutu (approved):

specification/trace/api.md
Oberon00: This is already written below:

> The behavior is defined as follows:
> ...
> - `IsRecording` MUST return `false` to signal that events, attributes and other elements
>  are not being recorded, i.e. they are being dropped.

bogdandrutu: Now I am confused about the discussion in the issue, need to re-read.

Oberon00: I *think* @tsloughter understood that PropagatedSpan was meant to be distinguishable by users like with isinstance checks and wondered why IsRecording is not enough for that.

tsloughter:
I just don't get why any of this is needed. I think I'd be fine with not using a separate context key if none of this was added to the API spec and the spec of the extractor simply said a Span should be created from propagated trace context with `is_recording` as `false`.

Oberon00:
Also, from a performance perspective, a non-recording span can be implemented more efficiently and lightweight than a full span.

Oberon00:
And there is also no other API to create a non-recording span (except if you muck around with samplers maybe).

carlosalberto (approved):

jmacd (approved):

anuraaga (approved):
This is much easier for me to understand, thanks

specification/trace/api.md
```diff
@@ -637,13 +637,16 @@ timestamps to the Span object:
...
+publicly if possible (e.g. by only exposing a function that returns something
+with the Span interface type). If a new type is required to be publicly exposed
```
arminru, line 648:
```suggestion
with the Span interface type). If a new type is required to be publicly exposed,
```

arminru (approved):

carlosalberto:
@open-telemetry/specs-approvers Please review. The more eyes/approvals, the better ;)

bogdandrutu:
@yurishkuro @SergeyKanzhelev ?

carlosalberto:
Ping @SergeyKanzhelev @yurishkuro

yurishkuro (approved):

SergeyKanzhelev (approved):

PR #1135, opened 2020-10-23
Span processor clarifications

rachelleahklein:
## Changes

This PR attempts to clarify information about span processors in the Trace SDK spec. It contains:
* More information about when one might want to use a simple vs. batching span processor
* More information about scenarios in which span processors may be used without exporters
* Minor grammatical fixes

### Related issues
The need for clarification in the spec came to my attention when it was mentioned in this ticket: https://github.com/open-telemetry/opentelemetry-ruby/issues/397.

### DIFF
--- a/specification/trace/sdk.md
+++ b/specification/trace/sdk.md
@@ -293,11 +293,11 @@ exporters such as [z-pages](../../experimental/trace/zpages.md).
 Span processors can be registered directly on SDK `TracerProvider` and they are
 invoked in the same order as they were registered.
 
-Each processor registered on `TracerProvider` is a start of pipeline that consist
-of span processor and optional exporter. SDK MUST allow to end each pipeline with
+Each processor registered on `TracerProvider` is a start of a pipeline that consists
+of a span processor and an optional exporter. The SDK MUST allow to end each pipeline with
 individual exporter.
 
-SDK MUST allow users to implement and configure custom processors and decorate
+The SDK MUST allow users to implement and configure custom processors and decorate
 built-in processors for advanced scenarios such as tagging or filtering.
 
 The following diagram shows `SpanProcessor`'s relationship to other components
@@ -392,13 +392,13 @@ make the flush timeout configurable.
 
 The standard OpenTelemetry SDK MUST implement both simple and batch processors,
 as described below. Other common processing scenarios should be first considered
-for implementation out-of-process in [OpenTelemetry Collector](../overview.md#collector)
+for implementation out-of-process in an [OpenTelemetry Collector](../overview.md#collector).
 
 #### Simple processor
 
 This is an implementation of `SpanProcessor` which passes finished spans
-and passes the export-friendly span data representation to the configured
-`SpanExporter`, as soon as they are finished.
+and the export-friendly span data representation to the configured
+`SpanExporter` as soon as they are finished.
 
 **Configurable parameters:**
 
@@ -406,21 +406,21 @@ and passes the export-friendly span data representation to the configured
 
 #### Batching processor
 
-This is an implementation of the `SpanProcessor` which create batches of finished
+This is an implementation of `SpanProcessor` which creates batches of finished
 spans and passes the export-friendly span data representations to the
 configured `SpanExporter`.
 
 **Configurable parameters:**
 
 * `exporter` - the exporter where the spans are pushed.
-* `maxQueueSize` - the maximum queue size. After the size is reached spans are
+* `maxQueueSize` - the maximum queue size. After this limit is reached, spans are
   dropped. The default value is `2048`.
 * `scheduledDelayMillis` - the delay interval in milliseconds between two
   consecutive exports. The default value is `5000`.
 * `exportTimeoutMillis` - how long the export can run before it is cancelled.
   The default value is `30000`.
 * `maxExportBatchSize` - the maximum batch size of every export. It must be
-  smaller or equal to `maxQueueSize`. The default value is `512`.
+  smaller than or equal to `maxQueueSize`. The default value is `512`.
 
 ## Span Exporter
 
--- a/specification/trace/sdk.md
+++ b/specification/trace/sdk.md
@@ -411,9 +411,8 @@ in the background.
 
 #### Batching processor
 
-This is an implementation of `SpanProcessor` which creates batches of finished
-spans and passes the export-friendly span data representations to the
-configured `SpanExporter`.
+This is an implementation of `SpanProcessor` which collects finished spans into batches
+and passes the export-friendly span data representations to the associated `SpanExporter`.
 
 Typically, the batching processor will be more suitable for production environments
 than the simple processor.
--- a/specification/trace/sdk.md
+++ b/specification/trace/sdk.md
@@ -399,11 +399,11 @@ This is an implementation of `SpanProcessor` which passes finished spans
 and the export-friendly span data representation to the configured
 `SpanExporter` as soon as they are finished.
 
-Typically, the simple processor will be most suitable for use in testing. Exceptions to this
-include scenarios where creating multiple threads is not desirable (since batching
-span processors will typically be threaded in order to pass spans on to exporters
-in the background) and scenarios where different custom attributes should be added to individual
-spans based on code scopes.
+Typically, the simple processor will be most suitable for use in testing; it should be used with
+caution in production. It may be appropriate for production use in scenarios where creating
+multiple threads is not desirable (since batching span processors will typically be threaded
+in order to pass spans on to exporters in the background) as well as scenarios where different
+custom attributes should be added to individual spans based on code scopes.
 
 **Configurable parameters:**
 
--- a/specification/trace/sdk.md
+++ b/specification/trace/sdk.md
@@ -400,10 +400,11 @@ This is an implementation of `SpanProcessor` which passes finished spans
 and the export-friendly span data representation to the configured
 `SpanExporter` as soon as they are finished.
 
-Typically, the simple processor will be most suitable for use in testing and/or
-in scenarios where creating multiple threads is not desirable, since batching
+Typically, the simple processor will be most suitable for use in testing. Exceptions to this
+include scenarios where creating multiple threads is not desirable (since batching
 span processors will typically be threaded in order to pass spans on to exporters
-in the background.
+in the background) and scenarios where different custom attributes should be added to individual
+spans based on code scopes.
 
 **Configurable parameters:**
 
--- a/specification/trace/sdk.md
+++ b/specification/trace/sdk.md
@@ -399,8 +399,8 @@ This is an implementation of `SpanProcessor` which passes finished spans
 and the export-friendly span data representation to the configured
 `SpanExporter` as soon as they are finished.
 
-Typically, the simple processor will be most suitable for use in testing; it should be used with
-caution in production. It may be appropriate for production use in scenarios where creating
+The simple processor must not be used lightly in production. It runs on application thread and may
+degrade performance. It may be appropriate for production use in scenarios where creating
 multiple threads is not desirable (since batching span processors will typically be threaded
 in order to pass spans on to exporters in the background) as well as scenarios where different
 custom attributes should be added to individual spans based on code scopes.
--- a/specification/trace/sdk.md
+++ b/specification/trace/sdk.md
@@ -297,8 +297,7 @@ Each processor registered on `TracerProvider` is a start of a pipeline that cons
 of one or more span processors and, optionally, one or more exporters. The SDK MUST
 allow ending each pipeline with an individual exporter.
 
-The SDK MUST allow users to implement and configure custom processors and decorate
-built-in processors for advanced scenarios such as tagging or filtering.
+The SDK MUST allow users to implement and configure custom processors.
 
 The following diagram shows `SpanProcessor`'s relationship to other components
 in the SDK:
--- a/specification/trace/sdk.md
+++ b/specification/trace/sdk.md
@@ -294,8 +294,8 @@ Span processors can be registered directly on SDK `TracerProvider` and they are
 invoked in the same order as they were registered.
 
 Each processor registered on `TracerProvider` is a start of a pipeline that consists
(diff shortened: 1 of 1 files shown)


specification/trace/sdk.md
```diff
@@ -391,35 +392,43 @@ make the flush timeout configurable.
...
 
 #### Simple processor
 
 This is an implementation of `SpanProcessor` which passes finished spans
-and passes the export-friendly span data representation to the configured
-`SpanExporter`, as soon as they are finished.
+and the export-friendly span data representation to the configured
+`SpanExporter` as soon as they are finished.
+
+Typically, the simple processor will be most suitable for use in testing and/or
```
SergeyKanzhelev, line 403: it may be used in production for adding attributes like here: https://medium.com/opentelemetry/opentelemetry-beyond-getting-started-5ac43cd0fe26#ffbb:~:text=Custom%20attributes%20in%20code%20scopes

SergeyKanzhelev (changes requested):
thank you for PR. I added comment on one more scenario for processor. Current edit may mislead

rachelleahklein:
Thanks for the quick response! I will add that further clarification.

specification/trace/sdk.md
```diff
@@ -287,16 +287,17 @@ invocations. The span processors are invoked only when
 [`IsRecording`](api.md#isrecording) is true.
 
 Built-in span processors are responsible for batching and conversion of spans to
-exportable representation and passing batches to exporters.
+exportable representation and passing batches to exporters or alternatives to
```
reyang, line 290: Not sure if a GA spec should point to an experimental spec, thoughts?

specification/trace/sdk.md
```diff
@@ -391,35 +392,43 @@ make the flush timeout configurable.
...
 **Configurable parameters:**
 
 * `exporter` - the exporter where the spans are pushed.
 
 #### Batching processor
 
-This is an implementation of the `SpanProcessor` which create batches of finished
+This is an implementation of `SpanProcessor` which creates batches of finished
 spans and passes the export-friendly span data representations to the
 configured `SpanExporter`.
```
Oberon00, line 416:
```suggestion
This is an implementation of `SpanProcessor` which collects finished spans into batches
and passes the export-friendly span data representations to the associated `SpanExporter`.
```

specification/trace/sdk.md
```diff
@@ -391,35 +392,43 @@ make the flush timeout configurable.
...
 
+Typically, the batching processor will be more suitable for production environments
+than the simple processor.
```
Oberon00, line 419: Depends on the language, e.g. in Java exporters do batching themselves sometimes. But typically, I think that BatchSpanProcessor should be the expected scenario, yes.

specification/trace/sdk.md
```diff
@@ -391,35 +392,43 @@ make the flush timeout configurable.
...
+
 **Configurable parameters:**
 
 * `exporter` - the exporter where the spans are pushed.
-* `maxQueueSize` - the maximum queue size. After the size is reached spans are
+* `maxQueueSize` - the maximum queue size. After this limit is reached, spans are
   dropped. The default value is `2048`.
```
Oberon00, line 425: This is now misleading because the spans cannot really be dropped (as in sampled out) because the SpanProcessor has no back-channel to the TracerProvider. Would be nice if this was possible, but for now I'd say

```suggestion
  skipped by this BatchSpanProcessor. The default value is `2048`.
```

specification/trace/sdk.md
```diff
@@ -287,16 +287,17 @@ invocations. The span processors are invoked only when
...
 
 Span processors can be registered directly on SDK `TracerProvider` and they are
 invoked in the same order as they were registered.
 
-Each processor registered on `TracerProvider` is a start of pipeline that consist
-of span processor and optional exporter. SDK MUST allow to end each pipeline with
-individual exporter.
+Each processor registered on `TracerProvider` is a start of a pipeline that consists
+of one or more span processors and, optionally, one or more exporters. The SDK MUST
+allow ending each pipeline with an individual exporter.
```
Oberon00, line 299: I think we should remove this paragraph. Or do you know what is meant by it? AFAIK the TracerProvider must allow the registration of multiple SpanProcessors, but each built-in span processor only supports a single exporter.

```suggestion
```

specification/trace/sdk.md
```diff
@@ -287,16 +287,17 @@ invocations. The span processors are invoked only when
...
 
-SDK MUST allow users to implement and configure custom processors and decorate
+The SDK MUST allow users to implement and configure custom processors and decorate
 built-in processors for advanced scenarios such as tagging or filtering.
```
Oberon00, line 302: I think we should remove this paragraph too. Unless someone can tell what "tagging" means here or how the SDK can allow "filtering". CC @lmolkova 

```suggestion
```

specification/trace/sdk.md
rachelleahklein: Would it be true then to say "and passing batches to exporters or other span processors," or are there still other alternatives?

specification/trace/sdk.md
rachelleahklein: Thanks for your valuable feedback on this section as well as the batch processor one, @Oberon00. Do you think more should be added to clarify that while these statements may "typically" be true, it also depends on language implementation? I'm not sure how much detail to get into here.

specification/trace/sdk.md
rachelleahklein: Personally, I don't feel qualified to recommend whether this paragraph should be removed. I wasn't trying to clarify any concepts here, just cleaning up. Hopefully others can weigh in.

specification/trace/sdk.md
rachelleahklein: Also, to @reyang's point, I am not even sure referring to zPages specifically makes sense here. My only goal is to have a clear definition of span processors in this sentence. Reading this section, I was confused by the fact that span processors were defined as "responsible for... passing batches to exporters" but then a few lines down, exporters are described as optional.

specification/trace/sdk.md
iNikem: The paragraph itself is needed, at least the part about adding custom processors. Examples "tagging or filtering" are indeed misleading and may be removed.

specification/trace/sdk.md
iNikem: I think we definitely need some better description of the possible compositions of `SpanProcessors` and `SpanExporters`. E.g. we MUST indicate somewhere that there should be a `SpanProcessor` before each `SpanExporter` (purely from methods signatures point of view). We also MUST stress and clearly describe that multiple registered `SpanProcessor` are "siblings" and do NOT call each other. All this has already created some confusion in the past.

iNikem (approved):
This PR does not solve existing problems with the current description, but it is a nice cleanup. Thank you! :)

rachelleahklein:
I have attempted to address @SergeyKanzhelev's requested change. Please let me know if more clarification is needed there.

Thanks to all who have submitted feedback. I think there are a few things that are potentially outside the scope of this PR, but I have tried to incorporate fixes that are straightforward.

One outstanding question I have: does the overall definition of span processors (see discussion on ln. 290) need to be further fixed?

specification/trace/sdk.md
SergeyKanzhelev: I don't see a problem referring to experimental feature as it makes the point that this is an extensibility point in SDK that can be useful for purposes including experimentation with the new features.

specification/trace/sdk.md
```diff
@@ -391,35 +391,43 @@ make the flush timeout configurable.
...
 
 #### Simple processor
 
 This is an implementation of `SpanProcessor` which passes finished spans
-and passes the export-friendly span data representation to the configured
-`SpanExporter`, as soon as they are finished.
+and the export-friendly span data representation to the configured
+`SpanExporter` as soon as they are finished.
+
+Typically, the simple processor will be most suitable for use in testing. Exceptions to this
```
SergeyKanzhelev, line 402: I don't want to position these non-testing scenarios as exceptions. These should be used with caution, but they are valid use cases for production use

specification/trace/sdk.md
SergeyKanzhelev: I like how it's written. Wdyt @Oberon00?

SergeyKanzhelev:
@rachelleahklein thanks again for the PR. The only outstanding feedback from me is to position simple span processor scenarios as dangerous, but valid for production (not as exceptions).

specification/trace/sdk.md
Oberon00: I think the PR's wording is fine here 👍

SergeyKanzhelev (approved):
This is an improvement to the wording without semantical change. Thank you!

justinfoote (approved):
Thank's for the clarifications @rachelleahklein!

specification/trace/sdk.md
jmacd: I support removing this paragraph. We don't need to specify that pipelines end in exporters, just that the two builtin span processors end in exporters. The metric SDK hasn't thus far mentioned support for multiple processors, and why should we? It's possible to define a multi-processor that calls a series of individual processors. Should the default trace SDK include a builtin multi-processor? I suppose they should if it's common to implement multiple trace exporters with different processor settings, but that does not sound common. Wouldn't you prefer to have a single processor call multiple exporters? Why do we need to specify these things?

> multiple registered SpanProcessor are "siblings" and do NOT call each other

If you only specify SDK support for a single processor, this problem vanishes. Potentially you can also specify a multi-processor and/or a multi-exporter, but are you sure these are commonly needed?

rachelleahklein: For what it's worth, as someone relatively new to OpenTelemetry concepts, it was helpful to me to have a textual description the span processor pipeline in addition to the diagram that also exists in this section. There may be some additional nuance that needs to be added, but having some form of this paragraph seems like a good idea to me.

specification/trace/sdk.md
```diff
@@ -391,35 +391,43 @@ make the flush timeout configurable.
...
 
 #### Simple processor
 
 This is an implementation of `SpanProcessor` which passes finished spans
-and passes the export-friendly span data representation to the configured
-`SpanExporter`, as soon as they are finished.
+and the export-friendly span data representation to the configured
+`SpanExporter` as soon as they are finished.
+
+Typically, the simple processor will be most suitable for use in testing; it should be used with
```
cijothomas, line 402: Not sure this should be mentioned here.
We have production systems running using SimpleProcessors. In those systems, the OpenTelemetry SDK uses SimpleProcessor and exports the span as they arrive, without any batching. 
The export may be done to some agent, which in turn may do batching.
Or Export could be done to the actual backend, to enable scenarios like "Live telemetry".  Eg in Azure: https://docs.microsoft.com/azure/azure-monitor/app/live-stream

specification/trace/sdk.md
```diff
@@ -391,35 +391,43 @@ make the flush timeout configurable.
...
+caution in production. It may be appropriate for production use in scenarios where creating
```
SergeyKanzhelev, line 403:
```suggestion
Typically, the simple processor will be most suitable for use in testing; it should be used with
The simple processor must not be used lightly in production. It runs on application thread and may
degrade performance. It may be appropriate for production use in scenarios where creating
```

specification/trace/sdk.md
```diff
@@ -391,35 +391,43 @@ make the flush timeout configurable.
...
+multiple threads is not desirable (since batching span processors will typically be threaded
+in order to pass spans on to exporters in the background) as well as scenarios where different
+custom attributes should be added to individual spans based on code scopes.
```
SergeyKanzhelev, line 407:
```suggestion
custom attributes should be added to individual spans based on code scopes. Simple processors
might also be used for scenarios where the callback needs to be called before sampling. For
example, z-pages can be implemented this way.

```

specification/trace/sdk.md
SergeyKanzhelev: Added couple suggestions. @anuraaga, @cijothomas are these suggestions looks OK for you?

specification/trace/sdk.md
Oberon00: This is wrong though. The simple processor is also not invoked for DROPped spans.

specification/trace/sdk.md
rachelleahklein: Thank you for adding more suggestions, @SergeyKanzhelev. I have incorporated the first (though happy to revert or refine if there are further comments).

The second is still open and I'm hoping @Oberon00 or others can help clarify a more accurate version.

Thank you all!

specification/trace/sdk.md
SergeyKanzhelev: hm, I'm referring to the sampled flag as oppose to the recorded flag. Recorded, but sampled out spans should not make it to the batch processor, correct? Recorded will be processed by simple processors. dropped will not reach either

rachelleahklein:
I don't seem to be able to reopen this PR, but I still do hope to work on it and get it approved. It sounds like there are some clarifications that people have found helpful. Please let me know if there is any further feedback I can address. Thank you.

specification/trace/sdk.md
```diff
@@ -287,17 +287,17 @@ invocations. The span processors are invoked only when
 [`IsRecording`](api.md#isrecording) is true.
 
 Built-in span processors are responsible for batching and conversion of spans to
-exportable representation and passing batches to exporters.
+exportable representation and passing batches to exporters or alternatives to
```
bogdandrutu, line 290: I think we are mixing 2 different things in this sentence:
1. What is a SpanProcessor, and how can this be used to interact with the SDK.
  - The SpanProcessor in general does not need to batch or talk to the exporter, it is a simple "hook" that allows custom processors to interact with the Span lifetime events (on start and on end). For example somebody can implement a processor that tracks all the active requests (one of the functionality that we want to offer in the zPages). Others can implement different other functionality.
2. What are the "built-in" SpanProcessors.
  - We will provide 2 flavors of SpanProcessors (Simple/Batch) that are able to batch (or not in case of simple), transform the Spans in an "exportable" format, and call the "exporter".

PR #1137, opened 2020-10-23
Overhaul process command-line resource conventions.

### LINKED ISSUE #831: process.command_line accepts multiple value types
I noticed `process.command_line` accepts different types of values. I don't think we generally do this but was this intended? I would just use a string for it in all cases.

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/resource/semantic_conventions/process.md#process

### LINKED ISSUE #997: Reconsider `required` property for `process.*` semantic conventions
Currently we define `process.pid` and "at least one of `process.executable.name`, `process.executable.path`, `process.command`, or `process.command_line`" as required, if _any_ resource attribute in the process namespace is present. I don't think this is a sensible requirement, since there are multiple concerns (or multiple features from a backend's perspective) mixed together in the process namespace. I could well imagine that one might want to report the runtime (added in #882) but has no need for reporting the PID and/or the entire commmand line. This would currently not be "legal", however.
I think we should drop the entire `required` column for the process namespace.

This is related to #653. I do, however, think of the trace semantic conventions more as distinct features, where defining some set of attributes as essential to provide some reasonable output makes more sense than for the process resource attributes.

Oberon00:
Fixes #831, fixes #997

## Changes

* Splits process.runtime as a separate semantic convention from process (fixes #997)
* No longer require PID in any case (would also fix #997, but this is mainly for languages where the pid is not available)
* Split process.command_args from process.command_line (fixes #831).

### DIFF
--- a/specification/resource/semantic_conventions/process.md
+++ b/specification/resource/semantic_conventions/process.md
@@ -6,17 +6,34 @@
 
 | Attribute  | Description  | Example  | Required |
 |---|---|---|--|
-| process.pid | Process identifier (PID). | `1234` | Yes |
+| process.pid | Process identifier (PID). | `1234` | No |
 | process.executable.name | The name of the process executable. On Linux based systems, can be set to the `Name` in `proc/[pid]/status`. On Windows, can be set to the base name of `GetProcessImageFileNameW`. | `otelcol` | See below |
 | process.executable.path | The full path to the process executable. On Linux based systems, can be set to the target of `proc/[pid]/exe`. On Windows, can be set to the result of `GetProcessImageFileNameW`. | `/usr/bin/cmd/otelcol` | See below |
 | process.command | The command used to launch the process (i.e. the command name). On Linux based systems, can be set to the zeroth string in `proc/[pid]/cmdline`. On Windows, can be set to the first parameter extracted from `GetCommandLineW`. | `cmd/otelcol` | See below |
-| process.command_line | The full command used to launch the process. The value can be either a list of strings representing the ordered list of arguments, or a single string representing the full command. On Linux based systems, can be set to the list of null-delimited strings extracted from `proc/[pid]/cmdline`. On Windows, can be set to the result of `GetCommandLineW`. | Linux: `[ cmd/otecol, --config=config.yaml ]`, Windows: `cmd/otecol --config=config.yaml` | See below |
+| process.command_line | The full command used to launch the process as a single string representing the full command. On Windows, can be set to the result of `GetCommandLineW`. Do not set this if you have to assemble it just for monitoring; use `process.command_args` instead. | `C:\cmd\otecol --config="my directory\config.yaml"` | See below |
+| process.command_args | All the command arguments (including the command/executable itself) as received by the process. On Linux-based systems (and some other Unixoid systems supporting procfs), can be set according to the list of null-delimited strings extracted from `proc/[pid]/cmdline`. For libc-based executables, this would be the full argv vector passed to `main`. | `[ cmd/otecol, --config=config.yaml ]` | See below |
 | process.owner | The username of the user that owns the process. | `root` | No |
 | process.runtime.name | The name of the runtime of this process. For compiled native binaries, this SHOULD be the name of the compiler. | `OpenJDK Runtime Environment` | No |
 | process.runtime.version | The version of the runtime of this process, as returned by the runtime without modification. | `14.0.2` | No |
 | process.runtime.description | An additional description about the runtime of the process, for example a specific vendor customization of the runtime environment. | `Eclipse OpenJ9 openj9-0.21.0` | No |
 
-At least one of `process.executable.name`, `process.executable.path`, `process.command`, or `process.command_line` is required.
+Between `process.command_args` and `process.command_line`, usually `process.command_args` should be preferred.
+On Windows and other systems where the native format of process commands is a single string,
+`process.command_args` can additionally (or instead) be used.
+
+At least one of `process.executable.name`, `process.executable.path`, `process.command`, `process.command_line` or `process.command_args` is required.
+
+## Process runtimes
+
+**type:** `process.runtime`
+
+**Description:** The single (language) runtime instance which is monitored.
+
+| Attribute  | Description  | Example  | Required |
+|---|---|---|--|
+| process.runtime.name | The name of the runtime of this process. For compiled native binaries, this SHOULD be the name of the compiler. | `OpenJDK Runtime Environment` | No |
+| process.runtime.version | The version of the runtime of this process, as returned by the runtime without modification. | `14.0.2` | No |
+| process.runtime.description | An additional description about the runtime of the process, for example a specific vendor customization of the runtime environment. | `Eclipse OpenJ9 openj9-0.21.0` | No |
 
 `process.runtime.name` SHOULD be set to one of the values listed below, unless more detailed instructions are provided.
 If none of the listed values apply, a custom value best describing the runtime CAN be used.
--- a/specification/resource/semantic_conventions/process.md
+++ b/specification/resource/semantic_conventions/process.md
@@ -13,9 +13,6 @@
 | process.command_line | The full command used to launch the process as a single string representing the full command. On Windows, can be set to the result of `GetCommandLineW`. Do not set this if you have to assemble it just for monitoring; use `process.command_args` instead. | `C:\cmd\otecol --config="my directory\config.yaml"` | See below |
 | process.command_args | All the command arguments (including the command/executable itself) as received by the process. On Linux-based systems (and some other Unixoid systems supporting procfs), can be set according to the list of null-delimited strings extracted from `proc/[pid]/cmdline`. For libc-based executables, this would be the full argv vector passed to `main`. | `[ cmd/otecol, --config=config.yaml ]` | See below |
 | process.owner | The username of the user that owns the process. | `root` | No |
-| process.runtime.name | The name of the runtime of this process. For compiled native binaries, this SHOULD be the name of the compiler. | `OpenJDK Runtime Environment` | No |
-| process.runtime.version | The version of the runtime of this process, as returned by the runtime without modification. | `14.0.2` | No |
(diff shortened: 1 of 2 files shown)


specification/resource/semantic_conventions/process.md
```diff
@@ -6,17 +6,34 @@
...
 | process.runtime.name | The name of the runtime of this process. For compiled native binaries, this SHOULD be the name of the compiler. | `OpenJDK Runtime Environment` | No |
 | process.runtime.version | The version of the runtime of this process, as returned by the runtime without modification. | `14.0.2` | No |
 | process.runtime.description | An additional description about the runtime of the process, for example a specific vendor customization of the runtime environment. | `Eclipse OpenJ9 openj9-0.21.0` | No |
 
-At least one of `process.executable.name`, `process.executable.path`, `process.command`, or `process.command_line` is required.
+Between `process.command_args` and `process.command_line`, usually `process.command_args` should be preferred.
+On Windows and other systems where the native format of process commands is a single string,
+`process.command_args` can additionally (or instead) be used.
+
+At least one of `process.executable.name`, `process.executable.path`, `process.command`, `process.command_line` or `process.command_args` is required.
```
arminru, line 24: What if someone only wants to report the PID for whatever reason? What about:

```suggestion
At least one of `process.executable.name`, `process.executable.path`, `process.command`, `process.command_line` or `process.command_args` is required to allow back ends to identify the executable.
```

Oberon00:
@open-telemetry/technical-committee The auto-assign script seems not to be triggered when going via a draft PR.

arminru:
> @open-telemetry/technical-committee The auto-assign script seems not to be triggered when going via a draft PR.

I filed https://github.com/kentaro-m/auto-assign/issues/136.

specification/resource/semantic_conventions/process.md
Oberon00: This paragraph was already there previously, I just added the command_args alternative.

Oberon00:
@arminru 

> I filed kentaro-m/auto-assign#136.

I think this is an issue with our triggers which are only on open of PR not for open of draft or ready-for-review https://github.com/open-telemetry/opentelemetry-specification/blob/master/.github/workflows/auto-assign-tc-members.yml#L4

arminru:
> @arminru
> 
> > I filed [kentaro-m/auto-assign#136](https://github.com/kentaro-m/auto-assign/issues/136).
> 
> I think this is an issue with our triggers which are only on open of PR not for open of draft or ready-for-review [https://github.com/open-telemetry/opentelemetry-specification/blob/master/.github/workflows/auto-assign-tc-members.yml#L4](https://github.com/open-telemetry/opentelemetry-specification/blob/master/.github/workflows/auto-assign-tc-members.yml?rgh-link-date=2020-10-23T14%3A51%3A21Z#L4)

According to the Github event documentation I linked on that issue, `ready_for_review` is only triggered at the time the PR is created: _"A pull request was created that is not in draft mode."_
Let's have a try nevertheless: #1138

specification/resource/semantic_conventions/process.md
```diff
@@ -6,17 +6,38 @@
...
+| process.command_args | All the command arguments (including the command/executable itself) as received by the process. On Linux-based systems (and some other Unixoid systems supporting procfs), can be set according to the list of null-delimited strings extracted from `proc/[pid]/cmdline`. For libc-based executables, this would be the full argv vector passed to `main`. | `[ cmd/otecol, --config=config.yaml ]` | See below |
 | process.owner | The username of the user that owns the process. | `root` | No |
 | process.runtime.name | The name of the runtime of this process. For compiled native binaries, this SHOULD be the name of the compiler. | `OpenJDK Runtime Environment` | No |
 | process.runtime.version | The version of the runtime of this process, as returned by the runtime without modification. | `14.0.2` | No |
 | process.runtime.description | An additional description about the runtime of the process, for example a specific vendor customization of the runtime environment. | `Eclipse OpenJ9 openj9-0.21.0` | No |
```
arminru:
```suggestion
```

You moved these down into their own section.

Oberon00: Whoops, fixed in de9b8ec

specification/resource/semantic_conventions/process.md
```diff
@@ -6,18 +6,36 @@
...
 | process.owner | The username of the user that owns the process. | `root` | No |
+
+Between `process.command_args` and `process.command_line`, usually `process.command_args` should be preferred.
+On Windows and other systems where the native format of process commands is a single string,
+`process.command_args` can additionally (or instead) be used.
+
+For backwards compatibility with older versions of this semantic convention,
+it is possible but deprecated to use an array as type for `process.command_line`.
+In that case it MUST be interpreted as if it was `process.command_args`.
+
```
arminru, line 24: I'm not sure if we want to keep backwards compatibility here since the previous solution was a bit unclean IMHO. We don't have multiple types (string and array of strings in this case) for the same attribute anywhere else, which is also why we weren't able to pour these into YAML definitions for the markdown and constant generator tool yet.
I would be +1 for a breaking change here if we can settle on this before GA (which I hope we can). The strict trace spec freeze does not apply to semantic conventions as far as I know.

specification/resource/semantic_conventions/process.md
```diff
@@ -6,18 +6,36 @@
...
+| process.command_args | All the command arguments (including the command/executable itself) as received by the process. On Linux-based systems (and some other Unixoid systems supporting procfs), can be set according to the list of null-delimited strings extracted from `proc/[pid]/cmdline`. For libc-based executables, this would be the full argv vector passed to `main`. | `[ cmd/otecol, --config=config.yaml ]` | See below |
 | process.owner | The username of the user that owns the process. | `root` | No |
+
+Between `process.command_args` and `process.command_line`, usually `process.command_args` should be preferred.
+On Windows and other systems where the native format of process commands is a single string,
+`process.command_args` can additionally (or instead) be used.
+
+For backwards compatibility with older versions of this semantic convention,
+it is possible but deprecated to use an array as type for `process.command_line`.
+In that case it MUST be interpreted as if it was `process.command_args`.
```
Oberon00, line 23: I added this paragraph to keep the change non-breaking but I would be happy to remove it if deemed unnecessary.

arminru, line 23:
```suggestion
```
I'm not sure if we want to keep backwards compatibility here since the previous solution was a bit unclean IMHO. We don't have multiple types (string and array of strings in this case) for the same attribute anywhere else, which is also why we weren't able to pour these into YAML definitions for the markdown and constants generator tool yet. While not making interpretation impossible, I think it would still be easier for consumers and overall cleaner if we can stick to one well-defined type per attribute.

I would be +1 for a breaking change here if we can settle on this before GA (which I hope we can). The strict trace spec freeze does not apply to semantic conventions as far as I know.

arminru (approved):
LGTM. I'd prefer to remove the backwards-compatibility part as mentioned in the comment, however.

specification/resource/semantic_conventions/process.md
Oberon00: I think we would still declare command_line as non-array string in the YAML and have this text just there in the markdown as prose. We should clarify that the MUST only applies for those that choose to support this at at all (which should be a MAY).

jmacd: I will register that I don't understand why it matters if a value can take on more than one type. I didn't see a problem with `command_line` being a string or a list. I really wish I knew how people plan on using this information such that the this distinction matters. 😁

jmacd (approved):

specification/resource/semantic_conventions/process.md
Oberon00: For one, this is not supported by the current semantic convention tool, which is why this file has no corresponding YAML yet.

specification/resource/semantic_conventions/process.md
```diff
@@ -6,18 +6,36 @@
...
+| process.command_line | The full command used to launch the process as a single string representing the full command. On Windows, can be set to the result of `GetCommandLineW`. Do not set this if you have to assemble it just for monitoring; use `process.command_args` instead. | `C:\cmd\otecol --config="my directory\config.yaml"` | See below |
+| process.command_args | All the command arguments (including the command/executable itself) as received by the process. On Linux-based systems (and some other Unixoid systems supporting procfs), can be set according to the list of null-delimited strings extracted from `proc/[pid]/cmdline`. For libc-based executables, this would be the full argv vector passed to `main`. | `[ cmd/otecol, --config=config.yaml ]` | See below |
 | process.owner | The username of the user that owns the process. | `root` | No |
+
+Between `process.command_args` and `process.command_line`, usually `process.command_args` should be preferred.
+On Windows and other systems where the native format of process commands is a single string,
+`process.command_args` can additionally (or instead) be used.
```
anuraaga, line 19: command_line here I think since we talked about args just before? If not I don't quite follow the sentence.

Oberon00:
```suggestion
`process.command_line` can additionally (or instead) be used.
```

Indeed, a confusing mistake 😃

Oberon00:
@anuraaga Could you please review again? I fixed the issue you [mentioned](https://github.com/open-telemetry/opentelemetry-specification/pull/1137#discussion_r513427123).

Oberon00:
This PR formally fulfills the requirements for being merged. So I guess if you want to still review (@anuraaga?) please do so quickly.

anuraaga (approved):
So sorry for the late review - having trouble differentiating mentions from review requests in GitHub notifications and need to figure out a way to better handle them.

arminru:
@anuraaga On https://github.com/pulls ("Pull requests" in Github's title bar) there is a filter called "Mentioned". That one shows all PRs on which you were mentioned _at any point_, thus also PRs where you were only mentioned in the description or an old comment, but it's better than nothing 🙂

anuraaga:
@arminru Yeah I generally use github.com/notifications and it has a link for mentions. I need to get used to clicking it at least once a day :D

PR #1138, opened 2020-10-23
Run "Auto Assign" workflow once draft PRs are ready for review

arminru:
see https://github.com/open-telemetry/opentelemetry-specification/pull/1137#issuecomment-715378919

Event documentation:
https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request

### DIFF
--- a/.github/workflows/auto-assign-tc-members.yml
+++ b/.github/workflows/auto-assign-tc-members.yml
@@ -1,7 +1,7 @@
 name: 'Auto Assign'
 on: 
   pull_request_target:
-    types: [opened]
+    types: [opened, ready_for_review]
 
 jobs:
   add-owner:


Oberon00 (approved):

anuraaga (approved):
Nice!

carlosalberto (approved):

bogdandrutu (approved):

bogdandrutu:
@arminru can we keep only the "ready_for_review"?

arminru:
@bogdandrutu I'm not sure if the event is only triggered for PRs that transition from draft to ready or also for ones that are created as ready already, the documentation is a bit sparse. We could try but I don't see an issue with having both (unless it ends up assigning two users at once).

PR #1076, opened 2020-10-08
Database metric semantic conventions

justinfoote:
Fixes #1012

This PR attempts to specify semanatic conventions for database metric instruments.  

There are a few spicy parts that I'd love feedback about:

### 1. `db.operation` and `db.table`

We'd like to correlate the golden signal metrics to the spans that represent the same 
operation. The easiest way to do that may have been to include the `db.statement` as a 
label, but this has the potential to drastically increase cardinality.
The alternative that I see is to include `db.operation` and `db.table` as labels. 

This has clearly been discussed already, and the trace semantic convention includes
this line:

> While it would semantically make sense to set this, e.g., to a SQL keyword like `SELECT` or `INSERT`, it is not recommended to attempt any client-side parsing of `db.statement` just to get this property (the back end can do that if required).

But in my experience, these values are frequently available from an ORM library _without_
requiring SQL string parsing, and if this instrumentation may be eventually written into
those ORM libraries, I'd like to encourage them to provide this information on the 
metric instruments.

### 2. Connection pool metrics

There is some prior art in go-redis of some very detailed metrics describing database
connection pool behavior. While this is pretty interesting, I think these values will not
be widely available to all database client libraries. So I've recommended that we record
a simpler `db.connection_pool.limit` and `db.connection_pool.usage`. 

I've also included a MAY to allow instrumentation to collect more detailed information if
it is available. I'm thinking that if these concepts will be used by more than just
go-redis, it would be ideal if they looked the same.

## Changes

This PR adds a new semantic conventions at  specification/metrics/semantic_conventions/database.md

PR #1141, opened 2020-10-23
Database semantic conventions: Add db.table, allow SQL keyword as db.operation

### LINKED ISSUE #1104: Database table / collection semantic conventions

Use a database table or collection name as a dimension to search, filter, or aggregate spans by.

What did you expect to see?

An database semantic convention around `db.table` or similar. It looks like there is a mongodb specific attribute for this, `db.mongodb.collection` but this is a fairly common attribute for most databases.

**Additional context.**

There is already a PR to introduce something similar on the metrics side in #1076 as `db.table`, so it would make sense to be consistent. Many drivers / ORMs have a concept of the table / collection they are querying so it would be possible to provide those as part of those integrations. That being said, for cases where this is difficult it would make sense to recommend avoiding client-side parsing (similar to the rule for `db.operation`).

For systems where this concept doesn't exist (redis, for instance), cases where there are multiple tables in one query (sql join), or where it is difficult for the instrumentation to determine this without additional parsing, the field could be omitted and the trace storage could classify this as some default value (such as "unknown" or "other"). For these cases, `db.name` would become a much more relevant attribute.

justinfoote:
Fixes #1104

## Changes

This PR adds a new attribute to the trace semantic conventions for database operations: `db.table`, and adds guidance that this attribute should be set only if the value is readily available from the library being instrumented. 
It also changes the guidance around `db.operation` to indicate that it should be set to the SQL keyword, but also only if the value is readily available from the library being instrumented.

### DIFF
--- a/semantic_conventions/trace/database.yaml
+++ b/semantic_conventions/trace/database.yaml
@@ -191,12 +191,27 @@ groups:
             conditional: Required, if `db.statement` is not applicable.
           brief: >
             The name of the operation being executed, e.g. the [MongoDB command name](https://docs.mongodb.com/manual/reference/command/#database-operations)
-            such as `findAndModify`.
+            such as `findAndModify`, or the SQL keyword.
           note: >
-            While it would semantically make sense to set this, e.g., to a SQL keyword like `SELECT` or `INSERT`,
-            it is not recommended to attempt any client-side parsing of `db.statement` just to get this property
-            (the back end can do that if required).
-          examples: ['findAndModify', 'HMSET']
+            It is not recommended to attempt any client-side parsing of
+            `db.statement` just to get this property, but it should be set if 
+            the operation name is provided by the library being instrumented.
+            If the statement has an ambiguous operation, or performs more than
+            one operation, this value may be omitted.
+          examples: ['findAndModify', 'HMSET', 'SELECT']
+        - id: table
+          tag: call-level
+          type: string
+          required:
+            conditional: Recommended if available.
+          brief: The name of the primary table, collection, segment, etc... that the operation is acting upon, including the schema name (if applicable).
+          note: >
+            It is not recommended to attempt any  client-side parsing of
+            `db.statement` just to get this property, but it should be set if
+            it is provided by the library being instrumented.
+            If the operation is acting upon an anonymous table, or more than one
+            table, this value may be omitted.
+          examples: ['public.users', 'customers']
         - ref: net.peer.name
           tag: connection-level
           required:
--- a/semantic_conventions/trace/database.yaml
+++ b/semantic_conventions/trace/database.yaml
@@ -193,11 +193,12 @@ groups:
             The name of the operation being executed, e.g. the [MongoDB command name](https://docs.mongodb.com/manual/reference/command/#database-operations)
             such as `findAndModify`, or the SQL keyword.
           note: >
-            It is not recommended to attempt any client-side parsing of
-            `db.statement` just to get this property, but it should be set if
-            the operation name is provided by the library being instrumented.
-            If the statement has an ambiguous operation, or performs more than
-            one operation, this value may be omitted.
+            When setting this to an SQL keyword, it is not recommended to
+            attempt any client-side parsing of `db.statement` just to get this
+            property, but it should be set if the operation name is provided by
+            the library being instrumented.
+            If the SQL statement has an ambiguous operation, or performs more
+            than one operation, this value may be omitted.
           examples: ['findAndModify', 'HMSET', 'SELECT']
         - ref: net.peer.name
           tag: connection-level
--- a/semantic_conventions/trace/database.yaml
+++ b/semantic_conventions/trace/database.yaml
@@ -199,19 +199,6 @@ groups:
             If the statement has an ambiguous operation, or performs more than
             one operation, this value may be omitted.
           examples: ['findAndModify', 'HMSET', 'SELECT']
-        - id: table
-          tag: call-level
-          type: string
-          required:
-            conditional: Recommended if available.
-          brief: The name of the primary table, collection, segment, etc... that the operation is acting upon, including the schema name (if applicable).
-          note: >
-            It is not recommended to attempt any client-side parsing of
-            `db.statement` just to get this property, but it should be set if
-            it is provided by the library being instrumented.
-            If the operation is acting upon an anonymous table, or more than one table, this
-            value may be omitted.
-          examples: ['public.users', 'customers']
         - ref: net.peer.name
           tag: connection-level
           required:
@@ -309,6 +296,27 @@ groups:
             The collection being accessed within the database stated in `db.name`.
           examples: [ 'customers', 'products' ]
 
+    - id: db.sql
+      prefix: 'db.sql'
+      extends: 'db'
+      brief: >
+        Call-level attrbiutes for SQL databases
+      attributes:
+        - id: table
+          tag: call-level-tech-specific
+          type: string
+          required:
+            conditional: Recommended if available.
+          brief: The name of the primary table that the operation is acting upon, including the schema name (if applicable).
+          note: >
+            It is not recommended to attempt any client-side parsing of
+            `db.statement` just to get this property, but it should be set if
+            it is provided by the library being instrumented.
+            If the operation is acting upon an anonymous table, or more than one table, this
+            value may be omitted.
+          examples: ['public.users', 'customers']
+
+
     - id: db.tech
       brief: "Semantic convention group for specific technologies"
       constraints:
@@ -316,3 +324,4 @@ groups:
         - include: 'db.hbase'
         - include: 'db.redis'
         - include: 'db.mongodb'
+        - include: 'db.sql'
--- a/semantic_conventions/trace/database.yaml
+++ b/semantic_conventions/trace/database.yaml
@@ -314,7 +314,7 @@ groups:
             `db.statement` just to get this property, but it should be set if
             it is provided by the library being instrumented.
             If the operation is acting upon an anonymous table, or more than one table, this
-            value may be omitted.
+            value MUST NOT be set.
           examples: ['public.users', 'customers']
 
 
--- a/semantic_conventions/trace/database.yaml
+++ b/semantic_conventions/trace/database.yaml
@@ -194,7 +194,7 @@ groups:
             such as `findAndModify`, or the SQL keyword.
           note: >
             It is not recommended to attempt any client-side parsing of
-            `db.statement` just to get this property, but it should be set if 
+            `db.statement` just to get this property, but it should be set if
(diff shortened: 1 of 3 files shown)


iNikem (approved):

specification/trace/semantic_conventions/database.md
```diff
@@ -135,15 +135,18 @@ Usually only one `db.name` will be used per connection though.
...
+| `db.table` | string | The name of the primary table, collection, segment, etc... that the operation is acting upon, including the schema name (if applicable). [5] | `public.users`<br>`customers` | Conditional<br>Recommended if available. |
 
 **[1]:** In some SQL databases, the database name to be used is called "schema name".
 
 **[2]:** Required, if applicable and no more-specific attribute is defined.
 
 **[3]:** The value may be sanitized to exclude sensitive information.
 
-**[4]:** While it would semantically make sense to set this, e.g., to a SQL keyword like `SELECT` or `INSERT`, it is not recommended to attempt any client-side parsing of `db.statement` just to get this property (the back end can do that if required).
+**[4]:** It is not recommended to attempt any client-side parsing of `db.statement` just to get this property, but it should be set if  the operation name is provided by the library being instrumented. If the statement has an ambiguous operation, or performs more than one operation, this value may be omitted.
```
anuraaga, line 147:
```suggestion
**[4]:** It is not recommended to attempt any client-side parsing of `db.statement` just to get this property, but it should be set if the operation name is provided by the library being instrumented. If the statement has an ambiguous operation, or performs more than one operation, this value may be omitted.
```

anuraaga (approved):
Table is important, this seems nice

semantic_conventions/trace/database.yaml
```diff
@@ -191,12 +191,27 @@ groups:
...
+            such as `findAndModify`, or the SQL keyword.
           note: >
-            While it would semantically make sense to set this, e.g., to a SQL keyword like `SELECT` or `INSERT`,
-            it is not recommended to attempt any client-side parsing of `db.statement` just to get this property
-            (the back end can do that if required).
-          examples: ['findAndModify', 'HMSET']
+            It is not recommended to attempt any client-side parsing of
+            `db.statement` just to get this property, but it should be set if 
+            the operation name is provided by the library being instrumented.
+            If the statement has an ambiguous operation, or performs more than
```
Oberon00, line 199: This seems to imply that operation is always derived from statement, which is not true. In fact, the main purpose of this attribute is to support DBs that do not use a statement/query language as their API, such as CouchDB.

justinfoote: Yeah, I see your point. This certainly was not my intention.  

In the main "brief" description of this attribute, we still list the mongo example first. We could add a few more examples there. 

Or I wonder if it would be better to include something in the footnote like this:

```yaml
note: >
    In the case of SQL databases, the operation should be set to the SQL keyword
    (example: `SELECT` or `INSERT`). 
    It is not recommended to attempt any client-side parsing of
    `db.statement` just to get this property, but it should be set if 
    the operation name is provided by the library being instrumented.
    If the statement has an ambiguous operation, or performs more than
    one operation, this value may be omitted.
```

Is this redundant with the note in the 'required' column: "Required, if `db.statement` is not available."

What do you think?

jmacd (approved):

semantic_conventions/trace/database.yaml
```diff
@@ -191,12 +191,27 @@ groups:
...
+          tag: call-level
+          type: string
+          required:
+            conditional: Recommended if available.
+          brief: The name of the primary table, collection, segment, etc... that the operation is acting upon, including the schema name (if applicable).
+          note: >
+            It is not recommended to attempt any  client-side parsing of
+            `db.statement` just to get this property, but it should be set if
+            it is provided by the library being instrumented.
+            If the operation is acting upon an anonymous table, or more than one
```
thisthat, line 212:
```suggestion
            it is provided by the library being instrumented. If the operation
            is acting upon an anonymous table, or more than one
```
See https://github.com/open-telemetry/opentelemetry-specification/pull/1141/files#r511390753

justinfoote:
@open-telemetry/technical-committee Is this ready to merge?

semantic_conventions/trace/database.yaml
```diff
@@ -191,12 +191,27 @@ groups:
...
           brief: >
             The name of the operation being executed, e.g. the [MongoDB command name](https://docs.mongodb.com/manual/reference/command/#database-operations)
-            such as `findAndModify`.
+            such as `findAndModify`, or the SQL keyword.
           note: >
-            While it would semantically make sense to set this, e.g., to a SQL keyword like `SELECT` or `INSERT`,
-            it is not recommended to attempt any client-side parsing of `db.statement` just to get this property
-            (the back end can do that if required).
-          examples: ['findAndModify', 'HMSET']
+            It is not recommended to attempt any client-side parsing of
```
Oberon00, line 196:
```suggestion
            When setting this to an SQL keyword, it is not recommended to attempt any client-side parsing of
```

semantic_conventions/trace/database.yaml
```diff
@@ -191,12 +191,27 @@ groups:
...
+            `db.statement` just to get this property, but it should be set if
+            the operation name is provided by the library being instrumented.
+            If the statement has an ambiguous operation, or performs more than
```
Oberon00, line 199:
```suggestion
            If the SQL statement has an ambiguous operation, or performs more than
```

semantic_conventions/trace/database.yaml
Oberon00: I suggested a simpler change by emphasizing that that part only applies to SQL.

semantic_conventions/trace/database.yaml
```diff
@@ -191,12 +191,27 @@ groups:
...
+            one operation, this value may be omitted.
+          examples: ['findAndModify', 'HMSET', 'SELECT']
+        - id: table
```
Oberon00, line 202: Actually I just realized that the term `table` is probably specific to SQL. Thus I recommend adding this to a new semantic convention `db.sql` in this file that `extends: db` and has this table attribute, so that the full name is `db.sql.table`.

Oberon00 (changes requested):
Requesting changes because I have concerns for adding this SQL-specific attribute to the very general DB conventions.

semantic_conventions/trace/database.yaml
Oberon00: An alternative might be to have a more generic term, but since a table likely has no 1:1 mapping to other DB concepts (there might be several concepts that match a bit), I think having this in a SQL-specific convention would be easier.

justinfoote: This is a good idea.  I'll leave the operation where it is, but more the table to an SQL specific section in _Call-level attributes for specific technologies_.

semantic_conventions/trace/database.yaml
justinfoote: :+1: changed in ff2e13b

semantic_conventions/trace/database.yaml
justinfoote: :+1: changed in ff2e13b

semantic_conventions/trace/database.yaml
justinfoote: Done in 657670c

Oberon00 (approved):
LGTM! Thanks for responding to my comments so quickly!

justinfoote:
@open-telemetry/technical-committee I think this actually is ready to merge now.

semantic_conventions/trace/database.yaml
```diff
@@ -294,10 +297,32 @@ groups:
...
+          type: string
+          required:
+            conditional: Recommended if available.
+          brief: The name of the primary table that the operation is acting upon, including the schema name (if applicable).
+          note: >
+            It is not recommended to attempt any client-side parsing of
+            `db.statement` just to get this property, but it should be set if
+            it is provided by the library being instrumented.
+            If the operation is acting upon an anonymous table, or more than one table, this
+            value may be omitted.
```
arminru, line 317:
```suggestion
            value SHOULD be omitted.
```
We could even make this a SHOULD.

arminru (approved):

semantic_conventions/trace/database.yaml
Oberon00: Or must, as it does not make sense.

arminru:
```suggestion
            value MUST NOT be set.
```
then

justinfoote:
@bogdandrutu @jmacd -- this is the PR I mentioned in the metrics SIG today that's holding up the [database metrics semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/pull/1076).

arminru:
@justinfoote Sorry I've missed that before: Could you please add an entry in the changelog to make people aware of the new `table` and the changed `operation` attribute? Once that's done we can merge this PR right away.

justinfoote:
@arminru Added to the changelog in 862361f.

jamal:
One question about this change, I noticed that we still have `db.mongodb.collection`. Is it expected to use both or should the other be deprecated in favor of `db.table`?

arminru:
@jamal Actually we should have updated the PR name and changelog entry - it is called `db.sql.table` now and explicitly only applies to SQL-like databases and is thus _not_ a replacement for the dedicated `db.mongodb.collection`.

jamal:
@arminru Ah, I missed that, thank you. This definitely solves _some_ of the problems I wanted to solve.

An aside but being new to all of this, I'm a bit curious. Was this discussed in a specification meeting? I don't see any conversations around this in the PR, and I would have loved to be able to provide some feedback.

Having a specific solution for sql makes sense, since table is not a generic name (for example, Mongo). But, there are cases I wanted to cover as well that don't make sense. For example, Redis doesn't have a concept of table / collection. But, in our own implementation we introduced a wrapper over Lettuce to allow us to specify a collection name at the connection level. This allows us to segment metrics / traces, and I was hoping to extend our own instrumentation to tag this in traces. Obviously, this is not standard so it wouldn't be part of the spec.

But, because this change is specific to SQL, vendors / backends may not support this for Redis and allow us to provide this enhancement. I wonder if there are any thoughts on how to support cases like this?

I'll review other cases to see if I can think of other gaps I would want to make sure are covered and will open a separate issue if it comes up.

Thanks so much!

Oberon00:
> Was this discussed in a specification meeting? I don't see any conversations around this in the PR, and I would have loved to be able to provide some feedback.

There is more discussion in the linked issue #1104. I don' remember this being discussed in a meeting.

jamal:
@Oberon00 Oh, sorry, I feel silly. Totally missed that conversation when I checked the Issue again. Thanks! I'll think through this and will start a separate conversation if I have any recommendations on how this could be extended. Thanks so much!

Issue #129, opened 2019-06-15
Should API be namespaced by version or not and how exactly


tigrannajaryan:
The OpenTelemetry API can be namespaced by major version numbers, e.g. API 0.x.x is in io.opentelemetry, 1.x.x is in io.opentelemetry.v1, etc.

This allows one SDK to implement multiple major version APIs and one application to contain a mix of code that calls different API versions.

With a namespaced API SDK will typical implement lower version numbers are thin shims that map the API calls to the latest major version of API.

Actual mechanism of namespacing is language dependent, e.g. reverse domain notation for package in Java, [import paths in Go](https://github.com/golang/go/wiki/Modules#why-must-major-version-numbers-appear-in-import-paths), [namespaces](https://en.cppreference.com/w/cpp/language/namespace) in C++, etc.

We had a discussion about namespaces here: https://github.com/open-telemetry/opentelemetry-specification/pull/115#discussion_r293169930

tigrannajaryan:
For some reason I can't self-assign issues. Can someone with admin rights have a look at the permissions?

SergeyKanzhelev:
I suggest we give approvers a triager role. Any objections?

![image](https://user-images.githubusercontent.com/9950081/65057184-16bdcc00-d927-11e9-99f6-a77d0df645b4.png)

Wanshell:
can read ,clone, and push to this repository. can also manage issues and pull requests.

bogdandrutu:
@SergeyKanzhelev can you check that the CODEOWNERS works with Triage? I think we need to give them Write and based on my understanding Write include Triager

SergeyKanzhelev:
yes, my bad. every member of https://github.com/orgs/open-telemetry/teams/specs-approvers must have write permissions. Either @tigrannajaryan needs to be added to approvers or we need even broader triagers role for specs repo. I'd suggest we add Tigran to approvers

tigrannajaryan:
@SergeyKanzhelev thank you, I am able to assign issues to myself.

tedsuo:
@tigrannajaryan just followinig up on this. I believe we are not namespacing with a version number; at least, not with the initial version - who knows what OTel v2.0 would bring. If you agree it is resolved, can you close the issue?

tigrannajaryan:
Agreed, closing.

Issue #1084, opened 2020-10-09
Update Metrics Semantic Conventions README


justinfoote:
We have several open PRs related to semantic conventions:
* https://github.com/open-telemetry/opentelemetry-specification/pull/937
* https://github.com/open-telemetry/opentelemetry-specification/pull/1076
* https://github.com/open-telemetry/opentelemetry-specification/pull/1077
* https://github.com/open-telemetry/opentelemetry-specification/pull/1052

Only [one of them](https://github.com/open-telemetry/opentelemetry-specification/pull/937/files#diff-fc4d651756163e7a0dad101edc2d83e7) is attempting to update the [metrics semantic conventions README](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/semantic_conventions/README.md).  

This issue is a placeholder for us to sort out the readme once all these PRs have resolved.
This issue will be fixed once all the metrics semantic conventions have merged or closed and the readme has been updated to be coherent, and to include links to all the child semantic conventions.

justinfoote:
Also, potentially update the general timing convention (assuming it gets merged). See: https://github.com/open-telemetry/opentelemetry-specification/pull/1142#discussion_r517762323

ocelotl:
I asked Claude to review this issue and Claude suggests it be closed with this explanation:

> Issue is purely a 2020 placeholder to reconcile the metrics semantic conventions README (specification/metrics/semantic_conventions/README.md) once a set of pre-GA PRs merged. That entire semantic_conventions directory no longer exists in the spec repo (API returns 404) because all semantic conventions were migrated to open-telemetry/semantic-conventions (cf. cross-referenced tracking issue #2753, Semantic Conventions Working Group). The work now belongs in that repo, not here.

To guard against a mistaken suggestion, Claude then independently re-verified this conclusion:

> Closeable: 2020 placeholder to reconcile specification/metrics/semantic_conventions/README.md (now 404) once gating PRs resolved (#937 merged; #1076/#1077/#1052/#1142 closed); semconv migrated to open-telemetry/semantic-conventions, WG tracking #2753 CLOSED. Lean obsolete over moved-repo since no live equivalent task exists in the new repo.

Claude evaluated this issue with high confidence that it can be closed.

Claude has classified this issue as obsolete.

The intention of this comment is to make it easier for a human being to review this issue before making a decision.

ocelotl:
[...]

Issue #705, opened 2020-07-15
Standardization of Units needed


MrAlias:
Looking across language implementation the unit field for the [`metric instrument definition`](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md#metric-event-format) is implemented differently. 

In some cases it is implemented as a definite type, restricting it to well defined values:

* https://github.com/open-telemetry/opentelemetry-go/blob/master/api/unit/unit.go

But, in most cases it is an unconstrained string:

* https://github.com/open-telemetry/opentelemetry-python/blob/master/opentelemetry-sdk/src/opentelemetry/sdk/metrics/__init__.py#L151
* https://github.com/open-telemetry/opentelemetry-java/blob/master/sdk/src/main/java/io/opentelemetry/sdk/metrics/AbstractInstrument.java#L102
* https://github.com/open-telemetry/opentelemetry-js/blob/master/packages/opentelemetry-metrics/src/export/types.ts#L83

I wasn't able to find a part of the specification that specifies a unit. Given the metric SDK is still a work in progress, I'm guessing that it would eventually at least be referenced there.

However, the proto does give some specification on what values a unit can have: https://github.com/open-telemetry/opentelemetry-proto/blob/b54688569186e0b862bf7462a983ccf2c50c0547/opentelemetry/proto/metrics/v1/metrics.proto#L117-L119

This uses the [Unified Code For Units of Measure](http://unitsofmeasure.org/ucum.html).

## Questions

1. Should implementation of units be standardized similar to semantic conventions and be supported as canonical strings? Or, should they be specified as distinct types? Could they be specified so that either implementation would suffice, meaning that languages that support a strong type system could used a dedicated type, and weaker typing systems could have a more flexible implementation? 
2. Should we stick with this UCUM standards body like in the proto to specify the unit structure we expect? What about [IEEE](https://www.ewh.ieee.org/soc/ias/pub-dept/abbreviation.pdf), or [SI](https://www.nist.gov/pml/weights-and-measures/metric-si/si-units)? Should we specify our own and have a subset of these standards?
3. Should the prefix be decoupled from the unit? If the unit is to be it's own type, should the prefix just be a dedicated field?
4. How should the extensibility of the unit system be handled? Does it need to be specified?
5. Should things like `count`, `total`, or `sum` be included in the unit specification?

**Additional context.**

[...]

Issue #557, opened 2020-04-09
Span name: Both low-cardinality (grouping key) and human-readable (display name)


Oberon00:
This came up when I suggested the handler name instead of the route name for the HTTP semantic conventions (#548). I think that the handler name is a better grouping key than the route, but the route is a better display name, as @yurishkuro [pointed out](https://github.com/open-telemetry/opentelemetry-specification/pull/548#discussion_r404397569). So I think we should relieve span name from this double duty. I'll post ideas as separate comments to allow reactions.

EDIT:

Cf. current spec of span name https://github.com/open-telemetry/opentelemetry-specification/blob/e7fe34ad/specification/trace/api.md#span:

> The span name is a human-readable string which concisely identifies the work represented by the Span, for example, an RPC method name, a function name, or the name of a subtask or stage within a larger computation. The span name should be the most general string that identifies a (statistically) interesting class of Spans, rather than individual Span instances. That is, "get_user" is a reasonable name, while "get_user/314159", where "314159" is a user ID, is not a good name due to its high cardinality.

I used the term "grouping key" here because I think using the span name as a key to group / sample similar spans is the reason why a Span name should have a low cardinality.

Oberon00:
1) Add a new field "GroupingKey" to the span that defines a grouping key. The span name should be used if this is not set.

Oberon00:
2) Add a semantic convention for a span attribute `grouping_key`.

Oberon00:
3) Define a resource semantic convention `semconv.isgroupingkey:<ATTRIBUTE_NAME>` that defines a given span attribute as being viable as a grouping key (Problem: what if multiple exist? Have the value be an integer priority?).

Oberon00:
I forgot to mention this in the SIG meeting today, but this topic was talked about there, so it seems this is an actual issue.

jkwatson:
Can you define what a grouping key is? I'm not familiar with this term.

Oberon00:
@jkwatson Thank you for that question, I think I accidentally coined that term here. I edited the description to be more clear. The "grouping key" requirement can be found in the spec as a requirement for low cardinality of the span name. I think the intention here was to allow grouping / sampling per span name.

arminru:
@jkwatson You can think of the proposed grouping key as what an SQL query will look at in a "GROUP BY" clause. All spans with the same grouping key will be grouped together on the backend when being analyzed or presented in the UI.
[...]

PR #1142, opened 2020-10-23
Add semantic convention for timing durations

### LINKED ISSUE #1035: Define semantic conventions for timing metrics
Documented semantic conventions for any generic timing measurement (e.g. latency, response time, throughput) along with labels that allow faceting by e.g. error.

Semantic conventions for timing metrics documented in markdown generated from YAML definitions as per #547.

**Additional context.**
These should be roughly analogous to the trace semantic conventions for generic spans [defined here](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/span-general.md). For reference, metric semantic conventions for HTTP operations are already [defined here](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/semantic_conventions/http-metrics.md).

justinfoote:
Fixes #1035

## Changes

This PR adds semantic conventions for metrics that represent how long an operation takes to run. It includes guidance about how to incorporate tracing semantic conventions, but only in the case that the operation is already described there.

In addition to guiding the creation of metric instruments, I'm hoping that this is sort of the _base class_ for all of the other duration metric semantic conventions, and it can serve as a guide to future spec writers about our general conventions around this sort of metric.

**Question:** Does this warrant a changelog update?

Related [oteps](https://github.com/open-telemetry/oteps) [#129](https://github.com/open-telemetry/oteps/pull/129)

### DIFF
--- a/specification/metrics/semantic_conventions/timing-metrics.md
+++ b/specification/metrics/semantic_conventions/timing-metrics.md
@@ -1,0 +1,66 @@
+# General Semantic Conventions for Timed Operations
+
+<!-- Re-generate TOC with `markdown-toc --no-first-h1 -i` -->
+
+<!-- toc -->
+
+- [Overview](#overview)
+- [Duration Instrument](#duration-instrument)
+  * [Value](#value)
+  * [Metric Labels](#metric-labels)
+
+<!-- tocstop -->
+
+## Overview
+
+This document describes the metric instruments and labels to use when measuring
+the timing of any operation.
+
+This is the generic specification; there are context-specific specifications
+that follow from it for the following:
+
+* [HTTP operations](./http-metrics.md)
+* TODO: add the rest of these when they're written
+
+## Duration Instrument
+
+For each category of operation, a single `ValueRecorder` metric instrument
+should be created.
+Its name should include the name of the category, followed by the kind of
+operation being timed, like this:
+
+```
+{category}.{subcategory}.duration
+```
+
+If the duration is derived from spans that follow one of the common
+semantic-conventional _areas_, the category should be the label prefix used in
+that semantic convention, and the subcategory should be the span kind.
+
+Example names:
+
+* `http.server.duration`
+* `db.client.duration`
+* `messaging.producer.duration`
+* `rpc.client.duration`
+
+### Value
+
+The duration of each operation **in seconds** should be recorded onto this value
+recorder.
+
+### Metric Labels
+
+It is up to the implementor to annotate the duration instruments with labels
+specific to the represented operation, but care should be taken when adding
+labels to avoid excessive cardinality.
+
+If there is a corresponding span for the operation with defined trace semantic
+conventions, the labels should follow the attribute guidelines found in those
+semantic conventions, with the following caveats:
+
+1. Do not include any high-cardinality attributes, or replace them with a
+low-cardinality substitution.
+2. Replace non-string values with reasonable string representations when
+possible, and do not include attributes that cannot be easily represented as a
+string.
--- a/specification/metrics/semantic_conventions/timing-metrics.md
+++ b/specification/metrics/semantic_conventions/timing-metrics.md
@@ -59,7 +59,7 @@ If there is a corresponding span for the operation with defined trace semantic
 conventions, the labels should follow the attribute guidelines found in those
 semantic conventions, with the following caveats:
 
-1. Do not include any high-cardinality attributes, or replace them with a
+1. Avoid including any high-cardinality attributes, or replace them with a
 low-cardinality substitution.
 2. Replace non-string values with reasonable string representations when
 possible, and do not include attributes that cannot be easily represented as a
--- a/specification/metrics/semantic_conventions/README.md
+++ b/specification/metrics/semantic_conventions/README.md
@@ -2,6 +2,7 @@
 
 The following semantic conventions surrounding metrics are defined:
 
+* [Timing](timing-metrics.md): General semantic conventions for timed operations.
 * [HTTP Metrics](http-metrics.md): Semantic conventions and instruments for HTTP metrics.
 * [System Metrics](system-metrics.md): Semantic conventions and instruments for standard system metrics.
 * [Process Metrics](process-metrics.md): Semantic conventions and instruments for standard process metrics.


specification/metrics/semantic_conventions/timing-metrics.md
```diff
@@ -0,0 +1,66 @@
...
+Example names:
+
+* `http.server.duration`
+* `db.client.duration`
+* `messaging.producer.duration`
+* `rpc.client.duration`
+
+### Value
+
+The duration of each operation **in seconds** should be recorded onto this value
```
kenfinnigan, line 49: Any reason why _seconds_?

Why not milliseconds?

justinfoote: I'm following the precedent set in our system metrics like [system.cpu.time](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/semantic_conventions/system-metrics.md#systemcpu---processor-metrics). 

I see this as a placeholder until a PR lands for [standardization of units](https://github.com/open-telemetry/opentelemetry-specification/issues/705), which I expect will define a `units` label with one of the UCUM standard unit and prefix abbreviations.

jmacd: +1 @justinfoote

I was hoping that with [standardization of units](https://github.com/open-telemetry/opentelemetry-specification/issues/705) we would not have to specify which timing unit will be used. The semantics are the unchanged by the units, for data carried over OpenTelemetry or any protocol that can preserve units. 

Exporters may want to convert time units, of course. Statsd exporters would use milliseconds, conventionally. If the stopwatch measures nanoseconds, it should be OK to record that precision too. @kenfinnigan does this address your concern? If so, could we add a TODO about this?

specification/metrics/semantic_conventions/timing-metrics.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Duration Instrument
+
+For each category of operation, a single `ValueRecorder` metric instrument
+should be created.
+Its name should include the name of the category, followed by the kind of
+operation being timed, like this:
+
+```
+{category}.{subcategory}.duration
```
Oberon00, line 33: Why not use the span name as prefix here?

```suggestion
{spanname}.duration
```

It is supposed to have low cardinality.

specification/metrics/semantic_conventions/timing-metrics.md
kenfinnigan: If I'm understanding [standardization of units](https://github.com/open-telemetry/opentelemetry-specification/issues/705) correctly, there would be a "units" label that has, for instance, "seconds", and the meter would just have a value of "5"?

For me, that aligns with how Micrometer does units, and I'd be good with that

specification/metrics/semantic_conventions/timing-metrics.md
justinfoote: It would have low cardinality, and it would support correlation between spans and metrics, which is certainly a goal of mine.  There are two reasons I see. 

First, there may not be a span for the operation being timed here (though really, in this case, I think we should recommend _both_ a span and a metric for an operation whose duration is important). 

Second, I don't think the most meaningful aggregation is per-span-name. That is, I'd like to see `http.server.duration` to aggregate response time across all of my service's endpoints, and I'd like to be able to facet that graph by the HTTP method and path.

justinfoote: How would you feel about a convention to include `span.name` as a label when there is a matching span?

Oberon00: > First, there may not be a span for the operation being timed here

That is not really relevant. You can still use the name that the span tracing that operation *would have*.

> Second, I don't think the most meaningful aggregation is per-span-name.

This is something that was discussed a bit in #557, but I guess is still an open point. However, if you come up with some categorization scheme here, it would certainly also be desirable as span attribute (`span.category`?) and warrant broader discussion.

justinfoote: OK, I think we agree about the first point, but in practice, an instrumentation writer may find a scenario in which there is no span, and the spec doesn't include any semantics for such a span, but they still want to create a timing metric for the operation. I think the correct approach would be to encourage that instrumentation author to provide semantics for the sort of operation they are timing. 

I do indeed think it would be useful to have a `span.category` attribute. We have an implied category already, with separate semantic conventions for `database`, `rpc`, `faas`, etc... I think it would be worthwhile to add that explicitly to spans. 
...but, I don't want to tackle that discussion in this PR. This is about generic timing metrics. It's intended to be guidance to spec writers about how to write category-specific specs, and guidance to instrumentation authors about how to construct a timing metric for something that isn't explicitly called out in one of our category-specific specs.

How about if I create an issue to have the `span.category` discussion separately?

specification/metrics/semantic_conventions/timing-metrics.md
aabmass: I don't think a new label is needed because the unit is already a part of the instrument definition. This could effect the value type (float or int) of the instruments though, should we recommend using an int of the original precision?

> If the stopwatch measures nanoseconds, it should be OK to record that precision too.

Like in this case, prefer int nanoseconds?

aabmass (approved):

specification/metrics/semantic_conventions/timing-metrics.md
justinfoote: Yes, good point. It's part of the instrument, and it's a separate field in the proto. 

I would support the idea of allowing int nanoseconds.  But I think this is part of the larger can of worms that is #705.

specification/metrics/semantic_conventions/timing-metrics.md
```diff
@@ -0,0 +1,66 @@
...
+## Overview
+
+This document describes the metric instruments and labels to use when measuring
+the timing of any operation.
+
+This is the generic specification; there are context-specific specifications
+that follow from it for the following:
+
+* [HTTP operations](./http-metrics.md)
+* TODO: add the rest of these when they're written
```
MrAlias, line 23: Can we track this in an issue or issues?

specification/metrics/semantic_conventions/timing-metrics.md
MrAlias, line 33: Is there a situation where the `subcategory` would be omitted? Should it be optional?

MrAlias (approved):

specification/metrics/semantic_conventions/timing-metrics.md
justinfoote: Good point. I've added it to the existing issue #1084.

jmacd (approved):
This is great. I'd like to refer any remaining questions over units to https://github.com/open-telemetry/opentelemetry-specification/pull/1177.

jmacd:
There are two remaining questions associated with the use of "seconds" here that are partly addressed in #1177:
- Should it be OK to mix units with different order-of-magnitude prefixes? (Yes, but they can't be aggregated together.)
- Should it be OK to mix integer and floating-point values for the same metric name? (Open question? I'd say: Yes, but they can't be aggregated together.)

jgals:
@justinfoote and @jmacd, would one of you be able to reopen this PR? I don't think it should have been closed as inactive. 🤖

PR #1154, opened 2020-10-29
Add OpenMetrics guidance

### LINKED ISSUE #600: More prescriptive guidance on metric naming
The metric specification should include normative statements about naming metric instruments, including:

- When to use labels instead of components in a name?
- Avoid units in the name
- Use of plural vs singular (latencies or latency)?
- OpenMetrics guidance
- Name-spacing instruments.

This was one of the "TODOs" mentioned mentioned in #598.

justinfoote:
Fixes #600

This PR adds some guidelines for interoperating with OpenMetrics systems.
I have not fully specified the behavior of accepting OpenMetrics metrics -- I've delegated that to the existing design document in the collector's prometheus receiver. I am making the assumption that this is the only place we expect to accept OpenMetrics data into OpenTelemetry, and that their design document is sufficient for now.  

I have included guidelines about metric and label renaming when converting OpenTelemetry metrics to OpenMetrics. 
I have not included details about the values, or about how many OpenMetrics metrics should be created for each OpenTelemetry aggregation type. I'd love to chat more about the details here before it gets written up. 

## Changes

* Add new openmetrics-guidelines.md to metrics semantic conventions.


specification/metrics/semantic_conventions/openmetrics-guidelines.md
```diff
@@ -0,0 +1,48 @@
+# Guidance for Interoperating with OpenMetrics
+
+## Overview
```
SergeyKanzhelev, line 3: I'd suggest to add "This document is work in progress and will be updated with the OpenMetrics development" note here. Not sure if we also need to say something w.r.t. why OTel protocol is not the same as OpenMetrics.

CC: @open-telemetry/governance-committee

yurishkuro:
@jmacd ptal

MrAlias (approved):
This looks like a great start to build from in subsequent PRs. :+1:

jmacd (approved):

SergeyKanzhelev (approved):

PR #1155, opened 2020-10-30
Clarify primitive values in an array are nullable

anuraaga:
This is just an idea - it may just be me but I generally feel a primitive type isn't nullable. Does anyone else think that? Having the word nullable here makes me less confused, "why does it talk about null below even though we're talking about primitive types".

### DIFF
--- a/specification/common/common.md
+++ b/specification/common/common.md
@@ -17,7 +17,7 @@ Attributes are a list of zero or more key-value pairs. An `Attribute` MUST have 
 - The attribute key, which MUST be a non-`null` and non-empty string.
 - The attribute value, which is either:
   - A primitive type: string, boolean, double precision floating point (IEEE 754-1985) or signed 64 bit integer.
-  - An array of primitive type values. The array MUST be homogeneous,
+  - An array of nullable primitive type values. The array MUST be homogeneous,
     i.e. it MUST NOT contain values of different types. For protocols that do
     not natively support array values such values SHOULD be represented as JSON strings.
 


Oberon00:
Hmm, I think the intention here is still that `null`  should not be used even in an array if it's avoidable. I hope we never use nulls in arrays in a semantic convention at least.

anuraaga:
I see - for background, the reason I checked this spect was I was thinking of changing an API from `Boolean...` to `boolean...` to disallow `null`. But after looking at the spec in detail, I noticed `null` must be allowed. It took me a double-take though since it sounded like we are talking about primitive values. With the intention of avoiding `null` as much as possible, it's a bit unclear which is better for the API though.

Oberon00:
I think we might want some clarification here in the spec. AFAIR, the null-wording was added to make sure that parallel arrays (e.g. keys[i] mapping to values[i]) stay in sync. I don't think the intention was to require allowing null for arrays. But as the spec is now, it is indeed unclear.

carlosalberto:
cc @tigrannajaryan (who was the original author of this change, IIRC)

specification/common/common.md
```diff
@@ -17,7 +17,7 @@ Attributes are a list of zero or more key-value pairs. An `Attribute` MUST have
 - The attribute key, which MUST be a non-`null` and non-empty string.
 - The attribute value, which is either:
   - A primitive type: string, boolean, double precision floating point (IEEE 754-1985) or signed 64 bit integer.
-  - An array of primitive type values. The array MUST be homogeneous,
+  - An array of nullable primitive type values. The array MUST be homogeneous,
```
tigrannajaryan, line 20: > ... but I generally feel a primitive type isn't nullable ...

Note that the spec does not take this Java-centric view. "primitive" in this context means merely what is written in the spec:

>A primitive type: string, boolean, double precision floating point (IEEE 754-1985) or signed 64 bit integer.

That said, I agree with your clarification, since strictly speaking the current wording is contradicting itself: it prohibits "null" on lines 19-20, then allows "null" on lines 33-39.

I am not sure "nullable" is fully understandable though. Should we be more explicit, e.g.:

>An array of primitive type values or null values. The array MUST be homogeneous,
>i.e. it MUST NOT contain values of different types. The only allowed exception to this rule is that certain array values may be null, while other values are not null (see more on arrays and nulls below).

PR #1156, opened 2020-10-30
Updating gRPC status semantic conventions

### LINKED ISSUE #1044: should StatusCanonicalCode be applied for rpc?

After the merge of this PR: https://github.com/open-telemetry/opentelemetry-specification/pull/966, we can see that the `StatusCanonicalCode` got changed to `Error`, `Unset`, and `Ok`.

But, if we see the rpc status: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/rpc.md#status, they kept the same.

**Additional context.**

Should the status be the same? How should we map the rpc status to the new `StatusCanonicalCode`? For example: `Unkown` is an error or an unset status?

cc: @cijothomas @reyang

toumorokoshi:
Since the simplification of trace status codes,
they no longer align with gRPC.

Clarifying the expected status code field for gRPC, as
well as the mappings between gRPC and OpenTelemetry status.

Fixes #1044 

Related issues #

* https://github.com/open-telemetry/opentelemetry-python/pull/1308


iNikem (approved):

specification/trace/semantic_conventions/rpc.md
```diff
@@ -96,15 +96,13 @@ Note that _method_ in this context is about the called remote procedure and _not
 For remote procedure calls via [gRPC][], additional conventions are described in this section.
 
 `rpc.system` MUST be set to `"grpc"`.
+`rpc.grpc.status_code` MUST be set to the status code of the gRPC operation.
```
Oberon00, line 99: While I generally agree with that, you cannot write it like that. Please make sure a new `rpc.grpc` convention that `extends: rpc` is added to the YAML model and the property is defined there.
In doing that, you will notice that you do not say here whether this semantic convention should use the code (number) or any name of that code (string). I think it must use the integer code, because the names can vary across languages.

Oberon00 (changes requested):
Requesting changes, because it is not clear whether numeric or symbolic status code must be used (and if symbolic, which symbolic name)

semantic_conventions/trace/rpc.yaml
```diff
@@ -33,3 +33,14 @@ groups:
...
+      prefix: rpc.grpc
+      extends: rpc
+      brief: 'Call-level attributes for gRPC'
+      attributes:
+        - id: status_code
+          type: number
+          required: always
+          tag: call-level-tech-specific
+          brief: "the status code number of the gRPC request."
+          examples: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16]
```
Oberon00, line 46: OK, unfortunately the examples are kinda pointless here, but since they are required, maybe use a bit fewer at least.

```suggestion
          examples: [0, 1, 16]
```

semantic_conventions/trace/rpc.yaml
```diff
@@ -33,3 +33,14 @@ groups:
           - net.peer.ip
           - net.peer.name
         - include: network
+    - id: rpc.grpc
+      prefix: rpc.grpc
+      extends: rpc
+      brief: 'Call-level attributes for gRPC'
+      attributes:
+        - id: status_code
+          type: number
```
Oberon00, line 42: Hmm, this would logically be an integer enum but I don't think we support that yet, @thisthat ?

specification/trace/semantic_conventions/rpc.md
```diff
@@ -96,15 +96,13 @@ Note that _method_ in this context is about the called remote procedure and _not
 For remote procedure calls via [gRPC][], additional conventions are described in this section.
 
 `rpc.system` MUST be set to `"grpc"`.
+`rpc.grpc.status_code` MUST be set to the status code number of the gRPC operation.
```
Oberon00, line 99: This should now be replaced by the generated attribute table for grpc:

```suggestion

### Attributes

<!-- semconv rpc.grpc -->
<!-- endsemconv -->
```

And then run `make table-generation` (see https://github.com/open-telemetry/opentelemetry-specification/tree/master/semantic_conventions#generation)

semantic_conventions/trace/rpc.yaml
```diff
@@ -33,3 +33,14 @@ groups:
...
+          required: always
+          tag: call-level-tech-specific
```
Oberon00, line 44: Since you never filter by tag in the markdown, I think you don't need this.

semantic_conventions/trace/rpc.yaml
```diff
@@ -33,3 +33,14 @@ groups:
...
+          brief: "the status code number of the gRPC request."
```
Oberon00, line 45:
```suggestion
          brief: "The [numeric status code](https://github.com/grpc/grpc/blob/v1.33.2/doc/statuscodes.md) of the gRPC request."
```

Oberon00 (commented):
Thank you for the update! 😃  Just a few smaller bits & pieces left before I think this is ready.

Were you able to find the documentation for the markdown tool? If not, where would you have looked? Maybe we can improve the situation here by placing more links to it in different places?

semantic_conventions/trace/rpc.yaml
```diff
@@ -33,3 +33,14 @@ groups:
           - net.peer.ip
           - net.peer.name
         - include: network
+    - id: rpc.grpc
+      prefix: rpc.grpc
+      extends: rpc
+      brief: 'Call-level attributes for gRPC'
```
Oberon00, line 39: Why Call-level?

```suggestion
      brief: 'Tech-specific attributes for gRPC.'
```

semantic_conventions/trace/rpc.yaml
toumorokoshi: not strongly opinionated, can do. I figured since it was short might as well enumerate all values.

specification/trace/semantic_conventions/rpc.md
anuraaga: I guess if we replace the MD we lose the other attributes for measage. Should we move those into the yml too?

semantic_conventions/trace/rpc.yaml
Oberon00: No strong opinion either. Fine for me to keep as-is.

specification/trace/semantic_conventions/rpc.md
Oberon00: Which do you mean? The message attributes are still literally in the markdown after my suggestion.

The generator does not really support events yet.

semantic_conventions/trace/rpc.yaml
toumorokoshi: I didn't see any... https://github.com/open-telemetry/build-tools/issues/19

semantic_conventions/trace/rpc.yaml
toumorokoshi: I was considering adding tags to make them consistent with what I've seen the in the YAML, but can remove.

Oberon00 (approved):
Thank you!

semantic_conventions/trace/rpc.yaml
toumorokoshi: for posterity, I did go with the suggestion because otherwise the generated table results in a very tall row (due to each example being on a different row).

specification/trace/semantic_conventions/rpc.md
```diff
@@ -97,14 +97,19 @@ For remote procedure calls via [gRPC][], additional conventions are described in
 
 `rpc.system` MUST be set to `"grpc"`.
 
+### Attributes
```
arminru, line 100: Please run `markdown-toc` or add it manually above

specification/trace/semantic_conventions/rpc.md
```diff
@@ -97,14 +97,19 @@ For remote procedure calls via [gRPC][], additional conventions are described in
...
+
 [gRPC]: https://grpc.io/
 
 ### Status
 
-Implementations MUST set status which MUST be the same as the gRPC client/server
-status. The mapping between gRPC canonical codes and OpenTelemetry status codes
-is 1:1 as OpenTelemetry canonical codes is just a snapshot of grpc codes which
-can be found [here](https://github.com/grpc/grpc-go/blob/master/codes/codes.go).
+The [Span Status](../api.md#set-status) MUST be left unset for an OK gRPC status code, and `Error` for all others.
```
arminru, line 112:
```suggestion
The [Span Status](../api.md#set-status) MUST be left unset for an `OK` gRPC status code, and set to `Error` for all others.
```

arminru (approved):

specification/trace/semantic_conventions/rpc.md
toumorokoshi: done! the headings ended up colliding so I renamed them.

semantic_conventions/trace/rpc.yaml
thisthat: Sorry for the late response :sweat_smile: 
It is already possible to declare an integer enum with the current tool :sunglasses: 
```suggestion
          type:
            allow_custom_values: false
            members:
              - id: OK
                value: 0
              - id: CANCELLED
                value: 1
              ....
```

thisthat (approved):

semantic_conventions/trace/rpc.yaml
SergeyKanzhelev: let's add this?

Oberon00 (approved):

Oberon00 (approved):

SergeyKanzhelev (approved):

SergeyKanzhelev:
No semantical change for 2 days and enough approvals. Merging

Issue #1158, opened 2020-10-30
Complete the Metric SDK specification


jmacd:
**What are you trying to achieve?**

The metric SDK specification is unfinished. There are a number of TODOs in https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/sdk.md

These have to be addressed before we can finish the SDK compliance matrix for metrics.

hdost:
Is there a permalink to the SDK?

SergeyKanzhelev:
> Is there a permalink to the SDK?

Answering the question: https://github.com/open-telemetry/opentelemetry-specification/blob/135d99df75beb66836673e0fadc9fd5c67abef03/specification/metrics/sdk.md (just before this PR: https://github.com/open-telemetry/opentelemetry-specification/pull/1401/)

However, if you want to work on it, please make sure to join the discussion of metrics API group. Easiest to join the meeting (see OpenTelemetry calendar), but also can reach out to @reyang for details on how to get involved

reyang:
I'm closing this issue since it is too broad. We'll be using [milestones](https://github.com/open-telemetry/opentelemetry-specification/milestones) in this repo to track the items.

PR #1159, opened 2020-10-30
Metric SDK: add asynchronous instrument details; add export pipeline terminology details

jmacd:
## Changes

Fills in two missing parts of the metric SDK specification, with terminology taken from the OTel-Go implementation and requirements gathered from the API specification.

Part of #1158.


jkwatson:
This all seems reasonable; I don't know how to map it to the Java implementation, unfortuntely.

jmacd:
@carlosalberto has begun auditing and coming up to speed on this implementation, thankfully. 😀

specification/metrics/sdk.md
```diff
@@ -452,9 +476,47 @@ considered language-specific details.
 
 ### Export pipeline detail
 
-TODO: define AggregatorSelector, Aggregator, Accumulation, ExportKind,
-ExportKindSelector, Aggregation, AggregationKind ExportRecord,
-ExportRecordSet
+#### Export pipeline data types
+
+**Accumulation**: Passed from the Accumulator to the Processor, this
+consists of the Instrument, the Label Set, the Resource, and a
```
MrAlias, line 482: This should only contain a description of a metric instrument, not the instrument itself, right?

```suggestion
consists of the Metric Descriptor, the Label Set, the Resource, and a
```

specification/metrics/sdk.md
```diff
@@ -452,9 +476,47 @@ considered language-specific details.
...
+**AggregationKind**: This is determined by the Aggregation, which is
+returned by an Aggregator, chosen by an AggregatorSelector.
+Aggregation kinds determine what data can be read from an Aggregation
+(i.e., the format).
+
+**ExportKind**: This is determined by the ExportKindSelector and may
+be cumulative or delta.  Note that the term ExportKind is used in the
+SDK to refer to this choice, while the same concept is called
+AggregationTemporality when stored as a field in the OpenTelemetry
+protocol.  TODO: rename ExportKind to AggregationTemporality?
```
MrAlias, line 508: I like the idea of consistency :+1:

specification/metrics/sdk.md
```diff
@@ -452,9 +476,47 @@ considered language-specific details.
...
+
+**ExportRecordSet**: The set of ExportRecords computed during a
+Processor checkpoint interval.
+
+#### Export pipeline configuration
+
+**AggregationKind**: This is determined by the Aggregation, which is
+returned by an Aggregator, chosen by an AggregatorSelector.
+Aggregation kinds determine what data can be read from an Aggregation
+(i.e., the format).
```
MrAlias, line 502: Could we describe these `*Kind` definition more by what they are instead of what determines them and what calls them?

Possibly:

```suggestion
**AggregationKind**: The format which an Aggregator aggregates data into and ultimately sends it as.
```

MrAlias (approved):

specification/metrics/sdk.md
```diff
@@ -266,12 +269,33 @@ collection period while the current one is processed.  An Accumulation
...
 
 ### Processor
 
 TODO _Processor functional requirements_
 
+### Processor checkpoint interval
+
+The Processor checkpoint interval brackets Accumulations into a single
+unit of data.  The Processor checkpoint interval starts and finishes
+before and after calling Accumulator.Collect to process Acumulations
```
jkwatson, line 295:
```suggestion
unit of data.  The Processor checkpoint interval starts before and finishes
after calling Accumulator.Collect to process Acumulations
```

jkwatson: FYI, I'm very interested to see the Processor spec finished. I don't have a good idea of why "start" is necessary on the Processor and how its supposed to react to that call.

jmacd: This is a good point. Linking these two conversations together here: https://github.com/open-telemetry/opentelemetry-specification/pull/1198#discussion_r527802577

@jkwatson you wrote:

> why not have the controller pass this data to the processor, and decouple the accumulator from the processor?

See also https://github.com/open-telemetry/opentelemetry-go/issues/1362, where this coupling makes trouble for setting up the SDK (FYI @seanschade). Thinking in terms of the OTel-Go implementation, it would make sense for Collect() from the Accumulator to write into a channel and for the Processor to read from a channel.

I can imagine in other languages it would be more natural to use an iterator pattern to iterate over the results of collection using a `ForEach()` pattern, the way exporters consume the output of the `Processor` (although it means executing asynchronous instrument callbacks during the iteration). It won't be easy to do such a refactoring in Go because of how map iteration works--I'll try to make this explanation allow more language-specific approach and encourage decoupling, then file issues to fix the OTel-Go implementation.

PR #1160, opened 2020-10-30
Specify process.runtime for Python.

Oberon00:
Fixes open-telemetry/opentelemetry-python#1127

## Changes

Specifies `process.runtime.*` attributes for Python.

CC @open-telemetry/python-approvers

### DIFF
--- a/specification/resource/semantic_conventions/process.md
+++ b/specification/resource/semantic_conventions/process.md
@@ -120,9 +120,14 @@ Examples for some Python runtimes:
 | CPython 3.7.3 on Windows | cpython | 3.7.3 | 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 22:22:05) [MSC v.1916 64 bit (AMD64)] |
 | CPython 3.8.6 on Linux | cpython | 3.8.6 | 3.8.6 (default, Sep 30 2020, 04:00:38) <br>[GCC 10.2.0] |
 | PyPy 3 7.3.2 on Linux | pypy | 3.7.4 | 3.7.4 (?, Sep 27 2020, 15:12:26)<br>[PyPy 7.3.2-alpha0 with GCC 10.2.0] |
+| Jython 2.7.2 on Linux | jython | 2.7.2 | 2.7.2 (v2.7.2:925a3cc3b49d, Mar 21 2020, 10:03:58)<br>[Java HotSpot(TM) 64-Bit Server VM (Oracle Corporation)] |
 
 Note that on Linux, there is an actual newline in the `sys.version` string,
 and the CPython string had a trailing space in the first line.
+
+Pypy provided a CPython-compatible version in `sys.implementation.version` instead of the actual implementation version which is available in `sys.version`.
+
+Jython does not support Python 3, so there the fallbacks without `sys.implementation` were used.
 
 ***Ruby Runtimes:***
 
--- a/specification/resource/semantic_conventions/process.md
+++ b/specification/resource/semantic_conventions/process.md
@@ -103,7 +103,7 @@ Python instrumentation should fill in the values as follows:
   ```python
   vinfo = sys.implementation.version
   result = '.'.join(map(str, vinfo[:3])) + (
-      '' if vinfo.releaselevel == "final" and not vinfo.serial
+      '' if vinfo.releaselevel == 'final' and not vinfo.serial
       else vinfo.releaselevel + '.' + str(vinfo.serial))
   ```
 
--- a/specification/resource/semantic_conventions/process.md
+++ b/specification/resource/semantic_conventions/process.md
@@ -88,16 +88,41 @@ TODO(<https://github.com/open-telemetry/opentelemetry-dotnet/issues/1281>): Conf
 
 ***Python Runtimes:***
 
-TODO(<https://github.com/open-telemetry/opentelemetry-python/issues/1127>): Confirm the contents here
+Python instrumentation should fill in the values as follows:
 
-| Value | Description |
-| --- | --- |
-| `cpython` | CPython |
-| `graalvm` | GraalVM |
-| `ironpython` | IronPython |
-| `jython` | Jython |
-| `pypy` | PyPy|
-| `pythonnet` | PythonNet |
+- `process.runtime.name` -
+  Fill in the value of [`sys.implementation.name`][py_impl]
+  (fall back to [`platform.python_implementation().lower()`](https://docs.python.org/3/library/platform.html#platform.python_implementation) if `sys.implementation` is not available)
+- `process.runtime.version` -
+  Fill in the [`sys.implementation.version`][py_impl] values separated by dots.
+  Fall back to using [`sys.version_info`](https://docs.python.org/3/library/sys.html#sys.version_info) if that is not available.
+  Leave out the release level and serial if the release level
+  equals `final` and the serial equals zero
+  (leave out either both or none).
+
+  This can be implemented with the following Python snippet:
+
+  ```python
+  vinfo = sys.implementation.version
+  result = '.'.join(map(str, vinfo[:3])) + (
+      '' if vinfo.releaselevel == "final" and not vinfo.serial
+      else vinfo.releaselevel + str(vinfo.serial))
+  ```
+
+- `process.runtime.description` - Fill in the value of [`sys.version`](https://docs.python.org/3/library/sys.html#sys.version) as-is.
+
+[py_impl]: https://docs.python.org/3/library/sys.html#sys.implementation
+
+Examples for some Python runtimes:
+
+| Name | `process.runtime.name` | `process.runtime.version` | `process.runtime.description` |
+| --- | --- | --- | --- |
+| CPython 3.7.3 on Windows | cpython | 3.7.3 | 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 22:22:05) [MSC v.1916 64 bit (AMD64)] |
+| CPython 3.8.6 on Linux | cpython | 3.8.6 | 3.8.6 (default, Sep 30 2020, 04:00:38) <br>[GCC 10.2.0] |
+| PyPy 3 7.3.2 on Linux | pypy | 3.7.4 | 3.7.4 (?, Sep 27 2020, 15:12:26)<br>[PyPy 7.3.2-alpha0 with GCC 10.2.0] |
+
+Note that on Linux, there is an actual newline in the `sys.version` string,
+and the CPython string had a trailing space in the first line.
 
 ***Ruby Runtimes:***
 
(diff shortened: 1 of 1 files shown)


mariojonke (approved):

specification/resource/semantic_conventions/process.md
```diff
@@ -88,16 +88,46 @@ TODO(<https://github.com/open-telemetry/opentelemetry-dotnet/issues/1281>): Conf
...
+| --- | --- | --- | --- |
+| CPython 3.7.3 on Windows | cpython | 3.7.3 | 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 22:22:05) [MSC v.1916 64 bit (AMD64)] |
+| CPython 3.8.6 on Linux | cpython | 3.8.6 | 3.8.6 (default, Sep 30 2020, 04:00:38) <br>[GCC 10.2.0] |
+| PyPy 3 7.3.2 on Linux | pypy | 3.7.4 | 3.7.4 (?, Sep 27 2020, 15:12:26)<br>[PyPy 7.3.2-alpha0 with GCC 10.2.0] |
+| Jython 2.7.2 on Linux | jython | 2.7.2 | 2.7.2 (v2.7.2:925a3cc3b49d, Mar 21 2020, 10:03:58)<br>[Java HotSpot(TM) 64-Bit Server VM (Oracle Corporation)] |
+
+Note that on Linux, there is an actual newline in the `sys.version` string,
+and the CPython string had a trailing space in the first line.
+
+Pypy provided a CPython-compatible version in `sys.implementation.version` instead of the actual implementation version which is available in `sys.version`.
```
Oberon00, line 128: This is a bit sad. We could another special case for PyPy to the runtime.version determination to use

```
>>>> sys.pypy_version_info
(major=7, minor=3, micro=2, releaselevel='alpha', serial=0)
```

specification/resource/semantic_conventions/process.md
```diff
@@ -88,16 +88,46 @@ TODO(<https://github.com/open-telemetry/opentelemetry-dotnet/issues/1281>): Conf
...
-| `ironpython` | IronPython |
-| `jython` | Jython |
-| `pypy` | PyPy|
-| `pythonnet` | PythonNet |
+- `process.runtime.name` -
+  Fill in the value of [`sys.implementation.name`][py_impl]
+  (fall back to [`platform.python_implementation().lower()`](https://docs.python.org/3/library/platform.html#platform.python_implementation) if `sys.implementation` is not available)
+- `process.runtime.version` -
+  Fill in the [`sys.implementation.version`][py_impl] values separated by dots.
+  Fall back to using [`sys.version_info`](https://docs.python.org/3/library/sys.html#sys.version_info) if that is not available.
```
aabmass, line 98: I think we may need an additional field for the version of the _language_ (which defines the syntax and standard library API) after reading the docs for [`sys.implementation`](https://docs.python.org/3/library/sys.html#sys.implementation):

<blockquote>
<p><em>version</em> is a named tuple, in the same format as
<a class="reference internal" href="#sys.version_info" title="sys.version_info"><code class="xref py py-data docutils literal notranslate"><span class="pre">sys.version_info</span></code></a>.  It represents the version of the Python
<em>implementation</em>.  This has a distinct meaning from the specific
version of the Python <em>language</em> to which the currently running
interpreter conforms, which <code class="docutils literal notranslate"><span class="pre">sys.version_info</span></code> represents.  For
example, for PyPy 1.8 <code class="docutils literal notranslate"><span class="pre">sys.implementation.version</span></code> might be
<code class="docutils literal notranslate"><span class="pre">sys.version_info(1,</span> <span class="pre">8,</span> <span class="pre">0,</span> <span class="pre">'final',</span> <span class="pre">0)</span></code>, whereas <code class="docutils literal notranslate"><span class="pre">sys.version_info</span></code>
would be <code class="docutils literal notranslate"><span class="pre">sys.version_info(2,</span> <span class="pre">7,</span> <span class="pre">2,</span> <span class="pre">'final',</span> <span class="pre">0)</span></code>.  For CPython they
are the same value, since it is the reference implementation.</p>
</blockquote>

For CPython they are the same but for PyPy they are independent.

specification/resource/semantic_conventions/process.md
aabmass: Sounds reasonable since PyPy is not conforming to the docs here

specification/resource/semantic_conventions/process.md
```diff
@@ -88,16 +88,46 @@ TODO(<https://github.com/open-telemetry/opentelemetry-dotnet/issues/1281>): Conf
...
+| CPython 3.8.6 on Linux | cpython | 3.8.6 | 3.8.6 (default, Sep 30 2020, 04:00:38) <br>[GCC 10.2.0] |
+| PyPy 3 7.3.2 on Linux | pypy | 3.7.4 | 3.7.4 (?, Sep 27 2020, 15:12:26)<br>[PyPy 7.3.2-alpha0 with GCC 10.2.0] |
+| Jython 2.7.2 on Linux | jython | 2.7.2 | 2.7.2 (v2.7.2:925a3cc3b49d, Mar 21 2020, 10:03:58)<br>[Java HotSpot(TM) 64-Bit Server VM (Oracle Corporation)] |
+
+Note that on Linux, there is an actual newline in the `sys.version` string,
+and the CPython string had a trailing space in the first line.
+
+Pypy provided a CPython-compatible version in `sys.implementation.version` instead of the actual implementation version which is available in `sys.version`.
+
+Jython does not support Python 3, so there the fallbacks without `sys.implementation` were used.
```
aabmass, line 130: Same comment as above. Would it simplify things to remove it?

specification/resource/semantic_conventions/process.md
```diff
@@ -88,16 +88,46 @@ TODO(<https://github.com/open-telemetry/opentelemetry-dotnet/issues/1281>): Conf
...
+[py_impl]: https://docs.python.org/3/library/sys.html#sys.implementation
+
+Examples for some Python runtimes:
+
+| Name | `process.runtime.name` | `process.runtime.version` | `process.runtime.description` |
+| --- | --- | --- | --- |
+| CPython 3.7.3 on Windows | cpython | 3.7.3 | 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 22:22:05) [MSC v.1916 64 bit (AMD64)] |
+| CPython 3.8.6 on Linux | cpython | 3.8.6 | 3.8.6 (default, Sep 30 2020, 04:00:38) <br>[GCC 10.2.0] |
+| PyPy 3 7.3.2 on Linux | pypy | 3.7.4 | 3.7.4 (?, Sep 27 2020, 15:12:26)<br>[PyPy 7.3.2-alpha0 with GCC 10.2.0] |
+| Jython 2.7.2 on Linux | jython | 2.7.2 | 2.7.2 (v2.7.2:925a3cc3b49d, Mar 21 2020, 10:03:58)<br>[Java HotSpot(TM) 64-Bit Server VM (Oracle Corporation)] |
```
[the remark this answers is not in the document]
Oberon00: I could add a note advising users of that. But just showing different versions of CPython 3 would be a bit boring 😃

specification/resource/semantic_conventions/process.md
Oberon00: For PyPy they are also the same, unfortunately (see note below). We could add additional fields for that, true, maybe python.runtime.lang_version. But I did not want to add new fields in this PR.

specification/resource/semantic_conventions/process.md
Oberon00: It would simplify things. If we are OK with losing spec on Python < 3.3, I can remove the wording and the example.
I think Jython might be an interesting example for a runtime on top of another runtime, but maybe it is not worth it.

specification/resource/semantic_conventions/process.md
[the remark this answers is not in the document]
Oberon00: I added fc4d621, that should resolve this by removing Jython on thus all the examples are 3.5+.

specification/resource/semantic_conventions/process.md
Oberon00: I added fc4d621 to remove considerations for Python < 3.3

specification/resource/semantic_conventions/process.md
```diff
@@ -88,16 +88,41 @@ TODO(<https://github.com/open-telemetry/opentelemetry-dotnet/issues/1281>): Conf
...
+  equals `final` and the serial equals zero
+  (leave out either both or none).
+
+  This can be implemented with the following Python snippet:
+
+  ```python
+  vinfo = sys.implementation.version
+  result = '.'.join(map(str, vinfo[:3])) + (
+      '' if vinfo.releaselevel == "final" and not vinfo.serial
+      else vinfo.releaselevel + str(vinfo.serial))
```
Oberon00, line 107: I just noticed we are missing a dot here according to the spec above.

```suggestion
      else vinfo.releaselevel + '.' + str(vinfo.serial))
```

specification/resource/semantic_conventions/process.md
aabmass: Ya, that is funny considering the quote from the python docs no longer holds true for PyPy 😕 . But could this effect other languages and would be worth adding `process.runtime.language_version` or similar?

Oberon00: I think that would be useful, but I think it is out of scope for this PR.

The new attribute might better be Python-specific though. For example, I don't think it makes much sense for JVMs or .NET runtimes which support multiple languages.

specification/resource/semantic_conventions/process.md
```diff
@@ -88,16 +88,44 @@ TODO(<https://github.com/open-telemetry/opentelemetry-dotnet/issues/1281>): Conf
...
+  This can be implemented with the following Python snippet:
+
+  ```python
+  vinfo = sys.implementation.version
+  result =  ".".join(map(
+      str,
+      vinfo[:3]
+      if vinfo.releaselevel == "final" and not vinfo.serial
+      else vinfo
+  ))
```
Oberon00, line 110: This should be OK now. Example result might be

```python
In [1]: from packaging import version

In [2]: version.parse('3.9.1.alpha.3')
Out[2]: <Version('3.9.1a3')>
```

so it is still a correct version.

lzchen (approved):
Nice!

arminru (approved):
Rubber-stamping since we have approvals from an OTel-Python maintainer and an approver

carlosalberto (approved):

carlosalberto:
@codeboten Any opinion/approval?

specification/resource/semantic_conventions/process.md
Oberon00: I guess it is too late to change this in this PR now, unless there is a strong opinion. @open-telemetry/python-approvers feel free to suggest a follow-up spec PR if this comes up again when implementing.

Issue #855, opened 2020-08-21
Proposal: Dictionary of common Attribute/Label definitions


justinfoote:
Hey all,
In a [recent PR](https://github.com/open-telemetry/opentelemetry-specification/pull/739) regarding metric semantic conventions, we repeatedly diverted the conversation to these topics:
* There's a lot of duplication between the labels in the HTTP metric semantic convention and the attributes in the HTTP tracing semantic convention. 
* If we remove that duplication and reference the tracing convention from the metrics convention, it implies some connection between the underlying span and metric data structures.
* Labels and attributes have different restrictions on value types.  Does it even make sense to use the same key for a metric label and a span attribute if their value is different (for example, http.response_code of `"200"` on a label or `200` on an attribute).

I propose that we solve this issue by creating a single data dictionary that can be referenced from all semantic conventions.  This will clarify what each attribute name signifies, so we don't have to redefine the same names in multiple places.

We use a system like this at New Relic.  We've even made this data dictionary publicly-available and searchable.  Check it out here: https://docs.newrelic.com/attribute-dictionary?attribute_name=traceId

In this example, because the definition of `traceId` exists in a single place, when it is included in a spec for Spans, Transactions, and Logs, each spec does not need to include a full definition of the attribute.

At the moment, OpenTelemetry has defined traceId in at least two places:
* [Logs](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/logs/data-model.md#field-traceid): 
> Field: TraceId
>
> Type: byte sequence.
>
> Description: Request trace id as defined in W3C Trace Context. Can be set for logs that are part of request processing and have an assigned trace id. This field is optional.
* [SpanContext](https://github.com/open-telemetry/opentelemetry-specification/blob/b409fb1833cbb90e19969d82a1209fb31e976a0d/specification/overview.md#spancontext):
> TraceId is the identifier for a trace. It is worldwide unique with practically sufficient probability by being made as 16 randomly generated bytes. TraceId is used to group all spans for a specific trace together across all processes.

I propose something like this, in a big table of all attributes:

| Name | Description | Serialization | Example |
| ---- | ----------- | ------------- | ------- |
| traceId | The identifier for a trace. It is worldwide unique with practically sufficient probability by being made as 16 randomly generated bytes. TraceId is used to group all spans for a specific trace together across all processes, and to correlate other data points to those spans. | Byte Sequence       |   `8ac95d01265752b841aa418a0ed4cb48` |

[...]

PR #1162, opened 2020-10-30
Provide semantic conventions for RPC metrics

### LINKED ISSUE #1016: Define metric semantic conventions for RPC
Documented metric semantic conventions for remote procedure calls (RPC).

Metric semantic conventions for RPC documented in markdown generated from YAML definitions as per #547.

**Additional context.**
These should be roughly analogous to the trace semantic conventions for RPC [defined here](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/rpc.md). For reference, metric semantic conventions for HTTP operations are already [defined here](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/semantic_conventions/http-metrics.md).

jsuereth:
Fixes #1016 

## Changes

Adds semantic conventions for RPC metrics.
- Re-use existing RPC attribute spec YAML.  Note: this interacts oddly pulling from trace => metrics.
- Define 3 (x 2) key metrics for RPC:   `rpc.?.duration`, `rpc.?.request_size` and `rpc.?.response_size`, crossed by `server` or `client`
- Add warning around reducing cardinality in attributes for metrics (cross-reference the same information for spans).

Deviations from [OpenCensus](https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/gRPC.md)
- This does not track a "count" of messages that have been issued.
- This does not pull different metrics per-method vs. per-rpc, aggregation across labels is assumed.


specification/metrics/semantic_conventions/rpc.md
```diff
@@ -0,0 +1,82 @@
...
+
+*Note: RPC server and client metrics are split to allow correlation across client/server boundaries, e.g. Lining up an RPC method latency to determine if the server is responsible for latency the client is seeing.*
+
+### RPC Server
+
+Below is a table of RPC server metric instruments.
+
+| Name                       | Instrument    | Units        | Description | Status |
+|----------------------------|---------------|--------------|-------------|--------|
+| `rpc.server.duration`      | ValueRecorder | milliseconds | measures duration of inbound RPC | Recommended |
```
MrAlias, line 24: Sorry for the ignorance, but is this a timing of when the request is received to when a response is sent? How would this be used for a streaming RPC?

specification/metrics/semantic_conventions/rpc.md
```diff
@@ -0,0 +1,82 @@
...
+
+Below is a table of labels that SHOULD be included on metric events and whether
+or not they should be on the server, client or both.
+
+<!-- semconv rpc -->
+| Attribute  | Type | Description  | Example  | Required |
+|---|---|---|---|---|
+| [`rpc.system`](../../trace/semantic_conventions/rpc.md) | string | A string identifying the remoting system. | `grpc`<br>`java_rmi`<br>`wcf` | Yes |
+| [`rpc.service`](../../trace/semantic_conventions/rpc.md) | string | The full name of the service being called, including its package name, if applicable. | `myservice.EchoService` | Conditional<br>No, but recommended |
+| [`rpc.method`](../../trace/semantic_conventions/rpc.md) | string | The name of the method being called, must be equal to the $method part in the span name. | `exampleMethod` | Conditional<br>No, but recommended |
```
MrAlias, line 48: Guessing this was copy/pasted, but we probably do not want to include a reference to a span name here.

```suggestion
| [`rpc.method`](../../trace/semantic_conventions/rpc.md) | string | The name of the method being called. | `exampleMethod` | Conditional<br>No, but recommended |
```

justinfoote (approved):
This looks good to me. 
My only question is: should the rpc metrics also include HTTP labels if the RPC calls are happening over HTTP?

specification/metrics/semantic_conventions/rpc.md
jsuereth: This is actually how the markdown-generation tool works.   This opens a question on whether we should SHARE rpc attributes between Trace/Metrics.  Right now Trace owns the single-id `rpc` so I just re-used it here because the attributes SHOULD be the same (in my opinion).    However, if we change the documentation for one, we change both (the downside).

So my question to you:

1. Should the attributes be shared?
2. Should I update the tool which generates this markdown from YAML to allow different descriptions?

specification/metrics/semantic_conventions/rpc.md
```diff
@@ -0,0 +1,82 @@
...
+
+### RPC Server
+
+Below is a table of RPC server metric instruments.
+
+| Name                       | Instrument    | Units        | Description | Status |
+|----------------------------|---------------|--------------|-------------|--------|
+| `rpc.server.duration`      | ValueRecorder | milliseconds | measures duration of inbound RPC | Recommended |
+| `rpc.server.request.size`  | ValueRecorder | bytes        | measures size of RPC request messages | Optional |
+| `rpc.server.response.size` | ValueRecorder | bytes        | measures size of RPC response messages | Optional |
```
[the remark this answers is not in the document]
jsuereth: Yeah, meant to have these be uncompressed with compressed size as an optional you can also use.

specification/metrics/semantic_conventions/rpc.md
jsuereth: This is the time from when the request was received to when the response is set.

## Streaming RPCs
This likely needs a callout for how to handle streaming RPCs/understand them.

Initially, my thinking is this would be the duration to complete the stream: https://grpc.io/docs/what-is-grpc/core-concepts/#server-streaming-rpc which doesn't apply in the bi-directional streaming.

I'll update duration to denote its applicability to streaming RPCs and a convention where we can get throughput metrics for streaming RPCs

specification/metrics/semantic_conventions/rpc.md
```diff
@@ -0,0 +1,82 @@
+# General
```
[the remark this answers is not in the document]
jsuereth: No this mirrors the trace `semantic_conventions` where there are "General RPC" conventions then we provide guidance for specific technologies if necessary (e.g. the `gRPC` callout).

I can rename to "General RPC"

specification/metrics/semantic_conventions/rpc.md
jsuereth: We discussed in spec sig.  I'm going to ask to defer fixing the shared RPC attribute labels to another PR.   There is an existing related proposal (#855), but I'll put forward another proposal that addresses this specific concern and works with existing proposal (or is compatible with it).

jsuereth:
@justinfoote Regarding RPC having HTTP labels if it's over HTTP.  While Personally I could see reasons to include it or not, I think perhaps we should follow from TRACE: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/rpc.md#distinction-from-http-spans

TL;DR:  The metrics is more about the fact that it's an RPC then it is about HTTP GET/POST/PUT/DELETE resources, so we shouldn't cross-report.

specification/metrics/semantic_conventions/rpc.md
```diff
@@ -0,0 +1,86 @@
...
+
+| Name                       | Instrument    | Units        | Description | Status | Streaming |
+|----------------------------|---------------|--------------|-------------|--------|-----------|
+| `rpc.server.duration`      | ValueRecorder | milliseconds | measures duration of inbound RPC | Recommended | N/A |
+| `rpc.server.request.size`  | ValueRecorder | bytes        | measures size of RPC request messages (uncompressed) | Optional | Recorded per message in a streaming batch |
+| `rpc.server.response.size` | ValueRecorder | bytes        | measures size of RPC response messages (uncompressed) | Optional | Recorded per response in a streaming batch |
+| `rpc.server.requests_per_rpc` | ValueRecorder | count | measures the number of messages received per RPC.  Should be 1 for all non-streaming RPCs | Optional | Required |
+| `rpc.server.responses_per_rpc` | ValueRecorder | count | measures the number of messages sent per RPC.  Should be 1 for all non-streaming RPCs | Optional | Required |
+
+### RPC Client (non-streaming)
```
yurishkuro, line 30: there is a Streaming column in the table

jmacd (approved):

specification/metrics/semantic_conventions/rpc.md
jmacd: Are you suggestion to remove " (non-streaming)" from the title? (I agree.)

specification/metrics/semantic_conventions/rpc.md
```diff
@@ -0,0 +1,86 @@
...
+
+*Note: RPC server and client metrics are split to allow correlation across client/server boundaries, e.g. Lining up an RPC method latency to determine if the server is responsible for latency the client is seeing.*
+
+### RPC Server
+
+Below is a table of RPC server metric instruments.
+
+| Name                       | Instrument    | Units        | Description | Status | Streaming |
+|----------------------------|---------------|--------------|-------------|--------|-----------|
+| `rpc.server.duration`      | ValueRecorder | milliseconds | measures duration of inbound RPC | Recommended | N/A |
```
MrAlias, line 24: Does N/A here mean that this instrument is not applicable to a streaming request, as it it shouldn't be used in that situation?

carlosalberto (approved):

specification/metrics/semantic_conventions/rpc.md
jsuereth: Doh, missed a refactor, thanks!

specification/metrics/semantic_conventions/rpc.md
jsuereth: I added clarification to why it's N/A.

jsuereth:
@MrAlias Is there anything else I need to do for this PR?

MrAlias (approved):
LGTM :+1:

jsuereth:
@ carlosalberto What else do I need to do for this PR?

specification/metrics/semantic_conventions/rpc.md
```diff
@@ -0,0 +1,86 @@
...
+| `rpc.client.response.size` | ValueRecorder | bytes        | measures size of RPC response messages (uncompressed) | Optional | Recorded per message in a streaming batch |
+| `rpc.client.requests_per_rpc` | ValueRecorder | count | measures the number of messages received per RPC.  Should be 1 for all non-streaming RPCs | Optional | Required |
+| `rpc.client.responses_per_rpc` | ValueRecorder | count | measures the number of messages sent per RPC.  Should be 1 for all non-streaming RPCs | Optional | Required |
+
+## Labels
+
+Below is a table of labels that SHOULD be included on metric events and whether
+or not they should be on the server, client or both.
+
+<!-- semconv rpc -->
```
arminru, line 48:
```suggestion
```
This is not actually generated by the generator tool (yet), right?

specification/metrics/semantic_conventions/rpc.md
```diff
@@ -0,0 +1,86 @@
...
+| [`net.peer.ip`](../../trace/semantic_conventions/span-general.md) | string | Remote address of the peer (dotted decimal for IPv4 or [RFC5952](https://tools.ietf.org/html/rfc5952) for IPv6) | `127.0.0.1` | See below |
+| [`net.peer.name`](../../trace/semantic_conventions/span-general.md) | string | Remote hostname or similar, see note below. | `example.com` | See below |
+| [`net.peer.port`](../../trace/semantic_conventions/span-general.md) | number | Remote port number. | `80`<br>`8080`<br>`443` | Conditional<br>See below |
+| [`net.transport`](../../trace/semantic_conventions/span-general.md) | string enum | Transport protocol used. See note below. | `IP.TCP` | Conditional<br>See below |
+
+**Additional attribute requirements:** At least one of the following sets of attributes is required:
+
+* [`net.peer.ip`](../../trace/semantic_conventions/span-general.md)
+* [`net.peer.name`](../../trace/semantic_conventions/span-general.md)
+<!-- endsemconv -->
```
arminru, line 63:
```suggestion
```

specification/metrics/semantic_conventions/rpc.md
```diff
@@ -0,0 +1,86 @@
+# General RPC conventions
+
+The conventions described in this section are RPC specific.  When RPC operations
+occur, metric events about those operations will be generated and reported to
+provide insight into those operations.   By adding RPC labels to metric events
+it allows for finely tuned filtering.
+
+**Disclaimer:** These are initial RPC metric instruments and labels but more may
+be added in the future.
+
```
arminru, line 10:
```suggestion
```
This is implicitly assumed for all of our semantic conventions and doesn't need any explicit disclaimer I'd say.

specification/metrics/semantic_conventions/rpc.md
```diff
@@ -0,0 +1,86 @@
+# General RPC conventions
+
+The conventions described in this section are RPC specific.  When RPC operations
+occur, metric events about those operations will be generated and reported to
+provide insight into those operations.   By adding RPC labels to metric events
+it allows for finely tuned filtering.
```
arminru, line 6:
```suggestion
The conventions described in this section are RPC specific. When RPC operations
occur, metric events about those operations will be generated and reported to
provide insight into those operations. By adding RPC labels to metric events
it allows for finely tuned filtering.
```

specification/metrics/semantic_conventions/rpc.md
```diff
@@ -0,0 +1,86 @@
+# General RPC conventions
+
```
arminru, line 2: It would be nice if you could add a TOC here (have a look at some of the other documents, these make use of `markdown-toc` for that).

arminru (commented):
Please add an entry in CHANGELOG.md for your new semantic convention. Thanks!

arminru:
@jsuereth The build checks required to pass for merging are currently saying "Waiting for status to be reported" because we switched to Github actions in the meantime. If you merge or rebase onto the current master, those changes should be picked up and the build should pass.

specification/metrics/semantic_conventions/rpc.md
jsuereth: added

specification/metrics/semantic_conventions/rpc.md
jsuereth: removed

specification/metrics/semantic_conventions/rpc.md
jsuereth: It is

specification/metrics/semantic_conventions/rpc.md
jsuereth: see above

jsuereth:
> Please add an entry in CHANGELOG.md for your new semantic convention. Thanks!

Done!

PR #1166, opened 2020-10-31
Rename trace/ to tracing/.

### LINKED ISSUE #1165: Editorial: Rename trace/ to tracing/
> A Trace can be thought of as a directed acyclic graph (DAG) of Spans

but this is not exactly what the tracing spec is about.

Related: open-telemetry/opentelemetry-java#1948

Oberon00:
"A Trace can be thought of as a directed acyclic graph (DAG) of Spans"

but this is not exactly what the tracing spec is about.

Fixes #1165

Related issues open-telemetry/opentelemetry-java#1948

---

⚠️ IMPORTANT: If this PR is accepted, we should also rename the spec:trace label and update it's description.


Oberon00:
One issue here is that we already have `TRACE_`  as prefix in environment variable names. I do not change this here as it would be more than an editorial change, and for envvar names, the shortness might actually even matter.

.github/ISSUE_TEMPLATE/trace.md
```diff
@@ -1,7 +1,7 @@
 ---
-name: Trace
-about: Issues related to the specification/trace directory
-labels: spec:trace
+name: Tracing
+about: Issues related to the specification/tracing directory
+labels: spec:tracing
```
Oberon00, line 4: The label itself would have to be renamed in the github repository settings when merging this.

specification/library-layout.md
```diff
@@ -15,7 +15,7 @@ api
    ├── context
    │   └── propagation
    ├── metrics
-   ├── trace
+   ├── tracing
```
Oberon00, line 18: As this is only a guideline, I think changing this is OK as SIGs can still keep the old name if they want (Java was actually using `tracing`, see e.g. https://github.com/open-telemetry/opentelemetry-java/tree/v0.9.1/sdk/tracing.

anuraaga (approved):

specification/library-layout.md
carlosalberto: (Although, to be clear, the package name itself still uses `trace`)

carlosalberto:
We already had this discussion long time ago, although I can't find the issue right now (@bogdandrutu @yurishkuro or @SergeyKanzhelev may remember this).

Personally I don't see any actual value in this, and I'm _slightly_ opposed to it - this is not a big change in itself, but we would probably end up de facto forcing SIGs to change their packages.

tigrannajaryan (changes requested):
We are breaking external links to spec files with this rename (I personally used plenty in outside repos, source codes and documents). I am opposed to this change unless I hear a good justification or someone knows a way to avoid breaking the links (there is no way to 301 redirect github content, is it?).

Oberon00:
@tigrannajaryan You should usually not link to master but a specific commit or version tag (e.g. https://github.com/open-telemetry/opentelemetry-specification/blob/v0.6.0/specification/trace/api.md will never break, also not after this PR). That said, in practice many probably don't follow this best-practice.

tigrannajaryan:
> @tigrannajaryan You should usually not link to master but a specific commit or version tag (e.g. https://github.com/open-telemetry/opentelemetry-specification/blob/v0.6.0/specification/trace/api.md will never break, also not after this PR). That said, in practice many probably don't follow this best-practice.

@Oberon00 that's a good advice that I will follow for the future, but unfortunately doesn't help with already existing links. It will be very difficult for me to go back, find and fix all links that I used in the past. I am guessing it is not just me, others may have done it too.

Oberon00:
Shortly talked about this in SIG spec, too little value, too risky now. See issue.

PR #1177, opened 2020-11-02
Add specification of metric instrument units

### LINKED ISSUE #705: Standardization of Units needed

justinfoote:
Fixes #705

## Changes

This PR adds some requirements about units for Metric Instruments. 

I made some decisions that I'd love feedback on:

1. We will not do any reaggregation or reconciliation of units in the SDK (this is to eliminate the need to parse unit strings or provide a more feature-rich units API, which is a long rabbit hole).
2. Metric instruments should assume that any unit string they get is valid.
3. I have left it up to API implementors to decide whether to provide constants or composition functions for the required units.
4. I've identified a handful of base units, along with a few prefixes that we consider important. I chose these based on the current semantic conventions, and I included a few other frequently used units. As we write more semantic conventions, we should add more units as needed.
5. I chose the case-insensitive form from the UCUM specification. I'm willing to be convinced that I made the wrong choice here. This is the guidance from UCUM:
    > Some computer systems or programming languages still have the requirement of case insensitivity and some humans who are not familiar with SI units tend to confuse upper and lower case or can not interpret the difference in upper and lower case correctly. For this reason the case insensitive symbols are defined. Although The Unified Code for Units of Measure does not encourage use of case insensitive symbols where not absolutely necessary, in some circumstances the case insensitive representation may be the greatest common denominator. Thus some systems that can handle case sensitivity may end up using case insensitive symbols in order to communicate with less powerful systems.


specification/metrics/semantic_conventions/README.md
```diff
@@ -105,14 +105,28 @@ instrument creation, but can be added if there is ambiguity.
...
+the range `[0, 1]` SHOULD use the default unity unit (`1`).
+Instruments that measure the fraction out of a total as a percentage, whose
+values are in the range `[0, 100]` SHOULD use the percent unit (`%`).
+
+Instruments that represent the number of discrete instances of a countable
+quantity SHOULD use either the default unity unit (`1`) or a
+[non-unit](https://ucum.org/ucum.html#para-50) (example: `{errors}` or
+`{packets}`)
+
+Implementors may choose the correct prefix to use for the scale of values being
```
MrAlias, line 122:
```suggestion
Implementors MAY choose the correct prefix to use for the scale of values being
```

MrAlias (approved):

specification/metrics/semantic_conventions/README.md
```diff
@@ -105,14 +105,28 @@ instrument creation, but can be added if there is ambiguity.
...
+measured. The choice of prefix may affect the value type to be recorded.
+
+> Example: it may be reasonable to record time as an integer when using an
```
justinfoote, line 125: @aabmass - I didn't see your comment on #705 before I wrote this up. I'd love your feedback about the wording of this section. How can we add the guidance you were suggesting (if the values to be recorded are whole numbers, instrumenters should prefer recording in a smaller unit (like nanoseconds) rather than recording a floating point in a larger unit)

aabmass: I wonder if others think this is a good idea in the first place. The reason I suggested it is:
- Higher precision
- Potentially smaller on wire if we ever switched OTLP to varint

However, Prometheus recommends base units (seconds) and statsd uses milliseconds for time. Is the gain in precision worth it over sticking to prior art? There is also a concern of overflow for very large values, but I can't think of an actual example that would overflow.

specification/metrics/semantic_conventions/README.md
```diff
@@ -105,14 +105,28 @@ instrument creation, but can be added if there is ambiguity.
...
+> instrument with a unit of nanosecond, when the same duration would be recorded
+> as a floating point on an instrument with a unit of second.
+
+OpenTelemetry APIs provide constants for all units used in the semantic
+conventions in these specifications (including
```
aabmass, line 130: Does this conflict with the above?

> The API MAY provide constants for each of the above required units.

justinfoote: Maybe. And now I'm re-reading the PR, I'm not sure these lines really even belong here.  I'm trying to say that implementations must be kept up to date as the spec evolves, but that's already expected, right? Maybe it doesn't need to be explicit here.

specification/metrics/semantic_conventions/README.md
justinfoote: I'm happy to change this, but I lowercased it on purpose. 
I don't think it's really a normative statement -- it's an instruction to people, it's not a declarative statement about the functionality of the implementation. Should it still be capitalized?

MrAlias: Your original intention sounds good.

specification/metrics/semantic_conventions/README.md
justinfoote: :+1: Removed in c151496

specification/metrics/api.md
```diff
@@ -326,6 +326,61 @@ Synchronous events have one additional property, the distributed
 [Context](../context/context.md) (containing Span, Baggage, etc.)
 that was active at the time.
 
+### Units of Measure
+
+All unit strings used in OpenTelemetry are expected to be valid according to The
```
jmacd, line 331: Can we write "Unit strings in OpenTelemetry SHOULD be valid according to ..."? I was wondering why you wrote "are expected to".

specification/metrics/api.md
```diff
@@ -326,6 +326,61 @@ Synchronous events have one additional property, the distributed
...
+[Unified Code for Units of Measure](https://ucum.org/ucum.html) (UCUM)
+specification using the case-insensitive form.
```
jmacd, line 333: You've used upper-case below, but we could equally well use lower-case, and when we talking about "reaggregating metrics of dissimilar units" below maybe we recommend SDKs SHOULD force the case one way or another so that differently-cased units aggregate together in an export pipeline.

As for whether to use upper- or lower-case, RFC 7540 states "header field names MUST be converted to lowercase prior to their encoding in HTTP/2". I have a preference for lower-case names, so naturally wonder what led you to choose uppercase?

specification/metrics/api.md
```diff
@@ -326,6 +326,61 @@ Synchronous events have one additional property, the distributed
...
+
+Metric instruments SHOULD NOT validate that the unit string passed to them is a
+valid value. The default unit to be used by a metric instrument when unit is not
+provided SHOULD be the non-annotated unity unit (that is: `1`).
```
jmacd, line 337: I worry a bit about the use of units jargon that may be lost on our readers. On the other hand, I want to embrace it. We should introduce terms to explain the syntax, like "{thing}" is a unity unit counting a quanity of things, to start with, so that a non-annotated unity unit is what you get when all you know is "quantity of things".

Also in the UCUM document, the terms "metric" vs "nonmetric". I'm afraid we also use the term "metric", so this overloading can be confusing, but it's also quite handy. 

Some of the lingering confusion about the choice between ValueRecorder/ValueObserver (a.k.a. "Grouping" instruments) vs the others (a.k.a. "Adding" instruments) can be reflected in terms used in the UCUM document.

For example,

> Furthermore, for a unit to be metric it must be a quantity on a ratio scale where multiplication and division with scalars are defined.

I think we can say you SHOULD NOT use a nonmetric unit with an Adding instrument.

> Those symbols that are used as units that imply a measurement on a scale other than a ratio scale (e.g., interval scale, logarithmic scale) are defined differently. These do not represent proper units as elements of the group (U,·). Therefore those special semantic entities are called special units, as opposed to proper units.

We could say that ValueRecorder/ValueObserver may be used with metric and nonmetric units, on the other hand, and the API has no way to say which it is. We might say that when you use a Grouping instrument for inputs with known-proper units, we can always correctly aggregate these using a sum (because they are "metric"). The SDK can automatically switch between sum and histogram aggregation when units are proper.

Just encouraging you to add a bit more , if you're up to it. :grinning:

specification/metrics/api.md
jkwatson: I don't think we can enforce with a normative statement what end-users are going to do. I think that's why this was left non-normative.

jmacd: Can we say "End-users SHOULD use valid unit strings according to ..."? 
Otherwise, let's make the statement less passive somehow, maybe "Users are expected to use valid unit strings according to ...".

justinfoote: I think we can make a normative statement here. We can't dictate user's behavior, but we can dictate that SDKs SHOULD assume that any metric they get is valid, and that they SHOULD NOT attempt to parse the value.

justinfoote: Though this phrase is in the API spec at the moment, and it's really dictating the behavior of the default SDK, so maybe it belongs somewhere else if we're going to make a normative statement about SDK behavior.

specification/metrics/api.md
justinfoote: 👍 I'm happy to use lowercase instead. I don't have a strong preference. I used uppercase because all of the case-insensitive examples in the UCUM spec are written in uppercase.

justinfoote: OK, I'm rewriting this portion now. I'm on board specifying these unit values as lowercase strings, but I'm hesitant to suggest that SDKs should coerce all units to lowercase. If a customer is manually instrumenting, and they have a special use case that requires case-sensitive unit values, I don't want the SDKs to stand in their way. 

I really think that SDKs should not mutate units at all -- they just assume that they're getting a valid unit, and that they can aggregate like instruments with the same unit together.

specification/metrics/api.md
justinfoote: OK, after giving this some thought, I think I'm in favor of less jargon. 
* I want to remove the term Unity. In the semantic convention table, I'd prefer we use the name `none`, to indicate that this is an instrument measuring something unitless.
* I don't love the idea of introducing a new definition of the term _metric_. In this case, it means a unit that can be _meaningfully_ scaled, especially by orders of magnitude. (They bring up the example that celsius is _not_ metric, even though the SI allows metric prefixes to be used with it, because, for example, 30 celsius is not 3 times hotter than 10 celsius.) This definition is so enormously different from those used elsewhere in the spec, I think it's inviting misunderstanding.
* While I could certainly see the convenience of an SDK choosing the appropriate aggregator based on the units of the instrument, I fear that it's putting too much meaning into a value that I don't think the SDK should need to reason about. There is _a lot_ of meaning that our SDKs could parse out of the units field, but it's a slippery slope.  This is part of why this PR took me so long to write -- I kept adding a whole bunch of complexity, then I'd have to scale it back and delete a bunch of paragraphs. 
* Having said all that, I would be in favor of adding guidance about what sort of units to use with adding or grouping instruments; I think that would be useful. I'll try putting something like that together today.

justinfoote:
Hey, anyone who's interested in seeing this PR merged, I'd love to get my build-tools PR reviewed. It generates markdown and code for these units. Find it here: https://github.com/open-telemetry/build-tools/pull/21

specification/metrics/api.md
justinfoote: OK, I'm curious about the implications of [another part of the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md#instrument-naming-requirements) related to instrument naming:
```markdown
Metric instruments are primarily defined by their name, which is how
we refer to them in external systems.  Metric instrument names conform
to the following syntax:

1. They are non-empty strings
2. They are case-insensitive
...
```

If instrument names are case-insensitive, does this imply that two instruments whose names differ only in casing MUST be aggregated together? If so, what is responsible for resolving the casing difference of those strings? And should the same logic be in place for unit strings?

justinfoote: I'm digging through the go reference implementation, and I can't find a place where a Descriptor's name is coerced to lowercase. It seems like _maybe_ this functionality belongs in the registry's `keyOf` or `uniqueInstrumentMeterImpl.checkUniqueness`, but I don't see it there.

jmacd: That sounds right, yes. I would be happy to implement a down-casing (or up-casing) there.

jgals:
@justinfoote and @jmacd, I'm commenting here to avoid having this PR closed by the bot. 🤖

specification/metrics/api.md
```diff
@@ -326,6 +326,61 @@ Synchronous events have one additional property, the distributed
...
+| megabaud    | signal transmission rate | MIBBD         |
+| gigabaud    | signal transmission rate | GIBBD         |
+| terabaud    | signal transmission rate | TIBBD         |
+| connections | connections              | {connections} |
+| errors      | errors                   | {errors}      |
+| faults      | faults                   | {faults}      |
+| operations  | operations               | {operations}  |
+| packets     | packets                  | {packets}     |
+| processes   | processes                | {processes}   |
+| threads     | threads                  | {threads}     |
```
jkwatson, line 374: would it make sense to put these into some yaml so the could be auto-generated to support them?

jmacd: (as a TODO?)

spec-compliance-matrix.md
```diff
@@ -80,7 +80,7 @@ status of the feature is not known.
 
 |Feature                                       |Go|Java|JS |Python|Ruby|Erlang|PHP|Rust|C++|.Net|
 |----------------------------------------------|--|----|---|------|----|------|---|----|---|----|
-|TBD|
+|Provide commonly used units                   |  |    |   |      |    |      |   |    |   |    |
```
jmacd, line 83: Maybe add `Metric SDK Accumulator differentiates aggregations based on unit value`.

specification/metrics/api.md
jmacd: How about "identity unit"? "unitless unit"?

specification/metrics/api.md
justinfoote: :100: There is a corresponding PR here: https://github.com/open-telemetry/build-tools/pull/21

I had considered adding the yaml to this PR according to that proposal, but I thought it more prudent to wait until that PR merges. It's looking like that's imminent, so maybe I'll go ahead and add the yaml now.

specification/metrics/api.md
justinfoote: :+1: I think I prefer unitless.

spec-compliance-matrix.md
justinfoote: :+1: Added in 1130c9c

specification/metrics/api.md
justinfoote: I added the yaml and the generated markdown table in 6fe5382. 
This introduces a hard dependency between these PRs though. We can't merge this one until the build-tools one is merged.

specification/metrics/api.md
justinfoote: I went with unitless for now. I'm open to more feedback about it. Naming things is hard.

justinfoote:
Doh! This got closed over the holiday. I'm reopening it now.  I think I've actually responded to all the feedback on it.
@open-telemetry/specs-metrics-approvers, I'd love some more reviews. If possible, I'd love to see this PR merged this week.

jmacd (approved):
Thank you @justinfoote

jmacd:

weyert:
@open-telemetry/specs-metrics-approvers Was it intended to close this PR?

jmacd:
@weyert I ❤️ this PR and want to get it approved.

Oberon00:
In the context of https://github.com/open-telemetry/build-tools/pull/65, this came up again. Do you still plan to merge this? If so, would someone care to update the generator docs to document this "unit" feature?

Issue #823, opened 2020-08-18
Consideration when using OTel SDK (and maybe collector) with non-OTLP backend


anuraaga:
OpenTelemetry by default collects a lot of information, and backends that speak OTLP natively will generally be equipped to deal with this. But I worry we don't have a good story right now for connecting to non-OTLP backends like Zipkin (I will use Zipkin as the example since I'm most familiar with it, I wouldn't be surprised if other systems can have similar issues with storage). Zipkin instrumentation generally defaults to collecting less information since users run their own backends and pay significant cost for data. But currently, in the specification exporters are required to convert almost all information in OTLP. For example, [library info](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk_exporters/zipkin.md#instrumentationlibrary) and all the [attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk_exporters/zipkin.md#attribute) which given the breadth of semantic conventions can be a lot. Resource hasn't been specced [yet](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk_exporters/zipkin.md#attribute) but with this pattern I expect it to also end up populating a lot of information for resources because the current stance is to preserve data during conversion

https://github.com/open-telemetry/opentelemetry-specification/pull/800#discussion_r470857107

I think the goal of preserving all data is a nice goal but it could provide challenges to organizations using the opentelemetry sdk with a zipkin backend. We can imagine a software development team that switches their instrumentation to opentelemetry, and when they release the new version, their metrics look fine and they go to prod, taking down the entire tracing infrastructure due to an explosion in data that may provide low value given the backend doesn't natively recognize it. We can expect tracing infrastructure to have count-based ingestion controls, but I think size-based controls are less common and this would be very hard to deal with.

One piece we have that can help is the collector's attribute processor. It can be used to filter out attributes. This is a nice start, but we may need to consider improvements

- Prebaked configuration or at least an example of how to configure it to get a result that is similar to what most Zipkin users have to effect a low impact on the backend

- Support an allow mode instead of only deletion which would allow new otel conventions to be opt-in.

[...]

PR #1185, opened 2020-11-03
Remove resource reporting from zipkin exporter spec for now

anuraaga:
I had never noticed the word Resource here before. I think this is problematic since the overwhelming size / cardinality of resource will likely kill most zipkin deployments if attached to every span (zipkin data format does not have any mechanism for flattening these into a shared struct).

#823 discusses this but has been pushed to post-GA - I would recommend we also push Resource reporting for Zipkin to post-GA to coincide with that, or otherwise prioritize #823 (e.g., spec out a way for SDKs to allow filtering resource)


Oberon00 (approved):
Maybe it would make sense to a "MAY" for Resources and a short note describing the problem and suggesting to make it configurable.

anuraaga:
@Oberon00 Yeah it makes sense but I'm a little worried about diverging configuration style. Especially if the setting is exposed as an environment variable, we could have inconsistency of the variable across languages. Do you think the recommendation makes sense with an added caveat of not being exposed as env var, which we usually spec in order to have consistency?

Oberon00:
You are right, that is rather complicated. Let's leave this PR as-is then, and this configuration stuff can be done after GA.

Oberon00:
Actually, we might want an issue for that, so it can be triaged.

anuraaga:
Thanks - filed #1196

jmacd:
This issue is remarkably similar to an issue in the metrics world facing most of the common exposition formats. I would like to see an OTel-wide solution. In Metrics we've discussed the outcome we'd like:
1. exporters w/o explicit resource supports will drop resources
2. users are expected to configure the resource->(label or attribute) translation they want
3. the OTel collector has (will have) a resource processor capable of moving resources into metric labels
The same should probably be said for trace exposition formats without resources.

jmacd:
(I'm not sure we have a good GH issue capturing the above "issue", but it's been discussed in the context of Prometheus remote-write exporters recently.)

anuraaga:
Oops lost track of this issue - is it ok to continue with it?

jmacd (approved):

iNikem:
@jmacd As assignee, are you Ok with merging this or do we need more feedback?

carlosalberto:
Overall LGTM - but as Jaeger Spans will keep this part, can we add a note here? something along "Resource as tags has been explicitly discarded for now, etc, etc".

carlosalberto (approved):

jmacd:
I think we should merge this because #1196 has been filed.

Issue #586, opened 2020-05-06
obtain a tracer/meter without a name


zeitlinger:
# Summary

## Background

In [Obtaining a tracer](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#obtaining-a-tracer) and [Obtaining a Meter](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api-user.md#obtaining-a-meter), it says that `name` is required.

## Proposed change

The `name` should be optional.

## Use case

Obtaining a tracer/meter for the main application, rather than a library.

## What should "name" be, if not provided?

Something like [service.name](https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/resource/semantic_conventions#service).

## Rationale

1. Make the user API easier for the main use case (not a library).
2. Have a consistent naming scheme for application tracer/meter, e.g. `service.name`.

# Analysis

## Problem

Provided that the resource should already be sufficient to identify an application, it seems redundant for the user to provide this information again.

In the best case, the application developer will provide the correct name (e.g. the value of `service.name`) - where "correct" is defined either by semantic convention, or a company internal standard.

In the worst case, the application developer will fill in some random value, which will make it harder to identify the trace/meter as being the one from the main application, which in turn could make further processing (e.g. in the collector) harder.

### Static name

The `name` should be static, e.g. `main`.

Rationale: The tracer/meter can still be identified by the resource, e.g. `service.name`.

Pro:
1. It's very easy to identify the tracer/meter of the main application.

Con:
1. You always need the resource for context. However, this is already the case for library, e.g. you don't know which application created a span from the tracer `io.opentelemetry.contrib.mongodb` without the resource.

### Service.name

The `name` should be filled with `service.name`.

Rationale: `service.name` is the same as (or the closest to) the application name.

Pro/Con: Inverse of the above.

### Span exporter

Let the span exporter decide how to fill the tracer/meter name.

Rationale: The setup of the span exporter can be extracted into a library and be reused. In this setup logic, you can e.g. read the application name from an environment variable.

dyladan:
My main worry about this is if you make the name optional, library developers will leave it out. If it is required it can be statically type checked to ensure it is there.

Oberon00:
[...]

PR #1187, opened 2020-11-03
Clarify the location of tracing's Context interaction.

carlosalberto:
Clarifies the location of the tracing interaction with `Context`, i.e. we want to say that this logic can exist **anywhere** in the API, as long as it exists in a single place.

The mention of a `TracingContextUtils` class seemed like a *too-specific* suggestion, hence the removal (in Java we had this class, but ended up putting the `Context` interaction as static/instance `Span` directly).

### DIFF
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -149,9 +149,8 @@ the following functionality:
 - Set the currently active span to the implicit context. This is equivalent to getting the implicit context, then inserting the `Span` to the context.
 
 All the above functionalities operate solely on the context API, and they MAY be
-exposed as either static methods on the trace module or as static methods on a class
-inside the trace module. This functionality
-SHOULD be fully implemented in the API when possible.
+exposed as either static methods on the trace module, or as static methods on a class
+inside the trace module. This functionality SHOULD be fully implemented in the API when possible.
 
 ## Tracer
 
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -149,7 +149,7 @@ the following functionality:
 - Set the currently active span to the implicit context. This is equivalent to getting the implicit context, then inserting the `Span` to the context.
 
 All the above functionalities operate solely on the context API, and they MAY be
-exposed as either static methods on the trace module, as static methods on a class
+exposed as either static methods on the trace module or as static methods on a class
 inside the trace module. This functionality
 SHOULD be fully implemented in the API when possible.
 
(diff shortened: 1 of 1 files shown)


specification/trace/api.md
```diff
@@ -149,10 +149,9 @@ the following functionality:
 - Set the currently active span to the implicit context. This is equivalent to getting the implicit context, then inserting the `Span` to the context.
 
 All the above functionalities operate solely on the context API, and they MAY be
-exposed as static methods on the trace module, as static methods on a class
-inside the trace module (it MAY be named `TracingContextUtilities`), or on the
-[`Tracer`](#tracer) class. This functionality SHOULD be fully implemented in the
-API when possible.
+exposed as either static methods on the trace module, as static methods on a class
+inside the trace module, or on the [`Tracer`](#tracer) class. This functionality
```
Oberon00, line 153:
```suggestion
inside the trace module. This functionality
```

Alternatively, add Span to the list of suggestions.

Oberon00 (approved):

specification/trace/api.md
carlosalberto: Actually, anything that keeps things simple is better, I think.

carlosalberto:
Actually, we may need to also remove the mention of `static`, as IIRC Ruby poked with the idea of keeping these in the `Tracer` as instance methods. @open-telemetry/ruby-approvers ? ;)

Oberon00:
I am against removing "static". I mean I'm open to improving the wording to make it more language neutral, but a Tracer instance MUST NOT be required to interact with the current span. This often leads to misunderstandings, such as that each named tracer manages it's own active span, or complaint's such as "why do I need to provide an instrumentation library name to access the current span" (#586).

yurishkuro (approved):

specification/trace/api.md
```diff
@@ -149,10 +149,9 @@ the following functionality:
 - Set the currently active span to the implicit context. This is equivalent to getting the implicit context, then inserting the `Span` to the context.
 
 All the above functionalities operate solely on the context API, and they MAY be
-exposed as static methods on the trace module, as static methods on a class
-inside the trace module (it MAY be named `TracingContextUtilities`), or on the
-[`Tracer`](#tracer) class. This functionality SHOULD be fully implemented in the
-API when possible.
+exposed as either static methods on the trace module, as static methods on a class
```
iNikem, line 152:
```suggestion
exposed as either static methods on the trace module or as static methods on a class
```

bogdandrutu:
@carlosalberto we can do a separate PR for the "static" if you feel strongly.

bogdandrutu (approved):

PR #1117, opened 2020-10-20
Add OTEL_SAMPLING_PROBABILITY env variable definition

tigrannajaryan:
Fixes https://github.com/open-telemetry/opentelemetry-specification/issues/1105

Issue #1188, opened 2020-11-03
Support restarting the trace with a different trace ID


yurishkuro:
**What are you trying to achieve?**

When SaaS products from cloud providers participate in the tracing, the providers are often reluctant to respect customer-supplied trace ID because it can be abused by the caller (e.g. sending the same trace ID for all requests). W3C trace context explicitly allows nodes in the call graph to restart a trace with a new trace ID, and return it to the caller via the response header (currently in the design phase). However, the current OTEL SDK spec states that the tracer must always respect & reuse the inbound trace ID.

OTEL SDK may support a configurable mechanism to restart a trace. The inbound SpanContext in this case can still be recorded as a Link to external trace.

**Additional context.**

Previous discussion here: https://github.com/open-telemetry/opentelemetry-specification/pull/998#discussion_r516932534

**Alternatives.**

The behavior may also be implemented purely on the instrumentation side, e.g.

```
spanCtx := extract(request)
var span Span
if trusted(request) {
    span = tracer.StartSpan(parent=spanCtx)
} else {
    span = tracer.StartSpan(links=[spanCtx])
}
```

If OTEL Spec Committee decides not to support this directly in the SDKs, then the use case should be documented in some FAQ.

dyladan:
The response header format is not near ready for implementation. The basic fields and data format is not yet agreed on. I would strongly recommend against implementing anything based on it yet.

As far as restarting the trace, this could be a useful feature for some people and should be fairly straightforward.

svrnm:
The label removal we applied during triage stopped the `stale` bot from closing this issue, so we close it manually, if this is incorrect please re-open or raise a new issue, cc @mx-psi @jpkrohling

PR #611, opened 2020-05-18
Add requirements for probability sampler

bogdandrutu:
In https://github.com/open-telemetry/opentelemetry-specification/pull/331 there were questions about why different requirements are important when deciding on the probability sampling algorithm.

PR #1190, opened 2020-11-04
Add OTEL_TRACE_SAMPLER_TRACEIDRATIO_RATIO env variable definition

### LINKED ISSUE #1105: Add OTEL_SAMPLING_PROBABILITY (or similar named) to sdk environment variable spec

Consistency across languages

**Additional context.**

Java SDK currently uses `OTEL_CONFIG_SAMPLER_PROBABILITY`

Javascript SDK currently uses `OTEL_SAMPLING_PROBABILITY`

It would be nice to agree on name before the SDKs GA, so we don't need to change it after GA.

carlosalberto:
Fixes #1105 

Opened on behalf of #1117 - a few of the pending issues to discuss/address:

* Any additional instructions on how this should work?
* Is it enough to have a simple clarification on this env var used only in conjunction with the TraceIdRatio Sampler?
* A follow-up should come, mentioning how this and other env-vars (such as `OTEL_PROPAGATORS`) are meant to be primarily used by Auto-Instrumentation, and not by SDKs yet.

cc @jkwatson @cijothomas @Oberon00

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,6 @@ New:
 
 - Enforce that the Baggage API must be fully functional, even without an installed SDK.
   ([#1103](https://github.com/open-telemetry/opentelemetry-specification/pull/1103))
-- Add OTEL_TRACE_SAMPLING_PROBABILITY env variable definition
-  ([#1117](https://github.com/open-telemetry/opentelemetry-specification/pull/1117))
 - Rename "Canonical status code" to "Status code"
   ([#1081](https://github.com/open-telemetry/opentelemetry-specification/pull/1081))
 - Add Metadata for Baggage entries, and clarify W3C Baggage Propagator implementation
@@ -63,6 +61,8 @@ New:
   ([#937](https://github.com/open-telemetry/opentelemetry-specification/pull/937))
 - Add OTEL_TRACE_SAMPLER env variable definition
   ([#1136](https://github.com/open-telemetry/opentelemetry-specification/pull/1136/))
+- Add OTEL_TRACE_SAMPLER_TRACEIDRATIO_RATIO env variable definition
+  ([#1190](https://github.com/open-telemetry/opentelemetry-specification/pull/1190))
 
 Updates:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ New:
 
 - Enforce that the Baggage API must be fully functional, even without an installed SDK.
   ([#1103](https://github.com/open-telemetry/opentelemetry-specification/pull/1103))
+- Add OTEL_SAMPLING_PROBABILITY env variable definition
+  ([#1117](https://github.com/open-telemetry/opentelemetry-specification/pull/1117))
 - Rename "Canonical status code" to "Status code"
   ([#1081](https://github.com/open-telemetry/opentelemetry-specification/pull/1081))
 - Add Metadata for Baggage entries, and clarify W3C Baggage Propagator implementation
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ New:
 
 - Enforce that the Baggage API must be fully functional, even without an installed SDK.
   ([#1103](https://github.com/open-telemetry/opentelemetry-specification/pull/1103))
-- Add OTEL_SAMPLING_PROBABILITY env variable definition
+- Add OTEL_TRACE_SAMPLING_PROBABILITY env variable definition
   ([#1117](https://github.com/open-telemetry/opentelemetry-specification/pull/1117))
 - Rename "Canonical status code" to "Status code"
   ([#1081](https://github.com/open-telemetry/opentelemetry-specification/pull/1081))
--- a/spec-compliance-matrix.md
+++ b/spec-compliance-matrix.md
@@ -127,7 +127,11 @@ status of the feature is not known.
 |OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT               |   |    |   |      |    |      |   |    |   |    |
 |OTEL_SPAN_EVENT_COUNT_LIMIT                   |   |    |   |      |    |      |   |    |   |    |
 |OTEL_SPAN_LINK_COUNT_LIMIT                    |   |    |   |      |    |      |   |    |   |    |
+<<<<<<< HEAD
 |OTEL_TRACE_SAMPLER                            |   |    |   |      |    |      |   |    |   |    |
+=======
+|OTEL_SAMPLING_PROBABILITY                     |   |    |   |      |    |      |   |    |   |    |
+>>>>>>> Add OTEL_SAMPLING_PROBABILITY env variable definition
 
 ## Exporters
 
--- a/spec-compliance-matrix.md
+++ b/spec-compliance-matrix.md
@@ -127,11 +127,8 @@ status of the feature is not known.
 |OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT               |   |    |   |      |    |      |   |    |   |    |
 |OTEL_SPAN_EVENT_COUNT_LIMIT                   |   |    |   |      |    |      |   |    |   |    |
 |OTEL_SPAN_LINK_COUNT_LIMIT                    |   |    |   |      |    |      |   |    |   |    |
-<<<<<<< HEAD
 |OTEL_TRACE_SAMPLER                            |   |    |   |      |    |      |   |    |   |    |
-=======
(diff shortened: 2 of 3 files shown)


spec-compliance-matrix.md
```diff
@@ -128,6 +128,7 @@ status of the feature is not known.
 |OTEL_SPAN_EVENT_COUNT_LIMIT                   |   |    |   |      |    |      |   |    |   |    |
 |OTEL_SPAN_LINK_COUNT_LIMIT                    |   |    |   |      |    |      |   |    |   |    |
 |OTEL_TRACE_SAMPLER                            |   |    |   |      |    |      |   |    |   |    |
+|OTEL_TRACE_SAMPLING_PROBABILITY               |   |    |   |      |    |      |   |    |   |    |
```
cijothomas, line 131: You want to call it OTEL_TRACE_SAMPLER_TRACEIDRATIO_RATIO, right?

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ New:
 
 - Enforce that the Baggage API must be fully functional, even without an installed SDK.
   ([#1103](https://github.com/open-telemetry/opentelemetry-specification/pull/1103))
+- Add OTEL_TRACE_SAMPLING_PROBABILITY env variable definition
```
cijothomas, line 14: OTEL_TRACE_SAMPLER_TRACEIDRATIO_RATIO

spec-compliance-matrix.md
carlosalberto: Ooops, yes. Will fix.

cijothomas (approved):
LGTM - need typo (i think) fix. Rename OTEL_TRACE_SAMPLING_PROBABILITY to OTEL_TRACE_SAMPLER_TRACEIDRATIO_RATIO

cijothomas:
> A follow-up should come, mentioning how this and other env-vars (such as OTEL_PROPAGATORS) are meant to be primarily used by Auto-Instrumentation, and not by SDKs yet.

The following from the spec makes this clear already right? SDK's may chose to support ENV var based option. Its not a MUST have. If they support it, then use the names from the spec. (I am happy with the wordings :) )
_SDKs MAY choose to allow configuration via the environment variables in this specification, but are not required to. If they do, they SHOULD use the names listed in this document._

specification/sdk-environment-variables.md
```diff
@@ -9,7 +9,8 @@ The goal of this specification is to unify the environment variable names betwee
 | OTEL_RESOURCE_ATTRIBUTES | Key-value pairs to be used as resource attributes |                                   | See [Resource SDK](./resource/sdk.md#specifying-resource-information-via-an-environment-variable) for more details. |
 | OTEL_LOG_LEVEL           | Log level used by the SDK logger                  | "info"                            |                                     |
 | OTEL_PROPAGATORS         | Propagators to be used as a comma separated list  | "tracecontext,baggage"            | Values MUST be deduplicated in order to register a `Propagator` only once. Unrecognized values MUST generate a warning and be gracefully ignored. |
-| OTEL_TRACE_SAMPLER       | Sampler to be used for traces                     | "parentbased_always_on"                       | See [Sampling](./trace/sdk.md#sampling) |
+| OTEL_TRACE_SAMPLER       | Sampler to be used for traces                     | "parentbased_always_on"           | See [Sampling](./trace/sdk.md#sampling) |
+| OTEL_TRACE_SAMPLER_TRACEIDRATIO_RATIO | Sampling probability, a number in the [0..1] range, e.g. 0.25 | 1        | The specified value will only be used if OTEL_TRACER_SAMPLER is set to "traceidratio" |
```
Oberon00, line 13:
```suggestion
| OTEL_TRACE_SAMPLER_TRACEIDRATIO_RATIO | Sampling probability, a number in the [0..1] range, e.g. 0.25 | 1        | The specified value will only be used if OTEL_TRACER_SAMPLER is set to "traceidratio" or "parentbased_traceidratio" |
```

Oberon00 (commented):
LGTM, but a few typos & nits that should be fixed.

specification/sdk-environment-variables.md
Oberon00, line 13: I don't think the default should be `1` BTW, that would make it the same as alwayson. What about a default of `0.01` (1%)? You would usually switch to traceidratio from the default parentbased_always_on to reduce your load.

carlosalberto: Not a sampling expert here. I'm fine with the value you proposed, so let's take it for a run.

Oberon00: I'm also not a sampling expert, I just pulled this value out of thin air tbh. But anything less than 1 would be a more useful default I think.

carlosalberto: Done.

Oberon00 (approved):
Thank you!

carlosalberto:
Ready for review.

@cijothomas I created #1193 as a follow-up 😉

specification/sdk-environment-variables.md
```diff
@@ -9,7 +9,8 @@ The goal of this specification is to unify the environment variable names betwee
...
 | OTEL_LOG_LEVEL           | Log level used by the SDK logger                  | "info"                            |                                     |
 | OTEL_PROPAGATORS         | Propagators to be used as a comma separated list  | "tracecontext,baggage"            | Values MUST be deduplicated in order to register a `Propagator` only once. Unrecognized values MUST generate a warning and be gracefully ignored. |
-| OTEL_TRACE_SAMPLER       | Sampler to be used for traces                     | "parentbased_always_on"                       | See [Sampling](./trace/sdk.md#sampling) |
+| OTEL_TRACE_SAMPLER       | Sampler to be used for traces                     | "parentbased_always_on"           | See [Sampling](./trace/sdk.md#sampling) |
+| OTEL_TRACE_SAMPLER_TRACEIDRATIO_RATIO | Sampling probability, a number in the [0..1] range, e.g. 0.25 | 0.01        | The specified value will only be used if OTEL_TRACER_SAMPLER is set to "traceidratio" or "parentbased_traceidratio" |
```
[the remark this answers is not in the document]
carlosalberto: This a tricky one as, I suspect, there's no perfect solution. The previous name didn't suit anybody, but I think we are open to good alternatives.

(An alternative is to provide a general-purpose `OTEL_TRACE_SAMPLER_PROBABILITY` which would work on _any_ probability-based `Sampler`, but that can get easily tricky IMHO).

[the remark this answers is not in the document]
anuraaga: I think random_sample_rate / probability type of names could make sense, key word being random. Users don't really care whether the random is based on an inline computed random number or a random number that happens to be in the the trace id, so I would focus on that in the name. I don't think OTel needs to officially support more than one random sampler ever does it? Maybe this does bring up that the current traceidratio sampler is too specifically named and could just be "random". Seems like a nice change for users to me.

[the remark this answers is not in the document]
Oberon00: The original name of TraceIdRatioBased was Probability, and it was deliberately renamed to say what it does, see #611. It is not really random. If you have many spans in the same trace, all will have the same sampling decision. Thus, TraceIdRatioBased will not help you against bad actors that send you requests with a known-sampled trace ID (related #1188).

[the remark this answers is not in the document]
anuraaga: I see - well the main point that came up for me here is this name has trace appear twice, ratio appear twice, so lots of stuttering, and doesn't seem to indicate that it is picking a random sampling rate, one of the most common settings a tracing user applies in my experience, after service name. It just seems confusing.

> (An alternative is to provide a general-purpose OTEL_TRACE_SAMPLER_PROBABILITY which would work on any probability-based Sampler, but that can get easily tricky IMHO).

Perhaps this is still an ok approach to make clear what the variable does? We don't actually automatically enable the sampler without the other env var too so it seems ok.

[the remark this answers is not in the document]
yurishkuro: TC discussed this on the call today and the recommendation is to follow the pattern used in Jaeger SDK, which has two properties, SAMPLER_TYPE and SAMPLER_PARAM. The PARAM is numeric and is interpreted by the specific sampler as necessary, however for even more flexibility PARAM can be treated as a string, which allows it to even be JSON if a sampler requires more complex configuration.

I think @carlosalberto has an action item to propose alternative PR(s).

tigrannajaryan:
I suggest to close this in favor of https://github.com/open-telemetry/opentelemetry-specification/pull/1202 which has more approvals and IMO is a nicer solution.

PR #882, opened 2020-08-26
Add process.runtime.name / version describing executing runtime

anuraaga:
Fixes #830

## Changes

Adds attributes to fill the name and version of the runtime, e.g. openjdk 14.0.1, dotnet-framework 4.5, cpython 3.3, etc.

I tried to collect runtime names from various sources like wikipedia, sdkman, and random googling, but am not familiar enough with many of the languages to know how representative these are, let me know! Naming wise, I used a hyphen instead of a space for any name that had a space and otherwise just lowercased.

PR #789, opened 2020-08-12
Add semantic conventions for thread.name and thread.id span attributes

mateuszrzeszutek:
Fixes #788

## Changes

Added two new span attributes responsible for storing current thread information: `thread.name` and `thread.id`

Related issues: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/942

PR #1194, opened 2020-11-04
Add middleware to process semantic conventions

### LINKED ISSUE #1143: Possibility to support Application Server in semantic conventions?
I will be involved in contributing quite a few auto-instrumentations for different Java application servers over the next few months for OpenTelemetry Java Agent. Support and tests for Jetty, Tomcat, WildFly, GlassFish, JBoss, etc will be built with my involvement.

What I am currently struggling to understand is the means to send the application-server specific information using the OpenTelemetry. What I expect to send is the information that “This span originated from Apache Tomcat 7.0”, probably encoded in two different key-value pairs. 

I looked at the semantic conventions about Resources, especially [Process](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/resource/semantic_conventions/process.md), but it does not seem to offer such possibility. _process.runtime_ seems to be be reserved for JVM vendor/version. So I am looking at two (seemingly) bad alternatives as of now: 

- send this information just as span attributes (creating redundancy & bloating the protocol) or 
- go off the standard and add custom namespace to the Process which would be specific to one vendor (Splunk to be precise). 

At least in the Java world, the application server is an extremely important dimension when trying to make sense of the runtime architecture of yours, so I am trying to figure out whether or not this should be part of the semantic conventions. For other platforms the issue might not be as severe, but for example the difference between php-fpm and mod_php for PHP installations and the PyPy vs CPython for the Python world might reveal similar issues.

ivomagi:
As discovered while analyzing the request from https://github.com/open-telemetry/opentelemetry-specification/issues/1143, it became apparent that the current semantic conventions do lack the information about the middleware running on the runtime within the process. Adding this information makes the traces more valuable - the proposed change makes it possible to visualize that Tomcat 9.0.39 application server is sending messages to a Kafka message queue which in turn has a subscriber running on GlassFish 5.0 persisting the content to Druid 0.15.0. In the current form, it seems that the semantic conventions leave this information at the span level, which would be creating redundancy in outbound communication from the agent and/or not capture the information in unified manner, making it harder for Otel-compliant servers to visualize such traces.

Fixes #1143

## Changes

Added middleware definition to the resource semantic conventions.

Related issues #

### DIFF
--- a/specification/resource/semantic_conventions/process.md
+++ b/specification/resource/semantic_conventions/process.md
@@ -136,7 +136,7 @@ Examples for some Ruby runtimes
 
 **type:** `process.middleware`
 
-**Description:** The single middleware instance which is monitored on the runtime in this process.
+**Description:** The single middleware instance which is monitored on this process.
 
 | Attribute  | Description  | Example  | Required |
 |---|---|---|--|
--- a/specification/resource/semantic_conventions/process.md
+++ b/specification/resource/semantic_conventions/process.md
@@ -131,3 +131,59 @@ Examples for some Ruby runtimes
 | --- | --- | --- | --- |
 | MRI | ruby | 2.7.1 | ruby 2.7.1p83 (2020-03-31 revision a0c7c23c9c) [x86_64-darwin19] |
 | TruffleRuby | truffleruby | 2.6.2 | truffleruby (Shopify) 20.0.0-dev-92ed3059, like ruby 2.6.2, GraalVM CE Native [x86_64-darwin] |
+
+## Process middlewares
+
+**type:** `process.middleware`
+
+**Description:** The single middleware instance which is monitored on the runtime in this process.
+
+| Attribute  | Description  | Example  | Required |
+|---|---|---|--|
+| process.middleware.type | Standardized attribute, specifying the type of middleware on the runtime. This attribute adds value to traces, giving the opportunity to visualise different kind of middleware using different representation in server-side. The supported values of type are `as` (application server), `ws` (web server), `dbs` (database server) or `mom` (message-oriented middleware). | `as` | No |
+| process.middleware.name | The name of the middleware running on this process. | `Tomcat` | No |
+| process.middleware.version | The version of the middleware of this process, as returned by the middleware (preferably during runtime) without modification. | `9.0.39` | No |
+| process.middleware.description | An additional description about the middleware of the process. | `Apache Tomcat/9.0.39` | No |
+
+`process.middleware.type` MUST be set to one of the values listed below. 
+`process.middleware.name` SHOULD be set to one of the values listed below. If none of the listed values apply, a custom value best describing the runtime MAY be used.
+
+***Application server middlewares:***
+
+Information describing the application server should be captured using the values acquired from the middleware specific API, preferably during runtime. As an example - for Java application servers this information can be captured by invoking `ServletContext.getServerInfo()` during runtime and parsing the result.
+
+Examples for some application server middlewares:
+
+| Name | `process.middleware.name` | `process.middleware.version` | `process.middleware.description` |
+| --- | --- | --- | --- |
+| Apache Tomcat | `Tomcat` | `9.0.39` | `Apache Tomcat/9.0.39` |
+| Eclipse Jetty | `Jetty` | `9.4.33` | `jetty/9.4.33.v20201020` |
+| Eclipse GlassFish | `GlassFish` | `5.0` | `GlassFish Server Open Source Edition  5.0` |
+| WildFly | `WildFly` | `21.0.0` | `WildFly Full 21.0.0.Final (WildFly Core 13.0.1.Final) - 2.2.2.Final` |
+
+***Web server middlewares:***
+
+Examples for some web server middlewares:
+
+| Name | `process.middleware.name` | `process.middleware.version` | `process.middleware.description` |
+| --- | --- | --- | --- |
+| nginx | `nginx` | `1.19.3` | `nginx version: nginx/1.19.3` |
+| Apache HTTP Server | `apache` | `2.4.41` | `Apache/2.4.41 (Ubuntu) OpenSSL/1.1.1f` |
+
+***Message-oriented middlewares:***
+
+Examples for some message-oriented middlewares:
+
+| Name | `process.middleware.name` | `process.middleware.version` | `process.middleware.description` |
+| --- | --- | --- | --- |
+| Apache Kafka | `Kafka` | `2.6.0` | `2.6.0 (Commit:62abe01bee039651)` |
+
+***Database middlewares:***
+
+Examples for some database middlewares:
+
+| Name | `process.middleware.name` | `process.middleware.version` | `process.middleware.description` |
+| --- | --- | --- | --- |
+| Apache Druid | `Druid` | `0.15.0` | `0.15.0-incubating` |
+| MySQL | `MySQL` | `5.7.28` | `5.7.28-0ubuntu0.19.04.2` |
+
--- a/specification/resource/semantic_conventions/process.md
+++ b/specification/resource/semantic_conventions/process.md
@@ -141,12 +141,12 @@ Examples for some Ruby runtimes
 | Attribute  | Description  | Example  | Required |
 |---|---|---|--|
 | process.middleware.type | Standardized attribute, specifying the type of middleware on the runtime. This attribute adds value to traces, giving the opportunity to visualise different kind of middleware using different representation in server-side. The supported values of type are `as` (application server), `ws` (web server), `dbs` (database server) or `mom` (message-oriented middleware). | `as` | No |
-| process.middleware.name | The name of the middleware running on this process. | `Tomcat` | No |
-| process.middleware.version | The version of the middleware of this process, as returned by the middleware (preferably during runtime) without modification. | `9.0.39` | No |
-| process.middleware.description | An additional description about the middleware of the process. | `Apache Tomcat/9.0.39` | No |
+| process.middleware.name | The name of the middleware running on this process. | `WildFly` | No |
+| process.middleware.version | The version of the middleware of this process, as returned by the middleware (preferably during runtime) without modification. | `21.0.0` | No |
+| process.middleware.description | An additional description about the middleware of the process. | `WildFly Full 21.0.0.Final (WildFly Core 13.0.1.Final) - 2.2.2.Final` | No |
 
-`process.middleware.type` MUST be set to one of the values listed below. 
-`process.middleware.name` SHOULD be set to one of the values listed below. If none of the listed values apply, a custom value best describing the runtime MAY be used.
+`process.middleware.type` MUST be set to one of the values from `as` (application server), `ws` (web server), `dbs` (database server) or `mom` (message-oriented middleware).
+`process.middleware.name` SHOULD be set to the values listed in examples below. If none of the listed values apply, a custom value best describing the runtime MAY be used.
 
 ***Application server middlewares:***
 
--- a/specification/resource/semantic_conventions/process.md
+++ b/specification/resource/semantic_conventions/process.md
@@ -150,7 +150,7 @@ Examples for some Ruby runtimes
 
 ***Application server middlewares:***
 
-Information describing the application server should be captured using the values acquired from the middleware specific API, preferably during runtime. As an example - for Java application servers this information can be captured by invoking `ServletContext.getServerInfo()` during runtime and parsing the result.
+Information describing the application server should be captured using the values acquired from the middleware specific API, preferably during runtime. As an example - for Java application servers supporting Servlet API this information can be captured by invoking `ServletContext.getServerInfo()` during runtime and parsing the result.
 
 Examples for some application server middlewares:
 


Oberon00 (commented):
Please do not add this to process. All resources are process-wide, the middleware is IMHO not really more closely related to the process than, say the faas.instance_id.

CC @tigrannajaryan (cf #882, #1137)

specification/resource/semantic_conventions/process.md
```diff
@@ -131,3 +131,59 @@ Examples for some Ruby runtimes
 | --- | --- | --- | --- |
 | MRI | ruby | 2.7.1 | ruby 2.7.1p83 (2020-03-31 revision a0c7c23c9c) [x86_64-darwin19] |
 | TruffleRuby | truffleruby | 2.6.2 | truffleruby (Shopify) 20.0.0-dev-92ed3059, like ruby 2.6.2, GraalVM CE Native [x86_64-darwin] |
+
+## Process middlewares
+
+**type:** `process.middleware`
+
+**Description:** The single middleware instance which is monitored on the runtime in this process.
```
kenfinnigan, line 139: I'm not sure the description works as it refers to a middleware instance being monitored on a runtime, but the runtimes refer to specific language runtimes that don't always apply to middleware.

specification/resource/semantic_conventions/process.md
```diff
@@ -131,3 +131,59 @@ Examples for some Ruby runtimes
...
+
+| Attribute  | Description  | Example  | Required |
+|---|---|---|--|
+| process.middleware.type | Standardized attribute, specifying the type of middleware on the runtime. This attribute adds value to traces, giving the opportunity to visualise different kind of middleware using different representation in server-side. The supported values of type are `as` (application server), `ws` (web server), `dbs` (database server) or `mom` (message-oriented middleware). | `as` | No |
```
kenfinnigan, line 143: Is a database server middleware?

The means of communicating with a database server, such as JDBC, is middleware, but I'm not sure that the database server itself is.

specification/resource/semantic_conventions/process.md
```diff
@@ -131,3 +131,59 @@ Examples for some Ruby runtimes
...
+| process.middleware.name | The name of the middleware running on this process. | `Tomcat` | No |
+| process.middleware.version | The version of the middleware of this process, as returned by the middleware (preferably during runtime) without modification. | `9.0.39` | No |
+| process.middleware.description | An additional description about the middleware of the process. | `Apache Tomcat/9.0.39` | No |
```
kenfinnigan, line 146: What value above `name` and `version` does `description` offer? From the included example it doesn't appear to be of added value

specification/resource/semantic_conventions/process.md
```diff
@@ -131,3 +131,59 @@ Examples for some Ruby runtimes
...
+
+`process.middleware.type` MUST be set to one of the values listed below.
```
kenfinnigan, line 148: I don't see the list this refers to

specification/resource/semantic_conventions/process.md
```diff
@@ -131,3 +131,59 @@ Examples for some Ruby runtimes
...
+
+| Attribute  | Description  | Example  | Required |
+|---|---|---|--|
+| process.middleware.type | Standardized attribute, specifying the type of middleware on the runtime. This attribute adds value to traces, giving the opportunity to visualise different kind of middleware using different representation in server-side. The supported values of type are `as` (application server), `ws` (web server), `dbs` (database server) or `mom` (message-oriented middleware). | `as` | No |
+| process.middleware.name | The name of the middleware running on this process. | `Tomcat` | No |
+| process.middleware.version | The version of the middleware of this process, as returned by the middleware (preferably during runtime) without modification. | `9.0.39` | No |
+| process.middleware.description | An additional description about the middleware of the process. | `Apache Tomcat/9.0.39` | No |
+
+`process.middleware.type` MUST be set to one of the values listed below. 
+`process.middleware.name` SHOULD be set to one of the values listed below. If none of the listed values apply, a custom value best describing the runtime MAY be used.
```
kenfinnigan, line 149: Same for this list

specification/resource/semantic_conventions/process.md
```diff
@@ -131,3 +131,59 @@ Examples for some Ruby runtimes
...
+
+***Application server middlewares:***
+
+Information describing the application server should be captured using the values acquired from the middleware specific API, preferably during runtime. As an example - for Java application servers this information can be captured by invoking `ServletContext.getServerInfo()` during runtime and parsing the result.
```
kenfinnigan, line 153: What is meant by "middleware specific API"?

The example only works if `ServletContext` is present, which is not guaranteed for Java application servers.

iNikem:
> Please do not add this to process. All resources are process-wide, the middleware is IMHO not really more closely related to the process than, say the faas.instance_id.
> 
> CC @tigrannajaryan (cf #882, #1137)

`faas.instance_id` says something about _outside_ the process, the environment in which this process runs. `middleware` says something about _inside_ the process.

specification/resource/semantic_conventions/process.md
ivomagi: Being new to Otel I might be on a wrong track here as the databases seem to be treated as downstream dependencies vs part of the trace itself. But there is no reason why there could not be agents monitoring the databases themselves and building spans out of the database as well. And for example with Druid and other JVM-based databases, it is rather straightforward as of now. Also the middleware definition lists databases as one of the members of middleware family.

kenfinnigan: Which "middleware definition" are you referring to?

specification/resource/semantic_conventions/process.md
ivomagi: Indeed, but the middleware runs using a specific language runtime - for example Druid database and Apache Tomcat both run on a Java Virtual Machine. Or am I missing the point here?

specification/resource/semantic_conventions/process.md
ivomagi: I guess this is the unfortunate and vague definition of the application server for the Java world. Indeed, in theory, there can be a Java EE application server without the servlet container implementing the Servlet API, but I am aware of none of such implementations. So in practice this example should be guaranteed to work on any Java application server which at minimum seem to always implement the Servlet API (and then some, if they choose so).

specification/resource/semantic_conventions/process.md
kenfinnigan: Not all middleware does, but the statement implies it is.

Kafka, Nginx, and other middleware are not tied to a specific language runtime, at least to my understanding.

specification/resource/semantic_conventions/process.md
kenfinnigan: That's not the case. Quarkus, Helidon, and others can be built and run without the Servlet API, and they're considered an application server

specification/resource/semantic_conventions/process.md
ivomagi: Gotcha. I had built a hierarchical dependency of process-runtime-middleware, but it can be process-middleware or process-runtime-middleware as well. Changed the description to reflect this.

Oberon00:
That is not the point. The resource namespaces should make the name unique, not describe some hierarchy. Also, many app server (is Apache a middleware?) use multiple processes.

tigrannajaryan:
@ivomagi please sign the CLA.

ivomagi:
> [CLA Check](https://api.lfcla.com/v2/repository-provider/github/sign/9522512/184463891/1194)
> 
> * [❌](https://api.lfcla.com/v2/repository-provider/github/sign/9522512/184463891/1194)  ivomagi The commit ([da4425d](https://github.com/open-telemetry/opentelemetry-specification/commit/da4425de8400b77de4ea33537c9d6559162e8efc) ,c5ef9a64e5659520f720763a95bf8d06d6a3198b ,c011cd53f638892a7aa4b117ebad7992c2a52e18) is not authorized under a signed CLA. [Please click here to be authorized](https://api.lfcla.com/v2/repository-provider/github/sign/9522512/184463891/1194). For further assistance with EasyCLA, [please submit a support request ticket](https://jira.linuxfoundation.org/servicedesk/customer/portal/4).

I signed it

specification/resource/semantic_conventions/process.md
ivomagi: Good observation, for the example I used it indeed was seemingly "empty information". Replaced the example, hopefully it became more clear.

specification/resource/semantic_conventions/process.md
ivomagi: Again, unfortunately it is kind-of vague. Quarkus could also be considered to be a library and Helidon seems to run on Netty, which would be the middleware for this case. But expanded the definition and added reference to "Java application servers which support Servlet API" to clarify.

specification/resource/semantic_conventions/process.md
ivomagi: I guess I was too optimistic when refering to it as "definition", as for example Wikipedia version of it does not count the database as middleware and maybe rightfully so. This however does not change the fact that many of the databases (Druid, Cassandra, Derby) do run on a Java runtime and could very well be part of the trace itself (vs downstream dependencies). That being said, maybe Otel does treat databases as not part of the trace at all or treats them special members within the trace?

Oberon00: I think midleware is not really what you mean anyway. It is used by many webframeworks to mean something different, e.g. https://docs.djangoproject.com/en/3.1/topics/http/middleware/ or https://aspnetmonsters.com/2016/03/2016-02-28-what-is-middleware-anyway/

specification/resource/semantic_conventions/process.md
ivomagi: Inlined the reference to "below" to the sentence itself.

specification/resource/semantic_conventions/process.md
ivomagi: This refers to examples in following sections, clarified this in the sentence.

specification/resource/semantic_conventions/process.md
ivomagi: There are numerous traces in the wild internet, trying to define middleware. And maybe I am off the best path here with the naming, but it does not seem to change the need to enhance the spans in the trace with the information about what kind of "middleware" they are running. This is also something that most existing (non-otel-compliant) vendors offer and having this clearly defined in Otel should greatly reduce the friction in migrating from an existing APM vendor to standards-based solutions.

tigrannajaryan:
> That is not the point. The resource namespaces should make the name unique, not describe some hierarchy.

This is correct and has [been our approach](https://github.com/open-telemetry/opentelemetry-specification/pull/789#issuecomment-675506855) so far. Namespaces are not intended to describe a hierarchy of objects. Namespaces are intended to ensure global uniqueness of names. 

There is no need to create nested namespaces just because objects also logically are nested, although it is of course not prohibited.

specification/resource/semantic_conventions/process.md
kenfinnigan: I don't see the change as being any more valuable.

I'm not sure a description field is needed at all, even if it's optional.

anuraaga:
Conceptually these seem to live at the layer of InstrumentationLibrary - they aren't really constant to the app (one app can have multiple middlewares) but can still group some spans together. Does it make sense to add information about the instrumented library in that struct?

Oberon00:
>  they aren't really constant to the app (one app can have multiple middlewares)

I had the impression this PR is about a different kind of "middleware" that is more or less an aditional runtime within the language runtime, such as an Java AppServer (you can have only one per JVM -- at least usually)

ivomagi:
> > [CLA Check](https://api.lfcla.com/v2/repository-provider/github/sign/9522512/184463891/1194)
> > 
> > * [❌](https://api.lfcla.com/v2/repository-provider/github/sign/9522512/184463891/1194)  ivomagi The commit ([da4425d](https://github.com/open-telemetry/opentelemetry-specification/commit/da4425de8400b77de4ea33537c9d6559162e8efc) ,c5ef9a64e5659520f720763a95bf8d06d6a3198b ,c011cd53f638892a7aa4b117ebad7992c2a52e18) is not authorized under a signed CLA. [Please click here to be authorized](https://api.lfcla.com/v2/repository-provider/github/sign/9522512/184463891/1194). For further assistance with EasyCLA, [please submit a support request ticket](https://jira.linuxfoundation.org/servicedesk/customer/portal/4).
> 
> I signed it

I signed it

ivomagi:
> Conceptually these seem to live at the layer of InstrumentationLibrary - they aren't really constant to the app (one app can have multiple middlewares) but can still group some spans together. Does it make sense to add information about the instrumented library in that struct?

Apparently this is the path Steve Flanders also recommended to take in the conversation I was having on the exact time you provided this alternative. Will be closing the PR and will use the InstrumentationLibrary route to send the information.

ivomagi:
Will be using the InstrumentationLibrary.name and InstrumentationLibrary.version attributes which would (in case the OTLP protocol is used) be converted in the collector to otel.library.name and otel.library.version and sent to server.

Oberon00:
> Conceptually these seem to live at the layer of InstrumentationLibrary

While this is true, the values are not suitable for use as InstrumentationLibrary. A new concept on the same layer would be needed, e.g. https://github.com/open-telemetry/oteps/issues/78

ivomagi:
Well, it seems the InstrumentationLibrary.name and InstrumentationLibrary.version is not at the moment used to send this information. It seems that at least Java and PHP agents use these properties to send the module instrumenting the span and the module version. Two examples:
- name: io.opentelemetry.auto.servlet version: 0.10.0-SNAPSHOT
- name: io.opentelemetry.auto.spring-webmvc 0.10.0-SNAPSHOT
So in a sense it seems I am running in circles. For now giving up and sending what is needed as span attributes. Still think it makes a whole lot of sense to have means to send middleware information in standardized way, current approach will create a problems down the road (less valuable traces or incompatible vendors and redundancy)

PR #1197, opened 2020-11-05
Add guidance on when to use and not use nested namespaces

tigrannajaryan:
This topic has come up at least 3 times now. I believe a clarification is
warranted. The clarification is aligned with previous recommendations:
https://github.com/open-telemetry/opentelemetry-specification/pull/789#issuecomment-675506855
https://github.com/open-telemetry/opentelemetry-specification/pull/882#issuecomment-690945900
https://github.com/open-telemetry/opentelemetry-specification/pull/1194#issuecomment-722431982

### DIFF
--- a/specification/common/common.md
+++ b/specification/common/common.md
@@ -62,6 +62,10 @@ Names SHOULD follow these rules:
 - Namespaces can be nested. For example `telemetry.sdk` is a namespace inside
   top-level `telemetry` namespace and `telemetry.sdk.name` is an attribute
   inside `telemetry.sdk` namespace.
+  Note: the fact that an entity is located within another entity is typically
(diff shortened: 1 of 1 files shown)


Oberon00 (approved):
Would it make sense to specifically call out the `process` namespace here, which seems to be rather popular?

cijothomas (approved):

tigrannajaryan:
> Would it make sense to specifically call out the `process` namespace here, which seems to be rather popular?

I am not sure how exactly to do it nicely. Can be improved in a future PR by someone else.

arminru (approved):

bogdandrutu (approved):

Issue #731, opened 2020-07-23
Changing OTLP exporters to export cumulative metrics by default


huyan0:
*What are you trying to achieve?*
Currently, OTLP exporters in SDKs are pass-through and send delta metrics directly. This makes Prometheus Exporter in the collector exports delta counter values as "gauge". Because of this behavior of the SDK exporter, it is then up to the collector to perform aggregations before exporting to cumulative backends like Prometheus. This feature does not yet exist, but has been proposed in [#1422] (https://github.com/open-telemetry/opentelemetry-collector/issues/1422) in the collector repository.

However, in the case of multiple collector instances behind a load balancer, each individual collector possibly cannot aggregate cumulative values correctly, as metric events could be distributed across collectors. 

If OTLP exporters export cumulative by default, then it guarantees that events from the same metric or timeseries are aggregated correctly before sending to the collector. The collector could then correctly export the metric to cumulative backends.

*Additional context*
related issues: Collector [#1422](https://github.com/open-telemetry/opentelemetry-collector/issues/1422), [#1255](https://github.com/open-telemetry/opentelemetry-collector/issues/1255)

cc @alolita @huyan0 @jmacd

jmacd:
Following the discussion at yesterday's Metric SIG, I believe our best step forward is:

1. State the OTLP can be configured for Cumulative, Passthrough, or Delta by default. The machinery for this is already demonstrated by the Metrics Processor component in https://github.com/open-telemetry/opentelemetry-go/pull/840. I see this as a requirement that OTLP support both delta and cumulative for all kinds of metric. See https://github.com/open-telemetry/opentelemetry-proto/pull/168 for an OTLP proposal that meets this requirement.
2. Declare an environment variable or configuration field that determines this setting for OTLP
3. Set the default OTLP behavior to Cumulative, which gives status-quo behavior for Prometheus users
4. Add support in the Collector, when configured as an Agent, to support converting Passthrough OTLP to Cumulative OTLP. The assumption that there is a single agent allows this to work.
5. Collector exporters that wish to expose Deltas are impacted by this, as it's difficult to coordinate correctly in a horizontally scaled pool of collectors. These clients will have to be configured for DELTA OTLP.
6. Future possibility: the OTLP client would contact the service for a dynamic configuration and use the best option at startup.

bogdandrutu:
[...]

PR #1198, opened 2020-11-05
Metrics SDK: Accumulator and Processor state requirements

jmacd:
## Changes

Introduces basic requirements for the Processor and adds a missing requirement for the Accumulator.

Related issues #731 
Relates to https://github.com/open-telemetry/opentelemetry-java/issues/1976 and https://github.com/open-telemetry/opentelemetry-java/pull/1978

Part of #2000.


specification/metrics/sdk.md
```diff
@@ -268,9 +276,75 @@ Label Set, Resource, and metric Descriptor.
...
+
+##### Basic Processor: StatelessExportKindSelector
+
+The StatelessExportKindSelector configures a Metric export pipeline
+with no long-term memory requirements.  In this selector, the Counter,
+UpDownCounter, ValueRecorder, and ValueObserver instruments are
+configured for Delta aggregation temporality while SumObserver and
+UpDownSumObserver instruments are configured for Cumulative
+aggregation temporality.  This basic Processor configuration has no
+long-term memory requirements because each instrument is
```
jkwatson, line 330: I'm not sure what you mean by the instrument being "passed-through" here. Should this be "measurements", rather than "instrument"?

jkwatson: or "recordings" I guess?

specification/metrics/sdk.md
```diff
@@ -250,6 +250,14 @@ synchronous instrument updates.  The Accumulator SHOULD NOT hold an
 exclusive lock while calling an Aggregator (see below), since some
 Aggregators may have higher concurrency expectations.
 
+State managed in the Accumulator MUST be transient.  This requirement
+ensures that export pipelines written constructed for stateless
+exporters (e.g. Statsd, OTLP with a stateless ExportKindSelector) are
+not penalized by permanent state in the Accumulator.  This implies
+that the use of long-term state in a Metrics export pipeline should be
+elective, and such state if present should be managed in the Processor
+component.
```
jkwatson, line 259: I'm missing something here. The Accumulator holds on to a Map of aggregators. If the aggregations that are required are cumulative, how can the Accumulator not be holding on to the aggregators that are maintaining the state of the cumulative aggregations?

specification/metrics/sdk.md
```diff
@@ -268,9 +276,75 @@ Label Set, Resource, and metric Descriptor.
...
+The Processor component interface supports interaction from the
+Controller and Accumulator.  The Controller, responsible for
+initiating a new round of collection, informs the Processor when a
+collection interval starts and finishes.  After finishing the
+collection interval, the Controller gets the ExportRecordSet before
+calling the Exporter to export data.
+
+The Accumulator interacts with the Processor during the call to its
+`Collect()` method, during which it calls `Process()` once per
+ExportRecord on the Processor.
```
jkwatson, line 290: Is there a reason why the Controller doesn't mediate this? Couldn't the controller ask the accumulator for the data, then pass it to the Processor, rather than have the Accumulator know about the processor directly? What do we gain from the extra coupling between Accumulator and Processor?

specification/metrics/sdk.md
jkwatson: Trying to read between the lines, since we don't really have a description of how the Processor works yet...but does the Processor have it's own long-term, stateful aggregators, which keep the cumulative state state, and compute it, when needed based on the deltas that are passed in by the Accumulator?

Is this paragraph basically saying that the aggregators in the Accumulator MUST only keep delta-state (i.e. one collection interval's worth of state), and that any cumulative state is owned by the Processor?

jmacd: I could fill in more of the "model implementation" below, I guess, to make this clear. The idea is that the Accumulator will need a high-concurrency map of instrument/labelset in order to compute accumulations. For the synchronous instruments, these are accumulated in DELTA form.

This PR is trying to assert that the Processor component is the only one permitted to keep state, and it means the Processor also owns a map where it keeps long-term state. In the OTel-Go implementation, this map is built whether or not there is a state requirement, but it's because of other requirements that are not explicit here, even if implied.

In order to correctly emit CUMULATIVE values, the Processor is required to merge aggregations with equal instrument/labelset values, therefore a map is needed. This may sound implied by the Accumulator requirements, and indeed I believe is **when there is a single Accumulator**. We have not directly addressed the question of whether a Processor is required to support a single Accumulator or not. As implemented in the OTel-Go SDK, the basic Processor naturally supports multiple Accumulators. This is seen as a way to support multiple Resources from within a single Process--I mention this as an undecided topic for debate.

specification/metrics/sdk.md
jmacd: The Controller does mediate this--it's the exclusive "owner" of the pipeline and mediates everything, I'd say. Why ask the Accumulator to pass data to the Processor? The Processor has lots of flexibility this way, e.g., to simply ignore Accumulations.

specification/metrics/sdk.md
jmacd: I'll try to update the wording here. I'm trying to say that when the Accumulator has DELTA Accumluations (e.g., from Counters) the output will be AggregationTemporality=DELTA, and likewise when the Accumulator has CUMULATIVE values (e.g., from SumObservers) the output will be AggregationTemporality=CUMULATIVE. This configuration requires no long-term state in the Processor, as a result.

specification/metrics/sdk.md
jkwatson: My question is...why not have the controller pass this data to the processor, and decouple the accumulator from the processor?

specification/metrics/sdk.md
jmacd: The comment above wasn't very useful--mostly a tangent--so I struck it out. 

I re-read your feedback @jkwatson and understood partly what was needed. I've rewritten this section to be a lot more clear, I hope.

specification/metrics/sdk.md
jmacd: After thinking about this, I guess there's no reason to require the Accumulator to call the Processor directly. The Processor could instead implement an iterator pattern, to let the Processor scan through a collection representing the Accumulator state. 

Is that a better design? I'm not sure the code will be cleaner for me in Go--if I'm to avoid allocating an intermediate data structure--but it sounds logically cleaner.

specification/metrics/sdk.md
jmacd: Fixed, I think.

jmacd:
@open-telemetry/specs-metrics-approvers Please take a loot at this updated spec about Accumulator and Processor state.

specification/metrics/sdk.md
jkwatson: I'll ponder in my nascent java implementation what the appropriate degree of coupling might be. @breedx-splk this is something we should think about.

specification/metrics/sdk.md
```diff
@@ -250,6 +268,18 @@ synchronous instrument updates.  The Accumulator SHOULD NOT hold an
...
 
+The Accumulator MUST eliminate from memory (i.e. "forget") state
+associated with label sets used in earlier collection intervals, after
+they are not for a suitable amount of time.  A "suitable amount of
+time" is intentionally not specific, since implementations may wish to
+optimize memory management and have to contend with concurrent access.
+This requirement ensures that export pipelines constructed for
+stateless exporters (e.g. Statsd, OTLP with a stateless
+ExportKindSelector) are not forced into the use of permanent state in
+the Accumulator.  This implies that the use of long-term state in a
```
jkwatson, line 279: It would be great if the details of how the golang implementation manages this were included in the detailed description of the model implementation. [I think this is done via some clever reference counting, IIRC].

PR #1202, opened 2020-11-06
Add OTEL_TRACE_SAMPLER_ARG env variable definition.

### LINKED ISSUE #1105: Add OTEL_SAMPLING_PROBABILITY (or similar named) to sdk environment variable spec

carlosalberto:
Fixes #1105 
Closes #1190 (alternative solution)

## Changes

It was discussed in the latest TC meeting trying out the approach taken by [Jaeger](https://github.com/jaegertracing/jaeger-client-java/blob/master/jaeger-core/src/main/java/io/jaegertracing/Configuration.java#L122), which uses two environment variables to configure samplers:

1) JAEGER_SAMPLER_TYPE
2) JAEGER_SAMPLER_PARAM: the Sampler-specific parameters (as defined by each specific type).

This way, each `Sampler` will use the same environment variable, and we could even extend the future expected input, i.e. accept json.

### DIFF
--- a/specification/sdk-environment-variables.md
+++ b/specification/sdk-environment-variables.md
@@ -24,9 +24,9 @@ Known values for `OTEL_TRACE_SAMPLER` are:
 - `"parentbased_always_off"`: `ParentBased(root=AlwaysOffSampler)`
 - `"parentbased_traceidratio"`: `ParentBased(root=TraceIdRatioBased)`
 
-Known values for `OTEL_TRACE_SAMPLER_ARG` are:
+Depending on the value of `OTEL_TRACE_SAMPLER`, `OTEL_TRACE_SAMPLER_ARG` may be set as follows:
 
-- `traceidratio` and `parentbased_traceidratio`: Sampling probability, a number in the [0..1] range, e.g. "0.25".
+- For `traceidratio` and `parentbased_traceidratio` samplers: Sampling probability, a number in the [0..1] range, e.g. "0.25".
 
 ## Batch Span Processor
 
--- a/specification/sdk-environment-variables.md
+++ b/specification/sdk-environment-variables.md
@@ -10,7 +10,7 @@ The goal of this specification is to unify the environment variable names betwee
 | OTEL_LOG_LEVEL           | Log level used by the SDK logger                  | "info"                            |                                     |
 | OTEL_PROPAGATORS         | Propagators to be used as a comma separated list  | "tracecontext,baggage"            | Values MUST be deduplicated in order to register a `Propagator` only once. Unrecognized values MUST generate a warning and be gracefully ignored. |
 | OTEL_TRACE_SAMPLER       | Sampler to be used for traces                     | "parentbased_always_on"                       | See [Sampling](./trace/sdk.md#sampling) |
-| OTEL_TRACE_SAMPLER_ARG   | String value to be used as the sampler argument   |                                   | The specified value will only be used if OTEL_TRACE_SAMPLER is set. Each Sampler type defines its own expected input, if any. Invalid or unrecognized input MUST be logged and MUST NOT result in an error.  |
+| OTEL_TRACE_SAMPLER_ARG   | String value to be used as the sampler argument   |                                   | The specified value will only be used if OTEL_TRACE_SAMPLER is set. Each Sampler type defines its own expected input, if any. Invalid or unrecognized input MUST be logged and MUST be otherwise ignored, i.e. the SDK MUST behave as if OTEL_TRACE_SAMPLER_ARG is not set.  |
 
 Known values for OTEL_PROPAGATORS are: "tracecontext", "baggage", "b3", "jaeger".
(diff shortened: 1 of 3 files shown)


yurishkuro (approved):
lgtm

anuraaga (approved):

carlosalberto:
@open-telemetry/specs-approvers Please review this issue (will merge by EOD Monday).

Oberon00 (approved):
Nice solution!

specification/sdk-environment-variables.md
```diff
@@ -23,6 +24,10 @@ Known values for `OTEL_TRACE_SAMPLER` are:
 - `"parentbased_always_off"`: `ParentBased(root=AlwaysOffSampler)`
 - `"parentbased_traceidratio"`: `ParentBased(root=TraceIdRatioBased)`
 
+Known values for `OTEL_TRACE_SAMPLER_ARG` are:
+
+- `traceidratio` and `parentbased_traceidratio`: Sampling probability, a number in the [0..1] range, e.g. "0.25".
```
Oberon00, line 29: This is worded strangely: traceidratio is a known value for OTEL_TRACE_SAMPLER_ARG? Then what happens if I set OTEL_TRACE_SAMPLER_ARG=traceidratio? Please describe more clearly what is meant here, e.g. 

```suggestion
Depending on the value of `OTEL_TRACE_SAMPLER`, `OTEL_TRACE_SAMPLER_ARG` may be set as follows:

- For `traceidratio` and `parentbased_traceidratio` samplers: Sampling probability, a number in the [0..1] range, e.g. "0.25".
```

carlosalberto: Thanks for the clarification change!

arminru (approved):

specification/sdk-environment-variables.md
```diff
@@ -10,6 +10,7 @@ The goal of this specification is to unify the environment variable names betwee
 | OTEL_LOG_LEVEL           | Log level used by the SDK logger                  | "info"                            |                                     |
 | OTEL_PROPAGATORS         | Propagators to be used as a comma separated list  | "tracecontext,baggage"            | Values MUST be deduplicated in order to register a `Propagator` only once. Unrecognized values MUST generate a warning and be gracefully ignored. |
 | OTEL_TRACE_SAMPLER       | Sampler to be used for traces                     | "parentbased_always_on"                       | See [Sampling](./trace/sdk.md#sampling) |
+| OTEL_TRACE_SAMPLER_ARG   | String value to be used as the sampler argument   |                                   | The specified value will only be used if OTEL_TRACE_SAMPLER is set. Each Sampler type defines its own expected input, if any. Invalid or unrecognized input MUST be logged and MUST NOT result in an error.  |
```
tigrannajaryan, line 13: > and MUST NOT result in an error. 

Can you please clarify this? What does it mean to not result in an error?

tigrannajaryan (approved):
LGTM, just one minor comment left.

specification/sdk-environment-variables.md
bogdandrutu: I think the idea is to not crash the app, what is the fallback? go back to the default api recommended.

carlosalberto: Exactly, the idea is that this won't crash the app (no Exception thrown at all, that is), but report the invalid value in the usual logging facility we have used in other places. 

@tigrannajaryan any suggestion on how to re-phrase this to make it clear?

yurishkuro, line 13:
```suggestion
| OTEL_TRACE_SAMPLER_ARG   | String value to be used as the sampler argument   |                                   | The specified value will only be used if OTEL_TRACE_SAMPLER is set. Each Sampler type defines its own expected input, if any. Invalid or unrecognized input MUST be logged and treated as unset, i.e. MUST NOT result in an error.  |
```

However, I think handling of invalid input should be described once in this doc since it should apply to all variables.

tigrannajaryan: How about this:
```suggestion
| OTEL_TRACE_SAMPLER_ARG   | String value to be used as the sampler argument   |                                   | The specified value will only be used if OTEL_TRACE_SAMPLER is set. Each Sampler type defines its own expected input, if any. Invalid or unrecognized input MUST be logged and MUST be otherwise ignored, i.e. the SDK MUST behave as if OTEL_TRACE_SAMPLER_ARG is not set.  |
```

carlosalberto: @yurishkuro Applied @tigrannajaryan's (as I saw his change first), but I think it clarifies things in the same way. Let me know otherwise.

(and will create an issue to  follow up, mentioning invalid values should be ignored and be treated as unset).

bogdandrutu (approved):

PR #1208, opened 2020-11-08
Use MUST consistently instead of shall.

anuraaga:
Randomly noticed this - removes usage of SHALL for MUST which is much more common in our spec, and means the same. Consistency can reduce some cognitive load.

### DIFF
--- a/specification/trace/semantic_conventions/messaging.md
+++ b/specification/trace/semantic_conventions/messaging.md
@@ -75,7 +75,7 @@ Often such destinations are unnamed or have an auto-generated name.
 
 ## Conventions
 
-Given these definitions, the remainder of this section describes the semantic conventions that shall be followed for Spans describing interactions with messaging systems.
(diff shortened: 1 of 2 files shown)


specification/trace/semantic_conventions/messaging.md
```diff
@@ -75,7 +75,7 @@ Often such destinations are unnamed or have an auto-generated name.
 
 ## Conventions
 
-Given these definitions, the remainder of this section describes the semantic conventions that shall be followed for Spans describing interactions with messaging systems.
+Given these definitions, the remainder of this section describes the semantic conventions that MUST be followed for Spans describing interactions with messaging systems.
```
Oberon00, line 78: I think this shoud actually a SHOULD or 
```suggestion
Given these definitions, the remainder of this section describes the semantic conventions for Spans describing interactions with messaging systems.
```

Oberon00 (approved):

yurishkuro (approved):

PR #1209, opened 2020-11-08
Add default port column to database table

hasanfd:
There is a requirement that **net.peer.port** will be added only if the port is different than the default port. We need to know the default ports to implement this feature. A single source of default ports will help teams that work on different languages.

### DIFF
--- a/specification/trace/semantic_conventions/database.md
+++ b/specification/trace/semantic_conventions/database.md
@@ -49,51 +49,51 @@ Some database systems may allow a connection to switch to a different `db.user`,
 
 `db.system` MUST be one of the following or, if none of the listed values apply, a custom value:
 
-| Value  | Description |
-|---|---|
-| `other_sql` | Some other SQL database. Fallback only. See notes. |
-| `mssql` | Microsoft SQL Server |
-| `mysql` | MySQL |
-| `oracle` | Oracle Database |
-| `db2` | IBM Db2 |
-| `postgresql` | PostgreSQL |
-| `redshift` | Amazon Redshift |
-| `hive` | Apache Hive |
-| `cloudscape` | Cloudscape |
-| `hsqlsb` | HyperSQL DataBase |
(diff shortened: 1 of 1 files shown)


Oberon00 (commented):
While it might be useful for backends to know the default port per DB in absence of net.peer.port, I think there is a misunderstanding here.

> There is a requirement that net.peer.port will be added **only if** the port is different than the default port.

This is wrong. Correct would be

> There is **only a requirement** that net.peer.port will be added if the port is different than the default port.

So it is perfectly fine to add the net.peer.port attribute even if it equals the default port.

specification/trace/semantic_conventions/database.md
```diff
@@ -49,51 +49,51 @@ Some database systems may allow a connection to switch to a different `db.user`,
...
-| `hive` | Apache Hive |
-| `cloudscape` | Cloudscape |
-| `hsqlsb` | HyperSQL DataBase |
-| `progress` | Progress Database |
-| `maxdb` | SAP MaxDB |
-| `hanadb` | SAP HANA |
-| `ingres` | Ingres |
+| Value  | Description | Default port
+|---|---|---|
+| `other_sql` | Some other SQL database. Fallback only. See notes. | -
```
Oberon00, line 54: If you want to add the port to the enum (not sure if that is the best way) you'll need to add it to the `note` in the YAML where this is generated from: https://github.com/open-telemetry/opentelemetry-specification/blob/250721606406d51657ee8e1d5655d7beb87a8d18/semantic_conventions/trace/database.yaml#L14-L143 and then run `make table-generation`.

hasanfd: If that's the case, I guess, it is more straightforward to put net.port.peer to every span. I doubt that we would have difficulties to find net.port.peer if the library uses the default port. So knowing default ports does not ease the job. 

Thanks for the comments @Oberon00 :) I am closing the PR.

PR #1211, opened 2020-11-09
Add requirement section specification

### LINKED ISSUE #1210: Make it possible to automatically list all specification requirements
This is the original content of this issue:

*What are you trying to achieve?*
I am trying to quickly make a list of specification requirements in order to check the specification compliance of a particular implementation.

*What did you expect to see?*
I was expecting to easily find all the requirements that make up the specification.

*Additional context.*
I have been reading some parts of the specification and found that the usage of [RFC 2119](https://tools.ietf.org/html/rfc2119) keywords is sometimes inconsistent, some hard requirements are not specified with MUST, some optional requirements are not specified with SHOULD/MAY. Instead, some requirements are written using other terms that have the same meaning but not the exact syntax of RFC 2119. This makes it impossible to simply search for these keywords in the specification in order to find all the precise requirements that make up the document.

@jmacd [requested](https://github.com/open-telemetry/opentelemetry-specification/pull/2003#issuecomment-942569789) an issue to be filed. I prefer to reuse this one to minimize the amount of issues :+1:

### Motivation for this change

The spec is hard to understand because it is written in an inconsistently-applied formal language. Here are some examples:

- [This is a "Recommendations" section where MUST and SHOULD appear both](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-naming.md#recommendations-for-opentelemetry-authors)
- [This section states that a name SHOULD SHOULD NOT coincide with namespaces](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-naming.md#attribute-naming)
- [This section uses the terms "this is required for..." instead of using MUST](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attributes)
- [This section includes conditional requirements. It is hard to find which requirements an implementation has to fulfill if it requires finding any conditions in the specification first](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attribute-limits)
- [This section uses may in lower case](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/opencensus.md#opencensus-binary-context)

When implementing OpenTelemetry it is very hard to know what has to be implemented because there is no clear listing of all requirements. If the specifcation is written in the way intended in this PR it will be possible to generate automatically a list of every requirement and condition that the specification states.

It is also much easier to check that the specification is correct if we only need to check a list of well defined sections. Some checks (like having a RFC 2119 keyword, for example) can even be automated.

It is also easier to communicate between OpenTelemetry participants if the requirements are clearly identifiable, this approach uses markdown headings so that each requirement will have its own URL so that they can be used as reference in Slack, email, etc.

### Steps

Fixing this issue will require a big change, because it means reformatting every document that has been written to this moment, so it would be convenient to separate this into smaller steps:

- [ ] Merge the [specification-writing guidelines](https://github.com/open-telemetry/opentelemetry-specification/pull/2003/files#diff-0219b1a94eb4180d8dfca3754feb1363702cbef0d41f0672c7dce85df96bfa2c)
- [ ] Update the specification document status states to manage the transition
- [ ] Merge the scripts that generate the JSON files from the specification markdown files
- [ ] Reformat common/attribute-naming.md
- [ ] Reformat common/common.md
- [ ] Reformat sdk-configuration.md
- [ ] Reformat context/api-propagators.md
- [ ] Reformat context/context.md
- [ ] Reformat overview.md
- [ ] Reformat protocol/exporter.md
- [ ] Reformat protocol/otlp.md
- [ ] Reformat protocol/design-goals.md
- [ ] Reformat protocol/requirements.md
- [ ] Reformat performance-benchmark.md
- [ ] Reformat vendors.md
- [ ] Reformat logs/overview.md
- [ ] Reformat logs/data-model.md
- [ ] Reformat sdk-environment-variables.md
- [ ] Reformat compatibility/openmetrics.md
- [ ] Reformat compatibility/opentracing.md
- [ ] Reformat compatibility/opencensus.md
- [ ] Reformat resource/sdk.md
- [ ] Reformat resource/semantic_conventions/device.md
- [ ] Reformat resource/semantic_conventions/cloud_provider/aws/logs.md
- [ ] Reformat resource/semantic_conventions/cloud_provider/aws/README.md
- [ ] Reformat resource/semantic_conventions/cloud_provider/aws/eks.md
- [ ] Reformat resource/semantic_conventions/cloud_provider/aws/ecs.md
- [ ] Reformat resource/semantic_conventions/webengine.md
- [ ] Reformat resource/semantic_conventions/os.md
- [ ] Reformat resource/semantic_conventions/README.md
- [ ] Reformat resource/semantic_conventions/deployment_environment.md
- [ ] Reformat resource/semantic_conventions/k8s.md
- [ ] Reformat resource/semantic_conventions/host.md
- [ ] Reformat resource/semantic_conventions/process.md
- [ ] Reformat resource/semantic_conventions/faas.md
- [ ] Reformat resource/semantic_conventions/container.md
- [ ] Reformat resource/semantic_conventions/cloud.md
- [ ] Reformat versioning-and-stability.md
- [ ] Reformat baggage/api.md
- [ ] Reformat performance.md
- [ ] Reformat library-guidelines.md
- [ ] Reformat metrics/supplementary-guidelines.md
- [ ] Reformat metrics/sdk.md
- [ ] Reformat metrics/api.md
- [ ] Reformat metrics/README.md
- [ ] Reformat metrics/datamodel.md
- [ ] Reformat metrics/sdk_exporters/in-memory.md
- [ ] Reformat metrics/sdk_exporters/prometheus.md
- [ ] Reformat metrics/sdk_exporters/stdout.md
- [ ] Reformat metrics/sdk_exporters/otlp.md
- [ ] Reformat metrics/semantic_conventions/runtime-environment-metrics.md
- [ ] Reformat metrics/semantic_conventions/system-metrics.md
- [ ] Reformat metrics/semantic_conventions/rpc.md
- [ ] Reformat metrics/semantic_conventions/openmetrics-guidelines.md
- [ ] Reformat metrics/semantic_conventions/faas-metrics.md
- [ ] Reformat metrics/semantic_conventions/README.md
- [ ] Reformat metrics/semantic_conventions/http-metrics.md
- [ ] Reformat metrics/semantic_conventions/process-metrics.md
- [ ] Reformat library-layout.md
- [ ] Reformat trace/sdk.md
- [ ] Reformat trace/api.md
- [ ] Reformat trace/sdk_exporters/zipkin.md
- [ ] Reformat trace/sdk_exporters/jaeger.md
- [ ] Reformat trace/sdk_exporters/non-otlp.md
- [ ] Reformat trace/semantic_conventions/instrumentation/aws-sdk.md
- [ ] Reformat trace/semantic_conventions/instrumentation/aws-lambda.md
- [ ] Reformat trace/semantic_conventions/http.md
- [ ] Reformat trace/semantic_conventions/database.md
- [ ] Reformat trace/semantic_conventions/rpc.md
- [ ] Reformat trace/semantic_conventions/exceptions.md
- [ ] Reformat trace/semantic_conventions/messaging.md
- [ ] Reformat trace/semantic_conventions/README.md
- [ ] Reformat trace/semantic_conventions/faas.md
- [ ] Reformat trace/semantic_conventions/span-general.md
- [ ] Reformat error-handling.md

ocelotl:
Fixes #1210

## Changes

Adds a new way of specifying requirements. More details [here](https://github.com/open-telemetry/opentelemetry-specification/pull/1211/files?short_path=8adea91#diff-8adea912606edd980007df4620cadd058978b993f7b67bc3bf31576b0e7fc872). Already added requirement sections for the metrics api and sdk md files, more can be added later.


Oberon00 (dismissed):
The requirement matrix is supposed to (mostly) solve this: https://github.com/open-telemetry/opentelemetry-specification/blob/master/spec-compliance-matrix.md I guess it would be important to add links from the rows to the corresponding spec item though.

While I think a bit smaller linkable sections would be nice, I think in the current form, this PR makes the spec harder to read (and write).

In my opinion, if we want such a highly structured format, Markdown is not really suitable and we would have to move to something else, e.g. Asciidoc, ReStructuredText, YAML and move the spec from being viewable as-is on GH to be displayed on GH pages after some generation process.

ocelotl:
> The requirement matrix is supposed to (mostly) solve this: https://github.com/open-telemetry/opentelemetry-specification/blob/master/spec-compliance-matrix.md I guess it would be important to add links from the rows to the corresponding spec item though.
> 
I don't think the requirement matrix solves the issue this PR solves. How does the requirement matrix handle the different RFC 2119 keywords that can show up in the specification? That matrix it says nothing of the feature in the implementation being mandatory (I mean a MUST requirement) or not (I mean a MAY requirement). This makes it not possible for the reader to know how actually complete an implementation is because in a certain implementation, a MAY requirement can be marked as `+` while a MUST requirement can be marked as `-` but the requirement matrix reader can't tell that the implementation lacks a mandatory requirement because that information is not present there.

> While I think a bit smaller linkable sections would be nice, I think in the current form, this PR makes the spec harder to read (and write).
> 
I think it makes the specification easier to read. For example, let's consider someone who is looking for all the mandatory requirements in a certain document because that person wants to make sure a certain implementation has everything that MUST be implemented, anything that MAY be implemented is of no interest for this person. The way the specification is written right now makes it necessary for this person to read and understand the whole specification document when looking for MUST requirements because there is no check that makes sure that all the things that are mandatory are properly specified with a MUST keyword. The specification can say "a span MUST have a name" and the specification can also say "it is mandatory that every span has a name". Both statements are semantically equivalent but the fact that the second one can be present in the specification as it is now makes it necessary for the reader to read and understand the document of interest completely to make sure no mandatory requirement has been missed. If the specification was written how this PR suggests, it would be very easy for the specification reader to find the mandatory requirements, as it would only be a matter of looking for the occurrences of MUST.

I think it makes the specification a bit harder to write. I mean that it is an additional effort for the specification developer to add a requirement section with a proper RFC 2119 keyword in order to make sure that the requirement that is added is clear and specific. Nevertheless, that is how specifications should be :slightly_smiling_face: 

> In my opinion, if we want such a highly structured format, Markdown is not really suitable and we would have to move to something else, e.g. Asciidoc, ReStructuredText, YAML and move the spec from being viewable as-is on GH to be displayed on GH pages after some generation process.

I also think Markdown is not the perfect file format for this. Nevertheless, the main change I want to introduce in this PR is the strict specification of requirements to make it possible to extract them programatically and also check them in the same way. The format we use is secondary, so I did not want to introduce a change that is not central to this PR here. I think Markdown is good enough if we use it as it is suggested here (it seems like it may be necessary to adapt the CI tests for these changes to be accepted, though). Github is capable of rendering [RestructuredText](https://github.com/HPENetworking/topology/blob/master/README.rst) and [Asciidoc](https://github.com/asciidoc/asciidoc-py3/blob/master/README.asciidoc) directly without any additional process.

yurishkuro:
> "a span MUST have a name" ... "it is mandatory that every span has a name". Both statements are semantically equivalent

I think such instances MUST be corrected to use MUST.

> strict specification of requirements to make it possible to extract them programmatically and also check them in the same way

You'd be able to _enumerate_ them programmatically, but that's about it, isn't it?  Someone still needs to interpret the text.

ocelotl:
> > "a span MUST have a name" ... "it is mandatory that every span has a name". Both statements are semantically equivalent
> 
> I think such instances MUST be corrected to use MUST.
> 
> > strict specification of requirements to make it possible to extract them programmatically and also check them in the same way
> 
> You'd be able to _enumerate_ them programmatically, but that's about it, isn't it? Someone still needs to interpret the text.

Correct, someone still needs to interpret the text. I think enumerating the requirements is an advantage in itself as it allows for the implementation developers to have a clear list of requirements that can be used to measure compliance with the specification.

ocelotl:
> > The requirement matrix is supposed to (mostly) solve this: https://github.com/open-telemetry/opentelemetry-specification/blob/master/spec-compliance-matrix.md I guess it would be important to add links from the rows to the corresponding spec item though.
> 
> I don't think the requirement matrix solves the issue this PR solves. How does the requirement matrix handle the different RFC 2119 keywords that can show up in the specification? That matrix it says nothing of the feature in the implementation being mandatory (I mean a MUST requirement) or not (I mean a MAY requirement). This makes it not possible for the reader to know how actually complete an implementation is because in a certain implementation, a MAY requirement can be marked as `+` while a MUST requirement can be marked as `-` but the requirement matrix reader can't tell that the implementation lacks a mandatory requirement because that information is not present there.
> 
> > While I think a bit smaller linkable sections would be nice, I think in the current form, this PR makes the spec harder to read (and write).
> 
> I think it makes the specification easier to read. For example, let's consider someone who is looking for all the mandatory requirements in a certain document because that person wants to make sure a certain implementation has everything that MUST be implemented, anything that MAY be implemented is of no interest for this person. The way the specification is written right now makes it necessary for this person to read and understand the whole specification document when looking for MUST requirements because there is no check that makes sure that all the things that are mandatory are properly specified with a MUST keyword. The specification can say "a span MUST have a name" and the specification can also say "it is mandatory that every span has a name". Both statements are semantically equivalent but the fact that the second one can be present in the specification as it is now makes it necessary for the reader to read and understand the document of interest completely to make sure no mandatory requirement has been missed. If the specification was written how this PR suggests, it would be very easy for the specification reader to find the mandatory requirements, as it would only be a matter of looking for the occurrences of MUST.
> 
> I think it makes the specification a bit harder to write. I mean that it is an additional effort for the specification developer to add a requirement section with a proper RFC 2119 keyword in order to make sure that the requirement that is added is clear and specific. Nevertheless, that is how specifications should be 
> 
> > In my opinion, if we want such a highly structured format, Markdown is not really suitable and we would have to move to something else, e.g. Asciidoc, ReStructuredText, YAML and move the spec from being viewable as-is on GH to be displayed on GH pages after some generation process.
> 
> I also think Markdown is not the perfect file format for this. Nevertheless, the main change I want to introduce in this PR is the strict specification of requirements to make it possible to extract them programatically and also check them in the same way. The format we use is secondary, so I did not want to introduce a change that is not central to this PR here. I think Markdown is good enough if we use it as it is suggested here (it seems like it may be necessary to adapt the CI tests for these changes to be accepted, though). Github is capable of rendering [RestructuredText](https://github.com/HPENetworking/topology/blob/master/README.rst) and [Asciidoc](https://github.com/asciidoc/asciidoc-py3/blob/master/README.asciidoc) directly without any additional process.

For the sake of completeness, [here](https://github.com/github/markup#markups) are the supported Github markups, @Oberon00 :+1:

Oberon00:
Even though Github supports basic ReStructuredText, the real usefulness of rst would lie in the ability to write custom markup like `.. requirement` which I don't think GitHub supports. I'm not sure how AsciiDoc works here.

ocelotl:
> Even though Github supports basic ReStructuredText, the real usefulness of rst would lie in the ability to write custom markup like `.. requirement::` which I don't think GitHub supports.

Yes, but that is actually not central to this PR. The way that we can display the requirement sections in Github is possible with the code in this PR (although it is less than ideal), but the main idea of this PR is to separate the _hard_ requirements into their own sections from the rest of the specification so that they can be identified quickly and processed by a parser.

ocelotl:
Is it possible to reopen this PR? I don't seem able to do so. @carlosalberto

ocelotl:
Thanks @Oberon00 for reopening!

Looking around I found [this](https://www.w3.org/TR/qaframe-spec/#consistent-style). Funny thing, it pretty much includes everything I am suggesting in this PR. This W3C specification says: _It is important for readers to be able to differentiate requirements in the specification from non-requirements in order to either implement or review them._

This is exactly what I am attempting to do here :sweat_smile: by separating the specification parts that include an RFC Keyword into their own separately-labeled sections.

It even mentions this: _It will be easier to extract conformance requirements and better for accessibility_. This is precisely what I want with the parsing tool that extracts the requirements into a JSON file.

I'll have to take a better look at this document, probably everything that I have imagined for this PR is already well defined there.

mattmccleary:
+1 for aligning our "shalls", "musts", "required", "recommended", etc. to standards.

Can we start here, which I think is the same suggestion as @ocelotl above?
https://www.ietf.org/rfc/rfc2119.txt

For a short time, I worked on construction specifications, and I used to get my hand slapped for using "must" instead of "shall". As an emerging discipline, software conventions seem less established, and we can learn from the traditional engineering disciplines in tightening up our language.

Oberon00:
> +1 for aligning our "shalls", "musts", "required", "recommended", etc. to clearly defined standards.

We already have this in theory, see https://github.com/open-telemetry/opentelemetry-specification#notation-conventions-and-compliance. But there are some places where we use statements like "is", "has", "can" instead of MUST/SHOULD/MAY.

reyang:
> +1 for aligning our "shalls", "musts", "required", "recommended", etc. to clearly defined standards.
> 
> Can we start here, which I think is the same suggestion as @ocelotl above?
> https://www.ietf.org/rfc/rfc2119.txt
> 
> For a short time, I worked on construction specifications, and I used to get my hand slapped for using "must" instead of "shall". As an emerging discipline, software conventions seem less established, and we can learn from the traditional engineering disciplines in tightening up our language.

@mattmccleary check this https://github.com/open-telemetry/opentelemetry-specification#notation-conventions-and-compliance.

mattmccleary:
> We already have this in theory, see https://github.com/open-telemetry/opentelemetry-specification#notation-conventions-and-compliance. But there are some places where we use statements like "is", "has", "can" instead of MUST/SHOULD/MAY.

@Oberon00 @reyang Thank you for the clarification. Digging through the text to button-up the language where there's divergence and making the feature matrix auto-generable would be valuable experience for an intern.

reyang:
@ocelotl do you want to resurrect this? I'm willing to work with you to get this merged. We've discussed this during the [03/02/2021 Spec SIG Mtg](https://docs.google.com/document/d/1-bCYkN-DWJq4jw1ybaDZYYmx-WAe6HnwfWbkm8d57v8/edit#heading=h.e40nkw8m97kz) that we do need something like this.

I proposed something like [ECMA262](https://262.ecma-international.org/11.0/#sec-ordinary-and-exotic-objects-behaviours) then I noticed your PR was doing the same thing.

ocelotl:
> @ocelotl do you want to resurrect this? I'm willing to work with you to get this merged. We've discussed this during the [03/02/2021 Spec SIG Mtg](https://docs.google.com/document/d/1-bCYkN-DWJq4jw1ybaDZYYmx-WAe6HnwfWbkm8d57v8/edit#heading=h.e40nkw8m97kz) that we do need something like this.
> 
> I proposed something like [ECMA262](https://262.ecma-international.org/11.0/#sec-ordinary-and-exotic-objects-behaviours) then I noticed your PR was doing the same thing.

@reyang, yes, I personally would love to see this move forward. I'm not sure how much time I can dedicate to this PR now :disappointed: (sorry if this is a dumb question, but how does ECMA262 does the same thing as this PR?)

That being said, @reyang, @mattmccleary, @Oberon00:

My goal with this PR is to make OpenTelemetry compliant with [this](https://www.w3.org/TR/qaframe-spec/#consistent-style):

_It is important for readers to be able to differentiate requirements in the specification from non-requirements in order to either implement or review them._

This means:

1. Define a way to separately format the RFC2119-keyworded requirements in the OpenTelemetry documents.
2. Review all the current specification to make sure everywhere there should be an RFC2119-keyworded requirement, it is written following the previously defined format.

Only reviewing and fixing the current document to make sure that we are using an RFC2119 keyword everywhere it should is not enough. With time, we will very likely make the same mistake of not using the correct RFC2119 keyword where it is needed again and will end up with the same problem that we have now. We need this separate format for what I call "requirement sections" so that we clearly separate what needs to be exactly written with an RFC2119 keyword from what does not so that our readers can easily tell what they need to implement to be compliant.

Ok, the previous point 1 is in my opinion not that long or time consuming, it is mostly a matter of finding the right way to format these "requirement sections". @Oberon00 has previously raised very valid points regarding the inability of Github to render nicely these sections in HTML when we use markdown. I have some hope that using a different markup language (RestructuredText, maybe?) may allow us to work around this problem. Of course, rewriting the entire specification in a different markup language can be something hard to sell to the whole OpenTelemetry community because it may require developers to learn the differences between a new markup language and Github-flavored markdown.

I think those are the less time consuming problems that need solving (which does not mean that they will consume little time). The bulk of the work (if we decide to move forward with this PR, of course) will be reading, understanding and rewriting every occurrence of a requirement in the "requirement section" format.

I believe the changes suggested in this PR will make OpenTelemetry significantly better. Requirements will be clearly defined, implementations would be able to compare their compliance against all the "requirement sections" (and so, they will be able to know how compliant they are) in this PR and it will be easier for them (and us) to refer to a specific requirement when we communicate with each other. I also want to be honest and tell you all that I understand that this PR aims a review and rewrite of the whole specification and maybe even using a new markup language to write it, and that is of course a big change that can impact this project.

Sorry for the long post :sweat_smile: If someone has any idea on how to minimize the impact of this PR (or can provide resources to do this work (an intern, as @mattmccleary mentioned, maybe?)), please leave a comment below :slightly_smiling_face: 

Thank you all!

reyang:
@ocelotl here goes my suggestion:
1. I think this PR has a good shape, instead of waiting for a "hypothesis" intern to boil the ocean, I think we should focus on making small steps. Once the merge conflict is resolved, we should get it merged.
2. Once this PR get merged, I can take over if you don't have time. I'll clean up the other docs, and put CI enforcement to make sure folks get CI break if they don't follow the rule.
3. Figuring out another format could take time, and I don't see it as a blocker. The main effort here is to have the spec written in a more organized way. Once it is better organized, converting it to another format is quite straightforward.

specification_parser/specification_parser.py
```diff
@@ -0,0 +1,118 @@
+from re import finditer, findall
```
reyang, line 1: Instead of introducing another top level folder, could we put this under an existing folder? e.g. `https://github.com/open-telemetry/opentelemetry-specification/tree/main/internal/tools`

reyang (approved):
This is a good step towards our goal.
Please resolve the merge conflict and move the script to the `internal/tools` folder.

ocelotl:
> This is a good step towards our goal.
> Please resolve the merge conflict and move the script to the `internal/tools` folder.

All right, thanks @reyang! Will get this fixed ASAP

specification_parser/specification_parser.py
ocelotl: yes, fixed :+1:

specification/requirements.md
```diff
@@ -0,0 +1,131 @@
...
+
+###### requirement: concept_documentation
+> The concept SHOULD be documented in every implementation.
+```
+
+### Purpose of the requirement sections
+
+The idea behind writing the requirements in this manner is to make it easy for the reader to find all the
+requirements included in an OpenTelemetry specification document. In this way, it is also easy to find all
+the requirements a certain implementaiton must comply with. With all the requirements available for the
```
reyang, line 97:
```suggestion
the requirements a certain implementation must comply with. With all the requirements available for the
```

ocelotl:
> > This is a good step towards our goal.
> > Please resolve the merge conflict and move the script to the `internal/tools` folder.
> 
> All right, thanks @reyang! Will get this fixed ASAP

Fixed, @reyang :+1:

reyang:
@MrAlias @Oberon00 @yurishkuro I think this is good to go? Please give your approval/rejection. Thanks!

specification/metrics/api.md
```diff
@@ -108,12 +108,23 @@ can be configured at run time.
...
+
+###### requirement: meters_return_no_ops
+
+> Meters MUST return no-op implementations of any instruments.
+
+From a user's perspective, calls to these should be ignored without raising errors.
+
+###### requirement: no_null_references
+
+> No null references MUST be returned in languages where accessing these results in errors.
```
yurishkuro, line 123: nit: I recommend avoiding inverted statements
```suggestion
> Null references MUST NOT be returned in languages where accessing these results in errors.
```

yurishkuro:
I feel like there's still a gap between the intention in the original ticket and the proposal in the PR. I think the intent is to support a workflow like this:
  * a language SIG wants to release a new version of the API/SDK
  * they run the script against specification that produces a checklist of all requirements
  * they certify that each requirement is satisfied by the language implementation, by
    * having a test suite where there is an entry for each requirement, e.g. a unit test function with the name of the requirement id

The PR currently does the second step in this, but doesn't say anything more, specifically _how_ it should be used, other than saying that it assigns unique IDs that can be references externally.

specification/requirements.md
```diff
@@ -0,0 +1,131 @@
...
+The first six `#` symbols create a Markdown heading for the requirement section. The `requirement: `
+string that follows indicates that this particular header is part of a requirement section and not
+just any Markdown six `#` heading. The following string indicated by `unique_key_indetifier` MUST be
+unique in the document that contains it. The characters that make this string SHOULD only be
+alphanumeric characters and underscores.
+
+#### Description Format
+
+The description of every requirement section MUST be written as a
+[block quote](https://github.github.com/gfm/#block-quotes) immediately following a requirement section key:
```
yurishkuro, line 64: Why is a block quote required? Is the meaning of not-quoted text different?

specification/requirements.md
```diff
@@ -0,0 +1,131 @@
...
+
+Also, included in the same Markdown documents that make the OpenTelemetry specification
+are to be included specific sections named _requirement sections_ that follow a specific
+format. These sections are the part of the document that more formally describes each of
+the specific requirements that the OpenTelemetry specification has.
+
+Each of these requirement sections has 2 components:
+
+1. A unique **key**, a string that identifies the requirement section in a Markdown document
+2. A **description**, a string that MUST have one an only one of the [RFC 2119 keywords](https://tools.ietf.org/html/rfc2119).
```
MrAlias, line 39:
```suggestion
2. A **description**, a string that MUST have one and only one of the [BCP 14 keywords](https://tools.ietf.org/html/bcp14).
```

MrAlias: This 'one and only one' requirement will be hard to satisfy when there are normative statements like "some clause SHOULD apply, if it does it MUST ...", or "some clause is OPTIONAL but when included it MUST take this form...".

specification/requirements.md
```diff
@@ -0,0 +1,131 @@
...
+that includes at least one requirement section. With this JSON files, a testing schema can be produced for
+every implementation that can help developers know how compliant with the specification the implementation is.
+
+The parser can also work as a checker that makes sure that every requirement section is compliant with this
+specification. This can even be incorporated to the CI of the repo where the OpenTelemetry specification is
+in order to reject any change that adds a non-compliant requirement section.
+
+Finally, it makes the specification developer follow a "testing mindset" while writing requirements. For example,
+when writing a requirement, the specification developers ask themselves "can a test be written for this statement?".
+This helps writing short, concise requirements that are clear for the implementation developers.
```
MrAlias, line 118: This is a recipe for constrained and poor specification writing. The specification is written in English not a programming language. If this is going to impose upon authors of the English language a restriction that will take away the expressiveness needed to communicate concepts and ideas it is wrong. It should be engineered the other way, the parser should parse English.

MrAlias (commented):
This approach described here will require that normative requirements be manually consolidated into an additional section of the specification by a human instead of directly reading them from the specification. This suffers from the same problem discussed about the specification compliance matrix, it will form a derivative work. As such, it will compete for authority with the specification in situations where there are conflicts or stale statements.

My understanding was that we would try to generate a equivalent form of this requirements section using a program not having a human duplicate the specification into a more easily parsable form for a machine. I'm skeptical if this moves us closer to a document that is easier for implementers to check their implementation with or it is just rewrites the specification from English to a psudo-machine code and adds translators for other machine languages.

specification/requirements.md
ocelotl: The intention of having separate specification sections is not to impose upon the authors a restriction that will take away any expressiveness to communicate ideas. Any resource to communicate ideas can be part of the document (text, images, diagrams, etc.). This PR only intends to define how _certain specific sections_ of the specification are to be written, these sections are the "hard requirements" of the specification and it is convenient that they are defined in a clear manner. This is also necessary to define precise requirements that can be easily extracted from the specification. Everything else outside these sections can be expressed freely and the only limit is the imagination of the author.

specification/requirements.md
ocelotl: Good point, @MrAlias. I guess it MAY (hehe) be possible to use something like "nested" specification sections, something like:

# requirement: requirement_for_that

> That MAY be present

## requirement: requirement_for_that_if_it_is_present

> If that is present, then that MUST be this

Nevertheless, much more thought seems to be necessary to make sure that this is a way that can cover all the situations where a section specification can be written. Also, this goes against the previous idea of using six `#` to introduce a requirement section. Since I am not sure of how to address your concern completely and correctly now, I'll rewrite this section to be less constraining. :+1:

specification/requirements.md
ocelotl: No, it is not any different but it helps the parser "know" where the description begins and where it ends.

internal/tools/specification_parser/specification_parser.py
```diff
@@ -0,0 +1,118 @@
...
+            requirement_description = requirement["description"].strip()
+
+            RFC_2119_keyword_matches = []
+
+            for RFC_2119_keyword_regex in [
+                r"MUST NOT",
+                r"MUST(?! NOT)",
+                r"SHOULD NOT",
+                r"SHOULD(?! NOT)",
+                r"MAY"
```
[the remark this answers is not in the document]
ocelotl: Added :+1:

ocelotl:
> I feel like there's still a gap between the intention in the original ticket and the proposal in the PR. I think the intent is to support a workflow like this:
> 
> * a language SIG wants to release a new version of the API/SDK
> * they run the script against specification that produces a checklist of all requirements
> * they certify that each requirement is satisfied by the language implementation, by
>   
>   * having a test suite where there is an entry for each requirement, e.g. a unit test function with the name of the requirement id
> 
> The PR currently does the second step in this, but doesn't say anything more, specifically _how_ it should be used, other than saying that it assigns unique IDs that can be references externally.

I added a [paragraph](https://github.com/open-telemetry/opentelemetry-specification/pull/1211/files#diff-6994dc74d836ff3e312096163b03b6e6d979197a62271c0c40b276bfa197e61cR137-R139) to better explain what is to be done with the generated JSON files. I hope this makes the overall purpose of this PR more clear.

ocelotl:
> This approach described here will require that normative requirements be manually consolidated into an additional section of the specification by a human instead of directly reading them from the specification. This suffers from the same problem discussed about the specification compliance matrix, it will form a derivative work. As such, it will compete for authority with the specification in situations where there are conflicts or stale statements.

The requirement sections introduced in this PR won't compete for authority with the rest of the specification. The requirement sections will **be** the authority. This is because they will be the part of the specification that is written with a strict set of rules and is clearly delimited in a clearly marked section, that can be referenced as they will have an unique identifier. This is also necessary because implementation will be specifically looking at them to know what they should implement. Also, the requirement sections have [advantages](https://github.com/open-telemetry/opentelemetry-specification/pull/1211#issuecomment-725172134) over the compliance matrix because they indicate if the feature is mandatory or not.

> 
> My understanding was that we would try to generate a equivalent form of this requirements section using a program not having a human duplicate the specification into a more easily parsable form for a machine. I'm skeptical if this moves us closer to a document that is easier for implementers to check their implementation with or it is just rewrites the specification from English to a psudo-machine code and adds translators for other machine languages.

There is no need to duplicate anything in the requirement sections nor for them to be a rewrite of the specification from English to another language. The requirement sections are meant to be the short, concise part of the specification that includes one or more BCP 14 keywords, nothing less, nothing more. Any example, explanation or clarification of intention will remain in the rest of the specification and they will compliment each other, the former to clearly define what is to be implemented, the latter to make it clear for human beings why the specification is how it is.

carlosalberto:
@reyang Wondering whether a prototype for this is still on the works?

PR #1212, opened 2020-11-09
Remove ordering reqirement for attributes.

### LINKED ISSUE #1203: Clarify why attribute order should be preserved
The specification around attributes says that "Attributes SHOULD preserve the order in which they're set." I do not understand why this requirement is in place. What is the use-case that would benefit from insertion order being preserved?

** What did you expect to see? **

Some sort of reasoning for *why* attribute order should be preserved. I suspect that this will not be something that Java users will expect, as attributes act much like a java `Map`, which does not preserve insertion order, in general. I don't understand the rationale for this requirement.

**Additional context.**

In the Java repository, we have done significant work to optimize the memory usage of attributes. They do not current preserve insertion order, but instead are sorted lexigraphically, which enables some optimizations. Introducing the preservation of insertion order would make this more difficult.

Oberon00:
Fixes #1203

## Changes

Removes ordering requirement for attributes for now.

## Motivation

If we decide this is helpful, we can add it back after GA. But if we have it in GA and decide it is not worth it (e.g. for performance reasons), removing this would be a potentially breaking change for users.

CC @jkwatson

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,7 +64,8 @@ New:
 
 Updates:
 
-- Remove ordering SHOULD-requirement for attributes.
+- Remove ordering SHOULD-requirement for attributes
+  ([#1212](https://github.com/open-telemetry/opentelemetry-specification/pull/1212))
 - Make `process.pid` optional, split `process.command_args` from `command_line`
   ([#1137](https://github.com/open-telemetry/opentelemetry-specification/pull/1137))
 - Renamed `CorrelationContext` to `Baggage`:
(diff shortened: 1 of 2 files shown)


iNikem (approved):

jkwatson (approved):
Thanks!

trask (approved):

bogdandrutu (approved):

tigrannajaryan (approved):

anuraaga (approved):

arminru (approved):

carlosalberto (approved):

bogdandrutu:
8 approves (6 from owners)

PR #966, opened 2020-09-17
Error Flagging with Status Codes

tedsuo:
## Changes

* Changed StatusCodes to Unset, Error, and Ok
* Added StatusSource
* Updated HTTP status code mapping

Related issues #
#965 

Related [oteps](https://github.com/open-telemetry/oteps) #
* [OTEP 136](https://github.com/open-telemetry/oteps/blob/master/text/trace/0136-error_flagging.md)

PR #1216, opened 2020-11-10
Remove deprecated span status codes from http semantic conventions

### LINKED ISSUE #1204: The HTTP semantic conventions reference CanonicalStatusCodes that no longer exist in the spec
I found [this block](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/http.md#http-client) in the HTTP trace semantic conventions:

```
For status, the following special cases have canonical error codes assigned:

| Client error                | Trace status code  |
|-----------------------------|--------------------|
| DNS resolution failed       | `Unknown`     |
| Request cancelled by caller | `Cancelled`        |
| URL cannot be parsed        | `InvalidArgument`  |
| Request timed out           | `DeadlineExceeded` |

This is not meant to be an exhaustive list
but if there is no clear mapping for some error conditions,
instrumentation developers are encouraged to use `Unknown`
and open a PR or issue in the specification repository.
```

But since PR #966, we've only had three statuses specified: `Ok`, `Unset`, and `Error`.

I propose that we remove this whole section from the HTTP semantic convention. 

But should we replace it with different guidance about these error conditions?  
I'd expect most languages to present these as exceptions or errors, which could be added to a span as events. Is that what we should suggest here?

Related: #965

justinfoote:
Fixes #1204

This is an editorial change. These span status codes were made obsolete in #966, and should have been removed then.

### DIFF
--- a/specification/trace/semantic_conventions/http.md
+++ b/specification/trace/semantic_conventions/http.md
@@ -106,20 +106,6 @@ from the `net.peer.name`
 used to look up the `net.peer.ip` that is actually connected to.
 In that case it is strongly recommended to set the `net.peer.name` attribute in addition to `http.host`.
 
-For status, the following special cases have canonical error codes assigned:
-
-| Client error                | Trace status code  |
-|-----------------------------|--------------------|
-| DNS resolution failed       | `Unknown`     |
-| Request cancelled by caller | `Cancelled`        |
-| URL cannot be parsed        | `InvalidArgument`  |
-| Request timed out           | `DeadlineExceeded` |
-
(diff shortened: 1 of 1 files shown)


Oberon00 (approved):
Thank you!

iNikem (approved):

arminru (approved):

carlosalberto (approved):

bogdandrutu (approved):
Thank you

bogdandrutu:
Trivial change, can be merged before the 24/48h

PR #1217, opened 2020-11-10
Add more attributes for Cassandra to database semantic conventions

### LINKED ISSUE #1058: Additional Cassandra Attributes
I'd like to define more cassandra attributes, as requested by @trask in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1314#issuecomment-703134133
Additionally, some of these tags may apply to other DBs as well.

What did you expect to see?

I expected to see more detailed information, including table name, query execution information, etc. Listed below.
```java
  public static final String QUERY_PAGE_SIZE = "query.pageSize";
  public static final String QUERY_CL = "query.cl"; // consistency level
  public static final String DB_HOST_DC = "db.host.dc";
  public static final String DB_HOST_ID = "db.host.id";
  public static final String QUERY_SPECULATIVE_EXECUTION_COUNT = "query.speculativeExecutionCount";
  public static final String QUERY_IDEMPOTENCE = "query.idempotence";
  public static final String QUERY_TABLE = "query.table";
```
**Additional context.**

Some of these tags already exist in the opentracing cassandra instrumentation https://github.com/opentracing-contrib/java-cassandra-driver/blob/master/opentracing-cassandra-driver-3/src/main/java/io/opentracing/contrib/cassandra/TracingSession.java#L500

FrankSpitulski:
Fixes #1058 

## Changes

Adds additional optional cassandra tags.

Related issues #

Related [instrumentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1298)


semantic_conventions/trace/database.yaml
```diff
@@ -251,6 +251,49 @@ groups:
...
+          tag: call-level-tech-specific
+          brief: >
+            The page size of the query.
+          examples: [5000]
+        - id: consistency_level
+          type: string
+          tag: call-level-tech-specific
+          brief: >
+            The consistency level of the query.
+          examples: 'LOCAL_QUORUM'
```
arminru, line 265: This could be an enum (see the definition of `db.system` in line 8 above).

semantic_conventions/trace/database.yaml
```diff
@@ -251,6 +251,49 @@ groups:
...
+        - id: table
+          type: string
+          tag: call-level-tech-specific
+          brief: >
+            The table being queried.
+          examples: 'mytable'
```
arminru, line 271: We could consider using `db.sql.table` instead which has the same meaning and a more elaborate definition. CQL is close to SQL after all.

semantic_conventions/trace/database.yaml
```diff
@@ -251,6 +251,49 @@ groups:
...
+          tag: call-level-tech-specific
+          brief: >
+            Whether or not the query is idempotent.
+          examples: [true, false]
+        - id: speculative_execution_count
+          type: number
+          required: always
+          tag: call-level-tech-specific
+          brief: >
+            The number of times a query was speculatively executed. A query that was not speculatively executed will have a value of 0.
```
arminru, line 283:
```suggestion
        - id: speculative_execution_count
          type: number
          tag: call-level-tech-specific
          brief: >
            The number of times a query was speculatively executed. Not set or `0` if the query was not executed speculatively.
```

semantic_conventions/trace/database.yaml
```diff
@@ -251,6 +251,49 @@ groups:
...
+          type: string
+          tag: call-level-tech-specific
+          brief: >
+            The host id reported by the coordinating node for a query.
+          examples: 'be13faa2-8574-4d71-926d-27f16cf8a7af'
+        - id: host.dc
+          type: string
+          tag: call-level-tech-specific
+          brief: >
+            The host dc reported by the coordinating node for a query.
```
arminru, line 295:
```suggestion
            The host data center reported by the coordinating node for a query.
```

semantic_conventions/trace/database.yaml
```diff
@@ -251,6 +251,49 @@ groups:
...
+          tag: call-level-tech-specific
+          brief: >
+            The table being queried.
+          examples: 'mytable'
+        - id: idempotence
+          type: boolean
+          tag: call-level-tech-specific
+          brief: >
+            Whether or not the query is idempotent.
+          examples: [true, false]
```
arminru, line 277:
```suggestion
```
You could leave those out for booleans (it also looks a bit odd when rendered since the YAML parser normalizes it to its capital letter representation).

arminru (commented):
Hi @FrankSpitulski!
While some SIGs use PR titles as the one here, this is not common in the spec repo. Apart from that, please use the correct term "attribute" or people might get confused. What about "Add more attributes for Cassandra to database semantic conventions"?
Also please add an entry in `CHANGELOG.md`.
Thanks!

semantic_conventions/trace/database.yaml
Oberon00: Indeed, this should be an enum. Or at least you should describe where the string comes from. This looks like a constant/enumerator defined in some library. But note that we want consistency across languages, so this is not necessarily the best source, e.g. one language might call it `LOCAL_QUORUM`, another one `kLocalQuorum` a third `CASSANDRA_CONSISTENCY_LOCAL_QUORUM`. So we should have definitive list of those here, or a link to a definitive list (e.g. Cassandra docs).

semantic_conventions/trace/database.yaml
Oberon00: I think having a separate attribute here for cassandra is fine. Otherwise we might want to rename `db.sql.table` to `db.table` after all and describe to which technologies it applies. But I think I prefer the separate attribute a bit.

semantic_conventions/trace/database.yaml
Oberon00: I think they are enforced by the parser. Or did we have an exception for bool? Maybe.

thisthat: They are optional for boolean and enum: https://github.com/open-telemetry/opentelemetry-specification/blob/master/semantic_conventions/syntax.md#attributes

semantic_conventions/trace/database.yaml
FrankSpitulski: +1 I agree that should either be separate or use a top level db attribute

semantic_conventions/trace/database.yaml
FrankSpitulski: the docs have the full set https://docs.datastax.com/en/cassandra-oss/3.0/cassandra/dml/dmlConfigConsistency.html

FrankSpitulski: changed to use an enum

semantic_conventions/trace/database.yaml
```diff
@@ -251,6 +251,69 @@ groups:
...
+        - id: page_size
+          type: number
+          tag: call-level-tech-specific
+          brief: >
+            The page size of the query.
+          examples: [5000]
+        - id: consistency_level
+          tag: call-level-tech-specific
+          brief: >
+            The consistency level of the query. Based on consistency values from [cql](https://docs.datastax.com/en/cassandra-oss/3.0/cassandra/dml/dmlConfigConsistency.html).
```
arminru, line 263:
```suggestion
            The consistency level of the query. Based on consistency values from [CQL](https://docs.datastax.com/en/cassandra-oss/3.0/cassandra/dml/dmlConfigConsistency.html).
```
Or did you mean to write `cqlsh`?

arminru (commented):
LGTM!
It would be great if someone more familiar with Cassandra than me could take a look as well 🙂

specification/trace/semantic_conventions/database.md
```diff
@@ -162,6 +162,13 @@ For example, when retrieving a document, `db.operation` would be set to (literal
 | Attribute  | Type | Description  | Example  | Required |
 |---|---|---|---|---|
 | `db.cassandra.keyspace` | string | The name of the keyspace being accessed. To be used instead of the generic `db.name` attribute. | `mykeyspace` | Yes |
+| `db.cassandra.page_size` | number | The page size of the query. | `5000` | No |
+| `db.cassandra.consistency_level` | string enum | The consistency level of the query. Based on consistency values from [cql](https://docs.datastax.com/en/cassandra-oss/3.0/cassandra/dml/dmlConfigConsistency.html). | `ALL` | No |
+| `db.cassandra.table` | string | The table being queried. | `mytable` | No |
+| `db.cassandra.idempotence` | boolean | Whether or not the query is idempotent. |  | No |
+| `db.cassandra.speculative_execution_count` | number | The number of times a query was speculatively executed. Not set or `0` if the query was not executed speculatively. | `0`<br>`2` | No |
+| `db.cassandra.host.id` | string | The host id reported by the coordinating node for a query. | `be13faa2-8574-4d71-926d-27f16cf8a7af` | No |
```
yurishkuro, line 170: "host" is confusing here, how about "coordinator.id"?

specification/trace/semantic_conventions/database.md
```diff
@@ -162,6 +162,13 @@ For example, when retrieving a document, `db.operation` would be set to (literal
 | Attribute  | Type | Description  | Example  | Required |
 |---|---|---|---|---|
 | `db.cassandra.keyspace` | string | The name of the keyspace being accessed. To be used instead of the generic `db.name` attribute. | `mykeyspace` | Yes |
+| `db.cassandra.page_size` | number | The page size of the query. | `5000` | No |
+| `db.cassandra.consistency_level` | string enum | The consistency level of the query. Based on consistency values from [cql](https://docs.datastax.com/en/cassandra-oss/3.0/cassandra/dml/dmlConfigConsistency.html). | `ALL` | No |
+| `db.cassandra.table` | string | The table being queried. | `mytable` | No |
+| `db.cassandra.idempotence` | boolean | Whether or not the query is idempotent. |  | No |
```
yurishkuro, line 168: Who can determine whether a query is idempotent?  This does not sound like Cassandra specific.

semantic_conventions/trace/database.yaml
```diff
@@ -251,6 +251,69 @@ groups:
           brief: >
             The name of the keyspace being accessed. To be used instead of the generic `db.name` attribute.
           examples: 'mykeyspace'
+        - id: page_size
```
yurishkuro, line 254: the driver seems to call it "fetch size": https://docs.datastax.com/en/developer/java-driver/2.1/manual/paging/

The description below could be better, like "fetch size used for paging, i.e. how many rows will be returned at once"

specification/trace/semantic_conventions/database.md
FrankSpitulski: Every query is marked as idempotent or not. https://docs.datastax.com/en/devapp/doc/devapp/driversQueryIdempotence.html

specification/trace/semantic_conventions/database.md
FrankSpitulski: sure I can change it to coordinator

semantic_conventions/trace/database.yaml
FrankSpitulski: CQL is fine. Either work. I'll use CQL.

semantic_conventions/trace/database.yaml
FrankSpitulski: The most recent driver uses the term page size https://docs.datastax.com/en/developer/java-driver/4.9/manual/core/paging/

specification/performance-benchmark.md
```diff
@@ -1,70 +1,70 @@
-# Performance Benchmark of OpenTelemetry API
```
yurishkuro, line 1: can this file be reverted? doesn't seem related to the PR.

FrankSpitulski: yeah this is weird. I don't remember touching this file.

semantic_conventions/trace/database.yaml
```diff
@@ -251,6 +251,69 @@ groups:
...
+          type: number
+          tag: call-level-tech-specific
+          brief: >
+            The number of times a query was speculatively executed. Not set or `0` if the query was not executed speculatively.
+          examples: [0, 2]
+        - id: coordinator.id
+          type: string
+          tag: call-level-tech-specific
+          brief: >
+            The (node's) id of the coordinating node for a query.
```
arminru, line 309:
```suggestion
            The ID of the coordinating node for a query.
```

specification/trace/semantic_conventions/database.md
```diff
@@ -162,6 +162,13 @@ For example, when retrieving a document, `db.operation` would be set to (literal
 | Attribute  | Type | Description  | Example  | Required |
 |---|---|---|---|---|
 | `db.cassandra.keyspace` | string | The name of the keyspace being accessed. To be used instead of the generic `db.name` attribute. | `mykeyspace` | Yes |
+| `db.cassandra.page_size` | number | The fetch size used for paging, i.e. how many rows will be returned at once. | `5000` | No |
+| `db.cassandra.consistency_level` | string enum | The consistency level of the query. Based on consistency values from [CQL](https://docs.datastax.com/en/cassandra-oss/3.0/cassandra/dml/dmlConfigConsistency.html). | `ALL` | No |
+| `db.cassandra.table` | string | The table being queried. | `mytable` | No |
+| `db.cassandra.idempotence` | boolean | Whether or not the query is idempotent. |  | No |
+| `db.cassandra.speculative_execution_count` | number | The number of times a query was speculatively executed. Not set or `0` if the query was not executed speculatively. | `0`<br>`2` | No |
+| `db.cassandra.coordinator.id` | string | The (node's) id of the coordinating node for a query. | `be13faa2-8574-4d71-926d-27f16cf8a7af` | No |
```
arminru, line 170:
```suggestion
| `db.cassandra.coordinator.id` | string | The ID of the coordinating node for a query. | `be13faa2-8574-4d71-926d-27f16cf8a7af` | No |
```

arminru (approved):
LGTM!
It would be great if someone more familiar with Cassandra than me could take a look as well 🙂

semantic_conventions/trace/database.yaml
```diff
@@ -251,6 +251,69 @@ groups:
...
+            The fetch size used for paging, i.e. how many rows will be returned at once.
+          examples: [5000]
+        - id: consistency_level
+          tag: call-level-tech-specific
+          brief: >
+            The consistency level of the query. Based on consistency values from [CQL](https://docs.datastax.com/en/cassandra-oss/3.0/cassandra/dml/dmlConfigConsistency.html).
+          type:
+            members:
+              - id: ALL
+                value: 'ALL'
```
Oberon00, line 267: We usually use all-lowercase for these values. If there is any chance that we allow custom values in the future and these are the actual cassandra-native values (and not only used in the Java client lib), I'm fine with uppercase though.

semantic_conventions/trace/database.yaml
SergeyKanzhelev: `db.sql.table` is already mentioned in the rendered table. Having both is confusing:

semantic_conventions/trace/database.yaml
FrankSpitulski: This would fall under being cassandra native values. There was some previous discussion on it here. https://github.com/open-telemetry/opentelemetry-specification/pull/1217#discussion_r520797155

SergeyKanzhelev:
I am still concerned with this: https://github.com/open-telemetry/opentelemetry-specification/pull/1217#discussion_r524727177 two attributes with the similar semantic in the same table:

| Attribute  | Type | Description  | Example  | Required |
|---|---|---|---|---|
| `db.cassandra.table` | string | The table being queried. | `mytable` | No |
| `db.sql.table` | string | The name of the primary table that the operation is acting upon, including the schema name (if applicable). [2] | `public.users`<br>`customers` | Conditional<br>Recommended if available. |

semantic_conventions/trace/database.yaml
FrankSpitulski: There isn't much more to say @SergeyKanzhelev unless you have a new suggestion.

SergeyKanzhelev: the question is what having a cassandra-specific attribute should be defined? Why not use generic `db.sql.table`? If this is totally wrong, can we change the tool to at least not show both attributes in the same table?

FrankSpitulski: The issue with using db.sql.table is that it's not generic. It's defined in the sql specific section instead of in the generic DB section.

Oberon00: Since we have quite a few of them, it would make sense to have a separate table for cassandra, indeed. This can be achieved by using a different tag, or a different semantic convention group (that extends the generic db)

Oberon00: Also, the brief of `cassandra.table` should mention Cassandra and the brief of sql.table SQL.

FrankSpitulski: Bumping, will get to this eventually

FrankSpitulski: @Oberon00, I just had a closer look. It seems it already is a different convention group that extends the generic db.

```
groups:
    - id: db ...

    - id: db.mssql ...

    - id: db.cassandra
      prefix: db.cassandra
      extends: db
```

Oberon00: Cool, then it should be possible to just edit the magic HTML comments in the markdown to generate that in a separate table.

FrankSpitulski: @Oberon00 it's been split

specification/trace/semantic_conventions/database.md
```diff
@@ -178,6 +179,25 @@ For example, when retrieving a document, `db.operation` would be set to (literal
 **[2]:** It is not recommended to attempt any client-side parsing of `db.statement` just to get this property, but it should be set if it is provided by the library being instrumented. If the operation is acting upon an anonymous table, or more than one table, this value MUST NOT be set.
 <!-- endsemconv -->
 
+#### Cassandra
+
+Seperated for clarity.
```
FrankSpitulski, line 184:
```suggestion
Separated for clarity.
```

FrankSpitulski:
@SergeyKanzhelev It's been moved to a new table

SergeyKanzhelev (approved):
LGTM, but if somebody with the experience of running Cassandra can comment - would be great.

iNikem:
@open-telemetry/technical-committee does anything prevent this from merging? Either we don't have "somebody with the experience of running Cassandra" or they do not object. Either way there is no point in waiting forever, is it?

specification/trace/semantic_conventions/database.md
```diff
@@ -178,6 +179,25 @@ For example, when retrieving a document, `db.operation` would be set to (literal
 **[2]:** It is not recommended to attempt any client-side parsing of `db.statement` just to get this property, but it should be set if it is provided by the library being instrumented. If the operation is acting upon an anonymous table, or more than one table, this value MUST NOT be set.
 <!-- endsemconv -->
 
+#### Cassandra
+
+Separated for clarity.
+
+<!-- semconv db.tech(tag=call-level-tech-specific-cassandra) -->
```
Oberon00, line 186: I think this would be cleaner if you used

```suggestion
<!-- semconv db.cassandra -->
```

here. Then you could remove the `tag: call-level-tech-specific-cassandra` line everywhere.

Also, you can remove the `include:` for cassandra from `db.tech`.

specification/trace/semantic_conventions/database.md
```diff
@@ -178,6 +179,25 @@ For example, when retrieving a document, `db.operation` would be set to (literal
 **[2]:** It is not recommended to attempt any client-side parsing of `db.statement` just to get this property, but it should be set if it is provided by the library being instrumented. If the operation is acting upon an anonymous table, or more than one table, this value MUST NOT be set.
 <!-- endsemconv -->
 
+#### Cassandra
+
+Separated for clarity.
+
```
Oberon00, line 185:
```suggestion
```

Oberon00 (approved):

Oberon00:
IMHO: Let's merge it, and address my YAML nitpicks later. We can't make the review periods arbitrarily long and still expect contributors to respond.

PR #1219, opened 2020-11-10
Add a SQL span naming suggestion: <db.operation> <db.name>.<db.sql.table>

### LINKED ISSUE #1139: Is it okay to use SQL statement type (e.g. just "SELECT") as span name?

Spec states that "SQL statement without variable arguments" are a good choice for DB span name, but sometimes SQL queries (especially ones generated by ORM frameworks like Hibernate) get terribly long and may have very high cardinality. E.g.

```sql
SELECT t.col1 as column1, t.col2 as column2, t.col3 as column3, t.col4 as column4, t.col5 as column5 from table t where t.id = ?
```

just for selecting a single entry. Imagine a query that eagerly fetches 3 tables joined together, each one having 15-20 columns, with some complex conditions -- and everything generated completely dynamically using `session.createCriteria()` (Hibernate again).

Instead, would it be fine to use just `SELECT` (or `INSERT` or ...) as span name? Full query should be available as `db.statement` anyway.

mateuszrzeszutek:
Fixes #1139

## Changes

Add a suggestion that `<db.operation> <db.name>.<db.sql.table>` may be used for SQL DB spans; also modify the MySQL example to reflect that.

Related issues:
* https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1409

Related [oteps](https://github.com/open-telemetry/oteps): none


specification/trace/semantic_conventions/database.md
```diff
@@ -19,7 +19,9 @@
 **Span kind:** MUST always be `CLIENT`.
 
 The **span name** SHOULD be set to a low cardinality value representing the statement executed on the database.
-It may be a stored procedure name (without arguments), SQL statement without variable arguments, operation name, etc.
+It may be a stored procedure name (without arguments), DB statement without variable arguments, operation name, etc.
+Since SQL statements may have very high cardinality even without arguments, SQL spans MAY be named the following way:
+`<db.operation> <db.name>.<db.sql.table>`, provided that `db.operation` and/or `db.sql.table` are available.
```
arminru, line 24:
```suggestion
`<db.operation> <db.name>.<db.sql.table>`, provided that `db.operation` and `db.sql.table` are available.
```
What would you do if only one of both is available?

iNikem:
@mateuszrzeszutek  Just to clarify: your intention is that both old-way "normalized" sql AND `<db.operation> <db.name>.<db.sql.table>` are OK ?

specification/trace/semantic_conventions/database.md
```diff
@@ -19,7 +19,9 @@
 **Span kind:** MUST always be `CLIENT`.
 
 The **span name** SHOULD be set to a low cardinality value representing the statement executed on the database.
-It may be a stored procedure name (without arguments), SQL statement without variable arguments, operation name, etc.
+It may be a stored procedure name (without arguments), DB statement without variable arguments, operation name, etc.
```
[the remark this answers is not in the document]
mateuszrzeszutek: Good point, done.

specification/trace/semantic_conventions/database.md
mateuszrzeszutek: As discussed in https://github.com/open-telemetry/opentelemetry-specification/issues/1139#issuecomment-719670107, when `db.sql.table` is unavailable just skip it and use the `<db.operation> <db.name>` naming pattern.
I've added that case to the spec.

mateuszrzeszutek:
> your intention is that both old-way "normalized" sql AND <db.operation> <db.name>.<db.sql.table> are OK ?

Hmm, I think that the whole DB span naming spec is vague intentionally and it doesn't make any sense to say "but SQL MUST be named this way". I believe that we should at the very least recommend this low-cardinality naming scheme -- so I changed MAY to RECOMMENDED.

Oberon00:
Note that RECOMMENDED is defined to mean the same as SHOULD and we usually use the latter: https://github.com/open-telemetry/opentelemetry-specification#notation-conventions-and-compliance

mateuszrzeszutek:
> Note that RECOMMENDED is defined to mean the same as SHOULD and we usually use the latter: https://github.com/open-telemetry/opentelemetry-specification#notation-conventions-and-compliance

Right, a quick search has revealed that RECOMMENDED is hardly ever used in this project. Changed to SHOULD.

specification/trace/semantic_conventions/database.md
```diff
@@ -19,7 +19,10 @@
 **Span kind:** MUST always be `CLIENT`.
 
 The **span name** SHOULD be set to a low cardinality value representing the statement executed on the database.
-It may be a stored procedure name (without arguments), SQL statement without variable arguments, operation name, etc.
+It MAY be a stored procedure name (without arguments), DB statement without variable arguments, operation name, etc.
+Since SQL statements may have very high cardinality even without arguments, SQL spans SHOULD be named the
+following way: `<db.operation> <db.name>.<db.sql.table>`, provided that `db.operation` and `db.sql.table` are available.
```
arminru, line 24:
```suggestion
following way, unless the statement is known to be of low cardinality:
`<db.operation> <db.name>.<db.sql.table>`, provided that `db.operation` and `db.sql.table` are available.
```
The SHOULD requirement alone would already allow to deviate if there are legitimate reasons and if consequences are well considered, but this makes it a bit clearer that it's fine to use such statements in case the cardinality is well understood.

iNikem (approved):

arminru (approved):

trask (approved):

specification/trace/semantic_conventions/database.md
```diff
@@ -19,7 +19,11 @@
 **Span kind:** MUST always be `CLIENT`.
 
 The **span name** SHOULD be set to a low cardinality value representing the statement executed on the database.
-It may be a stored procedure name (without arguments), SQL statement without variable arguments, operation name, etc.
+It MAY be a stored procedure name (without arguments), DB statement without variable arguments, operation name, etc.
+Since SQL statements may have very high cardinality even without arguments, SQL spans SHOULD be named the
+following way, unless the statement is known to be of low cardinality:
+`<db.operation> <db.name>.<db.sql.table>`, provided that `db.operation` and `db.sql.table` are available.
+If `db.sq.table` is not available due to its semantics, the span SHOULD be named `<db.operation> <db.name>`.
```
yurishkuro, line 26:
```suggestion
If `db.sql.table` is not available due to its semantics, the span SHOULD be named `<db.operation> <db.name>`.
```

specification/trace/semantic_conventions/database.md
```diff
@@ -19,7 +19,11 @@
 **Span kind:** MUST always be `CLIENT`.
 
 The **span name** SHOULD be set to a low cardinality value representing the statement executed on the database.
-It may be a stored procedure name (without arguments), SQL statement without variable arguments, operation name, etc.
+It MAY be a stored procedure name (without arguments), DB statement without variable arguments, operation name, etc.
+Since SQL statements may have very high cardinality even without arguments, SQL spans SHOULD be named the
+following way, unless the statement is known to be of low cardinality:
+`<db.operation> <db.name>.<db.sql.table>`, provided that `db.operation` and `db.sql.table` are available.
```
yurishkuro, line 25: I am curious how a generic instrumentation would get access to this information. Quite often when instrumenting a DB driver abstraction (JDBC or DBI), all you have is a query string.

specification/trace/semantic_conventions/database.md
mateuszrzeszutek: Thanks for catching this!

specification/trace/semantic_conventions/database.md
mateuszrzeszutek: That is true. If the instrumented library does not provide an operation/table name you can always extract that info with a  relatively simple lexer.

Oberon00: I think we should make clear that the SQL string SHOULD NOT be parsed just for this. This is already explained in the documentation for db.operation, but could be re-iterated here.

specification/trace/semantic_conventions/database.md
```diff
@@ -24,6 +24,8 @@ Since SQL statements may have very high cardinality even without arguments, SQL
 following way, unless the statement is known to be of low cardinality:
 `<db.operation> <db.name>.<db.sql.table>`, provided that `db.operation` and `db.sql.table` are available.
 If `db.sql.table` is not available due to its semantics, the span SHOULD be named `<db.operation> <db.name>`.
+It is not recommended to attempt any client-side parsing of `db.statement` just to get these properties,
+they should be used if the library being instrumented provides them.
```
Oberon00, line 28:
```suggestion
they should only be used if the library being instrumented already provides them.
```

Oberon00 (approved):

anuraaga (approved):

specification/trace/semantic_conventions/database.md
arminru: This was clarified in 32de8ebdaeea6d92d9280a9ce8cf89fa8fbe3b71.

Issue #1148, opened 2020-10-28
OpenTelemetry should use an allowed port


bboreham:
**What are you trying to achieve?**

Conformance with Internet standards.

**What did you expect to see?**

A port number in the range of User Ports, and ideally one assigned by IANA.

**Additional context.**

OpenTelemetry is [using port 55680](https://github.com/open-telemetry/opentelemetry-specification/blob/cb25fe94d8d9378c04ee37d7802e7552cf5b6cc0/specification/protocol/exporter.md#configuration-options). According to [RFC 6335](https://tools.ietf.org/html/rfc6335#section-8.1.2), 
> a port number in [the Dynamic Ports range (49152-65535)] MUST NOT be used as a service identifier.

On a practical front, ports in the high range are usually allocated dynamically by the operating system, so there is some random chance 55680 is in use for that purpose and your OpenTelemetry receiver will not start.

andrewhsu:
from the spec issue triage mtg today, setting to P1 so we can have a quick decision. perhaps setting the port number a required parameter will allow us to set it to a default we want to choose later.

mtwo:
Following up from the same meeting. I'm making an IANA application now. I'll include the governance committee and @tigrannajaryan in all communications with them.

mtwo:
@tigrannajaryan @bogdandrutu three questions that I need answered before I file this:

- Do we have a [list of defined TXT keys](https://tools.ietf.org/html/rfc6763#section-6.1)? I've entered our service name as 'otlp'
- "Please describe how your protocol supports security" - what should I enter for this
- I'm assuming that we only use TCP and don't use UDP - is this accurate?

tigrannajaryan:
Does this proposal mean we need to change the default port? It will break a bunch of software that relies on the defaults.

mtwo:
It does. Do you think the benefits of having a reserved port outweigh the pre-GA change? Bogdan also proposed an alternative of making the port a required parameter during RC and then defaulting it (via an optional parameter) to the newly-assigned port at GA.

If you think that it's too late and we should stick with our current port, then we can close this and I won't submit the request to the IANA.

bboreham:
Server-side you can have the code listen to both old and new ports. The only hard break is for new-built clients that cannot be configured to talk to old servers.

mtwo:
I'll add this as a discussion topic at Wednesday's Collector SIG meeting

tigrannajaryan:
[...]

PR #1221, opened 2020-11-10
Change default OTLP port number

tigrannajaryan:
Contributes to https://github.com/open-telemetry/opentelemetry-specification/issues/1148

Note that a separate port is used for OTLP/HTTP for now. There is currently work
in progress to confirm that we can use the same port. Once we have the confirmation
I will update the spec again to use one port.

### DIFF
--- a/specification/protocol/otlp.md
+++ b/specification/protocol/otlp.md
@@ -289,7 +289,7 @@ some in beta.
 
 #### Default Port
 
-The default network port for OTLP/gRPC is 55680.
+The default network port for OTLP/gRPC is 4317.
 
 ### OTLP/HTTP
 
@@ -445,7 +445,7 @@ connections SHOULD be configurable.
 
 #### Default Port
 
-The default network port for OTLP/HTTP is 55681. There is currently an [open
+The default network port for OTLP/HTTP is 4318. There is currently an [open
 issue](https://github.com/open-telemetry/opentelemetry-collector/issues/1256) to
 use the same port for OTLP/gRPC and OTLP/HTTP. In that case this spec will be
 updated to use the same default port for OTLP/gRPC and OTLP/HTTP.
--- a/specification/protocol/otlp.md
+++ b/specification/protocol/otlp.md
@@ -445,10 +445,7 @@ connections SHOULD be configurable.
 
 #### Default Port
 
-The default network port for OTLP/HTTP is 4318. There is currently an [open
-issue](https://github.com/open-telemetry/opentelemetry-collector/issues/1256) to
-use the same port for OTLP/gRPC and OTLP/HTTP. In that case this spec will be
-updated to use the same default port for OTLP/gRPC and OTLP/HTTP.
+The default network port for OTLP/HTTP is 4317.
 
 ## Implementation Recommendations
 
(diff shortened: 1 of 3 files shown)


arminru (approved):
Please also update the default mentioned in `specification/protocol/exporter.md#configuration-options` and the examples below.

Should we add a compliance matrix entry for using the correct default port?

tigrannajaryan:
Corresponding change in the Collector, allowing to receive on new and legacy ports simultaneously to ease the transition https://github.com/open-telemetry/opentelemetry-collector/pull/2104

tigrannajaryan:
> Please also update the default mentioned in `specification/protocol/exporter.md#configuration-options` and the examples below.

Done.

 
> Should we add a compliance matrix entry for using the correct default port?

I added this to the CHANGELOG, hopefully it is sufficient.

yurishkuro (approved):

carlosalberto (approved):

bogdandrutu (approved):

bogdandrutu:
@tigrannajaryan is the number confirmed?

tigrannajaryan:
> @tigrannajaryan is the number confirmed?

@mtwo [submitted](https://github.com/open-telemetry/opentelemetry-specification/issues/1148#issuecomment-725067404) the request with IANA. It will take a while to get the confirmation. I think we should move forward with the assumption that it will be accepted. On the minor chance that it will be rejected and a different port number suggested we can update the spec again, we still have time.

carlosalberto:
> I think we should move forward with the assumption that it will be accepted. On the minor chance that it will be rejected and a different port number suggested we can update the spec again, we still have time.

+1

carlosalberto:
@open-telemetry/specs-approvers Anybody objects to going forward with this proposal, given that the request with IANA **has been filled** but it may take a while to get a confirmation? If not, I suggest merging early tomorrow or on Monday.

yurishkuro:
sgtm

I actually care less about having an official IANA port than about not using ephemeral port, because the latter has a higher chance of causing instability at deployment.

carlosalberto:
@bogdandrutu Any opinion? ;)

codeboten (approved):

bogdandrutu:
SGTM, let's merge this.

specification/protocol/otlp.md
```diff
@@ -445,7 +445,7 @@ connections SHOULD be configurable.
 
 #### Default Port
 
-The default network port for OTLP/HTTP is 55681. There is currently an [open
+The default network port for OTLP/HTTP is 4318. There is currently an [open
```
bogdandrutu, line 448: Did we not say that we try to not have 2 ports?

carlosalberto: Ping @tigrannajaryan

carlosalberto:
Since now we have the two-ports update, it looks we are ready to merge today, right? @tigrannajaryan @bogdandrutu

specification/protocol/otlp.md
tigrannajaryan: Fixed.

tigrannajaryan: The spec now lists the same port for both OTLP/gRPC and OTLP/HTTP.

bogdandrutu (approved):

bogdandrutu:
4 approves + author (TC member) - merging

PR #998, opened 2020-09-24
Specify Span ID creation with sampling.

Oberon00:
Fixes #1060.

Partial fix for #864 (remainder would require an API change to not pass a trace ID to the sampler if there is none).

## Changes

* Clarifies how span IDs are generated for DROP and RECORD_ONLY.

Issue #864, opened 2020-08-24
Option to allow "default" IDs for unsampled traces


awssandra:
Follow-up from the Sampling SIG on 08/21/2020.

Historically, we've seen a performance hit on the X-Ray SDKs from generating IDs on every request (as we were generating it regardless of sampling).

We have since moved to a model of only generating an ID when the sampling decision is positive. Otherwise we stick in a valid default ID (just all zero'd out). Is this kind of construct supported on the current sampling/trace model?

arminru:
> Otherwise we stick in a valid default ID (just all zero'd out).

Please note that as per the W3C Trace Context specification all zeroes means invalid.

> trace-id         = 32HEXDIGLC  ; 16 bytes array identifier. All zeroes forbidden
parent-id        = 16HEXDIGLC  ; 8 bytes array identifier. All zeroes forbidden
trace-id: All bytes as zero (00000000000000000000000000000000) is considered an invalid value.
parent-id: All bytes as zero (0000000000000000) is considered an invalid value.

-- https://www.w3.org/TR/trace-context/#version-format

Oberon00:
We would need to inherit the parent context instead of generating a new one (do we do that already? What about the no-op case?). For traces that are unsampled from the root on, we do need to generate something though. I wonder if the W3C spec should be amended to allow e.g. a special trace id of all F to indicate "no ID available but please propagate traceflags and (if applicable) tracestate anyway.

Oberon00:
I just noticed that we do not specify at all how span or trace IDs are to be generated when creating spans. I think this needs to be resolved before GA. It should be done after #932

Oberon00:
Assigned p2, but could also qualify as p1

Oberon00:
From the SIG meeting: If you have an unsampled child, propagate the parent span ID as my span ID.

Oberon00:
I think #998 addresses the most important part of this issue, but not the actual issue in the title, which would be a "default" **trace** ID. For that we would need to change the sampler API to not accept but optionally return a trace ID, or pass a function that lazily generates the trace ID or something.

Oberon00:
@andrewhsu @open-telemetry/technical-committee I think while the issue in the title is after-ga, the part of the issue where the spec does not say at all what should happen with an unsampled span is required-for-ga (or should at least be allowed-for-ga, not sure what the guidelines are here).

PR #1225, opened 2020-11-13
Specify Span ID creation with sampling (non-recording spans included)

### LINKED ISSUE #1060: Missing spec for Span ID creation for dropped/record-only spans

I am trying to ensure that span ID creation for spans that are dropped or only recorded (but not sampled) is uniform across different language implementations.

**Additional context.**

As explained in https://github.com/open-telemetry/opentelemetry-specification/issues/864#issuecomment-689513021 the spec is silent on what to do for IDs of spans that are not sampled. #998 suggest to reuse the parent-span-id for dropped or use zero if there is none, but use a new span ID for everything else. A simpler alternative would be to simply add the sentence "A new Span ID MUST always be generated, even for dropped spans".

This is not purely an optimization issue, it impacts the observable behavior of the monitored application when the unsampled span is injected (and then potentially used as parent on the other side).

carlosalberto:
Fixes #1060 

Spin-off of #998 (closes #998)

After discussing pros and cons, [it was recommended](https://github.com/open-telemetry/opentelemetry-specification/pull/998/files#r517616726) to always generate new Span Ids even for non-recording instances.

> The TC recommendation is to generate a new span ID even for non-recording spans.

cc @yurishkuro @SergeyKanzhelev


spec-compliance-matrix.md
```diff
@@ -68,6 +68,7 @@ status of the feature is not known.
 |[Sampling](specification/trace/sdk.md#sampling)|
 |Allow samplers to modify tracestate           |   | +  |   | [-](https://github.com/open-telemetry/opentelemetry-python/issues/1220)    |    | +    |   | +  |   |    |
 |ShouldSample gets full parent Context         |   | +  | + | +    |    | +    |   |    |   |    |
+|[Span/Trace ID creation wrt. sampling is compliant](specification/trace/sdk.md#sdk-span-creation) |   |    |   |      |    |      |   |    |   |    |
```
Oberon00, line 71:
```suggestion
|[New Span ID created also for non-recording Spans](specification/trace/sdk.md#sdk-span-creation) |   |    |   |      |    |      |   |    |   |    |
```

Oberon00 (commented):
Comparison to the alternative PR #998: https://github.com/dynatrace-oss-contrib/opentelemetry-specification/compare/sdk-spanid-sampling...carlosalberto:sdk-spanid-sampling

spec-compliance-matrix.md
carlosalberto: Thanks! Applied.

Oberon00:
One point to consider: If we explicitly encourage users to rely on unsampled spans having an ID, we should probably close #864 "Option to allow "default" IDs for unsampled traces" as wontfix. CC @awssandra

bogdandrutu (approved):

specification/trace/sdk.md
```diff
@@ -61,11 +61,28 @@ The following table summarizes the expected behavior for each combination of
...
+3. Generate a new span ID for the `Span`, independently of the sampling decision.
+   This is done so other components (such as logs or exception handling) can rely on
+   a unique span ID, even if the `Span` is a non-recording instance.
+4. Create a span depending on the decision returned by `ShouldSample`:
+   see description of [`ShouldSample`'s](#shouldsample) return value below
+   for how to set `IsRecording` and `Sampled` on the Span,
+   and the [table above](#recording-sampled-reaction-table) on whether
+   to pass the `Span` to `SpanProcessor`s.
+   A non-recording span MAY be implemented using the same mechanism as when a
+   `Span` is created with no API-implementation installed or as described in
```
jkwatson, line 84:
```suggestion
   `Span` is created without an SDK installed or as described in
```

carlosalberto: Thanks!

yurishkuro (approved):

jmacd (approved):

PR #1227, opened 2020-11-13
Yaml files for process semantic convention

### LINKED ISSUE #547: Proposal: Define semantic conventions in JSON/YAML
This is a proposal to define all semantic conventions in a format that can be automatically parse by every language and generate constants without having to manually sync their constant definitions with the markdown text in this repository:
1. Move definitions in opentelemetry-semantic-conventions - languages can use `git submodules` to reference to that repository
1. Specificantion will automatically generate tabels like https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-http.md#common-attributes when the `git submodule` is updated.
1. Languages will automatically re-generate all constants when `git submodule` is updated.

This will allow us to keep consistency between languages and consistent description for every semantic convention defined.

Open questions:
1. Which format to use  JSON/YAML/etc.?
1. Does anyone know about a tool to generate the markdown tables from the JSON? A simple tool can be written, but would like to avoid that.
1. Should languages generate the constant code or read the files at runtime?
    * Personally I prefer generated code - fast initialization, compiler optimizations, etc.

thisthat:
Fixes #547

## Changes

**Editorial change**.
This PR introduces the YAML definitions of Process resource attributes.


specification/resource/semantic_conventions/process.md
```diff
@@ -22,15 +22,25 @@
...
+| `process.command` | string | The command used to launch the process (i.e. the command name). On Linux based systems, can be set to the zeroth string in `proc/[pid]/cmdline`. On Windows, can be set to the first parameter extracted from `GetCommandLineW`. | `cmd/otelcol	` | See below |
+| `process.command_line` | string | The full command used to launch the process as a single string representing the full command. On Windows, can be set to the result of `GetCommandLineW`. Do not set this if you have to assemble it just for monitoring; use `process.command_args` instead. | `C:\cmd\otecol --config="my directory\config.yaml"` | See below |
+| `process.command_args` | string[] | All the command arguments (including the command/executable itself) as received by the process. On Linux-based systems (and some other Unixoid systems supporting procfs), can be set according to the list of null-delimited strings extracted from `proc/[pid]/cmdline`. For libc-based executables, this would be the full argv vector passed to `main`. | `cmd/otecol`<br>`--config=config.yaml ` | See below |
```
thisthat, line 33: I've created https://github.com/open-telemetry/build-tools/issues/23

Oberon00 (approved):

specification/resource/semantic_conventions/process.md
```diff
@@ -22,15 +22,25 @@
...
+| `process.command_args` | string[] | All the command arguments (including the command/executable itself) as received by the process. On Linux-based systems (and some other Unixoid systems supporting procfs), can be set according to the list of null-delimited strings extracted from `proc/[pid]/cmdline`. For libc-based executables, this would be the full argv vector passed to `main`. | `cmd/otecol`<br>`--config=config.yaml` | See below |
+| `process.owner` | string | The username of the user that owns the process. | `root` | No |
+
+**Additional attribute requirements:** At least one of the following sets of attributes is required:
+
+* `process.executable.name`
+* `process.executable.path`
+* `process.command`
+* `process.command_line`
+* `process.command_args`
```
arminru, line 42: Can we split this into a top-level `process` convention only containing `process.pid` and a separate one for executable-related things? Otherwise, if enforced, only reporting the PID but none of the other values wouldn't be possible.

Oberon00: I think this is out of scope here. But if we do that, we can equivalently just remove the anyOf requirement

thisthat: This won't be an editorial change tho.

arminru: Just removing the `anyOf` would be fine with me.

Oberon00: Still, I think we should do that in a separate PR. This one can now be merged quickly (probably today) since it is only an editorial change. The removal of the requirement is a content change.

arminru: The condition in the current requirement makes it pretty weak already and allowed for just reporting the PID, which would no longer be possible with this PR:
> At least one of `process.executable.name`, `process.executable.path`, `process.command`, `process.command_line` or `process.command_args` is required **to allow back ends to identify the executable**.

It could also be interpreted as a note on why it is unconditionally required but I think it was intended as a condition, arguably it's a bit unclear.

Oberon00: I interpret this as "At least one of process.executable.name, process.executable.path, process.command, process.command_line or process.command_args is required. Note: This allows back ends to identify the executable.

thisthat: I don't think was possible to report only the PID with the current requirements.
I read that as a motivation why one of the 5 attributes is required and not as a "nice thing to have but not necessary".

arminru: I think the current wording is vague enough to allow for both interpretations. With the changes introduced in this PR, one of the two options is explicitly chosen, which is not an editorial change either. I don't think there is any justification for prohibiting users to report just PIDs without the command line and would therefore go for the more permissive option. This would either be going without the `anyOf` or splitting it up, where both would lead to the same result.

semantic_conventions/resource/process.yaml
```diff
@@ -0,0 +1,84 @@
...
+        brief: >
+          The version of the runtime of this process, as returned by the runtime
+          without modification.
+        examples: '14.0.2'
+      - id: description
+        type: string
+        brief: >
+          An additional description about the runtime of the process, for example
+          a specific vendor customization of the runtime environment.
+        examples: 'Eclipse OpenJ9 openj9-0.21.0'
```
arminru, line 84:
```suggestion
        examples: 'Eclipse OpenJ9 openj9-0.21.0'

```

specification/resource/semantic_conventions/process.md
Oberon00: Seems that wording was really meant to make it optional: https://github.com/open-telemetry/opentelemetry-specification/pull/1137#discussion_r510923926. Then I'm OK with removing it, if we update the PR title/description.
We could also consider not adding a anyOf but keeping the original wording in the markdown as-is.

thisthat: I prefer keeping the wording. I don't want to introduce any semantic change with this PR.

thisthat: Done in 650f9b1

arminru: We should resolve this ambiguity but I'm fine with restricting this PR to only adding the YAML representation. Now with the latest change the restrictions stay untouched.

arminru (approved):

semantic_conventions/resource/process.yaml
```diff
@@ -0,0 +1,87 @@
...
```
arminru, line 87:
```suggestion
        examples: 'Eclipse OpenJ9 Eclipse OpenJ9 VM openj9-0.21.0'
```
This was changed in #1242.

specification/resource/semantic_conventions/process.md
```diff
@@ -48,11 +50,13 @@ At least one of `process.executable.name`, `process.executable.path`, `process.c
...
-| process.runtime.name | The name of the runtime of this process. For compiled native binaries, this SHOULD be the name of the compiler. | `OpenJDK Runtime Environment` | No |
-| process.runtime.version | The version of the runtime of this process, as returned by the runtime without modification. | `14.0.2` | No |
-| process.runtime.description | An additional description about the runtime of the process, for example a specific vendor customization of the runtime environment. | `Eclipse OpenJ9 openj9-0.21.0` | No |
+<!-- semconv process.runtime -->
+| Attribute  | Type | Description  | Example  | Required |
+|---|---|---|---|---|
+| `process.runtime.name` | string | The name of the runtime of this process. For compiled native binaries, this SHOULD be the name of the compiler. | `OpenJDK Runtime Environment` | No |
+| `process.runtime.version` | string | The version of the runtime of this process, as returned by the runtime without modification. | `14.0.2` | No |
+| `process.runtime.description` | string | An additional description about the runtime of the process, for example a specific vendor customization of the runtime environment. | `Eclipse OpenJ9 openj9-0.21.0` | No |
```
arminru, line 58:
```suggestion
| `process.runtime.description` | string | An additional description about the runtime of the process, for example a specific vendor customization of the runtime environment. | `Eclipse OpenJ9 Eclipse OpenJ9 VM openj9-0.21.0` | No |
```
This was changed in #1242.

PR #1229, opened 2020-11-17
Add YAML definition for code-related semantic attributes.

vovencij:
## Changes
As requested in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1628#discussion_r523246765 this PR adds [source code related span attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/span-general.md#source-code-attributes) to the list of semantic attributes.

PR #1230, opened 2020-11-17
Add .gitattributes & normalize line endings

Oberon00:
See https://github.com/open-telemetry/opentelemetry-specification/pull/1229#issuecomment-728851744


docfx.cmd
```diff
@@ -1,15 +1,15 @@
-SETLOCAL
```
Oberon00, line 1:
```sh
$ file docfx.cmd
docfx.cmd: ASCII text, with CRLF line terminators
```

Seems it used LF before.

arminru (approved):

anuraaga (approved):

PR #1235, opened 2020-11-17
Add Post GA Workflow document.

carlosalberto:
It was discussed in the @open-telemetry/technical-committee  that a document briefly describing the after-GA workflow would be great to add clarity - the contents of this document reflect the related discussion.

### DIFF
--- a/post-ga-workflow.md
+++ b/post-ga-workflow.md
@@ -24,7 +24,8 @@ OTEPs are used as a mean to add entirely new components (such as logging) or to 
 
 ## Releases
 
-Minor releases are expected to approximately happen every 6 months, containing additive, non-breaking changes in the API, as mentioned above.
+Minor releases are expected to happen ad-hoc, containing additive, non-breaking changes in the API, as mentioned above.
+Releases with regular cadence will be considered in the future.
 
 Major releases are allowed to contain breaking changes in the API.
 
--- a/post-ga-workflow.md
+++ b/post-ga-workflow.md
@@ -4,12 +4,12 @@
 
 The following changes to the Specification v1.x will be allowed after GA:
 
-- Editorial changes.
-- API: additive, non-breaking changes. This is done in order to stay
+- API **additive** changes. This is done in order to stay
   compatible with previous versions. Adding entirely new API components,
   such as logging, is thus allowed.
 - SDK changes.
-- Semantic conventions for new libraries.
+- Editorial changes.
+- Semantic conventions additions.
 
 The `main` branch will always contain the latest changes, and branches will be used to track published releases.
 
--- a/post-ga-workflow.md
+++ b/post-ga-workflow.md
@@ -1,0 +1,31 @@
+# Post-GA Workflow
+
+## Specification Changes
+
+The following changes to the Specification v1.x will be allowed after GA:
+
+- Editorial changes.
+- API: additive, non-breaking changes. This is done in order to stay
+  compatible with previous versions. Adding entirely new API components,
+  such as logging, is thus allowed.
+- SDK changes.
+- Semantic conventions for new libraries.
+
+The `master` branch will always contain the latest changes, and branches will be used to track published releases.
+
+API-breaking changes can be accepted under `experimental` subdirectories.
+
+## OTEPS
+
+OTEPs are used as a mean to add entirely new components (such as logging) or to introduce API-breaking changes. They will have similar rules as Specification PRs regarding their lifetime:
+
+- OTEP PRs will be automatically marked stale after 14 days without activity.
+- OTEP PRs will be automatically closed 14 days after being marked stale.
+
+## Releases
+
+Minor releases are expected to approximately happen every 6 months, containing additive, non-breaking changes in the API, as mentioned above.
+
+Major releases are allowed to contain breaking changes in the API.
+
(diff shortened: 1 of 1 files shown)


post-ga-workflow.md
```diff
@@ -0,0 +1,31 @@
...
+The following changes to the Specification v1.x will be allowed after GA:
+
+- Editorial changes.
+- API: additive, non-breaking changes. This is done in order to stay
+  compatible with previous versions. Adding entirely new API components,
+  such as logging, is thus allowed.
+- SDK changes.
+- Semantic conventions for new libraries.
+
+The `master` branch will always contain the latest changes, and branches will be used to track published releases.
```
bogdandrutu, line 14: I think there is an ongoing effort to change the name of the default branch from `master` to `main`. So I would not use that word here.

carlosalberto: Oh, good point ;)

carlosalberto: Updated.

post-ga-workflow.md
```diff
@@ -0,0 +1,31 @@
...
+## OTEPS
+
+OTEPs are used as a mean to add entirely new components (such as logging) or to introduce API-breaking changes. They will have similar rules as Specification PRs regarding their lifetime:
+
+- OTEP PRs will be automatically marked stale after 14 days without activity.
+- OTEP PRs will be automatically closed 14 days after being marked stale.
+
+## Releases
+
+Minor releases are expected to approximately happen every 6 months, containing additive, non-breaking changes in the API, as mentioned above.
```
jkwatson, line 27: this seems extremely optimistic! I would have guessed more like 2 months.

post-ga-workflow.md
```diff
@@ -0,0 +1,31 @@
...
+  compatible with previous versions. Adding entirely new API components,
+  such as logging, is thus allowed.
+- SDK changes.
+- Semantic conventions for new libraries.
+
+The `main` branch will always contain the latest changes, and branches will be used to track published releases.
+
+API-breaking changes can be accepted under `experimental` subdirectories.
+
+## OTEPS
```
arminru, line 18:
```suggestion
## OTEPs
```

post-ga-workflow.md
```diff
@@ -0,0 +1,31 @@
...
+
+OTEPs are used as a mean to add entirely new components (such as logging) or to introduce API-breaking changes. They will have similar rules as Specification PRs regarding their lifetime:
+
+- OTEP PRs will be automatically marked stale after 14 days without activity.
+- OTEP PRs will be automatically closed 14 days after being marked stale.
```
arminru, line 23: I'm not sure if this is the right place to define rules like these. Wouldn't this better go into the community or OTEP repo?

carlosalberto: I'd rather keep it here, and have it in a single place, at least for now (also, there's the rationale that OTEPs are logically an extension of the Specification, so they should be controlled by rules imposed here). Not a strong feeling though.

post-ga-workflow.md
carlosalberto: Removed this, to use an ad-hoc approach for now. We will revisit the cycles in the future ;)

carlosalberto:
@open-telemetry/specs-approvers Please review. Releases cadence expectation was removed, and we will re-visit this part in the future.

post-ga-workflow.md
arminru: No strong opinion either but strictly this is not really about the Post GA spec workflow but applies to the OTEP repo in general.

arminru (approved):

post-ga-workflow.md
```diff
@@ -0,0 +1,32 @@
+# Post-GA Workflow
+
+## Specification Changes
+
+The following changes to the Specification v1.x will be allowed after GA:
+
+- Editorial changes.
+- API: additive, non-breaking changes. This is done in order to stay
```
Oberon00, line 8: Non-breaking changes should be accepted for everything. E.g. adding a new semantic attribute for an existing library is not allowed currently according to this list.

post-ga-workflow.md
```diff
@@ -0,0 +1,32 @@
...
+  compatible with previous versions. Adding entirely new API components,
+  such as logging, is thus allowed.
+- SDK changes.
+- Semantic conventions for new libraries.
```
Oberon00, line 12: I think "library" is the wrong word here. What about semantic conventions that are not related to a particular library? For example we might want to add a semantic convention for reporting information about the active auto-instrumenting agent, or a new attribute to the existing http convention.

In fact, I think we do not have any library-specific semantic conventions in the spec currently. We have some tech-specific ones, like cassandra-specific attributes.

carlosalberto: What about "Additional semantic conventions"? or "New semantic conventions"?

post-ga-workflow.md
carlosalberto: Oh, definitely - this was done to have a strong remark that for API, it's really, really important. Probably it can be worded as "API: only additive changes. This is done... " etc.

post-ga-workflow.md
Oberon00: Non-breaking changes only? I.e., mostly only add (non-required) attributes (also to existing conventions).

carlosalberto: Updated. Please review ;)

Issue #1215, opened 2020-11-09
Ability to name logger - GetLogger(name) and retain this information on event


maxgolov:
# What are you trying to achieve?

Most prominent frameworks across different programming languages allow to `name` a Logger. This is typically done at `GetLogger(name)` API call, where subsequently the knowledge of logger name is retained on Logger object. When SDK flows the data emitted from the Logger to Exporter, it should be possible for the Logger to pass-thru not only the Event name, but Logger name that emitted this event.

# Additional context.

While we are designing Logging API in OpenTelemetry C++ SDK, we are facing a challenge that the spec does not prescribe the Logger to have a `name` per se. Whereas if you build a shim or exporter from other prominent logging SDKs to OpenTelemetry SDK, you'd actually have to 'remember' the Logger name passed down to it.

# References

Logging frameworks that provide named loggers and provide ability to search/grep/export based on logger name:

- log4j: https://logging.apache.org/log4j/1.2/manual.html - Cit. : `Loggers are named entities`
- log4cxx: https://logging.apache.org/log4cxx/latest_stable/usage.html
- Python logging cookbook: https://docs.python.org/3/howto/logging-cookbook.html
- Xamarin / C# Logging API: https://docs.microsoft.com/en-us/dotnet/api/java.util.logging.logger.getlogger?view=xamarin-android-sdk-9

Example showing visually why it is important for an `Event` to carry information about what logger emitted this event:

![image](https://user-images.githubusercontent.com/34072974/98602401-40f1ec00-2295-11eb-84e2-5148d13afefb.png)

Column refers explicitly to what `logger name` it is.

Similar logic also applies to Android OS / platform code, where there is no `logger name` concept per se, but it can be implicitly mapped to `LOG_TAG`: https://stackoverflow.com/questions/8355632/how-do-you-usually-tag-log-entries-android - which allows to group / filter by your "named logger" events from the unified stream of all Logcat messages / events.

maxgolov:
@arminru @andrewhsu

Note that two most popular 3rd party C# libraries also provide the same API / method - `GetLogger(name)` :
- log4net: https://logging.apache.org/log4net/release/sdk/html/M_log4net_LogManager_GetLogger_2.htm
- NLog: https://nlog-project.org/documentation/v4.3.0/html/M_NLog_LogManager_GetLogger.htm

[...]

PR #1236, opened 2020-11-18
Add LoggerName to Log Data Model

tigrannajaryan:
Resolves https://github.com/open-telemetry/opentelemetry-specification/issues/1215

It appears that logger name is sufficiently different to warrant a new concept in
Otel logs. For detailed discussion see the link above.

### DIFF
--- a/specification/logs/data-model.md
+++ b/specification/logs/data-model.md
@@ -155,6 +155,7 @@ SpanId         |Request span id.
 TraceFlags     |W3C trace flag.
 SeverityText   |The severity text (also known as log level).
 SeverityNumber |Numerical value of the severity.
+LoggerName     |Name of the logger that is used for producing the log record.
 Name           |Short event identifier.
 Body           |The body of the log record.
 Resource       |Describes the source of the log.
@@ -354,6 +355,13 @@ comparisons `SeverityNumber` field should be used. `SeverityNumber` can be
 compared to another `SeverityNumber` or to numbers in the 1..24 range (or to the
 corresponding short names).
 
+### Field: `LoggerName`
+
+The name of the logger that is used for producing the log record. The logger
+name is typically chosen by the developer to group related log records. Often
+the class name, package name, module name or other logical unit of the
+application is selected as the logger name.
+
 ### Field: `Name`
 
 Type: string.
@@ -434,7 +442,8 @@ Example 1
   "SpanId": "43222c2d51a7abe3",
   "SeverityText": "INFO",
   "SeverityNumber": 9,
-  "Body": "20200415T072306-0700 INFO I like donuts"
+  "Body": "20200415T072306-0700 INFO I like donuts",
+  "LoggerName": "com.example.App.MyClass"
 }
 ```
 
@@ -717,6 +726,12 @@ Rest of SDIDs -> Attributes["syslog.*"]</td>
     <td>Severity</td>
   </tr>
   <tr>
+    <td>Logger</td>
+    <td>string</td>
+    <td>Logger name.</td>
+    <td>LoggerName</td>
+  </tr>
+  <tr>
     <td>Message</td>
     <td>string</td>
     <td>Human readable message.</td>
@@ -890,7 +905,7 @@ Field            | Type               | Description                             
 -----------------|--------------------| ------------------------------------------------------- | ---------------------------
 timestamp        | string             | The time the event described by the log entry occurred. | Timestamp
 resource         | MonitoredResource  | The monitored resource that produced this log entry.    | Resource
-log_name         | string             | The URL-encoded LOG_ID suffix of the log_name field identifies which log stream this entry belongs to. | Name
+log_name         | string             | The URL-encoded LOG_ID suffix of the log_name field identifies which log stream this entry belongs to. | LoggerName
 json_payload     | google.protobuf.Struct | The log entry payload, represented as a structure that is expressed as a JSON object. | Body
 proto_payload    | google.protobuf.Any | The log entry payload, represented as a protocol buffer. | Body
 text_payload     | string             | The log entry payload, represented as a Unicode string (UTF-8). | Body


specification/logs/data-model.md
```diff
@@ -434,7 +442,8 @@ Example 1
   "SpanId": "43222c2d51a7abe3",
   "SeverityText": "INFO",
   "SeverityNumber": 9,
-  "Body": "20200415T072306-0700 INFO I like donuts"
+  "Body": "20200415T072306-0700 INFO I like donuts",
+  "LoggerName": "com.example.App.MyClass"
```
bogdandrutu, line 446: Is this not the "InstrumentationName"?

bogdandrutu (dismissed):

specification/logs/data-model.md
tigrannajaryan: See discussion https://github.com/open-telemetry/opentelemetry-specification/issues/1215
The consensus is that InstrumentationName is different from LoggerName.

zenmoto (approved):

specification/logs/data-model.md
bogdandrutu: If this is the case we definitely failed with `InstrumentationName` and we should probably remove that. This is hitting again and again.

@Oberon00 @arminru as people who pushed on the `InstrumentationName` which creates again more confusion.

zenmoto: I would think that in this case `InstrumentationName` would be the instrumentation or adapter that is being used to collect these events - something like "Log4J". `LoggerName` is determined by the author of the code being instrumented, not the instrumenting code. In Java it might be a package and class name, in other languages it might be "database_access", or "database::access", or any number of schemes. The commonality is that it's determined without OpenTelemetry necessarily in mind, but is rather following the conventions of the language or framework or application developer rather than our conventions.

Another way to cut it would be in cardinality- for a Java application you could adapt both Log4J and Logback. Hopefully from the exporter events from each would be indistinguishable from the other, but each will contain many (almost always at least hundreds) of logger names.

A Spring application using Log4J will have logging events delivered from application code via Log4J and events from the Spring framework via Logback. Both types of events should be grouped in a containing trace. We may also want to see which ones came via one route or the other. I think that both are important, but distinct.

bogdandrutu: @zenmoto "Log4J" will be the [`telemetry.sdk.name`](https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/resource/semantic_conventions#telemetry-sdk). So definitely the name of the library has it's own place.

> `LoggerName` is determined by the author of the code being instrumented, not the instrumenting code.

It is the same for metrics and traces, but it depends who is the author of the "code". If the logging happens via a plugin or is embedded in the code itself.

Also the APIs that we right now have for getting a tracer (see https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#get-a-tracer) is exactly like the "GetLogger", I see no difference.

yurishkuro: +1 to what Bogdan said. I am not seeing a difference between LoggerName and InstrumentationName.

tigrannajaryan: @bogdandrutu @yurishkuro I am not sure I understand your suggestion. Are you suggesting to put the class name in the `InstrumentationLibrary.Name` field? How can a _library_ name field be a good place for a _class_ name? 
What if I want to record _both_ the class name and the library name? Logger name is very often used to store the class name. I imagine it to also become quite common to record the library name in the log record, and do this consistently and uniformly to how we do that for traces. If we use `InstrumentationLibrary.Name` for the logger name (which often is _not_ the library name) then we can no longer record the library name there.

jmacd: The difference I see is that the logger name examples above use a hierarchical name, e.g., so that "database::access" is a subset of "database". In the otel-go-contrib repo, we've been using a practice of using the URL of the repository as the `InstrumentationLibrary.Name`, but I'm not sure this is a spec-level recommendation, but if we recommend `InstrumentationLibrary.Name` be a URL, then I think we can satisfy the intended use-case for GetLogger using either a `/sub/path` or a `#fragment/name` in the `InstrumentationLibrary.Name` field.

tigrannajaryan: If we want to use `InstrumentationLibrary.Name` field as a place to record the Logger Name, then at the minimum we need to rename the field to something that has the semantics that allow the class name to be a valid value for it. As of now `InstrumentationLibrary.Name` is clearly a library name, class name does not fit there no matter how you slice it.
Renaming does not solve the second problem that I mentioned (inability to record both class and library name simultaneously) but perhaps we can live with that.

If we intend to do what you suggest then `InstrumentationLibrary` needs to be renamed to something like `ApplicationUnitName`. In that case `ApplicationUnitName.Name` can contain a library name or a class name as desired.

tigrannajaryan: >Also the APIs that we right now have for getting a tracer (see https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#get-a-tracer) is exactly like the "GetLogger", I see no difference.

I think this is exactly the difference we are raising. Unlike `GetTracer`, `GetLogger` API first of all should accept a logger name, and may also a library name as second parameter, but logger name is the important one. This is just how logging libraries historically worked and I think we cannot ignore this legacy.

zenmoto: I think that the logger name may be closest in concept to a log event 'category'. How specific that category is depends on the conventions (class vs module vs function), but it is usually intended as a way of grouping together related events. I don't see how that intent would correlate with `InstrumentationLibrary.Name`.

bogdandrutu: > I think this is exactly the difference we are raising. Unlike GetTracer, GetLogger API first of all should accept a logger name, and may also accept a library name as second parameter, but logger name is the important one. This is just how logging libraries historically worked and I think we cannot ignore this legacy.

Completely lost about this. "GetLogger API first of all should accept a logger name" what do you mean? Not sure I understand what you mean by historically worked like that, what we try to say is that the GetTracer API is equivalent with legacy GetLogger, and we don't see any difference.

tigrannajaryan: I mean historically in other logging libraries, not in Otel. `getLogger` exists in other logging libraries (e.g. Java [Logger](https://docs.oracle.com/javase/7/docs/api/java/util/logging/Logger.html#getLogger(java.lang.String))) and often the name of the class is passed the parameter to `getLogger()`. 

Otel logging SDK and logging protocol MUST be able to work with non-Otel logging libraries. It should be possible to use non-Otel logging library to generate Otel-compliant log record. It is one the [listed requirements](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/logs/overview.md#legacy-first-party-applications-logs).

When a non-Otel logging library allows the user to obtain a logger via `getLogger(<name>)` call, we will have to then map the name of the logger to one of the fields in Otel data model and in OTLP protocol when exporting. 

There is currently no field in Otel data model and in OTLP that is suitable for recording the logger name. I think you are suggesting that `InstrumentationLibrary.Name` is that field. If that is your suggestion then I believe it is not a good fit. `InstrumentationLibrary.Name` has a definition that semantically does not match the logger name definition in all other logging libraries.

maxgolov: Can this be `LogName` or `LoggerName` or `Category` or `Module` or `Source` ?

maxgolov: In `Event Tracing for Windows` [here](https://docs.microsoft.com/en-us/windows/win32/api/evntrace/ns-evntrace-event_trace_logfilea) - it is called `LoggerName`

yurishkuro: > `InstrumentationLibrary.Name` has a definition that semantically does not match the logger name definition in all other logging libraries.

@tigrannajaryan I think your argument is being weaken by mentioning the class name, because class names map pretty well here, especially if we were to rename InstrumentationLibrary to InstrumentationComponent or something. However, where it does not fit at all is when the logger name is NOT a class name but an arbitrary hierarchical name, which is totally allowed by most traditional loggers. For example, it can be even something dynamic like `/customer/{customerID}`.

I think it is worth analyzing this a bit more. What are logger names usually used for? Off the top of my head:
  * can be just a label added to the log lines
  * different threshold levels can be configured per logger
  * different sinks
  * different message formatters? (not sure if that's actually true, or if it's more per sink - makes more sense per sink)
  * at least in log4j the names are hierarchical and propagate the above settings to subordinate loggers

Given these dependencies on the logger name, I am not sure InstrumentationLibrary/Component even makes sense for loggers, since a custom name like `/customer/{customerID}` could be reused across different layers of the application, making InstrumentationX resource attribute meaningless.

pmm-sumo: There are some interesting examples from the specification that to me eyes do not make `InstrumentationLibrary.Name` a consistent choice for storing the logger name:

1) [Overview/Instrumentation Libraries](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/overview.md#instrumentation-libraries)

> A library that enables OpenTelemetry observability for another library is called an Instrumentation Library.

As @zenmoto has noticed, I think `Log4j` or so would be a better fit (though not ideal, I think the instrumentation library is a more fine granular concept that does not really exist for logging libraries) for this definition rather than `com.example.App.MyClass`  

2) [Tracing API Specification](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#get-a-tracer) says following:
> name (required): This name must identify the instrumentation library (e.g. `io.opentelemetry.contrib.mongodb`) and _not_ the instrumented library.

I think this is in conflict with how logger names are being used, where they frequently point to the class in which the record was generated.

3) [Tracing Source Code Attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/span-general.md#source-code-attributes)

> Often a span is closely tied to a certain unit of code that is logically responsible for handling the operation that the span describes (usually the method that starts the span) _(followed by a table with attributes, such as `code.function` or `code.namespace`)_

I think that `code.namespace` for spans is an example of a close (though not ideal) match for `LoggerName` 


To me, `InstrumentationLibrary.Name` would work for logs if there would exist e.g. some automagic libraries that create logs for a given framework (as might happen for traces and metrics)

kumoroku (approved):

specification/logs/data-model.md
jmacd: Let's save I've read this thread and agree that we should have a new field. Is there a name that would make this concept less logging specific?  There are ways that "LoggerName" is like a resource, and if so we could establish a semantic convention for "LoggerName". However, we've resisted adding non-process-identifying resource attributes, as far as I know, and LoggerName is scoped to part of a process.

I've advocated for "resource scopes" in the past: https://github.com/open-telemetry/oteps/issues/78

In that proposal, we might just add a resource-scoped "logger.name=<your_class_name_or_whatever>" attribute that would apply to all (spans, logs, metrics) when used.

tigrannajaryan: @jmacd I agree on several points of your proposal for resource scopes. However, I believe it is orthogonal and more complex issue, which also likely requires changes to traces and metrics.

Could the "logger name" be modeled as a "scope" if we had resource scopes? Possibly. However, we don't have them and I think it is likely that we won't have them for a while (if ever). I do not want to make logs wait for something that may not be accepted.

Again, I generally like your "scopes" proposal and I think it is very useful for representing certain types of data, but I think the "logger name" can be introduced as a first-class concept independently from the scopes.

tigrannajaryan: >However, where it does not fit at all is when the logger name is NOT a class name but an arbitrary hierarchical name, which is totally allowed by most traditional loggers. For example, it can be even something dynamic like /customer/{customerID}.

>Given these dependencies on the logger name, I am not sure InstrumentationLibrary/Component even makes sense for loggers, since a custom name like /customer/{customerID} could be reused across different layers of the application, making InstrumentationX resource attribute meaningless.

@yurishkuro since you agree that the "logger name" does not belong to the InstrumentationLibrary.Name do you also agree with the suggestion that this PR makes, i.e. to introduce a new LoggerName field or you have a different proposal?

yurishkuro: I'm ok with it in the data model.

yurishkuro (approved):
LGTM

tigrannajaryan:
@maxgolov this PR addresses the issues that you opened. Please review.

tigrannajaryan:
I am going to remove the "Stale" label to give people more time. Most of Otel is focused on making the GA release so this can wait a bit.

tigrannajaryan:
Re-opening. This is not stale, it is deliberately postponed until it is time to focus on logs.

pmm-sumo (approved):

Oberon00:
> Re-opening. This is not stale, it is deliberately postponed until it is time to focus on logs.

Maybe you want to convert it to a draft for now? The the stale bot will leave it alone. Or you can mark it as after-ga (see https://github.com/open-telemetry/opentelemetry-specification/blob/master/.github/workflows/stale-pr.yaml#L15)

tigrannajaryan:
@Oberon00 looks like converting to draft did not help. Judging from bot's config `release:after-ga` label should help. Trying that.

Oberon00:
Hmm, I thought the bot was configured that way. Maybe it should be?

reyang (dismissed):

SergeyKanzhelev (approved):

specification/logs/data-model.md
rakyll: This seems to be a logical grouping name like "log group". Vendors call it a group, folder, etc.

specification/logs/data-model.md
```diff
@@ -890,7 +905,7 @@ Field            | Type               | Description
 -----------------|--------------------| ------------------------------------------------------- | ---------------------------
 timestamp        | string             | The time the event described by the log entry occurred. | Timestamp
 resource         | MonitoredResource  | The monitored resource that produced this log entry.    | Resource
-log_name         | string             | The URL-encoded LOG_ID suffix of the log_name field identifies which log stream this entry belongs to. | Name
+log_name         | string             | The URL-encoded LOG_ID suffix of the log_name field identifies which log stream this entry belongs to. | LoggerName
```
rakyll, line 908: Maybe it could be named something like a "log_group" to be more representative it's about logical categorization.

reyang: In .NET it is called "log category name" https://docs.microsoft.com/dotnet/api/microsoft.extensions.logging.ilogger-1.

tigrannajaryan:
I converted this back to Ready To Review so that we can resume the discussion. We need a resolution on this so that Logs data model can be declared stable.

tigrannajaryan:
@bogdandrutu since you objected to this change can you please read the comment thread here https://github.com/open-telemetry/opentelemetry-specification/pull/1236#discussion_r526187895 
The discussion was a long time ago, so re-reading may be useful to remember what it was about.

We discussed this today in Log SIG meeting and we could not find good supporting arguments that justify the usage  `InstrumentationLibrary.Name` to record the logger name. On the contrary, we still see that `InstrumentationLibrary.Name`  is not a good match for what logger name is. The arguments that explain why are in that comment thread, please have a look and reply.

I will also add this to the next spec SIG meeting or if the time there does not permit I will invite you to the Log SIG meeting to talk about this.

tigrannajaryan:
Added to next Spec SIG meeting agenda. @open-telemetry/specs-logs-approvers if you can please attend to discuss (next Tuesday).

tigrannajaryan:
From spec SIG meeting today:

- We will explore if we can tell that instrumentation library name is intended to capture the scope of instrumentation (and library is one of the possible scopes).
- We will also look at the possibility to rename "instrumentation library" to "instrumentation scope" everywhere (should make sure it is not a breaking change).
- If the first item above is doable then we can probably declare that Logger Name goes into instrumentation library name, there is no longer a semantic mismatch.
- This results in inability to record which logging library and which appender was used (was previously thought to be the instrumentation library name). Do we need a separate field / semantic convention to record this?

pmm-sumo:
> * This results in inability to record which logging library and which appender was used (was previously thought to be the instrumentation library name). Do we need a separate field / semantic convention to record this?

There was a lot going on that topic today, so not sure if I understood it correctly but I think one of the ideas was to expand [`telemetry.sdk`](https://github.com/open-telemetry/opentelemetry-specification/blob/b46bcab5fb709381f1fd52096a19541370c7d1b3/specification/resource/semantic_conventions/README.md#telemetry-sdk) semantic conventions to capture such information (e.g. `telemetry.sdk.library.name` or so?)

tigrannajaryan:
> > * This results in inability to record which logging library and which appender was used (was previously thought to be the instrumentation library name). Do we need a separate field / semantic convention to record this?
> 
> There was a lot going on that topic today, so not sure if I understood it correctly but I think one of the ideas was to expand [`telemetry.sdk`](https://github.com/open-telemetry/opentelemetry-specification/blob/b46bcab5fb709381f1fd52096a19541370c7d1b3/specification/resource/semantic_conventions/README.md#telemetry-sdk) semantic conventions to capture such information (e.g. `telemetry.sdk.library.name` or so?)

@bogdandrutu I think suggested that perhaps existing [telemetry.sdk.name](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions#telemetry-sdk) convention to record logging library name. I am not sure that this the right approach. I would expect that we put `opentelemetry` there when using `Log4+Otel Logging SDK` combo.

tigrannajaryan:
Converting to draft again since we are currently looking in the possibility to avoid having this field (in which case this  PR is not needed).

zenmoto:
Using instrumentation library seems to be a mismatch in intents- as an example, a tracing instrumentation library name we use for Kafka is `io.opentelemetry.kafka-clients-0.11`. A logger name used in a Kafka client would be `org.apache.clients.ClientUtils`. A log entry output by Kafka is put out by the SLF4J API, which means that it's attached to a library or adapter like Logback or Log4j, which in turn will have an OpenTelemetry exporter. It seems to me that using instrumentation library for logger name both hides the logger name (which is a primary piece of data in a lot of logging schemes) and precludes storing the data for the OpenTelemetry exporter to know how the data got into the pipeline.

tigrannajaryan:
> Using instrumentation library seems to be a mismatch in intents- as an example, a tracing instrumentation library name we use for Kafka is `io.opentelemetry.kafka-clients-0.11`. A logger name used in a Kafka client would be `org.apache.clients.ClientUtils`. 

If we accept the suggestion that "instrumentation library" concept should be actually named "instrumentation scope" in Otel then it is not a mismatch anymore, is it? You can say that the logs emitted by Kafka client indeed happened in the `org.apache.clients.ClientUtils` instrumentation scope. Would that be wrong to say?

>A log entry output by Kafka is put out by the SLF4J API, which means that it's attached to a library or adapter like Logback or Log4j, which in turn will have an OpenTelemetry exporter. It seems to me that using instrumentation library for logger name both hides the logger name (which is a primary piece of data in a lot of logging schemes) 

If the logger name is mandated to be recorded in the "instrumentation library name" field then how does it hide the logger name? Anyone who wants to find out the logger name know exactly where to look it for, right?

>and precludes storing the data for the OpenTelemetry exporter to know how the data got into the pipeline.

What other information about "how the data got into the pipeline" is necessary to be stored which is prevented by the suggested approach?

zenmoto:
Looking at the Java Instrumentation libraries, I found [this explanation](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/d20961621e19c2c010be91475ca57400ddd343ae/docs/contributing/using-instrumenter-api.md?plain=1#L109):

> The instrumentation name, which indicates the _instrumentation_ library name, not the
>   _instrumented_ library name. The value passed here should uniquely identify the instrumentation
>   library so that during troubleshooting it's possible to determine where the telemetry came from.

Having the Kafka class name in this instance doesn't really help know how the log got there in this instance. It does, however, give critical context as to what the log message _means_, in that it immediately lets us know that this is client code- it is information context on the _instrumented_ code, rather than on the _instrumentation_.

It seems that the cardinality of the data is also an indication of the mismatch- in other parts of OpenTelemetry like tracing, there is one instrumentation library for one piece of functionality dedicated to extracting that instrumentation. With logging the logger name is unbounded and often associated with a class, so it's likely for a single application to have hundreds or even thousands of logger names. It's common practice to use a subset of the name (`org.apache.kafka.clients.*`) to filter on logs that belong to a particular application component. This isn't something we would do with the instrumentation library in any other context as far as I know.

That seems to be an entirely different concept than what's expressed in the instrumentation document.

tigrannajaryan:
>Having the Kafka class name in this instance doesn't really help know how the log got there in this instance. It does, however, give critical context as to what the log message means, in that it immediately lets us know that this is client code- it is information context on the instrumented code, rather than on the instrumentation.

Can we argue that obtaining a logger and emitting a log via the logger _is_ in this case the instrumentation action, so the Kafka class is where the instrumentation happens and thus it is valid to use Kafka class name as the "instrumentation scope".


> That seems to be an entirely different concept than what's expressed in the instrumentation document.

We discussed this in the Spec SIG meeting (I think last week or the one before that) and the general sentiment was that the current definition/naming of "instrumentation library" does not really reflect the original intent, which was about the "scope" (and a library is only possible scope). The entire thing was originally called a "Component", see the the PR which introduced the concept: https://github.com/open-telemetry/oteps/pull/84
It eventually got merged as an "Instrumentation Library", but I believe the recent Spec SIG discussion led to a conclusion that the original intent is what we really want, so expanding this to include class names and other "scopes" that have higher cardinality does not seem to be a bad idea.

So, as posted above we will also try this:
>We will also look at the possibility to rename "instrumentation library" to "instrumentation scope" everywhere (should make sure it is not a breaking change).

zenmoto:
I was convinced by this and arguments by @jack-berg and @djaglowski in the Log SIG meeting.

tigrannajaryan:
We are waiting for some benchmarking result from @jack-berg to make sure it is possible implement the LoggerName->InstrumentationLibraryName mapping in a reasonably cost-effective way.

jack-berg:
Ok so I've done some benchmarking. I tested 3 scenarios:

1. `logEmitterWithLookup` - the existing behavior where for each log you lookup a log emitter from a map based on instrumentation name, such that log emitters are reused when the same instrumentation name is used. E.g. `logEmitterProvider.get("my-logger").logBuilder().setBody(..).emit()`
2. `logEmitterNoLookup  ` - a slight modification of scenario 1. The API stays the same, but the implementation creates a new log emitter for each instrumentation name instead of looking up in a map and reusing. This is nice because it allows the API to better mirror the pattern of tracing & metrics. So the same `logEmitterProvider.get("my-logger").logBuilder().setBody(..).emit()` pattern.
3. `singleLogEmitter` - a variation in which there is a single log emitter, and instrumentation name is passed as an argument when building a log. This avoids the map lookup, but results in a different API compared to metrics and traces. Usage looks like: `logEmitterProvider.get().logBuilder("my-logger").setBody(...).emit()`

I tested these various scenarios with different number of loggers in play (10, 100, 1000, 10000). This is important because the number of loggers dictates how high of a fixed cost you pay with a map lookup. 

Other details:
- Ran with a single thread. Multiple threads could impact performance for the map lookup case, but shouldn't play too big of a role. 
- Warmup of 5 iterations, each for 5 seconds.
- After warmup, ran 5 iterations for 10 seconds each.
- Ran on my local macbook pro with all other programs closed. Not a perfect testing environment, but reasonable. 
- Source code for benchmarks lives [here](https://github.com/jack-berg/opentelemetry-java/blob/log-benchmark/sdk/logs/src/jmh/java/io/opentelemetry/sdk/logs/LogsBenchmarks.java).

Here are the results:
```
Benchmark                                                         (numLoggers)  Mode  Cnt     Score    Error   Units
LogsBenchmarks.logEmitterNoLookup                                           10  avgt    5   110.966 ±  5.679   ns/op
LogsBenchmarks.logEmitterNoLookup:·gc.alloc.rate                            10  avgt    5  1767.821 ± 91.435  MB/sec
LogsBenchmarks.logEmitterNoLookup:·gc.alloc.rate.norm                       10  avgt    5   216.000 ±  0.001    B/op
LogsBenchmarks.logEmitterNoLookup:·gc.churn.G1_Eden_Space                   10  avgt    5  1771.441 ± 88.419  MB/sec
LogsBenchmarks.logEmitterNoLookup:·gc.churn.G1_Eden_Space.norm              10  avgt    5   216.443 ±  1.218    B/op
LogsBenchmarks.logEmitterNoLookup:·gc.churn.G1_Old_Gen                      10  avgt    5     0.005 ±  0.008  MB/sec
LogsBenchmarks.logEmitterNoLookup:·gc.churn.G1_Old_Gen.norm                 10  avgt    5     0.001 ±  0.001    B/op
LogsBenchmarks.logEmitterNoLookup:·gc.count                                 10  avgt    5   421.000           counts
LogsBenchmarks.logEmitterNoLookup:·gc.time                                  10  avgt    5   172.000               ms
LogsBenchmarks.logEmitterNoLookup                                          100  avgt    5   109.505 ±  0.372   ns/op
LogsBenchmarks.logEmitterNoLookup:·gc.alloc.rate                           100  avgt    5  1791.178 ±  5.650  MB/sec
LogsBenchmarks.logEmitterNoLookup:·gc.alloc.rate.norm                      100  avgt    5   216.000 ±  0.001    B/op
LogsBenchmarks.logEmitterNoLookup:·gc.churn.G1_Eden_Space                  100  avgt    5  1792.352 ± 37.239  MB/sec
LogsBenchmarks.logEmitterNoLookup:·gc.churn.G1_Eden_Space.norm             100  avgt    5   216.142 ±  4.894    B/op
LogsBenchmarks.logEmitterNoLookup:·gc.churn.G1_Old_Gen                     100  avgt    5     0.004 ±  0.006  MB/sec
LogsBenchmarks.logEmitterNoLookup:·gc.churn.G1_Old_Gen.norm                100  avgt    5    ≈ 10⁻³             B/op
LogsBenchmarks.logEmitterNoLookup:·gc.count                                100  avgt    5   426.000           counts
LogsBenchmarks.logEmitterNoLookup:·gc.time                                 100  avgt    5   195.000               ms
LogsBenchmarks.logEmitterNoLookup                                         1000  avgt    5   108.054 ±  1.318   ns/op
LogsBenchmarks.logEmitterNoLookup:·gc.alloc.rate                          1000  avgt    5  1815.006 ± 22.740  MB/sec
LogsBenchmarks.logEmitterNoLookup:·gc.alloc.rate.norm                     1000  avgt    5   216.000 ±  0.001    B/op
LogsBenchmarks.logEmitterNoLookup:·gc.churn.G1_Eden_Space                 1000  avgt    5  1821.016 ± 44.408  MB/sec
LogsBenchmarks.logEmitterNoLookup:·gc.churn.G1_Eden_Space.norm            1000  avgt    5   216.715 ±  4.390    B/op
LogsBenchmarks.logEmitterNoLookup:·gc.churn.G1_Old_Gen                    1000  avgt    5     0.002 ±  0.008  MB/sec
LogsBenchmarks.logEmitterNoLookup:·gc.churn.G1_Old_Gen.norm               1000  avgt    5    ≈ 10⁻⁴             B/op
LogsBenchmarks.logEmitterNoLookup:·gc.count                               1000  avgt    5   361.000           counts
LogsBenchmarks.logEmitterNoLookup:·gc.time                                1000  avgt    5   180.000               ms
LogsBenchmarks.logEmitterNoLookup                                        10000  avgt    5   112.914 ±  1.862   ns/op
LogsBenchmarks.logEmitterNoLookup:·gc.alloc.rate                         10000  avgt    5  1737.001 ± 27.790  MB/sec
LogsBenchmarks.logEmitterNoLookup:·gc.alloc.rate.norm                    10000  avgt    5   216.000 ±  0.001    B/op
LogsBenchmarks.logEmitterNoLookup:·gc.churn.G1_Eden_Space                10000  avgt    5  1737.752 ± 43.936  MB/sec
LogsBenchmarks.logEmitterNoLookup:·gc.churn.G1_Eden_Space.norm           10000  avgt    5   216.098 ±  7.363    B/op
LogsBenchmarks.logEmitterNoLookup:·gc.churn.G1_Old_Gen                   10000  avgt    5     0.004 ±  0.005  MB/sec
LogsBenchmarks.logEmitterNoLookup:·gc.churn.G1_Old_Gen.norm              10000  avgt    5    ≈ 10⁻³             B/op
LogsBenchmarks.logEmitterNoLookup:·gc.count                              10000  avgt    5   413.000           counts
LogsBenchmarks.logEmitterNoLookup:·gc.time                               10000  avgt    5   194.000               ms
LogsBenchmarks.logEmitterWithLookup                                         10  avgt    5   110.716 ±  0.985   ns/op
LogsBenchmarks.logEmitterWithLookup:·gc.alloc.rate                          10  avgt    5  1377.803 ± 12.116  MB/sec
LogsBenchmarks.logEmitterWithLookup:·gc.alloc.rate.norm                     10  avgt    5   168.000 ±  0.001    B/op
LogsBenchmarks.logEmitterWithLookup:·gc.churn.G1_Eden_Space                 10  avgt    5  1380.064 ± 44.644  MB/sec
LogsBenchmarks.logEmitterWithLookup:·gc.churn.G1_Eden_Space.norm            10  avgt    5   168.277 ±  5.986    B/op
LogsBenchmarks.logEmitterWithLookup:·gc.churn.G1_Old_Gen                    10  avgt    5     0.003 ±  0.008  MB/sec
LogsBenchmarks.logEmitterWithLookup:·gc.churn.G1_Old_Gen.norm               10  avgt    5    ≈ 10⁻³             B/op
LogsBenchmarks.logEmitterWithLookup:·gc.count                               10  avgt    5   328.000           counts
LogsBenchmarks.logEmitterWithLookup:·gc.time                                10  avgt    5   155.000               ms
LogsBenchmarks.logEmitterWithLookup                                        100  avgt    5   114.369 ±  2.505   ns/op
LogsBenchmarks.logEmitterWithLookup:·gc.alloc.rate                         100  avgt    5  1524.208 ± 33.385  MB/sec
LogsBenchmarks.logEmitterWithLookup:·gc.alloc.rate.norm                    100  avgt    5   192.001 ±  0.005    B/op
LogsBenchmarks.logEmitterWithLookup:·gc.churn.G1_Eden_Space                100  avgt    5  1526.819 ± 44.564  MB/sec
LogsBenchmarks.logEmitterWithLookup:·gc.churn.G1_Eden_Space.norm           100  avgt    5   192.330 ±  3.775    B/op
LogsBenchmarks.logEmitterWithLookup:·gc.churn.G1_Old_Gen                   100  avgt    5     0.002 ±  0.005  MB/sec
LogsBenchmarks.logEmitterWithLookup:·gc.churn.G1_Old_Gen.norm              100  avgt    5    ≈ 10⁻⁴             B/op
LogsBenchmarks.logEmitterWithLookup:·gc.count                              100  avgt    5   363.000           counts
LogsBenchmarks.logEmitterWithLookup:·gc.time                               100  avgt    5   172.000               ms
LogsBenchmarks.logEmitterWithLookup                                       1000  avgt    5   127.349 ±  2.318   ns/op
LogsBenchmarks.logEmitterWithLookup:·gc.alloc.rate                        1000  avgt    5  1539.975 ± 27.791  MB/sec
LogsBenchmarks.logEmitterWithLookup:·gc.alloc.rate.norm                   1000  avgt    5   216.000 ±  0.001    B/op
LogsBenchmarks.logEmitterWithLookup:·gc.churn.G1_Eden_Space               1000  avgt    5  1539.511 ± 35.603  MB/sec
LogsBenchmarks.logEmitterWithLookup:·gc.churn.G1_Eden_Space.norm          1000  avgt    5   215.935 ±  3.414    B/op
LogsBenchmarks.logEmitterWithLookup:·gc.churn.G1_Old_Gen                  1000  avgt    5     0.002 ±  0.007  MB/sec
LogsBenchmarks.logEmitterWithLookup:·gc.churn.G1_Old_Gen.norm             1000  avgt    5    ≈ 10⁻⁴             B/op
LogsBenchmarks.logEmitterWithLookup:·gc.count                             1000  avgt    5   366.000           counts
LogsBenchmarks.logEmitterWithLookup:·gc.time                              1000  avgt    5   173.000               ms
LogsBenchmarks.logEmitterWithLookup                                      10000  avgt    5   146.745 ±  5.332   ns/op
LogsBenchmarks.logEmitterWithLookup:·gc.alloc.rate                       10000  avgt    5  1188.098 ± 42.280  MB/sec
LogsBenchmarks.logEmitterWithLookup:·gc.alloc.rate.norm                  10000  avgt    5   192.000 ±  0.001    B/op
LogsBenchmarks.logEmitterWithLookup:·gc.churn.G1_Eden_Space              10000  avgt    5  1190.559 ± 72.655  MB/sec
LogsBenchmarks.logEmitterWithLookup:·gc.churn.G1_Eden_Space.norm         10000  avgt    5   192.390 ±  5.837    B/op
LogsBenchmarks.logEmitterWithLookup:·gc.churn.G1_Old_Gen                 10000  avgt    5     0.002 ±  0.007  MB/sec
LogsBenchmarks.logEmitterWithLookup:·gc.churn.G1_Old_Gen.norm            10000  avgt    5    ≈ 10⁻⁴             B/op
LogsBenchmarks.logEmitterWithLookup:·gc.count                            10000  avgt    5   283.000           counts
LogsBenchmarks.logEmitterWithLookup:·gc.time                             10000  avgt    5   137.000               ms
LogsBenchmarks.singleLogEmitter                                             10  avgt    5   106.824 ±  3.812   ns/op
LogsBenchmarks.singleLogEmitter:·gc.alloc.rate                              10  avgt    5  1632.356 ± 58.141  MB/sec
LogsBenchmarks.singleLogEmitter:·gc.alloc.rate.norm                         10  avgt    5   192.000 ±  0.001    B/op
LogsBenchmarks.singleLogEmitter:·gc.churn.G1_Eden_Space                     10  avgt    5  1636.770 ± 66.696  MB/sec
LogsBenchmarks.singleLogEmitter:·gc.churn.G1_Eden_Space.norm                10  avgt    5   192.520 ±  4.568    B/op
LogsBenchmarks.singleLogEmitter:·gc.churn.G1_Old_Gen                        10  avgt    5     0.004 ±  0.007  MB/sec
LogsBenchmarks.singleLogEmitter:·gc.churn.G1_Old_Gen.norm                   10  avgt    5    ≈ 10⁻³             B/op
LogsBenchmarks.singleLogEmitter:·gc.count                                   10  avgt    5   389.000           counts
LogsBenchmarks.singleLogEmitter:·gc.time                                    10  avgt    5   181.000               ms
LogsBenchmarks.singleLogEmitter                                            100  avgt    5   106.506 ±  1.458   ns/op
LogsBenchmarks.singleLogEmitter:·gc.alloc.rate                             100  avgt    5  1636.967 ± 22.275  MB/sec
LogsBenchmarks.singleLogEmitter:·gc.alloc.rate.norm                        100  avgt    5   192.000 ±  0.001    B/op
LogsBenchmarks.singleLogEmitter:·gc.churn.G1_Eden_Space                    100  avgt    5  1640.944 ±  1.657  MB/sec
LogsBenchmarks.singleLogEmitter:·gc.churn.G1_Eden_Space.norm               100  avgt    5   192.468 ±  2.675    B/op
LogsBenchmarks.singleLogEmitter:·gc.churn.G1_Old_Gen                       100  avgt    5     0.003 ±  0.006  MB/sec
LogsBenchmarks.singleLogEmitter:·gc.churn.G1_Old_Gen.norm                  100  avgt    5    ≈ 10⁻³             B/op
LogsBenchmarks.singleLogEmitter:·gc.count                                  100  avgt    5   390.000           counts
LogsBenchmarks.singleLogEmitter:·gc.time                                   100  avgt    5   184.000               ms
LogsBenchmarks.singleLogEmitter                                           1000  avgt    5   106.881 ±  0.905   ns/op
LogsBenchmarks.singleLogEmitter:·gc.alloc.rate                            1000  avgt    5  1631.019 ± 12.754  MB/sec
LogsBenchmarks.singleLogEmitter:·gc.alloc.rate.norm                       1000  avgt    5   192.000 ±  0.001    B/op
LogsBenchmarks.singleLogEmitter:·gc.churn.G1_Eden_Space                   1000  avgt    5  1636.535 ± 36.295  MB/sec
LogsBenchmarks.singleLogEmitter:·gc.churn.G1_Eden_Space.norm              1000  avgt    5   192.650 ±  4.586    B/op
LogsBenchmarks.singleLogEmitter:·gc.churn.G1_Old_Gen                      1000  avgt    5     0.003 ±  0.004  MB/sec
LogsBenchmarks.singleLogEmitter:·gc.churn.G1_Old_Gen.norm                 1000  avgt    5    ≈ 10⁻³             B/op
LogsBenchmarks.singleLogEmitter:·gc.count                                 1000  avgt    5   389.000           counts
LogsBenchmarks.singleLogEmitter:·gc.time                                  1000  avgt    5   183.000               ms
LogsBenchmarks.singleLogEmitter                                          10000  avgt    5   109.662 ±  2.261   ns/op
LogsBenchmarks.singleLogEmitter:·gc.alloc.rate                           10000  avgt    5  1589.906 ± 32.063  MB/sec
LogsBenchmarks.singleLogEmitter:·gc.alloc.rate.norm                      10000  avgt    5   192.000 ±  0.001    B/op
LogsBenchmarks.singleLogEmitter:·gc.churn.G1_Eden_Space                  10000  avgt    5  1594.619 ± 35.658  MB/sec
LogsBenchmarks.singleLogEmitter:·gc.churn.G1_Eden_Space.norm             10000  avgt    5   192.571 ±  4.089    B/op
LogsBenchmarks.singleLogEmitter:·gc.churn.G1_Old_Gen                     10000  avgt    5     0.003 ±  0.005  MB/sec
LogsBenchmarks.singleLogEmitter:·gc.churn.G1_Old_Gen.norm                10000  avgt    5    ≈ 10⁻³             B/op
LogsBenchmarks.singleLogEmitter:·gc.count                                10000  avgt    5   379.000           counts
LogsBenchmarks.singleLogEmitter:·gc.time                                 10000  avgt    5   182.000               ms
```

There's a lot of information there, but my take aways are:

- Most important metric is probably the time per operation (i.e. `ns/op`). Its tempting to look at memory allocation, but consider that if more operations are occurring, more allocations are happening as a side affect. Its only fair to compare memory allocation if the number of iterations is the same. 
- `logEmitterWithLookup` gets worse as the number of loggers increases: at 10 loggers, we get `110.716 ns/op`, at 10000 we're down to `146.745 ns/op`. 
- `singleLogEmitter` does the best, but not much better than `logEmitterNoLookup`, which as mentioned has a more familiar API. The performance of both of these is fairly constant as number of loggers increases.

In my opinion we should stick with the current API design since its familiar, and we can get good performance fairly easily by just creating new emitters for each log processed (rather than looking up from a map).

tigrannajaryan:
Thanks @jack-berg, this is very useful. @lonewolf3739 based on benchmarks that Jack did the current API design has about the same performance as the alternate approach that we discussed (with LogEmitter not associated with instrumentation library). If the costs are similar I would prefer to keep the current design and avoid changes. Would you be able to verify the performance for Python as well to help us make a decision?

lonewolf3739:
> Thanks @jack-berg, this is very useful. @lonewolf3739 based on benchmarks that Jack did the current API design has about the same performance as the alternate approach that we discussed (with LogEmitter not associated with instrumentation library). If the costs are similar I would prefer to keep the current design and avoid changes. Would you be able to verify the performance for Python as well to help us make a decision?

Sure, I will do that and share the observations.

lonewolf3739:
@tigrannajaryan results are comparatively similar and didn't show significant difference. I also think we can stick to current api and let the sdk authors decide on implementation of achieving the loggerName->InstrumentationLibraryInfo (new emitter/map/something else).

tigrannajaryan:
Update: this was discussed in the spec SIG meeting a couple weeks ago ([11/16/2021](https://docs.google.com/document/d/1pdvPeKjA8v8w_fGKAN68JjWBmVJtPCpqdi9IZrd6eEo/edit#heading=h.izshar9qcc24)). The outcome of that discussion was the following:

1. We will explore if we can tell that instrumentation library name is intended to capture the scope of instrumentation (and library is one of the possible scopes).
2. We will also look at the possibility to rename "instrumentation library" to "instrumentation scope" everywhere (should make sure it is not a breaking change).
3. If the first item above is doable then we can probably declare that Logger Name goes into instrumentation library name, there is no longer a semantic mismatch.
4. This results in inability to record which logging library and which appender was used (was previously thought to be the instrumentation library name). Do we need a separate field / semantic convention to record this?

Since then we explored 1 and found that it is possible to do although it may have a (small) performance impact due to the need to lookup the logger name for every log record and choose/build the appropriate log emitter.

We also [investigated](https://github.com/open-telemetry/opentelemetry-specification/issues/2203) 2 and found that it is not doable, it would be a breaking change.

To summarize the current state:
- Putting Logger Name into Instrumentation Library Name is technically possible but has a small performance impact.
- Instrumentation Library Name concept must remain as is and cannot be renamed. As a consequence putting Logger Name into it creates a semantic mismatch. Loggers are NOT Instrumentation Libraries.

Based on the above I am proposing that we do NOT accept the Spec SIG proposal to put Logger Name into Instrumentation Library Name and instead go ahead with the alternate proposal to have Logger Name as a top level field as suggested in this PR.

@bogdandrutu let me know if you have other arguments on this topic.

bogdandrutu:
> Since then we explored 1 and found that it is possible to do although it may have a (small) performance impact due to the need to lookup the logger name for every log record and choose/build the appropriate log emitter.

Performance in doing what?

> Instrumentation Library Name concept must remain as is and cannot be renamed. As a consequence putting Logger Name into it creates a semantic mismatch. Loggers are NOT Instrumentation Libraries.

As mentioned before the "Instrumentation Library Name" was designed to match exactly the Logger pattern (or almost exactly that pattern) and now the logging SIG decides that is not the way :) I found this ... annoying to say at least, maybe name is bad but still...

cijothomas (approved):

PR #1238, opened 2020-11-18
Move markdownlint and markdown-link-check to github actions

bogdandrutu:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,11 +28,11 @@ install-markdown-link-check:
 markdown-link-check:
 	@for f in $(ALL_DOCS); do $(MARKDOWN_LINK_CHECK) --quiet --config .markdown_link_check_config.json $$f; done
 
-.PHONY: install-markdown-lint
+.PHONY: install-markdownlint
 install-markdown-lint:
 	npm install -g markdownlint-cli
 
-.PHONY: markdown-lint
+.PHONY: markdownlint
 markdown-lint:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -28,11 +28,11 @@ install-markdown-link-check:
...
 	@for f in $(ALL_DOCS); do $(MARKDOWN_LINK_CHECK) --quiet --config .markdown_link_check_config.json $$f; done
 
-.PHONY: install-markdown-lint
+.PHONY: install-markdownlint
 install-markdown-lint:
 	npm install -g markdownlint-cli
 
-.PHONY: markdown-lint
+.PHONY: markdownlint
 markdown-lint:
```
Oberon00, line 36: The PHONY no longer matches the actual target name now.

```suggestion
markdownlint:
```

same with install above

arminru (approved):
I would put both in the same workflow/YAML to reduce maintenance effort when changing triggers or the like.

carlosalberto (approved):

carlosalberto:
Any reason to not include `misspell` as well? Or is that planned in a follow-up? Just curious ;)

bogdandrutu:
Will be a follow-up

Makefile
Oberon00: @bogdandrutu Why did you resolve this without fixing the issue?

PR #1242, opened 2020-11-19
Add java.vm.name to process.runtime.description & more examples.

Oberon00:
Add java.vm.name to process.runtime.description. I think (except for Eclipse J9), the result reads cleaner and is more informative.

Related: https://github.com/open-telemetry/opentelemetry-java/issues/1908#issuecomment-728909456

### DIFF
--- a/specification/resource/semantic_conventions/process.md
+++ b/specification/resource/semantic_conventions/process.md
@@ -52,7 +52,7 @@ At least one of `process.executable.name`, `process.executable.path`, `process.c
 |---|---|---|--|
 | process.runtime.name | The name of the runtime of this process. For compiled native binaries, this SHOULD be the name of the compiler. | `OpenJDK Runtime Environment` | No |
 | process.runtime.version | The version of the runtime of this process, as returned by the runtime without modification. | `14.0.2` | No |
-| process.runtime.description | An additional description about the runtime of the process, for example a specific vendor customization of the runtime environment. | `Eclipse OpenJ9 openj9-0.21.0` | No |
+| process.runtime.description | An additional description about the runtime of the process, for example a specific vendor customization of the runtime environment. | `Eclipse OpenJ9 Eclipse OpenJ9 VM openj9-0.21.0` | No |
 
 `process.runtime.name` SHOULD be set to one of the values listed below, unless more detailed instructions are provided.
 If none of the listed values apply, a custom value best describing the runtime MAY be used.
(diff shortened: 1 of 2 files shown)


specification/resource/semantic_conventions/process.md
```diff
@@ -86,18 +86,21 @@ Java instrumentation should fill in the values by copying from system properties
...
-| Zulu OpenJDK | OpenJDK Runtime Environment | 11.0.8+10-LTS | Azul Systems, Inc Zulu11.41+23-CA |
-| Android 11 | Android Runtime | 0.9 | The Android Project 2.1.0 |
+| OpenJDK | OpenJDK Runtime Environment | 11.0.8+10 | Oracle Corporation OpenJDK 64-Bit Server VM 11.0.8+10 |
+| AdoptOpenJDK Eclipse J9 | OpenJDK Runtime Environment | 11.0.8+10 | Eclipse OpenJ9 Eclipse OpenJ9 VM openj9-0.21.0 |
+| AdoptOpenJDK Hotspot | OpenJDK Runtime Environment | 11.0.8+10 | AdoptOpenJDK OpenJDK 64-Bit Server VM 11.0.8+10 |
+| SapMachine | OpenJDK Runtime Environment | 11.0.8+10-LTS-sapmachine | SAP SE OpenJDK 64-Bit Server VM 11.0.8+10-LTS-sapmachine |
+| Zulu OpenJDK | OpenJDK Runtime Environment | 11.0.8+10-LTS | Azul Systems, Inc OpenJDK 64-Bit Server VM Zulu11.41+23-CA |
+| Oracle Hotspot 8 (32 bit) | Java(TM) SE Runtime Environment | 1.8.0_221-b11 | Oracle Corporation Java HotSpot(TM) Client VM 25.221-b11 |
+| IBM J9 8 | Java(TM) SE Runtime Environment | 8.0.5.25 - pwa6480sr5fp25-20181030_01(SR5 FP25) | IBM Corporation IBM J9 VM 2.9 |
+| Android 11 | Android Runtime | 0.9 | The Android Project Dalvik 2.1.0 |
```
anuraaga, line 103: Thanks for even finding the name for Android :-)

anuraaga (approved):

arminru (approved):
Technically this is a breaking change of existing semantics but I would not expect any back end to actually parse this value but rather just display it as-is. If there are no objections, I'm fine with it.

Please add a changelog to make people aware of it, however.

iNikem (approved):

specification/resource/semantic_conventions/process.md
Oberon00: Caveat: I did not actually try this on on Android, but just googled around. The most official source I found was https://android.googlesource.com/platform/art/+/master/openjdkjvmti/ti_properties.cc but Stackoverflow concurs.

Oberon00:
> Technically this is a breaking change of existing semantics.

Hmm, in the strictest sense it is, but if you just look at the attribute description, it says **Java instrumentation should fill in** and the description for process.runtime still says "An additional description about the runtime of the process, for example a specific vendor customization of the runtime environment".

I think this description is something that we should encourage people not to parse, and we should add runtime-specific attributes if desired e.g. `jvm.prop.java.vm.name` to contain the system property `java.vm.name`.

> Please add a changelog to make people aware of it, however.

This is not yet implemented in opentelemetry-java and I made them aware on the respective issue open-telemetry/opentelemetry-java#1908. A changelog entry still makes sense though, will do.

jmacd (approved):

PR #1249, opened 2020-11-20
[Draft] Add http metrics translated from http metrics semconv table

hstan:
In our project we found the yaml code as configuration for semconv very useful, however, this is missing for metrics. This PR tries to add the HTTP metrics and if this structure is ok we'd like to move on to implement other types.

## Changes
Translated https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/semantic_conventions/http-metrics.md into yaml

### DIFF
--- a/semantic_conventions/metric/http.yaml
+++ b/semantic_conventions/metric/http.yaml
@@ -1,5 +1,5 @@
 groups:
-  - id: http
+  - id: metric.http
     prefix: http
     brief: 'This document defines semantic conventions for HTTP client and server Metrics.'
     note: >
--- a/semantic_conventions/metric/http.yaml
+++ b/semantic_conventions/metric/http.yaml
@@ -6,37 +6,37 @@ groups:
       These conventions can be used for http and https schemes
       and various HTTP versions like 1.1, 2 and SPDY.
     labels:
-      - id: http.method
+      - id: method
         type: string
         example: 'GET'
         brief: 'The HTTP request method'
-      - id: http.host
+      - id: host
         type: string
         example: 'www.example.com'
         brief: 'The value of the HTTP host header. When the header is empty or not present, this label should be the same.'
-      - id: http.scheme
+      - id: scheme
         type: string
         example: 'https'
         brief: 'The URI scheme identifying the used protocol in lowercase'
-      - id: http.status_code
+      - id: status_code
         type: string
         example: '200'
         brief: 'HTTP response status code'
-      - id: http.flavor
+      - id: flavor
         type: string
         example: [ '1.0', '1.1', '2', 'SPDY', 'QUIC' ]
         brief: 'Kind of HTTP protocol used'
     parameterizedLabels: # To avoid high cardinality the following labels SHOULD substitute any parameters when added as labels to http metric events
-      - id: http.url
+      - id: url
         type: string
         example: "api/users/{user_id}"
         brief: 'The originally requested URL'
-      - id: http.target
+      - id: target
         type: string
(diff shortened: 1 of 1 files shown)


jmacd (approved):

semantic_conventions/metric/http.yaml
```diff
@@ -0,0 +1,92 @@
+groups:
+  - id: http
```
thisthat, line 2: This would get a conflict with the Span HTTP semantic convention. 
I would suggest to call it `metric.http`

semantic_conventions/metric/http.yaml
```diff
@@ -0,0 +1,92 @@
...
+    prefix: http
+    brief: 'This document defines semantic conventions for HTTP client and server Metrics.'
+    note: >
+      These conventions can be used for http and https schemes
+      and various HTTP versions like 1.1, 2 and SPDY.
+    labels:
+      - id: http.method
```
thisthat, line 9: The `prefix` tag is to avoid writing `http` for all attributes :) 
Furthermore, this will get out of sync with the Span HTTP attributes. 
I suggest to use `ref: http.method` (`ref` requires the fqn - [syntax](https://github.com/open-telemetry/opentelemetry-specification/blob/master/semantic_conventions/syntax.md#attributes))

semantic_conventions/metric/http.yaml
```diff
@@ -0,0 +1,92 @@
...
+        brief: 'The URI scheme identifying the used protocol in lowercase'
+      - id: http.status_code
+        type: string
+        example: '200'
+        brief: 'HTTP response status code'
+      - id: http.flavor
+        type: string
+        example: [ '1.0', '1.1', '2', 'SPDY', 'QUIC' ]
+        brief: 'Kind of HTTP protocol used'
+    parameterizedLabels: # To avoid high cardinality the following labels SHOULD substitute any parameters when added as labels to http metric events
```
thisthat, line 29: @justinfoote WDYT about this convention?

hstan: @thisthat @justinfoote  This was from https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/semantic_conventions/http-metrics.md#parameterized-labels
I'm not sure if it's worth adding a new section as it is still just a label with the same attributes. It would be great if you could suggest.

semantic_conventions/metric/http.yaml
Oberon00: @hstan It's best if you let the table be generated as markdown somewhere. E.g. in this example, the name of the attribute would have been `http.http.method`.

hstan: @Oberon00 It seems the script for generating table has a whitelist for fields which does not include `labels` (and any new fields) for metrics. Should we use attributes instead or add the new fields to the script?

Oberon00: @hstan The fields in the YAML do not have arbitrary names. When you use fields that the script does not know how to interpret, it will break.

justinfoote: There is definitely some work to do to support metric semantic conventions in that markdown generator. I suggest we leave the yaml out of this PR until that work is done.

semantic_conventions/metric/http.yaml
justinfoote: I'm not sure about whether we should include a top-level `parameterizedLabels` key in the yaml schema. I think I'd prefer that we include this as a generic _note_ or something similar.

PR #1265, opened 2020-11-30
Allow cloud.* ad faas.* resource attributes to be set on span

### LINKED ISSUE #1261: Convider making faas.id a span attribute instead

Currently `faas.id` is a resource attribute. In AWS Lambda functions its value (lambda function ARN) is only available as a property of the AWS lambda context passed to the function, so it's impossible to detect this part just from the environment.

mateuszrzeszutek:
Fixes #1261

## Changes

Add a note to `faas` and `cloud` resource attributes that allows them to be used as span attributes in serverless functions.

### DIFF
--- a/specification/resource/semantic_conventions/cloud.md
+++ b/specification/resource/semantic_conventions/cloud.md
@@ -22,3 +22,7 @@
 | `azure` | Microsoft Azure |
 | `gcp` | Google Cloud Platform |
 <!-- endsemconv -->
+
+Note: When used in serverless function, all `cloud` resource attributes MAY also be used as span attributes.
+Some FaaS APIs do not provide a way to detect those attributes from the environment - they are passed as function
+parameters, available after the TracerProvider is instantiated; that makes it impossible to use them in a Resource.
--- a/specification/resource/semantic_conventions/cloud.md
+++ b/specification/resource/semantic_conventions/cloud.md
@@ -23,6 +23,6 @@
 | `gcp` | Google Cloud Platform |
 <!-- endsemconv -->
 
-Note: When used in serverless function, all `cloud` resource attributes MAY also be used as span attributes.
+When used in serverless function, some or all `cloud` resource attributes MAY also be used as span attributes.
 Some FaaS APIs do not provide a way to detect those attributes from the environment - they are passed as function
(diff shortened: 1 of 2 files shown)


specification/resource/semantic_conventions/cloud.md
```diff
@@ -22,3 +22,7 @@
 | `azure` | Microsoft Azure |
 | `gcp` | Google Cloud Platform |
 <!-- endsemconv -->
+
+Note: When used in serverless function, all `cloud` resource attributes MAY also be used as span attributes.
```
Oberon00, line 26:
```suggestion
When used in serverless function, some or all `cloud` resource attributes MAY also be used as span attributes.
```

specification/resource/semantic_conventions/faas.md
```diff
@@ -16,3 +16,7 @@
 <!-- endsemconv -->
 
 Note: The resource attribute `faas.instance` differs from the span attribute `faas.execution`. For more information see the [Semantic conventions for FaaS spans](../../trace/semantic_conventions/faas.md#difference-between-execution-and-instance).
+
+Note: All `faas` resource attributes MAY also be used as span attributes.
```
Oberon00, line 20:
```suggestion
Some or all `faas` resource attributes MAY also be used as span attributes.
```

Oberon00 (approved):
Nit: Use "Note:" only for explanations, not for specifying/allowing additional things.

mateuszrzeszutek:
@open-telemetry/technical-committee @open-telemetry/specs-approvers Please take a look at this PR when you get a minute.

anuraaga (approved):
We might consider a general statement at the resource level that any attribute that cannot be determined at startup time can be added as span attributes.

Even better might be wording that it's ok for a resource to be mutable or dynamic. I think we assume it's constant, but I think it limits some use cases for attributes (docker cgroup CPU limit?).

But this still seems like a fine step to me.

carlosalberto (approved):

carlosalberto:
Overall good, but I'd like to get more eyes before we merge as I feel this change could be needed in other components too (I will add it to the agenda of the next SIG meeting, in case we haven't resolved this issue/PR by then).

Oberon00 (changes requested):
Blocking merge for now, I think it would be best to discuss the feasibility of #1298 first.

Issue #1237, opened 2020-11-18
Jaeger exporter service.name clarification needed


jkwatson:
**What are you trying to achieve?**

#1222 includes language that requires the Jaeger exporter to use the `service.name` attribute of the Resource, but does not explain how the exporter should behave if that attribute is missing from the Resource.

Should the exporter be configured with a fallback service name, or should it fail to export any spans without a valid Resource `service.name` entry?

arminru:
@yurishkuro is there any fallback in Jaeger if it's not provided?
If so, I'd just not set it in that case and we could fix the spec by adding an `, if present` here:
https://github.com/open-telemetry/opentelemetry-specification/blob/f2869a369ff4b93024e3437837320da76d2c0637/specification/trace/sdk_exporters/jaeger.md#L24-L26

yurishkuro:
Most Jaeger SDKs will refuse to initialize if the service name (a required parameter) is blank, e.g.
   * https://github.com/jaegertracing/jaeger-client-python/blob/4cde94a984e0ef7ba1fbfe670391514398fe20da/jaeger_client/config.py#L109
  * https://github.com/jaegertracing/jaeger-client-go/blob/ca996000929ca5e2c70b9bcd67375836ab410de4/config/config.go#L219

I would expect OTEL in general to guarantee that there is always a resource with non-empty service name. In other words, I don't think this should be solved at the level of Jaeger exporters or any Jaeger-specific functionality.

Whether the SDK should raise an exception or default to some `unknown_service` constant is up for debate.

To unblock Jaeger exporter work specifically, I would suggest going with `unknown_service`.

jkwatson:
Currently, from what I understand, there is no requirement that `service.*` is present in the Resource. What is required is that if anything `service.*` is provided, that `service.name` is included. [although, I don't know if any SDKs have implemented this particular validation].

The java implementation requires a "service name" to be configured into the Jaeger exporter, but it's not clear how that should interact with the `service.name` on the Resource (I have guesses about how it should, but it's not specified here).

yurishkuro:
> The java implementation requires a "service name" to be configured into the Jaeger exporter

That sounds like a bad take. Jaeger backend uses OTEL collector components and can accept OTLP data directly, so the service name must be present in the native OTEL data.

jkwatson:
> > The java implementation requires a "service name" to be configured into the Jaeger exporter
> 
[...]

Issue #1034, opened 2020-09-29
Guidance for filling (or consider removing?) service.instance.id


anuraaga:
Resource defines `service.instance.id` as a unique ID among horizontally scaled services

https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/resource/semantic_conventions#service

But we have many similar IDs in other places
`host.id`
`k8s.pod.name` or `uid`
`container.id`
`faas.instance`

Should one of these be copied into `service.instance.id` when it's available? One issue is ordering, `host.id` is the instance ID for non-container workloads but not container workloads, for example.

Or can we just remove `service.instance.id` since we would expect an ID of horizontal scaling to be present in another convention?

Oberon00:
A host can host multiple services. A POD might host multiple (related) services. A container might host multiple services (even though that's a unusual). I can't think of a scenario wher FaaS would have more than one service instance.

I agree though that *most of the time*, except for host, everything should map 1:1 to service.instance.id, so that would be reasonable choices.

However, since only the triplet service.(namespace, name, id) is supposed to be unique, I think that we should not auto-fill service.instance.id by default with anything -- it seems to be predestined for an "*inherent* unique ID" as the spec puts it.

I think it may make a lot of sense to add a `telemetry.instance_id` to be able to group resources that came from the very same initialization of OpenTelemetry, i.e. something that is usable as primary key for resources. Note that the service namespace/name/id triplet is not usable for that as it stays (or should stay at least) the same over restarts of the same service instance.

anuraaga:
> is not usable for that as it stays (or should stay at least) the same over restarts of the same service instance.

Hmm - in that case should we consider removing the recommendation to "generate a random Version 1 or Version 4 RFC 4122 UUID"? I don't think it can ever be possible for the UUID to be the same over restarts. I guess a user needs to set it appropriately in that case.

Oberon00:
Spec says:

> It is preferable for the ID to be persistent and stay the same for the lifetime of the service instance, however it is acceptable that the ID is ephemeral and changes during important lifetime events for the service (e.g. service restarts).

And also later:

> services aiming for reproducible UUIDs may also use Version 5, see RFC 4122 for more recommendations

It would be cool if the spec could also write the "nicknames" for these versions, e.g. V 5 requires a name and namespace as input, e.g. Python's implementation has this:

> `uuid.uuid5(namespace, name)`
[...]

PR #1269, opened 2020-12-01
Define the fallback case for service.name

### LINKED ISSUE #1241: Service name should be mandatory for all configured resources.

The resource specifications currently allow the entire `service.*` namespace to be missing from a Resource configured into the SDK. This means that exporters that require a `service.name` must be configured (or hard-coded) with a fallback name. Since all services should have a `service.name`, we should simply require that to be set on any resource that is attached to an SDK.

**Additional context.**

This came up via the following spec clarification request for how the Jaeger exporter should behave in the absence of a `service.name` in the Resource: https://github.com/open-telemetry/opentelemetry-specification/issues/1237

carlosalberto:
Fixes #1241, #1237 

This is a very minimalistic clarification based on what currently many SIGs do (providing a fallback based on the running service), finally defaulting to the `unknown_service` literal.


yurishkuro (approved):

specification/resource/semantic_conventions/README.md
```diff
@@ -47,6 +47,9 @@ Certain attribute groups in this document have a **Required** column. For these
 | `service.version` | string | The version string of the service API or implementation. | `2.0.0` | No |
 
 **[1]:** MUST be the same for all instances of horizontally scaled services.
+Although this is a required attribute, if the value could not be specified SDKs
+MUST fallback to a language-specific name based on the running service.
+If this fallback fails, the name MUST default to `unknown_service`.
```
arminru, line 52: The part between those two semconv markers is generated, you'll have to add to the note in `semantic_conventions/resource/service.yaml` and then run the generator as described in `semantic_conventions/README.md`. The build is failing since the current markdown and the one that would be generated from the YAML don't match.

arminru, line 52: Should we use a name that makes it more obvious this was not set by user code or an instrumentation but by the OpenTelemetry SDK itself?

specification/resource/semantic_conventions/README.md
```diff
@@ -47,6 +47,9 @@ Certain attribute groups in this document have a **Required** column. For these
 | `service.version` | string | The version string of the service API or implementation. | `2.0.0` | No |
 
 **[1]:** MUST be the same for all instances of horizontally scaled services.
+Although this is a required attribute, if the value could not be specified SDKs
```
arminru, line 50:
```suggestion
Although this is a required attribute, if the value was not specified, SDKs
```

specification/resource/semantic_conventions/README.md
```diff
@@ -47,6 +47,9 @@ Certain attribute groups in this document have a **Required** column. For these
...
+MUST fallback to a language-specific name based on the running service.
```
arminru, line 51: Can we provide some guidance on how this could or should be done?

Oberon00 (commented):
In my opinion, this is the wrong place to specify this. I mean, it's fine to additionally specify it here, but we need a list of required resources in the SDK specification. Maybe https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/resource/sdk.md, although in principle it applies to tracing, metrics and probably logging SDK.
I think we should not assume that SDK implementers will read all resource semantic conventions.

specification/resource/semantic_conventions/README.md
carlosalberto: Open to suggestions :)

carlosalberto:
> I think we should not assume that SDK implementers will read all resource semantic conventions.

Fair point. I think this documentation *needs* to be here, but we also need to mention this in the main `sdk.md` document, with a link to the specific case of `service.name`, as this is a rather important item.

Oberon00:
We probably also want that for the telemetry.sdk resources BTW.

specification/resource/semantic_conventions/README.md
carlosalberto: This is actually language specific - but I added a mention of typically used components, such as process name, language or platform where the service is running.

specification/resource/sdk.md
```diff
@@ -34,6 +34,10 @@ The interface MUST provide a way to create a new resource, from [`Attributes`](.
 Examples include a factory method or a constructor for a resource
 object. A factory method is recommended to enable support for cached objects.
 
+Note that certain **required** `Resource` specific attributes
+MUST have fallback values if they were not specified, e.g. `service.name`.
+See [Resource standard attributes]((semantic_conventions/README.md)).
```
Oberon00, line 39:
```suggestion
See [Resource standard attributes](semantic_conventions/README.md).
```

Oberon00, line 39: But there is no such "Resource standard attributes" section in there. I don't think we can require all the Required attributes in that markdown file, since there are many (linked) there.

carlosalberto: Sure, that's why I used "certain" (agree that not all Required attributes should have fallback values).

specification/resource/semantic_conventions/README.md
jmacd: How about`$(telemetry.sdk.name)_service` or `openetemetry_service`?

jmacd: After reading the other comments, I think that it would be best if we didn't offer a vague recommendation and then a fallback. A simpler rule will be: when the user does not set a required attribute, a determined value will be used (e.g., the two suggestions above). There are probably other semantic conventions for the attributes that might be used to infer "language-specific name based on the running service (typically inferred from process name, language, plaform, etc)."

Maybe just remove the "language-specific name" part and jump to a MUST and one of the suggestions above?

andrewhsu:
@open-telemetry/specs-approvers should this really be `P1` since the linked issues #1241, #1237 are also `P1`?

jmacd:
Based on this comment, https://github.com/open-telemetry/opentelemetry-specification/issues/1241#issuecomment-735972286, I'm inclined to promote my own suggestion above:

https://github.com/open-telemetry/opentelemetry-specification/pull/1269#discussion_r534552116

Explicitly, it means that we can remove some of the words in this PR and simply specify one of the two options in my comment.

austinlparker:
Just as a note, I think whatever the decision is here needs to be something that is obvious to an operator/consumer of the telemetry and something that is non-suitable for actual usage (such as a 'smart' derived name based on command line or whatnot). Ultimately the name of a service is a logical decision made by humans who are interpreting data, but if we're too clever about it, some users may defer to the cleverness which is good up until the moment it isn't. An obvious default that isn't suitable for long-term usage satisfies both of these conditions.

jmacd:
The semantic conventions in https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/resource/semantic_conventions/process.md already contain the command name, and the telemetry SDK details are built in (https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/resource/semantic_conventions/README.md), so based on @austinlparker's comment I think when `service.name` is not set exporters MUST use `opentelemetry_service` as a fallback.

carlosalberto:
After a few iterations myself, I think we either go with:

1) Something **standard** based on the semantic conventions (as @jmacd proposed), such as `process.executable.name` or `process.command` - with a fallback to `opentelemetry_service` or `unknown_service`, **OR**
2) Fallback directly to `opentelemetry_service` or `unknown_service`.

Opinions? cc @yurishkuro @Oberon00

yurishkuro:
I prefer "unknown" to be in the name. I don't have a strong opinion on auto-resolvers, but I share Austin's concern with them not doing the right thing and being worse than the default.

jmacd:
I support "unknown"

carlosalberto:
Ping @Oberon00 (as I'm also slightly inclined to use `unknown_service` without the fallback, for the operators interaction that Austin mentioned).

Oberon00:
"unknown" is fine. Or "unknown:EXE-NAME".

specification/resource/sdk.md
arminru: I would list the attributes required to be set by the SDK (currently only `service.name` and `telemetry.sdk.*`, right?) explicitly in `sdk.md`. This way, SDK authors/maintainers only have to refer to the SDK spec and everything that's relevant for the SDK is versioned therein rather than having to check the semantic convention docs separately for changes.

Oberon00: In any case, if you say "certain" you must specify what these "certain" are. Right now, I would not know.

carlosalberto: Sounds fair, will add a note.

austinlparker:
+1 for 'unknown_service'

carlosalberto:
Updated with:

1) `unknown_service:{process.executable.name}`, e.g. `unknown_service:bash`
2) `unknown_service` if `process.executable.name` is not available.

Let me know whether this looks acceptable, and whether this fallback case is sane ;)

jmacd (approved):

carlosalberto:
Ping @Oberon00 @austinlparker

austinlparker:
lgtm

austinlparker (approved):

specification/resource/semantic_conventions/README.md
```diff
@@ -32,6 +33,12 @@ Attributes are grouped logically by the type of the concept that they described.
 
 Certain attribute groups in this document have a **Required** column. For these groups if any attribute from the particular group is present in the Resource then all attributes that are marked as Required MUST be also present in the Resource. However it is also valid if the entire attribute group is omitted (i.e. none of the attributes from the particular group are present even though some of them are marked as Required in this document).
 
+## Attributes with Default Value
+
+The SDK MUST set a default value for the following attributes if they were not specified:
```
Oberon00, line 38:
```suggestion
The SDK MUST set a default value for the following attributes if they were not specified, even if the resource would otherwise not be present at all:
```

Oberon00 (changes requested):

specification/resource/semantic_conventions/README.md
```diff
@@ -46,7 +53,7 @@ Certain attribute groups in this document have a **Required** column. For these
 | `service.instance.id` | string | The string ID of the service instance. [3] | `627cc493-f310-47de-96bd-71410b7dec09` | Yes |
```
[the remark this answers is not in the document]
carlosalberto: I'd suggest going ahead with this PR (as it has been standing for long time now) and follow up with #1034 right after (Priority 2) so we can either set it as well or remote it altogether.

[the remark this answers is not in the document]
Oberon00: If we do that, we need to raise #1034 to P1 required-for-ga. If you think this is OK, I'm fine with merging as the PR is now. Note that doing it the other way round (resolving  #1034 first) would not result in the spec temporarily being in a contradictory state though.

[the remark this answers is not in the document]
Oberon00: FYI @anuraaga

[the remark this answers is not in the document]
Oberon00: A "quickfix" would be to make the instance-ID non-required.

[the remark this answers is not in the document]
jmacd: I am in favor of marking the instance-ID as non-required to let us merge this PR. Thanks for pointing out the inconsistency in #1034.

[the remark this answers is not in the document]
carlosalberto: +1 for making instance-ID non-required (after all, there's the related issue). Will update, thanks!

carlosalberto:
> I think there is awareness now for the problem with #1034, so not blocking this PR (trusting you to sort this out properly)

Thanks! Have made `service.instance.id` non-required, so this can be properly addressed in #1034

PR #1272, opened 2020-12-02
Remove opaque when referring to baggage metadata

anuraaga:
## Changes

I don't know if this is correct, but something probably is. In Java, we currently expose the String for a baggage metadata

https://github.com/open-telemetry/opentelemetry-java/blob/master/api/src/main/java/io/opentelemetry/api/baggage/EntryMetadata.java#L40

My understanding of opaque is to not expose internal details, for example the String. So I'd generally not use opaque for this - does it make sense without the word opaque? I guess the key point here is we use a wrapper, not a String directly, to allow for more functionality in the future.

### DIFF
--- a/specification/baggage/api.md
+++ b/specification/baggage/api.md
@@ -85,8 +85,8 @@ REQUIRED parameters:
 
 OPTIONAL parameters:
 
-`Metadata` Optional metadata associated with the name-value pair. This should be an opaque wrapper
-for a string with no semantic meaning. Left opaque to allow for future functionality.
+`Metadata` Optional metadata associated with the name-value pair. This should be a wrapper
+for a string with no semantic meaning. We use a wrapper to allow for future functionality.
 
 `Context` The context containing the `Baggage` in which to set the baggage entry.
 


specification/baggage/api.md
```diff
@@ -85,8 +85,8 @@ REQUIRED parameters:
 
 OPTIONAL parameters:
 
-`Metadata` Optional metadata associated with the name-value pair. This should be an opaque wrapper
-for a string with no semantic meaning. Left opaque to allow for future functionality.
+`Metadata` Optional metadata associated with the name-value pair. This should be a wrapper
+for a string with no semantic meaning. We use a wrapper to allow for future functionality.
```
arminru, line 89: I think the wording _opaque_ was introduced to say that the content and its semantics are not defined or known, and that the metadata is only passed through as it was received. Maybe we can find some clearer wording here?
cc @jkwatson

anuraaga: Well, an alternative is to just use the `String` directly as-is, but I guess we're not supposed to do that. I think this sentence is explaining why we have a wrapper - I feel like the literal reason is to allow adding functionality in the future, not really related to opaqueness.

That being said, perhaps the point is about making sure the downstream baggage propagator gets the same value in as the upstream one did. In this case, perhaps we want to add a point "The wrapper must expose the original string value of the metadata as-is for use in propagation."?

jkwatson: @arminru's take is correct. And, I agree..clearer language would be good. Baggage has no intrinsic meaning right now, but it probably will in the future. In order to account for future meaning, we want to have a backwards-compatible handle to attach that meaning to. And, as you point out, we also need to propagate it without knowing what meaning users might attach themselves.

I think your phraseology seems fine to me.

yurishkuro: Another option is to remove the notion of metadata altogether and only add it when W3C clearly defines the semantics/syntax. I don't think any non-W3C formats allow metadata anyway.

anuraaga: @yurishkuro If that is an option, I'd definitely love to take it while we still can pre-GA as metadata has been extremely confusing to me so far. Any way to make it happen?

tedsuo:
@anuraaga FYI please use a fork rather than a branch on the main repo. Do you mind deleting this branch?

anuraaga:
Hey sorry about that - I didn't know I have push access to this repo :O.

PR #1277, opened 2020-12-05
vendors: add details about vendors offering instrumented services

tsloughter:
## Changes

This patch expands the spec related to vendors to include vendors providing non-observability services, but instead provde services or applications that are instrumented, like a proxy or a database.

This change was motivated by the AWS announcments and distros which seem to still require their custom trace propagators: https://aws.amazon.com/blogs/opensource/distributed-tracing-with-opentelemetry/

### DIFF
--- a/specification/vendors.md
+++ b/specification/vendors.md
@@ -27,16 +27,30 @@ The goal is for users to be able to easily switch between vendors while also
 ensuring that any language with an OpenTelemetry SDK implementation is able to
 work with any vendor who claims support for OpenTelemetry.
 
+Vendor means both services and applications which provide observability related
+functionality, like accepting and analyzing traces, as well as services and
+applications that are instrumented and propagate context, like an HTTP load
+balancer.
+
 This document will explain what is required of a vendor to be considered to
 "Support OpenTelemetry" or "Implements OpenTelemetry".
 
 ## Supports OpenTelemetry
+
+### Observability Services and Applications
 
 "Supports OpenTelemetry" means the vendor must accept the output of the default
 SDK through one of two mechanisms:
 
 - By providing an exporter for the [OpenTelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector/) and / or the OpenTelemetry SDKs
 - By building a receiver for the [OpenTelemetry protocol](https://github.com/open-telemetry/opentelemetry-proto)
+
+### Instrumented Services and Applications
+
+A service or application can be said to "Support OpenTelemetry" if its
+instrumentation supports the W3C Trace Context propagation. The propagation used
+can be configurable, but when text based propagation is used then at least the
+W3C Trace Context must be offered.
 
 ## Implements OpenTelemetry
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ Updates:
   ([#1219](https://github.com/open-telemetry/opentelemetry-specification/pull/1219))
 - Add RPC semantic conventions for metrics
   ([#1162](https://github.com/open-telemetry/opentelemetry-specification/pull/1162))
+- Clarify what it will mean for a vendor of instrumented services to "support
+  OpenTelemetry" ([#1277](https://github.com/open-telemetry/opentelemetry-specification/pull/1277)).
 
 ## v0.7.0 (11-18-2020)
 
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ Technical committee holds regular meetings, notes are held
   - [Acronym](#acronym)
   - [Contributions](#contributions)
   - [License](#license)
+- [Vendor Guidelines](specification/vendors.md)
 
 ## Project Timeline
 


yurishkuro (changes requested):
GC had a discussion this week on this topic. There are many nuances that need to be worked out into a coherent policy. Until then I don't this we should be making rush changes. And I think the compliance requirements must be clearly rationalized.

tsloughter:
Ok. Good to know it is already being discussed and will be addressed. I'd argue it'd be good to get a clarification like this one in now, before working out all the nuances and a policy that covers all cases. The spec already defines the need to support the default SDK to claim to "support OpenTelemetry", but it only made explicit reference to exporting while forgetting propagation.

Let me know if I can help with developing the full policy and whether, assuming the decision is to stick to waiting until that fully policy is complete to change anything, if I should close this PR.

yurishkuro:
@tedsuo can provide more details

yurishkuro:
> The spec already defines the need to support the default SDK to claim to "support OpenTelemetry"

I think that is the challenge, "support OpenTelemetry" is very vague and likely has many different levels.

anuraaga:
While I suspect this may have been discussed at the GC, can't find notes. Let me add some insight here in case it helps the discussion.

First, want to confirm @tsloughter, have you seen any messaging that seems to indicate that AWS managed services, for example ELB, "support OpenTelemetry"? I don't think I've seen any, I agree that this would be misleading if it's mentioned anywhere - even before the consideration of w3c propagation, none of them can export in OTLP format anyways. In my blog post, I say the opposite, that "OpenTelemetry fully supports the AWS Tracing Header", which seems correct given we have artifacts supporting propagation using it, as well as many other formats.

But a broader statement of "AWS supports OpenTelemetry" seems fair - AWS is dedicated to providing support for our users in their journey through OpenTelemetry, no need to use x-ray even. So I suspect that the current document is already reasonable - perhaps it could be renamed to `tracing-vendors.md` to make clear it applies to tracing vendors. A different doc is probably needed for `cloud-vendors.md` and I suspect it'll have vastly different content, including distinction of overall support vs support by specific managed services.

tsloughter:
@anuraaga "OpenTelemetry fully supports the AWS Tracing Header" is what stuck out to me. I don't think this PR applies to the AWS posts, but it was that wording that got me thinking about the potential issue and lack of propagation being covered in this spec.

My hope would be that vendors will want to say they "support OpenTelemetry" so will work on using the W3C Trace Context if it is a requirement in this spec.

The "fully supports the AWS Tracing Header" is really more of a separate concern which is whether a pluggable piece of OpenTelemetry can be said to "support" something simply because it can be made to do "anything".

I would argue it can't be said otel "supports" any individual tracing header aside from W3C and B3 because those are the only ones required in the API. What it does support is implementing and plugging in alternative third party propagation headers.

tsloughter:
I also now notice the X-Ray TraceId has a specific format itself, requiring a custom generator. I don't know if that is required to send traces to X-Ray or just means better support/sorting if they are done using the custom generator?

I think a "distribution" spec is needed, in addition to this doc on what vendor support/implementation means -- possibly being split into separate "tracing-vendor" and "cloud-vendor"  docs ("cloud-vendor" doesn't really cover it since this should apply to projects like the many proxies and service meshes as well).

anuraaga:
> I would argue it can't be said otel "supports" any individual tracing header aside from W3C and B3 because those are the only ones required in the API. What it does support is implementing and plugging in alternative third party propagation headers.

I see - yeah admittedly I had Java strongly in my mind when writing that blog post but it's not really an OpenTelemetry concern. Perhaps the better wording would have been OpenTelemetry supports propagating context to managed AWS services without mentioning any header format. Since this doc is about "X supports OpenTelemetry", if adding guidelines for "OpenTelemetry supports X", perhaps that is yet another doc :)

> I also now notice the X-Ray TraceId has a specific format itself, requiring a custom generator. I don't know if that is required to send traces to X-Ray or just means better support/sorting if they are done using the custom generator?

It's required - X-Ray very much does not support / implement OpenTelemetry as per this spec, and if we ever stated that in any material, that would be an error for sure. In general, we'd consider it "possible for a user to use OpenTelemetry with X-Ray" and where we can, we may provide a distro or wrapper library to "provide an out-of-the-box experience with OpenTelemetry and X-Ray" as well as other tracing providers. I'm providing these quotes just as examples that might drive the spec discussions.

Agree in general though that an instrumentation or distro specific doc can be useful. For example, the Java distro enables tracecontext, aws, and b3 propagators by [default](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsAgentProperties.java#L25). So it confirms to the spirit of what you propose here, but that was only because I thought it's natural to support tracecontext, but it may be worth making that extra clear in the spec.

tsloughter:
> it's not really an OpenTelemetry concern

It is. I think the propagation and export format being defined is one of the most important parts of OpenTelemetry and how it differs from OpenTracing.

Working in Erlang has meant no vendors provide libraries, everything has to be implemented and maintained by the community, then discovered and configured by our users, so we are extra sensitive to these concerns :)

specification/vendors.md
```diff
@@ -27,17 +27,31 @@ The goal is for users to be able to easily switch between vendors while also
...
+functionality, like accepting and analyzing traces, as well as services and
+applications that are instrumented and propagate context, like an HTTP load
+balancer.
+
 This document will explain what is required of a vendor to be considered to
 "Support OpenTelemetry" or "Implements OpenTelemetry".
 
 ## Supports OpenTelemetry
 
+### Observability Services and Applications
```
jmacd, line 40: Not a problem per se, but all the examples and requirements here relate to tracing.

We've recently identified that in Metrics and the current OTLP, there are "Supports OpenTelemetry" statements that we should consider perfectly valid but may be additionally described by the subset of aggregation temporalities they accept. A system can be said to support OpenTelemetry without supporting DELTA points, for example, but that's why we've specified built-in DELTA-to-CUMULATIVE conversion by default.

PR #1283, opened 2020-12-10
Standardize TraceId/SpandId/TraceFlags field casing in log formats

tigrannajaryan:
OTLP/JSON is currently using trace_id, span_id field naming. The snake case
with underscores comes from Protobuf/JSON mapping rules.

Elsewhere we were previously using traceid, spanid, traceflags without underscores.

This changes standardizes field names to use underscores for all text log mediums
where these fields may be recorded.

Note: the alternate approach, i.e. keeping traceid, spanid, traceflags and renaming
them in OTLP/JSON would result in OTLP/JSON format having an inconsistent naming,
since all other fields use snake casing.

### DIFF
--- a/specification/logs/overview.md
+++ b/specification/logs/overview.md
@@ -417,9 +417,9 @@ so that they include the Request Context information in the emitted logs.
 trace context should be recorded in logs. To summarize, the following field
 names should be used in legacy formats:
 
-- "traceid" for [TraceId](data-model.md#field-traceid), hex-encoded.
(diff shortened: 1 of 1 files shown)


yurishkuro (approved):
+1

zenmoto (approved):

jmacd (approved):

anuraaga (approved):

kumoroku (approved):

Issue #1267, opened 2020-11-30
Define versioning and support for OpenTelemetry


tedsuo:
The trace specification frozen, we are rapidly approaching the stable release of tracing for OpenTelemetry. Due to the project's size and scope, versioning and support are tricky issues. We need to define these concepts in concrete terms before we can declare tracing to be stable.

* How do we define version numbers?
* How do we define releases?
* How do we define API, SDK, conventions, and contrib support?
* How do we define ongoing experimental and beta work after some packages become stable?

A WIP proposal is open for comment here: https://docs.google.com/document/d/1QHlPsMqwrBm7-IIPef9czuM8KFiLbEDMzGhPrmHcppA/edit#

Alternate proposals are also acceptable, but we are trying to move quickly. By Friday, this WIP will be submitted as an OTEP, with the goal of finding approval by Friday, December 11th. Please contribute this week! Package management and backwards compatibility can have language-specific requirements, so we need maintainers to review this and determine that this proposal it will work for their implementation.

Thank you,
@tedsuo

tedsuo:
OTEP has been released here: https://github.com/open-telemetry/oteps/pull/143

Request for SIG maintainers: please create a language-specific document which describes how versioning and support works for your language, based on this OTEP, and post a link to it here. We want to ensure that versioning and support as been thought through before it is added to the spec.

tedsuo:
Cat herding champions, who will work with maintainers to create language specific versioning documents, based on the maintainer's meeting on Monday:

Java - JWatson
JS - Daniel 
Go - Tyler, JBD & Alolita
Python - Alolita
Erlang - Tristan
C# .NET - Alolita & Carlos
Ruby - Matt W. & Daniel
Rust - Julian & Daniel
PHP - Bob & Daniel
C++ - Alolita & Lalit
Collector - Tigran

tedsuo:
Thank you everyone for your participation! We've unlocked v1.0. Since work has successfully moved to the SIGs and implementations; I am closing this issue.

PR #1291, opened 2020-12-14
Versioning and Stability for OpenTelemetry

tedsuo:
## Changes

* Adds new versioning procedures and stability guarantees for the OpenTelemetry Clients.
* Expands the overview to include a high level description of the  OpenTelemetry client architecture.
* Expands the glossary and adds clarifications.

#### Related issues
* https://github.com/open-telemetry/opentelemetry-specification/issues/1267

#### Related OTEPS
* Implements [OTEP 0143](https://github.com/open-telemetry/oteps/pull/143)


specification/glossary.md
```diff
@@ -26,8 +32,30 @@ Some other fundamental terms are documented in the [overview document](overview.
...
+
+The maintainer of a shared library which is depended upon by many applications, and targeted by OpenTelemetry instrumentation.
+
+### Instrumentation Author
+
+The maintainer of OpenTelemetry instrumentation written against the OpenTelemetry API. This may be instrumentation written within application code, within a shared library, or within an instrumentation library.
+
+### Plugin Author
+
+The maintainer of and OpenTelemetry Plugin, written against OpenTelemetry SDK plugin interfaces.
```
anuraaga, line 51: Can you define plugin as well? I don't think we have it - without knowing what a plugin is, don't know who a plugin author is :)

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,181 @@
...
+Package **version numbers** MAY include a suffix, such as -alpha, -beta, -rc, or -experimental, to differentiate stable and experimental packages.
+
+### Stability
+
+Once a signal component is marked as stable, the following rules MUST apply until the end of that signal’s existence.
+
+**API Stability -**
+Backward-incompatible changes to API packages MUST NOT be made unless the major version number is incremented. All existing API calls MUST continue to compile and function against all future minor versions of the same major version.
+
+ABI compatibility for the API is RECOMMENDED, and MAY be offered on a language by language basis.
```
anuraaga, line 73: Is a recommendation the same as a simpler `Languages SHOULD offer ABI compatibility for the API.`?

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,181 @@
...
+
+**SDK Stability -**
+Public portions of SDK packages MUST remain backwards compatible. There are two categories of public features: **plugin interfaces** and **constructors**. Examples of plugin include the SpanProcessor, Exporter, and Sampler interfaces. Examples of constructors include configuration objects, environment variables, and SDK builders.
+
+ABI compatibility for public portions of the SDK are RECOMMENDED, and MAY be offered on a language by language basis.
+
+**Semantic Conventions Stability -**
+Semantic Conventions MUST NOT be removed once they are stable. New conventions MAY be added to replace usage of older conventions, but the older conventions MUST NOT be removed. Older conventions MUST be marked as deprecated when they are replaces by newer conventions.
```
anuraaga, line 81:
```suggestion
Semantic Conventions MUST NOT be removed once they are stable. New conventions MAY be added to replace usage of older conventions, but the older conventions MUST NOT be removed. Older conventions MUST be marked as deprecated when they are replaced by newer conventions.
```

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,181 @@
...
+
+ABI compatibility for public portions of contrib packages is RECOMMENDED, and MAY be offered on a language by language basis.
+
+Telemetry produced by contrib instrumentation MUST remain stable and backwards compatible, to avoid breaking alerts and dashboard. Existing telemetry MUST NOT be mutated or removed without a major version bump. Additional telemetry MAY be added. This includes additional spans, metrics, resources, attributes, events, and any other data types that OpenTelemetry emits.
+
+### Deprecation
+
+Signals MAY eventually be replaced. When this happens, they are marked as deprecated.
+
+Signals SHALL only marked as deprecated when the replacement becomes stable. Deprecated code MUST abide by the same support guarantees as stable code.
```
anuraaga, line 96:
```suggestion
Signals SHALL only be marked as deprecated when the replacement becomes stable. Deprecated code MUST abide by the same support guarantees as stable code.
```

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,181 @@
...
+
+API stability, as defined above, MUST be maintained.
+
+A version of the SDK which supports the latest minor version of the last major version of the API will continue to be maintained during LTS. Bug and security fixes MUST be backported. Additional feature development is NOT RECOMMENDED.
+
+Contrib packages available when the API is versioned MUST continue to be maintained for the duration of LTS. Bug and security fixes will be backported. Additional feature development is NOT RECOMMENDED.
+
+### SDK Support
+
+SDK stability, as defined above, will be maintained for a minimum of **one year** after after the release of the next major SDK version.
```
anuraaga, line 170:
```suggestion
SDK stability, as defined above, will be maintained for a minimum of **one year** after the release of the next major SDK version.
```

anuraaga: I'm not sure, but is there a situation where old versions of the API only work with old versions of the SDK?

specification/versioning-and-stability.md
jsuereth, line 73: Would it be better to caveat this that "Languages which ship binary artifacts SHOULD offer ABI compatibility" or even MUST?

If we take the notion that "API compatible" = source compiles and "ABI compatible" = binaries link, then I think the nuance here is more around the distribution mechanism.   No binary distribution => ABI doesn't exist.

specification/versioning-and-stability.md
jsuereth, line 81: nit: spelling:  s/replaces/replaced/

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,181 @@
...
+
+**Semantic Conventions Stability -**
+Semantic Conventions MUST NOT be removed once they are stable. New conventions MAY be added to replace usage of older conventions, but the older conventions MUST NOT be removed. Older conventions MUST be marked as deprecated when they are replaces by newer conventions.
+
+**Contrib Stability -**
+Plugins and instrumentation MUST be kept up to date, and MUST released simultaneously (or shortly after) the latest release of the API. The goal is to ensure users can update to the latest version of OpenTelemetry, and not be held back by the plugins that they depend on.
+
+Public portions of contrib packages (constructors, configuration, interfaces) MUST remain backwards compatible.
+
+ABI compatibility for public portions of contrib packages is RECOMMENDED, and MAY be offered on a language by language basis.
```
jsuereth, line 88: Nit: Could we expand this to encompass the difficulty of a plugin's ecosystem vs. the language itself?

E.g. a prometheus plugin can only offer the guaranteees allowed by the prometheus downstream dependencies it relies on.

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,181 @@
...
+  * [Contrib Support](#contrib-support)
+- [OpenTelemetry GA](#opentelemetry-ga)
+
+<!-- tocstop -->
+
+This document defines the stability guarantees offered by the OpenTelemetry clients, along with the rules and procedures for meeting those guarantees.
+
+In this document, the term OpenTelemetry specifically refers to the OpenTelemetry clients. It does not refer to the specification or the Collector.
+
+Language implementations MUST take this cross-language proposal, and produce a language-specific proposal which details how these requirements will be met. This proposal shall be placed in the root of each repo and named `VERSIONING.MD`.
```
carlosalberto, line 29: Do we still need this paragraph? If yes, maybe it will help to think of this as a mandatory section, more than a proposal.

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,181 @@
...
+Semantic Conventions MUST NOT be removed once they are stable. New conventions MAY be added to replace usage of older conventions, but the older conventions MUST NOT be removed. Older conventions MUST be marked as deprecated when they are replaces by newer conventions.
+
+**Contrib Stability -**
+Plugins and instrumentation MUST be kept up to date, and MUST released simultaneously (or shortly after) the latest release of the API. The goal is to ensure users can update to the latest version of OpenTelemetry, and not be held back by the plugins that they depend on.
+
+Public portions of contrib packages (constructors, configuration, interfaces) MUST remain backwards compatible.
+
+ABI compatibility for public portions of contrib packages is RECOMMENDED, and MAY be offered on a language by language basis.
+
+Telemetry produced by contrib instrumentation MUST remain stable and backwards compatible, to avoid breaking alerts and dashboard. Existing telemetry MUST NOT be mutated or removed without a major version bump. Additional telemetry MAY be added. This includes additional spans, metrics, resources, attributes, events, and any other data types that OpenTelemetry emits.
```
jsuereth, line 90: Nit:  I think it's worth leaving a placeholder after this (which should be discussed outside this PR) about what to do when a contrib plugin has an instable release and we want to support BOTH versions of that ecosystem.   Hopefully this never happens, but....  It'd be nice to know what to do if, e.g. Prometheus decided to have some kind of major breaking change that made us unable to support our previous version in a stable/compatible way with a new version.

Note: I'm picking on prometheus just because it's the biggest non-OTel plugin I can think of, not because I expect this to happen.

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,181 @@
...
+
+Contrib packages available when the API is versioned MUST continue to be maintained for the duration of LTS. Bug and security fixes will be backported. Additional feature development is NOT RECOMMENDED.
+
+### SDK Support
+
+SDK stability, as defined above, will be maintained for a minimum of **one year** after after the release of the next major SDK version.
+
+### Contrib Support
+
+Contrib stability, as defined above, will be maintained for a minimum of **one year** after after the release of the next major version of a contrib package.
```
jsuereth, line 174: What is the plan to support drive-by-contributions to contrib?    E.g. let's say someone comes with a really meaty and great plugin for system X and then their life-choices change and they're no longer able to maintain this plugin in the community.   Now, contrib has a large and unmaintained portion of code that is a liability.   Do we expect other contrib authors/maintainers to jump in and learn that ecosystem to fix bugs?   Is contrib meant to be more "communtiy self-service"?

jsuereth (approved):
Thanks for the heroic effort to define this @tedsuo!  My comments are only "issues", but in general this a huge step for the community and I'm really glad to see it happen.

My comments are mostly things I think we should open issues for and resolve via separate discussions, but just wanted to check to see if others agree.

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,181 @@
...
+
+![long term support](../internal/img/long-term-support.png)
+
+### API support
+
+Major versions of the API MUST be supported for a minimum of **three years** after the release of the next major API version. API support is defined as follows.
+
+API stability, as defined above, MUST be maintained.
+
+A version of the SDK which supports the latest minor version of the last major version of the API will continue to be maintained during LTS. Bug and security fixes MUST be backported. Additional feature development is NOT RECOMMENDED.
```
carlosalberto, line 164: > Additional feature development is NOT RECOMMENDED.

This would cover adding new members to existing interfaces, correct?

specification/glossary.md
```diff
@@ -26,8 +32,30 @@ Some other fundamental terms are documented in the [overview document](overview.
...
+
 ## Common
 
+### Signals
+
+OpenTelemetry is structured around signals. Each signal represents a coherent, stand-alone set of functionality. Each signal follows a sperate lifecycle defining its current stability level.
```
yurishkuro, line 57:
```suggestion
OpenTelemetry is structured around signals, or categories of telemetry: metrics, logs, traces. Each signal represents a coherent, stand-alone set of functionality. Each signal follows a sperate lifecycle defining its current stability level.
```

specification/glossary.md
tedsuo: Good idea, added to the glossary along with constructors.

specification/versioning-and-stability.md
tedsuo: Yes, there is one: when the API makes a major version bump. In that case, a version of the SDK (and contrib) which works with the prior major version must be maintained for the duration of the API LTS.

specification/versioning-and-stability.md
tedsuo: Yes, RECOMMENDED, MAY, and SHOULD mean the same thing from the perspective of the [IETF](https://www.ietf.org/rfc/rfc2119.txt).

I like your wording better. :)

tedsuo: Thank you, much clearer. 

I've avoided saying MUST due to push back from maintainers. It is a bit difficult to make a sweeping, cross-language statement here.

specification/versioning-and-stability.md
tedsuo: Thanks, I updated the language here. We still need this paragraph IMHO, to make the requirement that VERSIONING.MD documents must exist in each repo.

specification/versioning-and-stability.md
tedsuo: That is a good point. I added an exception, let me know if it works.

Contrib stability is going to be a rough ride...

specification/versioning-and-stability.md
tedsuo: Yeah, generally speaking we want to maintain support for v3 of Foo, even when v4 of Foo becomes available. Added a clarification.

In general, immutability is our friend here. The less we mutate the old stuff, the better.

specification/versioning-and-stability.md
tedsuo: Actually, contrib is the opposite in that sense. Packages are only added to contrib with the understanding that the OpenTelemetry SIG is now on the hook for maintaining them. This is both a stability and a security issue.

Packages maintained by third parties can be added to the [registry](http://opentelemetry.io/registry) for discoverability.

specification/versioning-and-stability.md
tedsuo: Yes; don't make the old dogs learn new tricks. :)

specification/versioning-and-stability.md
tedsuo: Added additional language to the exception to clarify this.

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,184 @@
...
+
+Once a signal component is marked as stable, the following rules MUST apply until the end of that signal’s existence.
+
+**API Stability -**
+Backward-incompatible changes to API packages MUST NOT be made unless the major version number is incremented. All existing API calls MUST continue to compile and function against all future minor versions of the same major version.
+
+Languages which ship binary artifacts SHOULD offer ABI compatibility for the API.
+
+**SDK Stability -**
+Public portions of SDK packages MUST remain backwards compatible. There are two categories of public features: **plugin interfaces** and **constructors**. Examples of plugins include the SpanProcessor, Exporter, and Sampler interfaces. Examples of constructors include configuration objects, environment variables, and SDK builders.
```
anuraaga, line 76: @tedsuo Any thoughts on a very small part of the API being annotated as alpha and thus not stable yet? While navigating the world with separate artifacts / stability for metrics, I found myself possibly wanting that. https://github.com/open-telemetry/opentelemetry-java/pull/2330/files Maybe not, but wondering what the feasibility is vs this spec.

tedsuo: Hmmm, not sure what you mean by a "small part"? Do you mean keeping the metrics API alpha while the Tracing API becomes stable? Or something else?

anuraaga: Yeah - so for example the Tracing SDK may reference the metrics API if it wants to provide monitoring of span processors. To be able to configure this, the Tracing SDK would need to expose a small API to allow configuring a `MeterProvider`. Cross cutting concerns :)

tedsuo: Well, if a stable tracing SDK wants to add the metrics API to its public surface area, then the tracing SDK gets a major version bump when the metrics API breaks. So the options appear to be: 

* Keep the tracing SDK marked as experimental, only mark the trace API and trace API contrib as stable.
* Manage the relationship via a stable trace plugin interface (i.e. make the metrics bit a `SpanProcessor` plugin, rather than add a `MetricsProvider` to the surface of the tracing SDK).
* Refrain from integrating metrics into the tracing SDK until the metrics API is stable.

A related tricky question is instrumentation. Seems like metrics instrumentation should be kept in separate packages from tracing instrumentation until both are stable. I'm skeptical of "standardized metrics instrumentation" so that whole concept is a can of worms in general; nothing stable going on in that department any time soon.

anuraaga: I think you're saying that you don't like the idea of declaring parts of APIs as unstable, via an annotation in Java as an example. That sounds fine to me :)

anuraaga (approved):

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,184 @@
...
+
+Languages which ship binary artifacts SHOULD offer ABI compatibility for public portions of the SDK.
+
+**Semantic Conventions Stability -**
+Semantic Conventions MUST NOT be removed once they are stable. New conventions MAY be added to replace usage of older conventions, but the older conventions MUST NOT be removed. Older conventions MUST be marked as deprecated when they are replaced by newer conventions.
```
iNikem, line 81: What about non-required semantic attributes? Is it allowed for auto-instrumentation to provide them in one version and stop providing them in the next minor version? They are optional, after all, right?

iNikem (approved):

specification/versioning-and-stability.md
tedsuo: Yeah, I'd like us to at least start with some firm bright lines. If we can't do it then there's a smell of some kind. 

Longer term if there is a way to be more nuanced without giving up ABI compatibility or other important features I'm all ears. :)

specification/versioning-and-stability.md
tedsuo: I think that fall under a different section, "contrib stability," in reference to the data produced by instrumentation. This section covers the stability of the conventions themselves - we can't remove the constants or anything else once they have been added to the conventions package. 

But to actually answer your question - we can't remove any data produced by instrumentation without a major version bump of that instrumentation library, as it could break dashboards and alerting. 

So, yes, they can be removed. But data should be deprecated first, with the actual deletions bundled up for release along with other major instrumentation changes, rather than have the breaking changes dribble out one by one.

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,184 @@
...
+
+For clarity, it is still possible to create new, backwards incompatible versions of existing signals without actually moving to v2.0 and breaking support.
+
+For example, imagine we develop a new, better tracing API - let's call it AwesomeTrace. We will never mutate the current tracing API into AwesomeTrace. Instead, AwesomeTrace would be added as an entirely new signal which coexists and interoperates with the current tracing signal. This would make adding AwesomeTrace a minor version bump, *not* v2.0. v2.0 would mark the end of support for current tracing, not the addition of AwesomeTrace. And we don't want to ever end that support, if we can help it.
+
+This is not actually a theoretical example. OpenTelemetry already supports two tracing APIs: OpenTelemetry and OpenTracing. We invented a new tracing API, but continue to support the old one.
+
+## Version numbers
+
+OpenTelemetry follows [semver 2.0](https://semver.org/) conventions, with the following distinctions.
```
chulkilee, line 116: It would be better to group what's the distinction from semver. The current following section is to guide how to version packages altogether - not necessarily "distinctions" from semver.

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,184 @@
...
+
+OpenTelemetry clients have four components: API, SDK, Semantic Conventions, and Contrib.
+
+For the purposes of versioning, all code within a component MUST treated as if it were part of a single package, and versioned with the same version number, except for Contrib, which may be a collection of packages versioned separately.
+
+* API packages for all signals MUST version together, across all signals. Signals MUST NOT have separate version numbers. There is one version number that applies to all signals that are included in the API release that is labeled with that particular version number.
+* SDK packages for all signals MUST version together, across all signals. Signals MUST NOT have separate version numbers. There is one version number that applies to all signals that are included in the SDK release that is labeled with that particular version number.
```
chulkilee, line 123: I'm not sure this is practical to use one version number for all signals.

My understanding is, one of the design goal is to have independent signals so that each signal can be packaged independently. Then.. any breaking changes in one signal require bumping major version of all packages, even others do not need any changes. This might be okay, but then why don't we have just one package for all signal? :confused: (note: this is kind of Rails model - ActiveRecord, ActiveModel... all are bumped altogether with Rails)

Or will I be able to mix different versions when using them? (e.g. `1.0.0` for a signal and `1.0.1` for another signal, as far as they can be used with same API package) If so, then what's the point of aligning the signal SDK versions?

Also note that signals may be in separate stages per spec (stable or experimental, for instance) - and it may be better to give 0.x or add pre-release tag to experimental signals. Enforcing one version (e.g. `1.2.3`) to all signal packages would make less use of semver actually.

BTW, as a user (and a developer as well), I prefer to have one SDK (reference) package for all signals for easier upgrade and maintenance actually, with annotating each module/function/class whatsoever with the stage.

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,184 @@
...
+* Different language implementations are NOT REQUIRED to have matching version numbers. For example, it is fine to have `opentelemetry-python-api` at v1.2.8 when `opentelemetry-java-api` is at v1.3.2.
+* Language implementations are NOT REQUIRED to match the version of the specification they implement. For example, it is fine for v1.8.2 of `opentelemetry-python-api` to implement v1.1.1 of the specification.
+
+**Exception:** in some languages, package managers may react poorly to experimental packages having a version higher than 0.X. In these cases, experimental signals MAY version independently from stable signals, in order to retain a 0.X version number. When a signal becomes stable, the version MUST be bumped to match the other stable signals in the release.
+
+### Major version bump
+
+Major version bumps MUST only occur when there is a breaking change to a stable interface, or the removal of a deprecated signal. Major version bumps MUST NOT occur for any other reason.
```
chulkilee, line 134: > Major version bumps MUST NOT occur for any other reason.

This is not required by SemVer by the way. SemVer requires "bump major version when ...." - not "you must not bump the major version unless ...". Do we need to do this?

A potential case for major version bump for "any other reason" - drop supporting end-of-life runtime. Some package manager supports checking the runtime requirement, but not all do that. For that reason, it's not uncommon to bump the major version only to drop the support for old runtime so that old runtime users wouldn't upgrade to the next version accidentally.

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,184 @@
...
+  * [Contrib Support](#contrib-support)
+- [OpenTelemetry GA](#opentelemetry-ga)
+
+<!-- tocstop -->
+
+This document defines the stability guarantees offered by the OpenTelemetry clients, along with the rules and procedures for meeting those guarantees.
+
+In this document, the terms "OpenTelemetry" and "language implementations" both specifically refer to the OpenTelemetry clients. These terms do not refer to the specification or the Collector in this document.
+
+Each language implementation MUST take these versioning and stability requirements, and produce a language-specific document which details how these requirements will be met. This document SHALL be placed in the root of each repo and named `VERSIONING.MD`.
```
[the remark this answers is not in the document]
tedsuo: Sure, no problem.

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,184 @@
...
+**SDK Stability -**
+Public portions of SDK packages MUST remain backwards compatible. There are two categories of public features: **plugin interfaces** and **constructors**. Examples of plugins include the SpanProcessor, Exporter, and Sampler interfaces. Examples of constructors include configuration objects, environment variables, and SDK builders.
+
+Languages which ship binary artifacts SHOULD offer ABI compatibility for public portions of the SDK.
+
+**Semantic Conventions Stability -**
+Semantic Conventions MUST NOT be removed once they are stable. New conventions MAY be added to replace usage of older conventions, but the older conventions MUST NOT be removed. Older conventions MUST be marked as deprecated when they are replaced by newer conventions.
+
+**Contrib Stability -**
+Plugins and instrumentation MUST be kept up to date, and MUST released simultaneously (or shortly after) the latest release of the API. The goal is to ensure users can update to the latest version of OpenTelemetry, and not be held back by the plugins that they depend on.
```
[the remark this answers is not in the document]
tedsuo: For the terminology, a couple points. I have this at the top:
> In this document, the terms "OpenTelemetry" and "language implementations" both specifically refer to the OpenTelemetry clients. These terms do not refer to the specification or the Collector in this document.

But I am thinking that needs to change. In general, we are sloppy about the term OpenTelemetry.

1) Use the term "OpenTelemetry clients" everywhere, rather than a short hand.
2) OpenTelemetry clients are defined as the API, SDK, Semantic Conventions, and Contrib packages, collectively. This could be made clearer in the overview.

Note that Contrib is part of "Official" OpenTelemetry; that is defined already under the definition of Contrib in the Overview. 3rd party plugins hosted elsewhere are not part of contrib. 

As far as MUST vs SHOULD, I do think it is reasonable that a SIG update all relevant Contrib packages as part of releasing a new version of the SDK, API, or Semantic Conventions. This should be done in a coordinated fashion – it is unfair to users to claim that a new version of the SDK is available when supported plugins do not work with it. We are not in the business of making breaking changes left and right; keeping Contrib packages up to date should not be a burden. 

If it is a burden, then we are breaking too many things. If we cannot keep up with our own changes, then 3rd party packages will definitely not be able to keep up, and users will be stuck on older versions of the SDK due to a lack of compatible plugins. It is a priority to avoid this situation. This is defined under Design Goals, let me know if that section is clear enough.

Likewise, instrumentation should always include the latest conventions and remove any API calls to deprecated methods. If this is not tracked in a coordinated fashion, it will become a mess. So, no waiting here either. Again, the API and conventions must remain stable and backwards compatible, so releasing a new version does not mean a massive rewrite of instrumentation is in order.

I feel that MUST makes this clearer than SHOULD. Let me know your thoughts.

specification/versioning-and-stability.md
tedsuo: Maybe "clarifications" is a better term. We are not deviating from semver, but we have some rules as to how we follow it.

specification/versioning-and-stability.md
tedsuo: We definitely want as few version numbers flying around as possible, and no mixing and matching. There are a number of reasons for this. Stable signals may depend upon each other. Most implementations will offer some kind of unified API or SDK package, which pulls in all of the stable signals and provides additional connivence methods. Contrib instrumentation will contain multiple signals, and the SDK may wish to instrument itself. We provided tested and verified releases. All of this is supported by treating everything within a component as part of a single release with a single version number. The four components within a release (API, SDK, SemConv, Contrib) only have separate versions due to the different time scales and different audiences who depend upon them. 

Signals are in separate packages because otherwise they could not be developed experimentally without destabilizing existing stable packages.  Experimental signals cannot be brought into any unified API/SDK/Contrib package until they become stable. So, we end up with an architecture where each signal is in its own package, but as a user, it is still possible to bring in a single package which references everything stable. If you want to use a package while it is experimental, you have to reference it directly so that you are aware of what you are doing. Since these direct references do not break when the package becomes stable, we don't punish our beta users.

I agree, this is similar to how Rails handles versioning. Everything is well factored into separate components which are then wrapped up together wherever that creates convenience, and everything has the same version number so that you don't lose your mind. 

(BTW, the infographics attempt to describe these various relationships. I will make sure this information is also in the text, so that it is not missed by screen readers.)

Note that experimental signals may keep a 0.X version number. That is already defined in this section, under "exceptions."

specification/versioning-and-stability.md
tedsuo: Very good point. I will clarify that major version numbers SHOULD happen when we drop support for anything at all, and SHOULD NOT happen for other reasons. 

I received a number of questions along the lines of "if I add X does that bump the major version?" Which is why I wanted to emphasize that we bump the major version when we drop support, not when we add support.

specification/versioning-and-stability.md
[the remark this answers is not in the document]
chulkilee: Since defining "OpenTelemetry" as clients in this doc is confusing, I think it's better to just call it OpenTelemetry clients or any other appropriate term. :+1:

> As far as MUST vs SHOULD, I do think it is reasonable that a SIG update all relevant Contrib packages as part of releasing a new version of the SDK, API, or Semantic Conventions.

I also expect it (official contrib packages are supporting new changes when released..) - but as it is "MUST", a delay of contrib implementation for any reason (e.g. lack of contributions or complexity of implementation in specific language) MUST block any release of spec changes - or we may break the promise("shortly after") - which is not desirable.

Actually, the concept of "contrib package" is being introduced in this doc - what are the example of contrib packages? Why should OpenTelemetry team need to maintain it, instead of enforcing the rules?

From previous section:

> Plugins and instrumentation that make use of the API or SDK interfaces, but are not part of the core packages necessary for running OTel, are referred to as Contrib packages.

From https://github.com/open-telemetry/opentelemetry-specification/pull/1291/files#r544742481

> [Contrib]  Packages are only added to contrib with the understanding that the OpenTelemetry SIG is now on the hook for maintaining them. This is both a stability and a security issue.

specification/overview.md
```diff
@@ -38,6 +43,32 @@ OpenTelemetry supports and defines important fundamental terms.
...
+
+### Semantics
+
+**Semantic Conventions** are schema defining the attributes which describe common concepts and operations which the signal observes. See details [below](#semantic-conventions).
+
+### Contrib Packages
+
+Plugins and instrumentation that make use of the API or SDK interfaces, but are not part of the core packages necessary for running OTel, are referred to as Contrib packages. The term "contrib" specifically refers to the plugins and instrumentation maintained by the OpenTelemetry organization outside of the SDK; it does not refer to third party plugins hosted elsewhere, or core plugins which are required to be part of the SDK release, such as OTLP Exporters and TraceContext Propagators. **API Contrib** refers to packages which depend solely upon the API; **SDK Contrib** refers to packages which also depend upon the SDK.
```
chulkilee, line 70: Running OTel => using OpenTelementry ? (No Otel definitely)

specification/versioning-and-stability.md
iNikem: But if attribute is optional, then all backends have to work with that attribute missing. Essentially you are saying "the moment auto-instrumentation adds some optional attribute for a particular instrumented library, it has to support is forever". What if next version of that library will not provide that data?

Also, what is the deprecation mechanism for semantic attributes? How can we communicate to our clients (meaning every user of auto-instrumentation) that these attribute from from instrumented library is deprecated?

specification/versioning-and-stability.md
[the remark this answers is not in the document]
tedsuo: Thanks for the feedback, I rewrote the definition of Contrib packages and added examples for clarification.

I do believe that if we offer integrations, we must keep them up to date or we have broken a promise to our end users. If we say SHOULD, then we are weakening that promise. I agree that there could be extenuating circumstances; you make a good case for SHOULD. But... I have seen a pattern where maintainers may consider this code to be less important than the SDK code. It isn't. OpenTelemetry is only as good as it's instrumentation. We need to commit to maintaining this ecosystem. The fact that it may be difficult is not a sufficient reason to loosen the requirement. Backwards compatibility is also difficult to maintain; that does not make it optional. So SHOULD makes me concerned. I will discuss this point at the maintainer's meeting tomorrow and get more feedback. In the meantime, I added some clarifying language to this section. Thank you for the well reasoned discussion.

specification/glossary.md
```diff
@@ -54,9 +82,17 @@ Denotes the library that implements the *OpenTelemetry API*.
 See [Library Guidelines](library-guidelines.md#sdk-implementation) and
 [Library resource semantic conventions](resource/semantic_conventions/README.md#telemetry-sdk).
 
+### Constructors
+
+Constructors are public code used by Application Owners to initialize and configure the OpenTelemetry SDK and contrib packages. Examples of constructors include configuration objects, environment variables, and builders.
+
+### SDK Plugins
+
+Plugins are libraries which extend the OpenTelemetry SDK Examples of plugin interfaces include the `SpanProcessor`, `Exporter`, and `Sampler` interfaces.
```
arminru, line 91:
```suggestion
Plugins are libraries which extend the OpenTelemetry SDK. Examples of plugin interfaces are the `SpanProcessor`, `Exporter`, and `Sampler` interfaces.
```

specification/overview.md
```diff
@@ -33,12 +38,43 @@ Table of Contents
...
 </details>
 
-This document provides an overview of the pillars of telemetry that
-OpenTelemetry supports and defines important fundamental terms.
+This document provides an overview of the OpenTelemetry project and defines important fundamental terms.
 
 Additional term definitions can be found in the [glossary](glossary.md).
 
-## Distributed Tracing
+## OpenTelemetry Client Architecture
```
arminru, line 45: @tedsuo Please also check if the wording introduced here matches what's defined in [library-guidelines.md](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/library-guidelines.md) and update it where needed (or consider removing duplicated sections).

specification/overview.md
```diff
@@ -33,12 +38,43 @@ Table of Contents
...
+
+![Cross cutting concerns](../internal/img/architecture.png)
+
+At the highest architectural level, OpenTelemetry clients are organized into **signals**. Each signal provides a specialized form of observability. For example, tracing, metrics, and baggage are three separate signals. Signals share a common subsystem – **context propagation** – but they function independently from each other.
+
+Each signal provides a mechanism for software to describe itself. A codebase, such as web framework or a database client, takes a dependency on various signals in order to describe itself. OpenTelemetry instrumentation code is then mixed into the other code within that codebase. This makes OpenTelemetry a **cross-cutting concern** - a piece of software which must be mixed into many other pieces of software in order to provide value. Cross-cutting concerns, by their very nature, violate a core design principle – separation of concerns. As a result, OpenTelemetry client design requires extra care and attention to avoid creating issues for the codebase which depend upon these cross-cutting APIs.
```
arminru, line 51: > OpenTelemetry instrumentation code is then mixed into the other code within that codebase.

This is not the case for instrumentation libraries provided separately or even by OpenTelemetry SIGs themselves. How about:

```suggestion
Each signal provides a mechanism for software to describe itself. A codebase, such as web framework or a database client, takes a dependency on various signals in order to describe itself. OpenTelemetry instrumentation code can be mixed into the other code within that codebase. This makes OpenTelemetry a **cross-cutting concern** - a piece of software which is mixed into many other pieces of software in order to provide value. Cross-cutting concerns, by their very nature, violate a core design principle – separation of concerns. As a result, OpenTelemetry client design requires extra care and attention to avoid creating issues for the codebase which depend upon these cross-cutting APIs.
```

PS: Adding line breaks after each sentence (or wherever semantically appropriate) would make commenting and tracking changes easier :)

specification/overview.md
```diff
@@ -33,12 +38,43 @@ Table of Contents
...
+
+OpenTelemetry clients are designed to separate the portion of each signal which must be imported as cross-cutting concerns from the portions which can be managed independently. OpenTelemetry clients are also designed to be an extensible framework. To accomplish this these goals, each signal consists of four types of packages.
+
+### API
+
+API packages consist of the cross-cutting public interfaces used for instrumentation. Any portion of an OpenTelemetry client which imported into 3rd-party libraries and application code is considered part of the API. To manage different levels of stability, every signal has its own, independent API package. Stable APIs may then be bundled up into a unified API which provides additional convenience methods.
```
arminru, line 57:
```suggestion
API packages consist of the cross-cutting public interfaces used for instrumentation. Any portion of an OpenTelemetry client which is imported into 3rd-party libraries and application code is considered part of the API. To manage different levels of stability, every signal has its own, independent API package. Stable APIs may then be bundled up into a unified API which provides additional convenience methods.
```

specification/overview.md
```diff
@@ -33,12 +38,43 @@ Table of Contents
...
+
+### SDK
+
+The implementation of the API. The SDK is managed by the application owner. Note that the SDK includes additional public interfaces which are not considered part of the API package, as they are not cross-cutting concerns. These public interfaces are defined as **constructors** and **plugin interfaces**. Examples of plugin interfaces include the `SpanProcessor`, `Exporter`, and `Sampler` interfaces. Examples of constructors include **configuration objects**, **environment variables**, and **SDK builders**. Application owners use the SDK constructors; plugin authors use the SDK plugin interfaces. Instrumentation authors must never directly reference any SDK package of any kind, only the API.
```
arminru, line 61:
```suggestion
The implementation of the API. The SDK is managed by the application owner. Note that the SDK includes additional public interfaces which are not considered part of the API package, as they are not cross-cutting concerns. These public interfaces are defined as [**constructors**](glossary.md#constructors) and [**plugin interfaces**](glossary.md#sdk-plugins). Examples of plugin interfaces include the `SpanProcessor`, `Exporter`, and `Sampler` interfaces. Examples of constructors include **configuration objects**, **environment variables**, and **SDK builders**. Application owners use the SDK constructors; plugin authors use the SDK plugin interfaces. Instrumentation authors must never directly reference any SDK package of any kind, only the API.
```

specification/overview.md
```diff
@@ -33,12 +38,43 @@ Table of Contents
...
+
+### Semantics
+
+The **Semantic Conventions** define the keys and values which describe commonly observed concepts and operations within cloud computing. See details [below](#semantic-conventions).
```
arminru, line 65:
```suggestion
The **Semantic Conventions** define the keys and values which describe commonly observed concepts and operations in applications. See details [below](#semantic-conventions).
```
They also apply to software running in old-school (on-prem or remote) data centers and such.

specification/overview.md
tedsuo: Thanks, will do. But do we have rules here? Seems like the goal of line breaks is to keep the width to roughly 80 characters, not to break things up semantically.

specification/versioning-and-stability.md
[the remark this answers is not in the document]
tedsuo: Based on the maintainers meeting today, we are going with SHOULD for now. But in general, it is clear that we need a more well defined plan for what we allow in contrib, how we maintain it, and how we treat 3rd-party plugins and instrumentation which we do not maintain.

The situation I want to avoid is one where we agree to take something into contrib, but that results in issues and PRs not being managed properly.

On the one hand, if these plugins are left to 3rd parties, then there is no "blessed" plugin and it is easier for alternatives to be generated by new maintainers. On the other hand, we are concerned about the security and reliability implications of packaging up and distributing 3rd-party plugins in the auto-installers and releases we provide. 

We'd prefer to have official plugins which are reliably maintained, but we do not yet have a clear plan for how to do this in a manner which addresses all concerns. We need a plan, but that is outside of the scope of this current proposal. So, leaving it as SHOULD for now.

If you have a good idea on how to address this, please consider making an OTEP. :)

specification/overview.md
arminru: I remember we had some discussions around this topic but don't think there's a documented "style guide" or something like that. I think we agreed that we don't want to break sentences apart at some artificial hard limit but rather where it makes sense semantically to still get a nice diff (i.e., at the end of sentences and for long sentences between clauses). This makes changes easier to track (also in the blame view) and you  don't have 5 different threads/conversations for the same line if statements are split into separate lines.

specification/overview.md
tedsuo: I made some improvements to this document to match the new information. At some point, I believe we could take another pass at [library-guidelines](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/library-guidelines.md) and [library-layout](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/library-layout.md). But I would like that to be a separate PR as it would greatly expand this PR, which is already rather large.

specification/overview.md
tedsuo: Sounds good. I formatted my portion in this manner, but left pre-existing content alone. This could be cleaned up in another PR.

jkwatson:
A suggestion from the spec SIG meeting today was to punt on (remove) the requirements for instrumentation/contrib stability for this PR, because we think that instrumentation stability is a much thornier (and more important!) problem that shouldn't hold us back from a 1.0 release of the API and SDK.

tedsuo:
Hi all. I've addressed all of the change requests that have been made to-date. A major change from the original OTEP has been to remove the definition of telemetry stability. That definition has been replaced with a notice that telemetry stability is TBD, and that contrib instrumentation must not be marked as stable until the issue is resolved.

I've created an issue for discussing this problem: https://github.com/open-telemetry/opentelemetry-specification/issues/1301

At this point, I'm finished with edits until more change requests come in. @open-telemetry/specs-approvers please review, and either approve or leave a comment.

Thank you everyone for all of the valuable feedback!

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,256 @@
...
+Public portions of contrib packages (constructors, configuration, interfaces) SHOULD remain backwards compatible.
+
+Languages which ship binary artifacts SHOULD offer [ABI compatibility](glossary.md#abi-compatibility) for contrib packages.
+
+**Exception:** Contrib packages MAY break stability when a required downstream dependency breaks stability.
+For example, a database integration may break stability if the required database client breaks stability.
+However, it is strongly RECOMMENDED that older contrib packages remain stable.
+A new, incompatible version of an integration SHOULD be released as separate contrib package, rather than break the existing contrib package.
+
+#### NOT DEFINED: Telemetry Stability
```
GregMefford, line 129: I don't have strong feelings about this, but was curious if we need to have `NOT DEFINED` here in the title, and as a result, in the TOC as well. This section _does_ define how we want to manage Telemetry stability - it's just that the plan is not complete yet. When there is a more-detailed plan, it will still belong in this section, and there won't be a change needed to the title of the section or to the TOC.

```suggestion
#### Telemetry Stability
```

Maybe this is an OTel documentation convention I'm not familiar with, in which case just ignore this suggestion.

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,256 @@
...
+### Removal
+
+Support is ended by the removal of a signal from the release.
+The release MUST make a major version bump when this happens.
+
+### A note on replacing signals
+
+Note that we currently have no plans for creating a major version of OpenTelemetry past v1.0.
+
+For clarity, it is still possible to create new, backwards incompatible versions of existing signals without actually moving to v2.0 and breaking support.
```
GregMefford, line 154:
```suggestion
For clarity, it is still possible to create new, backwards-incompatible versions of existing signals without actually moving to v2.0 and breaking support.
```

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,256 @@
...
+OpenTelemetry already supports two tracing APIs: OpenTelemetry and OpenTracing.
+We invented a new tracing API, but continue to support the old one.
+
+## Version numbers
+
+OpenTelemetry clients follow [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html), with the following clarifications.
+
+OpenTelemetry clients have four components: API, SDK, Semantic Conventions, and Contrib.
+
+For the purposes of versioning, all code within a component MUST treated as if it were part of a single package, and versioned with the same version number,
```
GregMefford, line 173:
```suggestion
For the purposes of versioning, all code within a component MUST be treated as if it were part of a single package, and versioned with the same version number,
```

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,256 @@
...
+except for Contrib, which may be a collection of packages versioned separately.
+
+* API packages for all signals MUST version together, across all signals.
+Signals MUST NOT have separate version numbers.
+There is one version number that applies to all signals that are included in the API release that is labeled with that particular version number.
+* SDK packages for all signals MUST version together, across all signals.
+Signals MUST NOT have separate version numbers.
+There is one version number that applies to all signals that are included in the SDK release that is labeled with that particular version number.
+* Semantic Conventions are a single package with a single version number.
+* Each contrib package MAY have it's own version number.
```
GregMefford, line 183:
```suggestion
* Each Contrib package MAY have its own version number.
```

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,256 @@
...
+* The API, SDK, Semantic Conventions, and contrib components have independent version numbers.
```
GregMefford, line 184:
```suggestion
* The API, SDK, Semantic Conventions, and Contrib components have independent version numbers.
```

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,256 @@
...
+![long term support](../internal/img/long-term-support.png)
+
+### API support
+
+Major versions of the API MUST be supported for a minimum of **three years** after the release of the next major API version.
+API support is defined as follows.
+
+* API stability, as defined above, MUST be maintained.
+
+* A version of the SDK which supports the latest minor version of the last major version of the API will continue to be maintained during LTS.
```
GregMefford, line 235:
```suggestion
* A version of the SDK which supports the latest minor version of the last major version of the API MUST continue to be maintained during Long Term Support (LTS).
```

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,256 @@
...
+SDK stability, as defined above, will be maintained for a minimum of **one year** after the release of the next major SDK version.
+
+### Contrib Support
+
+Contrib stability, as defined above, will be maintained for a minimum of **one year** after after the release of the next major version of a contrib package.
+
+## OpenTelemetry GA
+
+The term “OpenTelemetry GA” refers to the point at which OpenTracing and OpenCensus will be fully deprecated.
+The **minimum requirements** for declaring GA are as followed.
```
GregMefford, line 253:
```suggestion
The **minimum requirements** for declaring GA are as follows.
```

specification/library-guidelines.md
```diff
@@ -1,52 +1,53 @@
...
+OpenTelemetry clients are expected to provide full features out of the box and allow for innovation and experimentation through extensibility.
 
-The document does not attempt to describe a language library API. For API specs see [specification](../README.md).
+Please read the [overview](overview.md) first, to understand the fundamental architecture of OpenTelemetry.
 
-_Note to Language Library Authors:_ OpenTelemetry specification, API and SDK implementation guidelines are work in progress. If you notice incomplete or missing information, contradictions, inconsistent styling and other defects please let specification writers know by creating an issue in this repository or posting in [Gitter](https://gitter.im/open-telemetry/opentelemetry-specification). As implementors of the specification you will often have valuable insights into how the specification can be improved. The Specification SIG and members of Technical Committee highly value your opinion and welcome your feedback.
+The document does not attempt to describe the details or functionality of the OpenTelemetry client API. For API specs see the [API specifications](../README.md).
```
GregMefford, line 9:
```suggestion
The document does not attempt to describe the details or functionality of the OpenTelemetry client API. For API specs, see the [API specifications](../README.md).
```

specification/library-guidelines.md
```diff
@@ -1,52 +1,53 @@
...
+
+_Note to OpenTelemetry client Authors:_ OpenTelemetry specification, API and SDK implementation guidelines are work in progress. If you notice incomplete or missing information, contradictions, inconsistent styling and other defects please let specification writers know by creating an issue in this repository or posting in [Gitter](https://gitter.im/open-telemetry/opentelemetry-specification). As implementors of the specification you will often have valuable insights into how the specification can be improved. The Specification SIG and members of Technical Committee highly value your opinion and welcome your feedback.
```
GregMefford, line 11:
```suggestion
_Note to OpenTelemetry Library Authors:_ OpenTelemetry specification, API and SDK implementation guidelines are works in progress. If you notice incomplete or missing information, contradictions, inconsistent styling or other defects please let the specification writers know by creating an issue in this repository or posting in [Gitter](https://gitter.im/open-telemetry/opentelemetry-specification). As implementors of the specification, you will often have valuable insights into how the specification can be improved. The Specification SIG and members of Technical Committee highly value your opinion and welcome your feedback.
```

specification/library-guidelines.md
```diff
@@ -1,52 +1,53 @@
...
 
 ## Requirements
 
 1. The OpenTelemetry API must be well-defined and clearly decoupled from the implementation. This allows end users to consume API only without also consuming the implementation (see points 2 and 3 for why it is important).
 
-2. Third party libraries and frameworks that add instrumentation to their code will have a dependency only on the API of OpenTelemetry language library. The developers of third party libraries and frameworks do not care (and cannot know) what specific implementation of OpenTelemetry is used in the final application.
+2. Third party libraries and frameworks that add instrumentation to their code will have a dependency only on the API of OpenTelemetry client. The developers of third party libraries and frameworks do not care (and cannot know) what specific implementation of OpenTelemetry is used in the final application.
```
GregMefford, line 17:
```suggestion
2. Third-party libraries and frameworks that add instrumentation to their code will have a dependency only on the API of OpenTelemetry client. The developers of third-party libraries and frameworks do not care (and cannot know) what specific implementation of OpenTelemetry is used in the final application.
```

specification/library-guidelines.md
```diff
@@ -1,52 +1,53 @@
...
 
 3. The developers of the final application normally decide how to configure OpenTelemetry SDK and what extensions to use. They should be also free to choose to not use any OpenTelemetry implementation at all, even though the application and/or its libraries are already instrumented.  The rationale is that third-party libraries and frameworks which are instrumented with OpenTelemetry must still be fully usable in the applications which do not want to use OpenTelemetry (so this removes the need for framework developers to have "instrumented" and "non-instrumented" versions of their framework).
 
 4. The SDK must be clearly separated into wire protocol-independent parts that implement common logic (e.g. batching, tag enrichment by process information, etc.) and protocol-dependent telemetry exporters. Telemetry exporters must contain minimal functionality, thus enabling vendors to easily add support for their specific protocol.
```
GregMefford, line 21:
```suggestion
4. The SDK must be clearly separated into protocol-independent parts that implement common logic (e.g. batching, tag enrichment by process information, etc.) and protocol-dependent telemetry exporters. Telemetry exporters must contain minimal functionality, thus enabling vendors to easily add support for their specific protocol.
```

specification/library-guidelines.md
```diff
@@ -1,52 +1,53 @@
...
 
-Third-party libraries and frameworks that want to be instrumented in OpenTelemetry-compatible way will have a dependency on the API package. The developers of these third-party libraries will add calls to telemetry API to produce telemetry data.
+Libraries, frameworks, and applications that want to be instrumented with OpenTelemetry take a dependency only on the API packages. The developers of these third-party libraries will make calls to the API to produce telemetry data.
 
-Applications that use third-party libraries that are instrumented with OpenTelemetry API will have a choice to enable or not enable the actual delivery of telemetry data. The application can also call telemetry API directly to produce additional telemetry data.
+Applications that use third-party libraries that are instrumented with OpenTelemetry API control whether or not to install the SDK and generate of telemetry data. When the SDK is not installed, the API calls should be no-ops which generate minimal overhead.
```
GregMefford, line 48:
```suggestion
Applications that use third-party libraries that are instrumented with OpenTelemetry API control whether or not to install the SDK and generate telemetry data. When the SDK is not installed, the API calls should be no-ops which generate minimal overhead.
```

specification/library-guidelines.md
```diff
@@ -1,52 +1,53 @@
...
 
-In order to enable telemetry the application must take a dependency on the OpenTelemetry SDK, which implements the delivery of the telemetry. The application must also configure exporters so that the SDK knows where and how to deliver the telemetry. The details of how exporters are enabled and configured are language specific.
+In order to enable telemetry the application must take a dependency on the OpenTelemetry SDK. The application must also configure exporters and other plugins so that telemetry can be correctly generated and delivered to their analysis tool of choice. The details of how plugins are enabled and configured are language specific.
```
GregMefford, line 50:
```suggestion
In order to enable telemetry, the application must take a dependency on the OpenTelemetry SDK. The application must also configure exporters and other plugins so that telemetry can be correctly generated and delivered to their analysis tool(s) of choice. The details of how plugins are enabled and configured are language-specific.
```

specification/library-guidelines.md
```diff
@@ -102,7 +103,7 @@ The end-user application may decide to take a dependency on alternative implemen
 
 SDK provides flexibility and extensibility that may be used by many implementations. Before developing an alternative implementation, please, review extensibility points provided by OpenTelemetry.
 
-An example use case for alternate implementations is automated testing. A mock implementation can be plugged in during automated tests. For example it can store all generated telemetry data in memory and provide a capability to inspect this stored data. This will allow the tests to verify that the telemetry is generated correctly. Language Library authors are encouraged to provide such mock implementation.
+An example use case for alternate implementations is automated testing. A mock implementation can be plugged in during automated tests. For example it can store all generated telemetry data in memory and provide a capability to inspect this stored data. This will allow the tests to verify that the telemetry is generated correctly. OpenTelemetry client authors are encouraged to provide such mock implementation.
```
GregMefford, line 106:
```suggestion
An example use-case for alternate implementations is automated testing. A mock implementation can be plugged in during automated tests. For example, it can store all generated telemetry data in memory and provide a capability to inspect this stored data. This will allow the tests to verify that the telemetry is generated correctly. OpenTelemetry client authors are encouraged to provide such a mock implementation.
```

specification/library-guidelines.md
```diff
@@ -102,7 +103,7 @@ The end-user application may decide to take a dependency on alternative implemen
...
 
 Note that mocking is also possible by using SDK and a Mock `Exporter` without needed to swap out the entire SDK.
```
GregMefford, line 108:
```suggestion
Note that mocking is also possible by using SDK and a Mock `Exporter` without needing to swap out the entire SDK.
```

specification/library-guidelines.md
```diff
@@ -112,9 +113,9 @@ The mocking approach chosen will depend on the testing goals and at which point
...
+Because API and SDK package version numbers are not coupled, every API and SDK package release MUST clearly mention the Specification version number that they implement. In addition, if a particular version of SDK package is only compatible with a specific version of API package, then this compatibility information must be also published by OpenTelemetry client authors. OpenTelemetry client authors MUST include this information in the release notes. For example, the SDK package release notes may say: "SDK 0.3.4, use with API 0.1.0, implements OpenTelemetry Specification 0.1.0".
 
 _TODO: How should third party library authors who use OpenTelemetry for instrumentation guide their end users to find the correct SDK package?_
```
GregMefford, line 120:
```suggestion
_TODO: How should third-party library authors who use OpenTelemetry for instrumentation guide their end users to find the correct SDK package?_
```

GregMefford (commented):
I was going through this in detail to make sure I understand it before working on [the relevant documentation for Erlang/Elixir](https://github.com/open-telemetry/opentelemetry-erlang/pull/188), and I figured it would be a good time to batch up suggestions for small grammatical tweaks.

carlosalberto (approved):

specification/glossary.md
```diff
@@ -26,8 +36,46 @@ Some other fundamental terms are documented in the [overview document](overview.
...
+
+In this specification, the term **package** describes a set of code which represents a single dependency, which may be imported into a program independently from other packages.
+This concept may map to a different term in some languages, such as "module."
+Note that in some languages, the term "package" refers to a different concept.
+
+### ABI Compatibility
+
+An ABI (application binary interface) is an interface which defines interactions between software components at the machine code level, for example between an application executable and a compiled binary of a shared object library. ABI compatibility means that a new compiled version of a library may be correctly linked to a target executable without the need for that executable to be recompiled.
+
+ABI compatibility is important for most languages which produce compiled binary executables. For interpreted languages which do not produce binary executables, ABI compatibility is a non-existent requirement.
```
jkwatson, line 77:
```suggestion
ABI compatibility is important for most languages which produce compiled binary executables. For interpreted languages which do not produce binary executables, ABI compatibility may be a non-existent requirement.
```

jkwatson: I suggest changing this to "may be" since Java does have ABI requirements, but doesn't produce a binary executable (and is technically interpreted at the JVM/bytecode level).

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,256 @@
...
+Public portions of SDK packages MUST remain backwards compatible.
+There are two categories of public features: **plugin interfaces** and **constructors**.
+Examples of plugins include the SpanProcessor, Exporter, and Sampler interfaces.
+Examples of constructors include configuration objects, environment variables, and SDK builders.
+
+Languages which ship binary artifacts SHOULD offer [ABI compatibility](glossary.md#abi-compatibility) for SDK packages.
+
+#### Semantic Conventions Stability
+
+Semantic Conventions MUST NOT be removed once they are stable.
```
jkwatson, line 110: Do we want to split the semconv into a separate package? 

Also, they are more related to telemetry and we are punting on telemetry stability in this PR?

jkwatson: I see this is alluded to below...it might be good to call it out along with the signals being packaged independently.

jkwatson (approved):
👍 Thanks! A couple of small clarifications, but this generally is excellent.

specification/overview.md
```diff
@@ -8,37 +8,90 @@ Table of Contents
...
+This makes OpenTelemetry a **cross-cutting concern** - a piece of software which is be mixed into many other pieces of software in order to provide value. Cross-cutting concerns, by their very nature, violate a core design principle – separation of concerns. As a result, OpenTelemetry client design requires extra care and attention to avoid creating issues for the codebase which depend upon these cross-cutting APIs.
+
+OpenTelemetry clients are designed to separate the portion of each signal which must be imported as cross-cutting concerns from the portions which can be managed independently. OpenTelemetry clients are also designed to be an extensible framework.
+To accomplish this these goals, each signal consists of four types of packages: API, SDK, Semantic Conventions, and Contrib.
+
+### API
+
+API packages consist of the cross-cutting public interfaces used for instrumentation. Any portion of an OpenTelemetry client which is imported into 3rd-party libraries and application code is considered part of the API.
+To manage different levels of stability, every signal has its own, independent API package.
```
jkwatson, line 61: Is this mandatory? Or can all stable signals be in the main API package, and only experimental ones split out separately?

specification/glossary.md
tedsuo: Thanks, I reworded this:

> ABI compatibility is important for some languages, especially those which provide a form of machine code. For other languages, ABI compatibility may not be a relevant requirement.

specification/versioning-and-stability.md
tedsuo: Sure. I have some about this in the overview. Is there a another section which should be expanded to include these descriptions?

specification/overview.md
tedsuo: Well, the issue is how to manage the transition from experimental to stable with minimal disruption. If packages must start out experimental, and thus in a separate package, then that separate package would probably stick around. 

For packages that exist as part of the initial v1.0, I'm not sure that the requirement matters, though it might be kind of ugly to have some packages to be laid out one way, and others another...

anuraaga: Recently thought about the experimental -> stable transition and am hoping that fair amount of disruption is fine, and is what a user signs up for when using the alpha versioned artifact. In fact, they will probably have disruption on minor versions since it's alpha - otherwise it would just be GA :-) For example of duplication, we have some duplication of API to metrics to support the version split - we have `GlobalOpenTelemetry` to manage a global `OpenTelemetry` which exposes all stable signals, and a `GlobalMeterProvider` to manage a global metrics signal because it can't be part of `OpenTelemetry` yet. We would like to remove `GlobalMeterProvider` when metrics goes GA - this will cause disruption though, and relatively significant disruption I think as it is basically changing the entry point to metrics.

I'm hoping that the only users of alpha signals are OpenTelemetry instrumentation itself, then we can manage disruption ourselves. And any potential end-user needs to know they're not going to have a good time - that's my expectation of the alpha signal and a versioned artifact like `1.3.1-alpha`.

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,256 @@
...
+#### SDK Stability
+
+Public portions of SDK packages MUST remain backwards compatible.
+There are two categories of public features: **plugin interfaces** and **constructors**.
+Examples of plugins include the SpanProcessor, Exporter, and Sampler interfaces.
+Examples of constructors include configuration objects, environment variables, and SDK builders.
+
+Languages which ship binary artifacts SHOULD offer [ABI compatibility](glossary.md#abi-compatibility) for SDK packages.
+
+#### Semantic Conventions Stability
```
tigrannajaryan, line 108: Later in the document we say that Telemetry is not declared stable. However, at least of the Telemetry emitted depends on semantic conventions. So, in effect by saying Semantic Conventions are stable we are placing at least partial limitations on Telemetry as well.
I think we should also remove semantic conventions from initial stability guarantees.

tigrannajaryan (commented):
Ted, thanks a lot for leading this!

specification/versioning-and-stability.md
jkwatson: In the overview doc, it is explicitly mentioned in the API section that: 
`To manage different levels of stability, every signal has its own, independent API package.`

But not in the semantic conventions section. I think that's where I'm thinking it should be added.

specification/error-handling.md
```diff
@@ -62,7 +62,7 @@ The mechanism by which end users set or register a custom error handler should f
 ### Examples
 
 These are examples of how end users might register custom error handlers.
-Examples are for illustration purposes only. Language library authors
+Examples are for illustration purposes only. OpenTelemetry client authors
```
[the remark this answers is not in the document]
tedsuo: Ah, no group decision here. I began using the term "OpenTelemetry client" because "language library" wasn't clear - we have many libraries in each language, and none of these libraries contain a language. "Client" is a more common term for the set of components we are referring too. Open to other options.

specification/overview.md
tedsuo: So, the problem is not alpha, but beta. The stability guarantee can only come after we have some production usage and feedback. Breaking alpha users is okay, but a huge break in support looming over the beta will discourage participation. Who could suggest adopting a beta in good conscience if the knew for a fact that all the code they wrote would break on them as part of the v1.0? I really would prefer there to be no breaking change between the final beta, the release candidate, and the v1.0 release, except for fixing the usage issues those releases are designed to catch.

jkwatson: I think the way we would do this is not to break anyone's code, but just remove the requirement to import that alpha package separately. Everything else (java packaging, naming, etc) would remain exactly the same when we move the code into the omnibus API module.

anuraaga: @jkwatson So are we committing to deprecating but maintaining until a future major version APIs we only added due to the artifact split, like `GlobalMeterProvider`? Ideally we didn't, but maybe we have to live with it.

jkwatson: right. we can deprecate things in the alpha module, and not carry them forward when we move everything into the main module.

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,256 @@
...
+
+Patch versions make no changes which would require recompilation or potentially break application code.
+The following are examples of patch fixes.
+
+* Bug fixes which don't require minor version bump per rules above.
+* Security fixes.
+* Documentation.
+
+Currently, the OpenTelemetry project does NOT have plans to backport bug and security fixes to prior minor versions of the SDK.
+Security and bug fixes MAY only applied to the latest minor version.
```
MrAlias, line 221:
```suggestion
Security and bug fixes MAY only be applied to the latest minor version.
```

MrAlias (approved):

specification/versioning-and-stability.md
tedsuo: Ok I am convinced :)

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,256 @@
...
+The experiment MAY be discarded and removed.
+
+**Stable –** Stability guarantees, as defined by component type below, MUST now apply to all changes within this signal.
+Long-term dependencies MAY now be taken against this signal.
+
+**Deprecated –** this signal has been replaced, but MUST maintain the same stability guarantees.
+
+**Removed -** a deprecated signal is no longer supported, and MUST be removed from the release.
+
+All signal components MAY become stable together, or MAY transition to stability component-by-component in the following order: API, Semantic Conventions, API Contrib, SDK, SDK Contrib.
```
cijothomas, line 81: What is the significance of this ordering? Does it mean, we cannot have stable SDK, *before* SemanticConventions are stable?

specification/overview.md
```diff
@@ -8,37 +8,90 @@ Table of Contents
...
+
+### Contrib Packages
+
+The OpenTelemetry project maintains integrations with popular OSS projects which have been identified as important for observing modern web services.
+Example API integrations include instrumentation for web frameworks, database clients, and message queues.
+Example SDK integrations include plugins for exporting telemetry to popular analysis tools and telemetry storage systems.
+
+Some plugins, such as OTLP Exporters and TraceContext Propagators, are required by the OpenTelemetry specification. These required plugins are included as part of the SDK.
+
+Plugins and instrumentation packages which are optional and separate from the SDK are referred to as **Contrib** packages.
```
cijothomas, line 89: Is this same as all the packages maintained in the -contrin github repos for individual languages?

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,256 @@
...
+## Version numbers
+
+OpenTelemetry clients follow [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html), with the following clarifications.
+
+OpenTelemetry clients have four components: API, SDK, Semantic Conventions, and Contrib.
+
+For the purposes of versioning, all code within a component MUST treated as if it were part of a single package, and versioned with the same version number,
+except for Contrib, which may be a collection of packages versioned separately.
+
+* API packages for all signals MUST version together, across all signals.
```
cijothomas, line 176: Did you mean to say "all stable signals" MUST version together?

specification/versioning-and-stability.md
cijothomas: > OpenTelemetry SIG is now on the hook for maintaining them

I think this needs to be clarified in some official place. It was raised in maintainers meeting, and it was told that an issue will be opened to track "how to accept contrib and maintain.". It may be better to add TODO in this PR to that issue.

specification/versioning-and-stability.md
tigrannajaryan: @tedsuo do you plan to update this doc to remove Semantic Conventions Stability guarantees?

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,256 @@
...
+Bug and security fixes will be backported.
+Additional feature development is NOT RECOMMENDED.
+
+### SDK Support
+
+SDK stability, as defined above, will be maintained for a minimum of **one year** after the release of the next major SDK version.
+
+### Contrib Support
+
+Contrib stability, as defined above, will be maintained for a minimum of **one year** after after the release of the next major version of a contrib package.
```
andrewhsu, line 248: minor suggestion: remove redundant word
```suggestion
Contrib stability, as defined above, will be maintained for a minimum of **one year** after the release of the next major version of a contrib package.
```

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,256 @@
...
+The goal is to ensure users can update to the latest version of OpenTelemetry, and not be held back by the plugins that they depend on.
+
+Public portions of contrib packages (constructors, configuration, interfaces) SHOULD remain backwards compatible.
+
+Languages which ship binary artifacts SHOULD offer [ABI compatibility](glossary.md#abi-compatibility) for contrib packages.
+
+**Exception:** Contrib packages MAY break stability when a required downstream dependency breaks stability.
+For example, a database integration may break stability if the required database client breaks stability.
+However, it is strongly RECOMMENDED that older contrib packages remain stable.
+A new, incompatible version of an integration SHOULD be released as separate contrib package, rather than break the existing contrib package.
```
andrewhsu, line 127: minor suggestion: add article
```suggestion
A new, incompatible version of an integration SHOULD be released as a separate contrib package, rather than break the existing contrib package.
```

specification/versioning-and-stability.md
```diff
@@ -0,0 +1,256 @@
...
+
+<!-- tocstop -->
+
+This document defines the stability guarantees offered by the OpenTelemetry clients, along with the rules and procedures for meeting those guarantees.
+
+In this document, the terms "OpenTelemetry" and "language implementations" both specifically refer to the OpenTelemetry clients.
+These terms do not refer to the specification or the Collector in this document.
+
+Each language implementation MUST take these versioning and stability requirements, and produce a language-specific document which details how these requirements will be met.
+This document SHALL be placed in the root of each repo and named `VERSIONING`.
```
andrewhsu, line 36: i suggest having a markdown file extension to be explicitly clear that the file is meant to be documentation and not a simple text file containing the latest version number
```suggestion
This document SHALL be placed in the root of each repo and named `VERSIONING.md`.
```

andrewhsu (approved):
LGTM

i put in minor suggestions for changes to fix grammar and make things more clear, but they are not blockers in my mind

specification/overview.md
tedsuo: Ok, I understand. There are multiple approaches to achieving these guarantees, and keeping the experimental code from affecting stable code. Separate packages is one, the approach @jkwatson mentions is another. A third would be to keep experimental work on a branch - .NET would like to take this approach.

Ultimately, the goal is to meet the stability goals, not dictate a particular approach. I'll re-write this section to make the goals clearer, and emphasize that the separate packages is just one example approach.

specification/error-handling.md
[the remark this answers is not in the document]
tedsuo: I haven't been able to elicit much comment on this one. There's some preference for "OpenTelemetry client" over "language library" as it is a more standard term 🤷‍♀️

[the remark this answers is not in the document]
yurishkuro: How is "OpenTelemetry client" different from "OpenTelemetry SDK", which is what we generally use?

specification/versioning-and-stability.md
tedsuo: Yes, sorry for the delay. I've updated the doc.

tigrannajaryan (approved):
LGTM

specification/versioning-and-stability.md
tedsuo: Based on discussion in other threads, I've removed references to this architecture proposal.

specification/overview.md
tedsuo: I've removed references to independent packages.

specification/error-handling.md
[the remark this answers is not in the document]
tedsuo: @yurishkuro  OpenTelemetry client replaces "language library." It contains: API, SDK, SemConv, and Contrib.

specification/versioning-and-stability.md
tedsuo: Removing. The only actual rule is that the API must become stable before other components (or at the same time).

specification/overview.md
tedsuo: Correct. Everything we provide, but is not included by default.

specification/versioning-and-stability.md
tedsuo: Ah, yes. I cleaned this language up.

specification/versioning-and-stability.md
tedsuo: I had that actually, but it was requested that markdown format not be required, so I removed it.

jmacd (approved):

cijothomas (approved):

tedsuo:
Thank you everyone for all of the input. @open-telemetry/technical-committee can I get a merge? I believe this is complete for now.

#### Future work:
* Data stability: we need to advance the work in metrics farther before we can define this.
* Opentelemetry client vs language library: terminology bikeshed anyone may take up if they care.

carlosalberto:
@tedsuo Thanks! Will merge now as all comments are resolved _and_ we have had enough approvals for some days now.

Would you mind the issues for anything that is still needed? If there's still no issue for that, that is (as we have at least one issue regarding telemetry data already, etc).

cc @arminru @yurishkuro

PR #596, opened 2020-05-12
Support nested Attribute values (#376)

shengxil:
Related issue #376 

Nested Attribute is useful for representing multi-level values. For example, HTTP headers:
```
'http': {
    'headers': [
        'x-forwarded-for': 'foo.bar'
        'keep-alive': 'timeout=5'
    ]
}
``` 
Currently such cases are usually represented by a period delimited key chain (e.g. `'http.headers.x-forwarded-for': 'foo.bar'`). This does not work if any key contains a period itself. It is also difficult for SDKs to iterate the Attributes under a certain level.

This change allows nested Attributes by adding an Attribute value type - child Attribute. An attribute can also contain a homogeneous array of children attributes.

### Why not map?
1. `Map` is somehow a high level data structure. It has different definitions in different languages. In some languages putting a map value replaces the original value, while in other languages it becomes a sibling of the original value. 
2. Programming-wise, it is hard to keep the homogeneity in map values.

* Actually, the OT spec never used map in any span data type

PR #1293, opened 2020-12-16
Adding engine definition under resources semantic conventions

ivomagi:
Fixes https://github.com/open-telemetry/opentelemetry-specification/issues/1143

## Changes
Splunk will be adding auto-instrumentation support to various Java application servers (Tomcat, Jetty, JBoss, Weblogic, Websphere, …) during the forthcoming months. During 2021, this support will be expanded for both Java application servers as well as engines for other runtimes besides JVM. 

While implementing the support, we discovered that as of now there is no semantic convention about sending the information covering “this span was captured from Apache Tomcat version 9.0.39”.  

Capturing and exposing this information would result in: 
* Richer information in service map / runtime application architecture, supporting directed troubleshooting during incident management (Why is half of the production cluster still running previous version of Apache Tomcat)
* Possibility to correlate configuration changes in version upgrades with incidents (Particular error appeared right after upgrading from Wildfly 20 to Wildfly 21)

With other APM vendors also capturing & exposing this information; standardizing it in OpenTelemetry is a way to guarantee vendor-compatibility.

Related issues #


Oberon00 (commented):
I think "engine" is a bit arbitrary. IMHO what we are really missing is some concept of listing technologies used in a process and their relationships to each other (who runs on top of what, etc).

semantic_conventions/resource/engine.yaml
```diff
@@ -0,0 +1,22 @@
...
+    prefix: engine
+    brief: >
+       Resource describing the packaged software running the application code. Engines are executed using process.runtime.
+    attributes:
+      - id: name
+        type: string
+        required: always
+        brief: >
+          The name of the engine.
+        examples: ['FildFly']
```
iNikem, line 12:
```suggestion
        examples: ['WildFly']
```

specification/resource/semantic_conventions/engine.md
```diff
@@ -0,0 +1,52 @@
...
+| `engine.name` | string | The name of the engine. | `FildFly` | Yes |
+| `engine.version` | string | The version of the engine, as returned by the engine (preferably during runtime) without modification. | `21.0.0` | No |
+| `engine.description` | string | Additional description about the engine. | `WildFly Full 21.0.0.Final (WildFly Core 13.0.1.Final) - 2.2.2.Final` | No |
+<!-- endsemconv -->
+
+Information describing the engine SHOULD be captured using the values acquired from the API provided by the engine, preferably during runtime, to avoid maintenance burden on engine version upgrades.
+
+A resource can be attributed to at most one engine. To illustrate, let's look at a Python application using Apache HTTP Server with mod_wsgi as the server and Django as the web framework. In this case:
+
+* Apache HTTP Server would be set as `process.runtime`
```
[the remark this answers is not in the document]
iNikem: First, I agree with @Oberon00 about `process.runtime` and that `engine` will probably be Apache HTTP if we want to be strict. If we want to be lenient, then both HTTP server and `mod_wsgi` can be described as engine.

Second, as usual :), I propose to keep scope small. Can `runtime_stack` be a separate discussion?

specification/resource/semantic_conventions/engine.md
```diff
@@ -0,0 +1,52 @@
...
+* `mod_wsgi` would be set as engine
+* Django would not be set as an engine (Django is not running on `process.runtime`).
+
+## Java Engines:
+
+Java engines are often but not always packaged as application servers. For Java application servers supporting Servlet API the required information can be captured by invoking ServletContext.getServerInfo() during runtime and parsing the result.
```
iNikem, line 25:
```suggestion
Java engines are often but not always packaged as application servers. For Java application servers supporting Servlet API the required information MAY be captured by invoking `ServletContext.getServerInfo()` during runtime and parsing the result.
```

specification/resource/semantic_conventions/engine.md
```diff
@@ -0,0 +1,52 @@
...
+
+Java engines SHOULD use the following for engine.name attribute If none of the listed values apply, custom value MAY be used:
```
iNikem, line 27:
```suggestion
Java engines SHOULD use the following for `engine.name` attribute. If none of the listed values apply, custom value MAY be used:
```

specification/resource/semantic_conventions/engine.md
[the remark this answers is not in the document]
Oberon00: I don't know if that should be a separate discussion. If we can define `engine` precise and useful, then we can discuss it separately. But I wonder if it is possible to find a good and reasonably unambiguous definition. This "engine" stuff is inherently a list of technologies that can run on top or maybe sometimes alongside each other.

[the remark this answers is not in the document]
iNikem: "list of technologies" is too vague IMO. I mean, in case of Spring MVC application using Jackson Json parser, running on Tomcat inside JVM, what will you include in that list? Everything? Some part? How to decide?

In case of java this "engine" is very simple: it is application server and similar things, a la servlet containers. It may be "arbitrary", but this is exactly the information that we want to collect because we think it is useful. 

During las specification meeting there was general agreement that this kind of information is relevant. Perhaps people from other language SIGs may provide rules specific for their language ecosystem? Or say that they are not interested in that.

And again, perfection is the enemy of the good :)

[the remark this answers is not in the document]
Oberon00: Maybe in this case it would be better to use the term `app_server` instead of `engine`? I feel like this PR should fix #1143. I see now that this issue was indeed created by the same author as this PR. 
@ivomagi please link related issues in your PR description, otherwise we might end up discussing the same things twice 😃  In this case, you can maybe even use it in the `Fixes #`  line you left empty so far.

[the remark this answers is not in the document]
Oberon00: Please also consider making this an attribute of the incoming HTTP span instead of the resource. That would be more useful for other languages too where it might not be known which "engine" is going to start at the time where telemetry is initialized. Resources are quite inflexible here.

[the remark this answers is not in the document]
iNikem: `app_server` is bad because in (our) past a lot of people very enthusiastically argued that Jetty and Tomcat are NOT application servers and MUST NOT be called as such. :)

Span attribute is a very bad idea :) This info does not change for a given JVM/process. Why would we send this with every span? Resource is the correct place for such immutable information, no?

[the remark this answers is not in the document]
ivomagi: > No, the process.runtime would describe the Python interpreter. Runtimes are always "language runtimes". We have no way to describe Apache here. Or maybe Apache would be the engine?
> What about adding (parallel) array attributes e.g. process.runtime_stack.*? Then you would have the first entry for apache, the second for mod_wsgi, the 3rd for Python (Python runs inside mod_wsgi which runs on Apache).

Thanks @Oberon00 for the feedback, appreciate it. The particular concern seems valid with the additional context you exposed. But it would be easy to solve - having the possibility to map multiple `engine` instances mapped to a resource. Would changing the PR to have multiple engines instead of single be a way to proceed in your opinion?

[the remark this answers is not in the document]
Oberon00: I think that would be a direction in which we can go. But then we need to define what order should be used inside that array. Also, for that particular case, we cannot represent that the process.runtime runs inside mod_wsgi and apache, whereas for a typical Java case, it is the other way round and the engine(s) run inside the process.runtime (JVM). Which would all be covered by runtime_stack.

It might be easier to come up with a more clear definition of what we want to have in `engine` so that only one technology is viable at any time (and then maybe a better name than `engine` if we know a more precise definition).


If that is not possible across languages, I suggest to narrow the scope and really have something like

> `java.loader`: "The thing (responsible for the classloader) inside which this OpenTelemetry instance was loaded."

(if app_server is not acceptable)

specification/resource/semantic_conventions/engine.md
```diff
@@ -0,0 +1,52 @@
...
+
+| Name | `engine.name` |
+|---|---|
+| Apache Tomcat | tomcat |
```
anuraaga, line 31: Would we set tomcat if using a spring-boot application which starts up embedded tomcat? I guess not since the user isn't even really aware of the version of tomcat, and it's sort of a spring-boot implementation detail (in such a case, it's spring-boot doing classloader trickery to e.g., load embedded JARs, and tomcat doesn't do anything).

If this case doesn't apply, then I'm not sure if `ServletContext.getServerInfo()` will be a precise check, and we may need more description about the caveats with regard to embedded container runtimes (effectively just libraries, not app servers).

specification/resource/semantic_conventions/engine.md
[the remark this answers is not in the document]
anuraaga: I think `java.loader` would work well vs my current worry about the ambiguity resulting from some of these being embeddable in an app that's loaded by something else such as spring-boot. Java app servers seem like a concept rather unique to Java.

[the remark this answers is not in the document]
iNikem: @Oberon00 @anuraaga where did this "class loading" thing come from all of a sudden? This proposal says nothing about class loading. This `java.loader` goes into completely undesired direction.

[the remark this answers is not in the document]
iNikem: > But then we need to define what order should be used inside that array

@Oberon00 do we? Why order is important for you? It is certainly not important for us in the context of this proposal.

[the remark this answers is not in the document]
Oberon00: >  This java.loader goes into completely undesired direction.

Maybe? My problem with "engine" is that I don't really know in which direction it should be going. The defining characteristic of Java app servers that I could come up with from the top of my head was that AFAIK they manage each application with an isolated class loader, so if OpenTelemetry is loaded within such an app, we could detect it easily (note: it the auto-instrumenting agent is loaded directly at JVM startup, there will be no "engine" running yet and it will be tricky to detect it at the point where resource initialization is required. That's why I suggested a span attribute)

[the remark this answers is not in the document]
Oberon00: Just writing up my thoughts here, I have not really made up my mind:

As https://github.com/open-telemetry/opentelemetry-specification/pull/1293#discussion_r545574298 @anuraaga's comment underlines for me, it is not even easy to come up with any Java-specific definition.

Which maybe brings us back to

> "list of technologies" is too vague IMO. I mean, in case of Spring MVC application using Jackson Json parser, running on Tomcat inside JVM, what will you include in that list? Everything? Some part? How to decide?

I think it is not needed to come up with an exact definition here, from a semantics perspective you can add as few or as little as you like. Of course, from a technical perspective, if the list becomes too large it might become an overhead concern. We could come up with a list of particularly interesting technologies per language, e.g. these "engine" technologies proposed here may be a good start.


> But then we need to define what order should be used inside that array
>> @Oberon00 do we? Why order is important for you? It is certainly not important for us in the context of this proposal.

Probably it isn't that important after all. Which then would make `engines` as an array and `technologies` very similar.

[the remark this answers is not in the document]
ivomagi: Back from holiday season, aiming to recover the discussion. 

I am looking to make a change allowing multiple engines to be set as resources, removing the need to change the definition of runtime (the example in PR would list Apache HTTP Server, mod_wsgi and Django all as engines). For this I am looking for advice how to change the semantic conventions. Maybe you can steer me to the right path here @Oberon00 with the following:

Engine would be represented by multiple attributes - name, version and description. Is there somewhere an example, how to encode an array of non-primitves in YAML for semconv? I only found some examples for array of arrays of primitives at best.

tigrannajaryan:
Was autoclosed due to holidays. Reopening for discussion.

specification/resource/semantic_conventions/engine.md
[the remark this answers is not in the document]
ivomagi: Discussed this with @tigrannajaryan in the background. What are your thoughts @Oberon00 and @tigrannajaryan - can/should we proceed with allowing multiple engines per span thus alleviating most of the concerns expressed in the comments? If yes, then I would be looking for means to enhance the generator to allow arrays of non-primitives set in YAML, otherwise the work in this front would not be beneficial. Thank you in advance!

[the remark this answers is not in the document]
Oberon00: > I would be looking for means to enhance the generator to allow arrays of non-primitives set in YAML

It was explicitly decided that we don't want any "map"-like structures in the data model for traces. This is not a matter of the semantic convention generator, but mainly of the (feature-frozen) attribute definition https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/common/common.md#attributes.
See also this (closed) PR: https://github.com/open-telemetry/opentelemetry-specification/pull/596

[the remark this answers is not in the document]
Oberon00: > Engine would be represented by multiple attributes - name, version and description. Is there somewhere an example, how to encode an array of non-primitves in YAML for semconv?

One way would be to have multiple arrays, e.g. `engine.names` with the names `engine.versions` with the versions such that for any index `i`, `engines.names[i]` has version `engine.versions[i]`. That would work today without changing the generator tool

Another way would be to not use arrays but instead use some special naming like `engines.1.name`, `engines.2.name`, etc. That could use support in the generator.

In any case if we need something not supported by the generator for these conventions it would IMHO be preferable to hand-write the markdown first, and if the PR is accepted we can later add support in the generator.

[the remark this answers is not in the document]
tigrannajaryan: >It was explicitly decided that we don't want any "map"-like structures in the data model for traces.

@Oberon00 I believe back then we decided that we don't want them "for now" unless there is a clear need. We may need to revisit this decision if the data at the source clearly dictates/requires more complex data structure. We may not necessarily need to do it now, but let's keep the possibility open.

@ivomagi: 

1. What is the reason we are aiming for an array of value? Do we believe that the concept that we want to express is multi-dimensional with variable number of dimensions and thus it is not possible to capture with a fixed set of attributes?
How about using the approach we use elsewhere, e.g. use attributes `engine.java.name`, `engine.java.version`. If there is a second engine dimension `foo` then its name should be in `engine.foo.name` attribute. If the reader needs to find all engines they can enumerate all `engine.*.name` attributes. 
However, this approach is only necessary if we need to record multiple engines simultaneously and I did not see examples of that in this PR. 
Finally, I assume there is never a need to record more than one name of an engine of the same dimension (i.e. you never have 2 Java engines simultaneously in one resource/span). If there ever is such a need my proposal wouldn't work.

2. It may be worth also placing the `engine` in the `process` namespace. We would then have `process.engine.java.name`. This would work unless we think there are engines that are not associated with a process (is there a such a thing?).

P.S. I am not sure "engine" is the right term to use, but I don't have a better suggestion.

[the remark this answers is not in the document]
ivomagi: @Oberon00 indeed, this workaround was what I was aiming to fall back for but wanted to verify with smarter guys whether or not we do have the "map-like" structures or not.

@tigrannajaryan 

1. The array of values is not a "must have", it was just a solution I was able to come up with. The fixed set of attributes as you refer might have some issues though. Lets look at the same example used in original pull request which in time was designed with just a single engine in mind (but it seems that the community looks at things differently and multiple engines might be required). Lets see a Python application using Apache HTTP Server with mod_wsgi as the server and Django as the web framework. Now all of these three could be set as engines and all of these seem to want to reside under `engine.python` namespace.
2. In here I do not have a bias. Maybe more eyes with more experience will share thoughts about whether the `process` namespace would be better or worse location. 

And about the term, indeed, naming stuff is hard. Open for suggestions but all the alternatives I was able to come up with (appserver, container, framework, library, ...) were either used for different purpose or felt overloaded or vague.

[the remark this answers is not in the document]
tigrannajaryan: >The array of values is not a "must have", it was just a solution I was able to come up with. The fixed set of attributes as you refer might have some issues though. Lets look at the same example used in original pull request which in time was designed with just a single engine in mind (but it seems that the community looks at things differently and multiple engines might be required). Lets see a Python application using Apache HTTP Server with mod_wsgi as the server and Django as the web framework. Now all of these three could be set as engines and all of these seem to want to reside under engine.python namespace.

We could do this:
```
apache.version=0.1.2
apache.mod_wsgi.version=1.2.3
python.django.version=2.3.4
```
I know this means we cannot enumerate engines without knowing what attributes to look for. Is it a requirement to be able to enumerate?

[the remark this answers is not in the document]
vovencij: Being able to enumerate will definitely make life at backend easier, if one would like to list all engines from a trace (which sounds like a reasonable wish). 

Having engine name as a part of a path in the tree - if an engine reports itself via its own public API which instrumentation can use like `Apache Tomcat` or `Eclipse Jetty`. How that should be encoded then? I think that having a name, which is potentially returned by 3rd party and can consist of anything is better suited to be represented as a value.

iNikem:
Honestly, I think we are making this PR harder/deeper than it should be. As both #1143 and https://github.com/open-telemetry/opentelemetry-specification/issues/1143#issuecomment-716784555 say, there is very useful debugging information that current telemetry lacks. The exact meaning of that information may be language specific and I don't think specification really has to dictate it. E.g. the aforementioned example of Django application running on Apache httpd via `mod_wsgi`. Can we let language specific auto-instrumentation to decide on exact value? This is very language specific question and thus not a specification concern.

This PR, at least in my eyes, says very simple thing: if you (application operator or auto-instrumentation) want to provide this runtime information, this is semantic attribute for it.

Which leaves two question:

1. Name of the semantic attribute
2. Is it single value or an array?

`engine` is not an ideal name and IMO it is good enough. `app_server` is loaded term in Java world and unknown in other languages. `technology` is too vague IMO, but I don't object it too much. One more option is `application runtime`.

The "single value or array" question, IMO, depends solely on our ability to have good structure for that array. I don't like two parallel arrays for names and versions. Map-like attribute would be the simplest option.

specification/resource/semantic_conventions/engine.md
iNikem: I think in case of embedded Tomcat/Jetty/Undertow we still want to report it.

tigrannajaryan:
> The "single value or array" question, IMO, depends solely on our ability to have good structure for that array. I don't like two parallel arrays for names and versions. Map-like attribute would be the simplest option.

@iNikem I agree that structurally map-like attribute would be best. The problem is that unlike OTLP most other telemetry formats are unable to represent such data. What are exporters supposed to do when they see such data? Do we want to define how map data is flattened/converted to fit the limitations of e.g. Jaeger's attributes type system?

specification/resource/semantic_conventions/engine.md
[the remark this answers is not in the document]
tigrannajaryan: @vovencij OK, if that's a requirement then what I suggested won't work. 

We either do what @Oberon00 proposed earlier with parallel arrays or with using array indices in attributes names (but neither looks nice to me) or we use an array of maps, which requires us to define how such arrays are translated to formats that are not capable of representing such data natively.

I do not have a strong opinion about what approach to use and I do not know of any better alternate.

Oberon00:
> The problem is that unlike OTLP most other telemetry formats are unable to represent such data. 

Also, the feature-frozen tracing API does not support such attributes, so it would not be possible to use such a semantic convention with current OpenTelemetry.

tigrannajaryan:
> > The problem is that unlike OTLP most other telemetry formats are unable to represent such data.
> 
> Also, the feature-frozen tracing API does not support such attributes, so it would not be possible to use such a semantic convention with current OpenTelemetry.

@Oberon00 to be fair, it is possible to extend the API in backwards compatible manner to allow this, so in theory from API's perspective it is doable (though it is more work and won't be in the GA). However, I don't know how we will solve nicely the exporting of such data to non-OTLP formats. We could double down on what the spec already suggests for homogeneous arrays and make the same recommendation for maps. To remind, here is what the spec says today:

>For protocols that do not natively support array values such values SHOULD be represented as JSON strings.

JSON obviously would also work for maps. But this is a slippery slope. We would be essentially making JSON attributes widespread and while for simple arrays it is very human-readable and probably does not require any special handling in the backend, the more complicated data structures we allow the more there will be a need for backends to be able to deal with such data as structures rather than as simple string.

iNikem:
I propose to have this attribute as single valued. The majority of useful information proposed by @Oberon00 as "technologies" above can already be obtained from span's `InstrumentationLibrary`. This also solves the format problem.

I propose to further stress that exact choice of what to report in this attribute (if any) is language specific and is NOT specification's concern.

As name I like `application_runtime`, but think that `engine` is good enough, even if not ideal.

If @Oberon00 and @tigrannajaryan agree, I can work with @ivomagi on updating this PR.

iNikem:
@Oberon00 , @tigrannajaryan do you agree with my proposal above?

tigrannajaryan:
> @Oberon00 , @tigrannajaryan do you agree with my proposal above?

@iNikem I do not understand the problem well enough to agree. I do not object with going forward with the approach that makes best sense to you. Feel free to disregard my suggestions. I was only trying to show some alternates that could potentially help, but apparently they are not a good git.

iNikem (approved):

anuraaga (approved):

semantic_conventions/resource/engine.yaml
Oberon00: I notice this was marked as resolved but not applied. Intentional?

specification/resource/semantic_conventions/engine.md
```diff
@@ -0,0 +1,22 @@
...
+
+<!-- semconv engine_resource -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `engine.name` | string | The name of the engine. | `FildFly` | Yes |
+| `engine.version` | string | The version of the engine, as returned by the engine (preferably during runtime) without modification. | `21.0.0` | No |
+| `engine.description` | string | Additional description about the engine. | `WildFly Full 21.0.0.Final (WildFly Core 13.0.1.Final) - 2.2.2.Final` | No |
+<!-- endsemconv -->
+
+Information describing the engine SHOULD be captured using the values acquired from the API provided by the engine, preferably during runtime, to avoid maintenance burden on engine version upgrades. As an example - Java engines are often but not always packaged as application servers. For Java application servers supporting Servlet API the required information MAY be captured by invoking `ServletContext.getServerInfo()` during runtime and parsing the result.
```
Oberon00, line 15: Is this a practical suggestion? Resources must be computed before the SDK is initialized (at least in most languages I think).

specification/resource/semantic_conventions/engine.md
```diff
@@ -0,0 +1,22 @@
...
+
+A resource can be attributed to at most one engine.
+
+The situations where there are multiple candidates, it is up to instrumentation library authors to choose the engine. To illustrate, let's look at a Python application using Apache HTTP Server with mod_wsgi as the server and Django as the web framework. In this situation:
+
+* Either Apache HTTP Server or `mod_wsgi` MAY be chosen as `engine`, depending on the decision made by the instrumentation authors.
```
Oberon00, line 21: According to the definition above, neither can be used as engine, since the runtime will be Python and neither Apache nor mod_wsgi are executed using Python.

ivomagi: Relaxed the definition to cover for this (Engines are **typically** executed using process.runtime).

specification/resource/semantic_conventions/engine.md
ivomagi: I really see the benefit of runtime capture - the work needed in manually setting this is tedious and likely to introduce additional maintenance burden. But indeed, up until https://github.com/open-telemetry/opentelemetry-specification/issues/1298 is resolved, the instrumentation library providers cannot do this in runtime.

Oberon00:
> I propose to further stress that exact choice of what to report in this attribute (if any) is language specific and is NOT specification's concern.
>
> As name I like application_runtime, but think that engine is good enough, even if not ideal.

I'm fine with a single-valued attribute. I'm also fine with not specifying exactly what to report as engine. But when we do both, I think we have so much ambiguity in this attribute that I wonder how useful it will be in practice. It seems to me that Java application-server-like engines might be the only thing that were really considered during the design, so it might not work that well for other languages. Ideally I'd like to see some thoughts from other SIG's members (e.g., @open-telemetry/python-approvers or @open-telemetry/dotnet-approvers) if and how this attribute seems usable in their languages.

iNikem:
@Oberon00 I can agree with you. This PR was created based on the specific need described in #1143. It seems both the problem and the solution did not get enough traction in other SIGs. IMO that does not make the original problem irrelevant or this solution unsuitable :) Will be happy hear other opinions from other SIGs.

Oberon00:
> IMO that does not make the original problem irrelevant or this solution unsuitable

Agreed. I was just wondering if the name should be under some Java namespace, but I guess we can use `engine` (even though that word makes me think of game engines).

CodeBlanch:
> Ideally I'd like to see some thoughts from other SIG's members (e.g., @open-telemetry/python-approvers or @open-telemetry/dotnet-approvers) if and how this attribute seems usable in their languages.

Not totally my area of expertise, but here's my perspective for .NET:

I don't really know of an easy way to get the name of the webserver being used. It is kind of abstracted and hidden away. The version could be tricky too. For .NET Framework IIS does pass some variables on the request, but that won't be available on startup. Might be able to look at the hosting process to figure it out. OWIN, not sure. In a [self-hosted](https://docs.microsoft.com/en-us/aspnet/aspnet/overview/owin-and-katana/getting-started-with-owin-and-katana#self-host-owin-in-a-console-application) scenario user code owns all of the OWIN config & port binding. In .NET Core, it is slightly easier. You can ask for [IServer](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.hosting.server.iserver) from the app dependency container and then inspect the implementation type. I think we would have to hardcode the types. Like `if (server.GetType().Name.Contains("Kestrel")) AddKestrel();` which isn't ideal. I also don't think it's providing a whole lot of information because .NET Core is most likely always Kestrel(?). There might be a reverse-proxy in front of Kestrel (IIS, Nginx, Apache) which we would be more interesting information, but we can't get at that I don't think.

IMO for .NET you really care most about the runtime version (.NET Framework 4.8, .NET Core 3.1, .NET 5, etc.).

So I like the feature but not totally a slam dunk for .NET 🏀 

@cijothomas Does AppInsights solve this somehow?

iNikem:
> IMO for .NET you really care most about the runtime version (.NET Framework 4.8, .NET Core 3.1, .NET 5, etc.).

You can pass this information in `engine` if that makes sense for your language :)

iNikem:
@Oberon00 seems like other SIGs don't care :) Is there a chance to get your approval? Then I can start pestering TC about merging this PR :)

semantic_conventions/resource/engine.yaml
```diff
@@ -0,0 +1,22 @@
+groups:
+  - id: engine_resource
+    prefix: engine
+    brief: >
+       Resource describing the packaged software running the application code. Engines are executed using process.runtime.
```
Oberon00, line 5:
```suggestion
       Resource describing the packaged software running the application code. Engines are typically executed using process.runtime.
```
see https://github.com/open-telemetry/opentelemetry-specification/pull/1293/files#r567805983

semantic_conventions/resource/engine.yaml
```diff
@@ -0,0 +1,22 @@
...
+      - id: name
+        type: string
+        required: always
+        brief: >
+          The name of the engine.
+        examples: ['WildFly']
+      - id: version
+        type: string
+        brief: >
+          The version of the engine, as returned by the engine (preferably during runtime) without modification.
```
Oberon00, line 16:
```suggestion
          The version of the engine.
```

I think the remainder is quite vague.

semantic_conventions/resource/engine.yaml
```diff
@@ -0,0 +1,22 @@
...
+        examples: ['21.0.0']
+      - id: description
+        type: string
+        brief: >
+          Additional description about the engine.
```
Oberon00, line 21:
```suggestion
          Additional description of the engine (e.g. detailed version and edition information).
```

Oberon00 (approved):
I'm not really happy with the name `engine` but have no better suggestion, and if I'm the only one complaining then it's probably fine 😃

iNikem:
I have only one more option: `application_runtime` :)

iNikem:
@open-telemetry/technical-committee Asking for a merge :)

specification/resource/semantic_conventions/engine.md
```diff
@@ -0,0 +1,22 @@
...
+* Django would SHOULD NOT be set as an engine as the required information is already available in insrumentation library and setting this into `engine` would duplicate the information.
```
tigrannajaryan, line 22:
```suggestion
* Django SHOULD NOT be set as an engine as the required information is already available in insrumentation library and setting this into `engine` would duplicate the information.
```

tigrannajaryan (approved):

tigrannajaryan:
Need to wait one more day (the rule is: 2 days since last substantial change).

iNikem:
@open-telemetry/technical-committee More than one day has passed :) Can we now merge?

specification/resource/semantic_conventions/engine.md
```diff
@@ -0,0 +1,22 @@
+# Engine
+
+**type:** `engine`
+
+**Description:** Resource describing the packaged software running the application code. Engines are typically executed using process.runtime.
```
yurishkuro, line 5: > Engines are typically executed using process.runtime.

I don't understand what this means.

Oberon00: For example, WildFly is executed using a JVM.

yurishkuro: I understand that an "engine" is (from examples), I am referring to the wording "using process.runtime". Is it referring to the resource attribute?

yurishkuro: in other words, I don't think this wording is helping or explaining it well.

Oberon00: Yes, I think so.

iNikem: @yurishkuro Do you think should block this PR from merging or can this be remedied in a follow up PR? Can you suggest a better wording?

yurishkuro: Have you considered scoping it more narrowly, like `jvm.app_server` instead of `engine`? What are the other examples of "engines"?

Oberon00: This was indeed considered (I would also be in favor): https://github.com/open-telemetry/opentelemetry-specification/pull/1293#discussion_r544579642

iNikem: Honestly, it is very disheartening to have a long round of discussions, reach an agreement with all active participants, have approvals from them, waiting for procedural delay of 2 days and then after 5 days receive a "have you considered this other solution" from a new participant :(

@yurishkuro do you want to block this PR from merging? If yes, please request changes and block. If not, I would kindly ask to merge it (enough number of approvals and grace period passed). Please don't drag this PR back into uncertain state of limbo.

yurishkuro: Reading some of the past comments now, I don't see where any consensus was reached. Even some of the approvals were done "under duress". There were a number of problems raised for non-Java languages, and none of them were resolved, almost none of the other SIGs responded saying that it makes sense to them. If you're only interested in solving this for Java, then the convention should be scoped to Java.

yurishkuro: For example, in the [runtime](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/process.md#process-runtimes) spec, there are already sub-sections that make sense for specific languages, like `process.runtime.dotnet-core`, or `process.runtime.nodejs` - the latter actually sounds very much like this "engine" proposal.

If you believe the engine applies more generally, then please include examples for several languages in the description.

Oberon00: > there are already sub-sections that make sense for specific languages, like process.runtime.dotnet-core, or process.runtime.nodejs

There is no `process.runtime.nodejs` semantic convention. These subsection are just very outdated (maybe we should bug @open-telemetry/javascript-approvers and @open-telemetry/dotnet-approvers to update them)? What is actually meant is that you should one of the values in that table as `process.runtime.name`. E.g. you would use `process.runtime.name=browser` (and `telemetry.sdk.language=webjs`) to indicate Javascript in the browser.

iNikem: > almost none of the other SIGs responded saying that it makes sense to them

My take on that is that none of other SIGs objected to this.

> I don't see where any consensus was reached

I thought that approval means agreement. Am I wrong?

@yurishkuro And I beg you pardon, but you haven't answered my question: do you want to block this PR from merging?

specification/resource/semantic_conventions/engine.md
```diff
@@ -0,0 +1,22 @@
...
+| `engine.name` | string | The name of the engine. | `WildFly` | Yes |
+| `engine.version` | string | The version of the engine. | `21.0.0` | No |
+| `engine.description` | string | Additional description of the engine (e.g. detailed version and edition information). | `WildFly Full 21.0.0.Final (WildFly Core 13.0.1.Final) - 2.2.2.Final` | No |
+<!-- endsemconv -->
+
+Information describing the engine SHOULD be captured using the values acquired from the API provided by the engine, preferably during runtime, to avoid maintenance burden on engine version upgrades. As an example - Java engines are often but not always packaged as application servers. For Java application servers supporting Servlet API the required information MAY be captured by invoking `ServletContext.getServerInfo()` during runtime and parsing the result.
+
+A resource can be attributed to at most one engine.
+
+The situations where there are multiple candidates, it is up to instrumentation library authors to choose the engine. To illustrate, let's look at a Python application using Apache HTTP Server with mod_wsgi as the server and Django as the web framework. In this situation:
```
yurishkuro, line 19: > it is up to instrumentation library authors to choose the engine

This is incorrect. In the example below the "instrumentation library" will be only instrumenting Django framework, regardless of where it is run standalone or via wsgi. The "enging" is a part of the resource definition, so only the SDK can populate it.

specification/resource/semantic_conventions/engine.md
yurishkuro: I am not in the habit of exercising veto power. I have voiced my concerns with the wording.

iNikem: So you actually don't care what will happen with this PR?

specification/resource/semantic_conventions/engine.md
ivomagi: There is an ongoing discussion at #1298 in order to allow resources to be added post-creation. But indeed, up until this is allowed, you are correct. It seems we have a dependency for that matter and I agree to wait with the merge until the #1298 gets a resolution.

SergeyKanzhelev:
I think two unresolved comments in this PR are:

- wording describing the field is not clear enough for unprepared reader
- give examples on how it is used in other languages, outside Java or consider making these attributes java-specific

specification/resource/semantic_conventions/engine.md
rakyll: I think having a conversation about naming is important, semantic conventions are about naming in the end of the day. "Engine" might not be the best name in this case. Is it standing for a runtime, a platform, an execution environment? It matters to have this conversation upfront because the naming sets the scope and sometimes affects the other specs. For example, if "engine" stands for a platform/runtime, we have to iterate on platform specific specs like Lambda, etc.

bogdandrutu (changes requested):
To please @iNikem (you were asking for someone to block this PR), I read this PR (first of all really sorry for the delay, we had a deadline last week for the specification and lot of the efforts were focused there) and decided that this is not ready to be merged:
1. From the content of the PR (description, definition of new entries) it is not clear how this applies to other languages than Java. Because of this I think we may do a mistake merging this PR now without proving that it is generic enough for all the cases and we don't cause troubles in the future.

I can see as options:
1. We do the due diligence and understand if this is a good enough. Pick at least 2 other languages that have the same concept and show how this design can be applied in that case.
2. Understand that this is a java specific only problem and make things java specific only.

carlosalberto:
> Understand that this is a java specific only problem and make things java specific only.

I'd vow to go for this, which seems to be the case all along (that would include updating the actual semantic convention names, to make them Java-specific).

carlosalberto:
Ping @ivomagi

iNikem:
@carlosalberto 
First, I already have one more use-case for this type of data. There is a [PR to add Otel integration with Nginx](https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/5). It will then be useful to report that this monitored process is Nginx version XYZ. Proposed `engine` was meant exactly for such use-case.

Second, we talked to @ivomagi  and he will reach out explicitly to a couple of language SIG and will get then answer if they see any benefit if this propose semantic convention.

carlosalberto:
@iNikem Fantastic, thanks for the follow up, will be waiting :)

ivomagi:
> @iNikem Fantastic, thanks for the follow up, will be waiting :)
Submitted the question to three SDK repositories, asking for feedback. Choose .NET, Python and PHP:

- https://github.com/open-telemetry/opentelemetry-dotnet/issues/1839
- https://github.com/open-telemetry/opentelemetry-python/issues/1617
- https://github.com/open-telemetry/opentelemetry-php/issues/254

ivomagi:
@carlosalberto and @bogdandrutu - it seems we got feedback from PHP community who would be using this to distinguish between mod_apache and php-fpm deployments and .NET who seems to find the use for determining the version of the runtime. Python community has not responded. So it kinda seems there is a wider use for the feature outside the JVM world as well so can we proceed?

Oberon00:
> .NET who seems to find the use for determining the version of the runtime.

But `process.runtime.version` already exists for that.

Oberon00:
Yeah, I did not understand that. That should be part of `process.runtime.description`. If .NET uses engine to describe a certain aspect of the runtime, then it should rather use it's own attribute like `dotnet.runtime_variant` or whatever.

iNikem:
> then it should rather use it's own attribute like `dotnet.runtime_variant` or whatever.

why? What is that dotnet specific attribute is better than `engine` proposed here?

Oberon00:
The `engine` proposed here is vaguely defined, but about the only thing that is clear to me is that it describes something that is *not* the runtime but something that runs inside it or something that the runtime runs inside. The .NET runtime variant is just additional information about the same runtime described by process.runtime.

owais:
Python could use this to differentiate between web servers like uWSGI, Gunicorn, etc but I don't think `engine` is a good fit at all for this. Something like appserver/webserver would be a much better fit for Python at least.

carlosalberto:
Summarizing, it sounds like this is something useful to a few languages, but we need to find consensus on the name (and maybe clarify things a little bit more). @owais @Oberon00 any suggestions for alternative names?

owais:
I can't think of anything that'd be a better fit for Python at least other than `appserver` or `webserver`. That said, I don't know how useful this addition would be specifically for Python or something like Ruby. These are WSGI servers and generally are instrumented  as a result so not sure if this proposal adds any value for Python.

carlosalberto:
@ivomagi Would using `webserver`, as suggested, would be a good compromise? Else, what about `webservice_engine` or `web_engine`?

ivomagi:
`engine` was chosen to avoid overloaded terms such as `webserver` or `appserver`. But honestly, I care the most of having this possibility, vs the naming debate. So if @carlosalberto you or anyone else has a strong opinion on what the naming convention should be, I am on board with continuing with an alternative.

carlosalberto:
@ivomagi I suggest we go with `webengine`, as a) it does not overload other common names such as `webserver` and b) blends the original `engine` with `web`, which is (more or less) what was originally. @owais @iNikem @Oberon00 please comment on whether you like this or not.

Oberon00:
Webengine is the name of some libraries that render HTML (e.g. Qt). I would still OK with it, but I would like webserver better.

> engine was chosen to avoid overloaded terms such as webserver or appserver

But isn't engine even more overloaded? It seems to me that webserver is exactly what this attribute seems to want to describe?

iNikem:
> But isn't engine even more overloaded? It seems to me that webserver is exactly what this attribute seems to want to describe?

Would you call WebLogic server a "webserver"?

Oberon00:
In the sense of this attribute, yes, close enough 😃

carlosalberto:
All right, let's go with `webengine` then, and try to get this PR finally merged :) @ivomagi please update this one.

ivomagi:
@carlosalberto @bogdandrutu - updated the PR to webengine. There seems to be one requested change from Bogdan, other than this we should be good to merge?

carlosalberto:
I think we can dismiss @bogdandrutu's review as we the diligence was done (let me know otherwise).

ivomagi:
> I think we can dismiss @bogdandrutu's review as we the diligence was done (let me know otherwise).

Indeed, it seems we captured this feedback, but the review requesting changes is still out there and blocking the merge, so how should we proceed - @carlosalberto @bogdandrutu

carlosalberto:
@bogdandrutu Dismissed your review - let me know if you think something needs to be addressed still, and I will work on a follow up ;)

PR #1294, opened 2020-12-16
Clarify/relax required resource attributes.

### LINKED ISSUE #1346: Update default resource semantics (formerly required service.name).

The current wording for required service.name resource attribute is cumbersome to implement and may be interpreted in a way that makes creating a truly empty resource without service.name empty.

**Additional context.**

This is a tracking issue for the following PRs:

- #1345 "Make Resource.Merge consistent with Span.SetAttribute." (per-requesite, as discussed in spec meeting)
- #1294 "Clarify/relax required resource attributes." (actual solution)

I also gave this the spec:trace label, because this spec is mostly for the Zipkin/Jaeger span exporters.

Oberon00:
Fixes #1346.

Depends on #1345 ✔️ .

## Changes

Clarify that it is enough to add required resource attributes ~~when the resource is associated with something~~ to the default resource. We do not need to somehow patch the attribute on every (temporary) resource object created by the user.

Related Java issues (see linked PRs there): open-telemetry/opentelemetry-java#2298

EDIT on 2020-12-19: I forgot to update this PR description to match the changes from 1698fd0  (2020-12-17). Done now.


specification/resource/sdk.md
```diff
@@ -24,6 +24,15 @@ When associated with a [`MeterProvider`](../metrics/api.md#meter-interface),
...
 associated with this `Resource`.
 
+
+Certain **required** `Resource` attributes MUST be set to a default value if they were not specified when
+the Resource is associate with a `TracerProvider` or `MeterProvider`
+(this will result in a new `Resource` being associated, as the original `Resource` is immutable).
+See [Attributes with Default Value](semantic_conventions/README.md#attributes-with-default-value).
+
+Note that the OpenTelemetry project documents certain ["standard
+attributes"](semantic_conventions/README.md) that have prescribed semantic meanings.
```
yurishkuro, line 34: * "OpenTelemetry project documents" - Yuri declares that referring to oneself in the third person is weird
* I am not sure what this sentence is trying to say, specifically in the context of the previous paragraph.

Oberon00: I can remove the paragraph entirely if you like. I think it is just a note to make you aware that semantic conventions exist for resources. Do you think it fit better in the previous location?

Oberon00: "semantic meaning" is also nice

specification/resource/sdk.md
```diff
@@ -24,6 +24,13 @@ When associated with a [`MeterProvider`](../metrics/api.md#meter-interface),
 all metrics produced by any `Meter` from the provider will be
 associated with this `Resource`.
 
+Certain **required** `Resource` attributes MUST be set to a default value if they were not specified when
+the Resource is associate with a `TracerProvider` or `MeterProvider`
+(this will result in a new `Resource` being associated, as the original `Resource` is immutable).
```
yurishkuro, line 29: The implication is that each provider would be responsible for setting those required attributes to default values. Should there be some framework for enriching resources that providers can share, to avoid duplication?

Also, do we document default fallback values for all required attributes? Service name alone too a number of iterations.

Oberon00:
Something was weird here with the auto-requested reviews (there seem to have been none?) so I manually requested reviews now.

specification/resource/sdk.md
```diff
@@ -24,6 +24,13 @@ When associated with a [`MeterProvider`](../metrics/api.md#meter-interface),
 all metrics produced by any `Meter` from the provider will be
 associated with this `Resource`.
 
+Certain **required** `Resource` attributes MUST be set to a default value if they were not specified when
+the Resource is associate with a `TracerProvider` or `MeterProvider`
```
jkwatson, line 28:
```suggestion
the Resource is associated with a `TracerProvider` or `MeterProvider`
```

jkwatson:
This is a good clarification. I think it should also be highlighted in the referenced document [`semantic_conventions/README.md#attributes-with-default-value)`]

Oberon00:
Since I have no approvals yet, I reworked this again to not always force the attributes into the Resource. If this is to lenient, I can go back to requiring setting the attributes at time of association with TracerProvider/MeterProvider (see f65800a).

yurishkuro (dismissed):

jkwatson (approved):
thanks! much cleaner/clearer than what was previously specified.

anuraaga (approved):
Thanks!

specification/resource/semantic_conventions/README.md
```diff
@@ -33,9 +34,10 @@ Attributes are grouped logically by the type of the concept that they described.
 
 Certain attribute groups in this document have a **Required** column. For these groups if any attribute from the particular group is present in the Resource then all attributes that are marked as Required MUST be also present in the Resource. However it is also valid if the entire attribute group is omitted (i.e. none of the attributes from the particular group are present even though some of them are marked as Required in this document).
 
-## Attributes with Default Value
+## Semantic Attributes with SDK-provided Default Value
 
-The SDK MUST set a default value for the following attributes if they were not specified, even if the resource would otherwise not be present at all:
+These are the the attributes which MUST be provided by the SDK
+as specified in the [Resource SDK specification](../sdk.md#):
```
arminru, line 40:
```suggestion
as specified in the [Resource SDK specification](../sdk.md#sdk-provided-resource-attributes):
```

arminru (approved):
Thanks! Please don't forget about the changelog 🙂

Oberon00:
Done. BTW the new/updated distinction seems to be lost currently.

jmacd (approved):

specification/resource/sdk.md
```diff
@@ -1,10 +1,12 @@
 # Resource SDK
 
 A [Resource](../overview.md#resources) is an immutable representation of the entity producing
-telemetry. For example, a process producing telemetry that is running in a
+telemetry as [Attributes](../common/common.md#attributes).
```
yurishkuro, line 4: A Resource is a description of the entity producing telemetry, represented as an immutable collection of Attributes.

Oberon00: Attributes are defined as a collection of key-value pairs already.

yurishkuro: my intent was to separate the semantic meaning of Resource (description of the entity producing telemetry) from its implementation (immutable collections of Attributes). Just "immutable Attributes" doesn't sound right.

specification/resource/sdk.md
```diff
@@ -24,6 +26,21 @@ When associated with a [`MeterProvider`](../metrics/api.md#meter-interface),
...
+This resource MUST be associated with a `TracerProvider` or `MeterProvider`
+if another resource was not explicitly specified.
+
+Note: This means that it is possible to create and associate a resource that
+does not have all or any of the SDK-provided attributes present. However, that
+does not happen by default. If a user wants to combine custom attributes with
+the default resource, they can use [`Merge`](#merge) with their custom resource
+or specify their attributes by implementing
+[Custom resource detectors](#detecting-resource-information-from-the-environment)
+instead of explicitly associating a resource.
```
jkwatson, line 42: I don't understand what this means. It sounds like the user can get rid of the mandatory `service.name`, and if I read this correctly, it will always get rid of the mandatory `service.name` unless the user knows exactly what they are doing. I think this needs to have some very explicit examples provided about what the SDK is supposed to do, as I can't make heads nor tails of it any more.

Oberon00: Ignoring this note, the important part is the paragraph above:

> The SDK MUST provide access to a Resource with at least the attributes listed at Semantic Attributes with SDK-provided Default Value.

For example, in Java this could look like a function used like` Resource defaultResource = Resource.getDefault()`.

> This resource MUST be associated with a `TracerProvider` or `MeterProvider` if another resource was not explicitly specified.

For example, if I do `TracerProvider tracerProvider = SdkTracerProvider.builder().build()`, then this `tracerProvider` MUST have the default resource associated as if by `SdkTracerProvider.builder().setResource(Resource.getDefault()).build()`.

And that's all the spec requires. This note elaborates the consequences of this, namely if you do:
`TracerProvider tracerProvider = SdkTracerProvider.builder().setResource(Resource.getEmpty()).build()` then the `tracerProvider`  did have another resource explicitly specified, so it will really have that empty resource without `service.name`. If a user want's to *combine* their custom attributes with the default resource instead of replacing it, they can use `Merge`, i.e. `SdkTracerProvider.builder().setResource(Resource.create(...).merge(Resource.getDefault())).build()`, or they do not call `setResource` explicitly but put a `ResourceProvider` SPI on the classpath that supplies the custom additional values (in Java terms).

jkwatson: So, then we're back where we started, and we don't know how the exporters are supposed to behave if they get a resource without a `service.name` in it. This seems to defeat the entire purpose of making the `service.name` absolutely mandatory.

Oberon00: Hmm, exporters that need it could merge the default resource in themselves, or query just that attribute. The advantage over always setting would be that exporters which don't need it don't have the attribute forced on them so strongly.

specification/resource/sdk.md
jmacd: I agree with your goal @yurishkuro but I think it belongs in another PR.

specification/resource/sdk.md
jmacd: @jkwatson I don't want an exporter to do what's suggested here. Exporters should just export what they receive as per the library guidelines (part #4 of https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/library-guidelines.md#requirements).

I think the two paragraphs above are :100: and in OTel-Go we have done this already: `sdk/resource.New()` takes a list of `Detector` objects and other options that enable configuring a resource without the obligatory attributes set, but it's not advised and not a default as @Oberon00 suggests.

jmacd (approved):

specification/resource/sdk.md
Oberon00: So I see two options to continue here:

1. Keep the PR as-is, but add wording to exporter's specs how exporters should behave if they receive a resource without their required attributes (fall back to fetch only these from the default resource would be my suggestion).
2. Open a new PR that proposes the state of this PR before 1698fd0, i.e. force the required SDK-provided attributes to always be present on all resources associated with a Tracer/MeterProvider.

jkwatson: If we make the exporter examine the Resource, it will have to do this for every single span that is sent in to it, which seems incredibly inefficient. I don't think that's a good pattern.

Oberon00: The exporter needs to examine every resource anyway if it wants to send it (which Zipkin doesn't do AFAIK).

Even if the exporter relied on the service.name always being in the resource, would that mean that we allow different service.name for different spans?

jkwatson: > @jkwatson I don't want an exporter to do what's suggested here. Exporters should just export what they receive as per the library guidelines (part #4 of https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/library-guidelines.md#requirements).

Yes, 100% on this statement.

> 
> I think the two paragraphs above are 💯 and in OTel-Go we have done this already: `sdk/resource.New()` takes a list of `Detector` objects and other options that enable configuring a resource without the obligatory attributes set, but it's not advised and not a default as @Oberon00 suggests.

The issue is that Jaeger and Zipkin absolutely require a service name to work (as I understand it). So... again... how should these exporters behave if there isn't a service.name on the Resource associated with spans? This was entirely the purpose of making the service.name mandatory.. to make it so the exporters can be confident that they will always have something and not have to be configured with a fallback service name one-by-one. This PR undoes that popular change, which leaves us back without an answer of how the exporters are supposed to be have.

Oberon00: > This PR undoes that popular change, which leaves us back without an answer of how the exporters are supposed to behave.

We still have a well-defined default service name, and a way to access it (via the default resource to which the SDK MUST provide access). We can easily add wording to use that if the exporter requires it but does not find it.

Oberon00: Looking at https://github.com/open-telemetry/opentelemetry-java/blob/f066003484451bb7a50e6c66b639cebd951c0188/exporters/zipkin/src/main/java/io/opentelemetry/exporter/zipkin/ZipkinSpanExporter.java#L167-L171 it would be trivial and just as efficient to replace the explicit per-exporter configuration with an access to the default resource's service.name. The logic to use the resource's service.name if present instead is already there.

Oberon00: For Jaeger (and also OTLP and conceptually), the service name and all other resources are only sent once per batch of spans, so the exporter already has to somehow sort the spans by associated resource if we really allow different resources on the same exporter (which I personally think we should not but it might take effort to prevent). E.g. Java does

https://github.com/open-telemetry/opentelemetry-java/blob/25f1f54ddad6886a046d7d78ea2b6b4c71f9094b/exporters/jaeger/src/main/java/io/opentelemetry/exporter/jaeger/JaegerGrpcSpanExporter.java#L118-L120

```java
    spans.stream()
        .collect(Collectors.groupingBy(SpanData::getResource))
        .forEach((resource, spanData) -> requests.add(buildRequest(resource, spanData)));
```

jkwatson:
To summarize from today's spec SIG discussion:

We have several options here:

1) Make all Resources contain a fallback `service.name` if they don't have one explicitly set when created (this is as-written in the spec today).
2) Make the SDK assign a fallback `service.name` to any Resource that is attached to an SDK *Provider if it is missing that.
3) Write the exporter specification to use a fallback `service.name` if the Resource attached to spans is missing that attributes.

As sub-options for option 3), we could
a) Have the Resource semantic conventions contain an optional default value, and expose that to users, and hence exporters.
b) Require the SDK expose a "fallback" Resource for exporters to use when the Resource is missing a required attribute (the way this PR is currently written).

In any case..the core issue here is that the exporters (especially Jaeger) are currently woefully underspecified.

Oberon00:
Thank you for doing the summary @jkwatson!

I would prefer option 2 (which is specified before 1698fd0 in this PR) or 3b (this PR as it is now). I think 1 may be a hassle to implement and may lead to surprising behavior if you use Resources in a slightly non-default way. As I understand, 3a would require a change to the semantic convention generator tool and to all templates, and I doubt we will often need that feature.

What I like about 3b (let the exporter fetch the default) over 2 (let the SDK put the default in) is that we don't lose the information on whether the service.name is explicitly set or not. Also, exporters that don't need the service.name, don't need to send it.

bogdandrutu:
@anuraaga @yurishkuro @arminru @jmacd I re-asked review because this PR changed since your review approval and want to make sure we are agreeing on what we merge.

Oberon00:
@bogdandrutu This is not really true. The only change done after approvals was to add a CHANGELOG entry. However, the PR description was still reflecting the old status before the changes.

jonatan-ivanov (commented):
I'm not sure this is the right place for this question but why this property is called `service.name`? Should not this be a little more general, e.g.: `application.name` (e.g.: in case of a thick client)?

jmacd (approved):

anuraaga (approved):

carlosalberto:
Not a strong feeling, but I'd vow for 3a as it's the simplest one (and should work _just fine_).

carlosalberto (approved):

carlosalberto:
@jonatan-ivanov In order to not lose focus here, would you mind opening an issue with this question? Or even better, come to the Spec SIG meeting on Tuesday, 8AM Pacific ;)

carlosalberto:
@open-telemetry/specs-approvers Please review the latest iteration. Once we get 1-2 more reviews we are good to go (usually we need less reviews, but this being a tricky one, I'd rather have more eyes on it). Else, we will merge in the next days ;)

specification/resource/sdk.md
```diff
@@ -24,6 +26,21 @@ When associated with a [`MeterProvider`](../metrics/api.md#meter-interface),
 all metrics produced by any `Meter` from the provider will be
 associated with this `Resource`.
 
+## SDK-provided resource attributes
+
+The SDK MUST provide access to a Resource with at least the attributes listed at
+[Semantic Attributes with SDK-provided Default Value](semantic_conventions/README.md#semantic-attributes-with-sdk-provided-default-value).
+This resource MUST be associated with a `TracerProvider` or `MeterProvider`
+if another resource was not explicitly specified.
```
yurishkuro, line 34: this is not clear to me.

> The SDK MUST provide access to a Resource

provide access to whom? because later it says

> This resource MUST be associated with a `TracerProvider` or `MeterProvider`

associated by whom?

Are we saying that SDK must expose an API like `getDefaultResource()` that the user MUST call and then pass to provider? What if another resource is passed, not the default one, and thus the required attributes are still not available?

Oberon00: >> The SDK MUST provide access to a Resource
> provide access to whom? because later it says

As a public API.

>> This resource MUST be associated with a TracerProvider or MeterProvider
> associated by whom?

By the Tracer/MeterProvider implemenentation. For that, a public API would not be necessary, but it does not hurt that case and is very useful if exporters want to access it, e.g. to get a fallback `service.name`.

> Are we saying that SDK must expose an API like getDefaultResource()

Yes

> that the user MUST call and then pass to provider?

No, we do not place MUST requirements on the user here. See my above answer in this comment.

>  What if another resource is passed, not the default one, and thus the required attributes are still not available?

Then whoever needs these attributes needs to fall back to the default resource. This could be done e.g. by `explicitlyPassedResource.Merge(defaultResource).get(theAttributeNeeded)`.

Oberon00:
As discussed in the spec meeting: If we change the order of arguments of Merge (i.e. last value wins like everywhere else in OTel), this PR would offer better UX.

anuraaga:
`last value wins`++! I always have to double-take when using `Resource.merge`

Oberon00:
I created a separate PR for the Resource.Merge change: #1345.

carlosalberto:
@open-telemetry/specs-approvers I will merge this by EOD (FYI ;) )

arminru (approved):

PR #1296, opened 2020-12-17
Introduce resource attribute `faas.max_memory`

z1c0:
Fixes #

## Changes

This PR introduces the new attribute `max_memory` to the `faas` resource attributes. Maximum memory allocated to a serverless function is a useful metric since e.g. too little memory can easily stop a Java AWS Lambda function from working correctly.

### DIFF
--- a/semantic_conventions/resource/faas.yaml
+++ b/semantic_conventions/resource/faas.yaml
@@ -39,10 +39,8 @@ groups:
         brief: >
           The amount of memory available to the serverless function in MiB.
         note: >
+          It's recommended to set this attribute in since e.g. too little memory can easily
+          stop a Java AWS Lambda function from working correctly.
           On AWS Lambda, the environment variable `AWS_LAMBDA_FUNCTION_MEMORY_SIZE`
           provides this information.
-        required:
-          conditional: >
-            Required for executions on serverless environments given that the
-            respective environment provides this information.
         examples: 128
--- a/semantic_conventions/resource/faas.yaml
+++ b/semantic_conventions/resource/faas.yaml
@@ -37,7 +37,7 @@ groups:
       - id: max_memory
         type: number
         brief: >
-          The amount of memory available to the serverless function in MB.
+          The amount of memory available to the serverless function in MiB.
         note: >
           On AWS Lambda, the environment variable `AWS_LAMBDA_FUNCTION_MEMORY_SIZE`
           provides this information.
--- a/semantic_conventions/resource/faas.yaml
+++ b/semantic_conventions/resource/faas.yaml
@@ -34,3 +34,15 @@ groups:
         brief: >
           The execution environment ID as a string.
         examples: ['my-function:instance-0001']
+      - id: max_memory
+        type: number
+        brief: >
+          The amount of memory available to the serverless function in MB.
+        note: >
+          On AWS Lambda, the environment variable `AWS_LAMBDA_FUNCTION_MEMORY_SIZE`
+          provides this information.
+        required:
+          conditional: >
+            Required for executions on serverless environments given that the
+            respective environment provides this information.
+        examples: 128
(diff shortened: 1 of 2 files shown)


arminru (approved):
Thanks, Wolfgang!

semantic_conventions/resource/faas.yaml
```diff
@@ -34,3 +34,15 @@ groups:
         brief: >
           The execution environment ID as a string.
         examples: ['my-function:instance-0001']
+      - id: max_memory
+        type: number
+        brief: >
+          The amount of memory available to the serverless function in MB.
```
Oberon00, line 40: I think we should use bytes here. AWS probably uses MiB (1024 based) instead of MB (1000 based), but using bytes would avoid such confusion and may be more exact for other FaaS environments.

I think we also use bytes instead of a derived unit for all other byte-related attributes (like message size, HTTP body size).

Even if we decide to keep using Mega- then I would prefer writing `MiB` as it is more clear that it is 1024-based (even though that is expected anyway for everything but filesystem/storage related stuff).

semantic_conventions/resource/faas.yaml
```diff
@@ -34,3 +34,15 @@ groups:
...
+        note: >
+          On AWS Lambda, the environment variable `AWS_LAMBDA_FUNCTION_MEMORY_SIZE`
+          provides this information.
+        required:
+          conditional: >
+            Required for executions on serverless environments given that the
+            respective environment provides this information.
```
arminru, line 47: Making it required should not be a big deal but I think stating it's recommended and also providing the reasoning you mention in the PR would be enough.

semantic_conventions/resource/faas.yaml
z1c0: Bytes is the wrong granularity here.
The suggestion to use MiB is legit though. I will change that.

semantic_conventions/resource/faas.yaml
z1c0: I'm not sure what to do with this comment.

semantic_conventions/resource/faas.yaml
Oberon00: Probably bytes is unusual as something shown/configured in the UI. But MiB is also somewhat arbitrary. Maybe some other FaaS provider will let you configure the memory in KiB? Or it only let's you configure MiB but the API to query the max memory returns bytes.

I would be happy if we just used the unit B (bytes) for all memory-related stuff, instead of mixing GiB, MiB, KiB, B, MB, KB, GB. IMHO these derived units are better used only for display purposes.

z1c0: It's not arbitrary but the most reasonable unit here. It's the lowest resolution needed regarding this metric.

Oberon00: > It's the lowest resolution needed regarding this metric.

If we call it aws_lambda.max_memory, then yes. And I agree that it is unlikely we will come across another FaaS provider that provides finer granularity (it would not make sense).

Still, this will the first attribute we introduce with MB/MiB as unit.

z1c0: Still no good reason not to use it in this case IMHO. Having different units based on the use case is quite common.

semantic_conventions/resource/faas.yaml
arminru: Consider dropping the strict requirement and just mention in the note that it's recommended for the reasons stated in the PR description. I think it's nice to have but not necessary to properly identify and represent FaaS executions as opposed to `faas.name` and `faas.id`, which are currently required.

anuraaga:
Just a general question mostly for other spec approvers - I think container can also have `max_memory` or CPU set e.g., as k8s pod limits. Would we want to have a general location for these that apply to all types or have faas and container-specific attributes?

Oberon00:
@anuraaga I was thinking about this too. But I could e.g. imagine that faas.max_memory for a container-based AWS lambda *might* be different from what we would detect as container.max_memory. E.g. a FaaS implementation might allow short memory surges if the host has enough memory free. Or it might use a process limit instead of a container limit. I think there is merit in having an attribute that is (more or less) defined to show what was configured as "Max Memory" for the Lambda.
This is no strong opinion though.

jmacd (approved):

Oberon00 (approved):

anuraaga (approved):

semantic_conventions/resource/faas.yaml
Oberon00: FYI: https://github.com/open-telemetry/opentelemetry-specification/pull/1307#discussion_r549656854 suggests to always use the smallest practical unit.

carlosalberto (approved):

carlosalberto:
Merging as a) It seems an non-polemical addition and b) The PR has enough reviews and it has been waiting the whole break to get merged already.

PR #1304, opened 2020-12-25
fix small typo error in tracing spec api.md

WPH95:
keep the list context uniform

### DIFF
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -204,7 +204,7 @@ The API MUST allow retrieving the `TraceId` and `SpanId` in the following forms:
 `TraceId` (result MUST be a 32-hex-character lowercase string) or `SpanId`
 (result MUST be a 16-hex-character lowercase string).
 * Binary - returns the binary representation of the `TraceId` (result MUST be a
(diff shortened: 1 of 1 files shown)


carlosalberto (approved):

specification/trace/api.md
```diff
@@ -204,7 +204,7 @@ The API MUST allow retrieving the `TraceId` and `SpanId` in the following forms:
 `TraceId` (result MUST be a 32-hex-character lowercase string) or `SpanId`
 (result MUST be a 16-hex-character lowercase string).
 * Binary - returns the binary representation of the `TraceId` (result MUST be a
-16-byte array) `SpanId` (result MUST be a 8-byte array).
+16-byte array) or `SpanId` (result MUST be a 8-byte array).
```
yurishkuro, line 207: if we're fixing typos:
```suggestion
16-byte array) or `SpanId` (result MUST be an 8-byte array).
```

WPH95: good catch!

yurishkuro (approved):

PR #1307, opened 2020-12-28
Give guideline of not using unit names in attributes (#1053)

### LINKED ISSUE #1053: Trace payload/message size in semantic convention is inconsistent

Consistency should exist between the naming of payload/message size attribute names. Currently, we have a mix of how the name is constructed over 3 different protocols. HTTP has `measure-name(length)`(_`compressed`) convention, gRPC has `message`.`un/compressed`_`measure_name(size)` and message systems have `message`.`measure-name(size)`_`unit(bytes)`. Here you have the complete list:

[HTTP](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/http.md#common-attributes) (attribute on a span):
- `http.request_content_length`
- `http.request_content_length_uncompressed`
- `http.response_content_length`
- `http.response_content_length_uncompressed`

[gRPC](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/rpc.md#events) (attribute on link in a stream)
- `message.compressed_size` 
- `message.uncompressed_size` 

[Messaging system](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md#messaging-attributes) (attribute on a span):
- `messaging.message_payload_size_bytes`
- `messaging.message_payload_compressed_size_bytes `

Although this doesn't strictly violates the spec (although **message** in gRPC is borderline), we should seek consistency in the naming of measures. I would have expected something more like this:

HTTP (keeping this as this is the most established)
- `http.request_content_length`
- `http.request_content_length_uncompressed`
- `http.response_content_length`
- `http.response_content_length_uncompressed`

gRPC and messaging:
- `messaging.payload_size`
- `messaging.payload_size_uncompressed `

changes for message: joined RPC and messaging and:
- Removed unit (bytes)
- Changed compressed to uncompressed to align with HTTP
- Removed message from message_payload_size

I think it would be wise too add a note in the [attributes naming convention](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/common/common.md) some guidelines for messages, like:
- not adding the unit (bytes, mm, ...) in the attribute name

alexvanboxel:
Add a paragraph in the attribute name guidelines that discourage the use
of unit names and prefixes.

Add note about historical naming violation in messaging.md, the attribute names
message_payload_size_bytes and message_payload_compressed_size_bytes about the
incorrect use of bytes in the name.

Fixes #1053

## Changes

Related issues #

### DIFF
--- a/specification/common/attribute-and-label-naming.md
+++ b/specification/common/attribute-and-label-naming.md
@@ -49,6 +49,10 @@ Names SHOULD follow these rules:
   name prohibits existence of an equally named namespace in the future, and vice
   versa: any existing namespace prohibits existence of an equally named
   attribute or label key in the future.
+
+- Names should not contain unit names; neither should it contain a fraction of
+  the name's unit. For example, `http.request_content_length` does not have
+  bytes in its name.
   
 ## Name Pluralization guidelines
 
--- a/specification/trace/semantic_conventions/messaging.md
+++ b/specification/trace/semantic_conventions/messaging.md
@@ -181,6 +181,8 @@ For batch receiving and processing (see the [Batch receiving](#batch-receiving) 
 Even though in that case one might think that the processing span's kind should be `INTERNAL`, that kind MUST NOT be used.
 Instead span kind should be set to either `CONSUMER` or `SERVER` according to the rules defined above.
 
+**Note:** The payload size names (`message_payload_compressed_size_bytes` and `message_payload_size_bytes`) don't comply with the [Attribute and Label Naming](../../common/attribute-and-label-naming.md), it should not contain `bytes`. This is due to historical reasons.
+
 ### Attributes specific to certain messaging systems
 
 #### RabbitMQ


specification/common/attribute-and-label-naming.md
```diff
@@ -49,6 +49,11 @@ Names SHOULD follow these rules:
   name prohibits existence of an equally named namespace in the future, and vice
   versa: any existing namespace prohibits existence of an equally named
   attribute or label key in the future.
+
+- Names should not contain unit names, and their values should be measured in the
+  smallest practical unit size. For example, `http.request_content_length` does not
```
Oberon00, line 54: I would agree but it seems that sometimes other units are more natural, e.g. see https://github.com/open-telemetry/opentelemetry-specification/pull/1296#discussion_r545178077, where faas.max_memory is introduced which will use MiB instead of B (but does not use a unit suffix).

specification/common/attribute-and-label-naming.md
```diff
@@ -49,6 +49,11 @@ Names SHOULD follow these rules:
...
+  have bytes in its name and uses the smallest unit size: byte. If the smallest
+  unit is a fraction, the prefix is not part of the name.
```
Oberon00, line 56: I think you need an example here. Would this mean that we use `execution_time` instead of `execution_time_nanos`?.

specification/common/attribute-and-label-naming.md
```diff
@@ -49,6 +49,11 @@ Names SHOULD follow these rules:
   name prohibits existence of an equally named namespace in the future, and vice
   versa: any existing namespace prohibits existence of an equally named
   attribute or label key in the future.
+
+- Names should not contain unit names, and their values should be measured in the
```
Oberon00, line 53: What is the justification for that vs they should always contain unit names?

specification/common/attribute-and-label-naming.md
yurishkuro: There may be many situations where "smallest practical unit" is not well defined, easiest example - durations. A guideline for not including units in the name needs to be accompanied by an alternative way of capturing units as metric metadata.

alexvanboxel: I will remove the sentence that talks about the smallest practical unit.

specification/common/attribute-and-label-naming.md
alexvanboxel: I've been going over all the attributes defined in the spec and none mention the **unit fraction**. The only attribute that uses the **unit name** is `message_payload_size_bytes`.  So yes, I think having `execution_time` would be more consistent with the rest of the attributes.

specification/common/attribute-and-label-naming.md
alexvanboxel: Having the **unit names** in the attribute names would in most cases redundant. Example `execution_time` would be silly to have `seconds` in the name.  The same for `message_payload_size_bytes`, the fact that it's `bytes` is redundant information.

We could _discuss_ the fraction names (nanos, MiB, ...) separately: here it's about the fact that it's not used in any of the attributes yet.

tigrannajaryan: This change is reasonable for metric names which support units natively. Span or log attributes have no such support. I don't think this is a good blanket recommendation for all cases. There may be cases when units are ambiguous. In the example given `execution_time` it is not obvious to me that the execution time should be in seconds. Why would I assume that? Why not milliseconds?

Oberon00: I think it would make sense to require all spec semantic conventions to always use nanoseconds for any timespans. That way it would be unambiguous.

specification/common/attribute-and-label-naming.md
alexvanboxel: I've changed to wording that it shouldn't rely on the `smallest practical` unit. I'll flag this conversation as resolved.

specification/common/attribute-and-label-naming.md
alexvanboxel: I reworded the wording not to talk about the smallest practical unit, just to focus on the naming. Talking about this could be interesting, but I propose that someone takes this out of the scope of this PR.

tigrannajaryan (changes requested):
@alexvanboxel sorry, I am not convinced that this change is an improvement for Spans. For metrics, sure, I completely agree, it is wrong to have units in the names. For span attributes, I disagree.
I would be fine to merge a change like this if it applied only to metric names. As-is it applies to spans too so I am blocking it until we figure this out (I am willing to be convinced).

alexvanboxel:
> @alexvanboxel sorry, I am not convinced that this change is an improvement for Spans. For metrics, sure, I completely agree, it is wrong to have units in the names. For span attributes, I disagree.
> I would be fine to merge a change like this if it applied only to metric names. As-is it applies to spans too so I am blocking it until we figure this out (I am willing to be convinced).

are you opposed to units (ex. bytes) or fractions (ex. nano, milli, ...) or both?

tigrannajaryan:
> are you opposed to units (ex. bytes) or fractions (ex. nano, milli, ...) or both?

I am opposed to the blanket requirement which says we cannot have units in attribute names. I think there are valid cases when it is useful to have units in span attribute names, because not having the unit in the name makes the value ambiguous. For example `execution_time` attribute does not convey in any way whether the attribute value is in seconds or maybe milliseconds? How would the recipient of such attribute know?

specification/trace/semantic_conventions/messaging.md
```diff
@@ -181,6 +181,8 @@ For batch receiving and processing (see the [Batch receiving](#batch-receiving)
 Even though in that case one might think that the processing span's kind should be `INTERNAL`, that kind MUST NOT be used.
 Instead span kind should be set to either `CONSUMER` or `SERVER` according to the rules defined above.
 
+**Note:** The payload size names (`message_payload_compressed_size_bytes` and `message_payload_size_bytes`) don't comply with the [Attribute and Label Naming](../../common/attribute-and-label-naming.md), it should not contain `bytes`. This is due to historical reasons.
```
jmacd, line 184: Should we change this spec?

alexvanboxel:
> > are you opposed to units (ex. bytes) or fractions (ex. nano, milli, ...) or both?
> 
> I am opposed to the blanket requirement which says we cannot have units in attribute names. I think there are valid cases when it is useful to have units in span attribute names, because not having the unit in the name makes the value ambiguous. For example `execution_time` attribute does not convey in any way whether the attribute value is in seconds or maybe milliseconds? How would the recipient of such attribute know?

I'm fine with dropping this PR if no consensus can be reached. I like to have consistency, and this PR only resulted from the issue I created. Do we drop this PR?

carlosalberto:
@alexvanboxel For metrics it looks right IMHO, so worth considering going metrics-only from here.

PR #1316, opened 2021-01-05
More fully describe OTEL_PROPAGATOR and OTEL_EXPORTERS, remove confus…

anuraaga:
…ing language about SDK-specific conventions for third party propagators.

## Changes

Adds more detail about the actual formats for propagators and exporters, and removes wording about third party propagator names being in SDK documentation to ensure different languages use the same configuration values even for third party propagators. Instead, it documents how SDKs should deal with enum-type variables.

Follow up to #1311

PR #1317, opened 2021-01-05
Handle unrecognized values of sampler and exporter consistently with …

anuraaga:
…propagator.

## Changes

We only specify what to do with unrecognized values of propagator. Presumably we do the same thing for sampler / exporter.


Oberon00:
I think we should specify that in general for all environment variables.

anuraaga:
@Oberon00 Thanks, makes sense. Since it fit perfectly, I went ahead and merged it into #1316

PR #1318, opened 2021-01-05
Split OTEL_EXPORTER to OTEL_TRACE_EXPORTER and OTEL_METRICS_EXPORTER

### LINKED ISSUE #1309: OTEL_EXPORTER not consistent with other variables
I noticed this when setting multiple variables at the same time, but `OTEL_EXPORTER` doesn't seem consistent with other variables. It accepts a list for multiple exporters but is singular. `OTEL_PROPAGATORS` (more importantly) and `OTEL_RESOURCE_ATTRIBUTES` are plural.

### LINKED ISSUE #1313: otlp_span not consistent with trace, what we use for the signal / concept
We have `otlp_span` exporter to specify an exporter of only traces over OTLP but it would be more consistent with others, including `otlp_metrics` as `otlp_trace` - trace is the corresponding signal to metrics, and it is an exporter of `trace`.

anuraaga:
Fixes #1309
Fixes #1313

## Changes

Splits `OTEL_EXPORTER` into two variables that only accept a single value. This makes the syntax of the variable consistent with others (singular name, singular value) and removes some corner cases in configuration caused by the previous list (multiple jaeger exporters accept different format for endpoint, push / pull metrics exporters can't be enabled at the same time, inconsistency in not having a way of enabling multiple exporters of a same type).

### DIFF
--- a/spec-compliance-matrix.md
+++ b/spec-compliance-matrix.md
@@ -126,7 +126,7 @@ status of the feature is not known.
 |OTEL_EXPORTER_OTLP_*                          |   | +  |   | +    | +  | -    | - |    | - | -  | -   |
 |OTEL_EXPORTER_JAEGER_*                        |   | +  |   | +    | +  | -    | - | +  | - | -  | -   |
 |OTEL_EXPORTER_ZIPKIN_*                        |   | +  |   | +    |    | -    | - |    | - | -  | -   |
-|OTEL_EXPORTER                                 |   | -  |   | +    |    |      |   |    |   |  - | -   |
+|OTEL_EXPORTERS                                |   | -  |   | -    |    |      |   |    |   | -  | -   |
 |OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT               |   | +  |   | +    | +  |      |   |    |   | -  | -   |
 |OTEL_SPAN_EVENT_COUNT_LIMIT                   |   | +  |   | +    | +  |      |   |    |   | -  | -   |
 |OTEL_SPAN_LINK_COUNT_LIMIT                    |   | +  |   | +    | +  |      |   |    |   | -  | -   |
--- a/spec-compliance-matrix.md
+++ b/spec-compliance-matrix.md
@@ -126,7 +126,8 @@ status of the feature is not known.
 |OTEL_EXPORTER_OTLP_*                          |   | +  |   | +    | +  | -    | - |    | - | -  | -   |
 |OTEL_EXPORTER_JAEGER_*                        |   | +  |   | +    | +  | -    | - | +  | - | -  | -   |
 |OTEL_EXPORTER_ZIPKIN_*                        |   | +  |   | +    |    | -    | - |    | - | -  | -   |
-|OTEL_EXPORTERS                                |   | -  |   | -    |    |      |   |    |   | -  | -   |
+|OTEL_TRACE_EXPORTER                           |   | -  |   | -    |    |      |   |    |   | -  | -   |
+|OTEL_METRICS_EXPORTERS                        |   | -  |   | -    |    |      |   |    |   | -  | -   |
 |OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT               |   | +  |   | +    | +  |      |   |    |   | -  | -   |
 |OTEL_SPAN_EVENT_COUNT_LIMIT                   |   | +  |   | +    | +  |      |   |    |   | -  | -   |
 |OTEL_SPAN_LINK_COUNT_LIMIT                    |   | +  |   | +    | +  |      |   |    |   | -  | -   |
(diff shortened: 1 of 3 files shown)


specification/sdk-environment-variables.md
```diff
@@ -77,9 +77,9 @@ See [OpenTelemetry Protocol Exporter Configuration Options](./protocol/exporter.
 
 | Name          | Description                                                                  | Default |
 | ------------- | ---------------------------------------------------------------------------- | ------- |
-| OTEL_EXPORTER | Exporter to be used, can be a comma-separated list to use multiple exporters | "otlp"  |
+| OTEL_EXPORTERS | Exporters to be used as a comma separated list | "otlp"  |
```
Oberon00, line 80: It seems a bit annoying that we support exporting to two different exporters at the same time, but not two differently configured exporters of the same kind (e.g. zipkin + jaeger works, but not two different jaeger URLs). But this is a edge case anyway, because contrary to propators, you should usually only have a single exporter. Maybe that was the reason for the name?

anuraaga: It seems to already be common to specify two exporters, one for trace and one for metrics (e.g., `otlp_span` and `prometheus`). Should we have two variables for each and drop the ability to set a list for either of them to simplify things? We also get the win of not having the `otlp_span` and `otlp_metrics` variables.

I agree it's extremely rare to want to set two exporters for the same signal, and it would still be possible programmatically if really needed.

Oberon00: I like the idea of having two variables `OTEL_SPAN/TRACE_EXPORTER` and `OTEL_METRIC_EXPORTER` 👍
I think this could also make implementation simpler. E.g. right now, probably you might have one place that searches the metric exporter and another place that searches the span exporter, and if there is an unknown exporter you might not even be able to warn the user because it might just be an exporter for another signal.

anuraaga: It was definitely an annoyance I preferred not to deal with :)

https://github.com/open-telemetry/opentelemetry-java/pull/2401/files#diff-946e384e90e9353b5ed9ea9dcf4f7bb69ef6f9c71f2cd0931aa5702fc056be70R47

anuraaga: Anyone else that likes signal-specific exporters, please thumbs up and after a couple more votes I'll rework the PR.

anuraaga: Went ahead and split them up. Wondering if it means others should be e.g., `OTEL_TRACE_EXPORTER_JAEGER_AGENT_HOST`

Oberon00 (approved):

carlosalberto:
@open-telemetry/specs-approvers Please review. This is a small yet important change, that I'd like to have a lot of eyes on it ;)

yurishkuro (approved):

specification/sdk-environment-variables.md
```diff
@@ -75,13 +75,24 @@ See [OpenTelemetry Protocol Exporter Configuration Options](./protocol/exporter.
 
 ## Exporter Selection
 
+We define environment variables for setting a single exporter per signal. SDKs SHOULD provide a means to
+set multiple exporters for a signal programmatically.
```
tigrannajaryan, line 79: I am confused. First it says setting "a single exporter". Then it says set "multiple exporters". Is it "single" or "multiple"?

anuraaga: The difference is whether it's an environment variable or programmatically. Not sure how to reword it to be less confusing but let me know if you have an idea :) Or if this point isn't helpful (doc is about env variables) I can delete it too.

yurishkuro: not helpful here - SDK API requirements should be in one place.

anuraaga: Got it - removed

tigrannajaryan (approved):

trask (approved):

spec-compliance-matrix.md
```diff
@@ -126,7 +126,8 @@ status of the feature is not known.
 |OTEL_EXPORTER_OTLP_*                          |   | +  |   | +    | +  | -    | - |    | - | -  | -   |
 |OTEL_EXPORTER_JAEGER_*                        |   | +  |   | +    | +  | -    | - | +  | - | -  | -   |
 |OTEL_EXPORTER_ZIPKIN_*                        |   | +  |   | +    |    | -    | - |    | - | -  | -   |
-|OTEL_EXPORTER                                 |   | -  |   | +    |    |      |   |    |   |  - | -   |
+|OTEL_TRACE_EXPORTER                           |   | -  |   | -    |    |      |   |    |   | -  | -   |
+|OTEL_METRICS_EXPORTERS                        |   | -  |   | -    |    |      |   |    |   | -  | -   |
```
fbogsany, line 130:
```suggestion
|OTEL_METRICS_EXPORTER                        |   | -  |   | -    |    |      |   |    |   | -  | -   |
```

carlosalberto (approved):

bogdandrutu (approved):

PR #1319, opened 2021-01-05
Suffix OTLP TIMEOUT variable with MILLIS

### LINKED ISSUE #1312: OTLP Exporter timeout env var not consistent with batch span processor
BatchSpanProcessor suffixes with MILLIS

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/sdk-environment-variables.md#batch-span-processor

OTLP doesn't

I guess they should both be suffixed with MILLIS

anuraaga:
Fixes #1312

## Changes

Suffixes OTLP timeout variables with millis for consistency with BSP variables and to allow the variable to be self-documenting in terms of what the unit of the value is.


carlosalberto:
Approving as I'm ok with this approach (I _think_ @jkwatson will like it too ;) )

cc @bogdandrutu

carlosalberto (approved):

jkwatson (approved):

anuraaga:
Superceded by #1325 - both solve the issue but that probably solves it better.

PR #1320, opened 2021-01-05
docs(logs): GCP log field "http_request" maps to the "Resource" unified model field

nicoleczhu:
GCP LogEntries have an [HttpRequest field type](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#HttpRequest), that according to [OTel log specs](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/master/processor/resourcedetectionprocessor) so far, should map to the `Resource` field. It's arguably the origin/context of a log. 

Though I can also see why http request context should map under the `Attributes` field instead... Disclaimer: I'm new to OTel community, please let me know if there is a process or format to raise these types of nits. Thanks!


tigrannajaryan:
> GCP LogEntries have an [HttpRequest field type](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#HttpRequest), that according to [OTel log specs](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/master/processor/resourcedetectionprocessor) so far, should map to the `Resource` field. It's arguably the origin/context of a log.
> 
> Though I can also see why http request context should map under the `Attributes` field instead... Disclaimer: I'm new to OTel community, please let me know if there is a process or format to raise these types of nits. Thanks!

`http_request` data should be mapped to `Attributes` since it can vary for each Log record. `Resource` is assumed to contain constant data (per source of telemetry).

specification/logs/data-model.md
```diff
@@ -890,6 +890,7 @@ Field            | Type               | Description
 -----------------|--------------------| ------------------------------------------------------- | ---------------------------
 timestamp        | string             | The time the event described by the log entry occurred. | Timestamp
 resource         | MonitoredResource  | The monitored resource that produced this log entry.    | Resource
+http_request     | HttpRequest        | The HTTP request associated with the log entry, if any. | Resource
```
tigrannajaryan, line 893:
```suggestion
http_request     | HttpRequest        | The HTTP request associated with the log entry, if any. | Attributes
```
Ideally this should also specify the attribute name to use.

tigrannajaryan (changes requested):

specification/logs/data-model.md
nicoleczhu: Question: Should Google's `httpRequest` object be flattened (Option A), i.e. tell users to map `http_request.requestMethod` into `Attributes["http.method"]` and so on

Or should it remain nested (Option B), i.e. tell users to map `http_request` into `Attributes["google.httpRequest"]`

Option A: 
```json
 "Attributes":{
      "http.scheme":"https",
      "http.host":"donut.mycie.com",
      "http.target":"/order",
      "http.method":"post",
      "http.status_code":500,
      "http.flavor":"1.1",
      "http.user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_0) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/80.0.3987.149 Safari/537.36",
   }
```

Option B:
```json
"Attributes":{
  httpRequest: {
    requestMethod: "POST"
    requestUrl: "/"
    userAgent: "Go-http-client/1.1"
    protocol: "HTTP/1.1"
  }
}
```

tigrannajaryan: This is a good question, and I don't think Otel has a good answer to it.

It is not clear whether we should recommend to preserve the original data structure or we should recommend coercing it into Otel-like attribute names. There are cons and pros with each approach. This applies not just to this particular attribute of GCP logs but to any data that comes from existing sources. 

Can you please open a separate issue to discuss this and come up with a general recommendation?

tigrannajaryan: For now I suggest that we merge this PR as is since it is an improvement over previous state.

tigrannajaryan (approved):

specification/logs/data-model.md
nicoleczhu: Sounds great! Will do

jmacd (approved):

nicoleczhu:
Hi @jmacd @tigrannajaryan , I dont have permission to merge this PR (if it should still be merged). Can you help to merge? Thank you

PR #1322, opened 2021-01-06
Define default value for traceid ratio.

### LINKED ISSUE #1310: TraceIdRatioBased default ratio when configured with OTEL_TRACE_SAMPLER
Does TraceIdRatioBased have a default ratio or should SDKs throw an error if a ratio isn't provided? This would generally not be an in issue in programatic configuration, but if using environment variables as [here](https://github.com/open-telemetry/opentelemetry-specification/blob/f2869a369ff4b93024e3437837320da76d2c0637/specification/sdk-environment-variables.md#general-sdk-configuration), the sampler name could be set without an arg. I think 1.0 could be a reasonable default, or an error could make sense too.

anuraaga:
Fixes #1310

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,9 @@ Updates:
   ([#1269](https://github.com/open-telemetry/opentelemetry-specification/pull/1269))
 - Add requirement that the SDK allow custom generation of Trace IDs and Span IDs
   ([#1006](https://github.com/open-telemetry/opentelemetry-specification/pull/1006))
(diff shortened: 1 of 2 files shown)


carlosalberto (approved):

Oberon00 (approved):

tigrannajaryan (approved):

yurishkuro (approved):

carlosalberto:
We are ready to merge. @anuraaga think you could add a Changelog entry?

anuraaga:
@carlosalberto Yup added

PR #1325, opened 2021-01-07
Remove MILLIS from BatchSpanProcessor variables and define semantics for...

### LINKED ISSUE #1312: OTLP Exporter timeout env var not consistent with batch span processor

anuraaga:
… duration values.

Fixes #1312

## Changes

Removes MILLIS from BatchSpanProcessor time variables in favor of defining durations to always be interpreted as milliseconds.


specification/sdk-environment-variables.md
```diff
@@ -2,6 +2,22 @@
 
 The goal of this specification is to unify the environment variable names between different OpenTelemetry SDK implementations. SDKs MAY choose to allow configuration via the environment variables in this specification, but are not required to. If they do, they SHOULD use the names listed in this document.
 
+## Special configuration types
+
+### Duration
```
anuraaga, line 7: @mwear I'm imaging Endpoint to also be added to this list after #1234 (awesome PR number!)

yurishkuro: I am not sure this is worth it. In Go this can be parsed with stdlib, but that's not always the case in other langs, so it introduces an extra burden on SDK implementations with not a lot of benefits.

anuraaga: @yurishkuro We're discussing on #1312. @jsuereth has started working on adding it to all the languages and not sure how best to proceed with it without a definition of the unit strings. https://github.com/open-telemetry/opentelemetry-java/pull/2439 I'm personally fine with anything that makes the variable names consistent :)

specification/sdk-environment-variables.md
```diff
@@ -2,6 +2,22 @@
...
+
+Any value that represents a duration, for example a timeout, MUST interpret a purely numeric value as a number of
```
jsuereth, line 9: FYI - I didn't notice you already had this PR.   I described this a bit differently: https://github.com/open-telemetry/opentelemetry-specification/compare/master...jsuereth:wip-duration-defaults?expand=1

I think your specification is a lot smaller, but is missing the bit about `PositiveInteger`, i.e. I think we also need to limit the numeric value so it's reasonable to contribute a parser in all languages.


Suggested change:

```
Any value that represents a duration, for example a timeout, MUST interpret a purely positive integer value as a number of milliseconds. Formatted strings with a positive integer followed immediately by a unit string MAY be interpreted for the following units:
```

jsuereth (changes requested):

specification/sdk-environment-variables.md
carlosalberto: > Formatted strings with a positive integer followed immediately by a unit string MAY be interpreted for the following units:

I'm afraid that languages not supporting this specific sentence will soon be approached by users asking: "Hey, why doesn't this work? In Go it does but it doesn't in X language", converting it the de facto required.

(I know this is a small piece of work - but if we allow this to creep, we then may allow other 20 small additions and drag the release more than it is - and, well, we can *easily* add support for this parsing by the time we do the next release, which should include **stable** metrics :) )

jsuereth: I understand the concern.

1. I'm not sure how else you would phrase this in a way to ALLOW Go's handling of duration and also allow parsing units in the future to become required (post-GA).  If you have a recommendation for how we *allow* this change to occur post-GA, let us know.  Currently, I think having `_MILLIS` in env names actually prevents this format from being a good idea post-GA so at a minimum I'd like to remove that from the ENV variable.
2. I think it's on the maintainers of those langauge maintainers to PUNT the request or ask community for PRs.  i.e. This spec change is intended to NOT force languages to implement parsers, but allow them to come post GA, but will need discipline from maintainers to focus on more important issues.
3. I'm more than happy to contribute parsers for this to languages which need it, feel free to assign it to me.  Anything I can do to help maintainers focus on stable metrics :)

bogdandrutu:
@jsuereth looks like you requested changes, is there any concrete issue/reason that we need to solve?

jsuereth:
@bogdandrutu Yeah, here's the quote:

> FYI - I didn't notice you already had this PR. I described this a bit differently: https://github.com/open-telemetry/opentelemetry-specification/compare/master...jsuereth:wip-duration-defaults?expand=1
> 
> I think your specification is a lot smaller, but is missing the bit about `PositiveInteger`, i.e. I think we also need to limit the numeric value so it's reasonable to contribute a parser in all languages.
> 
> Suggested change:
> 
> ```
> Any value that represents a duration, for example a timeout, MUST interpret a purely positive integer value as a number of milliseconds. Formatted strings with a positive integer followed immediately by a unit string MAY be interpreted for the following units:
> ```

anuraaga:
@jsuereth Sorry for missing the comment about positive numbers! Will address

@carlosalberto So do you want to leave out discussion of duration units in this PR? What happens if different languages end up with different units? Maybe a low enough probability not to matter though.

carlosalberto:
> So do you want to leave out discussion of duration units in this PR?

Not quite, but have milliseconds being the default. IIRC the parsers @jsuereth mentioned take milliseconds as the default, unless I'm wrong?

jsuereth:
@carlosalberto The parser I contributed to Java (and plan for other languages) takes milliseconds as the default. 

 I think your plan for "allow-for-GA" with specific changes is exactly in line with what I was hoping to accomplish!

Parsers can be added when it makes sense to do so.

anuraaga:
@carlosalberto Ah ok - I just added the note about negative numbers not being supported. I think my wording of MUST vs MAY captured the aspect of the default being numeric milis, let me know if this is still missing anything.

carlosalberto:
Based on the discussion we had today at the Spec call, we decided to go with:

1) Remove `_MILLIS` for all related environment variables and have milliseconds as the actual default unit.
2) Fill a follow-up issue (allowed for GA) to have the duration unit parsing in place. We will merge this as soon as the main SIGs have support in place. See #1351 

@anuraaga Please amend appropriately ;)

anuraaga:
@carlosalberto Ok updated to only talk about the number.

carlosalberto (approved):

carlosalberto:
@anuraaga Thanks for the update!

@open-telemetry/specs-approvers please review/approve ;)

specification/sdk-environment-variables.md
```diff
@@ -2,10 +2,25 @@
...
+### Enum value
+
 For variables which accept a known value out of a set, i.e., an enum value, SDK implementations MAY support additional values not listed here.
 For variables accepting an enum value, if the user provides a value the SDK does not recognize, the SDK MUST generate a warning and gracefully ignore the setting.
 
+### Duration
+
+Any value that represents a duration, for example a timeout, MUST be an integer representing a number of
+milliseconds. The value is non-negative - if a negative value is provided, the SDK MUST generate a warning and
+gracefully ignore the setting.
```
tigrannajaryan, line 20:
```suggestion
milliseconds. The value is non-negative - if a negative value is provided, the SDK MUST generate a warning,
gracefully ignore the setting and use the default value if it is defined.
```

tigrannajaryan (approved):

jmacd (approved):

tedsuo (approved):
LGTM. Appreciate the simplification and, dare I say it, standardization.

PR #1327, opened 2021-01-08
Clarify tracer name for self-instrumenting apps/libraries

maxgolov:
Contradictory statement in Get Tracer API vs Glossary definition of "Instrumentation Library" meaning.

## Changes

- Get Tracer API definition suggests that `instrumentation library` is NOT the `instrumented library`.

- However, the Glossary elaborates on a case when both [MAY be the same thing](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/glossary.md#instrumentation-library)

It would be good to clear this ambiguity. For manually-instrumented applications or libraries that invoke OpenTelemetry SDK directly, when there is no ready-to-use `instrumentation library` - the apps should be able to pass their library name or module name, or component name to Get Tracer API. This would also resolve the multi-tenancy scenarios, where one app or library with many modules that do not have existing `instrumentation library` - themselves may provide their own component name to Tracer API.

Related [oteps](https://github.com/open-telemetry/oteps) https://github.com/open-telemetry/oteps/issues/145


specification/trace/api.md
```diff
@@ -102,10 +102,13 @@ The `TracerProvider` MUST provide the following functions:
...
 - `name` (required): This name must identify the [instrumentation library](../overview.md#instrumentation-libraries)
-  (e.g. `io.opentelemetry.contrib.mongodb`) and *not* the instrumented library.
-  In case an invalid name (null or empty string) is specified, a working
-  default Tracer implementation as a fallback is returned rather than returning
-  null or throwing an exception.
+  (e.g. `io.opentelemetry.contrib.mongodb`). [Instrumented library](../glossary.md#instrumented-library)
+  and [Instrumentation library](../glossary.md#instrumentation-library) may be
+  the same library if application or library has built-in OpenTelemetry instrumentation.
+  In that scenario the `name` denotes a module name or component name within the library
+  or application being instrumented. In case an invalid name (null or empty string)
```
Oberon00, line 109: Why only in that scenario? Also, in that case "being instrumented" is confusing, as it is the same as the library or application doing the instrumentation.

tigrannajaryan (approved):

specification/trace/api.md
maxgolov: I agree that in more than just that scenario. Can you suggest a better wording?

maxgolov: Key point is that application may have many modules. Some live in their own library. Some modules, as in, compiled into the main executable itself. Not every component would be having "standard function", i.e. not necessarily sql, http, db access or alike. Some could be renderers or processors of certain formats. For example, MS Word document processor; or some renderer process in Chromium browser (client). Then component has to uniquely identify itself to the `Tracer` , that is providing its own `name` , assuming it is `instrumentationName`. Obviously there are going to be no pre-existing instrumentation libraries, as it's the case of manually instrumenting some random component that cannot be categorized beforehand. In that flow, the component itself may provide its own unique name, module name, or class name, or component name, or other logical entity name (if spanned across multiple compilation units or libraries). And the `Tracer` allocated for that component then gets passed around... `being instrumented` as in - _manually_ instrumenting itself by invoking OpenTelemetry `TracerProvider.GetTracer(...)` API for itself.

Oberon00: So the change you want to introduce here is that you want to allow applications to not specify their name but the name of a submodule within the application (even if it is not its own library)?

maxgolov: Mostly yes. Not related to "applications". But related to the actual components or libraries / SDKs within the app identify themselves to Tracer, i.e. if I'm designing a `File Open Dialog` component or library (embeddable in another application), I can pass my module or component name `MyCompany.File.Dialog` as instrumentation name to Tracer. There would be no ready-to-use instrumentation for my scenario, so as a Library Developer I need to supply my component name.

In a heavily multi-tenant apps (say, 40 different teams working on 40 different components, all integrated in one app), I'd organize my module names to be unique.

**NOT the instrumented library** wording here in original spec is confusing. And prevents me from implementing per-component / per-module tracers within the app. As a minimum, I'd like to remove the  `and *not* the instrumented library` statement from spec.

maxgolov: @Oberon00 - did I answer your question? Would you suggest some better wording if you don't have objections about this semantical meaning of the `instrumentation name` ?

Oberon00: > "NOT the instrumented library" wording in original spec is confusing. It prevents me from implementing per-component / per-module tracers within the app.

I see now that this statement can indeed be misleading. Maybe adding `(unless it's the same as the instrumenting library, i.e., a self-instrumented library)` would be enough?

Oberon00, line 109:
```suggestion
  In that scenario the `name` denotes a module name or component name within that library
  or application. In case an invalid name (null or empty string)
```

With this adjustment, I'm fine with the change.

maxgolov: I committed your suggestion.

cijothomas (approved):

Oberon00 (approved):

specification/trace/api.md
```diff
@@ -102,10 +102,13 @@ The `TracerProvider` MUST provide the following functions:
...
-  In case an invalid name (null or empty string) is specified, a working
-  default Tracer implementation as a fallback is returned rather than returning
-  null or throwing an exception.
+  (e.g. `io.opentelemetry.contrib.mongodb`). [Instrumented library](../glossary.md#instrumented-library)
+  and [Instrumentation library](../glossary.md#instrumentation-library) may be
+  the same library if application or library has built-in OpenTelemetry instrumentation.
+  In that scenario the `name` denotes a module name or component name within that library
+  or application. In case an invalid name (null or empty string)
+  is specified, a working default Tracer implementation as a fallback is returned
+  rather than returning null or throwing an exception.
```
arminru, line 111:
```suggestion
  (e.g. `io.opentelemetry.contrib.mongodb`).
  If an application or library has built-in OpenTelemetry instrumentation, both
  [Instrumented library](../glossary.md#instrumented-library) and
  [Instrumentation library](../glossary.md#instrumentation-library) may refer to the same library.
  In that scenario, the `name` denotes a module name or component name within that library
  or application.
  In case an invalid name (null or empty string) is specified, a working
  default Tracer implementation as a fallback is returned rather than returning
  null or throwing an exception.
```

I think phrasing it this way makes it a bit easier to read and understand.

arminru (approved):

maxgolov:
@tigrannajaryan - looks like we got a quorum. How do we get this merged?

yurishkuro (approved):

PR #1328, opened 2021-01-08
Specify required integer limits

### LINKED ISSUE #1315: Numeric environment variables don't define limit
Numeric environment variables currently don't define an upper limit on a value

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/sdk-environment-variables.md

In Java, we currently read some variables as integer, others as long. I suspect none of these will ever need to be such huge values and we could state that all numbers have an upper limit of the maximum of a signed integer, otherwise we should document which are integer and which are long.

punya:
Fixes #1315.

## Changes

Adds verbiage specifying integer limit handling for SDK configuration environment variables.

### DIFF
--- a/specification/sdk-environment-variables.md
+++ b/specification/sdk-environment-variables.md
@@ -2,7 +2,7 @@
 
 The goal of this specification is to unify the environment variable names between different OpenTelemetry SDK implementations. SDKs MAY choose to allow configuration via the environment variables in this specification, but are not required to. If they do, they SHOULD use the names listed in this document.
 
-If an SDK chooses to support an integer-valued environment variable, it SHOULD support nonnegative values between 0 and 2<sup>31</sup>&minus;1 (inclusive). Individual SDKs MAY choose to support a larger range of values.
+If an SDK chooses to support an integer-valued environment variable, it SHOULD support nonnegative values between 0 and 2³¹ − 1 (inclusive). Individual SDKs MAY choose to support a larger range of values.
 
 ## General SDK Configuration
 
--- a/specification/sdk-environment-variables.md
+++ b/specification/sdk-environment-variables.md
@@ -2,6 +2,8 @@
 
 The goal of this specification is to unify the environment variable names between different OpenTelemetry SDK implementations. SDKs MAY choose to allow configuration via the environment variables in this specification, but are not required to. If they do, they SHOULD use the names listed in this document.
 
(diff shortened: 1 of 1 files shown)


specification/sdk-environment-variables.md
```diff
@@ -2,6 +2,8 @@
 
 The goal of this specification is to unify the environment variable names between different OpenTelemetry SDK implementations. SDKs MAY choose to allow configuration via the environment variables in this specification, but are not required to. If they do, they SHOULD use the names listed in this document.
 
+If an SDK chooses to support an integer-valued environment variable, it SHOULD support nonnegative values between 0 and 2<sup>31</sup>&minus;1 (inclusive). Individual SDKs MAY choose to support a larger range of values.
```
Oberon00, line 5: I think this is a reasonable requirement, it allows every number to fit in a signed 32 bit int.

However, I don't really like the raw HTML in here:

```suggestion
If an SDK chooses to support an integer-valued environment variable, it SHOULD support nonnegative values between 0 and 2³¹ − 1 (inclusive). Individual SDKs MAY choose to support a larger range of values.
```

Oberon00 (approved):

anuraaga (approved):
Thanks!

specification/sdk-environment-variables.md
```diff
@@ -2,6 +2,8 @@
 
 The goal of this specification is to unify the environment variable names between different OpenTelemetry SDK implementations. SDKs MAY choose to allow configuration via the environment variables in this specification, but are not required to. If they do, they SHOULD use the names listed in this document.
 
+If an SDK chooses to support an integer-valued environment variable, it SHOULD support nonnegative values between 0 and 2³¹ − 1 (inclusive). Individual SDKs MAY choose to support a larger range of values.
+
```
carlosalberto, line 6: Not a strong opinion, but I'd leave out "Individual SDKs MAY choose to support a larger range of values." (just to avoid tempting SDKs to do this ;) )

carlosalberto (approved):

bogdandrutu (approved):

specification/sdk-environment-variables.md
punya: I put that line in so that SDKs wouldn't feel compelled to look through all existing parsing logic to force a smaller range (e.g. if the default is to parse to a 64-bit int).

PR #1332, opened 2021-01-11
Opencensus compatibility specification

### LINKED ISSUE #766: Add a section on compatibility with OpenCensus to Compatibility doc
This issue is to track just the OpenCensus part of #114

jsuereth:
Fixes #766

## Changes

Provide a specification of OpenCensus migration to open telemetry.
- Specifies shims for TRACE that can be implemented per-langauage.
- Specifies overarching goals for migration
- ONLY accounts for specification<->specification issues and calls out possible incompatibility issues for unspecified behavior.
- Does *not* specify Metric compatibility as that will wait for Metrics to be declared GA / stable.

A few notes for discussion:

1. This spec calls for semantic-name mapping so that *users* of an OpenCensus bridge will get OpenTelemetry-looking metrics even from OpenCensus instrumented libraries.   The expectation is this eases the migration from OC => OTel, but that labels may "break" from OC-only implementation, and won't break as OC instrumented lirbaries are migrated to OTel.
2. This attempts to identify all known incompatibilites we've found between OpenCensus + OpenTelemetry APIs from the Java + Go implementations, as well as directly comparing the Specifications.  .  Unfortunately, a lot of incompatibilities are due to unspecified API, SDK or other behaviors, and cannot be (fully) addressed in this Specification.   Instead a set of principles for how to make tradeoffs are proposed up-front for implementors of OpenCensus-bridge components.
3. The specification is expected to evolve with known incompatibilities as needed, in an attempt to keep language-specific design decisions around unspecified behavior as uniform as possible.


specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,152 @@
...
+
+## Abstract
+
+The OpenTelemetry project aims to provide backwards compatibility with the
+[OpenCensus](https://opencensus.io) project in order to ease migration of
+instrumented codebases.
+
+This functionality will be provided as a bridge layer implementing the
+[OpenCensus API](https://github.com/census-instrumentation/opencensus-specs)
+using the OpenTelemetry API. This layer MUST NOT rely on implementation specific
```
anuraaga, line 11: Also I believe, it MUST NOT depend on the OpenCensus implementation, only the OpenCensus API.

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,152 @@
...
+OpenCensus supports two primary types of telemetry: Traces and Stats (Metrics).
+Compatibility for these is defined separately.
+
+> The overridding philosophy for compatibility is that OpenCensus instrumented
+> libraries and applications need make *no change* to their API usage in order
+> to use OpenTelemetry.   All changes should be solely configuration / setup.
+
+## Goals
+
+OpenTelemetry<->OpenCensus tracing compatibility has the following goals:
```
anuraaga, line 34:
```suggestion
OpenTelemetry<->OpenCensus compatibility has the following goals:
```

Tracing is mentioend below

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,152 @@
...
+
+1. OpenCensus has no hard dependency on OpenTelemetry
+2. Minimal changes to OpenCensus for implementation
+3. Easy for users to use, ideally no change to their code
+
+Additionally, for tracing there are the following goals:
+
+1. Maintain parent-child span relationship from applications and libraries
```
anuraaga, line 42:
```suggestion
1. Maintain parent-child span relationship between applications and libraries
```

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,152 @@
...
+2. Maintain unscoped span relationships from applications and libraries
```
anuraaga, line 43: What does unscoped span relationship mean?

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,152 @@
...
+4. TraceFlags in both OpenCensus and OpenTelemetry only specify the single
+   `sampled` flag ([OpenTelemetry](../trace/api.md#spancontext),
+   [OpenCensus](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/TraceConfig.md#traceparams)).
+   Some OpenCensus APIs support "debug" and "defer" tracing flags in additon to
+   "sampled".  In this case, the OpenCensus bridge will do its best to support
+   and translate unspecified flags into the closest OpenTelemetry equivalent.
+
+### Context Propagation
+
+The shim will provide an OpenCensus `PropogationComponent` implementation which
```
anuraaga, line 93:
```suggestion
The shim will provide an OpenCensus `PropagationComponent` implementation which
```

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,152 @@
...
+maps OpenCenus binary and text progation to OpenTelemetry context.
```
anuraaga, line 94:
```suggestion
maps OpenCenus binary and text propagation to OpenTelemetry context.
```

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,152 @@
...
+
+#### Text Context
+
+This adapter MUST use an OpenTelemetry `TextMapPropogator` to implement the
```
anuraaga, line 98:
```suggestion
This adapter MUST use an OpenTelemetry `TextMapPropagator` to implement the
```

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,152 @@
...
+OpenCensus `TextFormat`.
+
+This adapter SHOULD use configured OpenTelemetry `TextMapPropogator` on the
```
anuraaga, line 101:
```suggestion
This adapter SHOULD use configured OpenTelemetry `TextMapPropagator` on the
```

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,152 @@
...
+OpenTelemetry `TraceProvider` for text format propagation.
+
+This adapter MUST provide a default `W3CTraceContextPropogator` in lieu of a
```
anuraaga, line 104:
```suggestion
This adapter MUST provide a default `W3CTraceContextPropagator` in lieu of a
```

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,152 @@
...
+propogator defined against the `TraceProvider`.
```
anuraaga, line 105:
```suggestion
propagator defined against the `TraceProvider`.
```

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,152 @@
...
+
+#### B3 Context
+
+This adapter SHOULD use an contributed OpenTelemetry `B3Propogator` for the
```
anuraaga, line 109:
```suggestion
This adapter SHOULD use n contributed OpenTelemetry `B3Propagator` for the
```

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,152 @@
...
+B3 text format.
+
+#### OpenCensus Binary Context
+
+This adapter MUST implement OpenCensus `BinaryPropogator` to write OpenCensus
```
anuraaga, line 114:
```suggestion
This adapter MUST implement OpenCensus `BinaryPropagator` to write OpenCensus
```

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,152 @@
+# OpenCensus Compatibility
+
+## Abstract
+
+The OpenTelemetry project aims to provide backwards compatibility with the
+[OpenCensus](https://opencensus.io) project in order to ease migration of
```
anuraaga, line 6: I think it could help to have an example migration. Dunno, but for example something like

- Start with OC app that uses spring-boot and mongodb instrumented
- Add OTel shim and OTel SDK as dependencies, configure OTel SDK instead of OC implementation. Should still work mostly the same
- Replace spring-boot OC instrumentation with OTel instrumentation. Should still work mostly the same
- Replace MongoDB OC instrumentation with OTel instrumentation
- Drop OC

specification/compatibility/opencensus.md
jsuereth: The current shims (at least for Java) actually do, and require latest version of OpenCensus.

anuraaga: Oh - I wonder why we reimplement binaryformat then instead of using the one in the implementation, thought we were avoiding that dependency. Or maybe we haven't completed stripping it out?

https://github.com/open-telemetry/opentelemetry-java/blob/228e26ad9b4e35c4bcab8de7654ea270c5447196/opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryBinaryFormatImpl.java#L20

jsuereth: It turns out we can't really strip the dependency (in the shim).   The API surface area of OpenCenuss and the extension points require us to directly tie into the SDK.

TL;DR: the requirement that an "OpenTelemetry Sandwich" works means that we need to directly touch some SDK components.

````
|-- Applicaiton - Configured OpenTelemetry SDK --------------------------------- |
    |--  Library A -> Using OpenCensus to genereate Trace A  --------- |
      |--  Library B -> Using OpenTelemetry to generate Trace B-- |
```

If we want Trace B to discover Trace A via context and denote it as a parent, then the current approach is what we came up with.

specification/compatibility/opencensus.md
jsuereth: That was Paraagraph 4.  Should I expand that with this?  I do want to make sure it's clear the path-of-adoption and goals.

anuraaga: Yeah paragraph 4 is a general statement, I think an example with actual library names, etc can help a lot

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,152 @@
...
+propagator defined against the `TraceProvider`.
+
+#### B3 Context
+
+This adapter SHOULD use n contributed OpenTelemetry `B3Propagator` for the
+B3 text format.
+
+#### OpenCensus Binary Context
+
+This adapter MUST implement OpenCensus `BinaryPropagator` to write OpenCensus
```
anuraaga, line 114: So based on https://github.com/open-telemetry/opentelemetry-specification/pull/1332/files#r557580446, since we do need a dependency on the opencensus coreimpl anyways, is there any reason to require an implementation in the adapter then? May as well save some work if this isn't really providing a benefit (complete decoupling from coreimpl).

jsuereth: Two reasons, both technical and not ideal:

1. The utilities for writing the binary format are package-private, we can't get em (at least in Java).
2. We DO need to override it for TextPropagater.

I'm ok adjusting the specification however we see fit here.

specification/compatibility/opencensus.md
jsuereth: Links and link-like behavior from other features.

specification/compatibility/opencensus.md
jsuereth: Ok, updated.

specification/compatibility/opencensus.md
anuraaga: How about using the term `link` here then - we don't have this concept of scoped span elsewhere I think, some [mentions](https://github.com/open-telemetry/opentelemetry-specification/search?q=scope) are not the same as the meaning here I think.

specification/compatibility/opencensus.md
anuraaga: Many of these comments have been resolved without being fixed it seems, did you miss pushing a commit?

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,191 @@
...
+upstream OpenTelemetry API.
+
+The OpenCensus Shim and the OpenTelemetry API/SDK are expected to be consumed
+simultaneously in a running service, in order to ease migration from the former
+to the latter.  It is expected that application owners will begin the migration
+process towards OpenTelemetry via the shim and adding new telemetry information
+via OpenTelemetry.  Slowly, libraries and integrations will also migrate
+towards opentelemetry until the shim is no longer necessary.
+
+For example, an applciation may have traces today of the following variety:
```
anuraaga, line 25:
```suggestion
For example, an application may have traces today of the following variety:
```

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,191 @@
...
+
+```
+|-- Applicaiton - Configured OpenCenus --------------------------------- |
+    |--  gRPC -> Using OpenCensus to genereate Trace A  --------- |
```
anuraaga, line 29:
```suggestion
    |--  gRPC -> Using OpenCensus to generate Trace A  --------- |
```

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,191 @@
...
+      |--  Application -> Using OpenCensus to generate a sub Trace B-- |
+```
+
+In this case, the application should be able to update its outer layer to
+OpenTelemetry, without having to wait for all downstream dependencies to
+have updated to OpenTelemetry (or deal with incompatibilities therein). The
+Application also doesn't need to rewrite any of its own instrumentation.
+
+```
+|-- Applicaiton - Configured Otel w/ OpenCensus Shim ------------------- |
```
anuraaga, line 39:
```suggestion
|-- Application - Configured Otel w/ OpenCensus Shim ------------------- |
```

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,191 @@
...
+    |--  gRPC -> Using OpenCensus to genereate Trace A  --------- |
+      |--  Application -> Using OpenCensus to generate a sub Trace B-- |
+```
+
+Next, the application can update its own instrumentation in a piecemeal fashion:
+
+```
+|-- Applicaiton - Configured Otel w/ OpenCensus Shim ---------------------- |
```
anuraaga, line 47:
```suggestion
|-- Application - Configured Otel w/ OpenCensus Shim ---------------------- |
```

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,191 @@
...
+    |--  gRPC -> Using OpenCensus to genereate Trace A  --------- |
+      |--  Application -> Using OpenTelemetry to generate a sub Trace B-- |
+```
+
+> This layer of Otel -> OpenCensus -> Otel tracing can be thought of as the
+> "OpenTelemetry sandwhich" problem, and is the key motivating factor for
+> this specification.
+
+Finally, the APplication would update all usages of OpenCensus to OpenTelemetry.
```
anuraaga, line 56:
```suggestion
Finally, the Application would update all usages of OpenCensus to OpenTelemetry.
```

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,191 @@
...
+
+```
+|-- Applicaiton - Configured Otel standalong ----------------------------- |
```
anuraaga, line 59:
```suggestion
|-- Application - Configured Otel standalone ----------------------------- |
```

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,191 @@
...
+```
+
+In this case, the application should be able to update its outer layer to
+OpenTelemetry, without having to wait for all downstream dependencies to
+have updated to OpenTelemetry (or deal with incompatibilities therein). The
+Application also doesn't need to rewrite any of its own instrumentation.
+
+```
+|-- Applicaiton - Configured Otel w/ OpenCensus Shim ------------------- |
+    |--  gRPC -> Using OpenCensus to genereate Trace A  --------- |
```
anuraaga, line 40:
```suggestion
    |--  gRPC -> Using OpenCensus to generate Trace A  --------- |
```

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,191 @@
...
+      |--  Application -> Using OpenCensus to generate a sub Trace B-- |
+```
+
+Next, the application can update its own instrumentation in a piecemeal fashion:
+
+```
+|-- Applicaiton - Configured Otel w/ OpenCensus Shim ---------------------- |
+    |--  gRPC -> Using OpenCensus to genereate Trace A  --------- |
```
anuraaga, line 48:
```suggestion
    |--  gRPC -> Using OpenCensus to generate Trace A  --------- |
```

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,191 @@
...
+
+> This layer of Otel -> OpenCensus -> Otel tracing can be thought of as the
+> "OpenTelemetry sandwhich" problem, and is the key motivating factor for
+> this specification.
+
+Finally, the APplication would update all usages of OpenCensus to OpenTelemetry.
+
+```
+|-- Applicaiton - Configured Otel standalong ----------------------------- |
+    |--  gRPC -> Using Otel to genereate Trace A  --------- |
```
anuraaga, line 60:
```suggestion
    |--  gRPC -> Using Otel to generate Trace A  --------- |
```

specification/compatibility/opencensus.md
anuraaga: Sent https://github.com/open-telemetry/opentelemetry-java/pull/2605 to demonstrate not copying the binary format implementation.

This may not be possible in all languages, but I think this could be worded to `MUST provide an implementation of ... This MAY be the same implementation as OpenCensus if it is available.`

jsuereth:
@anuraaga Looks like i had a dropped commit when rebasing that lost all the committed suggestions from upstream, trying to recollect those now.

specification/compatibility/opencensus.md
jsuereth: done.   Thanks for catching that, I had missed the class was not final.

specification/compatibility/opencensus.md
jsuereth: Done

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,192 @@
...
+
+The shim will provide an OpenCensus `PropagationComponent` implementation which
+maps OpenCenus binary and text propagation to OpenTelemetry context.
+
+#### Text Context
+
+This adapter MUST use an OpenTelemetry `TextMapPropagator` to implement the
+OpenCensus `TextFormat`.
+
+This adapter SHOULD use configured OpenTelemetry `TextMapPropagator` on the
```
anuraaga, line 140: Just checking, is this something Java can't do but other languages may? For example, we have explicit extension points for B3 and W3C

https://github.com/open-telemetry/opentelemetry-java/blob/main/opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryPropagationComponentImpl.java#L28

So it seems like it wouldn't make sense to use the OTel configured propagator instead of directly exposing the correct implementations.

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,192 @@
...
+OpenTelemetry `TraceProvider` for text format propagation.
+
+This adapter MUST provide a default `W3CTraceContextPropagator` in lieu of a
```
anuraaga, line 143: Not sure what this sentence means. Is it the same as my above point, where we actually don't want to use the configured provider? More concretely, this line of code in Java?

https://github.com/open-telemetry/opentelemetry-java/blob/main/opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryPropagationComponentImpl.java#L34

anuraaga (approved):
Just a few smaller points but this seems to be good, thanks!

specification/compatibility/opencensus.md
jsuereth: I can clarify this.  I actually wanted to specify it so that OpenCensus shim will use the globallly configured propogator (see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#get-global-propagator).   Right now the implementation DOES NOT do this, but I'd like to fix it.  Happy to hear thoughts/.copncerns.

My main reasoning for this shift from the Java impl is because  I'd like to give OTel configuration the ability to control the impl of the propogator.  One of the "principles" here is when adopting the shim, OTEL configuration should override OC, always.

specification/compatibility/opencensus.md
jsuereth: NO, this is related to selecting the W3C propagatoor off the global (see: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#get-global-propagator) and if there is none available, providing a default.

I'll take a stab at rewording this, need to think about a better phrasing.

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,192 @@
...
+Below are listed known incompatibilities between OpenTelemetry and OpenCensus
+specifications.   Applications leveraging unspecified behavior from OpenCensus
+that *is* specified incompatibly within OpenTelemetry are not eligble for
+using the OpenCensus <-> OpenTelemetry bridge.
+
+1. In OpenCensus, there is [no specification](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/Span.md#span-creation)
+   when parent spans can be specified on a child span.   OpenTelemetry specifies
+   that [parent spans must be specified during span creation](../trace/api.md#span-creation).
+   This leads to some issues with OpenCensus APIs that allowed flexible
+   speciifcaiton of parent spans post-initialization.
```
bogdandrutu, line 115: s/speciifcaiton/specification

bogdandrutu, line 115: I don't think any implementation does that. As the title says that is "span-creation" so I think this is just a misinterpretation of the OpenCensus specs.

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,192 @@
...
+   without the links.
+3. OpenTelemetry specifies that samplers are
+   [attached to an entire Trace provider](../trace/sdk.md#sampling)
+   while [OpenCensus allows custom samplers per span](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/Sampling.md#how-can-users-control-the-sampler-that-is-used-for-sampling).
+4. TraceFlags in both OpenCensus and OpenTelemetry only specify the single
+   `sampled` flag ([OpenTelemetry](../trace/api.md#spancontext),
+   [OpenCensus](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/TraceConfig.md#traceparams)).
+   Some OpenCensus APIs support "debug" and "defer" tracing flags in additon to
+   "sampled".  In this case, the OpenCensus bridge will do its best to support
+   and translate unspecified flags into the closest OpenTelemetry equivalent.
```
bogdandrutu, line 128: facepalm :)))

specification/compatibility/opencensus.md
bogdandrutu: Probably you may want to state this principal somewhere.

bogdandrutu (approved):
Some minor comments, LGTM

specification/compatibility/opencensus.md
bogdandrutu: https://github.com/open-telemetry/opentelemetry-go/issues/1515

specification/compatibility/opencensus.md
jsuereth: I'd encourage you to go look up the specifics here.   I know for Java, it's allowed by the API.

specification/compatibility/opencensus.md
jsuereth: Take look at the rewording, hopefully it's more clear.

specification/compatibility/opencensus.md
bogdandrutu: Prove me wrong please, I checked Go and Java

bogdandrutu: https://github.com/census-instrumentation/opencensus-go/blob/master/trace/trace.go#L173

https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/trace/Tracer.java#L307

jsuereth: Right, it was Typescript.  Sorry it's been a while since I wrote this: https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-core/src/trace/model/span.ts#L59

The point is, it's allowable by the API and we're not going to support it.

bogdandrutu: So you can call it that only typescript has this problem.

jsuereth: Python does too, and likely other dynamic languages.   Do we call out specific languages in other areas of the specification?

bogdandrutu: sure

bogdandrutu (approved):
Please add a status

jsuereth:
@bogdandrutu for "status" do you mean the stable, experimental etc?  I'll go look up the rules and annotate sections accordingly.

bogdandrutu:
@jsuereth yes what we call the `document status` (e.g. experimental, stable, etc.)

bogdandrutu:
@jsuereth ping :)

bogdandrutu:
@jsuereth ping :)

jsuereth:
@bogdandrutu Will get to this later today/tomorrow. Sorry for delays, had a lot of urgent things / Scala.Love conference prep to do last week.

jsuereth:
@bogdandrutu Ok, so I updated stability guarantees.   Here's the plan:

- This document declares "Trace" as Experimental, but "feature frozen".  We won't do any additional mapping beyond what's outlined in this document.
- If we discover more issues in "Trace" as we implement (or adapt) the compatibility shims in languages, we have room to fix the specification.
- Once we have N languages implemented against the specification we mark "Trace" as "Stable, Feature Freeze".
  - At least ONE dynamic language (Python, Node, etc. missing today)
  - At least ONE statically typed language (Java, Go, C#)
  - Note: Today we have shim prototypes in Go + Java
- We follow with Metric definition in similar fashion once metrics are stable in the OTEL spec.

Sound reasonable?

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,197 @@
...
+
+### Methods on Spans
+
+All specified methods in OpenCensus will delegate to the underlying `Span` of
+OpenTelemetry.
+
+#### Known Incompatibilities
+
+Below are listed known incompatibilities between OpenTelemetry and OpenCensus
+specifications.   Applications leveraging unspecified behavior from OpenCensus
```
reyang, line 111:
```suggestion
specifications.  Applications leveraging unspecified behavior from OpenCensus
```

reyang: There are other places in the doc that also have 3 SPCs.

specification/compatibility/opencensus.md
```diff
@@ -0,0 +1,197 @@
...
+[OpenCensus](https://opencensus.io) project in order to ease migration of
+instrumented codebases.
+
+This functionality will be provided as a bridge layer implementing the
+[OpenCensus API](https://github.com/census-instrumentation/opencensus-specs)
+using the OpenTelemetry API. This layer MUST NOT rely on implementation specific
+details of the OpenTelemetry SDK.
+
+More specifically, the intention is to allow OpenCensus instrumentation to be
+recorded using OpenTelemetry. This Shim Layer MUST NOT publicly expose any
```
reyang, line 17: Might need to clarify where the shim should go - in OpenTelemetry project or OpenCensus project.

https://github.com/open-telemetry/opentelemetry-specification/blob/9047c91412d3d4b7f28b0f7346d8c5034b509849/specification/versioning-and-stability.md#a-note-on-replacing-signals

_OpenTelemetry already supports two tracing APIs: OpenTelemetry and OpenTracing. We invented a new tracing API, but continue to support the old one._

reyang (approved):
LGTM as "Experimental".

specification/compatibility/opencensus.md
jsuereth: I can do.  In practice, we've actually released new OpenCensus versions that expose hooks for the shim, and the shim lives on the OTEL side (as that's where you configure your pipeline).

As an FYI for you, I expect we might have to do something similar in Python/JavaScript if we run into similar issues as what we found in Go/Java.

bogdandrutu:
@jsuereth please fix the build 👯

jsuereth:
@bogdandrutu fixed :)  Sorry, for some reason can't run that check locally.

bogdandrutu (approved):

jmacd (commented):
LGTM

PR #1335, opened 2021-01-13
Add Instrumentation to Glossary

DawgFan:
## Changes

PR intends to increase clarity and build consensus by adding the following terms to glossary:
- Manual Instrumentation
- Automatic Instrumentation

### DIFF
--- a/specification/glossary.md
+++ b/specification/glossary.md
@@ -51,7 +51,7 @@ Metrics, logs, and traces exported to telemetry backends are examples of out-of-
 
 ### Manual Instrumentation
 
-Coding against the OpenTelemetry API such as [Tracing API](trace/api.md) and [Metrics API](metrics/api.md) to collect telemetry from end-user code or shared frameworks (e.g. MongoDB, Redis, etc.).
+Coding against the OpenTelemetry API such as the [Tracing API](trace/api.md) and/or [Metrics API](metrics/api.md) to collect telemetry from end-user code or shared frameworks (e.g. MongoDB, Redis, etc.).
 
 ### Automatic Instrumentation
 
--- a/specification/glossary.md
+++ b/specification/glossary.md
@@ -50,20 +50,13 @@ usually asynchronously by background routines
 rather than from the critical path of the business logic.
 Metrics, logs, and traces exported to telemetry backends are examples of out-of-band data.
 
-### Instrumentation
-
-In the narrowest sense, coding against the OpenTelemetry [Tracing API](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md) or [Metrics API](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md), to collect telemetry from end user code or from shared frameworks.
-
 ### Manual Instrumentation
 
-Refers to OpenTelemetry enablement via code, which typically includes:
-- Importing the OpenTelemetry [Tracing API](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md) or [Metrics API](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md)
-- Importing the OpenTelemetry [tracing SDK](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md), [metrics SDK](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/sdk.md), or [logging SDK](https://github.com/open-telemetry/opentelemetry-specification/issues/894)
-- Importing and/or configuring one or more [exporter libraries](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/glossary.md#exporter-library)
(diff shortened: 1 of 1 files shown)


specification/glossary.md
```diff
@@ -47,6 +50,23 @@ usually asynchronously by background routines
 rather than from the critical path of the business logic.
 Metrics, logs, and traces exported to telemetry backends are examples of out-of-band data.
 
+### Instrumentation
+
+In the narrowest sense, coding against the OpenTelemetry [Tracing API](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md) or [Metrics API](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md).
```
yurishkuro, line 55:
```suggestion
In the narrowest sense, coding against the OpenTelemetry [Tracing API](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md) or [Metrics API](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md), to collect telemetry from end user code or from shared frameworks.
```

yurishkuro: Not sure I am getting the gist of "In the narrowest sense" prefix.

specification/glossary.md
```diff
@@ -47,6 +50,23 @@ usually asynchronously by background routines
...
+
+### Manual Instrumentation
+
+Refers to OpenTelemetry enablement via code, which typically includes:
+- Importing the OpenTelemetry [Tracing API](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md) or [Metrics API](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md)
+- Importing the OpenTelemetry [tracing SDK](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md), [metrics SDK](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/sdk.md), or [logging SDK](https://github.com/open-telemetry/opentelemetry-specification/issues/894)
+- Importing and/or configuring one or more [exporter libraries](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/glossary.md#exporter-library)
```
yurishkuro, line 62: I disagree with this definition, especially the last two points. Instrumentation, to me, means changing either business or framework code to generate telemetry via OTEL API. It can be done explicitly (manual instr), or implicitly (monkey-patching or byte code manipulation, auto-instr). It is, however, completely orthogonal to enabling and configuring SDKs - those activities are _not_ instrumentation.

specification/glossary.md
cijothomas: Is this better?
_"Instrumentation is the process of describing the telemetry signal, using one of the OpenTelemetry APIs."_

yurishkuro: I liked "coding" better than "describing". I could "describe" things by waving hands.

yurishkuro (approved):

specification/glossary.md
```diff
@@ -47,6 +49,16 @@ usually asynchronously by background routines
 rather than from the critical path of the business logic.
 Metrics, logs, and traces exported to telemetry backends are examples of out-of-band data.
 
+### Manual Instrumentation
+
+Coding against the OpenTelemetry [Tracing API](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md) or [Metrics API](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md), to collect telemetry from end-user code or shared frameworks (e.g. MongoDB, Redis, etc.).
```
cijothomas, line 54: Please use relative links, like trace/api.md

cijothomas: shared frameworks or libraries

cijothomas (approved):
Please replace absolute links with relative.

specification/glossary.md
cijothomas: Also - could you say OpenTelemetry API to cover more than just Tracing and Metrics alone? (like PropagatorsAPI)

cijothomas:
```suggestion
Coding against the OpenTelemetry API ([Tracing API](trace/api.md), [Metrics API](metrics/api.md) etc.), to collect telemetry 
from end-user code or shared frameworks, libraries (e.g. MongoDB, Redis, etc.).
```

specification/glossary.md
cijothomas: @DawgFan
We would also need to define a term to refer to the process of instrumenting (i.e leverage Otel.API) and setting up/configuring the SDK to enable actual telemetry flow. Would "OpenTelemetry enablement" be a good term? Manual enablement involves installing SDK and modifying code to enable it (typically involves setting up some exporter for telemetry). Auto-enablement would achieve the same, but without modifying code. (typically some env variables or config files).
Can be a separate PR :)

specification/glossary.md
DawgFan: Are "shared frameworks" and "libraries" and "Instrumented Libraries" synonymous? If so, any objection to using "instrumented libraries", which is already defined in the glossary?

SergeyKanzhelev (approved):

SergeyKanzhelev:
I wonder if we need to define the `pre-instrumented` or `built-in` instrumentation to differentiate from automatic instrumentation

SergeyKanzhelev:
If no concerns will merge tomorrow

specification/glossary.md
DawgFan: Thank you for the feedback Cijo. I agree there's an opportunity for a second PR.

PR #1339, opened 2021-01-14
Lowercased ECS launch types

willarmiros:
## Changes
* Lower case options for launch type in Amazon ECS spec for consistency with other lowercased values


willarmiros:
Looks like the semantic-conventions check is failing for an unrelated reason. Verified locally that `make table-generation` doesn't update anything.

anuraaga (approved):

arminru (approved):

bogdandrutu (approved):

bogdandrutu (commented):
Can you please add an entry in the changelog about this?

bogdandrutu:
> Can you please add an entry in the changelog about this?

For @willarmiros :)

willarmiros:
@bogdandrutu done :)

Not sure why I'm getting failures on semantic-conventions again :/

Oberon00:
The output says:
`File /spec/metrics/semantic_conventions/rpc.md contains a table that would be reformatted.`

willarmiros:
@Oberon00 Right, and I'm confused because I didn't touch `rpc.yaml` or `rpc.md` in my change, and the check runs just fine on my local branch:

```bash
$ docker run --rm -v $(pwd)/semantic_conventions:/source -v $(pwd)/specification:/spec otel/semconvgen -f /source markdown -md /spec --md-check
58 files left unchanged.
```

Oberon00:
@thisthat Do you have an idea what could be the problem?

thisthat:
> @thisthat Do you have an idea what could be the problem?

It looks like the wrong version is used. 
@willarmiros I suggest deleting the local image and let it download again from docker hub. You can do that by running `docker image rm otel/semconvgen`

Now that we have a release with a proper version, we should pin the semantic tool to [tag 0.2.1](https://hub.docker.com/layers/otel/semconvgen/0.2.1/images/sha256-0690afb3d065277d4c5c4b897f4fd9d8fb34c7577d5858c3eb838224eeed6d7d?context=explore).

willarmiros:
@thisthat you're right that was it, thanks! Looks like a lot of markdown just got updated with whitespace and other auto-generated changes.

PR #1340, opened 2021-01-14
Add OTEL_EXPORTER_ZIPKIN_TRANSPORT_FORMAT to sdk-environment-variables

### LINKED ISSUE #1336: Environment variable for zipkin exporter transport format
Add environment variable in [sdk env spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/sdk-environment-variables.md#zipkin-exporter) for zipkin exporter transport format (`OTEL_EXPORTER_ZIPKIN_TRANSPORT_FORMAT`) with some recommended default value. The known transport formats are `v1_json`, `v1_thrift`, `v2_json` and `v2_protobuf`.

lonewolf3739:
Fixes #1336 

## Changes

This change adds environment variable to specify transport format for Zipkin exporter with default value `V2_PROTOBUF`.

### DIFF
--- a/specification/sdk-environment-variables.md
+++ b/specification/sdk-environment-variables.md
@@ -72,9 +72,17 @@ See [OpenTelemetry Protocol Exporter Configuration Options](./protocol/exporter.
 
 ## Zipkin Exporter
 
-| Name                          | Description                | Default                                                                                                      |
-| ----------------------------- | -------------------------- | ------------------------------------------------------------------------------------------------------------ |
-| OTEL_EXPORTER_ZIPKIN_ENDPOINT | Endpoint for Zipkin traces | <!-- markdown-link-check-disable --> "http://localhost:9411/api/v2/spans"<!-- markdown-link-check-enable --> |
+| Name                                      | Description                           | Default                                                                                                       |
+| -----------------------------             | --------------------------            | ------------------------------------------------------------------------------------------------------------  |
+| OTEL_EXPORTER_ZIPKIN_ENDPOINT             | Endpoint for Zipkin traces            | <!-- markdown-link-check-disable --> "http://localhost:9411/api/v2/spans"<!-- markdown-link-check-enable -->  |
+| OTEL_EXPORTER_ZIPKIN_TRANSPORT_FORMAT     | Transport format for Zipkin traces    | V2_JSON                                                                                                       |
+
+Known values for OTEL_EXPORTER_ZIPKIN_TRANSPORT_FORMAT are:
+
+- `"V1_JSON"`
+- `"V1_THRIFT"`
+- `"V2_JSON"`
+- `"V2_PROTOBUF"`
 
 ## Prometheus Exporter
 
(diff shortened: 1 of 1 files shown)


anuraaga (approved):

specification/sdk-environment-variables.md
```diff
@@ -72,9 +72,17 @@ See [OpenTelemetry Protocol Exporter Configuration Options](./protocol/exporter.
...
-| ----------------------------- | -------------------------- | ------------------------------------------------------------------------------------------------------------ |
-| OTEL_EXPORTER_ZIPKIN_ENDPOINT | Endpoint for Zipkin traces | <!-- markdown-link-check-disable --> "http://localhost:9411/api/v2/spans"<!-- markdown-link-check-enable --> |
+| Name                                      | Description                           | Default                                                                                                       |
+| -----------------------------             | --------------------------            | ------------------------------------------------------------------------------------------------------------  |
+| OTEL_EXPORTER_ZIPKIN_ENDPOINT             | Endpoint for Zipkin traces            | <!-- markdown-link-check-disable --> "http://localhost:9411/api/v2/spans"<!-- markdown-link-check-enable -->  |
+| OTEL_EXPORTER_ZIPKIN_TRANSPORT_FORMAT     | Transport format for Zipkin traces    | V2_PROTOBUF                                                                                                   |
```
iNikem, line 78: Does this mean that all language SDKs have to support all these formats? Do they already? If not, how much work will it be to add this before GA?

lonewolf3739: It seems certain language SDKs already have support for some formats and not sure about the progress on others but everything mentioned here is already part of [spec-compliance-matrix](https://github.com/open-telemetry/opentelemetry-specification/blob/master/spec-compliance-matrix.md#exporters).

iNikem (approved):

tigrannajaryan (approved):

specification/sdk-environment-variables.md
jkwatson: I think @iNikem 's question is still a good one. Are all languages required to implement all protocols and transport formats in order to get to a 1.0 release? I don't think so, but if so, that's a very significant change to what maintainers believe are required for 1.0.

jsuereth:
We discussed this in the Specifcation SiG.  I'm capturing the resulting decisions/actions from that.

- Adding this environment variable would be "allowed-for-GA" but should not be done in 1.0 unless we can verify SDK compliance in time for 1.0.
- There will be *no* specified default for format on SDKs.   This means no change for 1.0 of trace is required, but there is still room to add this environment variable post-1.0 (if we can't get tthe change in before 1.0 in all SDKs).   Effectively, this change will be ok to add post 1.0.
- We will also remove the specification for the OTLP exporter format to be consistent with the decision on ZIPKIN here for 1.0.
- At a minimum, SDKs should already be documenting Zipkin exporter usage (and the format that is default), outside the specification.

I'll be opening a specification PR to update the OTLP exporter specification.

carlosalberto:
Re-opening again. I think it's about time we start working on it ;)

lonewolf3739:
@carlosalberto I see that `OTEL_EXPORTER_ZIPKIN_PROTOCOL` is added in https://github.com/open-telemetry/opentelemetry-specification/pull/1358. Is this still relevant?

carlosalberto:
@lonewolf3739 It has been reserved, but not defined yet, so yes, we definitely to work on that (non an urgent matter, although worth keeping it in mind).

reyang (approved):

seemk:
I think a similar env var, `OTEL_EXPORTER_JAEGER_TRANSPORT_FORMAT`,  is needed for Jaeger as well to allow configuring between `protobuf` and `thrift`.

PR #1341, opened 2021-01-14
Add `k8s.node` in semantic conventions

### LINKED ISSUE #1334: Include k8s.node in semantic conventions

I'd like to include `k8s.node.name` and `k8s.node.ip` in k8s semantic conventions.

**Additional context.**

Resources like `k8s.node` can provide more information to debug. For instance, we can tell which physical machine the data is coming from; `host.name` cannot do that since it already reveals the pod's name in the k8s Pod.

XSAM:
Resolve #1334

### DIFF
--- a/semantic_conventions/resource/k8s.yaml
+++ b/semantic_conventions/resource/k8s.yaml
@@ -10,6 +10,22 @@ groups:
           The name of the cluster.
         examples: ['opentelemetry-cluster']
 
+  - id: k8s.node
+    prefix: k8s.node
+    brief: >
+      A Kubernetes Node.
+    attributes:
+      - id: name
+        type: string
+        brief: >
+          The name of the node.
+        examples: ['opentelemetry-node']
+      - id: ip
+        type: string
+        brief: >
+          The address of the node (dotted decimal for IPv4 or [RFC5952](https://tools.ietf.org/html/rfc5952) for IPv6.
+        examples: ['127.0.0.1']
+
   - id: k8s.namespace
     prefix: k8s.namespace
     brief: >
--- a/specification/resource/semantic_conventions/k8s.md
+++ b/specification/resource/semantic_conventions/k8s.md
@@ -20,9 +20,22 @@ Kubernetes object, but "name" is usually more user friendly so can be also set.
 **Description:** A Kubernetes Cluster.
 
 <!-- semconv k8s.cluster -->
-| Attribute  | Type | Description  | Example  | Required |
+| Attribute  | Type | Description  | Examples  | Required |
 |---|---|---|---|---|
 | `k8s.cluster.name` | string | The name of the cluster. | `opentelemetry-cluster` | No |
+<!-- endsemconv -->
+
+## Node
+
+**type:** `k8s.node`
+
+**Description:** A Kubernetes Node.
+
+<!-- semconv k8s.node -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `k8s.node.name` | string | The name of the node. | `opentelemetry-node` | No |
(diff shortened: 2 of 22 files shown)


XSAM:
It looks like `make table-generation` made lots of changes.

Oberon00:
I think it would be important to first clarify what we want to have in `host.name`. Probably the hostname that the app itself sees when calling `gethostname`, etc. which would be the hostname of the innermost container in the innermost VM.

Oberon00:
> It looks like make table-generation made lots of changes.

@XSAM This was just fixed in #1342

specification/resource/semantic_conventions/k8s.md
```diff
@@ -25,6 +25,19 @@ Kubernetes object, but "name" is usually more user friendly so can be also set.
...
+
+**type:** `k8s.node`
+
+**Description:** A Kubernetes Node.
+
+<!-- semconv k8s.node -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `k8s.node.name` | string | The name of the node. | `opentelemetry-node` | No |
+| `k8s.node.ip` | string | The address of the node (dotted decimal for IPv4 or [RFC5952](https://tools.ietf.org/html/rfc5952) for IPv6. | `127.0.0.1` | No |
```
tigrannajaryan, line 38: Is this the external IP or internal IP? (see https://kubernetes.io/docs/concepts/architecture/nodes/#addresses). It may be worth explicitly using `k8s.node.externalip` and `k8s.node.internalip` instead.

semantic_conventions/resource/k8s.yaml
```diff
@@ -10,6 +10,22 @@ groups:
...
+    attributes:
+      - id: name
+        type: string
+        brief: >
+          The name of the node.
+        examples: ['opentelemetry-node']
+      - id: ip
+        type: string
+        brief: >
+          The address of the node (dotted decimal for IPv4 or [RFC5952](https://tools.ietf.org/html/rfc5952) for IPv6.
```
SergeyKanzhelev, line 26: Can you pls comment on what IP to prefer?

SergeyKanzhelev:
CC: @dashpole

specification/resource/semantic_conventions/k8s.md
XSAM: You're right; explicitly declare the address type might better.

Besides, I found k8s may have multiple `ExternalIP` and `InternalIP`. Maybe we should use `k8s.node.external_ips`and `k8s.node.internal_ips` and use `string[]` as type instead.

For instance: `['127.0.0.1', '127.0.0.2']`

tigrannajaryan: @XSAM can you give me a link to a doc which says there may be multiple ExternalIP and InternalIP? I haven't been able to find this in Kubernetes docs.

dashpole: I haven't been able to find any documentation specifically about multiple internal or external, but the [API definition](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#nodestatus-v1-core) definitely supports it.  I did find an issue that references someone setting multiple IPs of the same type as well: [kubernetes/kubernetes#61921](https://github.com/kubernetes/kubernetes/issues/61921)

tigrannajaryan: We do not use arrays for ip address elsewhere in semantic conventions (e.g. `net.peer.ip`) but using arrays here here looks justified.

I would prefer to avoid the complication and just have a single ip address, but I do not see a reasonable good behavior that uses just one of the ip addresses if the source of the data can have multiple.

bogdandrutu:
Closed by (#1390)

Issue #1184, opened 2020-11-02
Clarification required for Resource attribute values.


cijothomas:
The [ResourceCreation](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/resource/sdk.md#resource-creation) section says the `Attributes` for Resource follows the common attributes rules, which means value can be  string,bool,double,float, array of these.

When [Merge](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/resource/sdk.md#merge) faces a collision, it says if the primary has "empty string", override it. 
Question: What if primary is not even string? for eg: what if primary is array?

[RetrieveAttribute](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/resource/sdk.md#retrieve-attributes) section says the attributes names and values are just strings.
"_The attributes should consist of the name and values, both of which should be strings._"

[Specify Resource from ENV variable](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/resource/sdk.md#specifying-resource-information-via-an-environment-variable) says all attrubute values Must be considered strings.
"_All attribute values MUST be considered strings._"

Is the attribute value for Resource expected to contain just strings only?
Or should it support the [common Attribute](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/common/common.md#attributes)?
If common, then how should Merge handle collisions?
if common, then how to configure it using ENV variables?

Oberon00:
> When Merge faces a collision, it says if the primary has "empty string", override it.

I think we should remove this wording. It makes less sense now that resources support different attribute types.

Oberon00:
#1345 partially resolves this.

cijothomas:
The current wording in the spec addresses the question in this issue. We can close this.

PR #1345, opened 2021-01-15
Make Resource.Merge consistent with Span.SetAttribute.

Oberon00:
Required for #1346.
Resolves parts of #1184.

As agreed (except for the empty-string change) in the spec meeting, this is a pre-requesite for the updated default resource wording of #1294.

## Changes

- Remove special case for empty string when merging resources.
- Let the second argument take precedence, i.e. x.Merge(y) will have values from y on conflict.

### DIFF
--- a/specification/resource/sdk.md
+++ b/specification/resource/sdk.md
@@ -43,7 +43,7 @@ Required parameters:
 
 ### Merge
 
-The interface MUST provide a way for a old resource and a
+The interface MUST provide a way for an old resource and an
 updating resource to be merged into a new resource.
 
 Note: This is intended to be utilized for merging of resources whose attributes
--- a/specification/resource/sdk.md
+++ b/specification/resource/sdk.md
@@ -43,27 +43,21 @@ Required parameters:
 
 ### Merge
 
-The interface MUST provide a way for a primary resource and a
-secondary resource to be merged into a new resource.
+The interface MUST provide a way for a old resource and a
+updating resource to be merged into a new resource.
 
 Note: This is intended to be utilized for merging of resources whose attributes
 come from different sources,
 such as environment variables, or metadata extracted from the host or container.
 
 The resulting resource MUST have all attributes that are on any of the two input resources.
-Conflicts (i.e. a key for which attributes exist on both the primary and secondary resource)
-MUST be handled as follows:
-
-* If the value on the primary resource is an empty string, the result has the value of the secondary resource.
-* Otherwise, the value of the primary resource is used.
-
-Attribute key namespacing SHOULD be used to prevent collisions across different
-resource detection steps.
+If a key exists on both the old and updating resource, the value of the updating
+resource MUST be picked (even if the updated value is empty).
 
 Required parameters:
 
-- the primary resource whose attributes take precedence.
-- the secondary resource whose attributes will be merged in.
+- the old resource
+- the updating resource whose attributes take precedence
 
 ### The empty resource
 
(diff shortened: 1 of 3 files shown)


anuraaga (approved):
Yes please!

specification/resource/sdk.md
```diff
@@ -43,27 +43,21 @@ Required parameters:
 
 ### Merge
 
-The interface MUST provide a way for a primary resource and a
-secondary resource to be merged into a new resource.
+The interface MUST provide a way for a old resource and a
```
carlosalberto, line 46: *an old

carlosalberto (approved):

carlosalberto:
LGTM

specification/resource/sdk.md
Oberon00:
```suggestion
The interface MUST provide a way for an old resource and an
```

👍  also *an updating I think.

Oberon00:
Actually, with the new behavior, I would like to change the specified name from `Merge` to `UpdatedWith`. If you agree, please upvote this comment and I can change the PR.

anuraaga:
Preference for Merge over UpdatedWith since it's shorter.

For some reference, JS has the reasonably short `Object.assign` which is similar to the proposed behavior.

tigrannajaryan (approved):

yurishkuro (approved):

specification/resource/sdk.md
```diff
@@ -43,27 +43,21 @@ Required parameters:
 
 ### Merge
 
-The interface MUST provide a way for a primary resource and a
-secondary resource to be merged into a new resource.
+The interface MUST provide a way for an old resource and an
```
jkwatson, line 46:
```suggestion
The interface MUST provide a way for an existing resource and an
```

specification/resource/sdk.md
```diff
@@ -43,27 +43,21 @@ Required parameters:
...
 The resulting resource MUST have all attributes that are on any of the two input resources.
-Conflicts (i.e. a key for which attributes exist on both the primary and secondary resource)
-MUST be handled as follows:
-
-* If the value on the primary resource is an empty string, the result has the value of the secondary resource.
-* Otherwise, the value of the primary resource is used.
-
-Attribute key namespacing SHOULD be used to prevent collisions across different
-resource detection steps.
+If a key exists on both the old and updating resource, the value of the updating
```
jkwatson, line 54:
```suggestion
If a key exists on both the existing and updating resource, the value of the updating
```

jkwatson (approved):

specification/resource/sdk.md
Oberon00: The other resource also exists.

Oberon00: Maybe "updated", and rename "updating" to "new"?

Oberon00: Or "previous" and "new".

jkwatson: "old" makes it sound like it's.... old, which is might not be at all.

Oberon00: I named it "old" to emphasize that the values in there may be overwritten by the other resource. But I agree that the name is a bit odd. Maybe we could specify it as a member function `resource.Merge(updating)`. Then we don't have to come up with a name for the left-hand-side.

cijothomas (approved):

specification/resource/sdk.md
```diff
@@ -43,27 +43,21 @@ Required parameters:
...
+resource MUST be picked (even if the updated value is empty).
```
bogdandrutu, line 55: Are we sure about this?

Oberon00: About the behavior for empty strings? I hope so, it is one special case we lose.

carlosalberto: Ping @bogdandrutu ;)

carlosalberto:
@Oberon00 Please rebase your branch ;)

Oberon00:
@carlosalberto Done.

PR #1358, opened 2021-01-20
Modify Exporter format variables for 1.0 (allowing further specification post 1.0)

jsuereth:
Adapts #1340

## Changes
- Reserve ZIPKIN format environment variable for post 1.0 implementation.
- Update specification to keep OTLP protocol + ZIPKIN protocol story consistent for 1.0 (request from @bogdandrutu, note: seeing this, I'm not sure I agree, looking for feedback)

Related issues #1336

### DIFF
--- a/specification/protocol/exporter.md
+++ b/specification/protocol/exporter.md
@@ -64,7 +64,7 @@ the future:
 - `OTEL_EXPORTER_OTLP_TRACE_PROTOCOL`
 - `OTEL_EXPORTER_OTLP_METRIC_PROTOCOL`
 
-SDKs are free to choose a default, if no configuration is provided.
+SDKs have an unspecified default, if no configuration is provided.
 
 ### Specifying headers via environment variables
 
--- a/specification/protocol/exporter.md
+++ b/specification/protocol/exporter.md
@@ -9,7 +9,6 @@ The following configuration options MUST be available to configure the OTLP expo
 | Configuration Option | Description                                                  | Default           | Env variable                                                 |
 | -------------------- | ------------------------------------------------------------ | ----------------- | ------------------------------------------------------------ |
 | Endpoint             | Target to which the exporter is going to send spans or metrics. The endpoint MUST be a valid URL with scheme (http or https) and host, and MAY contain a port and path. A scheme of https indicates a secure connection. When using `OTEL_EXPORTER_ENDPOINT` with OTLP/HTTP, exporters SHOULD follow the collector convention of appending the version and signal to the path (e.g. `v1/traces` or `v1/metrics`). The per-signal endpoint configuration options take precedence and can be used to override this behavior. See the [OTLP Specification][otlphttp-req] for more details. | `https://localhost:4317` | `OTEL_EXPORTER_OTLP_ENDPOINT` `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` |
-| Protocol             | The protocol used to transmit the data. One of `grpc`,`http/json`,`http/protobuf`. | `grpc`               | `OTEL_EXPORTER_OTLP_PROTOCOL` `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL` |
 | Certificate File     | Path to certificate file for TLS credentials of gRPC client. Should only be used for a secure connection. | n/a               | `OTEL_EXPORTER_OTLP_CERTIFICATE` `OTEL_EXPORTER_OTLP_TRACES_CERTIFICATE` `OTEL_EXPORTER_OTLP_METRICS_CERTIFICATE` |
 | Headers              | Key-value pairs to be used as headers associated with gRPC or HTTP requests. See [Specifying headers](./exporter.md#specifying-headers-via-environment-variables) for more details.                   | n/a               | `OTEL_EXPORTER_OTLP_HEADERS` `OTEL_EXPORTER_OTLP_TRACES_HEADERS` `OTEL_EXPORTER_OTLP_METRICS_HEADERS` |
 | Compression          | Compression key for supported compression types. Supported compression: `gzip`| No value              | `OTEL_EXPORTER_OTLP_COMPRESSION` `OTEL_EXPORTER_OTLP_TRACES_COMPRESSION` `OTEL_EXPORTER_OTLP_METRICS_COMPRESSION` |
@@ -52,6 +51,20 @@ export OTEL_EXPORTER_OTLP_PROTOCOL=grpc
 export OTEL_EXPORTER_OTLP_METRICS_ENDPOINT=http://collector.example.com/v1/metrics
 export OTEL_EXPORTER_OTLP_METRICS_PROTOCOL=http/json
 ```
+
+### Specify Protocol
+
+Currently, OTLP haas more than one transport protocol it can support, e.g. 
+`grpc`,  `http/json`, `http/protobuf`.   As of 1.0 of the specification, there 
+*is no specified default, or configuration via environment variables*.  We
+reserve the following environment variables for configuration of protocols in
+the future:
+
+- `OTEL_EXPORTER_OTLP_PROTOCOL`
+- `OTEL_EXPORTER_OTLP_SPAN_PROTOCOL`
+- `OTEL_EXPORTER_OTLP_METRIC_PROTOCOL`
+
+SDKs are free to choose a default, if no configuration is provided.
 
 ### Specifying headers via environment variables
 
--- a/specification/protocol/exporter.md
+++ b/specification/protocol/exporter.md
@@ -61,7 +61,7 @@ reserve the following environment variables for configuration of protocols in
 the future:
 
 - `OTEL_EXPORTER_OTLP_PROTOCOL`
-- `OTEL_EXPORTER_OTLP_SPAN_PROTOCOL`
+- `OTEL_EXPORTER_OTLP_TRACE_PROTOCOL`
 - `OTEL_EXPORTER_OTLP_METRIC_PROTOCOL`
 
 SDKs are free to choose a default, if no configuration is provided.
--- a/specification/protocol/exporter.md
+++ b/specification/protocol/exporter.md
@@ -54,7 +54,7 @@ export OTEL_EXPORTER_OTLP_METRICS_PROTOCOL=http/json
 
 ### Specify Protocol
 
-Currently, OTLP haas more than one transport protocol it can support, e.g.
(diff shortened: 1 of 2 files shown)


iNikem:
Can we define environment variables for protocols (both OTLP and Zipkin) together with possible values for those variables, but NOT require SDKs to implement ALL of them before 1.0?

specification/protocol/exporter.md
```diff
@@ -53,6 +52,20 @@ export OTEL_EXPORTER_OTLP_METRICS_ENDPOINT=http://collector.example.com/v1/metri
 export OTEL_EXPORTER_OTLP_METRICS_PROTOCOL=http/json
 ```
 
+### Specify Protocol
+
+Currently, OTLP haas more than one transport protocol it can support, e.g.
```
anuraaga, line 57:
```suggestion
Currently, OTLP has more than one transport protocol it can support, e.g.
```

jsuereth:
> Can we define environment variables for protocols (both OTLP and Zipkin) together with possible values for those variables, but NOT require SDKs to implement ALL of them before 1.0?

@iNikem I think for OTLP yes.  For Zipkin, there was some concern over what things meant.  I think it'd be fine to open a  PR following this one where that can be discussed, but the idea is to allow it to be specified after the 1.0 next week, in the event the discussion doesn't finish in time for 1.0.

specification/protocol/exporter.md
```diff
@@ -53,6 +52,20 @@ export OTEL_EXPORTER_OTLP_METRICS_ENDPOINT=http://collector.example.com/v1/metri
...
+### Specify Protocol
+
+Currently, OTLP has more than one transport protocol it can support, e.g.
+`grpc`,  `http/json`, `http/protobuf`.   As of 1.0 of the specification, there
+*is no specified default, or configuration via environment variables*.  We
+reserve the following environment variables for configuration of protocols in
+the future:
+
+- `OTEL_EXPORTER_OTLP_PROTOCOL`
+- `OTEL_EXPORTER_OTLP_SPAN_PROTOCOL`
```
obecny, line 64: should this be:
```suggestion
- `OTEL_EXPORTER_OTLP_TRACE_PROTOCOL`
```
?

andrewhsu:
@jsuereth is this PR ready for review?

jsuereth:
> @jsuereth is this PR ready for review?

yes!  Sorry, didn't mean to leave WIP on it for that long.

carlosalberto (approved):

iNikem:
I am concerned that if SIGs can choose their own default protocols then, after we update the spec, SIGs will have to change their defaults, which will be a breaking change. What am I missing?

jsuereth:
@!iNIkem we're not going to specify the default.  If you want to force a protocol, you'll need to use the variable or configuration.

specification/protocol/exporter.md
```diff
@@ -53,6 +52,20 @@ export OTEL_EXPORTER_OTLP_METRICS_ENDPOINT=http://collector.example.com/v1/metri
...
+`grpc`,  `http/json`, `http/protobuf`.   As of 1.0 of the specification, there
+*is no specified default, or configuration via environment variables*.  We
+reserve the following environment variables for configuration of protocols in
+the future:
+
+- `OTEL_EXPORTER_OTLP_PROTOCOL`
+- `OTEL_EXPORTER_OTLP_TRACE_PROTOCOL`
+- `OTEL_EXPORTER_OTLP_METRIC_PROTOCOL`
+
+SDKs are free to choose a default, if no configuration is provided.
```
anuraaga, line 67: @jsuereth the issue seems to be this. If sdks can pick a default, they can end up with different defaults. The intent of the spec has generally been to have shared behavior outside of api idioms. For something completely contained in the SDK it could still be ok, but protocol leaks out. For example, if given an app written in lang1 and another in lang2, we would expect both to connect to the same collector with default settings. But if the collector only enables proto, and lang2 only supports json, the behavior of the defaults changes drastically.

No snark intended, if even JS, the most json-oriented language around, can support gRPC, I'm hoping we could converge on it being a required default to prevent surprises.  Having no default, rather than an SDK-specific default, therefore requiring explicit user configuration seems fine though.

jsuereth: I agree that this isn't an ideal situation.   This is called "limiting what we do for 1.0, and providing a stable release".   We'll improve over time.  When we're ready to make a breaking change (and understand this problem better) we could specify a default.

The proposal here is to unblock 1.0 and allow something better to come later.

carlosalberto: To be clear: the plan is to improve this done by the time we do a 'full' release along with stable Metrics (or maybe even sooner). Hopefully we will triage the remaining bits accordingly ;)

andrewhsu:
discussed at the sig mtg today, and it is desired to have a resolution merged for GA so applying labels to this PR

specification/protocol/exporter.md
bogdandrutu: @jsuereth if we say that SDK should not have a default and require this? It will be annoying for the users while we fix this, but this will make us fix it sooner than later and we have all the freedom?

carlosalberto:
@open-telemetry/specs-approvers Please review/approve this PR ;)

specification/protocol/exporter.md
jmacd: I suggest calling this an "unspecified default".

jmacd (approved):

carlosalberto:
@open-telemetry/specs-approvers Please review (or approve) this one ;)

specification/protocol/exporter.md
carlosalberto: Ping @bogdandrutu

anuraaga:
While I understand the pressure to release something, I still feel like this is brushing under the rug too much. We've identified a point that could create compatibility problems - if we ignore it, we'll have even more to deal with than the issues we didn't think about in advance which are definitely going to appear!

Rather than converging on an ambiguous spec, no spec should be better right? These options come to mind

- Don't mark environment variables as stable for the 1.0 release. These are the last thing anyone implements - the fact that we find issues late in the game for them is understandable, and just reflects low baking time. Other features are better baked and could go out first. Languages add disclaimers in LOUD WORDS that the variables may change until a future date.

- Don't mark environment variables stable ever. From what I've seen people really get unhappy when a breaking change forces them to fix their code and rebuild it again. But breaking changes in environment variables do not have this aspect, if a change was needed a rebuild is never required. Never committing to env vars on breaking changes means even if languages have different defaults now, they could converge in the future. I think this lack of stability needs to be more clearly called out than just an unspecified default, and could apply generally, not just this particular change.

And btw if the intent of this change is along these lines, than I think it's not explicit enough - it should make clear what we want the languages to commit to in terms of messaging about their stability guarantees.

iNikem:
I have nothing to say about @anuraaga concerns, but I disagree with your options :) I think we have to declare env variable stable.

anuraaga:
> I think we have to declare env variable stable.

Do you think this means prioritizing stabilizing them over consistency among languages which is questionable with unspecified defaults? If we're good with it then that's fine too (it's been so hard to get divergence among languages of API concerns didn't think env vars would be so easy ;)

iNikem:
> > I think we have to declare env variable stable.
> 
> Do you think this means prioritizing stabilizing them over consistency among languages which is questionable with unspecified defaults? If we're good with it then that's fine too (it's been so hard to get divergence among languages of API concerns didn't think env vars would be so easy ;)

Yes, I think stable 1.0 which includes env variables is more important that consistency of unspecified behaviour across languages :)

specification/protocol/exporter.md
bogdandrutu: Ok, hopefully everyone understands this way :) Right now it is required for user to specify the values and SDK will not work without a value specified :) otherwise they actually define a default.

bogdandrutu:
@jsuereth please rebase

jsuereth:
> Languages add disclaimers in LOUD WORDS that the variables may change until a future date.

This is only as effective as users reading the docs.  In my experience, users will ignore all "warnings" if they have a strong need for something and then be upset with you when you break it.   We only have so much friction we can cause and this doesn't seem like a good use of social capital.

> * Don't mark environment variables stable ever. From what I've seen people really get unhappy when a breaking change forces them to fix their code and rebuild it again. But breaking changes in environment variables do not have this aspect, if a change was needed a rebuild is never required. Never committing to env vars on breaking changes means even if languages have different defaults now, they could converge in the future. I think this lack of stability needs to be more clearly called out than just an unspecified default, and could apply generally, not just this particular change.

This hasn't been my experience.  Also, I don't think anything is locked down such that it can't be slowly evolved/migrated.  We *should* be avoiding unnecessary churn (e.g. ENV variable renames) post 1.0 because it just causes friction and unrest for users.  If we really want this project to be stable and widely adopted, then we need to make sure users are succesful and not surprised.   Changing environment variables is probably one of the MOST surprising things, especially if you switch from something on-default to something default.

> And btw if the intent of this change is along these lines, than I think it's not explicit enough - it should make clear what we want the languages to commit to in terms of messaging about their stability guarantees

Suggestions welcome.   The intent here was to specify a minimum (what we felt was already common) so SDKs won't diverge past that minimum, but not spin cycles gaining consensus on the full specification.  

Regarding whether ENVIRONMENT variables are considered in Semantic Conventions + versioning, I think that belongs in a different PR than this, but I'm happy to take a stab at it.

jsuereth:
@bogdandrutu it's rebased now.

anuraaga (approved):

bogdandrutu (approved):

tigrannajaryan (approved):

cijothomas (approved):

carlosalberto:
We have enough approvals. Let's merge this by the end of the day 😃

codeboten (approved):

arminru (commented):
Please add an entry to `CHANGELOG.md`. Thanks!

specification/protocol/exporter.md
```diff
@@ -53,6 +52,20 @@ export OTEL_EXPORTER_OTLP_METRICS_ENDPOINT=http://collector.example.com/v1/metri
...
+### Specify Protocol
+
+Currently, OTLP has more than one transport protocol it can support, e.g.
+`grpc`,  `http/json`, `http/protobuf`.   As of 1.0 of the specification, there
+*is no specified default, or configuration via environment variables*.  We
+reserve the following environment variables for configuration of protocols in
+the future:
+
+- `OTEL_EXPORTER_OTLP_PROTOCOL`
+- `OTEL_EXPORTER_OTLP_TRACE_PROTOCOL`
```
codeboten, line 64: Should this be `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL`?

bogdandrutu: Yes please :)

carlosalberto:
Merging to expedite things. Will cook a follow-up PR myself.

PR #1365, opened 2021-01-22
sdk env: change OTEL_PROPAGATORS to OTEL_TEXT_MAP_PROPAGATORS to be explicit

tsloughter:
## Changes

This changes the name of the propagators OS environment variable to `OTEL_TEXT_MAP_PROPAGATORS ` to match with what it is actually setting up, https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/api-propagators.md#textmap-propagator

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,8 @@ Updates:
 - Resource SDK: Reverse (suggested) order of Resource.Merge parameters, remove
   special case for empty strings
   ([#1345](https://github.com/open-telemetry/opentelemetry-specification/pull/1345))
+- Change environment variable `OTEL_PROPAGATORS` to `OTEL_TEXT_MAP_PROPAGATORS`
+  ([#1365](https://github.com/open-telemetry/opentelemetry-specification/pull/1365))
 
 ## v0.7.0 (11-18-2020)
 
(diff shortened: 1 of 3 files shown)


carlosalberto:
So in theory the binary (or other upcoming formats) could be differentiated by their names itself, i.e.

```sh
OTEL_PROPAGATORS=tracecontext,binarytracecontext,b3,etc
```

The SDK in this case would know already the specified format's type. Or is this too bad?

tsloughter:
I guess it depends on how implementations are using them. My understanding, and how I implemented it, is that "text map propagators" are stored as a list and binary a separate list. And then a instrumented library, like an HTTP client or server, would pass headers to the "text map propagators".

This separation is why I assumed it better to have them set through separate environment variables.

Not that as the value of `$OTEL_PROPAGATORS` is parsed I can't hardcode its type and put it in the correct list of propagators. I'm fine doing that, the spec just made it seem like it was only for "text map" versions so I had gone about it the simpler way.

iNikem:
I agree with @carlosalberto that there is no need for a separate environment variable for every possible "type" of propagators.

tsloughter:
Sounds like I'm the only one who thinks its simpler to split them :), I'll close this PR.

PR #1374, opened 2021-01-25
Fix TODOs in Jaeger exporter spec

yurishkuro:
Provide additional details for mappings needed to be supported in Jaeger exporters.

Related issues #1237

### DIFF
--- a/specification/trace/sdk_exporters/jaeger.md
+++ b/specification/trace/sdk_exporters/jaeger.md
@@ -6,7 +6,10 @@ Jaeger accepts spans in two formats:
 * Thrift `Batch`, defined in [jaeger-idl/.../jaeger.thrift](https://github.com/jaegertracing/jaeger-idl/blob/master/thrift/jaeger.thrift), accepted via UDP or HTTP
 * Protobuf `Batch`, defined in [jaeger-idl/.../model.proto](https://github.com/jaegertracing/jaeger-idl/blob/master/proto/api_v2/model.proto), accepted via gRPC
 
-See also: [Jaeger APIs](https://www.jaegertracing.io/docs/latest/apis/).
+See also:
+
+* [Jaeger APIs](https://www.jaegertracing.io/docs/latest/apis/)
+* [Reference implementation of this translation in the OpenTelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector/blob/master/translator/trace/jaeger/traces_to_jaegerproto.go)
 
 ## Summary
 
--- a/specification/trace/sdk_exporters/jaeger.md
+++ b/specification/trace/sdk_exporters/jaeger.md
@@ -117,10 +117,11 @@ Array values MUST be serialized to string like a JSON list as mentioned in
 OpenTelemetry `Link`(s) MUST be converted to `SpanReference`(s) in Jaeger,
 using `FOLLOWS_FROM` reference type. The Link's attributes cannot be represented
 in Jaeger explicitly. The exporter MAY convert link attributes to span `Log`(s):
-  * use Span start time as the timestamp of the Log
-  * set Log tag `event=link`
-  * set Log tags `trace_id` and `span_id` from `SpanContext`'s fields
-  * store `Link`'s attributes and Log tags
+
+* use Span start time as the timestamp of the Log
+* set Log tag `event=link`
+* set Log tags `trace_id` and `span_id` from `SpanContext`'s fields
+* store `Link`'s attributes and Log tags
 
 ### Events
 
--- a/specification/trace/sdk_exporters/jaeger.md
+++ b/specification/trace/sdk_exporters/jaeger.md
@@ -102,7 +102,7 @@ OpenTelemetry `SpanKind` field MUST be encoded as `span.kind` tag in Jaeger span
 
 In Jaeger Thrift format the timestamps and durations MUST be represented in
 microseconds (since epoch for timestamps). If the original value in OpenTelemetry
-is expressed in nanoseconds, it MUST be rounded to microseconds.
+is expressed in nanoseconds, it MUST be rounded or truncated to microseconds.
 
 In Jaeger Proto format the timestamps and durations MUST be represented
 with nanosecond precision using `google.protobuf.Timestamp` and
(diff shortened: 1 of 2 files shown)


specification/trace/sdk_exporters/jaeger.md
```diff
@@ -36,9 +50,41 @@ OpenTelemetry Span's `InstrumentationLibrary` MUST be reported as span `tags` to
 | `InstrumentationLibrary.name`|`otel.library.name`|
 | `InstrumentationLibrary.version`|`otel.library.version`|
 
-### Attribute
+### Parent ID
+
+Jaeger Thrift format allows capturing parent ID in a top-level Span field.
+Jaeger Proto format does not support parent ID field; instead the parent
+must be recorded as a `SpanReference` of type `CHILD_OF`, e.g.:
```
carlosalberto, line 57: MUST?

carlosalberto (approved):

carlosalberto:
cc @jkwatson

specification/trace/sdk_exporters/jaeger.md
```diff
@@ -36,9 +50,63 @@ OpenTelemetry Span's `InstrumentationLibrary` MUST be reported as span `tags` to
...
+| `SpanKind.CONSUMER`|`"consumer"`|
+| `SpanKind.PRODUCER`|`"producer"` |
+| `SpanKind.INTERNAL`| do not add `span.kind` tag |
 
-TBD
+### Unit of time
+
+In Jaeger Thrift format the timestamps and durations MUST be represented in
+microseconds (since epoch for timestamps). If the original value in OpenTelemetry
+is expressed in nanoseconds, it MUST be rounded to microseconds.
```
anuraaga, line 105: Is floor acceptable instead of rounding mathematically? e.g., what happens normally when doing `TimeUnit.NANOSECONDS.toMicros` in Java

specification/trace/sdk_exporters/jaeger.md
```diff
@@ -54,9 +122,29 @@ The following table defines the OpenTelemetry `Status` to Jaeger `tags` mapping.
...
+### Links
+
+OpenTelemetry `Link`(s) MUST be converted to `SpanReference`(s) in Jaeger,
+using `FOLLOWS_FROM` reference type. The Link's attributes cannot be represented
+in Jaeger explicitly. The exporter MAY convert link attributes to span `Log`(s):
+
+* use Span start time as the timestamp of the Log
+* set Log tag `event=link`
+* set Log tags `trace_id` and `span_id` from `SpanContext`'s fields
+* store `Link`'s attributes and Log tags
```
anuraaga, line 146:
```suggestion
* store `Link`'s attributes as Log tags
```

specification/trace/sdk_exporters/jaeger.md
```diff
@@ -54,9 +122,29 @@ The following table defines the OpenTelemetry `Status` to Jaeger `tags` mapping.
...
+Primitive types MUST be represented by the corresponding types of Jaeger tags.
+
+Array values MUST be serialized to string like a JSON list as mentioned in
+[semantic conventions](../../overview.md#semantic-conventions).
+
+### Links
+
+OpenTelemetry `Link`(s) MUST be converted to `SpanReference`(s) in Jaeger,
+using `FOLLOWS_FROM` reference type. The Link's attributes cannot be represented
+in Jaeger explicitly. The exporter MAY convert link attributes to span `Log`(s):
```
anuraaga, line 141:
```suggestion
in Jaeger explicitly. The exporter MAY convert links to span `Log`(s) to preserve attributes:
```

Looking at the below conversion, it seems like it is more than just attributes being stored in Log (trace_id, span_id)

specification/trace/sdk_exporters/jaeger.md
yurishkuro: +1

anuraaga (approved):

specification/trace/sdk_exporters/jaeger.md
```diff
@@ -36,9 +50,63 @@ OpenTelemetry Span's `InstrumentationLibrary` MUST be reported as span `tags` to
...
+
+OpenTelemetry `SpanKind` field MUST be encoded as `span.kind` tag in Jaeger span.
+
+| OpenTelemetry | Jaeger |
+| ------------- | ------ |
+| `SpanKind.CLIENT`|`"client"`|
+| `SpanKind.SERVER`|`"server"`|
+| `SpanKind.CONSUMER`|`"consumer"`|
+| `SpanKind.PRODUCER`|`"producer"` |
+| `SpanKind.INTERNAL`| do not add `span.kind` tag |
```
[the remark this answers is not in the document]
yurishkuro: It's fine either way, there's no special processing for "internal". We can allow both, or for consistency require "internal".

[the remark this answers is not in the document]
yurishkuro: My preference is to not include internal since it increases data size without adding any useful info.

specification/trace/sdk_exporters/jaeger.md
```diff
@@ -54,9 +122,29 @@ The following table defines the OpenTelemetry `Status` to Jaeger `tags` mapping.
...
+
+Primitive types MUST be represented by the corresponding types of Jaeger tags.
+
+Array values MUST be serialized to string like a JSON list as mentioned in
+[semantic conventions](../../overview.md#semantic-conventions).
+
+### Links
+
+OpenTelemetry `Link`(s) MUST be converted to `SpanReference`(s) in Jaeger,
+using `FOLLOWS_FROM` reference type. The Link's attributes cannot be represented
```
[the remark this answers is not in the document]
yurishkuro: What you're describing is the behavior needed to represent ParentID, and I will clarify that it must be the first in the list. But this is a section for additional Links.

yurishkuro:
btw, @tigrannajaryan , I wouldn't mind pointing to Collector's code as the reference implementation of this conversion, any concerns?

tigrannajaryan:
> btw, @tigrannajaryan , I wouldn't mind pointing to Collector's code as the reference implementation of this conversion, any concerns?

@yurishkuro No concerns, I think it is a good idea. Here is the link: https://github.com/open-telemetry/opentelemetry-collector/blob/master/translator/trace/jaeger/traces_to_jaegerproto.go

specification/trace/sdk_exporters/jaeger.md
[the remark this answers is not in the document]
tigrannajaryan: If you do not mind I would prefer not to change Collector's translation since I am worried that it may break people's stuff who rely on the current behavior (unlikely but possible). So, if using "internal" does not result in any problems for Jaeger let's allow it.

PR #1362, opened 2021-01-22
Consolidate OTLP variables to use TRACES / METRICS to describe talking…

anuraaga:
… to TraceService / MetricsService

Figured we need a PR to try to knock these out :)

Fixes #1352
Fixes #1353

## Changes

Environment variables for configuring OTLP exporters are changed from SPAN / METRIC to TRACE / METRICS. This is to correspond to the services being connected to, TraceService and MetricsService. It seems natural, and leaves out ambiguity by just matching up with the service names. Hopefully these variables won't be used all that much anyways in favor of the polyglot variables.

PR #1375, opened 2021-01-26
Add configuration for client TLS auth and rename variable for trusted…

### LINKED ISSUE #1363: Environment variables for OTLP client TLS key / certificate

Ability to set client key / certificate for mTLS.

**Additional context.**

We currently only have a CERTIFICATE_FILE environment variable, which is a bit misleading since this only refers to trusted certificates. For mTLS, there would need to be two additional variables, one for the client certificate and one for the client private key. I think we should have three variables

OTEL_EXPORTER_OTLP_TLS_PRIVATE_KEY
OTEL_EXPORTER_OTLP_TLS_CERTIFICATE
OTEL_EXPORTER_OTLP_TLS_TRUSTED_CERTIFICATE

Reference: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1829

anuraaga:
… cert of server.

Fixes #1363

## Changes

Adds variables for configuring client TLS auth and rename / clarify the current configuration for a server's trusted certificate.


iNikem:
Changelog?

anuraaga:
@iNikem Yeah generally add it after some approvals since until then it just causes daily merge conflicts :P

iNikem (approved):

specification/protocol/exporter.md
```diff
@@ -10,7 +10,8 @@ The following configuration options MUST be available to configure the OTLP expo
...
 | Protocol             | The protocol used to transmit the data. One of `grpc`,`http/json`,`http/protobuf`. | `grpc`               | `OTEL_EXPORTER_OTLP_PROTOCOL` `OTEL_EXPORTER_OTLP_SPAN_PROTOCOL` `OTEL_EXPORTER_OTLP_METRIC_PROTOCOL` |
-| Certificate File     | Path to certificate file for TLS credentials of gRPC client. Should only be used for a secure connection. | n/a               | `OTEL_EXPORTER_OTLP_CERTIFICATE` `OTEL_EXPORTER_OTLP_SPAN_CERTIFICATE` `OTEL_EXPORTER_OTLP_METRIC_CERTIFICATE` |
+/ Client certificate / key | The TLS certificate and private key to authenticate the client for mTLS. Should only be used for a secure connection.. | n/a | `OTLP_EXPORTER_OTLP_TLS_CERTIFICATE` `OTLP_EXPORTER_OTLP_SPAN_TLS_CERTIFICATE` `OTLP_EXPORTER_OTLP_METRIC_TLS_CERTIFICATE`, `OTLP_EXPORTER_OTLP_TLS_PRIVATE_KEY` `OTLP_EXPORTER_OTLP_SPAN_TLS_PRIVATE_KEY` `OTLP_EXPORTER_OTLP_METRIC_TLS_PRIVATE_KEY` |
```
arminru, line 13:
```suggestion
| Client certificate / key | The TLS certificate and private key to authenticate the client for mTLS. Should only be used for a secure connection. | n/a | `OTEL_EXPORTER_OTLP_TLS_CERTIFICATE` `OTEL_EXPORTER_OTLP_SPAN_TLS_CERTIFICATE` `OTEL_EXPORTER_OTLP_METRIC_TLS_CERTIFICATE`, `OTEL_EXPORTER_OTLP_TLS_PRIVATE_KEY` `OTEL_EXPORTER_OTLP_SPAN_TLS_PRIVATE_KEY` `OTEL_EXPORTER_OTLP_METRIC_TLS_PRIVATE_KEY` |
```

I suppose you didn't rename it to `OTLP_EXPORTER_OTLP_...` on purpose ;-)

arminru: Would it make sense to split this in two lines rather than listing 6 vars at once?

specification/protocol/exporter.md
```diff
@@ -10,7 +10,8 @@ The following configuration options MUST be available to configure the OTLP expo
...
+| Trusted certificate     | The trusted certificate to use when verifying a server's TLS credentials. Should only be used for a secure connection. | n/a               | `OTEL_EXPORTER_OTLP_TLS_TRUSTED_CERTIFICATE` `OTEL_EXPORTER_OTLP_SPAN_TLS_TRUSTED_CERTIFICATE` `OTEL_EXPORTER_OTLP_METRIC_TLS_TRUSTED_CERTIFICATE` |
```
arminru, line 14: As per #1362 this should all be `_TRACE_` instead of `_SPAN_`. Please don't forget updating the PR that gets merged last.

specification/protocol/exporter.md
anuraaga: Hmm - I think it makes sense to group them when thinking about an option. Both have to be configured at the same time or it doesn't make sense, e.g., the method should be something like `setClientCertificate(certificate, privateKey)`. But that means the environment variables end up grouped together and long. I lean towards this just a bit.

tigrannajaryan:
Lets use `TRACES` instead of `SPAN` as it is proposed by https://github.com/open-telemetry/opentelemetry-specification/pull/1362

specification/protocol/exporter.md
```diff
@@ -10,7 +10,8 @@ The following configuration options MUST be available to configure the OTLP expo
...
 | Protocol             | The protocol used to transmit the data. One of `grpc`,`http/json`,`http/protobuf`. | `grpc`               | `OTEL_EXPORTER_OTLP_PROTOCOL` `OTEL_EXPORTER_OTLP_SPAN_PROTOCOL` `OTEL_EXPORTER_OTLP_METRIC_PROTOCOL` |
-| Certificate File     | Path to certificate file for TLS credentials of gRPC client. Should only be used for a secure connection. | n/a               | `OTEL_EXPORTER_OTLP_CERTIFICATE` `OTEL_EXPORTER_OTLP_SPAN_CERTIFICATE` `OTEL_EXPORTER_OTLP_METRIC_CERTIFICATE` |
+| Client certificate / key | The TLS certificate and private key to authenticate the client for mTLS. Should only be used for a secure connection. | n/a | `OTEL_EXPORTER_OTLP_TLS_CERTIFICATE` `OTEL_EXPORTER_OTLP_SPAN_TLS_CERTIFICATE` `OTEL_EXPORTER_OTLP_METRIC_TLS_CERTIFICATE`, `OTEL_EXPORTER_OTLP_TLS_PRIVATE_KEY` `OTEL_EXPORTER_OTLP_SPAN_TLS_PRIVATE_KEY` `OTEL_EXPORTER_OTLP_METRIC_TLS_PRIVATE_KEY` |
```
SergeyKanzhelev, line 13: Found a usage of this variable, example here: https://github.com/open-telemetry/opentelemetry-python/blob/db74594d93b7e42d26ed2061b7672d1ad823f30c/exporter/opentelemetry-exporter-otlp/tests/test_otlp_metric_exporter.py#L74 Can you please make sure to check that this is fine to change

anuraaga: That would be renamed to `OTEL_EXPORTER_OTLP_METRIC_TLS_TRUSTED_CERTIFICATE` I think. Is that what you'd like to confirm?

jsuereth: Could we still allow the old Environment variable but mark it deprecated with a migration?

Given how CLOSE we are to 1.0 I think it'd be good if we pretend we have users we don't want to break, to get in the habit of how these changes will have to be made going forward.

anuraaga: If languages want to support multiple variables I think they could, even without it being mentioned in the spec right?

anuraaga:
Any more updates needed on this?

specification/protocol/exporter.md
SergeyKanzhelev: My comment was about checking with python maintainers whether it is already heavily in use. As @jsuereth pointed we need to be careful with renaming. If not much adoption so far, PR good with me.

anuraaga: Well, there aren't any open source usage

https://github.com/search?q=OTEL_EXPORTER_OTLP_METRIC_CERTIFICATE&type=code

And if we were done making changes, we'd already be 1.0 - this doc doesn't have any label on it, not even feature-freeze like some of the other ones so it seems a bit unfair to require too much process. The variable seems to be a bug really since the description doesn't describe it, these are credentials of a server, not a client.

Languages can use a deprecation cycle to change the variable if they wish, but I don't think that's something to have here, or if we did, it would be a much more general document detailing a mandate for and how to deprecate features.

SergeyKanzhelev: Not asking for much process. Maybe just a ack from @open-telemetry/python-approvers

anuraaga: Ok - @open-telemetry/python-approvers does this change look ok? Thanks!

anuraaga:
@codeboten @ocelotl @lzchen Could you take a look at this PR? Thanks!

ocelotl:
Sure, will do this tonight :+1:

codeboten (approved):
This looks fine to me, not sure if there is a heavy use of this variable just yet.

specification/protocol/exporter.md
```diff
@@ -10,7 +10,8 @@ The following configuration options MUST be available to configure the OTLP expo
...
 | Protocol             | The protocol used to transmit the data. One of `grpc`,`http/json`,`http/protobuf`. | `grpc`               | `OTEL_EXPORTER_OTLP_PROTOCOL` `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL` |
-| Certificate File     | Path to certificate file for TLS credentials of gRPC client. Should only be used for a secure connection. | n/a               | `OTEL_EXPORTER_OTLP_CERTIFICATE` `OTEL_EXPORTER_OTLP_TRACES_CERTIFICATE` `OTEL_EXPORTER_OTLP_METRICS_CERTIFICATE` |
+| Client certificate / key | The TLS certificate and private key to authenticate the client for mTLS. Should only be used for a secure connection. | n/a | `OTEL_EXPORTER_OTLP_TLS_CERTIFICATE` `OTEL_EXPORTER_OTLP_TRACES_TLS_CERTIFICATE` `OTEL_EXPORTER_OTLP_METRICS_TLS_CERTIFICATE`, `OTEL_EXPORTER_OTLP_TLS_PRIVATE_KEY` `OTEL_EXPORTER_OTLP_TRACES_TLS_PRIVATE_KEY` `OTEL_EXPORTER_OTLP_METRICS_TLS_PRIVATE_KEY` |
```
ocelotl, line 13: Just a question, mostly out of curiosity. I see a replacing of `Path to certificate file...` for `The TLS certificate and private key...`. The idea behind that is to use the contents of the certificate as the value of the environment variable?

ocelotl (approved):

specification/protocol/exporter.md
anuraaga: Thanks for calling this out. This was tough for me since this table has both configuration options and environment variables. I don't think the SDK has to expose a programmatic interface that only accepts a path, and often it'd probably not a good idea since it makes it difficult to provide credentials in non-file ways such as embedding into the binary or using a KMS client. The environment variable makes sense to point to a path though. So I made the description generic to not specify the type to apply to both.

anuraaga: Perhaps we should move the environment variables out of here to the sdk-environment-variables doc since they seem somewhat mysterious to be defined here and would give us the flexibility when describing. That'd be for another PR though I think.

lzchen (approved):

carlosalberto:
@jsuereth Any opinion on this one?

PR #1378, opened 2021-01-26
Add metric for currently-in-flight HTTP requests

evankanderson:
## Changes

Adds an `http.server.requests` metric to count in-flight HTTP requests on the server.

I was starting to implement some monitoring using the HTTP semantic conventions, and realized that it was also useful to have a count of the current in-flight HTTP requests (those which had started but not yet completed). This allows monitoring both the concurrency of the server and detecting unfortunate queueing or backlog events before the requests have completed (for example, if the request is blocked indefinitely).

(Will update CHANGELOG.md once I have a PR #)


specification/metrics/semantic_conventions/http-metrics.md
```diff
@@ -18,6 +18,7 @@ Below is a table of HTTP server metric instruments.
 | Name                   | Instrument    | Units        | Description |
 |------------------------|---------------|--------------|-------------|
 | `http.server.duration` | ValueRecorder | milliseconds | measures the duration of the inbound HTTP request |
+| `http.server.requests` | ValueObserver | requests     | measures the number of concurrent HTTP requests handled |
```
jkwatson, line 21: To me, this name evokes the total number of requests served, rather than concurrent requests. I recommend having a naming pattern that evokes the concurrent nature of what is being measured.

anuraaga: Suggestion is active_requests

evankanderson: `active_requests` would be fine with me, and answers #1370 as well.

evankanderson:
(Merged CHANGELOG.md -- let me know if I should squash)

evankanderson:
Is this a reasonable addition? Do I need to bring this up at a contributor meeting?

specification/metrics/semantic_conventions/http-metrics.md
```diff
@@ -48,6 +49,16 @@ and whether they should be on server, client, or both types of HTTP metric event
 | `net.host.name`    | `server`            | see [2] in [label alternatives](#label-alternatives) | See [general network connection attributes](../../trace/semantic_conventions/span-general.md#general-network-connection-attributes) |
 | `net.host.port`    | `server`            | see [2] in [label alternatives](#label-alternatives) | See [general network connection attributes](../../trace/semantic_conventions/span-general.md#general-network-connection-attributes) |
 
+The following labels SHOULD be included in the `http.server.requests` observation:
```
anuraaga, line 52:
```suggestion
The following labels SHOULD be included in the `http.server.active_requests` observation:
```

evankanderson: Thanks, fixed!

anuraaga (approved):
Thanks!

specification/metrics/semantic_conventions/http-metrics.md
```diff
@@ -17,9 +17,10 @@ type and units.
 
 Below is a table of HTTP server metric instruments.
 
-| Name                   | Instrument    | Units        | Description |
-|------------------------|---------------|--------------|-------------|
-| `http.server.duration` | ValueRecorder | milliseconds | measures the duration of the inbound HTTP request |
+| Name                          | Instrument    | Units        | Description |
+|-------------------------------|---------------|--------------|-------------|
+| `http.server.duration`        | ValueRecorder | milliseconds | measures the duration of the inbound HTTP request |
+| `http.server.active_requests` | ValueObserver | requests     | measures the number of concurrent HTTP requests handled |
```
reyang, line 23: Do we want "Active Requests", or "Requests that we've received" and "Requests that we've finished processing" (so we can calculate the currently-in-flight ones)?

reyang, line 23:
```suggestion
| `http.server.active_requests` | ValueObserver | requests     | measures the number of concurrent HTTP requests that are currently in-flight |
```

reyang (changes requested):
Please see my comment [here](https://github.com/open-telemetry/opentelemetry-specification/pull/1378/files#r578093122).

jmacd:
@reyang In other cases, we've discussed how it may be acceptable to specify semantic conventions without saying how they should be generated. For the case of "in flight HTTP requests" we could either materialize a dedicated metric or we could calculate the number in-flight by subtracting one metric from another. It's clearly much harder to do the second, so most users will prefer the first. But if you have a sophisticated metric system, it's redundant to record an additional metric and you shouldn't have to.

The example we discussed was metrics with `.utilization`, `.usage`, and `.limit` suffixes. With any two of these you can calculate the third, so which should we record? All three are useful to monitor, so it's hard to argue that in some cases you shouldn't just record all three.

specification/metrics/semantic_conventions/http-metrics.md
evankanderson: I think `http.server.duration` includes "requests we've finished processing".

It would be possible to calculate `http.server.started - count_of_distribution(http.server.duration)` to get the currently in-flight requests, but I think "started requests over all time" is probably the least useful of the three measures (I have no idea what I'd use it for), and you could derive it as `count_of_distribution(http.server.duration) + http.server.active_requests` if necessary.

In my dream world, I could actually have a histogram of "requests in flight by current in-flight time" here, but I don't have enough interest to make people suffer through temporary-histogram-buckets as yet another metric type for this resource.

evankanderson:
Sorry for the delay, the extra comments caught me in the middle of vacation/school break, and it took some time to dig out.

reyang (approved):
LGTM.

evankanderson:
I'm not authorized to merge this PR, but it looks like it does have the requisite reviews. What's the next step?

evankanderson:
@jmacd -- I think this is ready for merge?

specification/metrics/semantic_conventions/http-metrics.md
```diff
@@ -17,9 +17,10 @@ type and units.
 
 Below is a table of HTTP server metric instruments.
 
-| Name                   | Instrument    | Units        | Description |
-|------------------------|---------------|--------------|-------------|
-| `http.server.duration` | ValueRecorder | milliseconds | measures the duration of the inbound HTTP request |
+| Name                          | Instrument    | Units        | Description |
+|-------------------------------|---------------|--------------|-------------|
+| `http.server.duration`        | ValueRecorder | milliseconds | measures the duration of the inbound HTTP request |
+| `http.server.active_requests` | ValueObserver | requests     | measures the number of concurrent HTTP requests that are currently in-flight |
```
jmacd, line 23: I'd like this to be UpDownSumObserver, which is what you'd get from using an UpDownCounter instrument (delta temporality) with cumulative conversion.

evankanderson: Done. Not sure how I'd missed that.

jmacd (approved):

specification/metrics/semantic_conventions/http-metrics.md
jmacd: > requests in flight by current in-flight time

This is a GaugeHistogram :grinning: please add feedback on https://github.com/open-telemetry/opentelemetry-proto/issues/274

jmacd:
@evankanderson sorry for the long delay. Thank you for this contribution!

Issue #1373, opened 2021-01-25
Release v1.0.0


tedsuo:
This is a tracking issue for the v1.0 release of the specification. Please post any concerns or blockers here.

Remaining issues required for v1.0 can be found here: https://github.com/open-telemetry/opentelemetry-specification/issues?q=is%3Aopen+label%3Arelease%3Arequired-for-ga+-label%3Aspec%3Ametrics+label%3Apriority%3Ap1

v1.0 PR: https://github.com/open-telemetry/opentelemetry-specification/pull/1372

awssandra:
Hey tedsuo,

Where can we view the diff between 0.7.0 and the 1.0 spec?

Oberon00:
There are a bunch of issues in with the label `area:ga-tracking` https://github.com/open-telemetry/opentelemetry-specification/labels/area%3Aga-tracking and without priority. What do we do about these? Some of them seem to be for tracking the GA of non-spec repos, maybe these should be transferred to the community repo.

tedsuo:
@Oberon00 those are "non-spec" issues which are project wide. It's true, they could be in the community repo instead. Either way, we need to ensure progress is being made on them. (None are a blocker for releasing v1.0 of the spec). Let's review these in the maintainers meeting on Monday.

tedsuo:
@awssandra you can check out the changlelog to see what was added. Everything currently marked as "unreleased" will be added to the v1.0.0 release: https://github.com/open-telemetry/opentelemetry-specification/blob/main/CHANGELOG.md#unreleased

If you want to see the actual text changes, you can check out the `main` branch and then use `git diff v0.7.0` to see the changes directly. (Apologies if that's a pedantic answer!)

tedsuo:
I believe that all non-metrics issues which are required for GA are now resolved. At Monday's maintainer's meeting, I suggest we merge the PR (assuming the TC does not do so before then).

bogdandrutu:
Waiting for one more TC member to sign on that :)

PR #1372, opened 2021-01-23
Release v1.0.0

tedsuo:
This pull request versions the OpenTelemetry specification to v1.0.

If there are any open PRs or issues which you believe would block this merge, please list them here. 

Remaining issues marked as required for tracing GA can be found here: https://github.com/open-telemetry/opentelemetry-specification/issues?q=is%3Aopen+label%3Arelease%3Arequired-for-ga+-label%3Aspec%3Ametrics+label%3Apriority%3Ap1

PR #1385, opened 2021-01-27
Add lifecycle statuses to all documents

tedsuo:
## Changes

* Adds the current lifecycle stage to each document's status. 
* Clarifies that the feature-freeze status is a community process, not a definition of stability or feature-completeness.
* Sets the status for Tracing SDK, Resource SDK, Context, Propagation, Baggage API, Jaeger Exporter, and Zipkin Exporter to stable.

Related issues #

* https://github.com/open-telemetry/opentelemetry-specification/pull/1372
* https://github.com/open-telemetry/opentelemetry-specification/issues/1373


bogdandrutu:
Please rebase and checks will run

tedsuo:
@bogdandrutu resolved

iNikem (approved):

carlosalberto (approved):

specification/resource/semantic_conventions/README.md
```diff
@@ -1,5 +1,7 @@
 # Resource Semantic Conventions
 
+**Status**: [Experimental](../../document-status.md)
```
cijothomas, line 3: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#semantic-attributes-with-sdk-provided-default-value - This section of the semantic convention is reffered to by the main Resource spec (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#sdk-provided-resource-attributes). We need to get this portion to the main Resource spec, else it'll be difficult to justify the MUST requirement for "SDK-provided resource attributes", since it refers to an experimental portion.

lzchen:
I know that environment variables are deemed optional in the [spec matrix](https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md#environment-variables) but some languages have already implemented some of them. In this case, will the env var names be marked as "stable" [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md) as well or will there possible breaking changes in the future (such as the recent OTEL_EXPORTERS -> OTEL_METRICS_EXPORTER + OTEL_TRACE_EXPORTER)? We need to know whether or not we can release env var functionality as part of 1.0.0.

jsuereth (approved):

jsuereth:
> I know that environment variables are deemed optional in the [spec matrix](https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md#environment-variables) but some languages have already implemented some of them. In this case, will the env var names be marked as "stable" [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md) as well or will there possible breaking changes in the future (such as the recent OTEL_EXPORTERS -> OTEL_METRICS_EXPORTER + OTEL_TRACE_EXPORTER)? We need to know whether or not we can release env var functionality as part of 1.0.0.

I'd be nice if we could lock down the semantics folks were already relying on here without having to fully lock down everything about environment variables.  The OTEL_EXPORTERS change was one where we likely WILL be doing deprecation + migration in the future, and we could have done that here for existing SDKs rather than force migrate prior to 1.0.0.

It'd be nice if we agreed to "1.0" environment variables (even if they're optional, we still won't change semantics in breaking ways).  so +1 to answering this question.

specification/context/api-propagators.md
```diff
@@ -1,5 +1,7 @@
 # Propagators API
 
+**Status**: [Stable, Feature-Freeze](../document-status.md)
```
bogdandrutu, line 3: Since you removed `Feature-Freeze` from the status should we remove it from here as well?

tedsuo: Feature freeze is still a status, just separate from the lifecycle statuses. If we are once again open to reviewing new (backwards compatible) propagator features, then we can remove it. Otherwise, we can keep it.

specification/resource/semantic_conventions/README.md
tedsuo: That is a good point. But I don't think this link is wrong – we have not agreed to make the conventions stable yet, but we have agreed to use them. So from a spec perspective, the reference from a stable SDK to an experimental section is correct, in that it reflects the current state of affairs that we've agreed to as a group. The MUST in that section of the SDK means that the current set of default resources must be provided, not that the default resources themselves cannot change.

However, I don't really like this state of affairs – it would be better to make the conventions stable. You are correctly pointing out that it is problematic to leave them experimental. There's no way for the SDK to avoid referencing an experimental package until we make them stable. 

Changing any semantic convention means that bumping to the latest version of the SDK could cause a transitive compilation error, if the SDK forces an update to a newer version of the semantic conventions. We can mitigate that issue by having the SDK maintain its own internal copy of the conventions that it uses. That isn't pretty.

But, again, this situation is what we have explicitly decided to live with, by leaving the semantic conventions experimental. 

Since all of our instrumentation relies on these conventions, the user is in for a bumpy ride if we break any convention – especially if we fail to roll those changes out quickly. We better respect that fact when we touch the conventions – experimental doesn't mean we can be thoughtless. We can't dribble out breaking changes to the conventions.

tedsuo:
@lzchen @jsuereth I think there are two questions about env vars.

* Have we marked them as stable yet? The answer to that is no.
* Should we mark them as stable? We better do that soon, IMHO.

Since they are currently experimental, I would like to mark them here as such, and then a separate issue can be opened about making them stable. Does that sound good to you two?

lzchen:
@tedsuo 
Discussed during 02/01/2021 maintainer's meeting.
Environment variables that are related to tracing will be marked as stable by the 1.0.0 release.

bogdandrutu (approved):

specification/protocol/otlp.md
```diff
@@ -1,5 +1,7 @@
 # OpenTelemetry Protocol Specification
 
+**Status**: [Stable](../document-status.md)
```
bogdandrutu, line 3: There is are discussion in this document about JSON being supported for OTLP/HTTP which I think we are not yet ready to make it stable. How can we solve that?

bogdandrutu (changes requested):

cijothomas (approved):

specification/protocol/otlp.md
tedsuo: Ok, I marked OTLP/HTTP as experimental. The JSON and binary formats for HTTP were intermixed; if we want proto over OTLP/HTTP to be stable this section should be re-written.

tigrannajaryan: OTLP/HTTP-Protobufs are stable. Only OTLP/HTTP-JSON is unstable. Please file a TODO to fix this.

We should probably also note that only Traces are currently stable. Looks like we are going to end up with a lot of clarifications for our stability declarations in this file to be true, but that's the reality.

carlosalberto: @bogdandrutu Please review. I think that we are ready to merge this and tune the OTLP/JSON part in the related follow up.

reyang (approved):
:shipit:

bogdandrutu (approved):

anuraaga (approved):

Sachin1O1:
I guess, we need to update the status of the doc again or is it still in experimental?

PR #1386, opened 2021-01-28
Small clarification for default's service.name for exporters.

### LINKED ISSUE #1237: Jaeger exporter service.name clarification needed

### LINKED ISSUE #1380: Zipkin spec should suggest to use Resource.GetDefault
These 2 PRs modified default resource requirements. 
https://github.com/open-telemetry/opentelemetry-specification/pull/1269
https://github.com/open-telemetry/opentelemetry-specification/pull/1294

Zipkin exporter spec must be updated to reflect this change.
https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk_exporters/zipkin.md#service-name

carlosalberto:
Fixes #1237 
Fixes #1380 

Small clarification on using the default `Resource` is no `service.name` was used.

PS - Don't think we need a CHANGELOG entry for this, but happy to add one if needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,7 +52,7 @@ Updates:
   ([#1339](https://github.com/open-telemetry/opentelemetry-specification/pull/1339))
 - Trace Exporters: Fix TODOs in Jaeger exporter spec
   ([#1374](https://github.com/open-telemetry/opentelemetry-specification/pull/1374))
-- Clarify that Jaeger/Zipkin exporters must to rely on the default Resource to
+- Clarify that Jaeger/Zipkin exporters must rely on the default Resource to
   get service.name if none was specified.
   ([#1386](https://github.com/open-telemetry/opentelemetry-specification/pull/1386))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,9 @@ Updates:
   ([#1339](https://github.com/open-telemetry/opentelemetry-specification/pull/1339))
 - Trace Exporters: Fix TODOs in Jaeger exporter spec
   ([#1374](https://github.com/open-telemetry/opentelemetry-specification/pull/1374))
+- Clarify that Jaeger/Zipkin exporters must to rely on the default Resource to
+  get service.name if none was specified.
+  ([#1386](https://github.com/open-telemetry/opentelemetry-specification/pull/1386))
 
 ## v0.7.0 (11-18-2020)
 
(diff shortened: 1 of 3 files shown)


jkwatson (approved):

specification/trace/sdk_exporters/jaeger.md
```diff
@@ -43,6 +43,8 @@ single process and exporters need to handle this case accordingly.
 Critically, Jaeger backend depends on `Span.Process.ServiceName` to identify the service
 that produced the spans. That field MUST be populated from the `service.name` attribute
 of the [`service` resource](../../resource/semantic_conventions/README.md#service).
+If no `service.name` was specified, that field MUST be populated from the
```
Oberon00, line 46:
```suggestion
If no `service.name` is contained in a Span's Resource, that field MUST be populated from the
```

Oberon00 (commented):
Nit: "was specified" sounds a bit like there may be another way beyond resources to specify a service name.

Oberon00 (approved):

arminru (approved):

yurishkuro (approved):

codeboten (approved):

specification/trace/sdk_exporters/zipkin.md
```diff
@@ -53,7 +53,9 @@ and Zipkin.
 
 Zipkin service name MUST be set to the value of the
 [resource attribute](../../resource/semantic_conventions/README.md):
-`service.name`. In Zipkin it is important that the service name is consistent
+`service.name`. If no `service.name` is contained in a Span's Resource, it MUST be populated from the
+[default](../../resource/sdk.md#sdk-provided-resource-attributes) `Resource`.
+In Zipkin it is important that the service name is consistent
```
cijothomas, line 58: few lines down, it says "Zipkin MUST allow to configure the default "fall back" name". We need to remove that as well, as the only source from which service name should be obtained is the Resource.

carlosalberto: Ooh, good catch! Will adjust that part, thanks! Updated.

cijothomas (approved):

andrewhsu (approved):
LGTM

cijothomas:
> PS - Don't think we need a CHANGELOG entry for this, but happy to add one if needed

I'd add it, as this is easy to overlook by language SIGs. A changelog would make it more easy.

CHANGELOG.md
```diff
@@ -52,6 +52,9 @@ Updates:
   ([#1339](https://github.com/open-telemetry/opentelemetry-specification/pull/1339))
 - Trace Exporters: Fix TODOs in Jaeger exporter spec
   ([#1374](https://github.com/open-telemetry/opentelemetry-specification/pull/1374))
+- Clarify that Jaeger/Zipkin exporters must to rely on the default Resource to
```
iNikem, line 55:
```suggestion
- Clarify that Jaeger/Zipkin exporters must rely on the default Resource to
```

PR #1388, opened 2021-01-29
Update api.md

aman-raza:
Fixes #

## Changes

Removed the optional tag from the "version" param and the "required" tag from name param.

### DIFF
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -101,7 +101,7 @@ The `TracerProvider` MUST provide the following functions:
 
 This API MUST accept the following parameters:
 
-- `name` (required): This name must identify the [instrumentation library](../overview.md#instrumentation-libraries)
+- `name` : This name must identify the [instrumentation library](../overview.md#instrumentation-libraries)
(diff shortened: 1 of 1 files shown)


yurishkuro:
what is the motivation for this change?

aman-raza:
> what is the motivation for this change?

Both the name and version is required by the traceprovider
currently it is optional

Oberon00:
> Both the name and version is required by the traceprovider

Which language are you talking about?

aman-raza:
> > Both the name and version is required by the traceprovider
> 
> Which language are you talking about?

php

Oberon00:
I think this is an issue with PHP then. This repository is a specification not a documentation. That is, it states how the implementations *should* be. It is not (primarily) a user manual.

aman-raza:
> I think this is an issue with PHP then. This repository is a specification not a documentation. That is, it states how the implementations _should_ be. It is not (primarily) a user manual.

ohh so I have to look for the issue in .php then, got it

PR #1393, opened 2021-01-30
Remove contradictory and confusing statement from Resource Attributes

cijothomas:
Fixes some portion of #1184 

## Changes
Spec says Resource attribute can be any type (string, int, bool etc.) allowed in common attribute (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attributes). Yet the Retrieve Attribute section hints that name/value should be strings. Its best to remove this sentence to avoid any confusion. 

This is just editorial change, not changing actual spec.


reyang (approved):
LGTM.

iNikem (approved):

yurishkuro (approved):

arminru (approved):

Oberon00 (approved):

PR #1400, opened 2021-02-04
OTLP stability clarifications

### LINKED ISSUE #1397: Clarify OTLP stability
Currently the OTLP specification is a mix of stable and experimental sections. Some of these sections are mixed together to the point that it is hard to clarify what is stable vs what is experimental. For example: 
* Proto over HTTP is stable, while JSON over HTTP is experimental. 
* The tracing protocol is stable, while the metrics protocol is experimental.

This document should be reformatted to more clearly separate the stable and experimental sections.

carlosalberto:
Fixes #1397 

Adds minimal clarifications.

Specifically, for OTLP/HTTP it simply mentions that we are stable for the Binary part, while being experimental for the JSON one. I'm also open to rewrite these sections to have them clearly separated, although that would probably be an overkill, given that they both share **a lot**.

### DIFF
--- a/specification/protocol/otlp.md
+++ b/specification/protocol/otlp.md
@@ -301,7 +301,8 @@ The default network port for OTLP/gRPC is 4317.
 
 ### OTLP/HTTP
 
-**Status**: [Experimental](../document-status.md)
+**Binary Format Status**: [Stable](../document-status.md)
+**JSON Format Status**: [Experimental]((../document-status.md))
 
 OTLP/HTTP uses Protobuf payloads encoded either in binary format or in JSON
 format. The Protobuf schema of the messages is the same for OTLP/HTTP and
--- a/specification/protocol/otlp.md
+++ b/specification/protocol/otlp.md
@@ -48,8 +48,7 @@ of OpenTelemetry project.
 ## Signals Maturity Level
 
 Each signal has different support and stability in OTLP, described through its
-own maturity level, which in turn applies to **all** the OTLP Transports listed below,
-e.g. OTLP/gRPC.
+own maturity level, which in turn applies to **all** the OTLP Transports listed below.
 
 * Tracing: **Stable**
 * Metrics: **Beta**
--- a/specification/protocol/otlp.md
+++ b/specification/protocol/otlp.md
@@ -11,6 +11,7 @@ nodes such as collectors and telemetry backends.
 Table of Contents
 </summary>
 
+- [Signal Status](#signal-status)
 - [Protocol Details](#protocol-details)
   * [OTLP/gRPC](#otlpgrpc)
     + [OTLP/gRPC Concurrent Requests](#otlpgrpc-concurrent-requests)
@@ -43,6 +44,16 @@ Table of Contents
 
 OTLP is a general-purpose telemetry data delivery protocol designed in the scope
 of OpenTelemetry project.
+
+## Signal Status
+
+Each signal has different support and stability in OTLP, described through its
+own status, which in turn applies to **all** the OTLP Transports listed below,
+e.g. OTLP/gRPC.
+
+* Tracing: [Stable](../document-status.md)
+* Metrics: [Experimental](../document-status.md)
+* Logs: [Experimental](../document-status.md)
 
 ## Protocol Details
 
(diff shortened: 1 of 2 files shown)


carlosalberto:
cc @tigrannajaryan

specification/protocol/otlp.md
```diff
@@ -44,6 +45,16 @@ Table of Contents
 OTLP is a general-purpose telemetry data delivery protocol designed in the scope
 of OpenTelemetry project.
 
+## Signal Status
```
Oberon00, line 48: Should we defer to the table at https://github.com/open-telemetry/opentelemetry-proto#maturity-level?

carlosalberto: Not entirely opposed, although I'm wondering about a (temporary) state where Metrics is Stable but its Collector counterpart is not. Happy to defer it that table however.

specification/protocol/otlp.md
```diff
@@ -44,6 +45,16 @@ Table of Contents
...
+
+Each signal has different support and stability in OTLP, described through its
+own status, which in turn applies to **all** the OTLP Transports listed below,
+e.g. OTLP/gRPC.
+
+* Tracing: [Stable](../document-status.md)
+* Metrics: [Experimental](../document-status.md)
+* Logs: [Experimental](../document-status.md)
```
tigrannajaryan, line 56: Metrics are marked "Beta", Logs are marked "Alpha" here https://github.com/open-telemetry/opentelemetry-proto#maturity-level
Should we use the same wording here?

specification/protocol/otlp.md
```diff
@@ -301,7 +312,8 @@ The default network port for OTLP/gRPC is 4317.
 
 ### OTLP/HTTP
 
-**Status**: [Experimental](../document-status.md)
+**Binary Format Status**: [Stable](../document-status.md)
+**JSON Format Status**: [Experimental](../document-status.md)
```
tigrannajaryan, line 316: Same for JSON. Should we say "Alpha"?

specification/protocol/otlp.md
carlosalberto: Updated, thanks for pointing out this out ;) Included a link to that table, just in case.

specification/protocol/otlp.md
carlosalberto: As the sections in this document use the values in document-status.md, I think it's worth keeping it as "Experimental".

specification/protocol/otlp.md
carlosalberto: Included a "See..." note ;)

specification/protocol/otlp.md
tigrannajaryan: Works either way for me. No strong opinion.
It may make sense to align document-status.md with our "Maturity levels" used elsewhere. Doesn't have to be in this PR, can be done later.

specification/protocol/otlp.md
```diff
@@ -44,6 +45,18 @@ Table of Contents
 OTLP is a general-purpose telemetry data delivery protocol designed in the scope
 of OpenTelemetry project.
 
+## Signals Maturity Level
+
+Each signal has different support and stability in OTLP, described through its
+own maturity level, which in turn applies to **all** the OTLP Transports listed below,
+e.g. OTLP/gRPC.
```
iNikem, line 52: If it applies to all described below, isn't it strange to make an individual example? Why this specific transport is more important that others?

iNikem (approved):

arminru (approved):

specification/protocol/otlp.md
carlosalberto: Removed it ;)

specification/protocol/otlp.md
carlosalberto: @tigrannajaryan Thanks. Let's do a full level review as a follow up and go ahead with this PR in order to get things settled for now ;)

tigrannajaryan (approved):

carlosalberto:
Merging to include this as part of our release. Also, this should not be polemical as it's merely a clarification.

bogdandrutu:
Thanks @carlosalberto

PR #1404, opened 2021-02-05
Fix capitalization of MUST NOT/SHOULD NOT

yurishkuro:
Per https://tools.ietf.org/html/rfc2119, the work "NOT" must also be capitalized.

Skipping changelog since this is a stylistic change.

### DIFF
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -211,7 +211,7 @@ The API MUST allow retrieving the `TraceId` and `SpanId` in the following forms:
 * Binary - returns the binary representation of the `TraceId` (result MUST be a
 16-byte array) or `SpanId` (result MUST be an 8-byte array).
 
-The API should not expose details about how they are internally stored.
+The API SHOULD NOT expose details about how they are internally stored.
 
 ### IsValid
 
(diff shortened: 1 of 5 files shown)


specification/trace/api.md
```diff
@@ -241,7 +241,7 @@ All mutating operations MUST return a new `TraceState` with the modifications ap
 `TraceState` MUST at all times be valid according to rules specified in [W3C Trace Context specification](https://www.w3.org/TR/trace-context/#tracestate-header-field-values).
 Every mutating operations MUST validate input parameters.
 If invalid value is passed the operation MUST NOT return `TraceState` containing invalid data
-and MUST follow the [general error handling guidelines](../error-handling.md) (e.g. it usually must not return null or throw an exception).
+and MUST follow the [general error handling guidelines](../error-handling.md) (e.g. it usually should not return null or throw an exception).
```
yurishkuro, line 244: "usually must" does not make sense

reyang: I think "and MUST follow the general error handling guidelines" is good enough - perhaps we just remove "(e.g. it usually should not return null or throw an exception)" at all?

reyang (approved):
LGTM with one non-blocking suggestion.

specification/trace/api.md
yurishkuro: +1

bogdandrutu (approved):

specification/trace/api.md
```diff
@@ -241,7 +241,7 @@ All mutating operations MUST return a new `TraceState` with the modifications ap
 `TraceState` MUST at all times be valid according to rules specified in [W3C Trace Context specification](https://www.w3.org/TR/trace-context/#tracestate-header-field-values).
 Every mutating operations MUST validate input parameters.
 If invalid value is passed the operation MUST NOT return `TraceState` containing invalid data
-and MUST follow the [general error handling guidelines](../error-handling.md) (e.g. it usually must not return null or throw an exception).
```
Oberon00, line 244: Why was this removed? The point about `null` is IMHO important. Maybe it should be refactored to not be inside the `e.g.`

PR #1402, opened 2021-02-05
Add OT Trace Propagator.

carlosalberto:
Fixes #1391 

Adds  `OT Trace` propagator as a (very) optional extension component.

cc @yurishkuro

PR #1406, opened 2021-02-08
Update ottrace environment variable to match spec.

anuraaga:
Presuming this was just missing from #1402. Hope it can be added without much ceremony, otherwise recommend renaming OT Trace to OT Tracer for the spec entry (matches the header name)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ New:
 
 Updates:
 
+- Update OT Trace propagator environment variable to match latest name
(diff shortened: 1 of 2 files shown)


carlosalberto:
Had been thinking about it myself when doing this change, but was afraid to change too many things. Then again, now is the last time for doing it, so let's go with it.

carlosalberto (approved):

Oberon00 (approved):

codeboten (approved):

cijothomas (approved):

PR #1407, opened 2021-02-08
[Blocking 1.0] Add status to environment variables document

bogdandrutu:

### DIFF
--- a/specification/sdk-environment-variables.md
+++ b/specification/sdk-environment-variables.md
@@ -73,7 +73,7 @@ Depending on the value of `OTEL_TRACES_SAMPLER`, `OTEL_TRACES_SAMPLER_ARG` may b
 
 ## Span Collection Limits
 
-**Status**: [Stable](document-status.md)
+**Status**: [Experimental](document-status.md)
 
 | Name                            | Description                          | Default | Notes |
 | ------------------------------- | ------------------------------------ | ------- | ----- |
--- a/specification/sdk-environment-variables.md
+++ b/specification/sdk-environment-variables.md
@@ -1,8 +1,12 @@
 # OpenTelemetry Environment Variable Specification
 
+**Status**: [Mixed](document-status.md)
+
 The goal of this specification is to unify the environment variable names between different OpenTelemetry SDK implementations. SDKs MAY choose to allow configuration via the environment variables in this specification, but are not required to. If they do, they SHOULD use the names listed in this document.
 
 ## Special configuration types
+
+**Status**: [Stable](document-status.md)
 
 ### Numeric value
 
@@ -22,6 +26,8 @@ gracefully ignore the setting and use the default value if it is defined.
 For example, the value `12000` indicates 12000 milliseconds, i.e., 12 seconds.
 
 ## General SDK Configuration
+
+**Status**: [Stable](document-status.md)
 
 | Name                     | Description                                       | Default                           | Notes                               |
 | ------------------------ | ------------------------------------------------- | --------------------------------- | ----------------------------------- |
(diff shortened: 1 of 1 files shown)


carlosalberto (approved):

specification/sdk-environment-variables.md
```diff
@@ -1,5 +1,9 @@
 # OpenTelemetry Environment Variable Specification
 
+Tracing: [Stable](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/document-status.md)
```
Oberon00, line 3: What category are variables like:
* OTEL_LOG_LEVEL
* OTEL_PROPAGATORS
* OTEL_RESOURCE_ATTRIBUTES

This categorization is to coarse or incomplete.

Oberon00: Also there is general wording like how a duration should look. Is this also stable?

carlosalberto: There's also https://github.com/open-telemetry/opentelemetry-specification/pull/1409 btw - and in general we are in a similar boat to Proto, as overall we are stable, with a few parts (Metrics/Logging) not being as stable.

Hence I propose we mark this *entire* document as stable (as #1409 does), mentioning that the metrics/logs sections are unstable though.

tedsuo: * If a document is mixed, please mark it as **Status**: [Mixed](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/document-status.md) at the top, and then mark each section as `Stable` or `Experimental`
* Please use a relative link, not an absolute link.

carlosalberto: After reviewing this very document, I see the **only** covered environment variable is `OTLP_METRICS_EXPORTER`, which IMHO shouldn't change _at all_. 

For logging, we can go without it **for now**, as there are no log variables at this moment here (and `OTEL_LOG_LEVEL` is not quite the same, hah).

So let's go stable IMHO.

bogdandrutu: `OTEL_EXPORTER_PROMETHEUS*` as well, it is not stable correct?

carlosalberto: Oh well. I'd say they _should_ be stable as it's only two basic options, but yes, I guess those ones are unstable :(

bogdandrutu: Hope I covered all the comments. I think we agreed on the duration to be stable and we can extend in the future.

tedsuo (approved):
LGTM. @Oberon00 did we miss anything?

iNikem (approved):

SergeyKanzhelev (approved):

specification/sdk-environment-variables.md
```diff
@@ -65,6 +73,8 @@ Depending on the value of `OTEL_TRACES_SAMPLER`, `OTEL_TRACES_SAMPLER_ARG` may b
 
 ## Span Collection Limits
 
+**Status**: [Stable](document-status.md)
```
Oberon00, line 76: This should not really be stable IMHO. We don't even have a spec at all for span limits, apart from these environment variables.

jsuereth: Do you expect what's specified to change?  Totally agree that spin limits should be spec'd out better, but do we think that would break the current definition of these?

IIUC Stability = Adding is ok, removing is not.

jsuereth (approved):

specification/sdk-environment-variables.md
bogdandrutu: I marked this as experimental for the moment. Will followup.

PR #1414, opened 2021-02-09
[For 1.0] Add WARNING for TraceIdRatioBased algorithm stability.

Oberon00:
Required for releasing 1.0 (#1373).

Alternative to PR #1412.

----

See https://github.com/open-telemetry/opentelemetry-specification/pull/1372#issuecomment-775934345

TraceIdRatioBased is only vaguely specified and has a TODO in it. We should call attention to the resulting unstable-ness of the algorithm.


specification/trace/sdk.md
```diff
@@ -136,20 +136,27 @@ the `ParentBased` sampler specified below.
...
-traces that any `TraceIdRatioBased` sampler with a lower sampling rate would
-sample. This is important when a backend system may want to run with a higher
-sampling rate than the frontend system, this way all frontend traces will
-still be sampled and extra traces will be sampled on the backend only.
+  traces that any `TraceIdRatioBased` sampler with a lower sampling rate would
+  sample. This is important when a backend system may want to run with a higher
+  sampling rate than the frontend system, this way all frontend traces will
+  still be sampled and extra traces will be sampled on the backend only.
+* **WARNING:** Since the exact algorithm is not specified yet (see TODO above),
+  there will probably be be breaking changes to it in any language SDK once it is.
```
arminru, line 154:
```suggestion
  there will probably be breaking changes to it in any language SDK once it is specified.
```

arminru (approved):
Thank you!

specification/trace/sdk.md
jkwatson, line 154: This language makes it sound like we will be required to bump to a 2.0.0 version when this is specified. Is that the intent?

Oberon00: The intent was exactly to avoid that. Maybe there is a better word than "breaking" here? Or "if you rely on the algorithm, your code will break"?

bogdandrutu (commented):
I am confused about the description of the PR, ParentBased seems simple and stabel

specification/trace/sdk.md
Oberon00: 5399f14

punya (approved):
This is a pragmatic choice that avoids churning APIs/SDKs just before the release, with a fairly limited downside risk for future releases.

jsuereth (approved):

carlosalberto (approved):

specification/trace/sdk.md
```diff
@@ -136,20 +136,28 @@ the `ParentBased` sampler specified below.
...
-still be sampled and extra traces will be sampled on the backend only.
+  traces that any `TraceIdRatioBased` sampler with a lower sampling rate would
+  sample. This is important when a backend system may want to run with a higher
+  sampling rate than the frontend system, this way all frontend traces will
+  still be sampled and extra traces will be sampled on the backend only.
+* **WARNING:** Since the exact algorithm is not specified yet (see TODO above),
+  there will probably be changes to it in any language SDK once it is, which
+  would break code that relies on the algorithm results.
+  Only the configuration and creation APIs can be considered stable.
+  It is recommended to use this sampler algorithm only for root spans
```
yurishkuro, line 157: I think what this is saying is that with the current state of the spec the sampler MUST be used as a delegate of the `ParentBased` sampler, not as a top-level sampler.

Oberon00: It's hard to place requirements on users 😃

yurishkuro: True, but we can place reqs on the documentation of SDK classes, to say that in the current state there is no expected interop behavior across languages for the sampling logic.

cijothomas (approved):

bogdandrutu (approved):

bogdandrutu:
@yurishkuro are you ok to move with this option then?

yurishkuro (approved):

yurishkuro:
sgtm

PR #1416, opened 2021-02-09
Finalize SpanLimits definition, add option to configure it

bogdandrutu:
Followup from https://github.com/open-telemetry/opentelemetry-specification/pull/1407 when we were asked to exclude the environment variables for the span limits.

**Update:**
Discussed with the @open-telemetry/technical-committee if we want to have this or not, and decision was to include this PR in the version 1.0 that will be released today:
* This PR does not add a new functionality, the limits were defined before and defined only as configurable via ENV variables. Having the limits was and will remain optional. 
* This PR indeed does make required to add the code configuration as well and standardize it, for languages that implement this feature, but all languages which support the ENV variables have a way to configure this via code anyway, so this PR tries only to standardize on that mechanism and will be a small rename (if any).

### DIFF
--- a/spec-compliance-matrix.md
+++ b/spec-compliance-matrix.md
@@ -70,6 +70,7 @@ status of the feature is not known.
 | ShouldSample gets full parent Context                                                            |          |    | +    | +  | +      | +    | +      |     |      | +   | -    | +     |
 | [New Span ID created also for non-recording Spans](specification/trace/sdk.md#sdk-span-creation) |          |    |      |    | +      | +    |        |     |      |     | -    | +     |
 | [IdGenerators](specification/trace/sdk.md#id-generators) ]                                       |          |    |      |    |        |      |        |     |      |     |      |       |
+| [SpanLimits](specification/trace/sdk.md#span-limits) ]                                           |          |    |      |    |        |      |        |     |      |     |      |       |
 
 ## Baggage
 
--- a/spec-compliance-matrix.md
+++ b/spec-compliance-matrix.md
@@ -70,7 +70,7 @@ status of the feature is not known.
 | ShouldSample gets full parent Context                                                            |          |    | +    | +  | +      | +    | +      |     |      | +   | -    | +     |
 | [New Span ID created also for non-recording Spans](specification/trace/sdk.md#sdk-span-creation) |          |    |      |    | +      | +    |        |     |      |     | -    | +     |
 | [IdGenerators](specification/trace/sdk.md#id-generators) ]                                       |          |    |      |    |        |      |        |     |      |     |      |       |
-| [SpanLimits](specification/trace/sdk.md#span-limits) ]                                           |          |    |      |    |        |      |        |     |      |     |      |       |
+| [SpanLimits](specification/trace/sdk.md#span-limits) ]                                           | X        |    |      |    |        |      |        |     |      |     |      |       |
 
 ## Baggage
 
--- a/specification/trace/sdk.md
+++ b/specification/trace/sdk.md
@@ -303,9 +303,11 @@ To protect against such errors, SDK Spans MAY discard attributes, links, and
 events that would increase the number of elements of each collection beyond
 the configured limit.
 
-It the SDK implements the limits above it MUST provide a way to change these limits, via a configuration to the
-TracerProvider, by allowing users to configure individual limits like in the Java
-example bellow. The name of the configuration class MAY be `SpanLimits`, and the
+It the SDK implements the limits above it MUST provide a way to change these
+limits, via a configuration to the TracerProvider, by allowing users to
+configure individual limits like in the Java example bellow.
+
+The name of the configuration class MAY be `SpanLimits`, and the
 name of the properties SHOULD be `AttributeCountLimit`, `EventCountLimit` and
 `LinkCountLimit`. Implementations MAY provide additional configuration such as
 `AttributePerEventCountLimit` and `AttributePerLinkCountLimit`.
@@ -324,13 +326,13 @@ public final class SpanLimits {
 
 **Configurable parameters:**
 
-* `AttributeCountLimit` - Maximum allowed span attribute count;
-* `EventCountLimit` - Maximum allowed span event count;
-* `LinkCountLimit` - Maximum allowed span link count;
-* `AttributePerEventCountLimit` - Maximum allowed attribute per span event count;
-* `AttributePerLinkCountLimit` - Maximum allowed attribute per span link count;
+* `AttributeCountLimit` (Default=1000) - Maximum allowed span attribute count;
+* `EventCountLimit` (Default=1000) - Maximum allowed span event count;
+* `LinkCountLimit` (Default=1000) - Maximum allowed span link count;
+* `AttributePerEventCountLimit` (Default=128) - Maximum allowed attribute per span event count;
+* `AttributePerLinkCountLimit` (Default=128) - Maximum allowed attribute per span link count;
 
-If there is a configurable provided, the SDK SHOULD honor the environment variables
+If there is a configuration provided, the SDK SHOULD honor the environment variables
 specified in [SDK environment variables](../sdk-environment-variables.md#span-collection-limits).
 
 There SHOULD be a log emitted to indicate to the user that an attribute, event,
--- a/specification/trace/sdk.md
+++ b/specification/trace/sdk.md
@@ -307,10 +307,10 @@ It the SDK implements the limits above it MUST provide a way to change these
 limits, via a configuration to the TracerProvider, by allowing users to
 configure individual limits like in the Java example bellow.
 
-The name of the configuration class MAY be `SpanLimits`, and the
-name of the properties SHOULD be `AttributeCountLimit`, `EventCountLimit` and
-`LinkCountLimit`. Implementations MAY provide additional configuration such as
-`AttributePerEventCountLimit` and `AttributePerLinkCountLimit`.
+The name of the configuration options SHOULD be `AttributeCountLimit`,
+`EventCountLimit` and `LinkCountLimit`. The options MAY be bundled in a class,
+which then SHOULD be called `SpanLimits`. Implementations MAY provide additional
+configuration such as `AttributePerEventCountLimit` and `AttributePerLinkCountLimit`.
 
 ```java
 public final class SpanLimits {
@@ -331,9 +331,6 @@ public final class SpanLimits {
 * `LinkCountLimit` (Default=1000) - Maximum allowed span link count;
 * `AttributePerEventCountLimit` (Default=128) - Maximum allowed attribute per span event count;
 * `AttributePerLinkCountLimit` (Default=128) - Maximum allowed attribute per span link count;
-
-If there is a configuration provided, the SDK SHOULD honor the environment variables
-specified in [SDK environment variables](../sdk-environment-variables.md#span-collection-limits).
 
 There SHOULD be a log emitted to indicate to the user that an attribute, event,
 or link was discarded due to such a limit. To prevent excessive logging, the log
--- a/specification/trace/sdk.md
+++ b/specification/trace/sdk.md
@@ -303,12 +303,12 @@ To protect against such errors, SDK Spans MAY discard attributes, links, and
 events that would increase the number of elements of each collection beyond
 the configured limit.
 
-The SDKs MUST provide a way to change these limits, via a configuration to the
-TracerProvider, by allowing users to configure individual limits like the java
-example bellow. The name of the cofiguration class MAY be `SpanLimits`, and the
-name of the properties SHOULD be `AttributeCountLimit`, `EventCountLimit`,
-`LinkCountLimit`. Implementation MAY provide additional configuration such as
-`AttributePerEventCountLimit`, `AttributePerLinkCountLimit`.
+It the SDK implements the limits above it MUST provide a way to change these limits, via a configuration to the
+TracerProvider, by allowing users to configure individual limits like in the Java
+example bellow. The name of the configuration class MAY be `SpanLimits`, and the
+name of the properties SHOULD be `AttributeCountLimit`, `EventCountLimit` and
+`LinkCountLimit`. Implementations MAY provide additional configuration such as
+`AttributePerEventCountLimit` and `AttributePerLinkCountLimit`.
 
 ```java
 public final class SpanLimits {
--- a/specification/trace/sdk.md
+++ b/specification/trace/sdk.md
@@ -9,7 +9,7 @@
 * [Tracer Provider](#tracer-provider)
 * [Additional Span Interfaces](#additional-span-interfaces)
 * [Sampling](#sampling)
-* [Limits on Span Collections](#limits-on-span-collections)
+* [Span Limits](#span-limits)
 * [Id Generator](#id-generators)
 * [Span Processor](#span-processor)
 * [Span Exporter](#span-exporter)
@@ -26,9 +26,10 @@ supplied to the `TracerProvider` must be used to create an
 [`InstrumentationLibrary`][otep-83] instance which is stored on the created
 `Tracer`.
 
-Configuration (i.e., [Span processors](#span-processor), [IdGenerator](#id-generators),
-and [`Sampler`](#sampling)) MUST be managed solely by the `TracerProvider` and it
-MUST provide some way to configure them, at least when creating or initializing it.
+Configuration (i.e., [SpanProcessors](#span-processor), [IdGenerator](#id-generators),
+[SpanLimits](#span-limits) and [`Sampler`](#sampling)) MUST be managed solely by
+the `TracerProvider` and it MUST provide some way to configure all of them that
+are implemented in the SDK, at least when creating or initializing it.
 
 The TracerProvider MAY provide methods to update the configuration. If
 configuration is updated (e.g., adding a `SpanProcessor`),
@@ -292,7 +293,7 @@ Optional parameters:
 |present|false|true|`localParentSampled()`|
 |present|false|false|`localParentNotSampled()`|
 
-## Limits on Span Collections
+## Span Limits
 
 Erroneous code can add unintended attributes, events, and links to a span. If
 these collections are unbounded, they can quickly exhaust available memory,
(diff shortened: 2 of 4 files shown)


specification/trace/sdk.md
```diff
@@ -284,18 +285,43 @@ Thus, the SDK specification defines sets of possible requirements for
...
 To protect against such errors, SDK Spans MAY discard attributes, links, and
 events that would increase the number of elements of each collection beyond
-the recommended limit of 1000 elements. SDKs MAY provide a way to change this limit.
+the recommended limit of 1000 elements.
 
-If there is a configurable limit, the SDK SHOULD honor the environment variables
-specified in [SDK environment variables](../sdk-environment-variables.md#span-collection-limits).
+The SDKs MAY provide a way to change these limits, via a configuration to the
+TracerProvider, by allowing users to configure individual limits like the java
+example bellow (the name of the cofiguration class MAY be `SpanLimits`, and the
```
tigrannajaryan, line 300:
```suggestion
example bellow. The name of the cofiguration class MAY be `SpanLimits`, and the
```

specification/trace/sdk.md
```diff
@@ -284,18 +285,43 @@ Thus, the SDK specification defines sets of possible requirements for
...
+name of the properties SHOULD be `AttributeCountLimit`, `EventCountLimit`,
+`LinkCountLimit` to be consistent with the environment variables; implementation
```
tigrannajaryan, line 302:
```suggestion
`LinkCountLimit` to be consistent with the environment variables. Implementation
```

specification/trace/sdk.md
```diff
@@ -284,18 +285,43 @@ Thus, the SDK specification defines sets of possible requirements for
...
+MAY provide additional configuration such as `AttributePerEventCountLimit`,
+`AttributePerLinkCountLimit`):
```
tigrannajaryan, line 304:
```suggestion
`AttributePerLinkCountLimit`:
```

specification/trace/sdk.md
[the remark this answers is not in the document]
tigrannajaryan: Maybe remove the actual names used here and simply say that names should be consistent with environment variables.

specification/trace/sdk.md
```diff
@@ -284,18 +285,43 @@ Thus, the SDK specification defines sets of possible requirements for
...
+}
+```
+
+Known properties explanation:
+
+* `AttributeCountLimit` - Maximum allowed span attribute count;
+* `EventCountLimit` - Maximum allowed span event count;
+* `LinkCountLimit` - Maximum allowed span link count;
+* `AttributePerEventCountLimit` - Maximum allowed attribute per span event count;
+* `AttributePerLinkCountLimit` - Maximum allowed attribute per span link count;
```
tigrannajaryan, line 324: It is not clear what this section describes.

specification/trace/sdk.md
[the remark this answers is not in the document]
bogdandrutu: Everywhere else we define the names in the code, and env follows that not the other way around. See BatchSpanProcessor. So probably just remove the part `to be consistent with the environment variables` and maybe keep the link to the env as suggested.

specification/trace/sdk.md
bogdandrutu: Done.

specification/trace/sdk.md
bogdandrutu: Done.

specification/trace/sdk.md
bogdandrutu: Done.

specification/trace/sdk.md
bogdandrutu: Consistency with other parts of the document.

tigrannajaryan (approved):

carlosalberto (approved):

jmacd (approved):

tedsuo (approved):
Since these limits are not required to be implemented, I think it is fine to add these limits now. We don't currently define what happens when these limits are surpassed. That is also fine; it should be left to the implementor to take what whatever action makes these limits effective in their language, should they need to implement them.

specification/trace/sdk.md
```diff
@@ -292,17 +293,44 @@ Optional parameters:
...
+  public int getLinkCountLimit();
+}
+```
+
+**Configurable parameters:**
+
+* `AttributeCountLimit` - Maximum allowed span attribute count;
+* `EventCountLimit` - Maximum allowed span event count;
+* `LinkCountLimit` - Maximum allowed span link count;
+* `AttributePerEventCountLimit` - Maximum allowed attribute per span event count;
```
anuraaga, line 330: This generally seems like a safe change but I would double check these two variables. For example, an alternative approach could be to allow the attribute count limit above to apply to all of the span, events, and links. Less knobs for similar safety - without properly evaluating the options I'm uncomfortable if this blocks other approaches or not. If it doesn't, it's fine.

anuraaga (approved):

specification/trace/sdk.md
bogdandrutu: I think usually people would like a limit way large for the span attributes than what they want for the attributes per event. I don't expect users to set the limit of attributes per event to be 1000, more or less around 32.

specification/trace/sdk.md
```diff
@@ -292,17 +293,44 @@ Optional parameters:
...
 these collections are unbounded, they can quickly exhaust available memory,
 resulting in crashes that are difficult to recover from safely.
 
 To protect against such errors, SDK Spans MAY discard attributes, links, and
 events that would increase the number of elements of each collection beyond
-the recommended limit of 1000 elements. SDKs MAY provide a way to change this limit.
+the recommended limit of 1000 elements.
 
-If there is a configurable limit, the SDK SHOULD honor the environment variables
+The SDKs MAY provide a way to change these limits, via a configuration to the
```
yurishkuro, line 306: L30 says MUST, which makes more sense, given that the default value of 1000 is completely arbitrary (and imho completely unreasonable in the general case - a span with 1000 attributes / events?)

bogdandrutu: Good call, I think the default is unreasonable.

bogdandrutu: Maybe `128`, but I will do a separate PR.

bogdandrutu:
@yurishkuro please approve if you think is ok, I updated to must

yurishkuro (approved):

bogdandrutu:
@tedsuo your comment is no longer applying based on the feedback from @yurishkuro.

iNikem (approved):

specification/trace/sdk.md
```diff
@@ -292,17 +293,44 @@ Optional parameters:
...
 resulting in crashes that are difficult to recover from safely.
 
 To protect against such errors, SDK Spans MAY discard attributes, links, and
 events that would increase the number of elements of each collection beyond
-the recommended limit of 1000 elements. SDKs MAY provide a way to change this limit.
+the recommended limit of 1000 elements.
 
-If there is a configurable limit, the SDK SHOULD honor the environment variables
+The SDKs MUST provide a way to change these limits, via a configuration to the
+TracerProvider, by allowing users to configure individual limits like the java
```
arminru, line 307:
```suggestion
TracerProvider, by allowing users to configure individual limits like in the Java
```

specification/trace/sdk.md
```diff
@@ -292,17 +293,44 @@ Optional parameters:
...
+example bellow. The name of the cofiguration class MAY be `SpanLimits`, and the
```
arminru, line 308:
```suggestion
example bellow. The name of the configuration class MAY be `SpanLimits`, and the
```

specification/trace/sdk.md
```diff
@@ -292,17 +293,44 @@ Optional parameters:
...
+name of the properties SHOULD be `AttributeCountLimit`, `EventCountLimit`,
+`LinkCountLimit`. Implementation MAY provide additional configuration such as
+`AttributePerEventCountLimit`, `AttributePerLinkCountLimit`.
```
arminru, line 311:
```suggestion
and `LinkCountLimit`. Implementations MAY provide additional configuration such as
`AttributePerEventCountLimit` and `AttributePerLinkCountLimit`.
```

arminru, line 311:
```suggestion
example bellow. The name of the configuration options SHOULD be `AttributeCountLimit`, `EventCountLimit`, and `LinkCountLimit`.
The options MAY be bundled in a class, which then SHOULD be called `SpanLimits`.
Implementation MAY provide additional configuration such as
`AttributePerEventCountLimit` and `AttributePerLinkCountLimit`.
```
Alternative to the two comments since I'd make it optional to group them in a class.

specification/trace/sdk.md
```diff
@@ -292,17 +293,44 @@ Optional parameters:
...
 |present|false|false|`localParentNotSampled()`|
 
-## Limits on Span Collections
+## Span Limits
 
 Erroneous code can add unintended attributes, events, and links to a span. If
 these collections are unbounded, they can quickly exhaust available memory,
 resulting in crashes that are difficult to recover from safely.
 
 To protect against such errors, SDK Spans MAY discard attributes, links, and
```
arminru, line 302: Since the limits MUST be configurable anyway now, should we also make this a MUST?

specification/trace/sdk.md
arminru: Should we also adapt the default for these two then? Currently it's 1000 for all.

specification/trace/sdk.md
```diff
@@ -292,17 +293,44 @@ Optional parameters:
...
+
+**Configurable parameters:**
+
+* `AttributeCountLimit` - Maximum allowed span attribute count;
+* `EventCountLimit` - Maximum allowed span event count;
+* `LinkCountLimit` - Maximum allowed span link count;
+* `AttributePerEventCountLimit` - Maximum allowed attribute per span event count;
+* `AttributePerLinkCountLimit` - Maximum allowed attribute per span link count;
+
+If there is a configurable provided, the SDK SHOULD honor the environment variables
```
arminru, line 333:
```suggestion
If there is a configuration provided, the SDK SHOULD honor the environment variables
```

arminru, line 333: The meaning of "SHOULD honor" is not clear to me. I'd clearly specify if the environment variables overrule the limits set on the TracerProvider in code or the other way round.
Also, "SHOULD honor" sounds like if the environment variables would _only_ be applicable if a configuration was provided in code.

specification/trace/sdk.md
```diff
@@ -292,17 +293,44 @@ Optional parameters:
...
 these collections are unbounded, they can quickly exhaust available memory,
 resulting in crashes that are difficult to recover from safely.
 
 To protect against such errors, SDK Spans MAY discard attributes, links, and
 events that would increase the number of elements of each collection beyond
-the recommended limit of 1000 elements. SDKs MAY provide a way to change this limit.
+the recommended limit of 1000 elements.
 
-If there is a configurable limit, the SDK SHOULD honor the environment variables
+The SDKs MUST provide a way to change these limits, via a configuration to the
```
arminru, line 306: Please add an entry to the spec compliance matrix for the new MUST requirements.

Oberon00 (changes requested):
This adds a nontrivial new MUST requirement right before release. Note that support for environment variables in SDKs is only a MAY, so SDKs are likely to not have this implemented at all.
Also, since we immediately make this stable without review period (except for this PR), our options to fix issues later are limited.

I don't understand why this seems to be so urgent. The affected env vars are marked experimental now, so this is not blocking release of 1.0, right?

Oberon00:
@tedsuo 

> Since these limits are not required to be implemented, I think it is fine to add these limits now.

The PR makes it a MUST to make it configurable though. Technically you could make the configuration interface a no-op I suppose, but is this really the intention here?

yurishkuro:
> This adds a nontrivial new MUST requirement right before release.

Good call. I think it's better to say "*if the SDK implements the limits* then the limits MUST be configurable"

carlosalberto:
> if the SDK implements the limits then the limits MUST be configurable

I like this, seems like a good trade-off.

bogdandrutu:
@Oberon00 please unblock the PR just in case I don't fix this until you close the day, if you are ok with @yurishkuro suggestion and I will do that.

specification/trace/sdk.md
bogdandrutu: Done.

specification/trace/sdk.md
bogdandrutu: Done.

specification/trace/sdk.md
bogdandrutu: done.

specification/trace/sdk.md
bogdandrutu: Done.

specification/trace/sdk.md
bogdandrutu: We agreed that if the limits are implemented then this is a MUST.

Oberon00:
I still don't understand why we must rush this into 1.0, but since the TC seems to be in broad agreement and the MAY/MUST issue was fixed, I will do as requested and lift my request-changes.

carlosalberto:
@Oberon00 

One factor is the desire to keep environment variables (if implemented by the SIG) as stable as possible, as they are user-facing API points. At the same time, I think there are potential improvements that can be done on top without breaking this (as #1425

bogdandrutu:
@Oberon00 all languages that are in RC or close to GA implemented this functionality in different ways, I made a search and saw that js has something, java has something else etc. That configuration was in the RC so I think is better to have it in the specs since at least 5 different languages implemented this (indeed differently).

yurishkuro (approved):

specification/trace/sdk.md
```diff
@@ -26,9 +26,10 @@ supplied to the `TracerProvider` must be used to create an
 [`InstrumentationLibrary`][otep-83] instance which is stored on the created
 `Tracer`.
 
-Configuration (i.e., [Span processors](#span-processor), [IdGenerator](#id-generators),
-and [`Sampler`](#sampling)) MUST be managed solely by the `TracerProvider` and it
-MUST provide some way to configure them, at least when creating or initializing it.
```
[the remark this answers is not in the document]
Oberon00: This MUST did not include span limits.

specification/trace/sdk.md
```diff
@@ -26,9 +26,10 @@ supplied to the `TracerProvider` must be used to create an
...
+Configuration (i.e., [SpanProcessors](#span-processor), [IdGenerator](#id-generators),
+[SpanLimits](#span-limits) and [`Sampler`](#sampling)) MUST be managed solely by
+the `TracerProvider` and it MUST provide some way to configure them, at least
```
Oberon00, line 31:
```suggestion
the `TracerProvider` and it MUST provide some way to configure all of them that are implemented in the SDK, at least
```

Otherwise we may accidentally introduce a MUST for the SpanLimits.

specification/trace/sdk.md
```diff
@@ -292,17 +293,46 @@ Optional parameters:
...
+## Span Limits
 
 Erroneous code can add unintended attributes, events, and links to a span. If
 these collections are unbounded, they can quickly exhaust available memory,
 resulting in crashes that are difficult to recover from safely.
 
 To protect against such errors, SDK Spans MAY discard attributes, links, and
 events that would increase the number of elements of each collection beyond
-the recommended limit of 1000 elements. SDKs MAY provide a way to change this limit.
+a recommended limit.
```
Oberon00, line 304:
```suggestion
a configured limit.
```

specification/trace/sdk.md
```diff
@@ -292,17 +293,46 @@ Optional parameters:
...
+
+**Configurable parameters:**
+
+* `AttributeCountLimit` (Default=1000) - Maximum allowed span attribute count;
+* `EventCountLimit` (Default=1000) - Maximum allowed span event count;
+* `LinkCountLimit` (Default=1000) - Maximum allowed span link count;
+* `AttributePerEventCountLimit` (Default=128) - Maximum allowed attribute per span event count;
+* `AttributePerLinkCountLimit` (Default=128) - Maximum allowed attribute per span link count;
+
+If there is a configuration provided, the SDK SHOULD honor the environment variables
```
Oberon00, line 335: This seems counter-intuitive. The env vars SHOULD override programmatic setting of these values?

Oberon00:
Seeing as there are many comments / discussion is still somewhat active I think we should at least white the usual 48 h period before merging this.

specification/trace/sdk.md
bogdandrutu: Done.

specification/trace/sdk.md
bogdandrutu: Done.

bogdandrutu:
@Oberon00 all the comments are about "wording" not about the meaning of the change. I think we all agreed that this is a MUST is you implement the limits which is a MAY. I don't think the discussion is controversial in any way.

Oberon00:
I agreee the discussion is not controversial. But on the other hand, if the goal is to keep this stable, a review period of the *usual* length will help to have a solid wording where we don't need any "hotfixes" later. I just have a bad feeling about rushing this in so fast for 1.0.

specification/trace/sdk.md
bogdandrutu: Done.

specification/trace/sdk.md
bogdandrutu: What about removing this sentence completely?

bogdandrutu: Decided to remove it, since we don't have the same statement in any other place where we offer env options. Consistency wins.

specification/trace/sdk.md
bogdandrutu: This sentence is not necessary, anywhere else we don't refer to the env, so I think it is better to just remove it.

specification/trace/sdk.md
arminru: I think we should clearly specify if the environment variables overrule the limits set on the TracerProvider in code or the other way round. Ideally we would specify such precedence globally to be consistent but that's out of scope of this PR.

specification/trace/sdk.md
bogdandrutu: Done.

spec-compliance-matrix.md
```diff
@@ -70,6 +70,7 @@ status of the feature is not known.
 | ShouldSample gets full parent Context                                                            |          |    | +    | +  | +      | +    | +      |     |      | +   | -    | +     |
 | [New Span ID created also for non-recording Spans](specification/trace/sdk.md#sdk-span-creation) |          |    |      |    | +      | +    |        |     |      |     | -    | +     |
 | [IdGenerators](specification/trace/sdk.md#id-generators) ]                                       |          |    |      |    |        |      |        |     |      |     |      |       |
+| [SpanLimits](specification/trace/sdk.md#span-limits) ]                                           |          |    |      |    |        |      |        |     |      |     |      |       |
```
cijothomas, line 73: Should we mark this as optional in this matrix, as the span-limits are optional sdk feature?

bogdandrutu: Sure.

bogdandrutu:
@Oberon00 I documented in the PR description the decision discussed with the TC members.

Issue #182, opened 2019-07-13
Limitation on number of attributes, links, events and behavior


lmolkova:
In Java SDK implementation, there is a limit and behavior is to discard in FIFO.

The limitations are not part of the spec and behavior when there are more than max is not defined.

austinlparker:
Seems like there’s a couple of options:

- Spec defines a hard max limit, unconfigurable. 
- Spec defines a default max, user-adjustable. 
- Spec defines no limit, exporters can select a limit (config of this limit is an exporter-specific concern)

The third option is the most similar to OpenTracing imo. I think the second option might be the best because there’s always going to be some sort of maximum, might as well put it in a well-known place.

discostu105:
I'd have a slight preference for the third option:

If the limits are user-adjustable via API, aren't we mixing different personas? An application developer (who would set the limit) might not know the exporter and ultimatively the backend that uses the data. The backend might have difficulties with 10 or with 10k attributes. It should be a concern of the exporter/backend only.

austinlparker:
That's a good point.

mtwo:
#3 also seems like the most reasonable choice to me, though I'm not extremely opinionated about this topic

bogdandrutu:
3 is not a good option because the intent for this is to not grow memory unbounded, this is not to protect the exporter or the backend, is to help Application owners to control how much memory their instrumentation uses.

I think the best compromise is 2, the limit is adjustable via the SDK not via the API (so an instrumentation dev, as you pointed does not know the backend), but the Application owner knows the backend in almost all the situations.

SergeyKanzhelev:
2 and 3 are almost identical in a sense that if exporter can configure this SDK setting, than it is the same. Multiple exporters will just need to set the max of options and then do additional trimming on their side. There is no way to prevent exporters to configure SDK as long as this setting is available for end user.

jmacd:
Removed this from the 0.3 milestone as it's not a blocker. We can add this in the 0.4 milestone.

iNikem:
Do we also want to impose/allow to configure some limits on the _size_ of attributes values?

Oberon00:
For attribute and label keys, there is discussion about a maximum length (among other things) on #504.

toumorokoshi:
[...]

PR #1419, opened 2021-02-10
Update default for span limits to 128, 1000 seems off

bogdandrutu:
128 is based on the value used in OpenCensus and it was not a problem there. See https://github.com/census-instrumentation/opencensus-java/blob/3d7f0b511dea392bbf68ee5268dd95857552dd87/api/src/main/java/io/opencensus/trace/config/TraceParams.java#L36

The limits for attributes and links were smaller, but I applied a max(of_all) function.

### DIFF
--- a/specification/sdk-environment-variables.md
+++ b/specification/sdk-environment-variables.md
@@ -77,9 +77,9 @@ Depending on the value of `OTEL_TRACES_SAMPLER`, `OTEL_TRACES_SAMPLER_ARG` may b
 
 | Name                            | Description                          | Default | Notes |
 | ------------------------------- | ------------------------------------ | ------- | ----- |
-| OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT | Maximum allowed span attribute count | 1000    |       |
-| OTEL_SPAN_EVENT_COUNT_LIMIT     | Maximum allowed span event count     | 1000    |       |
-| OTEL_SPAN_LINK_COUNT_LIMIT      | Maximum allowed span link count      | 1000    |       |
+| OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT | Maximum allowed span attribute count | 128     |       |
+| OTEL_SPAN_EVENT_COUNT_LIMIT     | Maximum allowed span event count     | 128     |       |
+| OTEL_SPAN_LINK_COUNT_LIMIT      | Maximum allowed span link count      | 128     |       |
 
 ## OTLP Exporter
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ New:
 
 Updates:
 
+- Update default limit for span attributes, events, links to 128([#1419](https://github.com/open-telemetry/opentelemetry-specification/pull/1419))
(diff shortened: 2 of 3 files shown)


yurishkuro (approved):

specification/sdk-environment-variables.md
```diff
@@ -77,9 +77,9 @@ Depending on the value of `OTEL_TRACES_SAMPLER`, `OTEL_TRACES_SAMPLER_ARG` may b
 
 | Name                            | Description                          | Default | Notes |
 | ------------------------------- | ------------------------------------ | ------- | ----- |
-| OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT | Maximum allowed span attribute count | 1000    |       |
-| OTEL_SPAN_EVENT_COUNT_LIMIT     | Maximum allowed span event count     | 1000    |       |
-| OTEL_SPAN_LINK_COUNT_LIMIT      | Maximum allowed span link count      | 1000    |       |
+| OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT | Maximum allowed span attribute count | 128     |       |
```
anuraaga, line 80: Do we need to reset the compliance matrix for these variables?

bogdandrutu: Is it an entry there?

bogdandrutu: Oh, reset the `x`es

bogdandrutu: I will do it in a followup PR, or after I merge #1420

anuraaga (approved):

arminru (approved):

Oberon00:
I can tell you where the 1000 comes from: https://github.com/open-telemetry/opentelemetry-specification/issues/182#issuecomment-690883490

It was chosen because

> And to @arminru's point,we just choose a number that's well beyond 99.99% of use cases. Don't want to throw numbers out without data, but 1000 seems well above the needed amount.
-- @toumorokoshi


128 was also discussed as limit but thrown out because it might prevent valid use cases, see https://github.com/open-telemetry/opentelemetry-specification/issues/182#issuecomment-690285768

As this is only an experimental part of the spec, I think it's OK to rush a change here, but if combined with #1416, I would oppose it.

arminru:
I still think 1000 is a good default to allow for 99.9% of all use cases _if this is not configurable_. With #1416 making configurability a MUST, however, going for a default of 128 should be fine since users have the option to adapt it if they're doing something special that would require them to go beyond that.

Oberon00 (approved):
I see. I am a bit confused then too why this became 1000, given that the 1000 was based on the assumption that it is *not* configurable while the 1000 appears exactly as default value for a *configuration* option.

toumorokoshi:
> I see. I am a bit confused then too why this became 1000, given that the 1000 was based on the assumption that it is not configurable while the 1000 appears exactly as default value for a configuration option.

IIRC the original PR made span limits optionally configurable, and stated that the default or unconfigurable value is 1000.

FWIW 1000 was chosen over 128 because 128 could be ambiguous as some code-level limitation (e.g. right above an 8-bit signed integer), and 1000 is clearly a human-defined limit.

I'm not really sure why 128 vs 1000 (or visa versa) necessitates a change. but not here to debate the change, just add clarity.

PR #1095, opened 2020-10-15
Add semantic conventions for the AWS SDK.

anuraaga:
Fixes #968 (by being the first instrumentation-specific instrumentation)

## Changes

This defines semantic attributes for instrumentation of the AWS SDK. We have instrumentation in several languages and they fill in some traditional semantic fields with varying coverage. AWS has also identified fields in request / response that are valuable for monitoring and we think this is a good list, though will naturally grow more, of attributes to get meaningful information from AWS API calls without the problems of copying the entire request/response.

The attributes are sort of "denormalized" as they are intended to be extracted from API calls, with instructions for each, and some attributes are common among the calls. This seems easy to implement, but let me know how this structure looks. Particularly interested in @thisthat's advice in terms of the YAML representations.

PR #1422, opened 2021-02-10
AWS SDK common and dynamodb conventions

### LINKED ISSUE #968: Where to put instrumentation-specific semantic conventions.
I think it's common for there to be a set of semantic conventions for specific instrumentation. These are often for frameworks with a cross-language scope, or client libraries for a service that exist in many languages. Information that are useful in observability that are always present in these situations make sense to have consistent in spans across languages.

We have an example already in our spec of instrumentation-specific conventions for gRPC.

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/rpc.md#grpc

I am working on semantic conventions for the AWS SDK client libraries

https://github.com/anuraaga/aws-opentelemetry-java-instrumentation/blob/master/docs/aws-sdk-semantic-conventions.md

These are conventions that AWS has collected through working with customers and are a set of data that we find helps with observability. Of course, we're happy to increase this list even more :) But the main idea is we think it is a good idea to have these conventions implemented in all instrumentations of the AWS SDK, regardless of language and want to make sure there is a reference point for instrumentation writers.

In Java instrumentation, we already have some attributes that are useful to users that are not documented anywhere, but should presumably be in the spec and filled by all languages instrumenting the same system (or not filled anywhere).

https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/602

For example, Kafka tombstone, elasticsearch actions.

Is it appropriate to have an `instrumentations` subfolder in `semantic_conventions` to put these sort of conventions in? We can start by extracting out `grpc.md` since it's already in this spec.

anuraaga:
Fixes #968 (by being the first instrumentation-specific instrumentation)

## Changes

This is a reopening of #1095 - towards the end of the year things got busy and I wasn't able to continue on it, but now that we are implementing AWS SDK instrumentation in various languages, we could use some conventions to base them off of.

This defines semantic attributes for instrumentation of the AWS SDK. We have instrumentation in several languages and they fill in some traditional semantic fields with varying coverage. AWS has also identified fields in request / response that are valuable for monitoring and we think this is a good list, though will naturally grow more, of attributes to get meaningful information from AWS API calls without the problems of copying the entire request/response.


semantic_conventions/trace/instrumentation/aws-sdk.yml
```diff
@@ -0,0 +1,337 @@
...
+        type: number
+        brief: "The value of the `ProvisionedThroughput.ReadCapacityUnits` request parameter."
+        examples:
+          - 1
+          - 2
+      - id: provisioned_throughput.write_capacity_units
+        type: number
+        brief: "The value of the `ProvisionedThroughput.WriteCapacityUnits` request parameter."
+        examples:
+          - 1
```
anuraaga, line 103: I think these are supposed to be `double`, but the generator only supports `number` I think, and the examples have to be integer.

Oberon00: CC @thisthat I somehow thought this was already fixed, but https://github.com/open-telemetry/build-tools/issues/13 is still open, so I may be mistaken.

willarmiros (approved):
LGTM!

semantic_conventions/trace/instrumentation/aws-sdk.yml
```diff
@@ -0,0 +1,337 @@
...
+          - S3
+      - id: operation
+        type: string
+        brief: "The name of the operation corresponding to the request, as returned by the AWS SDK"
+        examples:
+          - GetItem
+          - PutItem
+
+  - id: dynamodb.shared
+    prefix: aws
```
Oberon00, line 26: Why is `dynamodb` not included in the prefix? For example `aws.consumed_capacity` does not have any connection with `dynamodb` without looking it up in the docs.

semantic_conventions/trace/instrumentation/aws-sdk.yml
```diff
@@ -0,0 +1,337 @@
...
+        examples:
+          - ALL_ATTRIBUTES
+          - COUNT
+
+  - id: dynamodb.batchgetitem
+    brief: DynamoDB.BatchGetItem
+    prefix: aws
+    constraints:
+      - any_of:
+        - [aws.table_names, aws.consumed_capacity]
```
Oberon00, line 143: What is the motivation for this any_of? How does consumed_capacity replace table_names?

semantic_conventions/trace/instrumentation/aws-sdk.yml
```diff
@@ -0,0 +1,337 @@
...
+                }
+              ],
+              "Projection": {
+                "NonKeyAttributes": [ "string" ],
+                "ProjectionType": "string"
+              }
+            }'
+    constraints:
+      - any_of:
+        - [ db.name, aws.consumed_capacity, aws.item_collection_metrics, aws.provisioned_throughput.read_capacity_units, aws.provisioned_throughput.write_capacity_units ]
```
Oberon00, line 199: What do you think you are declaring here? It currently means that at least one of all these attributes is required. So for example if I supply `aws.provisioned_throughput.write_capacity_units`, then I don't need `db.name` anymore or vice versa. Is this really what you want? These seem to be completely unrelated items.
Maybe you want to list them with `ref:` in the attribute list instead?

anuraaga: As far as I know, it's any one of the _set_ of attributes. There is only one set here, and they're all required within the set.

I want to include these attributes in the documentation for how to instrument this method and couldn't find another way. Do you have a pointer to `ref:`?

anuraaga: More descriptively, ideally I want output like

```
The following shared attributes are also defined for this method:

- `db.name`
- `aws.consumed_capacity`
```

I could of course just do it in markdown if this isn't an appropriate use for the yaml but all the features of the yaml aren't very clear to me.

Oberon00: Oh, I missed the nested array! But if there is only one set, this also becomes misleading (I'd even say wrong) in the rendered markdown.

Oberon00: You can use an `include` constraint if you want to reference a whole other semantic convention. You can reference other attributes by using `ref:` instead of `id:` in the attribute list.

Oberon00: I recommend just grepping for those in the existing conventions, or look in syntax.md

anuraaga: Thanks a lot, I didn't understand `ref:` well enough, it seems to be what I needed. Also added `.dynamodb` to the prefix.

specification/trace/semantic_conventions/instrumentation/aws-sdk.md
```diff
@@ -0,0 +1,181 @@
...
+<!-- endsemconv -->
+
+### DynamoDB.DeleteTable
+
+<!-- semconv dynamodb.deletetable -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| [`db.name`](../database.md) | string | The value of the TableName request parameter. [1] | `customers`; `main` | No |
+
+**[1]:** In some SQL databases, the database name to be used is called "schema name".
```
anuraaga, line 73: Unfortunately somthing like `note: ''` didn't work to override this to remove it (we already have a clearer `Description`)

anuraaga:
Can anyone take another look at this? Thanks!

semantic_conventions/trace/instrumentation/aws-sdk.yml
```diff
@@ -0,0 +1,373 @@
+groups:
+  - id: aws
+    prefix: aws
+    brief: >
+      The `aws` conventions apply to operations using the AWS SDK. They map request or response parameters
+      in AWS SDK API calls to attributes on a Span. The conventions have been collected over time based
+      on feedback from AWS users of tracing and will continue to increase as new interesting conventions
```
iNikem, line 7:
```suggestion
      on feedback from AWS users of tracing and will continue to evolve as new interesting conventions
```

semantic_conventions/trace/instrumentation/aws-sdk.yml
```diff
@@ -0,0 +1,373 @@
...
+      - id: operation
+        type: string
+        brief: "The name of the operation corresponding to the request, as returned by the AWS SDK"
+        examples:
+          - GetItem
+          - PutItem
+
+  - id: dynamodb.shared
+    extends: aws
+    prefix: aws.dynamodb
```
iNikem, line 28: I am still confused about when to use nested namespaces and when not :)

semantic_conventions/trace/instrumentation/aws-sdk.yml
```diff
@@ -0,0 +1,373 @@
...
+    attributes:
+      - id: table_names
+        type: string[]
+        brief: The keys in the `RequestItems` object field in a request operating on multiple tables.
+        examples:
+          - Users
+          - Cats
+      - id: consumed_capacity
+        type: string[]
+        brief: "The JSON-serialized value of each item in the `ConsumedCapacity` response field."
```
iNikem, line 39: I am confused :) Comment above says "Attributes that exist for multiple DynamoDB request types", but these attributes seems like unique for specific commands?

semantic_conventions/trace/instrumentation/aws-sdk.yml
anuraaga: We'll have `aws.sqs`, `aws.s3` so I guess it makes sense to namespace by aws service.

semantic_conventions/trace/instrumentation/aws-sdk.yml
anuraaga: Yeah, they don't exist for _all_ DynamoDB request types, but exist for multiple, specific ones. Should I reword it?

anuraaga:
I don't believe I have any action items currently - can anyone give this another look? Thanks!

semantic_conventions/trace/instrumentation/aws-sdk.yml
```diff
@@ -0,0 +1,373 @@
...
+                          "NULL": boolean,
+                          "S": "string",
+                          "SS": [ "string" ]
+                       }
+                    },
+                    "SizeEstimateRangeGB": [ number ]
+                 }
+              ]
+             }'
+      - id: provisioned_throughput.read_capacity_units
```
Oberon00, line 95: While it is nice to have clear names, this name is also very long. The table does at least not fit in the "rich diff" view of GitHub with this, and might not fit in the normal view either. Also, depending on the protocol used to export spans, this may increase the overhead.

If there are any well-established acronyms for these I'd prefer to use them. Otherwise, `provisioned_read_capacity` or some similar shortening might work (the suffix `_units` does not seem to add much information).

semantic_conventions/trace/instrumentation/aws-sdk.yml
```diff
@@ -0,0 +1,373 @@
...
+      - ref: aws.dynamodb.table_names
+      - ref: aws.dynamodb.consumed_capacity
+      - ref: aws.dynamodb.item_collection_metrics
+
+  - id: dynamodb.deletetable
+    brief: DynamoDB.DeleteTable
+    extends: aws
+    prefix: aws.dynamodb
+    attributes:
+      - ref: db.name
```
Oberon00, line 223: This does not fit `db.name`. `db.name` is the name of the whole database, not a single table (there is `db.sql.table` though). I suggest to use `aws.dynamodb.table_names` but require only a single name is used.

Oberon00 (commented):
I wonder if it would be more helpful to make Dynamodb a `db` convention instead of / in addition to `aws` one.

You would probably use fields such as `db.system=dynamodb` (which is already a specified value by the way) and `db.operation` instead of/in addition to `aws.operation`.

semantic_conventions/trace/instrumentation/aws-sdk.yml
thisthat: I have created https://github.com/open-telemetry/build-tools/pull/30 :)

semantic_conventions/trace/instrumentation/aws-sdk.yml
anuraaga: Oops got confused for some reason, fixed

semantic_conventions/trace/instrumentation/aws-sdk.yml
anuraaga: Good point, I shortened a few that I found.

anuraaga (commented):
About sharing too much with the `db` convention, I think it only makes sense if any attributes can be shared with other DBs. There might be some like `limit` but I'd like to think about that later since it requires a bigger look at things.

For service / operation specifically, I realized that this is basically just RPC so may as well use those. It did bring to light #1462 which I think is ok.

semantic_conventions/trace/instrumentation/aws-sdk.yml
```diff
@@ -10,14 +10,12 @@ groups:
       Some descriptions are also provided for populating general OpenTelemetry semantic conventions based on
       these APIs.
     attributes:
-      - id: service
-        type: string
+      - ref: rpc.service
         brief: "The name of the service to which a request is made, as returned by the AWS SDK."
```
Oberon00, line 14: > as returned by the AWS SDK

That seems a bit strange to me. Is the name really *returned* by the SDK? Also, do we need some normalization here across different languages (e.g. Python might call it `create_key`, Java `createKey` and the AWS HTTP API reference `CreateKey`)?

anuraaga: Hmm - so the intent of my wording here was precisely to avoid the normalization. I think all AWS SDKs (at least should) return in a structured way the AWS service name and operation name, which correspond to the actual API docs and are of fixed case.

But as with everything, I found one language out of the four I just checked where maybe not :(

Java
https://github.com/aws/aws-xray-sdk-java/blob/master/aws-xray-recorder-sdk-aws-sdk-v2/src/main/java/com/amazonaws/xray/interceptors/TracingInterceptor.java#L123
`SERVICE_NAME`, `OPERATION_NAME`

Go
https://github.com/aws/aws-xray-sdk-go/blob/master/xray/aws.go#L351
`ServiceName`, `Operation.Name`

JS
https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/patchers/aws_p.js#L35
https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/patchers/aws_p.js#L63
`serviceIdentifier`, `operation`

Python
https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/ext/botocore/patch.py#L31
https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/ext/boto_utils.py#L42
`endpointPrefix`, `operation_name` (or just positional argument for some cases)

Given there does seem to be reasonable precedence for calling service / operation I think what I have here may be ok, but open to any suggestion.

Oberon00: OK, so there is runtime information about the called service name/operation, I didn't know that. Then this  seems fine to me.

Oberon00:
I wonder how using rpc.method/service will work if you use the AWS SDK to invoke a remote lambda method. In that case, maybe you would want to use the lambda's function name for rpc.method. But on the other hand, there is also faas.invoked_name, so it's probably no big issue (BTW, I wonder if we also want to introduce faas.invoked_id).

anuraaga:
@Oberon00 Yeah I think we run into similar issues on the server side of faas too, and it usually just affects the span name. Just need to pick which takes preference for name. Which reminded me I had forgotten to mention that! Added.

anuraaga:
Can anyone give this another look? Thanks!

Oberon00:
For sharing with `db`: I think at least you should require to use the (already existing) `db.system=dynamodb` attribute, so that the spans are recognizable as DB spans. Then I would also recommend to use `db.operation` instead of `rpc.method`, and `rpc.service` could be optional. If the technology-specific attributes are prefixed with `db.dynamodb.` or `aws.dynamodb.` is a matter of taste IMHO and I'm fine with either way.

But I think that using db.operation and db.system would already be valuable because it can allow a backend that does not know AWS semantic conventions to recognize such spans as DB-related. I don't know if there is something sensible for `db.name` to use, but that would also be valuable. The DB conventions also require at least one of `net.peer.name` or `net.peer.ip` to be set. That would in that case be the IP or hostname of the REST endpoint, which I think makes sense because at least it contains the region and may be useful for

Oberon00:
The other thing that looks worrisome to me is the long JSON strings in many attributes. Not only because the table format of markdown semantic conventions makes the examples look bad, but mainly because I think that tracing should not cause such large payloads to be sent. Won't this often be larger than the actual data that was queried by the user?

anuraaga:
@Oberon00 For DB, I think this relates to a somewhat general issue where spans can wear "multiple hats". Some backend logic that has been implemented to support `db.operation` will work for free if provided. Another backend logic that has been implemented to support `rpc.method` will work for free if provided. So we could just set them both to make backend logic as generic as possible. Or we may consider this to be too wasteful and it's up to the backend to support this specific request, e.g., `operation = attributes[rpc.method] || (attributes[db.system] == dynamodb ? attributes[db.operation] : null)`. This might not be too bad but it does bring to question why did we go with db.operation instead of rpc.method, do we have a priority that db somehow wins over rpc? To avoid the ambiguity, we'd just set both attributes though. Wondering what you think about these tradeoffs.

Oberon00:
My feeling is that db.operation is more well-fitting that rpc.method and I would only set db.operation. But I can see that setting both may also make sense in some cases, and semantically I see no drawback. It just feels wasteful.

anuraaga:
@Oberon00 Went with defining both for now as have a slight preference to having the `rpc.*` consistent among the service types, but interested to hear if anyone else has ideas on it.

> The other thing that looks worrisome to me is the long JSON strings in many attributes. 

Most of the long strings are request information, and aim to be a more (but not perfectly) structured form of the request object with fields that have not been very useful to customers excluded, so it's in some sense a compressed form of `db.statement`. There are a couple of response fields based on capacity / timing metrics returned from the service - cross-checking span performance with these can be useful, but if they seem a bit much for a first version we can leave them out for now.

anuraaga:
@open-telemetry/technical-committee Any suggestions on how to proceed to get this merged?

semantic_conventions/trace/instrumentation/aws-sdk.yml
Oberon00: I think it's fine as-is. @iNikem ?

specification/trace/semantic_conventions/instrumentation/aws-sdk.md
```diff
@@ -0,0 +1,173 @@
...
+parameters in AWS SDK API calls to attributes on a Span. The conventions have been collected over time based
+on feedback from AWS users of tracing and will continue to increase as new interesting conventions
+are found.
+
+Some descriptions are also provided for populating general OpenTelemetry semantic conventions based on these APIs.
+
+## Common Attributes
+
+The span name MUST be of the format `Service.Operation` as per the AWS HTTP API, e.g., `DynamoDB.GetItem`,
+`S3.ListBuckets`. This is equivalent to concatenating `rpc.service` and `rpc.method` with `.`.
```
Oberon00, line 15:
```suggestion
`S3.ListBuckets`. This is equivalent to concatenating `rpc.service` and `rpc.method` with `.` and consistent with the naming guidelines for RPC client spans.
```

specification/trace/semantic_conventions/instrumentation/aws-sdk.md
```diff
@@ -0,0 +1,173 @@
...
+
+<!-- semconv aws -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| [`rpc.method`](../rpc.md) | string | The name of the operation corresponding to the request, as returned by the AWS SDK | `GetItem`; `PutItem` | No |
+| [`rpc.service`](../rpc.md) | string | The name of the service to which a request is made, as returned by the AWS SDK. | `DynamoDB`; `S3` | No |
+| [`rpc.system`](../rpc.md) | string | The value `aws-api`. | `aws-api` | Yes |
+<!-- endsemconv -->
+
+## DynamoDB
```
Oberon00, line 25: As discussed at https://github.com/open-telemetry/opentelemetry-specification/pull/1422#issuecomment-787771585:

I want to have the DB attribute `db.system="dynamodb"` required for all DynamoDB spans. And please also add the suggestion (fine as MAY, SHOULD or MUST for me) to set `db.operation` to the same as `rpc.method`.

This is my only blocking comment.

anuraaga: @Oberon00 Oops, sorry about that I didn't notice my YAML change wasn't in a way that would get into the markdown, fixed it.

specification/trace/semantic_conventions/instrumentation/aws-sdk.md
```diff
@@ -0,0 +1,184 @@
...
+
+### Common Attributes
+
+These attributes are filled in for all DynamoDB request types.
+
+<!-- semconv dynamodb.all -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| [`db.system`](../database.md) | string | The value `dynamodb`. | `dynamodb` | No |
```
Oberon00, line 35: Please make it required.

anuraaga: Oops, done

carlosalberto:
@anuraaga would you mind marking the comments as resolved? this way it will be easier to review this PR (and hopefully we will have it merged _soon_ ;) )

anuraaga:
@carlosalberto Sure resolved

semantic_conventions/trace/instrumentation/aws-sdk.yml
arminru: https://github.com/open-telemetry/opentelemetry-specification/pull/1550 bumps the semconvgen reference to v0.3.0 which includes this change.

anuraaga:
Is this blocked on #1550? Is it possible to at least get approvals so this would get merged soon after? Thanks!

Oberon00 (approved):

Oberon00:
> Is this blocked on #1550? 

Don't think so. Whoever is merged second, will have to resolve the build errors of the other before though 😃

Oberon00:
@open-telemetry/specs-approvers @open-telemetry/technical-committee Trusting @anuraaga with the AWS-specifics, I think these semantic conventions are pretty sound and useful. The only thing I would prefer more opinions on is the JSON values for the attributes, which can be rather large. The only other attribute that can be similarly large that I know from the top of my head is the exception.stacktrace.

anuraaga:
Is it fair for me to give this a daily ping? Hoping to merge seen been more than a month :)

anuraaga:
Ah I need to update now that the numbers PR is in, will do it

anuraaga:
Ok updated to new number style

carlosalberto (approved):

carlosalberto:
> The only thing I would prefer more opinions on is the JSON values for the attributes, which can be rather large. The only other attribute that can be similarly large that I know from the top of my head is the exception.stacktrace.

If it gets on average as long as a stacktrace, we are cool, I think. What's your own impression on this @anuraaga ?

anuraaga:
@carlosalberto I think they'd be about the size of a typical web server stack trace in Java. An even closer reference is probably MySQL EXPLAIN query - things like the index stats and collection metrics are very similar to what an EXPLAIN query would return so the size is sort of in the same range (meaning the more indexes, the more potential output too)

carlosalberto:
Trusting @anuraaga expertise, merging it, thanks!

anuraaga:
Thanks @carlosalberto - added changelog

specification/trace/semantic_conventions/instrumentation/aws-sdk.md
```diff
@@ -0,0 +1,184 @@
+# Semantic conventions for AWS SDK
```
arminru, line 1: Please also mention this new file in the TOC in `specification/trace/semantic_conventions/README.md` to make it better discoverable.

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ release.
 
 ### Semantic Conventions
 
+- Add semantic conventions for the AWS operations and DynamoDB([#1442](https://github.com/open-telemetry/opentelemetry-specification/pull/1422))
```
arminru, line 30:
```suggestion
- Add semantic conventions for AWS SDK operations and DynamoDB ([#1442](https://github.com/open-telemetry/opentelemetry-specification/pull/1422))
```

anuraaga:
@arminru Thanks, updated

anuraaga:
@arminru I went ahead and moved AWS Lambda into the new section on README it seems more consistent (and hoping for more from various cross-language instrumentations!)

specification/trace/semantic_conventions/README.md
```diff
@@ -21,9 +21,13 @@ The following semantic conventions for spans are defined:
 * [RPC/RMI](rpc.md): Spans for remote procedure calls (e.g., gRPC).
 * [Messaging](messaging.md): Spans for interaction with messaging systems (queues, publish/subscribe, etc.).
 * [FaaS](faas.md): Spans for Function as a Service (e.g., AWS Lambda).
-  * [AWS Lambda](instrumentation/aws-lambda.md): Details on populating attributes for AWS Lambda.
 * [Exceptions](exceptions.md): Attributes for recording exceptions associated with a span.
 
+The following instrumentation-specific semantic conventions are defined:
```
arminru, line 26:
```suggestion
The following vendor-specific semantic conventions are defined:
```
Or what would "instrumentation-specific" mean?

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ release.
 
 ### Semantic Conventions
 
+- Add semantic conventions for AWS SDK operations and DynamoDB ([#1442](https://github.com/open-telemetry/opentelemetry-specification/pull/1422))
```
arminru, line 32:
```suggestion
- Add semantic conventions for AWS SDK operations and DynamoDB ([#1422](https://github.com/open-telemetry/opentelemetry-specification/pull/1422))
```

specification/trace/semantic_conventions/instrumentation/aws-sdk.md
```diff
@@ -0,0 +1,184 @@
...
+| `aws.dynamodb.consumed_capacity` | string[] | The JSON-serialized value of each item in the `ConsumedCapacity` response field. | `[{ "CapacityUnits": number, "GlobalSecondaryIndexes": { "string" : { "CapacityUnits": number, "ReadCapacityUnits": number, "WriteCapacityUnits": number } }, "LocalSecondaryIndexes": { "string" : { "CapacityUnits": number, "ReadCapacityUnits": number, "WriteCapacityUnits": number } }, "ReadCapacityUnits": number, "Table": { "CapacityUnits": number, "ReadCapacityUnits": number, "WriteCapacityUnits": number }, "TableName": "string", "WriteCapacityUnits": number }]` | No |
+| `aws.dynamodb.item_collection_metrics` | string | The JSON-serialized value of the `ItemCollectionMetrics` response field. | `{ "string" : [ { "ItemCollectionKey": { "string" : { "B": blob, "BOOL": boolean, "BS": [ blob ], "L": [ "AttributeValue" ], "M": { "string" : "AttributeValue" }, "N": "string", "NS": [ "string" ], "NULL": boolean, "S": "string", "SS": [ "string" ] } }, "SizeEstimateRangeGB": [ number ] } ] }` | No |
+| `aws.dynamodb.table_names` | string[] | The keys in the `RequestItems` object field in a request operating on multiple tables. | `[Users, Cats]` | No |
```
arminru, line 54: Most attributes are repeated in each table. Would it make sense to define them once at the beginning and then define for each `DynamoDB.*` operation, which of these attributes need to be set? Otherwise one would have to look at each definition to double-check that it is consistent with the others for the same key. It seems as if the meaning/content would not differ context-wise (other than single vs multiple elements in `table_names`).

specification/trace/semantic_conventions/instrumentation/aws-sdk.md
```diff
@@ -0,0 +1,184 @@
...
+<!-- endsemconv -->
+
+### DynamoDB.DeleteTable
+
+<!-- semconv dynamodb.deletetable -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `aws.dynamodb.table_names` | string[] | A single-element array with the value of the TableName request parameter. | `[Users, Cats]` | No |
```
arminru, line 86: > A single-element array with the value ... | `[Users, Cats]`

Description and example don't match here.

Same for some of the definitions below.

specification/trace/semantic_conventions/README.md
anuraaga: gRPC is a cross language library which I imagine would also go in this list with its instrumentation guidelines. Or DB client libraries like elastic search, couchbase, kafka, etc. We have some conventions for these sort of cases in the general conventions but I think we can have more detail and cover more specific use cases if we move them to instrumentation conventions.

Particularly don't like the word vendor since we usually refer to a tracing vendor but these aren't tracing related.

arminru: Ah I see! Yeah vendor is not suitable then.
I just found instrumentation in this case rather odd because a) all semantic conventions are defined for instrumentations and b) they are defined so that all instrumentations for the same framework/library/technology use the same conventions - so it's the exact opposite of something instrumentation-specific, isn't it? What about framework-specific, technology-specific or maybe a vague "specialized" (as opposed to general)? Sigh, naming... 😄

anuraaga: I see I think you're reading instrumentation as a general concept, while I get it but actually use the word more often to refer to instrumentation for a certain library - e.g. AWS SDK instrumentation, gRPC instrumentation, etc so instrumentation-specific. But library-specific seems like it'd be appropriate and remove the ambiguity

specification/trace/semantic_conventions/instrumentation/aws-sdk.md
anuraaga: If you look at the yaml, that is actually how they're defined. I didn't print a shared table at the top since it seems the generator inlines the spec anyways. IIUC, your main concern is the maintenance, in which case I guess this is ok? I personally do like the final rendering with it inlined since it makes each section easier to read.

specification/trace/semantic_conventions/instrumentation/aws-sdk.md
anuraaga: Oops yeah didn't notice I have to override both brief and examples for the single-item use case, fixed

carlosalberto:
@arminru Is this ready to go?

specification/trace/semantic_conventions/instrumentation/aws-sdk.md
arminru: I also like that all methods/tables can be read self-contained but my other concern was that instrumentation authors and span consumer reading it will have to compare the tables to see if the semantics of an attribute differ between the different methods/tables. Not a blocking issue though and can also be fixed later if desired, since it's just a change in how the tables are organized but not underlying definitions.

arminru (approved):
> @arminru Is this ready to go?

@carlosalberto Just a minor concern left that's not addressed but isn't blocking and could also be done afterwards. Feel free to merge it.

specification/trace/semantic_conventions/instrumentation/aws-sdk.md
anuraaga: I'd like to keep it as is if that's ok. This is both defining conventions but also being provided as an implementation guide - I think the current format is nicer to guide implementation, though I guess there is some tradeoff for when understanding them for the backends but I'd err on the side of implementation here.

arminru: Sure! Both options have their pros and cons but both are fine.

semantic_conventions/trace/instrumentation/aws-sdk.yml
z1c0: @Oberon00 why that?
> ... require only a single name is used.
Multiple affected tables are a valid use case.

PR #1439, opened 2021-02-12
Add none for exporter selection enums.

anuraaga:
## Changes

Adds `none` as an option for exporter selection. Because the default is `otlp`, there is no way to disable signal export via env vars.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ release.
 
 ## Unreleased
 
+- Adds `none` as a possible value for OTEL_TRACES_EXPORTER and OTEL_METRICS_EXPORTER to disable export([#1439](https://github.com/open-telemetry/opentelemetry-specification/pull/1439))
+
 ## v1.0.1 (2021-02-11)
 
 - Fix rebase issue for span limit default values ([#1429](https://github.com/open-telemetry/opentelemetry-specification/pull/1429))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ release.
 
 ## Unreleased
 
-- Adds `none` as a possible value for OTEL_TRACES_EXPORTER and OTEL_METRICS_EXPORTER to disable export([#1439](https://github.com/open-telemetry/opentelemetry-specification/pull/1439))
+- Adds `none` as a possible value for OTEL_TRACES_EXPORTER and OTEL_METRICS_EXPORTER to disable export
+  ([#1439](https://github.com/open-telemetry/opentelemetry-specification/pull/1439))
 
 ## v1.0.1 (2021-02-11)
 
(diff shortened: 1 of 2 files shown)


specification/sdk-environment-variables.md
```diff
@@ -141,11 +141,13 @@ Known values for OTEL_TRACES_EXPORTER are:
 - `"otlp"`: [OTLP](./protocol/otlp.md)
 - `"jaeger"`: [Jaeger gRPC](https://www.jaegertracing.io/docs/1.21/apis/#protobuf-via-grpc-stable)
 - `"zipkin"`: [Zipkin](https://zipkin.io/zipkin-api/) (Defaults to [protobuf](https://github.com/openzipkin/zipkin-api/blob/master/zipkin.proto) format)
+- `"none"`: Disables export of traces.
```
Oberon00, line 144: Nit: No sure about "disable". If the application adds its own exporter, it probably won't get disabled by this setting.

Oberon00 (approved):

specification/sdk-environment-variables.md
anuraaga: Thanks tried tweaking

iNikem (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ release.
 
 ## Unreleased
 
+- Adds `none` as a possible value for OTEL_TRACES_EXPORTER and OTEL_METRICS_EXPORTER to disable export([#1439](https://github.com/open-telemetry/opentelemetry-specification/pull/1439))
```
arminru, line 10:
```suggestion
- Adds `none` as a possible value for OTEL_TRACES_EXPORTER and OTEL_METRICS_EXPORTER to disable export
  ([#1439](https://github.com/open-telemetry/opentelemetry-specification/pull/1439))
```

arminru (approved):

carlosalberto (approved):

bogdandrutu (changes requested):
I am confused about this change. What is the scenario that we want people to "disable" trace but still link all the trace SDK?

trask (approved):
> What is the scenario that we want people to "disable" trace but still link all the trace SDK?

We use this in the javaagent, so that vendors can disable the OTLP exporter and provide their own exporter, while still getting all of the other auto-configuration goodness.

reyang (approved):

anuraaga:
@bogdandrutu Yeah this is for the case we provide a batteries included package like the Java agent. Another case is wanting to only export traces without metrics.

SergeyKanzhelev (approved):

bogdandrutu (approved):

bogdandrutu:
@anuraaga @trask thanks for the explanation. Next time please describe this in the PR description or an issue, it is important to make sure that we have good reasons (documented) for what we change/add :) Every API/property we add will add support cost so we need to understand why we are paying this cost :)

jsuereth (approved):

PR #1442, opened 2021-02-16
Add semantic conventions for instrumenting AWS Lambda.

anuraaga:
## Changes

Provides direction for instrumenting AWS Lambda request handlers to generate faas spans. There aren't any new attributes defined, just explanation of how to fill them as well as concerns about parenting, etc. This is mostly based on the [Java implementation](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/aws-lambda-1.0/library) but is in no ways set in stone.

/cc @kubawach @wangzlei


semantic_conventions/trace/instrumentation/aws-lambda.yaml
```diff
@@ -0,0 +1,61 @@
+groups:
+  - id: aws.lambda
+    brief: "Lambda function invocation spans."
+    attributes:
+      - ref: faas.execution
+        brief: "The value of the AWS Request ID from the Lambda `Context`."
```
Oberon00, line 6: This is not what faas.execution is. The execution ID should be the same for all requests that the same runtime instance handles. For AWS Lambda I would specify the log stream name, or the hex-id part thereof.

semantic_conventions/trace/instrumentation/aws-lambda.yaml
```diff
@@ -0,0 +1,61 @@
...
+        examples:
+          - 943ad105-7543-11e6-a9ac-65e093327849
+      - ref: faas.id
+        brief: "The value of the invocation arn for the function from the Lambda `Context`."
+        examples:
+          - arn:aws:lambda:us-east-2:123456789012:function:my-function:1
+      - ref: cloud.account.id
+        brief: "The account ID for the function. If not provided on Lambda `Context`, it SHOULD be parsed from the value of `faas.id` as the fifth item when splitting on `:`."
```
Oberon00, line 14: That "brief" is not very brief

```suggestion
        brief: "The account ID for the function."
        note: >
          If not provided on Lambda `Context`, it SHOULD be parsed from the value of `faas.id`
          as the fifth item when splitting on `:`.
```

semantic_conventions/trace/instrumentation/aws-lambda.yaml
```diff
@@ -0,0 +1,61 @@
...
+  - id: aws.lambda.api-gateway-proxy
```
Oberon00, line 15:
```suggestion
  - id: aws.lambda.api_gateway_proxy
```

We never use `-` in our semantic convention names, see also https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-and-label-naming.md#attribute-and-label-naming

semantic_conventions/trace/instrumentation/aws-lambda.yaml
```diff
@@ -0,0 +1,61 @@
...
+          - process
+  - id: aws.lambda.sqs-message
+    brief: "These conventions apply to an SQS message."
+    attributes:
+      - ref: faas.trigger
+        brief: "MUST be `pubsub`."
+        examples:
+          - pubsub
+      - ref: messaging.system
+        brief: "MUST be `AmazonSQS`."
```
Oberon00, line 51: This pattern seems commonly useful enough that special support in the semantic convention generator would be nice as in `fixed_value: "AmazonSQS"`.

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,104 @@
+# Instrumenting AWS Lambda
+
+**Status**: [Experimental](../../../document-status.md)
+
+This document defines how to fill semantic conventions when instrumenting an AWS Lambda request handler. AWS
```
Oberon00, line 5:
```suggestion
This document defines how to apply semantic conventions when instrumenting an AWS Lambda request handler. AWS
```

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,104 @@
...
+
+<!-- semconv aws.lambda -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| [`cloud.account.id`](../../../resource/semantic_conventions/cloud.md) | string | The account ID for the function. If not provided on Lambda `Context`, it SHOULD be parsed from the value of `faas.id` as the fifth item when splitting on `:`. | `opentelemetry` | No |
+| [`faas.execution`](../faas.md) | string | The value of the AWS Request ID from the Lambda `Context`. | `943ad105-7543-11e6-a9ac-65e093327849` | No |
+| [`faas.id`](../../../resource/semantic_conventions/faas.md) | string | The value of the invocation arn for the function from the Lambda `Context`. [1] | `arn:aws:lambda:us-east-2:123456789012:function:my-function:1` | No |
+
+**[1]:** For example, in AWS Lambda this field corresponds to the [ARN](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html) value, in GCP to the URI of the resource, and in Azure to the [FunctionDirectory](https://github.com/Azure/azure-functions-host/wiki/Retrieving-information-about-the-currently-running-function) field.
```
Oberon00, line 24: That copied generic note is confusing. Does `note: ""` work in the YAML? Otherwise this would be worth a bugfix to the tool IMHO.

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,104 @@
...
+<!-- endsemconv -->
+
+The parent of the span MUST be determined by considering both the environment and any headers or attributes
+available from the event.
+
+If the `_X_AMZN_TRACE_ID` environment variable is set, it SHOULD be parsed into an OpenTelemetry `Context` using
```
Oberon00, line 30: It seems to me that the application owner should be able to decide this. Ideally, the X-Ray propagator (or a specialized LambdaXrayPropagator) would be written in such a way that it checks the environment variable itself. The instrumentation should not decide this.

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,104 @@
...
+## API Gateway
+
+API Gateway allows a user to trigger a Lambda function in response to HTTP requests. It can be configured to be
+a pure proxy, where the information about the original HTTP request is passed to the Lambda function, or as a
+configuration for a REST API, in which case only a deserialized body payload is available.  In the case the API
+gateway is configured to proxy to the Lambda function, the instrumented request handler will have access to all
+the information about the HTTP request in the form of an API Gateway Proxy Request Event.
+
+The Lambda span name SHOULD be set to the `Resource` from the proxy request event, which corresponds to the user
+configured HTTP route instead of the function name.
```
Oberon00, line 48: Would it make sense to set `http.route` too?

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,104 @@
...
+This document defines how to fill semantic conventions when instrumenting an AWS Lambda request handler. AWS
+Lambda largely follows the conventions for [FaaS](../faas.md) while [HTTP](../http.md) conventions are also
+applicable when handlers are for HTTP requests.
+
+There are a variety of triggers for Lambda functions, and this document will grow over time to cover all the
+use cases.
+
+## All triggers
+
+For all events, a span with kind `SERVER` MUST be created corresponding to the function invocation. Unless
```
Oberon00, line 14:
```suggestion
For all events, a span with kind `SERVER` MUST be created corresponding to the function invocation unless stated otherwise below. Unless
```

SQS wants CONSUMER

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,104 @@
...
+gateway is configured to proxy to the Lambda function, the instrumented request handler will have access to all
+the information about the HTTP request in the form of an API Gateway Proxy Request Event.
+
+The Lambda span name SHOULD be set to the `Resource` from the proxy request event, which corresponds to the user
+configured HTTP route instead of the function name.
+
+<!-- semconv aws.lambda.api-gateway-proxy -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| [`faas.trigger`](../faas.md) | string | MUST be `http`. | `http` | No |
```
Oberon00, line 53: Consider using `required: always` in the YAML.

semantic_conventions/trace/instrumentation/aws-lambda.yaml
anuraaga: That's `faas.instance` right? `faas.execution`  is defined as the request ID.

Oberon00: Oh, you are right, I mixed them up. 🙈  Then this looks fine!

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
anuraaga: We generate two additional spans in addition for SQS. IIRC this was to have a span that contains the function name.

I notice after reading through the code to write this that for HTTP we actually are overwriting the name of the top span instead of generating a new one. I guess this should make sense for SQS as well, I can reflect that (and we'd need to update the Java).

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
anuraaga: Nope same issue in my other PR

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
anuraaga: The user can currently decide by enabling or disabling XRay so I think this reflects that choice well. If we had another option in the instrumentation it's two settings related to XRay which I think is just an extra setting.

It reminds me that in a follow-up to the SDK PR I need to describe using the AWS propagator directly as we do in Java. It's the only recognized format for the next years so there is no point for an option at least until that changes it's just an extra option. I think it's similar for Lambda.

semantic_conventions/trace/instrumentation/aws-lambda.yaml
anuraaga: It reminded me I'll need to send another PR for the Resource side too :)

semantic_conventions/trace/instrumentation/aws-lambda.yaml
```diff
@@ -0,0 +1,61 @@
...
+    attributes:
+      - ref: faas.trigger
+        brief: "MUST be `http`."
+        examples:
+          - http
+      - ref: http.method
+      - ref: http.status_code
+      - ref: http.url
+      - ref: http.user_agent
+  - id: aws.lambda.sqs-event
```
rakyll, line 28: aws.lambda.sqs_event

semantic_conventions/trace/instrumentation/aws-lambda.yaml
```diff
@@ -0,0 +1,61 @@
...
+          - pubsub
+      - ref: messaging.system
+        brief: "MUST be `AmazonSQS`."
+        examples:
+          - AmazonSQS
+      - ref: messaging.operation
+        brief: "MUST be `process`."
+        examples:
+          - process
+  - id: aws.lambda.sqs-message
```
rakyll, line 43: Ditto.

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,104 @@
...
+
+## All triggers
+
+For all events, a span with kind `SERVER` MUST be created corresponding to the function invocation. Unless
+stated otherwise below, the name of the span MUST be set to the function name from the Lambda `Context`.
+
+<!-- semconv aws.lambda -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| [`cloud.account.id`](../../../resource/semantic_conventions/cloud.md) | string | The account ID for the function. If not provided on Lambda `Context`, it SHOULD be parsed from the value of `faas.id` as the fifth item when splitting on `:`. | `opentelemetry` | No |
```
rakyll, line 20: A better example here is 123456789012.

semantic_conventions/trace/instrumentation/aws-lambda.yaml
```diff
@@ -0,0 +1,61 @@
...
+    attributes:
+      - ref: faas.execution
+        brief: "The value of the AWS Request ID from the Lambda `Context`."
+        examples:
+          - 943ad105-7543-11e6-a9ac-65e093327849
+      - ref: faas.id
+        brief: "The value of the invocation arn for the function from the Lambda `Context`."
+        examples:
+          - arn:aws:lambda:us-east-2:123456789012:function:my-function:1
+      - ref: cloud.account.id
```
rakyll, line 13: What about region?

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,104 @@
+# Instrumenting AWS Lambda
```
rakyll, line 1: This document doesn't mention its relationship to the FaaS spec, https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/trace/faas.yaml. Is it extending it or overriding it?

semantic_conventions/trace/instrumentation/aws-lambda.yaml
```diff
@@ -0,0 +1,61 @@
+groups:
+  - id: aws.lambda
+    brief: "Lambda function invocation spans."
+    attributes:
+      - ref: faas.execution
+        brief: "The value of the AWS Request ID from the Lambda `Context`."
+        examples:
+          - 943ad105-7543-11e6-a9ac-65e093327849
+      - ref: faas.id
```
rakyll, line 9: If the attributes here are not Lambda specific, can we extend the https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/trace/faas.yaml instead? If not, maybe we should namespace them differently. For example, "aws.lambda.id", "aws.lambda.execution", etc.

semantic_conventions/trace/instrumentation/aws-lambda.yaml
Oberon00: Maybe you could add an `include`  constraint for `faas`  and `cloud`?

semantic_conventions/trace/instrumentation/aws-lambda.yaml
Oberon00: I think this was the intention of @anuraaga. FaaS IMHO sufficiently describes the common lambda attributes (e.g. it tells you that the ARN is the faas.id in the faas' note already). This document just describes how to apply existing semantic conventions to Lambda. So it is IMHO right to refer to (note that `ref`  is used here instead of introducing a new attribute with `id`) the existing, fitting attributes.

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
anuraaga: Doesn't the first paragraph clarify this, including an actual link to faas? Let me know if something's unclear.

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
anuraaga: Yeah think so

anuraaga (commented):
Thanks, updated

semantic_conventions/trace/instrumentation/aws-lambda.yaml
anuraaga: This doc is only about span attributes for now, I'll need to follow up with `Resource`, which is where region is. Account ID would be there too if not for the arn-at-runtime restriction we're familiar with. Is it ok to clarify this later?

Also, @Oberon00, I've tried include but couldn't find any change in the markdown. Is it supposed to affect the rendering?

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
anuraaga: @Oberon00 Updated to only have two SQS spans. @kubawach can you confirm this should be ok?

anuraaga:
Can anyone give this another look? Thanks!

semantic_conventions/trace/instrumentation/aws-lambda.yaml
Oberon00: `include` should be more or less the same as `ref`-ing all attributes I think.

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
Oberon00: This seems like a mistake to me, should faas.trigger really be optional here?

Oberon00:
I wonder if using the semantic convention generator here is more confusing than useful. Just writing a plain markdown file and explaining (without listing every single attribute) which conventions to use in which combination for Lambda could be enough IMHO.

anuraaga:
@Oberon00 Tried inlining into the markdown, how does it look?

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,90 @@
...
+
+## All triggers
+
+For all events, a span with kind `SERVER` MUST be created corresponding to the function invocation unless stated
+otherwise below. Unless stated otherwise below, the name of the span MUST be set to the function name from the
+Lambda `Context`.
+
+The following attributes SHOULD be set.
+
+- [`faas.execution`](../faas.md) - This is always available through an accessor on the Lambda `Context`
```
Oberon00, line 20: IIRC this should be set to the request ID?

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,90 @@
...
+- [`faas.id`](../../../resource/semantic_conventions/faas.md) - This is always available through an accessor on the Lambda `Context`
+- [`cloud.account.id`](../../../resource/semantic_conventions/cloud.md) - In some languages, this is available as an accessor on the Lambda `Context`. Otherwise, it can be parsed from the value of `faas.id` as the fifth item when splitting on `:`
+
+The parent of the span MUST be determined by considering both the environment and any headers or attributes
+available from the event.
```
Oberon00, line 25:
```suggestion
available from the event, as explained in the remainder of this section.
```

Or add a subheading above here.

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,90 @@
...
+
+## API Gateway
+
+API Gateway allows a user to trigger a Lambda function in response to HTTP requests. It can be configured to be
+a pure proxy, where the information about the original HTTP request is passed to the Lambda function, or as a
+configuration for a REST API, in which case only a deserialized body payload is available.  In the case the API
+gateway is configured to proxy to the Lambda function, the instrumented request handler will have access to all
+the information about the HTTP request in the form of an API Gateway Proxy Request Event.
+
+The Lambda span name SHOULD be set to the `Resource` from the proxy request event, which corresponds to the user
```
Oberon00, line 44:
```suggestion
The Lambda span name and the [`http.route` span attribute](../http.md) SHOULD be set to the `Resource` from the proxy request event, which corresponds to the user
```

Resolves https://github.com/open-telemetry/opentelemetry-specification/pull/1442/files#r576699975

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,90 @@
...
+[`faas.trigger`](../faas.md) MUST be set to `http`. [HTTP attributes](../http.md) SHOULD be set based on the
+available information in the proxy request event.
+
+## SQS
+
+SQS is a message queue that triggers a Lambda function with a batch of messages. So we consider processing both
+of a batch and of each individual message. The function invocation span MUST correspond to the SQS event, which
+is the batch of messages. For each message, an additional span SHOULD be created to correspond with the handling
+of the SQS message. Because handling of a message will be inside user business logic, not the Lambda framework,
+automatic instrumentation mechanisms without code change will often not be able to be able to instrument the
```
Oberon00, line 56:
```suggestion
automatic instrumentation mechanisms without code change will often not be able to instrument the
```

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,90 @@
...
+For every message in the event, the message's system attributes (not message attributes, which are provided by
+the user) SHOULD be checked for the key `AWSTraceHeader`. If it is present, an OpenTelemetry `Context` SHOULD be
+parsed from the value of the attribute using the [AWS X-Ray Propagator](../../../context/api-propagators.md) and
+added as a link to the span. This means the span may have as many links as messages in the batch.
+
+[`faas.trigger`](../faas.md) MUST be set to `pubsub`.
+[`messaging.operation`](../messaging.md) MUST be set to `process`.
+[`messaging.system`](../messaging.md) MUST be set to `AmazonSQS`.
+
+For the SQS message span, the name MUST be `<event source> process`.  The parent MUST be the `CONSUMER` span
```
Oberon00, line 74: Please add subsection headings for message spans and event span for clarity.

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,90 @@
...
+be `<event source> process`. If there are multiple sources in the batch, the name MUST be
+`multiple_sources <process>`. The parent MUST be the `SERVER` span corresponding to the function invocation.
+
+For every message in the event, the message's system attributes (not message attributes, which are provided by
+the user) SHOULD be checked for the key `AWSTraceHeader`. If it is present, an OpenTelemetry `Context` SHOULD be
+parsed from the value of the attribute using the [AWS X-Ray Propagator](../../../context/api-propagators.md) and
+added as a link to the span. This means the span may have as many links as messages in the batch.
+
+[`faas.trigger`](../faas.md) MUST be set to `pubsub`.
+[`messaging.operation`](../messaging.md) MUST be set to `process`.
```
Oberon00, line 71: It irks me that both the span receiving the message and the spans processing the messages have the "process" operation. But I think you are right here and the "receive" operation is meant for synchronous receiving as it is currently defined.

Maybe this case would warrant a special `messaging.operation`, maybe `dispatch`? "Used for spans corresponding to a middleware component that gets spans received by a lower level component but contains no business logic for processing the messages, instead passing the message(s) on to other components which process it."

Although from what I understand, the processing would happen synchronously inside child spans of this span, or, if there is no instrumentation for the actual processing, there would be no child spans and this span is the closes equivalent to "process".

Oberon00 (approved):

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,90 @@
...
+There are a variety of triggers for Lambda functions, and this document will grow over time to cover all the
+use cases.
+
+## All triggers
+
+For all events, a span with kind `SERVER` MUST be created corresponding to the function invocation unless stated
+otherwise below. Unless stated otherwise below, the name of the span MUST be set to the function name from the
+Lambda `Context`.
+
+The following attributes SHOULD be set.
```
jsuereth, line 18: Does it make sense to use `MUST` for `faas.execution`/`faas.id`?  Asking because I'd like to better understand "instrumentation semconv" and how strict/lenient we're measuring compliance.

jsuereth (commented):
Non-Blocking comments:

I love the idea instrumentation conventions!

1. Can you provide a link to the java implementation in the PR header so it shows up in the commit record?
2. I see the implementation has "agent" and "library" usage.   Do you envision "library" being ubiquitously provided across OTel languages and "agent' when applicable?

Asking for two reasons:
- Likely Cloud Run / CloudFunctions should look very similar to what you've done here (baring the nuance/details of our environment).
- I think you may be paving the way for how the "instrumentation" work @tedsuo calls out in the OTel timeline can make progress.

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
anuraaga: Oops, thanks!

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
anuraaga: Yeah I think a separate operation type could make sense for the batch case. Presuming for now this is ok to go with

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
anuraaga: Yeah I'm not too sure about this so open to suggestions. I generally went with `SHOULD` for attributes read from the instrumentation library mainly because for me to comfortably say `MUST`, I'd need to verify every single language's Lambda framework to make sure these actually are available. Perhaps I should do so, though I think for some frameworks this can be extremely tedious (many languages have 2-3 versions of the AWS SDK by now...), so leaning towards SHOULD indicating if they're available, fill them plz.

anuraaga (commented):
> I see the implementation has "agent" and "library" usage. Do you envision "library" being ubiquitously provided across OTel languages and "agent' when applicable?

Yup I think this is what we'd expect. For Java, wherever possible we are using the "agent" module only to initialize the "library" module but not itself define any instrumentation. Most languages won't have the concept of an agent though, these conventions definitely refer to library instrumentation.

> I think you may be paving the way for how the "instrumentation" work @tedsuo calls out in the OTel timeline can make progress.

Cool - yeah I'm thinking we need these instrumentation "manuals" or else languages will end up diverging, losing a lot of the value of these specc'd out conventions. Some libraries are pretty easy, thinking HTTP clients like Okhttp, and may not need manuals, but many have corner cases to deal with. And very much hoping to avoid [this](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/internal/awsxray/aws.go#L23)...

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
Oberon00: CC @arminru

anuraaga:
@open-telemetry/technical-committee Any suggestions on how to proceed to get this merged?

arminru:
@anuraaga It would be great if you could get a ✔️ from one of our other AWS experts like @rakyll and @alolita, so that the @open-telemetry/specs-approvers know everything is sound from your side. Also please make sure that all conversation are resolved, there are quite a few still open. Thanks!

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
anuraaga: Added subheading, thanks

anuraaga:
Thanks @arminru I've resolved the ones that I made changes for, left open threads that I responded but may still have some feedback coming back.

@rakyll @willarmiros @kubawach Do you mind reviewing / approving this PR? Thanks!

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
kubawach: +1 for being more flexible here (number of SDKs x number of available instrumentations)

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,96 @@
...
+## SQS
+
+SQS is a message queue that triggers a Lambda function with a batch of messages. So we consider processing both
+of a batch and of each individual message. The function invocation span MUST correspond to the SQS event, which
+is the batch of messages. For each message, an additional span SHOULD be created to correspond with the handling
+of the SQS message. Because handling of a message will be inside user business logic, not the Lambda framework,
+automatic instrumentation mechanisms without code change will often not be able to instrument the processing of
+the individual messages.
+
+The span kind for both spans MUST be `CONSUMER`.
```
kubawach, line 61: If I follow this correctly, processing a single SQS event with 2 messages should yield following trace (parent - child links):
SQS event
 - SQS msg 1
 - SQS msg 2
 - spans from autoinstrumentation

kubawach (approved):
Great work, very happy to see these semantic conventions coming to life :)

willarmiros (approved):
LGTM overall!

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
anuraaga: Ah, `both spans` is a bit misleading, reworded a bit

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,96 @@
...
+
+### SQS Event
+
+For the SQS event span, if all the messages in the event have the same event source, the name of the span MUST
+be `<event source> process`. If there are multiple sources in the batch, the name MUST be
+`multiple_sources <process>`. The parent MUST be the `SERVER` span corresponding to the function invocation.
```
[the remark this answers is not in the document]
anuraaga: Thanks, a typo indeed.

Messaging spec is definitely not set in stone so could be worth starting a Discussion on it.

anuraaga:
@arminru @willarmiros has approved as a double check from AWS, and @kubawach has been the most involved in Lambda out of anyone in the community. Are there any more AIs or can we get this in?

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,96 @@
+# Instrumenting AWS Lambda
```
arminru, line 1: Please add this to the README one level above where the other semantic conventions are mentioned (preferably as a sub-item for FaaS).

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,96 @@
...
+## All triggers
+
+For all events, a span with kind `SERVER` MUST be created corresponding to the function invocation unless stated
+otherwise below. Unless stated otherwise below, the name of the span MUST be set to the function name from the
+Lambda `Context`.
+
+The following attributes SHOULD be set.
+
+- [`faas.execution`](../faas.md) - The value of the AWS Request ID, which is always available through an accessor on the Lambda `Context`
+- [`faas.id`](../../../resource/semantic_conventions/faas.md) - The value of the invocation arn for the function, which is always available through an accessor on the Lambda `Context`
```
arminru, line 21:
```suggestion
- [`faas.id`](../../../resource/semantic_conventions/faas.md) - The value of the invocation ARN for the function, which is always available through an accessor on the Lambda `Context`
```

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,96 @@
...
+There are a variety of triggers for Lambda functions, and this document will grow over time to cover all the
+use cases.
+
+## All triggers
+
+For all events, a span with kind `SERVER` MUST be created corresponding to the function invocation unless stated
+otherwise below. Unless stated otherwise below, the name of the span MUST be set to the function name from the
+Lambda `Context`.
+
+The following attributes SHOULD be set.
```
arminru, line 18:
```suggestion
The following attributes SHOULD be set:
```

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,96 @@
...
+
+## API Gateway
+
+API Gateway allows a user to trigger a Lambda function in response to HTTP requests. It can be configured to be
+a pure proxy, where the information about the original HTTP request is passed to the Lambda function, or as a
+configuration for a REST API, in which case only a deserialized body payload is available.  In the case the API
+gateway is configured to proxy to the Lambda function, the instrumented request handler will have access to all
+the information about the HTTP request in the form of an API Gateway Proxy Request Event.
+
+The Lambda span name and the [`http.route` span attribute](../http.md) SHOULD be set to the `Resource` from the
```
arminru, line 46: When reading this I briefly thought this would be about the OpenTelemetry Resource due to the capitalization and formatting. Could you link to the definition for this event here or in the previous sentence so one could look it up when reading?

This would also clear it up:
```suggestion
The Lambda span name and the [`http.route` span attribute](../http.md#http-server-semantic-conventions) SHOULD be set to the `resource` property from the
```

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,96 @@
...
+proxy request event, which corresponds to the user configured HTTP route instead of the function name.
+
+[`faas.trigger`](../faas.md) MUST be set to `http`. [HTTP attributes](../http.md) SHOULD be set based on the
+available information in the proxy request event.
+
+## SQS
+
+SQS is a message queue that triggers a Lambda function with a batch of messages. So we consider processing both
```
arminru, line 54:
```suggestion
Amazon Simple Queue Service (SQS) is a message queue that triggers a Lambda function with a batch of messages. So we consider processing both
```
Since you already provide a definition/introduction here.

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,96 @@
...
+`multiple_sources process`. The parent MUST be the `SERVER` span corresponding to the function invocation.
+
+For every message in the event, the message's system attributes (not message attributes, which are provided by
+the user) SHOULD be checked for the key `AWSTraceHeader`. If it is present, an OpenTelemetry `Context` SHOULD be
+parsed from the value of the attribute using the [AWS X-Ray Propagator](../../../context/api-propagators.md) and
+added as a link to the span. This means the span may have as many links as messages in the batch.
+
+[`faas.trigger`](../faas.md) MUST be set to `pubsub`.
+[`messaging.operation`](../messaging.md) MUST be set to `process`.
+[`messaging.system`](../messaging.md) MUST be set to `AmazonSQS`.
```
arminru, line 76: You could add `AmazonSQS` to the examples listed for `messaging.system` so instrumentation authors can also find it outside the Lambda context. (Eventually we should make this an open enum with suggested values like we have for `db.system`).

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,96 @@
...
+
+The span kind for both types of SQS spans MUST be `CONSUMER`.
+
+### SQS Event
+
+For the SQS event span, if all the messages in the event have the same event source, the name of the span MUST
+be `<event source> process`. If there are multiple sources in the batch, the name MUST be
+`multiple_sources process`. The parent MUST be the `SERVER` span corresponding to the function invocation.
+
+For every message in the event, the message's system attributes (not message attributes, which are provided by
```
arminru, line 69:
```suggestion
For every message in the event, the _message system attributes_ (not message attributes, which are provided by
```
As it's called here: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-metadata.html#sqs-message-system-attributes

Ideally also add a reference to the docs here.

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,96 @@
...
+added as a link to the span. This means the span may have as many links as messages in the batch.
+
+[`faas.trigger`](../faas.md) MUST be set to `pubsub`.
+[`messaging.operation`](../messaging.md) MUST be set to `process`.
+[`messaging.system`](../messaging.md) MUST be set to `AmazonSQS`.
+
+### SQS Message
+
+For the SQS message span, the name MUST be `<event source> process`.  The parent MUST be the `CONSUMER` span
+corresponding to the SQS event. The message's system attributes (not message attributes, which are provided by
```
arminru, line 81:
```suggestion
corresponding to the SQS event. The _message system attributes_ (not message attributes, which are provided by
```

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,96 @@
...
+be `<event source> process`. If there are multiple sources in the batch, the name MUST be
+`multiple_sources process`. The parent MUST be the `SERVER` span corresponding to the function invocation.
+
+For every message in the event, the message's system attributes (not message attributes, which are provided by
+the user) SHOULD be checked for the key `AWSTraceHeader`. If it is present, an OpenTelemetry `Context` SHOULD be
+parsed from the value of the attribute using the [AWS X-Ray Propagator](../../../context/api-propagators.md) and
+added as a link to the span. This means the span may have as many links as messages in the batch.
+
+[`faas.trigger`](../faas.md) MUST be set to `pubsub`.
+[`messaging.operation`](../messaging.md) MUST be set to `process`.
```
arminru, line 75:
```suggestion
[`messaging.operation`](../messaging.md#messaging-attributes) MUST be set to `process`.
```
The prelude there is quite lengthy so the anchor would be helpful.

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
arminru, line 76:
```suggestion
[`messaging.system`](../messaging.md) MUST be set to `AmazonSQS`.
[`messaging.destination`](../messaging.md) MUST be set to `queue`.
```
Setting `messaging.destination` is required "if the message destination is either a queue or topic", see
https://github.com/open-telemetry/opentelemetry-specification/blob/v1.1.0/specification/trace/semantic_conventions/messaging.md#messaging-attributes.

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,96 @@
...
+
+For the SQS message span, the name MUST be `<event source> process`.  The parent MUST be the `CONSUMER` span
+corresponding to the SQS event. The message's system attributes (not message attributes, which are provided by
+the user) SHOULD be checked for the key `AWSTraceHeader`. If it is present, an OpenTelemetry `Context` SHOULD be
+parsed from the value of the attribute using the [AWS X-Ray Propagator](../../../context/api-propagators.md) and
+added as a link to the span.
+
+[`faas.trigger`](../faas.md) MUST be set to `pubsub`.
+[`messaging.operation`](../messaging.md) MUST be set to `process`.
+[`messaging.system`](../messaging.md) MUST be set to `AmazonSQS`.
```
arminru, line 88:
```suggestion
[`messaging.system`](../messaging.md) MUST be set to `AmazonSQS`.
[`messaging.destination`](../messaging.md) MUST be set to `queue`.
```

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,96 @@
...
+
+Other [Messaging attributes](../messaging.md) SHOULD be set based on the available information in the SQS message
+event.
+
+Note that `AWSTraceHeader` is the only supported mechanism for propagating `Context` for SQS to prevent conflicts
```
arminru, line 93: Supported by whom?

arminru (commented):
Thanks, @anuraaga! It would be nice if you could add one complete example showing a composition of all spans with their relationship to each other and the attributes, similar to https://github.com/open-telemetry/opentelemetry-specification/blob/v1.1.0/specification/trace/semantic_conventions/messaging.md#examples. This would make it easier to understand, especially the relationship between the SQS Event and Message spans.

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
Oberon00: I think it should be possible to use any OpenTelemetry-compatible vendor with SQS without paying for X-Ray. I'd like it more if we say something to the effect that "Instrumentations SHOULD default to using AWSTraceHeader with AWS X-Ray Propagator, but SHOULD be configurable to use any other header and propagator". If that makes sense at all.

Oberon00: Also related: https://github.com/open-telemetry/opentelemetry-specification/pull/1442#discussion_r576698621

anuraaga: I meant supported by our instrumentation. `AWSTraceHeader` doesn't mention the word X-Ray :) It's how to propagate context within AWS services and by having AWS SDK instrumentation and Lambda instrumentation explicitly use the X-Amzn-Trace-Id format (which we generally incorrectly call x-ray, it's actually not really tied to x-ray), it precisely allows any vendor to use AWS asynchronous services with proper propagation. No payment for X-Ray or anything involved. While it might be possible to implement our own propagation through message attributes, I'm not confident it would be propagated between services (e.g., S3 -> SNS -> SQS is a possible flow) while I am confident that `AWSTraceHeader` will be, regardless of tracing vendor. @kubawach has done incredible testing on these propagation cases, I'm guessing for use with Splunk, not X-Ray :P

I am going to try adding some text with some of this information to allay the concerns, which are fair but I wouldn't be worried :)

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
anuraaga: Thanks - didn't keep the italics since with the link it renders blue

anuraaga (commented):
Cleaned things up. @Oberon00 @arminru Hope I addressed the concerns about propagation - everything about parenting is to ensure it works correctly for any vendor, not just X-Ray :) @kubawach has verified this in production for some of the craziest scenarios too based on our Java code, and this doc is just outlining what we've done there - as long as the wording makes sense to you, I would take @kubawach's approval as a sign that we're not locking in.

anuraaga:
@arminru Also added some examples

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
kubawach: I support @anuraaga here. 

With the first approach to AWS propagation (for SQS producer - consumer) I used "standard" HTTP-headers mechanism, creating a SQS message attribute (upon produce) and extracting parent from it (upon consume). While this approach worked in SQS - SQS scenario, it would fail with more complex (but generally used around the world) scenarios such as S3 - SQS or S3 - SNS - SQS. Therefore I had to switch back to using AWS feature, guaranteeing that if AWS Trace header is set during AWS SDK request, the value will be maintained and returned at the end of the chain (consume - as SQS system attribute). There was simply no other way to do it (ie without relying on AWS features).

To sum up, approach that came out of discussions and code reviews was:
- when interacting with AWS (using AWS SDK) we will enforce X-Ray propagator (which basically sets AWS trace header in an appropriate format) in order to ensure that context propagation will be maintained thru the infrastructure (which is beyond our control)
- if there is a system that conforms to AWS interface and supports AWS SDK, in order to maintain the propagation it will also need to support AWS tracing header / format

Frankly at the beginning it didn't feel right to implement propagation enforcing AWS trace format but I realised that it's just relying on how a closed system work, just as we do with instrumentation libraries, in order to support as many use cases as possible.

Oberon00 (approved):
Thank you, that makes sense. It is annoying that we are not able to safely transport arbitrary key-value headers (e.g. standard W3C traceparent), but since that is an issue of the AWS platform I won't blame these semantic conventions.

anuraaga:
Resolved merge conflicts :)

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ release.
 
 ### Semantic Conventions
 
+- Add details for filling semantic conventiosn for AWS Lambda ([#1442](https://github.com/open-telemetry/opentelemetry-specification/pull/1442))
```
arminru, line 29:
```suggestion
- Add details for filling semantic conventions for AWS Lambda ([#1442](https://github.com/open-telemetry/opentelemetry-specification/pull/1442))
```

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,188 @@
...
+
+**Status**: [Experimental](../../../document-status.md)
+
+This document defines how to apply semantic conventions when instrumenting an AWS Lambda request handler. AWS
+Lambda largely follows the conventions for [FaaS](../faas.md) while [HTTP](../http.md) conventions are also
+applicable when handlers are for HTTP requests.
+
+There are a variety of triggers for Lambda functions, and this document will grow over time to cover all the
+use cases.
+
```
arminru, line 11:
```suggestion

<!-- Re-generate TOC with `markdown-toc --no-first-h1 -i` -->

<!-- toc -->
// toc goes here
<!-- tocstop -->

```

A TOC would be nice now that the file ended up being a bit longer.

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -66,31 +72,117 @@ For the SQS event span, if all the messages in the event have the same event sou
...
+### API Gateway Request Proxy (Lambda tracing passive)
+
+Given a process C that sends an HTTP request to an API Gateway endpoint with path `/pets/{petId}` configured for
+a Lambda function F:
+
+```
+Process C: | Span Client        |
+--
+Function F:    | Span Function |
```
arminru, line 123: nit: That sounds a bit odd (what's a span client?), I'd prefer "client span" and "function span".

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -66,31 +72,117 @@ For the SQS event span, if all the messages in the event have the same event sou
...
+
+| Field or Attribute | Span Prod1 | Span Prod2 | Span ProcBatch | Span Proc1 | Span Proc2 |
+|-|-|-|-|-|-|
+| Span name | `Q send` | `Q send` | `Q process` | `Q process` | `Q process` |
+| Parent |  |  |  | Span ProcBatch | Span ProcBatch |
+| Links |  |  |  | Span Prod1 | Span Prod2 |
+| SpanKind | `PRODUCER` | `PRODUCER` | `CONSUMER` | `CONSUMER` | `CONSUMER` |
+| Status | `Ok` | `Ok` | `Ok` | `Ok` | `Ok` |
+| `messaging.system` | `AmazonSQS` | `AmazonSQS` | `AmazonSQS` | `AmazonSQS` | `AmazonSQS` |
+| `messaging.destination` | `Q` | `Q` | | `Q` | `Q` |
```
arminru, line 174:
```suggestion
| `messaging.destination` | `Q` | `Q` | `Q` | `Q` | `Q` |
```
Why would you not set the Queue name here?

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -66,31 +72,117 @@ For the SQS event span, if all the messages in the event have the same event sou
...
+| `messaging.destination_kind` | `queue` | `queue` | `queue` | `queue` | `queue` |
+| `messaging.operation` |  |  | `process` | `process"` | `process` |
```
arminru, line 176:
```suggestion
| `messaging.operation` |  |  | `process` | `process` | `process` |
```

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -66,31 +72,117 @@ For the SQS event span, if all the messages in the event have the same event sou
...
+| `messaging.message_id` | | | | `"a1"` | `"a2"` |
+
+The above requires user code change to create `Span Proc1` and `Span Proc2` - in Java, the user would inherit from
+[TracingSqsMessageHandler][] instead of Lambda's standard `RequestHandler` to enable them. Otherwise the spans would
+not exist.
+
+[TracingSqsMessageHandler]: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingSqsMessageHandler.java
```
arminru, line 183:
```suggestion
[TracingSqsMessageHandler]: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/v1.0.1/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingSqsMessageHandler.java
```

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -66,31 +72,117 @@ For the SQS event span, if all the messages in the event have the same event sou
...
+| Links |  |  |  | Span Prod1 | Span Prod2 |
+| SpanKind | `PRODUCER` | `PRODUCER` | `CONSUMER` | `CONSUMER` | `CONSUMER` |
+| Status | `Ok` | `Ok` | `Ok` | `Ok` | `Ok` |
+| `messaging.system` | `AmazonSQS` | `AmazonSQS` | `AmazonSQS` | `AmazonSQS` | `AmazonSQS` |
+| `messaging.destination` | `Q` | `Q` | | `Q` | `Q` |
+| `messaging.destination_kind` | `queue` | `queue` | `queue` | `queue` | `queue` |
+| `messaging.operation` |  |  | `process` | `process"` | `process` |
+| `messaging.message_id` | | | | `"a1"` | `"a2"` |
+
+The above requires user code change to create `Span Proc1` and `Span Proc2` - in Java, the user would inherit from
```
arminru, line 179:
```suggestion
The above requires user code change to create `Span Proc1` and `Span Proc2`. In Java, the user would inherit from
```

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -66,31 +72,117 @@ For the SQS event span, if all the messages in the event have the same event sou
...
+[TracingSqsMessageHandler][] instead of Lambda's standard `RequestHandler` to enable them. Otherwise the spans would
+not exist.
```
arminru, line 181:
```suggestion
[TracingSqsMessageHandler][] instead of Lambda's standard `RequestHandler` to enable them. Otherwise these two spans
would not exist.
```

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -66,31 +72,117 @@ For the SQS event span, if all the messages in the event have the same event sou
...
+| `http.host` | | `foo.execute-api.us-east-1.amazonaws.com` |
+| `http.target` | | `/pets/10` |
+| `http.route` | | `/pets/{petId}` |
+| `http.status_code` | `200` | `200` |
+
+### API Gateway Request Proxy (Lambda tracing active)
+
+Active tracing in Lambda means an API Gateway span `Span APIGW` and a Lambda runtime invocation span `Span Lambda`
+will be exported to AWS X-Ray. All attributes above are the same except that in this case, the parent of `APIGW`
+is `Span Client` and the parent of `Span Function` is `Span Lambda`.
```
arminru, line 151: The two new spans are not defined in our semantic conventions, right?
Is the Lambda span a child of the APIGW span?
You could also add a "drawing" of the relationship like:
```
Span Client --> Span APIGW --> Span Lambda --> Span Function
```
with `x --> y` = `x parent of y`

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -66,31 +72,117 @@ For the SQS event span, if all the messages in the event have the same event sou
...
+The above requires user code change to create `Span Proc1` and `Span Proc2` - in Java, the user would inherit from
+[TracingSqsMessageHandler][] instead of Lambda's standard `RequestHandler` to enable them. Otherwise the spans would
+not exist.
+
+[TracingSqsMessageHandler]: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingSqsMessageHandler.java
+
+### SQS (Lambda tracing active)
+
+Active tracing in Lambda means a Lambda runtime invocation span `Span Lambda` will be exported to X-Ray. In this
+case, all of the above is the same except `Span ProcBatch` will have a parent of `Span Lambda`.
```
arminru, line 188: Same as above

arminru (commented):
Thanks, @anuraaga!

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
anuraaga: I borrowed this pattern from messaging.md. Should I change it?

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
anuraaga: `Q` is the queue name - I copied this pattern from messaging.md too :)

anuraaga (commented):
Thanks @arminru updated

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
arminru: Ah I see. There we use "Prod1" and "Proc1" there as symbols but "span client" sounds like someone consuming the span and "span function" sounds like something applied to a span to me. Should be fine given the context, however.

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
arminru: Sure, my question was why you would >not< also set `Q` for the `ProcBatch` span here.

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,213 @@
...
+
+| Field or Attribute | Span Prod1 | Span Prod2 | Span ProcBatch | Span Proc1 | Span Proc2 |
+|-|-|-|-|-|-|
+| Span name | `Q send` | `Q send` | `Q process` | `Q process` | `Q process` |
+| Parent |  |  |  | Span ProcBatch | Span ProcBatch |
+| Links |  |  |  | Span Prod1 | Span Prod2 |
+| SpanKind | `PRODUCER` | `PRODUCER` | `CONSUMER` | `CONSUMER` | `CONSUMER` |
+| Status | `Ok` | `Ok` | `Ok` | `Ok` | `Ok` |
+| `messaging.system` | `AmazonSQS` | `AmazonSQS` | `AmazonSQS` | `AmazonSQS` | `AmazonSQS` |
+| `messaging.destination` | `Q` | `Q` | | `Q` | `Q` | `Q` |
```
arminru, line 195:
```suggestion
| `messaging.destination` | `Q` | `Q` | `Q` | `Q` | `Q` |
```
One column too many :-)

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,213 @@
...
+| `http.status_code` | `200` | `200` |
+
+### API Gateway Request Proxy (Lambda tracing active)
+
+Active tracing in Lambda means an API Gateway span `Span APIGW` and a Lambda runtime invocation span `Span Lambda`
+will be exported to AWS X-Ray by the infrastructure (not instrumentation). All attributes above are the same
+except that in this case, the parent of `APIGW` is `Span Client` and the parent of `Span Function` is
+`Span Lambda`. This means the hierarchy looks like:
+
+Span Client --> Span APIGW --> Span Lambda --> Span Function
```
arminru, line 172: Please put this in a fenced code block (suggesting "```" here doesn't work apparently 😄).

In UML the arrows would be the other way round (`parent <-- child`). I'm not sure what would be most clear here.

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
anuraaga: Oops - thanks for the text I understand better. Actually the batch span can't have it since the same catch can serve multiple queues. I applied the suggestion thinking I made a column-length typo heh.

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -0,0 +1,213 @@
...
+[TracingSqsMessageHandler]: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/v1.0.1/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingSqsMessageHandler.java
+
+### SQS (Lambda tracing active)
+
+Active tracing in Lambda means a Lambda runtime invocation span `Span Lambda` will be exported to X-Ray by the
+infrastructure (not instrumentation). In this case, all of the above is the same except `Span ProcBatch` will
+have a parent of `Span Lambda`. This means the hierarchy looks like:
+
+Span Lambda --> Span ProcBatch --> Span Proc1 (links to Span Prod1 and Span Prod2)
+                               \-> Span Proc2 (links to Span Prod1 and Span Prod2)
```
arminru, line 213: Same as above (fences and arrow direction).

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
arminru: Ah I see! Well in this case it happens to be the same so if the batch would be inspected at that stage it could be added but that would be optional, so either should be fine here.

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
anuraaga: I laughed that I ended up making the column length typo anyways!

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
anuraaga: I added text to indicate what should happen if multiple queues. We might want to copy the text into `messaging.md` too

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
anuraaga: Added the fence. Kept the arrow direction, in tracing we're watching a request flow through steps, IMO having the order model the request direction is clearest.

arminru (approved):

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
carlosalberto: Is this comment addressed @arminru ?

carlosalberto (approved):

PR #1445, opened 2021-02-17
erlang process.runtime attributes: remove TODO note and update descriptions

tsloughter:
No real change here, simply removes the TODO since it is done and updates the attributes into a table.

Related issue: https://github.com/open-telemetry/opentelemetry-erlang/issues/96

### DIFF
--- a/specification/resource/semantic_conventions/process.md
+++ b/specification/resource/semantic_conventions/process.md
@@ -70,6 +70,12 @@ In addition to these attributes, [`telemetry.sdk.language`](README.md#telemetry-
 - `process.runtime.version` -  The version of the runtime (ERTS - Erlang Runtime System), i.e., `erlang:system_info(version)`.
 - `process.runtime.description` - string | An additional description about the runtime made by combining the OTP version, i.e., `erlang:system_info(otp_release)`, and ERTS version.
 
+Example:
+
+| `process.runtime.name` | `process.runtime.version` | `process.runtime.description` |
+| --- | --- | --- |
+| BEAM | 11.1 |  Erlang/OTP 23 erts-11.1 |
+
 ### Go Runtimes
 
 TODO(<https://github.com/open-telemetry/opentelemetry-go/issues/1181>): Confirm the contents here
--- a/specification/resource/semantic_conventions/process.md
+++ b/specification/resource/semantic_conventions/process.md
@@ -66,17 +66,9 @@ In addition to these attributes, [`telemetry.sdk.language`](README.md#telemetry-
 
 ### Erlang Runtimes
 
-TODO(<https://github.com/open-telemetry/opentelemetry-erlang/issues/96>): Confirm the contents here
-
-- `process.runtime.name` - The name of the Erlang runtime being used. Usually will be BEAM.
-- `process.runtime.version` - The ERTS (Erlang Runtime System) version. For BEAM this is found with `application:get_key(erts, vsn)`.
-- `process.runtime.description` - The OTP version `erlang:system_info(otp_release)` and ERTS version combined.
-
-Example:
-
-| Name | `process.runtime.name` | `process.runtime.version` | `process.runtime.description` |
-| --- | --- | --- | --- |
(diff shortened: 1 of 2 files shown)


yurishkuro (approved):
Does Erland SDK populate these already?

tsloughter:
I don't know what the semantic convention failure is about. Are docs on this in the repo?

tsloughter:
@yurishkuro in the latest `main` branch, yes.

tsloughter:
This reminded me to update the spec matrix, so I've now included that in this PR as well.

SergeyKanzhelev (approved):

specification/resource/semantic_conventions/process.md
```diff
@@ -66,17 +66,13 @@ In addition to these attributes, [`telemetry.sdk.language`](README.md#telemetry-
...
-- `process.runtime.name` - The name of the Erlang runtime being used. Usually will be BEAM.
-- `process.runtime.version` - The ERTS (Erlang Runtime System) version. For BEAM this is found with `application:get_key(erts, vsn)`.
-- `process.runtime.description` - The OTP version `erlang:system_info(otp_release)` and ERTS version combined.
-
-Example:
-
-| Name | `process.runtime.name` | `process.runtime.version` | `process.runtime.description` |
-| --- | --- | --- | --- |
-| beam | BEAM | 11.0.3 | Erlang/OTP 24 erts-11.0.3 |
+<!-- semconv process.runtime -->
```
arminru, line 69: This would add the same table as above when running the semantic conventions generator tool. Since this table is maintained manually, you'll have to remove those markers or the build fails. Currently it's complaining about a mismatch with the (potentially) generated files. See how it's done with the other tables below.

arminru (approved):

specification/resource/semantic_conventions/process.md
SergeyKanzhelev, line 69: this requires the yaml update: https://github.com/open-telemetry/opentelemetry-specification/tree/main/semantic_conventions#yaml-model-for-semantic-conventions

tsloughter:
Ah, woops, thanks @arminru and @SergeyKanzhelev :). I had copied from the earlier table in the document without thinking, should be fixed now.

specification/resource/semantic_conventions/process.md
```diff
@@ -66,17 +66,11 @@ In addition to these attributes, [`telemetry.sdk.language`](README.md#telemetry-
...
-- `process.runtime.name` - The name of the Erlang runtime being used. Usually will be BEAM.
-- `process.runtime.version` - The ERTS (Erlang Runtime System) version. For BEAM this is found with `application:get_key(erts, vsn)`.
-- `process.runtime.description` - The OTP version `erlang:system_info(otp_release)` and ERTS version combined.
-
-Example:
-
-| Name | `process.runtime.name` | `process.runtime.version` | `process.runtime.description` |
-| --- | --- | --- | --- |
-| beam | BEAM | 11.0.3 | Erlang/OTP 24 erts-11.0.3 |
+| Attribute  | Type | Description  | Examples  | Required |
```
Oberon00, line 69: To be consistent with other languages, I would keep the list form above.

bogdandrutu (approved):
Please resolve conflicts

specification/resource/semantic_conventions/process.md
Oberon00: Looking at it again, this even repeats the `Type` and `Required` column. I think this is rather misleading. Please either:

1. Revert to the list notation as above, just remove the TODO, maybe add a table with examples (as for other runtimes like Java) if you have more than one; or
2. remove the Type and Required column from the table.

Oberon00 (changes requested):

tsloughter:
Switched it back to a list and merged the spec matrix with the latest in `main`.

Oberon00 (approved):
OK, but you might want to keep the example somehow.

tsloughter:
Ok, added an example table.

PR #1452, opened 2021-02-19
Add ForceFlush to Tracing SDK

### LINKED ISSUE #1451: Add ForceFlush to Tracing SDK

I want to add `ForceFlush` to [Tracing SDK](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#shutdown) like it was done for the `Shutdown` method.

**Additional context.**

I'm using OpenTelemetry for AWS lambda functions. And I'm using dotnet opentelemetry library to achieve that. After each call, lambda goes to sleep and freezes background threads. Therefore I need to send activities to the exporter forcibly if I want to use the batch processor instead of the simple one.

Thankfully, `BaseProcessor` already contains a `ForceFlush` method that works perfectly in this case.

There is a PR that I've made for the dotnet library: https://github.com/open-telemetry/opentelemetry-dotnet/pull/1837
My PR is based on a similar Java PR, which you can find [here](https://github.com/open-telemetry/opentelemetry-java/pull/1068). And the Java PR has already been merged.

cc: @reyang @utpilla @cijothomas

kipwoker:
Fixes #1451

## Changes

Added `ForceFlush` to [Tracing SDK](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#shutdown) like it was done for the `Shutdown` method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ release.
 
 - Adds `none` as a possible value for OTEL_TRACES_EXPORTER and OTEL_METRICS_EXPORTER to disable export
   ([#1439](https://github.com/open-telemetry/opentelemetry-specification/pull/1439))
+- Add `ForceFlush` to [Tracing SDK](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#shutdown) ([#1452](https://github.com/open-telemetry/opentelemetry-specification/pull/1452))
 
 ## v1.0.1 (2021-02-11)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ release.
 
 - Adds `none` as a possible value for OTEL_TRACES_EXPORTER and OTEL_METRICS_EXPORTER to disable export
   ([#1439](https://github.com/open-telemetry/opentelemetry-specification/pull/1439))
-- Add `ForceFlush` to [Tracing SDK](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#shutdown) ([#1452](https://github.com/open-telemetry/opentelemetry-specification/pull/1452))
+- Add [`ForceFlush`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#forceflush) to SDK's `TracerProvider` ([#1452](https://github.com/open-telemetry/opentelemetry-specification/pull/1452))
 
 ## v1.0.1 (2021-02-11)
 
--- a/specification/trace/sdk.md
+++ b/specification/trace/sdk.md
@@ -70,7 +70,7 @@ implemented as a blocking API or an asynchronous API which notifies the caller
 via a callback or an event. OpenTelemetry client authors can decide if they want to
 make the flush timeout configurable.
 
-`ForceFlush` MUST invoke `ForceFlush` on all `SpanProcessors` of the `TracerProvider`.
+`ForceFlush` MUST invoke `ForceFlush` on all registered `SpanProcessors`.
 
 ## Additional Span Interfaces
 
--- a/specification/trace/sdk.md
+++ b/specification/trace/sdk.md
@@ -70,7 +70,7 @@ implemented as a blocking API or an asynchronous API which notifies the caller
 via a callback or an event. OpenTelemetry client authors can decide if they want to
 make the flush timeout configurable.
 
-`ForceFlush` MUST be implemented at least by invoking `ForceFlush` within all internal processors.
+`ForceFlush` MUST invoke `ForceFlush` on all `SpanProcessors` of the `TracerProvider`.
 
 ## Additional Span Interfaces
 
--- a/specification/trace/sdk.md
+++ b/specification/trace/sdk.md
@@ -58,6 +58,20 @@ make the shutdown timeout configurable.
 
 `Shutdown` MUST be implemented at least by invoking `Shutdown` within all internal processors.
 
+### ForceFlush
+
+This method provides a way for provider to immediately export all spans that have not yet been exported for all the internal processors.
+
+`ForceFlush` SHOULD provide a way to let the caller know whether it succeeded,
+failed or timed out.
+
+`ForceFlush` SHOULD complete or abort within some timeout. `ForceFlush` can be
+implemented as a blocking API or an asynchronous API which notifies the caller
+via a callback or an event. OpenTelemetry client authors can decide if they want to
+make the flush timeout configurable.
+
+`ForceFlush` MUST be implemented at least by invoking `ForceFlush` within all internal processors.
+
 ## Additional Span Interfaces
 
 The [API-level definition for Span's interface](api.md#span-operations)
--- a/spec-compliance-matrix.md
+++ b/spec-compliance-matrix.md
@@ -16,6 +16,7 @@ status of the feature is not known.
 | Get a Tracer                                                                                     |          | +  | +    | +  | +      | +    | +      | +   | +    | +   | +    | +     |
 | Safe for concurrent calls                                                                        |          | +  | +    | +  | +      | +    | +      | +   | +    | +   | +    | +     |
 | Shutdown (SDK only required)                                                                     |          |    | +    | +  | +      | +    | -      |     | +    | +   | +    | +     |
+| ForceFlush (SDK only required)                                                                   |          |    | +    | -  | -      | -    | -      |     | -    | -   | +    | -     |
 | [Trace / Context interaction](specification/trace/api.md#context-interaction)                    |          |    |      |    |        |      |        |     |      |     |      |       |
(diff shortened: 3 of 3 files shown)


Oberon00 (approved):
Thank you!

iNikem (approved):

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ release.
 
 - Adds `none` as a possible value for OTEL_TRACES_EXPORTER and OTEL_METRICS_EXPORTER to disable export
   ([#1439](https://github.com/open-telemetry/opentelemetry-specification/pull/1439))
+- Add `ForceFlush` to [Tracing SDK](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#shutdown) ([#1452](https://github.com/open-telemetry/opentelemetry-specification/pull/1452))
```
Oberon00, line 12:
```suggestion
- Add `ForceFlush` to [Tracing SDK](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#forceflush) ([#1452](https://github.com/open-telemetry/opentelemetry-specification/pull/1452))
```

anuraaga (approved):

specification/trace/sdk.md
```diff
@@ -58,6 +58,20 @@ make the shutdown timeout configurable.
...
+
+`ForceFlush` SHOULD provide a way to let the caller know whether it succeeded,
+failed or timed out.
+
+`ForceFlush` SHOULD complete or abort within some timeout. `ForceFlush` can be
+implemented as a blocking API or an asynchronous API which notifies the caller
+via a callback or an event. OpenTelemetry client authors can decide if they want to
+make the flush timeout configurable.
+
+`ForceFlush` MUST be implemented at least by invoking `ForceFlush` within all internal processors.
```
bogdandrutu, line 73: What else can you do? The "at least" sounds that you can do more

Oberon00: Nothing comes to mind, but is this different from shutdown?

Oberon00:
```suggestion
`ForceFlush` MUST invoke `ForceFlush` on all `SpanProcessors` of the `TracerProvider`.
```

Might be a better formulation (and could also be used for Shutdown above)

carlosalberto (approved):

carlosalberto:
Overall good, but I suggest we wait for next Tuesday to discuss it (given the historical fear of adding this kind of functionality).

specification/trace/sdk.md
kipwoker: @bogdandrutu if you agree with @Oberon00 suggestion, I will commit it, otherwise feel free to suggest another sentence.

yurishkuro:
Observation: despite being in the SDK spec, this PR proposes a change to TraceProvider, which is a part of the API, and it is a breaking change (adding methods to interface breaks existing implementations of that interface). While it is fine to add new methods to concrete SDK classes implementing TraceProvider, this change is very easy to misinterpret. Also, it is not clear how the end user is supposed to interact with this functionality - the flush won't be available at the API level.

Oberon00:
> Observation: despite being in the SDK spec, this PR proposes a change to TraceProvider, which is a part of the AP

No, it does not. It proposes to add this to the SDK implementation class of the TracerProvider. The same applies to Shutdown.

reyang (approved):
Thank you!

bogdandrutu:
> Observation: despite being in the SDK spec, this PR proposes a change to TraceProvider, which is a part of the API

This is a public api for the SDK implementation of the TraceProvider, but it is not an "interface" and even if it is we do not expect others to implement the SDK interfaces that we have (only for the API we guarantee that).

specification/trace/sdk.md
```diff
@@ -58,6 +58,20 @@ make the shutdown timeout configurable.
...
+
+`ForceFlush` SHOULD provide a way to let the caller know whether it succeeded,
+failed or timed out.
+
+`ForceFlush` SHOULD complete or abort within some timeout. `ForceFlush` can be
+implemented as a blocking API or an asynchronous API which notifies the caller
+via a callback or an event. OpenTelemetry client authors can decide if they want to
+make the flush timeout configurable.
+
+`ForceFlush` MUST invoke `ForceFlush` on all `SpanProcessors` of the `TracerProvider`.
```
bogdandrutu, line 73:
```suggestion
`ForceFlush` MUST invoke `ForceFlush` on all registered `SpanProcessors`.
```

jsuereth (approved):
+1.  We already recommend ForceFlush for ephemeral situations like CloudRun.

cijothomas (approved):

bogdandrutu (approved):
@yurishkuro do you still have a concern?

yurishkuro (approved):

SergeyKanzhelev (approved):

specification/trace/sdk.md
```diff
@@ -58,6 +58,20 @@ make the shutdown timeout configurable.
...
+
+`ForceFlush` SHOULD provide a way to let the caller know whether it succeeded,
+failed or timed out.
+
+`ForceFlush` SHOULD complete or abort within some timeout. `ForceFlush` can be
+implemented as a blocking API or an asynchronous API which notifies the caller
+via a callback or an event. OpenTelemetry client authors can decide if they want to
+make the flush timeout configurable.
+
+`ForceFlush` MUST invoke `ForceFlush` on all registered `SpanProcessors`.
```
SergeyKanzhelev, line 73: what about exporters?

reyang: Exporters are taken care of by the processors.

ThomsonTan: nit: change `SpanProcessors` to `SpanProcessor`s or just `SpanProcessor` because `SpanProcessors` is not a real class/object name.

SergeyKanzhelev: my question was whether we want to define this method on [exporter](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#span-exporter) as well as it is defined on processors. Since the retry logic is expected to be implemented on exporter, it may require a signal when `ForceFlush` is requested.

bogdandrutu: Let's keep that as a separate PR :)

Oberon00: I was completely surprised now when I realized exporters don't have this spec'd. It seems like a defect to me. Java has it implemented by the way https://github.com/open-telemetry/opentelemetry-java/blob/v0.17.1/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/SpanExporter.java#L69-L77.

SergeyKanzhelev: @Oberon00 if not the wording of retry that has to be handled by the exporter, it can be implemented stateless without the need for `Flush`.  Thank you for creating a tracking bug

bogdandrutu: @SergeyKanzhelev suggest to file an issue for that.

tigrannajaryan:
Just to make it clear: interfaces that the SDK must implement (such as `TraceProvider` interface) are NOT part of our compatibility guarantees from the perspective of the _implementor_ of the SDK. The interface may change in such a way that in turn requires the SDK code to change. Our guarantees say that the interface cannot change in a way that the requires end-user code to change.

Interfaces that define the interaction between the API and SDK can be modified by us. There is no expectation that the end user implements a `TraceProvider` and then expects that the `TraceProvider` interface remains stable in such a way that their implementation will not be broken over time.

Note 1: we have a requirement that newer versions of the SDK package must work correctly with older versions of the API packages, so we cannot do arbitrary changes to the interfaces between API and SDK. From the perspective of the API the callable interface of the SDK is guaranteed to be backwards compatible. 

Likewise SDK's interfaces that are intended to be called by the end user have the exact same requirement: the may evolve in a way that remains callable with the existing end-user code without changing anything in the end-user code.

Adding `ForceFlush` satisfies all above requirements. It is part of `TraceProvider` which is callable by the end user or by the API package, but it is not intended to be implemented by the end user.

Note 2: interfaces that are intended to be implemented by the end user (for example `Sampler`) have a different, stricter requirement. We must not change them in a way that breaks existing end user code that implements the interfaces. Such interfaces are considered to be part of our contract with the end user and are an integral part of our API.

Oberon00:
> Note 2: interfaces that are intended to be implemented by the end user (for example Sampler) have a different, stricter requirement. We must not change them in a way that breaks existing end user code that implements the interfaces. Such interfaces are considered to be part of our contract with the end user and are an integral part of our API.

@bogdandrutu and I wrote a few words about that on #1454. Since comments on a merged PR are hard to notice but I think this is an important topic, I will open a new issue for that.

Oberon00:
@tigrannajaryan also note that this PR does not even modify the `TracerProvider` interface, it modifies the `SdkTracerProvider` implementation class.

PR #1453, opened 2021-02-20
Add additional env variables for Jaeger exporter

lonewolf3739:
## Changes

This change adds two additional env variables for Jaeger exporter `OTEL_EXPORTER_JAEGER_CERTIFICATE` for gRPC and `OTEL_EXPORTER_JAEGER_PROTOCOL` for specifying protocol to use.

### DIFF
--- a/specification/sdk-environment-variables.md
+++ b/specification/sdk-environment-variables.md
@@ -109,7 +109,7 @@ Should only be used for a secure connection.
 
 - `OTEL_EXPORTER_JAEGER_PROTOCOL`
 
-Jaeger exporter has more than one transport protocol it can support, e.g. grpc, http/thrift, udp/thrift.
+Defines protocol to be used by the Jaeger exporter: `grpc`, `http/thrift`, `udp/thrift`.
 
 As of 1.0 of the specification, there
 *is no specified default, or configuration via environment variables*.
--- a/specification/sdk-environment-variables.md
+++ b/specification/sdk-environment-variables.md
@@ -99,6 +99,21 @@ See [OpenTelemetry Protocol Exporter Configuration Options](./protocol/exporter.
 | OTEL_EXPORTER_JAEGER_USER       | Username to be used for HTTP basic authentication | -                                                                                                |
 | OTEL_EXPORTER_JAEGER_PASSWORD   | Password to be used for HTTP basic authentication | -                                                                                                |
 
+Addtionally, the following environment variables are reserved for future
+usage in Jaeger Exporter configuration:
+
+- `OTEL_EXPORTER_JAEGER_CERTIFICATE`
+
+This will be used to specify path to certificate file for TLS credentials of gRPC client.
+Should only be used for a secure connection.
+
+- `OTEL_EXPORTER_JAEGER_PROTOCOL`
+
+Jaeger exporter has more than one transport protocol it can support, e.g. grpc, http/thrift, udp/thrift.
+
+As of 1.0 of the specification, there
+*is no specified default, or configuration via environment variables*.
+
 ## Zipkin Exporter
 
 **Status**: [Stable](document-status.md)


specification/sdk-environment-variables.md
```diff
@@ -99,6 +99,21 @@ See [OpenTelemetry Protocol Exporter Configuration Options](./protocol/exporter.
 | OTEL_EXPORTER_JAEGER_USER       | Username to be used for HTTP basic authentication | -                                                                                                |
 | OTEL_EXPORTER_JAEGER_PASSWORD   | Password to be used for HTTP basic authentication | -                                                                                                |
 
+Addtionally, the following environment variables are reserved for future
+usage in Jaeger Exporter configuration:
+
+- `OTEL_EXPORTER_JAEGER_CERTIFICATE`
```
yurishkuro, line 105: how about `OTEL_EXPORTER_JAEGER_CLIENT_CERT_FILE`? It's almost self-explanatory.

But even that seems not sufficient. For example, in the Jaeger C# client there are 3 env vars for TLS: https://github.com/jaegertracing/jaeger-client-csharp#configuration-via-environment

anuraaga: For reference I had tried a pr for the three variables but couldn't get it in for 1.0 so need to rework it in a backwards compatible way.

I would go with just `CERTIFICATE` here to match otlp.

specification/sdk-environment-variables.md
```diff
@@ -99,6 +99,21 @@ See [OpenTelemetry Protocol Exporter Configuration Options](./protocol/exporter.
...
+
+This will be used to specify path to certificate file for TLS credentials of gRPC client.
+Should only be used for a secure connection.
+
+- `OTEL_EXPORTER_JAEGER_PROTOCOL`
+
+Jaeger exporter has more than one transport protocol it can support, e.g. grpc, http/thrift, udp/thrift.
```
anuraaga, line 112: Since the exporters are vastly different, would it make more sense to have different values for exportet, e.g. jaeger, jaeger_http, jaeger_udp?

specification/sdk-environment-variables.md
lonewolf3739: As I understand it the private key and certificate chain pair is optional and can be null even for the secure connection. What do you think about `OTEL_EXPORTER_JAEGER_CERTIFICATE`, `OTEL_EXPORTER_JAEGER_CLIENT_KEY` and `OTEL_EXPORTER_JAEGER_CLIENT_CHAIN`?

specification/sdk-environment-variables.md
lonewolf3739: I tried to follow along the similar lines of OTLP(grpc, http/protobuf, http/json), Zipkin for consistency. And also `jaeger` for Jaeger gRPC is misleading/not aligning with others.

specification/sdk-environment-variables.md
yurishkuro: Do we have any SDK that already implements this auth mode and supports various certs?  I would model the names after the APIs used (although different languages may call them differently).

BTW, I would recommend `OTEL_EXPORTER_JAEGER_TLS_` prefix for all three, it provides a clear scope.

specification/sdk-environment-variables.md
yurishkuro, line 112:
```suggestion
Defines protocol to be used by the Jaeger exporter: `grpc`, `http/thrift`, `udp/thrift`.
```

specification/sdk-environment-variables.md
lonewolf3739: Looks like Java provides way to set the ChannelCreds but doesn't have the support for loading from env. Python [implementation](https://github.com/open-telemetry/opentelemetry-python/blob/main/exporter/opentelemetry-exporter-jaeger/src/opentelemetry/exporter/jaeger/util.py#L52) uses `OTEL_EXPORTER_JAEGER_CERTIFICATE`. I couldn't find any other language SDK supporting them yet.

anuraaga: @yurishkuro I don't think we should deviate from what we have for otlp since the protocols are so similar. I tried to get the tls prefix in before 1.0 but it didn't happen - I would say at this point consistency would be simpler. All grpc transports follow the same pattern.

specification/sdk-environment-variables.md
anuraaga: I see - it's a bit difficult because we have udp here, unlike the others. So any variables that relate to http specifically just can't apply, for example http headers (we have for otlp and still need to add for the others) and even tls I think. Would leaving udp to be separate make sense?

lonewolf3739: Not sure If I understood you correctly but if headers are not applicable then they will not be set and I don't think tls applies to udp. What is the issue with udp not being separate?

anuraaga: Yeah ignoring unsupported values is another approach too, it's mostly just to compare them. I don't think we currently have many variables that have a dependency on other ones, e.g. var2 is only read if var1 has a certain value. The separation could be clearer or it could be less clear, not too sure.

Issue #622, opened 2020-05-26
Align Embedded Logs data model with Standalone Logs data model


tigrannajaryan:
Now that we have an official standalone logs data model it will be beneficial to see if we can align logs embedded in the Spans (Span events) with this data model.

Embedded spans should use the same or similar model except the following which is already recorded in the containing Span:
- Trace context fields: trace id, span id, trace flags.
- Resource.

The differences currently are:
Span Event|Standalone Log Record|Notes
-----------|-----------------------|---------
Name|ShortName|Equivalent semantic, only field names are different.
-|Severity|Missing in Span Event.
-|Body|Missing in Span Event.
Attributes|Attributes|Almost equivalent, but Standalone Logs allow complex nested values that Span Events don't.

We need to decide which of these differences are acceptable/necessary. My preliminary thoughts are the following:
- Rename `Name` to `ShortName`. There is no semantics change.
- There is no need to add `Body` to Span Events. For standalone logs the `Body` is necessary to record what cannot be represented in the attributes or other fields for log formats that have unusual/unknown requirements. There are no such requirements for Span Events because we know they are generated (using OpenTelemetry API).
- We should extend Span Event `Attributes` to allow the same nested values. This [was requested/discussed](https://github.com/open-telemetry/opentelemetry-specification/issues/376) in the past too.
- I am torn on `Severity`. We should use it if we believe the [Error Semantics](https://github.com/open-telemetry/oteps/blob/master/text/0097-log-data-model.md) it introduces are applicable to Span Events.

Oberon00:
* Body can be a semantic convention for attributes.
* Severity can be a semantic convention for attributes
* ShortName could be renamed to Name.
* Why are nested attributes needed for logs? Is this not cases where you could do JSON in / JSON out?

tigrannajaryan:
>ShortName could be renamed to Name.

I would like to know what others think. This seems reasonable to me.

tigrannajaryan:
>Body can be a semantic convention for attributes.

Yes, it can. The question is: do we need a Body for Span Events? Our API does not have anything like that and I do not think we need to add it to the API (but perhaps there is a good use case that I am not aware of).

tigrannajaryan:
>Severity can be a semantic convention for attributes

I agree, it can. However, I think there is a value in consistency and uniformness. Making it a field in Log Data model and a semantic convention in Span Event is somewhat inconsistent.

tigrannajaryan:
[...]

PR #1461, opened 2021-02-23
Clarify that Events and Log Records are conceptually the same thing

tigrannajaryan:
I get this question a lot. This clarifies that events and log records
are different names of the same concept (from Otel perspective).

This was discussed verbally many times and I believe it was also
in my initial logs proposal but got lost somewhere in the process.

This also fixes the TBD link in the spec that was supposed to point
to this explanation.


yurishkuro (commented):
Hm, not sure if this is a fair comparison. 

First, you're quoting a definition of a "log file", not a "log". A log is just a more abstract notion of a log file, i.e. a Kafka topic's partition is also a "log".

Second, log != event because of difference in cardinality. More accurately, log is a sequence of events.

Finally, I think it is worth addressing the difference in meaning between "logs" as in "metrics, traces, and logs", and log as an abstract notion of sequentially written events. I don't think OTEL is aiming to deal with the latter, we're mostly talking about _application logs_, which some people differentiate from business events, and change management events. In other words, we are talking about logs as a telemetry signal, not logs as a transport mechanism.

tigrannajaryan:
@yurishkuro you are right, if we look into the more nuanced definition of the log vs log record vs event. 

However, the confusion I often see is that some people think that events are an entirely new concept that somehow requires a separate treatment in OpenTelemetry. I am trying to reduce this particular confusion.

I updated the wording to make it clearer that I am equating Log Records and Events, which is a more fair comparison. Note: I am using Events in the sense that Wikipedia uses them in the quote I added in the PR or [Windows Events system](https://en.wikipedia.org/wiki/Event_Viewer) use them. They are nothing but log records from this perspective (also known as "events", "event logs", and sometimes plainly "logs").

Unfortunately Wikipedia has no definition of Log, so I had to link to Log File as he best approximation, since it serves my purpose anyway because it clearly talks about events.

yurishkuro (approved):

carlosalberto (approved):

kumoroku (approved):

reyang:
We need to clarify if `Span.AddEvent` is a logging API or not (I believe it is not).

jmacd (approved):

jmacd:
> We need to clarify if Span.AddEvent is a logging API or not (I believe it is not).

I read the text of this PR to say that `AddEvent` is a logging API.

iNikem (approved):

tigrannajaryan:
According to glossary.md:
>**Embedded Log**
>Log Records embedded inside a Span object, in the Events list.

This definition probably needs to be refined. But semantically it definitely is a log record. At one point I wanted to make sure the data models are aligned but it did not get enough traction at the time: https://github.com/open-telemetry/opentelemetry-specification/issues/622

kenfinnigan:
I may be misunderstanding the intent, but I see logs being used today as being different from an event both in their intent and their data modeling.

Having a log, as an example, with "ERROR", a timestamp, and an arbitrary message is very different from a structured event, that could happen to be written to a log.

Where does OpenTelemetry believe "logs" fit in terms of https://charity.wtf/2019/02/05/logs-vs-structured-events/ ?

tigrannajaryan:
> I may be misunderstanding the intent, but I see logs being used today as being different from an event both in their intent and their data modeling.
> 
> Having a log, as an example, with "ERROR", a timestamp, and an arbitrary message is very different from a structured event, that could happen to be written to a log.

It is not clear to me what is "very different" from OpenTelemetry perspective. Both are possible to represent and process using existing Otel log data model. 

> Where does OpenTelemetry believe "logs" fit in terms of https://charity.wtf/2019/02/05/logs-vs-structured-events/ ?

This PR attempts to describe what OpenTelemetry believes. Any real or mythical differences between "logs" and "structured events" don't matter for OpenTelemetry. We are not in the business that requires differentiating the two.

reyang (approved):
LGTM.

PR #1467, opened 2021-02-25
Add SpanExporter.ForceFlush.

### LINKED ISSUE #1454: ForceFlush missing from Exporter spec
We currently have ForceFlush only specified in the SpanProcessor. #1452 adds the missing "helper" method to call this to the SdkTracerProvider. But we are also missing ForceFlush in the spec for exporters (some languages, e.g. Java, have it implemented already (https://github.com/open-telemetry/opentelemetry-java/blob/v0.17.1/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/SpanExporter.java#L69-L77).

In fact this was proposed in the original issue to add flush/forceFlush (#351) which was titled "Add flush API to processor/**exporter**". However it seems to have been glanced over in #370 (in turn, then some SIGs seemed to have glanced over the fact that it was never spec'd and implemented it anyway 😉).

In some languages, adding a new method to an interface might be a breaking change for existing implementers. However, there are simple workarounds for that such as adding Flush to an new interface like `FlushableExporter extends Exporter`. I think it is OK to leave this implementation detail to the language SIGs.

Oberon00:
Fixes #1454.

## Changes

Adds `ForceFlush` (which already existed in SdkTracerProvider and SpanProcessor) to SpanExporter, and adds wording that SpanProcessor.ForceFlush must call that new SpanExporter.ForceFlush.


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ release.
 - Adds `none` as a possible value for OTEL_TRACES_EXPORTER and OTEL_METRICS_EXPORTER to disable export
   ([#1439](https://github.com/open-telemetry/opentelemetry-specification/pull/1439))
 - Add [`ForceFlush`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#forceflush) to SDK's `TracerProvider` ([#1452](https://github.com/open-telemetry/opentelemetry-specification/pull/1452))
+- Add `ForceFlush` to `Span Exporter` interface [#1467](https://github.com/open-telemetry/opentelemetry-specification/pull/1467).
```
arminru, line 13:
```suggestion
- Add `ForceFlush` to `Span Exporter` interface ([#1467](https://github.com/open-telemetry/opentelemetry-specification/pull/1467))
```

specification/trace/sdk.md
```diff
@@ -582,6 +588,24 @@ return a `Failure` result.
...
+This is a hint to ensure that any `Spans` passed to the exporter prior to the
+call to `ForceFlush` SHOULD be completed as soon as possible, preferably before
+returning from this method.
+
+`ForceFlush` SHOULD provide a way to let the caller know whether it succeeded,
+failed or timed out.
+
+`ForceFlush` SHOULD only be called in cases where it is absolutely necessary,
+such as when using some FaaS providers that may suspend the process after an
+invocation, but before the `Processor` exports the completed spans.
```
arminru, line 602:
```suggestion
invocation, but before the exporter exports the completed spans.
```

specification/trace/sdk.md
```diff
@@ -582,6 +588,24 @@ return a `Failure` result.
 and the destination is unavailable). OpenTelemetry client authors can decide if they
 want to make the shutdown timeout configurable.
 
+#### `ForceFlush()`
+
+This is a hint to ensure that any `Spans` passed to the exporter prior to the
+call to `ForceFlush` SHOULD be completed as soon as possible, preferably before
```
arminru, line 594:
```suggestion
call to `ForceFlush` SHOULD be exported as soon as possible, preferably before
```
to avoid confusion with `end`ing spans

specification/trace/sdk.md
```diff
@@ -582,6 +588,24 @@ return a `Failure` result.
...
+returning from this method.
+
+`ForceFlush` SHOULD provide a way to let the caller know whether it succeeded,
+failed or timed out.
+
+`ForceFlush` SHOULD only be called in cases where it is absolutely necessary,
+such as when using some FaaS providers that may suspend the process after an
+invocation, but before the `Processor` exports the completed spans.
+
+`ForceFlush` SHOULD complete or abort within some timeout. `ForceFlush` can be
```
arminru, line 604:
```suggestion
`ForceFlush` SHOULD complete or abort within some timeout. `ForceFlush` MAY be
```

specification/trace/sdk.md
```diff
@@ -467,14 +467,20 @@ make the shutdown timeout configurable.
...
 
 `ForceFlush` SHOULD provide a way to let the caller know whether it succeeded,
 failed or timed out.
 
 `ForceFlush` SHOULD only be called in cases where it is absolutely necessary,
 such as when using some FaaS providers that may suspend the process after an
-invocation, but before the `Processor` exports the completed spans.
+invocation, but before the `SpanProcessor` exports the completed spans.
 
 `ForceFlush` SHOULD complete or abort within some timeout. `ForceFlush` can be
```
arminru, line 485:
```suggestion
`ForceFlush` SHOULD complete or abort within some timeout. `ForceFlush` MAY be
```

Oberon00: I think this should be done for all the ForceFlush definitions and also Shutdown at once, it would be confusing if TracerProvider/SpanProcessor/Exporter differed only there.

specification/trace/sdk.md
```diff
@@ -582,6 +588,24 @@ return a `Failure` result.
 and the destination is unavailable). OpenTelemetry client authors can decide if they
 want to make the shutdown timeout configurable.
 
+#### `ForceFlush()`
+
+This is a hint to ensure that any `Spans` passed to the exporter prior to the
```
Oberon00, line 593:
```suggestion
This is a hint to ensure that the export of any `Spans` passed to the exporter prior to the
```

specification/trace/sdk.md
Oberon00: I will do this instead: https://github.com/open-telemetry/opentelemetry-specification/pull/1467/files#r583035984

specification/trace/sdk.md
Oberon00: Same as https://github.com/open-telemetry/opentelemetry-specification/pull/1467/files#r583035024

carlosalberto:
Overall LGTM, although this is precisely one of the interfaces that are exposed to the users and needs special care and some thought on whether some kind of 'default method' can be offered in the different SIGs.

cc @bogdandrutu

specification/trace/sdk.md
```diff
@@ -582,6 +588,24 @@ return a `Failure` result.
 and the destination is unavailable). OpenTelemetry client authors can decide if they
 want to make the shutdown timeout configurable.
 
+#### `ForceFlush()`
```
bogdandrutu, line 591: I think we can say, in case of stable releases a "no-op" implementation can be offered as a default. I think that is the best we can do :)

bogdandrutu (approved):

specification/trace/sdk.md
```diff
@@ -467,14 +467,20 @@ make the shutdown timeout configurable.
 
 #### ForceFlush()
 
-Exports all spans that have not yet been exported to the configured `Exporter`.
+This is a hint to ensure that any tasks associated with `Spans` for which the
+`SpanProcessor` had already received events prior to the call to `ForceFlush` SHOULD
+be completed as soon as possible, preferably before returning from this method.
+
+In particular, if the `SpanProcessor` has any associated exporters, it MUST call
+the exporter's `Export` with all spans for which this was not already done and
```
reyang, line 475: I think this shouldn't be "MUST". For example, if the processor has 1000 items, and it is calling the `exporter.Export` with a maximum size of 100 items, on the 200 items timeout limit is hit - the `processor.ForceFlush` should return as early as possible with an indication of timeout rather than push more items to the exporter.

reyang (changes requested):
I think timeout has higher priority, so the processor should not try to push more items to the exporter when timeout happened.

specification/trace/sdk.md
```diff
@@ -582,6 +588,24 @@ return a `Failure` result.
 and the destination is unavailable). OpenTelemetry client authors can decide if they
 want to make the shutdown timeout configurable.
 
+#### `ForceFlush()`
+
+This is a hint to ensure that the export of any `Spans` passed to the exporter prior to the
```
reyang, line 593: This is a bit vague, if another thread is calling `exporter.Export(batch)`, what do we expect `exporter.ForceFlush` to do?

Oberon00: This is up to the exporter I think. But if Export and ForceFlush are called concurrently, then by definition we have a race condition, even if the Exporter does locking internally, so we cannot possibly guarantee that any Spans in that Export call are exported or not. The most we could offer would be a "all or none Spans of each concurrent Export call are affected" gurantee, but I don't think that would be helpful.

specification/trace/sdk.md
Oberon00: Interesting point, we could add wording on the timeout here. But I don't know how close we want to specify that. For example: It might make sense to heuristically determine in the SpanProcessor "Oh no, we are coming close to the timeout, we took average x seconds per batch but still have y batches left and only z seconds of the timeout. Better call exporter.ForceFlush now and only then Export more spans if time is still left".

Probably that wouldn't make sense though because if Export takes so long that we run into the timeout, ForceFlush would maybe be a no-op -- but it might also not be, e.g. `write` for a filebased exporter can take long, and still needs a `flush` at the end.

Specifying the details is complicated here, so maybe we should say:

> `ForceFlush` MUST be implemented such that these conditions are fulfilled:
>
> * If there was no timeout, `ForceFlush`MUST have been called after all spans were passed to `Export` of the Exporter.
> * If there was a timeout, at least one `Export` or `ForceFlush` operation was attempted.
> * ForceFlush MAY be called more than once if at least one Export call is in-between. Note: SpanProcessors might want to heuristically determine that they could run into a timeout and in response call `ForceFlush` before finishing export.
>
> If `ForceFlush` does not support a timeout, or an infinite timeout was given, then a stricter condition is required: The operation MUST call the exporter's `Export` with all spans for which this was not already done and invoke ForceFlush on the export afterwards exactly once.

Oberon00: We could additionally specify that the timeout expressed as a deadline timepoint SHOULD be passed on to the exporter, i.e. if I call processor.ForceFlush(timeout=5) at T=2, then the deadline is D = T + timeout = 7, so the first export call may get timeout=5, and if it took 1, then the next export call invoked at T=3 will get timeout=4 (= D - T).

reyang: The "most permissive but maybe still useful spec" looks reasonable to me 👍

specification/trace/sdk.md
reyang: I guess we could change it from "any `Spans` passed to the exporter" to "any `Spans` that the exporter has received"?

specification/trace/sdk.md
bogdandrutu: > I think this shouldn't be "MUST". For example, if the processor has 1000 items, and it is calling the exporter.Export with a maximum size of 100 items, on the 200 items timeout limit is hit - the processor.ForceFlush should return as early as possible with an indication of timeout rather than push more items to the exporter.

I feel we are too picky, and try to over specify things. Maybe as simple as "In particular, if the `SpanProcessor` has any associated exporters, it MUST call the exporter's `Export` with all spans for which this was not already done and then invoke `ForceFlush` on it. " If a timeout is given then the operation should return when deadline is hit.

If users calls with such a small timeout they get half of the work done. We should not try to ensure at least one export and one flush is called. If timeout is 1us then maybe only an export can be called, etc. So just explain the order of execution and describe that timeout should be interrupt this.

bogdandrutu:
> I think timeout has higher priority, so the processor should not try to push more items to the exporter when timeout happened.

I agree with @reyang here, we should should return when the deadline happens and not try to over-engineer the logic, if too short you get done what was possible during the amount of time.

specification/trace/sdk.md
Oberon00: >  We should not try to ensure at least one export and one flush is called

I agree, that's why I suggested "or". But I also feel that my suggestion in the comment above is overly complicated.

specification/trace/sdk.md
Oberon00:
```suggestion
This is a hint to ensure that the export of any `Spans`  the exporter has received prior to the
```

specification/trace/sdk.md
Oberon00: I added 04e3e29.

reyang (approved):
LGTM. Thanks for driving this! 💯

specification/trace/sdk.md
```diff
@@ -582,6 +592,24 @@ return a `Failure` result.
...
+failed or timed out.
+
+`ForceFlush` SHOULD only be called in cases where it is absolutely necessary,
+such as when using some FaaS providers that may suspend the process after an
+invocation, but before the exporter exports the completed spans.
+
+`ForceFlush` SHOULD complete or abort within some timeout. `ForceFlush` can be
+implemented as a blocking API or an asynchronous API which notifies the caller
+via a callback or an event. OpenTelemetry client authors can decide if they want to
+make the flush timeout configurable.
```
ThomsonTan, line 611: For blocking API, I think it is very hard to abort within some timeout. Like the implementation complies to the spec if it just does abort its entry point without doing any work. Should we say `ForceFlush` should return (complete or abort) as soon as possible if the given `timeout` is reached?

bogdandrutu:
@Oberon00 please rebase :) that will also remove the stale

jsuereth (approved):

bogdandrutu:
Please rebase

specification/trace/sdk.md
Oberon00: This is copy & past from shutdown and the other ForceFlush. Please open an issue if you want tor reword this (should be consistent everywhere).

bogdandrutu:
@arminru can you rebase & review & and merge if you are ok with this change :)

specification/trace/sdk.md
```diff
@@ -473,14 +473,24 @@ make the shutdown timeout configurable.
...
 
-Exports all spans that have not yet been exported to the configured `Exporter`.
+This is a hint to ensure that any tasks associated with `Spans` for which the
+`SpanProcessor` had already received events prior to the call to `ForceFlush` SHOULD
+be completed as soon as possible, preferably before returning from this method.
+
+In particular, if any `SpanProcessor` has any associated exporter, it SHOULD
+try to call the exporter's `Export` with all spans for which this was not
+already done and then invoke `ForceFlush` on it.
+The [built-in SpanProcessors](#built-in-span-processors) MUST do so.
```
arminru, line 483: @Oberon00 Please add an entry in the compliance matrix for that, otherwise it might be overlooked easily.

arminru (approved):

spec-compliance-matrix.md
```diff
@@ -72,6 +72,7 @@ status of the feature is not known.
 | [New Span ID created also for non-recording Spans](specification/trace/sdk.md#sdk-span-creation) |          | +  | +    |    | +      | +    | +      |     |      |     | -    | +     |
 | [IdGenerators](specification/trace/sdk.md#id-generators)                                         |          |    | +    |    |        | +    |        |     |      |     |      | +     |
 | [SpanLimits](specification/trace/sdk.md#span-limits)                                             | X        |    | +    |    |        | +    |        |     |      |     |      | +     |
+| [SpanProcessor implements ForceFlushes spec](specification/trace/sdk.md#force-flush-1)                                             |          |    |      |    |        |      |        |     |      |     |      |       |
```
arminru:
```suggestion
| [Built-in `SpanProcessor`s implement `ForceFlush` spec](specification/trace/sdk.md#forceflush-1)                                             |          |    |      |    |        |      |        |     |      |     |      |       |
```

spec-compliance-matrix.md
```diff
@@ -142,6 +143,8 @@ Note: Support for environment variables is optional.
 
 | Feature                                               | Optional | Go | Java                                                                  | JS | Python                                                                  | Ruby | Erlang | PHP | Rust | C++ | .Net | Swift |
 |-------------------------------------------------------|----------|----|-----------------------------------------------------------------------|----|-------------------------------------------------------------------------|------|--------|-----|------|-----|------|-------|
+| [Exporter interface](specification/trace/sdk.md#span-exporter) |          |    | + |    | +                                                                       |      |        |     |      |     | +    |       |
+| [Exporter interface has `ForceFlush`](specification/trace/sdk.md#forceflush-1) |          |    | + |    |                                                                         |      |        |     |      |     |      |       |
```
arminru:
```suggestion
| [Exporter interface has `ForceFlush`](specification/trace/sdk.md#forceflush-2) |          |    | + |    |                                                                         |      |        |     |      |     |      |       |
```

spec-compliance-matrix.md
Oberon00: Outdated, will add the name change.

spec-compliance-matrix.md
```diff
@@ -72,6 +72,7 @@ status of the feature is not known.
 | [New Span ID created also for non-recording Spans](specification/trace/sdk.md#sdk-span-creation) |          | +  | +    |    | +      | +    | +      |     |      |     | -    | +     |
 | [IdGenerators](specification/trace/sdk.md#id-generators)                                         |          |    | +    |    |        | +    |        |     |      |     |      | +     |
 | [SpanLimits](specification/trace/sdk.md#span-limits)                                             | X        |    | +    |    |        | +    |        |     |      |     |      | +     |
+| [SpanProcessor implements ForceFlushes spec](specification/trace/sdk.md#forceflush-1)            |          |    |      |    |        |      |        |     |      |     |      |       |
```
Oberon00, line 75:
```suggestion
| [Built-in `SpanProcessor`s implement `ForceFlush` spec](specification/trace/sdk.md#forceflush-1)               |          |    |      |    |        |      |        |     |      |     |      |       |
```

Thank you @arminru

PR #1470, opened 2021-02-25
Clarify wording on TracerProvider -> Tracer -> SpanProcessor in ForceFlush

MrAlias:
Clarify the relationship between TracerProvider and SpanProcessors with regards to the ForceFlush method.


MrAlias:
This isn't correct. Please disregard.

Issue #1457, opened 2021-02-23
Interface stability for implementers


Oberon00:
We need to define our policy for adding new methods to existing interfaces *in the specification*, especially considering APIs that are intended to be implemented by users, namely "plugin interfaces" (Exporter, Sampler) and API implementation interfaces (Span, TracerProvider, ...). It is clear and already specified that at least for plugin interfaces we must stay ABI-compabible. However, there are several ways to achieve this (other than just never specifying new methods) in different languages (Java default interface implementations, C++ non-pure virtual methods, deriving a new interface in most languages come to mind).

----

Recently there were some discussions on adding `ForceFlush` to the SdkTracerProvider (#1451, #1452) and the Exporter (#1454). While it was agreed that the former is unproblematic (SdkTracerProvider being a class), the Exporter is what [our versioning spec](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.0.1/specification/versioning-and-stability.md) calls a [plugin interface](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.0.1/specification/versioning-and-stability.md#sdk-stability):

> #### SDK Stability
>
> Public portions of SDK packages MUST remain backwards compatible.
There are two categories of public features: **plugin interfaces** and **constructors**.
Examples of plugins include the SpanProcessor, Exporter, and Sampler interfaces.
Examples of constructors include configuration objects, environment variables, and SDK builders.
>
> Languages which ship binary artifacts SHOULD offer [ABI compatibility](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.0.1/specification/glossary.md#abi-compatibility) for SDK packages.

(I think this is more or less what @tigrannajaryan stated in https://github.com/open-telemetry/opentelemetry-specification/pull/1452#issuecomment-783730643)

In #1454 I remarked about adding `ForceFlush` to the existing `Exporter` interface:

> In some languages, adding a new method to an interface might be a breaking change for existing implementers. However, there are simple workarounds for that such as adding Flush to an new interface like FlushableExporter extends Exporter. I think it is OK to leave this implementation detail to the language SIGs.

and @bogdandrutu remarked in response:

> I think we need to add some text about this, languages should choose what is best for them that allows to extent the API (java8 has default implementation, go can rely o duci typing, etc.) that will allow them to extend the interfaces. And we should not consider breaking change adding a new functionality on an interface. Will file a separate issue.

[...]

PR #1473, opened 2021-02-26
Add method to allow setting multiple attributes at once to …

anuraaga:
…a Span.

## Changes

We've gotten use feedback that it would be convenient to have a way to set multiple attributes at once. This is particularly useful in languages like Java where because of the way generics work, something like `attributes.forEach(span::setAttribute)` can't be offered. It has consistency with many collection apis that offer singular / plural versions of mutators so seems like a reasonable API.

https://github.com/open-telemetry/opentelemetry-java/pull/2799


specification/trace/api.md
```diff
@@ -462,6 +462,11 @@ The Span interface MUST provide:
   as arguments. This MAY be called `SetAttribute`. To avoid extra allocations some
   implementations may offer a separate API for each of the possible value types.
 
+The Span interface MAY provide:
+
+- An API to set multiple `Attributes` at once, where all `Attributes` are passed as a
+  single argument. This MAY be called `SetAllAttributes`.
```
Oberon00, line 468:
```suggestion
- An API to set multiple `Attributes` at once, where the `Attributes` are passed as a
  in a single method call. This MAY be called `SetAllAttributes`.
```

Whether this would be a single (map) argument or multiple arguments (e.g. `**kwargs` in Python) is language dependent.

anuraaga: Yup thanks

iNikem (approved):

Oberon00:
This being a "MAY", I *think* it should be fine anyway, but this is another example where we should think of interface stability #1457. This is what we have in the spec:

> Major versions of the API MUST be supported for a minimum of three years after the release of the next major API version.

and

> A version of the SDK which supports the latest minor version of the last major version of the API will continue to be maintained during LTS.

so the question is now if this should be a major or minor version bump for the API. It probably is only minor because calls against earlier versions will continue to compile:

> Backward-incompatible changes to API packages MUST NOT be made unless the major version number is incremented. All existing API calls MUST continue to compile and function against all future minor versions of the same major version.

although you could argue that

> Major version bumps MUST occur when there is a breaking change to a stable interface

is also fulfilled in languages where there are no default interface implementations, although, as discussed in #1457, there are always ways to make it backwards compatible by using a new interface instead of modifying an existing one.

tigrannajaryan:
> 
> although you could argue that
> 
> > Major version bumps MUST occur when there is a breaking change to a stable interface
> 
> is also fulfilled in languages where there are no default interface implementations, although, as discussed in #1457, there are always ways to make it backwards compatible by using a new interface instead of modifying an existing one.

Given that `Span` interface is not intended to be implemented by the end user, I do not think addition of a method to the interface constitutes a breaking change.

yurishkuro:
> Given that `Span` interface is not intended to be implemented by the end user

I would disagree. If that was the case why bother having an API at all, instead of just SDK. This may not cause immediate problems since the API is too new, but it is theoretically possible that it's been already implemented somewhere.

Oberon00:
I just noticed we also have this in the spec under "Design goals" https://github.com/open-telemetry/opentelemetry-specification/blob/v1.0.1/specification/versioning-and-stability.md#design-goals:

> Ensure that application owners stay up to date with the latest release of the SDK. We want all users to stay up to date with the latest version of the OpenTelemetry SDK. We do not want to create hard breaks in support, of any kind, which leave users stranded on older versions. It MUST always be possible to upgrade to the latest minor version of the OpenTelemetry SDK, without creating compilation or runtime errors.

If I interpret this correctly, then newer SDKs must continue supporting older APIs? But I can't find a "normative" reference to that outside this "Design goals" section.

carlosalberto:
> This may not cause immediate problems since the API is too new, but it is theoretically possible that it's been already implemented somewhere.

Well, .Net was released and it doesn't have this functionality, but otherwise, yes ;)

And overall, I'd say user-interfaces (such as `SpanProcessor`) sit next to API when it comes to compatibility (as opposed to internals and relatively internal members of the SDK).

specification/trace/api.md
```diff
@@ -462,6 +462,11 @@ The Span interface MUST provide:
   as arguments. This MAY be called `SetAttribute`. To avoid extra allocations some
   implementations may offer a separate API for each of the possible value types.
 
+The Span interface MAY provide:
+
+- An API to set multiple `Attributes` at once, where the `Attributes` are passed in a
+  single method call. This method MAY be called `SetAllAttributes`.
```
reyang, line 468: I would vote for `SetAttributes` instead of `SetAllAttributes`.
`SetAllAttributes` gives a feeling that it will make previous calls invalid:
```text
span.SetAllAttributes(foo: 1, bar: 2);
span.SetAllAttributes(baz: 3);
// do we have foo, bar and baz, or we only have baz?
```

```text
span.SetAttributes(foo: 1, bar: 2);
span.SetAttributes(baz: 3);
// we know that we will have foo, bar and baz
```

reyang (changes requested):
Please see my comment [here](https://github.com/open-telemetry/opentelemetry-specification/pull/1473/files#r583872561)

specification/trace/api.md
anuraaga: Heh this is the obvious of Java, we would expect `SetAttributes` to clear existing ones and `SetAll` to keep them, similar to `putAll` in a Map. I can remove the naming recommendation if it seems too language specific?

iNikem:
> > Given that `Span` interface is not intended to be implemented by the end user
> 
> I would disagree. If that was the case why bother having an API at all, instead of just SDK. This may not cause immediate problems since the API is too new, but it is theoretically possible that it's been already implemented somewhere.

I find this line of thought very disturbing. I hear again and again that any changes, even additive ones, to our API are breaking and should not happen. Do we really believe that our API is perfect and should not be evolved? Or that any change must produce a new major version?

yurishkuro:
> I find this line of thought very disturbing. I hear again and again that any changes, even additive ones, to our API are breaking and should not happen. Do we really believe that our API is perfect and should not be evolved? Or that any change must produce a new major version?

The stdlib in Go language has evolved for 10yrs without breaking changes (and without major versions). I think it's fair to say the complexity of Go's stdlib is orders of magnitude higher that OTEL. Evolution of the APIs does not have to mean breaking users' code.

iNikem:
> > I find this line of thought very disturbing. I hear again and again that any changes, even additive ones, to our API are breaking and should not happen. Do we really believe that our API is perfect and should not be evolved? Or that any change must produce a new major version?
> 
> The stdlib in Go language has evolved for 10yrs without breaking changes (and without major versions). I think it's fair to say the complexity of Go's stdlib is orders of magnitude higher that OTEL. Evolution of the APIs does not have to mean breaking users' code.

I absolutely don't argue with that. But it seems there are different opinions about what is a breaking change. Adding a new method to the interface, which is not meant to be implemented by end users, is not a breaking change. See #1457 as well.

specification/trace/api.md
carlosalberto: @anuraaga Probably yes. We had in the past a (now forgotten) PR that tried to add this, and it was suggested to define the operation as a very general thing, and don't make the name a hard requirement (and mention languages can adapt it to their own guidelines, etc).

yurishkuro: `AddAttributes` is shorter and unambiguous

anuraaga: Sorry for the delay, went ahead and removed the naming recommendation.

> AddAttributes is shorter and unambiguous

I think it's a bit ambiguous, previous attributes (especially for list attributes) we could expect to remain in some form if it's just `Add`.

anuraaga:
@reyang Is your comment addressed now that the method name suggestion is removed?

anuraaga:
@open-telemetry/technical-committee any way to proceed with this PR? Thanks!

carlosalberto (approved):

carlosalberto:
I think @reyang request for changes has indeed been addressed, as we don't impose a formal name for the new operation. Also, we have now a clear path regarding compatibility (see https://github.com/open-telemetry/oteps/pull/147) so we should be fine to go ;)

PR #1477, opened 2021-02-26
Clarify when OpenTelemetry MUST provide a context implementation

rakyll:
Language implementations should review the OpenTelemetry context if there is an established context implementation in their ecosystem to see if OpenTelemetry requirements are satified by the existing context. If not, OpenTelemetry must provide a Context implementation.

I'm moving the language-specific note to the top and calling out that Context is an in-process context propagation mechanism.

### DIFF
--- a/specification/context/context.md
+++ b/specification/context/context.md
@@ -20,8 +20,8 @@ Table of Contents
 
 ## Overview
 
-A `Context` is an in-process propagation mechanism which carries 
-execution-scoped values across API boundaries and between logically associated 
+A `Context` is an in-process propagation mechanism which carries
+execution-scoped values across API boundaries and between logically associated
 execution units. Cross-cutting concerns access their data in-process using the
 same shared `Context` object.
 
@@ -30,7 +30,7 @@ if one exists for them. In the cases where an extremely clear, pre-existing
 option is not available, OpenTelemetry MUST provide its own `Context`
 implementation. If a language has a pre-existing `Context` implementation that
 is incompatible with OpenTelemetry, OpenTelemetry MUST provide
-its own `Context` implementation. Depending on the language, `Context` can 
+its own `Context` implementation. Depending on the language, `Context` can
 either be explicit or implicit. OpenTelemetry allows both options.
 
 A `Context` MUST be immutable, and its write operations MUST
--- a/specification/context/context.md
+++ b/specification/context/context.md
@@ -20,20 +20,22 @@ Table of Contents
 
 ## Overview
 
-A `Context` is a propagation mechanism which carries execution-scoped values
-across API boundaries and between logically associated execution units.
-Cross-cutting concerns access their data in-process using the same shared
-`Context` object.
-
-A `Context` MUST be immutable, and its write operations MUST
-result in the creation of a new `Context` containing the original
-values and the specified values updated.
+A `Context` is an in-process propagation mechanism which carries 
+execution-scoped values across API boundaries and between logically associated 
+execution units. Cross-cutting concerns access their data in-process using the
+same shared `Context` object.
 
 Languages are expected to use the single, widely used `Context` implementation
 if one exists for them. In the cases where an extremely clear, pre-existing
 option is not available, OpenTelemetry MUST provide its own `Context`
-implementation. Depending on the language, its usage may be either explicit
-or implicit.
+implementation. If a language has a pre-existing `Context` implementation that
+is incompatible with OpenTelemetry, OpenTelemetry MUST provide
+its own `Context` implementation. Depending on the language, `Context` can 
+either be explicit or implicit. OpenTelemetry allows both options.
+
+A `Context` MUST be immutable, and its write operations MUST
+result in the creation of a new `Context` containing the original
+values and the specified values updated.
 
 Users writing instrumentation in languages that use `Context` implicitly are
 discouraged from using the `Context` API directly. In those cases, users will


yurishkuro (approved):
+1 to rewording.

> If a language has a pre-existing `Context` implementation that is incompatible with OpenTelemetry

Just curious, what are specific examples of this?

rakyll:
This question came up as a part of a discussion for support for a new language. Not sure if I'm contributing at lot to the text here but at least it clarifies that implementors should review their choice of context implementation against the spec.

specification/context/context.md
```diff
@@ -20,20 +20,22 @@ Table of Contents
...
+execution units. Cross-cutting concerns access their data in-process using the
+same shared `Context` object.
 
 Languages are expected to use the single, widely used `Context` implementation
 if one exists for them. In the cases where an extremely clear, pre-existing
 option is not available, OpenTelemetry MUST provide its own `Context`
-implementation. Depending on the language, its usage may be either explicit
-or implicit.
+implementation. If a language has a pre-existing `Context` implementation that
+is incompatible with OpenTelemetry, OpenTelemetry MUST provide
```
anuraaga, line 32: Would a context that is mutable be considered incompatible? I guess even though when we provide an implementation we prefer it, if a language has a very idiomatic context API that everyone is already used to, not using it could be worse.

So yeah we need to define what it means to be incompatible better.

carlosalberto: Line 36 still says that a `Context` must be immutable, so, yes, we should definitely consider a mutable one being incompatible. As far as I remember, most existing context propagation systems are immutable (gRPC/Reactor in Java, contextvars for Python).

I don't _feel_ this could ever be a problem.

anuraaga: Not completely sure but I think cls or such on nodejs is mutable. And, assuming it's true for node or any other language, it shouldn't be a reason to reimplement context in otel.

carlosalberto: any information on this @open-telemetry/javascript-approvers ? ;)

Oberon00: @open-telemetry/python-approvers Is the context wrapper that is used instead of contextvars (if it is still used) actually immutable?

Flarna: The Javascript language has no built in context support.
In browsers OTel uses zone.js as base to transport the context; in node.js async-hooks/AsyncLocalStorage is used to transport the context. These transports do not mutate the context.

Both allow to transport any object (incl. mutable objects). The way how they are used by the OTel API ensures that context is not mutable.

rakyll: TBH, I can't think of a context that is not immutable. You should never be able to mutate the context in the call chain with side affects to your earlier call site.

Flarna: Even if the context object itself is imutable users will (and have to) store references to mutable objects into context (e.g. a span).

anuraaga: From what I can tell, zone.js does seem to be immutable in the sense that we define here, the only way to add properties is to create another zone

https://github.com/angular/angular/blob/master/packages/zone.js/lib/zone.ts#L174

AsyncLocalStorage seems to just use any JS object - there seems to be nothing from preventing code within a runnable to overwrite values in it affecting the callers

https://nodejs.org/api/async_hooks.html#async_hooks_asynclocalstorage_run_store_callback_args

At least to me, this seems to conflict with the wording about `MUST be immutable` - but the JS implementation I'm sure guards the user from this by hiding the context key despite the user being able to mutate the context. Assuming this is allowed, I think this new language needs to reflect it but without understanding better what's involved in being "compatible", I feel it currently doesn't.

Flarna: I think it's mostly the same as in Java. `ThreadLocal` is used for transporting the otel context. `ThreadLocal` doesn't enforce immutability but the way how it is exposed/used by OTel ensures that OTel context objects are immutable.

For JS see [`Context`](https://github.com/open-telemetry/opentelemetry-js-api/blob/493daa12235a3e2d4d6d263104b2bde4b59282b1/src/context/types.ts#L17-L41) interface for details.

anuraaga: Ah this seems like sort of a context implementation built using those other libraries as propagation mechanisms. In that case it does seem to be in the spirit of this text. Then it seems good to go.

rakyll:
Do we want to add this or do you think it's arbitrary? I can send out the wording changes in a separate PR and close this.

carlosalberto:
I think this is good to go. @anuraaga @Oberon00  any opinion?

anuraaga:
@carlosalberto I'm ok with the change but generally ambivalent, mostly because I don't think it really changes the fact that we need to understand what incompatible means on a case-by-case basis. For example, I still think the JS case is an example of using native context APIs that are strictly mutable, but by presenting an immutable API on top of it for opentelemetry itself - I don't know what we consider the "context implementation" in this case and feel like both the old and new language are similarly ambiguous on what to do in such a situation, which isn't a bad thing since it's language-specific and can be decided case-by-case

rakyll:
> but by presenting an immutable API on top of it for opentelemetry itself - I don't know what we consider the "context implementation"

If you are making the developer engage with the wrapper layer, it would be considered a context implementation.

carlosalberto:
@open-telemetry/javascript-approvers Would you mind chiming it? We definitely need your input/opinion ;)

Flarna:
What is required from JS side?

AsyncLocalStore used in Node.JS is mutable similar as ThreadLocal used in Java is mutable. I assume .NET is using AsyncLocal which is also mutable.

The context API on exposed by OTel-Js is completely hiding that what is used under the hood and the API deals with immutable objects. As far as I know the same is valid for Java.

anuraaga:
I think the crux it's still unclear what is a context implementation. I think there are generally two aspects to an OTel context implementation

1) An immutable copy-on-write map
2) Context propagation mechanisms

1) is specified by the methods defined in the spec, and 2) is specified I think by the first sentence "A `Context` is an in-process propagation mechanism which carries execution-scoped values across API boundaries and between logically associated execution units."

In Java, we implement both, 1) as a class called `Context` and 2) as methods `makeCurrent()` / `close`, and helpers such as `Executor`. I don't consider just the use of the `ThreadLocal` as using an existing context propagation mechanism

For JS, I think what I'm hearing is 1) is implemented as an interface in JS while 2) is implemented using mechanisms provided by NodeJS, which is great since users don't have to worry about context propagation usually if the language runtime is handling it

For Go 1) is in the standard library and 2) uses the Go convention of manual propagation, so nothing was implemented at all.

So does reimplementing just one or the other count as a language "implementing its own context implementation"? Maybe - I don't find this clear either in the current text or the proposed new text though.

anuraaga:
> we should offer as part of OTel a thin layer that makes it behave as immutable.

This is being very clear that the compatibility issue is with regards to immutability. So this directness seems like reasonable wording to me.

jsuereth (approved):

specification/context/context.md
jsuereth: My $.02 - If you do find a language where context is mutable we have known workaround for an immutable OTel context there.  However, in most cases users can still "be evil" and violate this (thinking Java, specifically).

I like the way this is worded and if we need to loosen it later based on experience, it's worth having that discussion with a  better-known example.

carlosalberto:
@rakyll Would you mind doing a last iteration to convey the idea that

> that Context MUST be immutable, and if a language defines a built-in mechanism that does not satisfy this, we should offer as part of OTel a thin layer that makes it behave as immutable.

?

PR #1478, opened 2021-02-26
Add SpanProcessor section to spec-compliance-matrix

MrAlias:
## Changes

The SpanProcessor section of the trace/sdk is added to the spec-compliance-matrix. This also includes Go SIG compliance with this section.

Resolves https://github.com/open-telemetry/opentelemetry-go/issues/1615


iNikem:
Do we really need to list all requirements for the API in compliance matrix? I don't think we did that for other parts of the API.

spec-compliance-matrix.md
```diff
@@ -72,6 +72,56 @@ status of the feature is not known.
...
+| `OnStart` called synchronously                                                                   |          | +  |      |    |        |      |        |     |      |     |      |       |
+| `OnEnd` called after span ended                                                                  |          | +  |      |    |        |      |        |     |      |     |      |       |
+| `OnEnd` called synchronously called within `Span.End()`                                          |          | +  |      |    |        |      |        |     |      |     |      |       |
+| `Shutdown` is called once for each registered SpanProcessor                                      |          | +  |      |    |        |      |        |     |      |     |      |       |
+| [Simple Span Processor](specification/trace/sdk.md#simple-processor)                             |          |    |      |    |        |      |        |     |      |     |      |       |
+| Configurable parameter SpanExporter                                                              |          | +  |      |    |        |      |        |     |      |     |      |       |
```
carlosalberto, line 85: Probably we can leave this out, as `SimpleSpanProcessor` (or any SpanProcessor) becomes essentially useless if it doesn't take an exporter, so we should take this one for granted.

spec-compliance-matrix.md
```diff
@@ -72,6 +72,56 @@ status of the feature is not known.
...
+| `OnStart` accepts a read/write span object                                                       |          | +  |      |    |        |      |        |     |      |     |      |       |
+| `OnStart` accepts the parent `Context`                                                           |          | +  |      |    |        |      |        |     |      |     |      |       |
+| `OnStart` returns void or nothing                                                                |          | +  |      |    |        |      |        |     |      |     |      |       |
+| `OnEnd` does not block or throw exceptions                                                       |          | +  |      |    |        |      |        |     |      |     |      |       |
+| `OnEnd` returns void or nothing                                                                  |          | +  |      |    |        |      |        |     |      |     |      |       |
+| `Shutdown` releases processor resources                                                          |          | +  |      |    |        |      |        |     |      |     |      |       |
```
carlosalberto, line 93: Probably "Shutdown is implemented" or "is supported" sounds more general (and thus better, at least for this matrix).

spec-compliance-matrix.md
```diff
@@ -72,6 +72,56 @@ status of the feature is not known.
...
+| `Shutdown` includes the effects of `ForceFlush`                                                  |          | +  |      |    |        |      |        |     |      |     |      |       |
+| `Shutdown` aborts after configured timeout                                                       |          | [-](https://github.com/open-telemetry/opentelemetry-go/issues/1616)  |      |    |        |      |        |     |      |     |      |       |
+| `ForceFlush` communicates failure of timeout                                                     |          | N/A|      |    |        |      |        |     |      |     |      |       |
+| `ForceFlush` aborts after configured timeout                                                     |          | N/A|      |    |        |      |        |     |      |     |      |       |
+| [Batching Span Processor](specification/trace/sdk.md#batching-processor)                         |          |    |      |    |        |      |        |     |      |     |      |       |
+| Configurable parameter SpanExporter                                                              |          | +  |      |    |        |      |        |     |      |     |      |       |
```
carlosalberto, line 101: Likewise.

spec-compliance-matrix.md
```diff
@@ -72,6 +72,56 @@ status of the feature is not known.
...
+| Configurable parameter export timeout                                                            |          | [-](https://github.com/open-telemetry/opentelemetry-go/issues/1613)  |      |    |        |      |        |     |      |     |      |       |
+| Default export timeout 30 seconds                                                                |          | [-](https://github.com/open-telemetry/opentelemetry-go/issues/1613)  |      |    |        |      |        |     |      |     |      |       |
+| Configurable parameter max batch size                                                            |          | +  |      |    |        |      |        |     |      |     |      |       |
+| Default max batch size 512                                                                       |          | +  |      |    |        |      |        |     |      |     |      |       |
+| Exports Span data as configured batches                                                          |          | +  |      |    |        |      |        |     |      |     |      |       |
```
carlosalberto, line 110: Similarly, this is taken for granted - otherwise this processor is not actually implemented at all. So maybe remove it? I don't think it exists anywhere without doing this.

carlosalberto (approved):

carlosalberto:
Overall LGTM. Left a pair of comments. Thanks for taking on this!

MrAlias:
> Do we really need to list all requirements for the API in compliance matrix? I don't think we did that for other parts of the API.

The Go SIG wanted to be sure that we addressed all normative requirements and found this helpful. We thought it might be helpful for other SIGs. Feel free to close if it is not.

spec-compliance-matrix.md
MrAlias: Hmm, I'm not sure a requirement being trivial should disqualify it from being stated in a checklist. Especially if there could be alternate implementations (e.g. the SpanProcessor takes a wrapped exporter of another SpanExporter). It seems like since the specification requires a particular signature to the Simple Span Processor we should check that it is implemented that way, no?

spec-compliance-matrix.md
MrAlias: I think it can be taken for grated that the processor exports in batches, but I think it should be validated that it exports in batches that match the configuration. I know there has been confusion over what period of time each timeout represents and should probably be validated to match what the specification states, no?

spec-compliance-matrix.md
```diff
@@ -72,6 +72,56 @@ status of the feature is not known.
...
+| [SpanProcessor](specification/trace/sdk.md#span-processor)                                       |          |    |      |    |        |      |        |     |      |     |      |       |
+| Only invoked for recorded Spans                                                                  |          | +  |      |    |        |      |        |     |      |     |      |       |
+| Supports user defined SpanProcessor                                                              |          | +  |      |    |        |      |        |     |      |     |      |       |
+| Invoked in order registered                                                                      |          | +  |      |    |        |      |        |     |      |     |      |       |
+| `OnStart` called when span is started                                                            |          | +  |      |    |        |      |        |     |      |     |      |       |
+| `OnStart` called synchronously                                                                   |          | +  |      |    |        |      |        |     |      |     |      |       |
```
arminru, line 80:
```suggestion
| `OnStart` called synchronously                                                                   |          | +  |      |    |        |      |        |     |      |     |      |       |
| `OnStart` accepts a read/write span object                                                       |          | +  |      |    |        |      |        |     |      |     |      |       |
```
This that should already be part of the interface definition. You can drop it from the two pre-defined implementations down below in return.

spec-compliance-matrix.md
```diff
@@ -72,6 +72,56 @@ status of the feature is not known.
...
+| `OnEnd` called synchronously called within `Span.End()`                                          |          | +  |      |    |        |      |        |     |      |     |      |       |
+| `Shutdown` is called once for each registered SpanProcessor                                      |          | +  |      |    |        |      |        |     |      |     |      |       |
+| [Simple Span Processor](specification/trace/sdk.md#simple-processor)                             |          |    |      |    |        |      |        |     |      |     |      |       |
+| Configurable parameter SpanExporter                                                              |          | +  |      |    |        |      |        |     |      |     |      |       |
+| Exports Span data immediately                                                                    |          | +  |      |    |        |      |        |     |      |     |      |       |
+| `OnStart` does not block or throw exceptions                                                     |          | +  |      |    |        |      |        |     |      |     |      |       |
```
ThomsonTan, line 87: Merge the lines for `OnStart` and `OnEnd` if the same requirement applies to both? Like `OnStart`/`OnEnd` do not block or throw exceptions.

MrAlias: It seems appropriate to have these as different items if an implementation can implement one but not the other. Otherwise, if unified, +, -, or N/A would not apply and it would be indeterminate what a SIG should put there.

carlosalberto:
> Do we really need to list all requirements for the API in compliance matrix? I don't think we did that for other parts of the API.

FWIW, I think this is useful, but probably we can 'trim' a few things, in order to keep things minimalistic.

MrAlias:
> FWIW, I think this is useful, but probably we can 'trim' a few things, in order to keep things minimalistic.

I think this work will need to be taken on by someone else. I've tried to list all the the explicit normative requirements for implementation SIGs that are included in the specification in a clear and precise manner. If there is a desire to have a subset of these requirements listed here I do not feel I'm in the group of people who have a pulse on what is subjectively needed here.

Please feel free to close this if that is the direction we want to take these changes.

PR #1480, opened 2021-03-01
Rename "faas" to function

rakyll:
The resource semantic conventions for serverless function uses "faas"
as a prefix. FaaS would be a name to identify the concept of "function
as a service", not the platforms or the workloads.
The semantic conventions we define here are used to
attribute to the functions not to the concept.
FaaS is also not as clear as "function" to developers who are not
familiar with industry terms such as IaaS/PaaS/FaaS.

This change renames FaaS to function to better represent that
the resource was a serverless function.


rakyll:
I will follow up with renaming https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/trace/faas.yaml if we think this is a good next step.

specification/resource/semantic_conventions/function.md
```diff
@@ -0,0 +1,23 @@
...
+**Status**: [Experimental](../../document-status.md)
+
+**type:** `function`
+
+**Description:** A serverless instance.
+
+<!-- semconv function_resource -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `function.name` | string | The name of the function being executed. | `my-function` | Yes |
```
anuraaga, line 12: Can this be too confusing vs just a normal "code function"?

Oberon00:
I am against this change (at least in the current form) for two reasons:

1. Renaming a whole semantic convention group is a large breaking change. Breaking changes are generally OK, since these conventions are still "experimental", but I think the name "faas" is not bad enough to justify this. If we are worried about discoverability, I'm sure we can add a prominent link e.g. in a Lambda section of the upcoming AWS semantic conventions.
2. I think when most users read "faas" they will not know what it is. But when they read "function" they think they know what it is, but it will be wrong. Thus, I think the name "function" is worse. @anuraaga already [brought up](https://github.com/open-telemetry/opentelemetry-specification/pull/1480#discussion_r584498168) the confusion with [`code.function`](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.0.1/specification/trace/semantic_conventions/span-general.md#source-code-attributes) that can arise for example. I think "serverless", "serverless_function", "faas_instance", "cloud_function" might be better names, but I am not convinced renaming is worth it.

rakyll:
Fair. I'm closing this.

PR #1482, opened 2021-03-01
Add container runtime to container conventions

rakyll:
This allows users to identify the engine and possible issues with the runtime when debugging.


anuraaga (approved):

carlosalberto (approved):

iNikem (approved):

semantic_conventions/resource/container.yaml
```diff
@@ -12,10 +12,15 @@ groups:
...
         brief: >
-          Container id. Usually a UUID, as for example used to
+          Container ID. Usually a UUID, as for example used to
           [identify Docker containers](https://docs.docker.com/engine/reference/run/#container-identification).
           The UUID might be abbreviated.
         examples: ['a3bf90e006b2']
+      - id: runtime
+        type: string
+        brief: >
+          Container runtime running the container.
```
Oberon00, line 22: In my understanding, the Linux kernel is always running the container. The container engine is just the user interface for starting/stopping the container.

So maybe:

```suggestion
          The container runtime managing this container.
```

semantic_conventions/resource/container.yaml
```diff
@@ -12,10 +12,15 @@ groups:
...
+        examples: ['docker', 'containerd', rkt']
```
arminru, line 23:
```suggestion
        examples: ['docker', 'containerd', 'rkt']
```
Otherwise this is interpreted as `rkt'`.

rakyll: Good catch, this was a typo.

rakyll:
Addressed the feedback, please take a look.

arminru:
Actually it would probably be better to use an enum like we did elsewhere, for example here:

https://github.com/open-telemetry/opentelemetry-specification/blob/703309156f6387365c967d26141ad90af0dda66b/semantic_conventions/resource/cloud.yaml#L7-L22
https://github.com/open-telemetry/opentelemetry-specification/blob/703309156f6387365c967d26141ad90af0dda66b/specification/resource/semantic_conventions/cloud.md#L22-L28

This will allow constant classes or helpers generated from the YAML definitions to provide constants for these well-known engines in order to ensure we always get the same values. If users have to provide the string values themselves, we will probably get `crio` in one case and `CRI-O` in another.
@rakyll Do you think this would make sense?

PR #1483, opened 2021-03-01
Allow to collect host architecture

rakyll:
Introduce "arch" to represent the CPU architecture of the host, e.g. the processor on a physical or virtual host.
We may predefine some commonly used architecture as a follow up to this change.


semantic_conventions/resource/host.yaml
```diff
@@ -21,6 +21,11 @@ groups:
         brief: >
           Type of host. For Cloud, this must be the machine type.
         examples: ['n1-standard-1']
+      - id: arch
+        type: string
+        brief: >
+            The CPU architecture the host system is running on.
```
anuraaga, line 27: If it seems manageable I think an enum could work better for this - if a specialized architecture exists and can use OpenTelemetry, then we'd always be able to add it anyways, otherwise there aren't that many archs.

Otherwise I'd recommend doing similar to runtime and add descriptions of how to get this value in various languages to include any normalization that could be needed, eg replacing hyphen with underscore, etc.

Oberon00: @rakyll You wrote in the PR description

> We may predefine some commonly used architecture as a follow up to this change.

But I think we should already have these in the first version. Restricting a free-form string to an enum (even if one with `allow_custom_values: true`) would be a breaking change.

Oberon00:
I think we should look a bit into how this would work with "mixed-arch" scenarios. E.g. it is common to have a 32 bit process running on an 64 bit OS on x86_64. That gives potential for `os.arch`, `process.arch` in addition to `host.arch` in this PR. Or is `host.arch` the only architecture, and the rest would only be ABIs? Interesting in that regard is the existence of the [x32 ABI](https://en.wikipedia.org/wiki/X32_ABI), i.e. 64 bit instructions with 32 bit pointers on a 64 bit Linux.

rakyll:
@Oberon00 good point about how this will affect the architecture we may collect for the process and OS. Let me think about this a bit before we merge anything. @anuraaga I'll also provide a predefined list on the next iteration.

semantic_conventions/resource/host.yaml
reyang, line 27: Is this the host CPU architecture or the host OS architecture?

For example, one could run x86 version of Windows OS on an AMD64 hardware. The architecture of an executable image, a process, an operating system (including a subsystem) and a hardware could be all different.

Oberon00: Can the image really have a different arch from it's process?

rakyll: This is the CPU architecture of the host system, the VM's CPU.

The OS architecture would be in the os.yaml.

reyang: > Can the image really have a different arch from it's process?

An image could be `AnyCPU` for .NET Core app. It could be an architecture independent package (e.g. Mach-O).

reyang (approved):
LGTM.

semantic_conventions/resource/host.yaml
Oberon00: > An image could be AnyCPU for .NET Core app. It could be an architecture independent package (e.g. Mach-O).

The process image would then be the image of the .NET Runtime executable IMHO.

Oberon00: @rakyll 
> This is the CPU architecture of the host system, the VM's CPU.

Sounds good 👍  Can you add that point to the PR?

jmacd (approved):

semantic_conventions/resource/host.yaml
rakyll: Done. Added predefined options as well.

Are you happy with formatting? ALLCAPS_WITHUNDERSCORES?

rakyll:
PTAL

carlosalberto (approved):

semantic_conventions/resource/host.yaml
Oberon00: Looking at other enums (e.g. faas.trigger, messaging.operation) it seems that we use all-lowercase with underscores elsewhere.
Note that when generating code from the semantic convention it is possible to uppercase if that is the target language's convention.

semantic_conventions/resource/host.yaml
```diff
@@ -21,6 +21,27 @@ groups:
...
+            - id: ARM_32
+              value: 'ARM_32'
+              brief: "ARM 32-bit"
+            - id: AMD_64
+              value: 'AMD_64'
+              brief: "AMD 64-bit"
+            - id: x86
+              value: 'x86'
+              brief: "x86 32-bit"
+            - id: x86_64
```
Oberon00, line 40: I think AMD 64 and x86_64 are synonyms.

Oberon00 (changes requested):

specification/resource/semantic_conventions/host.md
```diff
@@ -12,7 +12,18 @@
 | `host.id` | string | Unique host ID. For Cloud, this must be the instance_id assigned by the cloud provider. | `opentelemetry-test` | No |
 | `host.name` | string | Name of the host. On Unix systems, it may contain what the hostname command returns, or the fully qualified hostname, or another name specified by the user. | `opentelemetry-test` | No |
 | `host.type` | string | Type of host. For Cloud, this must be the machine type. | `n1-standard-1` | No |
+| `host.arch` | string | The CPU architecture the host system is running on. | `ARM_64` | No |
 | `host.image.name` | string | Name of the VM image or OS install the host was instantiated from. | `infra-ami-eks-worker-node-7d4ec78312`; `CentOS-8-x86_64-1905` | No |
 | `host.image.id` | string | VM image ID. For Cloud, this value is from the provider. | `ami-07b06b442921831e5` | No |
 | `host.image.version` | string | The version string of the VM image as defined in [Version Attributes](README.md#version-attributes). | `0.1` | No |
+
+`host.arch` MUST be one of the following or, if none of the listed values apply, a custom value:
```
maxgolov, line 20: Values here neither match the common convention for Linux / Debian arch, e.g. not matching what's reported by `uname -m`, nor match what Node.js `os.arch()` API would have reported; nor what Rust architecture uses for the standard set of arch.... For different languages - you'd have different strings. And none of them would match the list proposed here 😄  So it's going to be quite an adventure to normalize to that proposed list 🤷 

I believe a better set could be this:

Architecture | Applies to
-- | --
x86 | x86, wow64, i386, i486, etc.
amd64 | x86_64, amd64, x64
arm32 | arm32
arm64 | arm64, aarch64
ppc32 | 32-bit PowerPC
ppc64 | 64-bit PowerPC
ia64 | Itanium
other | indistinguishable, or not listed here. **Architecture as reported by the language of choice without normalizing it / without bringing it to common denominator.**

C++, Rust, or Node - typically have their own API to obtain the process arch string. This string is going to be in their own common format. For Node - this list is well-described here: https://nodejs.org/api/os.html#os_os_arch - I personally believe Node's list is quite reasonable.

More complete set of architectures is described here in Linux kernel tree:
https://github.com/torvalds/linux/tree/master/arch

If you truly intend to describe the host arch , it would be reasonable to respect the Linux standard names for the archs, as the most prominent host OS. That's also what `uname -m` command would _typically_ report, namely `[arm|arm64|ia64|mips|x86]`. I would've left the loophole in this list to allow reporting what OS language-specific API reports as architecture. Any data analysis within a slice of data reported by agents written in specific language is consistent: a language (e.g. `node`) would report only a fixed set of archs. There is also not going to be any need to normalize.

I don't like this proposed list because it does not seem to align with the arch lists typically captured by prominent language APIs, neither matches the Linux kernel arch list.

semantic_conventions/resource/host.yaml
maxgolov: Don't restrict this to enum, please.

specification/resource/semantic_conventions/host.md
rakyll: This looks so much better. Go has a similar approach here. I initially wanted to propose something more aligned with uname but the formatting of OS names made me want to follow the UPPER_CASE style here. I'm modifying this PR in a bit.

rakyll:
PTAL

semantic_conventions/resource/host.yaml
```diff
@@ -21,6 +21,36 @@ groups:
...
+            - id: ppc32
+              value: 'ppc32'
+              brief: "32-bit PowerPC"
+            - id: ppc64
+              value: 'ppc64'
+              brief: "64-bit PowerPC"
+            - id: x86
+              value: 'x86'
+              brief: "32-bit x86"
+            - id: other
```
anuraaga, line 49: Can we just leave out the attribute if unknown instead of having an enum for it?

rakyll:
PTAL

carlosalberto:
@Oberon00 @maxgolov Please confirm this is ready to go (I _think_ it is ;) )

@rakyll Usually marking comments as resolved makes it easier to review, btw.

anuraaga (approved):

rakyll:
@carlosalberto will resolve them from this point on, thanks for pointing it out.

carlosalberto:
@maxgolov @Oberon00 Feedback as been properly addressed. In case of any further detail, please fill a follow up ;)

PR #1048, opened 2020-10-01
Clarify where links can be added, format add-link as add-event

bogdandrutu:
Signed-off-by: Bogdan Drutu <bogdandrutu@gmail.com>

Updates #850

PR #1488, opened 2021-03-01
Remove the AddLink section from the spec-compliance-matrix

MrAlias:
This was removed in #1048

### DIFF
--- a/spec-compliance-matrix.md
+++ b/spec-compliance-matrix.md
@@ -56,9 +56,6 @@ status of the feature is not known.
 | Array of primitives (homogeneous)                                                                |          | +  | +    | +  | +      | +    | +      | +   | +    | +   | +    | +     |
 | `null` values documented as invalid/undefined                                                    |          |    | +    | +  | +      | +    | N/A    |     |      | +   |      | N/A   |
 | Unicode support for keys and string values                                                       |          | +  | +    | +  | +      | +    | +      | +   | +    | +   | +    | +     |
-| [Span linking](specification/trace/api.md#specifying-links)                                      |          |    |      |    |        |      |        |     |      |     |      |       |
-| AddLink                                                                                          |          | +  | +    | +  | +      | +    | +      | +   | +    | -   | +    | +     |
-| Safe for concurrent calls                                                                        |          | +  | +    | +  | +      | +    | +      | +   | +    | +   | +    | +     |
 | [Span events](specification/trace/api.md#add-events)                                             |          |    |      |    |        |      |        |     |      | +   |      |       |
 | AddEvent                                                                                         |          | +  | +    | +  | +      | +    | +      | +   | +    | +   | +    | +     |
 | Add order preserved                                                                              |          | +  | +    | +  | +      | +    | +      | +   | +    | +   | +    | +     |


carlosalberto (approved):

spec-compliance-matrix.md
```diff
@@ -56,9 +56,6 @@ status of the feature is not known.
 | Array of primitives (homogeneous)                                                                |          | +  | +    | +  | +      | +    | +      | +   | +    | +   | +    | +     |
 | `null` values documented as invalid/undefined                                                    |          |    | +    | +  | +      | +    | N/A    |     |      | +   |      | N/A   |
 | Unicode support for keys and string values                                                       |          | +  | +    | +  | +      | +    | +      | +   | +    | +   | +    | +     |
-| [Span linking](specification/trace/api.md#specifying-links)                                      |          |    |      |    |        |      |        |     |      |     |      |       |
```
jkwatson, line 59: We can't remove span linking altogether, can we? Even though you can't add links after a span has started, you can definitely create a span with links associated: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#specifying-links

MrAlias: Sure. This just removes the old method the spec used to require, it is not intended to signify linking is removed.

jkwatson: doesn't this PR explicitly delete the line that links to the links section? This line that I'm commenting on?

arminru: `AddLink` is just the proposed name for adding a link _during span creation_ for languages that use a builder pattern. In Java, for example, `addLink` is defined on `SpanBuilder` and this is what this entry is about. See https://github.com/open-telemetry/opentelemetry-java/blob/v1.0.0/api/all/src/main/java/io/opentelemetry/api/trace/SpanBuilder.java#L133-L143.

@MrAlias You could reword it so that it's clear that this is about specifying links at span creation, not later on.
E.g. "Links can be specified at span creation" instead of "AddLink" in the line below.

arminru (changes requested):

spec-compliance-matrix.md
jkwatson: My point is that if we accept this PR, then there will be no reference left to span linking in the compliance matrix, which doesn't seem right, since it's still a feature that we need to support.

MrAlias: > doesn't this PR explicitly delete the line that links to the links section? This line that I'm commenting on?

Gotcha :+1:, I didn't understand what you we recommending. I'll add the link back in to the section.

jkwatson (approved):

anuraaga (approved):

rakyll (approved):

anuraaga:
@mralias Can you update the PR title? Or if you're ok with me updating can do

arminru (approved):

spec-compliance-matrix.md
arminru: Thanks, Tyler!

MrAlias:
@anuraaga I've updated the title and description. Feel free to update yourself if you feel there are further changes needed. I welcome the help :smile:.

bogdandrutu (approved):

carlosalberto (approved):

PR #1492, opened 2021-03-02
Ignore invalid spancontext in addLink

### LINKED ISSUE #1486: What should happen when you add an invalid span as a link?

What should happen when you add a link to the invalid span, e.g. in Java `spanBuilder.addLink(Span.getInvalid())`? Right now this situation is not described by the spec, and in Java SDK it turns out that a link to the invalid span is added and can be even exported.

**Additional context.**

* https://github.com/open-telemetry/opentelemetry-java/issues/2955

anuraaga:
## Changes

Adds text that adding links should be ignored if the links are to an invalid spancontext. This seems like reasonable behavior.

Fixes #1486


bogdandrutu (approved):

specification/trace/api.md
```diff
@@ -405,7 +405,8 @@ The Span creation API MUST provide:
   arguments. This MAY be called `AddLink`. This API takes the `SpanContext` of
   the `Span` to link to and optional `Attributes`, either as individual
   parameters or as an immutable object encapsulating them, whichever is most
-  appropriate for the language.
+  appropriate for the language. If the `SpanContext` is not [valid](#isvalid), the
+  call SHOULD be ignored.
```
bogdandrutu, line 409: This seems to be an implementation detail. Do we want to have this as the contract for the API?

anuraaga: If an implementation doesn't do it, I guess a link with `00000...` IDs will be sent all the way to the exporter - so I think we need it here to decide on whether this is something we want or not.

anuraaga: Sorry for missing extra context too - instrumentation has a tendency to want to pass in an invalid link rather than having to check any time it adds a link, like here

https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2263/files#r584866125

It would be good to confirm it's ok for SDKs to ignore such usage, or alternatively require instrumentation to deal with it.

bogdandrutu: >If an implementation doesn't do it, I guess a link with 00000... IDs will be sent all the way to the exporter 

Not sure if we talk about "exporter" in that case :) a custom implementation may not use the exporters we have or OTLP at all.

bogdandrutu: I think the SDK can ignore that. So the API contract can be "A valid implementation of this API can ignore links with invalid SpanContext" then our SDK as an implementation of the API we do that :)

Does it make sense?

bogdandrutu: I am trying to keep the separation of "what is required for every implementation of the API" vs "what our SDK (API implementation) does"

anuraaga: Ah you're referring to API vs SDK, got it

bogdandrutu:
@anuraaga thanks

anuraaga:
Assigned @bogdandrutu (assuming this is why @yurishkuro removed assignment? :P)

iNikem (approved):

mateuszrzeszutek (approved):

arminru (approved):
@anuraaga Please add an entry in the changelog.

carlosalberto (approved):

rakyll (approved):

anuraaga:
@arminru Thanks updated changelog

reyang (approved):

anuraaga:
@open-telemetry/technical-committee Can this be merged? It has many approvals.

yurishkuro (approved):

PR #745, opened 2020-07-27
Clarify how to make progress on a PR

reyang:
## Changes

This was discussed in the 7/27/2020 specification SIG meeting.
The goal is to have a better clarity on how to make progress on PRs.

PR #1494, opened 2021-03-02
Kick off the ownership clarification

reyang:
Per discussion during the [03/02/2021 Spec SIG Mtg](https://docs.google.com/document/d/1-bCYkN-DWJq4jw1ybaDZYYmx-WAe6HnwfWbkm8d57v8/edit#heading=h.qjk2i7r08k0r), I'm sending this PR in hope that all the spec docs would have owners/experts clarified. This could become very useful for semantic convention spec.

@rakyll please review.

@bogdandrutu @jmacd @jsuereth I put your names as the "domain experts" for metrics API spec, please approve if you'd like to be listed in the doc. If I haven't got your approval, I'll assume that you don't want to be listed here so I'll remove your name.

### DIFF
--- a/specification/metrics/api.md
+++ b/specification/metrics/api.md
@@ -2,6 +2,16 @@
 
 **Status**: [Experimental](../document-status.md)
 
+**Owner:**
+
+* [Reiley Yang](https://github.com/reyang)
+
+**Domain Experts:**
+
+* [Bogdan Brutu](https://github.com/bogdandrutu)
+* [Josh Suereth](https://github.com/jsuereth)
+* [Joshua MacDonald](https://github.com/jmacd)
+
 Note: this specification is subject to major changes. To avoid thrusting
 language client maintainers, we don't recommend OpenTelemetry clients to start
 the implementation unless explicitly communicated via


rakyll (approved):

specification/metrics/api.md
```diff
@@ -2,6 +2,16 @@
 
 **Status**: [Experimental](../document-status.md)
 
+**Owner:**
```
Oberon00, line 5: We need to define what is (not) the responsibility of the Owner, and the Domain Experts.

rakyll: We may also add a few sentences lines when domain experts assigned to a change. Are they always added (1) or are they responsible to handle the disputes (2)? I'd prefer (1).

reyang: I'm happy to work on a similar PR like https://github.com/open-telemetry/opentelemetry-specification/pull/745 after this PR.
Do you think this is blocking? My thinking is that we need to clarify the ownership anyways, so we don't have to be blocked.

reyang: @rakyll I vote for (1).

SergeyKanzhelev: would (1) also mean that we can add @ jsuereth as metrics approver than: https://github.com/open-telemetry/community/blob/01acd5d2c39b764554ac3d87a64a18a09250e9a3/community-members.md#user-content-specifications-and-proto:~:text=Metrics%20Approvers%3A

SergeyKanzhelev: should this list be synchronized with metrics approvers https://github.com/open-telemetry/community/blob/01acd5d2c39b764554ac3d87a64a18a09250e9a3/community-members.md#user-content-specifications-and-proto:~:text=Metrics%20Approvers%3A?

reyang: > should this list be synchronized with metrics approvers https://github.com/open-telemetry/community/blob/01acd5d2c39b764554ac3d87a64a18a09250e9a3/community-members.md#user-content-specifications-and-proto:~:text=Metrics%20Approvers%3A?

My answer would be no, approvers don't necessarily need to own the spec (e.g. if the spec is outdated, the owner should be responsible to fix it, guess this could be a separate clarification PR like #745).

tigrannajaryan:
Should we also add the owner and experts to CODEOWNERS to explicitly assign PRs that touch the area?

specification/metrics/api.md
SergeyKanzhelev: If we go with (1), CODEOWNERS as suggested by @tigrannajaryan below would be more automated solution. This led me to think of an approvers list.

I like the idea of explicitly listing owners - it solves an immediate issue. Maybe without overthinking the right approach we can go with just a document header as you suggest and then re-work when needed.

SergeyKanzhelev (approved):
it may not be a long term solution, but it may be useful now.

specification/metrics/api.md
reyang: Yep, I think having explicit owners listed here could be the 1st step, once we have all the document with owners clarified, changing it to a different format (e.g. CODEOWNERS file) could be an easy job (and we can enforce that via CI).

reyang:
> Should we also add the owner and experts to CODEOWNERS to explicitly assign PRs that touch the area?

I think later we can explore the automation (e.g. ensure CODEOWNERS reflect the approvers by adding a simple CI script).

specification/metrics/api.md
iNikem: > We need to define what is (not) the responsibility of the Owner, and the Domain Experts.

+1 on this.

iNikem (approved):

jsuereth (approved):

carlosalberto (approved):

MrAlias (approved):

PR #1481, opened 2021-03-01
Clarify the cloud resource attributes

rakyll:
Provide better examples representing account IDs, regions and zones
from non-GCP providers. Improve the description to explain the
relationship between resource and cloud attributes:
The resource is created or running in the account, region and zones.

PR #1495, opened 2021-03-03
Rename cloud.zone to cloud.availability_zone

rakyll:
Only Google Cloud calls availability zones a zone. Renaming zone
to make the least common case an exception.

This change is a follow-up of https://github.com/open-telemetry/opentelemetry-specification/pull/1481.

### DIFF
--- a/semantic_conventions/resource/cloud.yaml
+++ b/semantic_conventions/resource/cloud.yaml
@@ -37,7 +37,7 @@ groups:
         type: string
         brief: >
           Cloud regions often have multiple, isolated locations known as zones
-          to increase availability. Availability zone represents the isolated
+          to increase availability. Availability zone represents the
           zone where the resource is running.
         note: >
           Google Cloud calls availability zones zones.
--- a/semantic_conventions/resource/cloud.yaml
+++ b/semantic_conventions/resource/cloud.yaml
@@ -33,15 +33,15 @@ groups:
           [AWS regions](https://aws.amazon.com/about-aws/global-infrastructure/regions_az/), [Azure regions](https://azure.microsoft.com/en-us/global-infrastructure/geographies/), or
           [Google Cloud regions](https://cloud.google.com/about/locations).
         examples: ['us-central1', 'us-east-1']
-      - id: zone
+      - id: availability_zone
         type: string
         brief: >
           Cloud regions often have multiple, isolated locations known as zones
-          to increase availability. Zone represents the zone where the resource 
-          is running.
+          to increase availability. Availability zone represents the isolated
+          zone where the resource is running.
         note: >
-          AWS and Azure calls them availability zones or AZs.
-        examples: ['us-central1-a', 'us-east-1c']
+          Google Cloud calls availability zones zones.
+        examples: ['us-east-1c']
       - id: infrastructure_service
         type:
           allow_custom_values: true
--- a/semantic_conventions/resource/cloud.yaml
+++ b/semantic_conventions/resource/cloud.yaml
@@ -40,7 +40,7 @@ groups:
           to increase availability. Availability zone represents the
           zone where the resource is running.
         note: >
-          Google Cloud calls availability zones zones.
+          Google Cloud calls availability zones just "zones".
         examples: ['us-east-1c']
       - id: infrastructure_service
         type:
--- a/semantic_conventions/resource/cloud.yaml
+++ b/semantic_conventions/resource/cloud.yaml
@@ -40,7 +40,7 @@ groups:
           to increase availability. Availability zone represents the
           zone where the resource is running.
         note: >
-          Google Cloud calls availability zones just "zones".
+          Availability zones are called "zones" on Google Cloud.
         examples: ['us-east-1c']
       - id: infrastructure_service
         type:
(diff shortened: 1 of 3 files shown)


semantic_conventions/resource/cloud.yaml
```diff
@@ -33,15 +33,15 @@ groups:
...
         brief: >
           Cloud regions often have multiple, isolated locations known as zones
-          to increase availability. Zone represents the zone where the resource 
-          is running.
+          to increase availability. Availability zone represents the isolated
+          zone where the resource is running.
         note: >
-          AWS and Azure calls them availability zones or AZs.
-        examples: ['us-central1-a', 'us-east-1c']
+          Google Cloud calls availability zones zones.
```
reyang, line 43: nit - avoid "zones zones"?

```suggestion
          Google Cloud calls them zones.
```

specification/resource/semantic_conventions/cloud.md
```diff
@@ -12,10 +12,10 @@
...
 | `cloud.region` | string | The geographical region the resource is running. Refer to your provider's docs to see the available regions, for example [AWS regions](https://aws.amazon.com/about-aws/global-infrastructure/regions_az/), [Azure regions](https://azure.microsoft.com/en-us/global-infrastructure/geographies/), or [Google Cloud regions](https://cloud.google.com/about/locations). | `us-central1`; `us-east-1` | No |
-| `cloud.zone` | string | Cloud regions often have multiple, isolated locations known as zones to increase availability. Zone represents the zone where the resource  is running. [1] | `us-central1-a`; `us-east-1c` | No |
+| `cloud.availability_zone` | string | Cloud regions often have multiple, isolated locations known as zones to increase availability. Availability zone represents the isolated zone where the resource is running. [1] | `us-east-1c` | No |
 | `cloud.infrastructure_service` | string | The cloud infrastructure resource in use. [2] | `aws_ec2`; `azure_vm`; `gcp_compute_engine` | No |
 
-**[1]:** AWS and Azure calls them availability zones or AZs.
+**[1]:** Google Cloud calls availability zones zones.
```
reyang, line 18: Or consider this:

```suggestion
**[1]:** Google Cloud calls availability zones **zones**.
```

reyang (approved):
It's like a political game, but your statement "only Google calls..." makes sense to me.
More importantly, I think availability zone is just a better term than "zone" since it is more specific.

iNikem (approved):

anuraaga (approved):

semantic_conventions/resource/cloud.yaml
```diff
@@ -33,15 +33,15 @@ groups:
...
           [Google Cloud regions](https://cloud.google.com/about/locations).
         examples: ['us-central1', 'us-east-1']
-      - id: zone
+      - id: availability_zone
         type: string
         brief: >
           Cloud regions often have multiple, isolated locations known as zones
-          to increase availability. Zone represents the zone where the resource 
-          is running.
+          to increase availability. Availability zone represents the isolated
```
Oberon00, line 40:
```suggestion
          to increase availability. Availability zone represents the
```

I would not repeat "isolated" here.

Oberon00 (approved):

alolita (approved):

semantic_conventions/resource/cloud.yaml
arminru: or
```suggestion
          Google Cloud calls availability zones just "zones".
```

arminru (approved):

specification/resource/semantic_conventions/cloud.md
jsuereth: Or:  
`Availability zones are called "zones" on Google Cloud`

rakyll:
PTAL

semantic_conventions/resource/cloud.yaml
```diff
@@ -33,15 +33,15 @@ groups:
           [AWS regions](https://aws.amazon.com/about-aws/global-infrastructure/regions_az/), [Azure regions](https://azure.microsoft.com/en-us/global-infrastructure/geographies/), or
           [Google Cloud regions](https://cloud.google.com/about/locations).
         examples: ['us-central1', 'us-east-1']
-      - id: zone
+      - id: availability_zone
```
yurishkuro, line 36: Did we consider just `az`? This is a very long string that will be repeated in every span.

bogdandrutu: It is in the resource not the Span :), but still worth considering `az` maybe

yurishkuro: That distinction only matters at emission side, in the storage the resource attributes are usually denormalized into each span / log.

rakyll: I think zone is more readable if we think availability_zone is too long. Happy to keep things as they are.

carlosalberto: I'd suggest going with either `az` or `zone`.

Oberon00: zone was the original name, before this PR. I suggested availability_zone, and I think it is more clear. az is not consistent with our other semantic convention names, which are all on the verboser side.

carlosalberto: @Oberon00 Sounds like `az` is a no-go. Would `zone` be a good compromise? i.e. what we currently have ;)

Oberon00: I would not say `az` is a no-go. Personally I think many of our conventions are more verbose than they should be.
I can also live with `zone`, i.e. what we currently have. I would have preferred `availability_zone` over it though.

bogdandrutu: What I am hearing from this thread is that we do not have clear requirements for the semantic conventions names, and probably I would say that should be our next action item before merging or making a decision on this:
1. human readable vs storage space is one of the main topic to decide on.

carlosalberto: Let's discuss human readable vs storage space in the next Spec meeting.

rakyll: I'd prefer to keep it as zone if length is an issue. `az` is not readable enough.

rakyll:
We discussed in the meeting and decided that readability matters more. We always have a way to compress the keys in the future on the wire and at rest if it's needed. This change should be good to be merged.

I will also send a change with recommendations on max key lengths as a guideline as we discussed earlier.

rakyll:
@Oberon00 @iNikem @arminru @reyang Can we merge this change?

PR #1498, opened 2021-03-03
Add owners section to all the existing specs

reyang:
This is a follow up to #1494.

## Changes

I've simply added owner section to existing specs, and leave the actual owners as TBD (we can use the upcoming spec SIG meeting to drive the ownership).
After this PR, I will send a small PR to improve the CI, so we don't have new documents added that don't have the owners section.


carlosalberto (approved):

cijothomas (approved):

bogdandrutu (changes requested):
I am blocking this and also I want to propose reverting #1494.

Main reasons:
1. Owner is a new concept undefined in otel org. we do have concept of maintainers approvers and the right set of people are part of these groups based on clear rules.
2. Domain Expert in not defined as well, and it is not clear what are the rules to be one.

SergeyKanzhelev:
I don't think adding TBD everywhere solves any purpose short term. It's better to have targeted updates for key documents.

carlosalberto:
@reyang Do we want to maybe close this PR for now, while we get the clarifications in-place (and decide how to proceed from here)?

PR #1502, opened 2021-03-04
http.url MUST NOT contain credentials

pellared:
## Why

Example request's URL: `https://username:password@example.com:8080/path/to/file.aspx?query=1#fragment`

What is the expected attribute's value in such a case? For sure we should not pass the password. The username should not there as well per https://github.com/open-telemetry/opentelemetry-specification/pull/1502#discussion_r587860542.

Why do I think it is important? Because I saw this in OTel .NET SDK: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs#L115

Reference:
- https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#common-attributes
- https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#general-identity-attributes

## What 

Mandate to get rid of username (thanks to @yurishkuro) and password if it is present in the URL.

### DIFF
--- a/semantic_conventions/trace/http.yaml
+++ b/semantic_conventions/trace/http.yaml
@@ -17,8 +17,8 @@ groups:
             Full HTTP request URL in the form `scheme://host[:port]/path?query[#fragment]`.
             Usually the fragment is not transmitted over HTTP, but if it is known, it should be included nevertheless.
         note: >
-          `http.url` MUST NOT contain the password if it is passed via URL in form of `https://username:password@www.example.com/`.
-          In such case the attribute's value can be set as `https://username:@www.example.com/` or `https://www.example.com/`.
+          `http.url` MUST NOT contain any credentials passed via URL in form of `https://username:password@www.example.com/`.
+          In such case the attribute's value should be `https://www.example.com/`.
         examples: ['https://www.foo.bar/search?q=OpenTelemetry#SemConv']
       - id: target
         type: string
--- a/semantic_conventions/trace/http.yaml
+++ b/semantic_conventions/trace/http.yaml
@@ -17,7 +17,7 @@ groups:
             Full HTTP request URL in the form `scheme://host[:port]/path?query[#fragment]`.
             Usually the fragment is not transmitted over HTTP, but if it is known, it should be included nevertheless.
         note: >
-          `http.url` MUST NOT contain any credentials passed via URL in form of `https://username:password@www.example.com/`.
+          `http.url` MUST NOT contain credentials passed via URL in form of `https://username:password@www.example.com/`.
           In such case the attribute's value should be `https://www.example.com/`.
         examples: ['https://www.foo.bar/search?q=OpenTelemetry#SemConv']
       - id: target
--- a/semantic_conventions/trace/http.yaml
+++ b/semantic_conventions/trace/http.yaml
@@ -16,6 +16,9 @@ groups:
         brief: >
             Full HTTP request URL in the form `scheme://host[:port]/path?query[#fragment]`.
             Usually the fragment is not transmitted over HTTP, but if it is known, it should be included nevertheless.
+        note: >
+          `http.url` MUST NOT contain the password if it is passed via URL in form of `https://username:password@www.example.com/`.
+          In such case the attribute's value can be set as `https://username:@www.example.com/` or `https://www.example.com/`.
         examples: ['https://www.foo.bar/search?q=OpenTelemetry#SemConv']
       - id: target
         type: string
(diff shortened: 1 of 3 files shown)


jsuereth (approved):
Would you mind updating changelog with this change?

pellared:
> Would you mind updating changelog with this change?

Done https://github.com/open-telemetry/opentelemetry-specification/pull/1502/commits/1bcdc0bce040dce7ef64ffe4191dc82c91a59af6

Oberon00 (approved):

pjanotti:
cc @cijothomas for the OTel.NET SDK mention

pjanotti (approved):

semantic_conventions/trace/http.yaml
```diff
@@ -16,6 +16,9 @@ groups:
         brief: >
             Full HTTP request URL in the form `scheme://host[:port]/path?query[#fragment]`.
             Usually the fragment is not transmitted over HTTP, but if it is known, it should be included nevertheless.
+        note: >
+          `http.url` MUST NOT contain the password if it is passed via URL in form of `https://username:password@www.example.com/`.
+          In such case the attribute's value can be set as `https://username:@www.example.com/` or `https://www.example.com/`.
```
yurishkuro, line 21: I would suggest going with a more strict rule and require stripping everything before `@`. If people want username then can always add it separately, but logging username could be a huge privacy concern given recent EU regulations like ePD.

pellared: I agree, GDPR etc. However the specs already mention stuff about extracting the user identity. https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#general-identity-attributes therefore I think this will be not consistent. I would rather merge this and ask the maintainers to discuss the username problem in a broader context.

pellared: @yurishkuro 
Addressed https://github.com/open-telemetry/opentelemetry-specification/pull/1502/commits/0bfa6d5b9e75bc0beafa22ef84165cde27432d98
Could you please double-check?

yurishkuro (approved):

bogdandrutu (approved):

anuraaga (approved):

arminru (approved):

carlosalberto (approved):

veera83372:
I understand because of sensitivity of info we have to clear everything but for enduser it won't be evident that URL was containing username and password when it got here. Instead if we have redacted version like `https://***:***@www.example.com/` it will be more evident in expressing URL contains username and password but they are redacted for security reasons.

If we don't go with redacted version it might cause ambiguity for someone checking if url contains creds or not because in span with creds and without creds will be same.

options:
1. Use redacted version like `https://***:***@www.example.com/`
2. An attribute to say that credential info is removed (As @matej-g suggested [here)](https://github.com/open-telemetry/opentelemetry-go/issues/1675#issuecomment-793550867)

pellared:
@veera83372 

Why?
1. If a malicious user accesses this data, he would gain knowledge that could help him in making deeper attacks. Removing it completely reduces the attack surface.
1. Doing `***` is anonymization which AFAIK according to GDPR is a kind of personal data processing. Not even touching this info makes sure that won't be processed.

If someone needs such data I would rather create a dedicated attribute that is OPTIONAL (disabled by default). I guess this is what is suggested here https://github.com/open-telemetry/opentelemetry-go/issues/1675#issuecomment-793550867 by @matej-g.

Oberon00:
> 2. Doing *** is anonymization which AFAIK according to GDPR is a kind of personal data processing. Not even touching this info makes sure that auth data won't be processed.

At the point where it reaches your instrumentation/exporter/... code it has already been processed (e.g. the HTTP request containing it has been parsed). IANAL, but I would be extremely surprised if the additional "processing" of doing anonymization would be a GDPR problem in itself (if the processing to get it to the point where you can anonymize it wasn't already).

veera83372:
I am fine option 2 as well but just want to clarify `***` i didn't mean process any credentials data. We have to generate URL string without creds anyway, What i am trying to say is just add `***:***@` before actual URL if creds are present. Don't need to do processing on data.

>
>     2. Doing `***` is anonymization which AFAIK according to GDPR is a kind of personal data processing. Not even touching this info makes sure that auth data won't be processed.
>

pellared:
@veera83372  I understand. My biggest concern is:

>If a malicious user accesses this data, he would gain knowledge that could help him in making deeper attacks. Removing it completely reduces the attack surface.

IANAL, but I guess that from a GDPR perspective, not persisting the data (or not passing data further) is better (safer) than anonymization.

pellared:
@Oberon00 @veera83372 

> But my understanding of the *** suggestion was to actually replace the user and password with the string ***:***, discarding the original data, leaving only the info that user+password was there at all but not what it was.

This is what `anonymization` is. It leaves traces that there was "something"😉 Malicious user could dig deeper to try to get this info from other sources. Defence in depth.

PR #1503, opened 2021-03-04
Revert "Kick off the ownership clarification"

bogdandrutu:
Reverts open-telemetry/opentelemetry-specification#1494

Main reasons:

* Owner is a new concept undefined in otel org.
  * How do you become owner?
* Domain Expert in not defined as well, and it is not clear what are the rules to be one.

We do have concept of maintainers approvers and the right set of people are part of these groups based on clear rules. Why do we need more roles?

**Update:**
If we want to have these new roles we need to first define them then we can add it to every document/area.

### DIFF
--- a/specification/metrics/api.md
+++ b/specification/metrics/api.md
@@ -2,16 +2,6 @@
 
 **Status**: [Experimental](../document-status.md)
 
-**Owner:**
-
-* [Reiley Yang](https://github.com/reyang)
-
-**Domain Experts:**
-
-* [Bogdan Brutu](https://github.com/bogdandrutu)
-* [Josh Suereth](https://github.com/jsuereth)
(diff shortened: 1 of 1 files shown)


SergeyKanzhelev:
@reyang and @rakyll you wanted to start using the same approach on other files. WDYT?

reyang:
Discussed during the maintainers meeting:
1. Scope the change to the specs document that are "Experimental" (e.g. don't touch "Stable" docs).
2. @reyang to send a PR clarifying the owners / domain experts.
3. 1) does not have to block on 2, they can make progress in parallel
4. No need to rollback the change since metrics API is "Experimental"

carlosalberto:
Ping @reyang

tigrannajaryan:
What's the plan with this? Was the decision made to go forward with owners or revert it?

PR #1510, opened 2021-03-06
Initial overview w/ text from gdoc draft

jmacd:
Fixes #

## Changes

Please provide a brief description of the changes here. Update the
`CHANGELOG.md` for non-trivial changes. If `CHANGELOG.md` is updated,
also be sure to update `spec-compliance-matrix.md` if necessary.

Related issues #

Related [oteps](https://github.com/open-telemetry/oteps) #

PR #1512, opened 2021-03-08
Initial cut at migrating jmacd's datamodel document into the spec

jsuereth:
From Task: https://github.com/open-telemetry/opentelemetry-specification/projects/3#card-56227942

- Imports an initial version of @jmacd's Metric DataModel document, which elucidates a LOT of important points/goals in the metric data model over the existing specification.
- Adds TODO sections for discussion beyond the initial "Events" - "Data Model" - "Time Series" fragmentation.
- Updates verbiage in root specification around the metrics data model.

Note: This PR is 100% focused on framing the problem with  the 3 models and separation of concerns, as well as defining some key use cases for evaluating the model.

Replaces #1510 with permission to do so.


aabmass:
Original doc: https://docs.google.com/document/d/1lwLZlQ2RttzZ60N9xhJuJlivs3m5kFjUEZgrfpaF09c/edit#

specification/metrics/datamodel.md
```diff
@@ -0,0 +1,232 @@
...
+
+In this low-level metrics data model, a Timeseries is defined by an entity
+consisting of several metadata properties:
+
+- Metric name and description
+- Label set
+- Kind of point
+- Unit of measurement
+
+The primary data of each timeseries are ordered (timestamp, value) points, for
```
aabmass, line 168: Is this referring to a general timeseries model or *the* timeseries model of OTel?

[This part of the doc](https://docs.google.com/document/d/1lwLZlQ2RttzZ60N9xhJuJlivs3m5kFjUEZgrfpaF09c/edit#heading=h.eh4o3y81nl8s) talks about resource uniqueness separating timeseries:
> [Presumption of a Single-Writer] requirement states that Timeseries are logically expected to have a single writer at a time, to ensure that the mapping from OTLP into Timeseries is correct.  Generally, this requirement suggests that all writers should have at least one unique resource attribute or attribute value.

But its not mentioned here as "logical" metadata

specification/metrics/datamodel.md
```diff
@@ -0,0 +1,232 @@
...
+detail as we link this model with the other two.
+
+### Timeseries Model
+
+In this low-level metrics data model, a Timeseries is defined by an entity
+consisting of several metadata properties:
+
+- Metric name and description
+- Label set
+- Kind of point
```
aabmass, line 165: Just because "kind" is a bit vague

```suggestion
- Kind of point (integer, floating point, etc)
```

specification/metrics/datamodel.md
```diff
@@ -0,0 +1,232 @@
...
+
+1. Counter (Monotonic, cumulative)
+2. Gauge
+3. Histogram
+
+This model may be viewed as an idealization of
+[Prometheus Remote Write](https://docs.google.com/document/d/1LPhVRSFkGNSuU1fBd81ulhsCPR4hkSZyyBj1SZ8fWOM/edit#heading=h.3p42p5s8n0ui).
+Like that protocol, we are additionally concerned with knowing when a point
+value is defined, as compared with being implicitly or explicitly absent. There
+is no delta temporality in the timeseries model. To precisely define presence
```
aabmass, line 179: > There is no delta temporality in the timeseries model

Is "the timeseries model" specifically referring to PRW's timeseries model? Or saying there is no such thing as delta in a general timeseries model (and why not)?

specification/metrics/datamodel.md
```diff
@@ -0,0 +1,232 @@
...
+    - Collector computes “up”, staleness marker
+    - Collector applies a distinct external label
+5. OTel collector receives Statsd and exports OTLP
+    - With delta temporality: stateless collector
+    - With cumulative temporality: stateful collector
+6. OTel SDK exports directly to 3P backend
+
+These are considered the "core" use-cases used to analyze tradeoffs and design
+decisions within the metrics data model.
+
```
hdost, line 119: As mentioned in the Metrics Model Meeting If there's any Non Cases.

jsuereth: Added

specification/metrics/datamodel.md
jsuereth: This is still talking about the event model, added that explcitly.  

Single-Writer isn't written yet, but I hope it clarifies this a bit more (as the verbiage aludes to)

specification/metrics/datamodel.md
jsuereth: The first sentence explicitly calls out it's an "idealized version of PRW". 


Regarding whether delta temporality should be in our timeseries model, I'll let @jmacd comment more here.   My assumption is we're sticking to PRW's model to ensure compatibility here as our Event model + TimeSeries model MAY need to be subsets of other Event/TimeSeries models for clear definitions.

specification/metrics/datamodel.md
```diff
@@ -0,0 +1,232 @@
...
+standard metric data transformations for each of the model’s basic points while
+ensuring that the result carries the intended meaning.
+
+As in OpenCensus Metrics, metrics data can be transformed into one or more
+Views, just by selecting the aggregation interval and the desired dimensions.
+One stream of OTLP data can be transformed into multiple timeseries outputs by
+configuring different Views, and the required Views processing may be applied
+inside the SDK or by an external collector.
+
+### Example Use-cases
```
hdost, line 86: Btw, when I was speaking of use cases. This is one of the examples where I was thinking there could be some expansion: https://youtu.be/tCqTHDI8NZ4?t=1490

Specifically where josh mentioned equivalency and a user case like that. Is that a case we care about or not. 

The things below seem like requirements of the implementation rather than a use case which I suppose are two side of the same coin, but none the less. Having some use cases which demonstrate what is and is not supported might help when having some of these conversations.

Example could be:

User wants to track the latency of their HTTP service. The expectation would be that a Histogram is used to represent this information.  This provides the user a snapshot of the instance, but allows for spacial and temporal aggregation of latencies to obtain quantile summaries across multiple instances of the running application.

I think my example is a bit of a blend of "recommendation" and "use case", but none the less rather than saying "some users do X" the question that instantly comes to mind is "why?"

specification/metrics/datamodel.md
```diff
@@ -0,0 +1,232 @@
...
+
+Popular existing metrics data formats can be unambiguously translated into the
+OpenTelemetry data model for metrics, without loss of semantics or fidelity.
+Translation from the Prometheus and Statsd exposition formats is explicitly
+specified.
+
+The data model specifies a number of semantics-preserving data transformations
+for use on the collection path, supporting flexible system configuration. The
+model supports reliability and statelessness controls, through the choice of
+cumulative and delta transport. The model supports cost controls, through
```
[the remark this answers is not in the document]
jmacd: Please comment in https://github.com/open-telemetry/opentelemetry-proto/issues/274

specification/metrics/datamodel.md
```diff
@@ -0,0 +1,246 @@
...
+      distinct `service.instance.id` values)
+    - Collector outputs to an OTLP destination
+4. Pool of OTel collectors receive OTLP and export Prometheus Remote Write
+    - Collector joins service discovery with metric resources
+    - Collector computes “up”, staleness marker
+    - Collector applies a distinct external label
+5. OTel collector receives Statsd and exports OTLP
+    - With delta temporality: stateless collector
+    - With cumulative temporality: stateful collector
+6. OTel SDK exports directly to 3P backend
```
jmacd, line 115: 3P == PRW?

specification/metrics/datamodel.md
jmacd: @hdost In the recording @jsuereth is talking about equivalency between floating points and integers. Are you asking about / recommending that the data model support integer values as well as floating point values?

specification/metrics/datamodel.md
jmacd: The way I was thinking about it, the Delta temporality data points of OTLP can't be mapped directly into the timeseries model, without them first going through Delta-to-Cumulative conversion.  Delta data points don't have a well-defined value at a point in time, which is somehow a key property of the timeseries.

jmacd (approved):

specification/metrics/datamodel.md
jsuereth: No, this is a question I was asking myself.  What API should we give to `Exporters` in the SDK.   Should we try to support the "same" API for Metric Exporter in Go SDK as in the Collector?

Without lots of counter evidence, I think using this DataModel as the API for "Metric Exporter" would be a good idea for many reasons.

specification/metrics/datamodel.md
aabmass: > Delta data points don't have a well-defined value at a point in time

@jmacd I don't completely understand. I think you're getting at how the "actual" value at t<sub>i</sub> = Sum(deltas from t<sub>0</sub>..t<sub>i</sub>) and so the value isn't directly in the timeseries. Couldn't you make the same argument for a cumulative timeseries that has had a reset and the value went back to zero? Or if you are reporting deltas for a rate and that's all you care about, the deltas are conceptually the "actual" value in that case.

specification/metrics/datamodel.md
```diff
@@ -0,0 +1,246 @@
...
+this list is not exhaustive, it is meant to be representative of the scope and
+breadth of OTel metrics usage.
+
+1. OTel SDK exports 10 second resolution to a single OTel collector, using
+  cumulative temporality for a stateful client, stateless server:
+    - Collector passes-through original data to an OTLP destination
+    - Collector re-aggregates into longer intervals without changing dimensions
+    - Collector re-aggregates into several distinct views, each with a subset of
+      the available dimensions, outputs to the same destination
+2. OTel SDK exports 10 second resolution to a single OTel collector, using delta
```
reyang, line 98: I don't know if this is a common case and if we want to mention here (or the SDK):

OTel SDK exports both 10 seconds resolution (e.g. CPU, request latency) and 15 minutes resolution (e.g. room temperature) to a single OTel Collector.

specification/metrics/datamodel.md
```diff
@@ -0,0 +1,246 @@
...
+
+This specification uses as its foundation a
+[Metrics API consisting of 6 model instruments](api.md), each having distinct
+semantics, that were prototyped in several OpenTelemetry SDKs between July 2019
+and June 2020. The model instruments and their specific use-cases are meant to
+anchor our understanding of the OpenTelemetry data model and are divided into
+three categories:
+
+- Synchronous vs. Asynchronous. The act of calling a Metrics API in a
+  synchronous context means the application calls the SDK, typically having
```
reyang, line 154:
```suggestion
  synchronous context means the application/library calls the SDK, typically having
```

specification/metrics/datamodel.md
```diff
@@ -0,0 +1,246 @@
...
+  associated trace context and baggage; an Asynchronous instrument is called at
+  collection time, through a callback, and lacks context.
+- Adding vs. Grouping. Whereas adding instruments express a sum, grouping
+  instruments characterize a group of measurements. The numbers passed to adding
+  instruments define division, in the algebraic sense, while the numbers passed
+  to grouping instruments are generally not. Adding instrument values are always
+  parts of a sum, while grouping instrument values are individual measurements.
+- Monotonic vs. Non-Monotonic. The adding instruments are categorized by whether
+  the  derivative of the quantity they express is non-negative. Monotonic
```
reyang, line 163:
```suggestion
  the derivative of the quantity they express is non-negative. Monotonic
```

specification/metrics/datamodel.md
```diff
@@ -0,0 +1,246 @@
...
+respectively). The instruments and model use-cases will be described in greater
+detail as we link the event model with the other two.
+
+### Timeseries Model
+
+In this low-level metrics data model, a Timeseries is defined by an entity
+consisting of several metadata properties:
+
+- Metric name and description
+- Label set
```
reyang, line 178: Do we want to change this to attributes now or later? (not blocking)

reyang: Or probably consider "dimensions", since dimensions could come from the attributes/labels, baggage and context.

reyang (approved):

specification/metrics/datamodel.md
jmacd: Got it.  Inside the OTel-Go SDK prototype, there are different data types between the Accumulator and the Processor and between the Processor and the Exporter, but they both fundamentally map into the data model. The Accumulator-to-Processor data is delta in nature, the Processor-to-Exporter data is whatever-the-exporter-wants.

specification/metrics/datamodel.md
jmacd: I would be happy to apply in another PR, but yes I think we all agree to the label->attribute rename.

specification/metrics/datamodel.md
jmacd: Thanks for helping clarify the question!

> Or if you are reporting deltas for a rate and that's all you care about, the deltas are conceptually the "actual" value in that case.

There's a problem with calling deltas actual values, though, since we still have to divide by the width of the interval to make a meaningful rate. I think we should replace "delta data points don't have well-defined value at a point in time" with something more clear. 

Maybe: a metric stream of delta data points defines time-interval values, not point-in-time values.

jmacd (approved):

specification/metrics/datamodel.md
jmacd: I hope we can leave a few things under-specified for this first few pages of the document. Our goal is to continue reviewing small-to-medium size documents.

specification/metrics/datamodel.md
aabmass: > Maybe: a metric stream of delta data points defines time-interval values, not point-in-time values.

👍 that sounds good to me

aabmass (approved):

specification/metrics/datamodel.md
jsuereth: Let's do it in a follow up PR based on the discussions we had today.

specification/metrics/datamodel.md
jsuereth: added

specification/metrics/datamodel.md
jsuereth: Ok, using that phrasing.

specification/metrics/datamodel.md
jsuereth: Ok, so do you disagree that if we can make Processor-to-Exporter data model be a subset of this one (or the same) then it would simplify both OTLP Exporter *and* the need to write different exporters between the OT Collector and SDK?

What's the actionable change here.

jmacd: I agree with this in principle.

In practice, I've seen that when we use the protocol message (a.k.a. the "DTO") as an internal data type it leads to practical problems (either performance or size of dependencies). As a specification, absolutely makes sense to write it this way. The OTel-Go implements three significant processor facilities, two of which transform OTLP to OTLP: (1) it can strip labels, which results in the natural aggregation being applied if configured, (2) it converts delta-to-cumulative unless configured to be "stateless", (3) it can keep a memory of all OTLP points ever written, including from prior collection intervals.

specification/metrics/datamodel.md
```diff
@@ -99,20 +99,23 @@ breadth of OTel metrics usage.
...
+5. Pool of OTel collectors receive OTLP and export Prometheus Remote Write
     - Collector joins service discovery with metric resources
     - Collector computes “up”, staleness marker
     - Collector applies a distinct external label
-5. OTel collector receives Statsd and exports OTLP
+6. OTel collector receives Statsd and exports OTLP
     - With delta temporality: stateless collector
     - With cumulative temporality: stateful collector
-6. OTel SDK exports directly to 3P backend
+7. OTel SDK exports directly to 3P backend
```
jmacd, line 118: (Is "3P" defined somewhere?)

jmacd (approved):

specification/metrics/datamodel.md
```diff
@@ -0,0 +1,249 @@
...
+Comparing the OpenTelemetry and Timeseries data models, OTLP carries an
+additional kind of point. Whereas an OTLP Monotonic Sum point translates into a
+Timeseries Counter point, and an OTLP Histogram point translates into a
+Timeseries Histogram point, there are two OTLP data points that become Gauges
+in the Timeseries model: the OTLP Non-Monotonic Sum point and OTLP Gauge point.
+
+The two points that become Gauges in the Timeseries model are distinguished by
+their built in aggregate function, meaning they define re-aggregation
+differently. Sum points combine using addition, while Gauge points combine into
+histograms.
```
bogdandrutu, line 223: Can be last value as well.

specification/metrics/datamodel.md
bogdandrutu: I would replace with the explicit text :)

specification/metrics/datamodel.md
```diff
@@ -0,0 +1,249 @@
...
+Here are a set of use cases that, while won't be outright unsupported, are not
+in scope for key design decisions:
+
+- Using OTLP as an intermediary format between two non-compatible formats
+  - Importing [statsd](https://github.com/statsd/statsd) => Prometheus PRW
+  - Importing [collectd](https://collectd.org/wiki/index.php/Binary_protocol#:~:text=The%20binary%20protocol%20is%20the,some%20documentation%20to%20reimplement%20it)
+    => Prometheus PRW
+  - Importing Prometheus endpoint scrape => [statsd push | collectd | opencensus]
+  - Importing OpenCensus "oca" => any non OC or OTel format
+- TODO: define others.
```
bogdandrutu, line 135: I think I fail a bit to understand this:

It is in scope to have `Prometheus -> OTLP -> PRW` and it is in scope to have `StatsD -> OTLP` then why is not in scope to have `StatsD -> OTLP -> PRW`? mathematically if the translation function `OTLP -> PRW` is defined and `StatsD -> OTLP` is defined then I can compose them to get `StatsD -> OTLP -> PRW`

bogdandrutu (approved):
Overall LGTM, I think we can iterate over some details later.

specification/metrics/datamodel.md
jmacd: I think `StatsD -> OTLP -> PRW` is in-scope. We haven't added support for raw sampled metric events, which is how they appear to the StatsD receiver, so somewhere in that picture is a stateful conversion to OTLP(cumulative) which is not a difficult task, just requires memory.

I believe there is an additional concept that belongs in the specification, and I'm sorry to introduce it here, that I think of as a counterpart to the Single-Writer definition we have discussed. It can be called a Single-Reader property, and it is required to perform a correct delta-to-cumulative translation.

*Single-Reader*: Having access to the whole stream of data for a metric. This happens when an OTel collector is configured as a per-node agent and every process on the node reports through that agent. This happens when an OTel collector pool runs the Kafka exporter and Kafka receiver with appropriate configuration.

`StatsD -> OTLP -> PRW` should be in-scope for a single-node deployment, and it ought to behave just as https://github.com/prometheus/statsd_exporter would.

specification/metrics/datamodel.md
jmacd: I am not sure and have an opinion. I suspect it is rare that a user will configure last-value selection across a spatial dimension in cases where they expect more than one logical series, meaning cases where there is not expected overlap. If we expect more than one value at an instant in time, last-value aggregation is not very meaningful and I think the default should be to compute a histogram.

jmacd (approved):

specification/metrics/datamodel.md
jsuereth: @bogdandrutu There's a lot of nuance to Statsd => PRW.   While TECHNICALLY you can shove bits into PRW, when I say "compatibility" i mean the way we defined it for Prometheus:  "A user of prometheus shouldn't be able to tell if OTLP was used to transport prometheues style metric to their backend".   Similarly, for users of Statsd today, we should HONOR the statsd conventions and formats from statsd => PRW.   The reason Statsd => PRW is out of scope is because of all the non-data-related nuance (e.g. naming conventions, namespace options etc. that a statsd user would expect).

To be frank:  I don't think it's OTLP's job to define how Statsd maps into PRW, and we shouldn't sacrifice the design of OTLP for that use case.    We *can* attempt to provide the best translation possible, but it should not "block" or otherwise inhibit progress.   There's a difference between "out of scope" and "can't do".   What I'm suggesting is we won't prioritize or make a ton of effort to solve statsd => PRW bugs that are inherent to the impedance mismatch of those two models.


@jmacd I'm actually tying your "Single-Reader" into the "SingleWriter rule.    Specifically the way I'm phrasing Single-WRiter in the not-yet-public PR is that any given timeseries should originate from a --SingleWriter.  This includes *aggregation* time series, meaning if the collector is manipulating a metrics, it becomes the new single-writer of that aggregation.  It is also required for delta-to-cumulative manipulation (which in my opinion, that's a new series of data).   To some extent I *think* this nuance is just terminology and we need to make sure we COVER the instance of when it's OK to do "Delta-Cumulative" and what that implies on your telemetry-pipeline architecture.

MrAlias (approved):

specification/metrics/datamodel.md
jmacd: :100:
 
 > It is also required for delta-to-cumulative manipulation (which in my opinion, that's a new series of data).

Yes, that's how I've been thinking about it too. It's permitted to output the same metric name, in this case, because the incoming data is "terminal".

PR #1514, opened 2021-03-08
Add semantic conventions spec

rakyll:
This is a follow up from the spec SIG model where we
discussed we should ask each langauge to autogenerate
the semantic conventions keys from the YAML files.

Adding a spec specifically for semantic conventions to
capture the implementation requirementes from the languages.


rakyll:
Will merge it to the existing semantic conventions section.

Issue #1497, opened 2021-03-03
Stabilizing service.name


anuraaga:
Currently all attributes are experimental, for example we see the Experimental label.

https://www.github.com/open-telemetry/opentelemetry-specification/tree/main/specification%2Fresource%2Fsemantic_conventions%2FREADME.md

But `service.name` is required to use Zipkin or Jaeger - those exporters are also effectively experimental because of this dependency. Is it possible to mark this attribute stable? I guess we should add support to the yaml generator for stable attributes if so.

anuraaga:
For reference this came up because we don't have a stable way is configuring the service name in Java

https://github.com/open-telemetry/opentelemetry-java/issues/2973

Oberon00:
👍 for making it stable. But we should consider it together with `service.namespace`, as they are closely related. The instance ID stuff should not be made stable yet though, see #1034.

rakyll:
+1 to considering to stabilize the name with namespace. We are thinking about representing service identifiers as a first-class thing in the backends. We should think about the service identity as namespace + name.

anuraaga:
@thisthat Is this something we can add to the yaml tool?

thisthat:
@anuraaga sorry for the late reply! Sure, this can be added to the tool! Currently, we support only **deprecated** attributes. It would be easy to add other markers!

jsuereth:
Fixed by #3202

PR #1515, opened 2021-03-08
Mention that semantic conventions should be autogenerated

rakyll:
Adding a section to semantic conventions to capture the implementation
requirements from the languages.

This is a follow up from the spec SIG model where we
discussed we should ask each language to autogenerate
the semantic conventions keys from the YAML files.

### DIFF
--- a/specification/overview.md
+++ b/specification/overview.md
@@ -77,7 +77,7 @@ The **Semantic Conventions** define the keys and values which describe commonly 
 
 Both the collector and the libraries SHOULD autogenerate semantic
 convention keys into constants (or language idomatic equivalent).
-The [YAML](../semantic_conventions) files MUST be used as the 
+The [YAML](../semantic_conventions/README.md) files MUST be used as the 
 source of truth for generation. Each language implementation SHOULD 
 provide language-specific support to the
 [code generator](https://github.com/open-telemetry/build-tools/tree/main/semantic-conventions#code-generator).
--- a/specification/overview.md
+++ b/specification/overview.md
@@ -75,6 +75,13 @@ The **Semantic Conventions** define the keys and values which describe commonly 
 * [Span Conventions](trace/semantic_conventions/README.md)
 * [Metrics Conventions](metrics/semantic_conventions/README.md)
 
+Both the collector and the libraries SHOULD autogenerate semantic
+convention keys into constants (or language idomatic equivalent).
+The [YAML](../semantic_conventions) files MUST be used as the 
+source of truth for generation. Each language implementation SHOULD 
+provide language-specific support to the
+[code generator](https://github.com/open-telemetry/build-tools/tree/main/semantic-conventions#code-generator).
+
 ### Contrib Packages
 
 The OpenTelemetry project maintains integrations with popular OSS projects which have been identified as important for observing modern web services.
--- a/specification/overview.md
+++ b/specification/overview.md
@@ -77,7 +77,7 @@ The **Semantic Conventions** define the keys and values which describe commonly 
 
 Both the collector and the client libraries SHOULD autogenerate semantic
 convention keys and enum values into constants (or language idomatic
-equivalent). Generated values shouldn't be distrubuted in stable packages
+equivalent). Generated values shouldn't be distributed in stable packages
 until semantic conventions are stable.
 The [YAML](../semantic_conventions) files MUST be used as the
 source of truth for generation. Each language implementation SHOULD
--- a/specification/overview.md
+++ b/specification/overview.md
@@ -75,9 +75,9 @@ The **Semantic Conventions** define the keys and values which describe commonly 
 * [Span Conventions](trace/semantic_conventions/README.md)
 * [Metrics Conventions](metrics/semantic_conventions/README.md)
 
-Both the collector and the libraries SHOULD autogenerate semantic
+Both the collector and the client libraries SHOULD autogenerate semantic
 convention keys into constants (or language idomatic equivalent).
-The [YAML](../semantic_conventions/README.md) files MUST be used as the
+The [YAML](../semantic_conventions) files MUST be used as the
 source of truth for generation. Each language implementation SHOULD
 provide language-specific support to the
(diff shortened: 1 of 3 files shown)


iNikem (approved):

specification/overview.md
```diff
@@ -75,6 +75,13 @@ The **Semantic Conventions** define the keys and values which describe commonly
 * [Span Conventions](trace/semantic_conventions/README.md)
 * [Metrics Conventions](metrics/semantic_conventions/README.md)
 
+Both the collector and the libraries SHOULD autogenerate semantic
+convention keys into constants (or language idomatic equivalent).
+The [YAML](../semantic_conventions/README.md) files MUST be used as the
```
arminru, line 80:
```suggestion
The [YAML](../semantic_conventions) files MUST be used as the
```
If you link to the folder itself one will see the subfolders containing the YAML files right away. These are not linked from the README.

specification/overview.md
```diff
@@ -75,6 +75,13 @@ The **Semantic Conventions** define the keys and values which describe commonly
 * [Span Conventions](trace/semantic_conventions/README.md)
 * [Metrics Conventions](metrics/semantic_conventions/README.md)
 
+Both the collector and the libraries SHOULD autogenerate semantic
```
arminru, line 78: I think we usually call them _OpenTelemetry language implementations_ - just "libraries" is ambiguous.

arminru, line 78: Please add an entry for this new SHOULD requirement to the spec compliance matrix so we can get an overview of which SIGs already implement this.

specification/overview.md
```diff
@@ -75,6 +75,13 @@ The **Semantic Conventions** define the keys and values which describe commonly
...
+convention keys into constants (or language idomatic equivalent).
+The [YAML](../semantic_conventions/README.md) files MUST be used as the
+source of truth for generation. Each language implementation SHOULD
+provide language-specific support to the
+[code generator](https://github.com/open-telemetry/build-tools/tree/main/semantic-conventions#code-generator).
```
arminru, line 83: The linked readme is updated and contains broken links. I opened https://github.com/open-telemetry/build-tools/issues/34 for that.

specification/overview.md
```diff
@@ -75,6 +75,13 @@ The **Semantic Conventions** define the keys and values which describe commonly
 * [Span Conventions](trace/semantic_conventions/README.md)
 * [Metrics Conventions](metrics/semantic_conventions/README.md)
 
+Both the collector and the libraries SHOULD autogenerate semantic
+convention keys into constants (or language idomatic equivalent).
```
SergeyKanzhelev, line 79: is it a good practice to expose them from API package for instrumentation libraries to consume? It may not be the good practice as semantic convention may be versioned and expanded on faster pace than API. Maybe we can note about it here

specification/overview.md
rakyll: We decided to use OpenTelemetry Client Libraries in the earlier SIG meeting. Making the change here and a large charge is coming.

specification/overview.md
rakyll: The other question what it means to have breaking changes in semantic conventions that are experimental and how it affects the client library versioning/releases.

What do you like me to note here? Something in the lines of versioning of the API should/shouldn't be affected by the semantic conventions until they are stable?

specification/overview.md
rakyll: I tried this and it broke the link checker. Let's try agiain.

rakyll (commented):
PTAL, this is good to review again.

specification/overview.md
SergeyKanzhelev: yes. And also that generating constants doesn't mean they have to be exposed as part of an API package. I know there is nothing imply that. But it may be read as such

specification/overview.md
arminru: Hm, docfx doesn't seem to like that. Let's try adding a `/`.

specification/overview.md
carlosalberto: @rakyll @SergeyKanzhelev is something still missing here?

carlosalberto (approved):

SergeyKanzhelev (approved):

SergeyKanzhelev:
PR is good as is. It may be worth it to add the note about the way to expose these constant though

specification/overview.md
```diff
@@ -75,6 +75,13 @@ The **Semantic Conventions** define the keys and values which describe commonly
 * [Span Conventions](trace/semantic_conventions/README.md)
 * [Metrics Conventions](metrics/semantic_conventions/README.md)
 
+Both the collector and the client libraries SHOULD autogenerate semantic
+convention keys into constants (or language idomatic equivalent).
+The [YAML](../semantic_conventions/README.md) files MUST be used as the
+source of truth for generation. Each language implementation SHOULD
+provide language-specific support to the
+[code generator](https://github.com/open-telemetry/build-tools/tree/main/semantic-conventions#code-generator).
```
anuraaga, line 83: Is it ok not to mention that stable and experimental attributes should be generated in a way indicating that, as per language idioms? e.g., in Java we would generate stable attributes into a stable package / artifact, and unstable separate.

anuraaga: One reason I ask is I consider the `+` on Java in this PR to not actually be ready yet since we need to figure out how to do this for people to use our attributes as per #1497

carlosalberto: @anuraaga Shall we go with your clarification as a follow-up? Otherwise I'm afraid this PR will get stuck longer than it needs to (unless you strongly feel we should clarify this **before** merging this PR).

anuraaga: I would be fine with merging as is if removing the compliance matrix for now since I foresee a follow-up making the `+` go away so it seems like just churn. Or otherwise think it should be added along with a blank matrix.

specification/overview.md
rakyll: Added a note.

rakyll (commented):
PTAL

anuraaga (approved):
Thanks!

bogdandrutu (approved):

PR #1516, opened 2021-03-08
Unify the terminology for OpenTelemetry libraries

rakyll:
Spec names the libraries either clients or libraries. Unifying the
terminology to use "library" because specs must be very consistent
about naming the same concept similarly to avoid confusion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Technical committee holds regular meetings, notes are held
 
 - [Overview](specification/overview.md)
 - [Glossary](specification/glossary.md)
-- [Versioning and stability for OpenTelemetry clients](specification/versioning-and-stability.md)
+- [Versioning and stability for OpenTelemetry Libraries](specification/versioning-and-stability.md)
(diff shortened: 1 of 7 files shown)


anuraaga (commented):
IIRC, @tedsuo meant to consolidate on client to reduce confusion with the library that is actually being instrumented.

tedsuo:
Just following up form today's meeting.

There was some confusion about the term "language libraries" and "OpenTelemetry libraries."

* Readers were confused about whether they referred to target libraries vs instrumentation libraries.
* Readers were confused about whether collector libraries were 

The term "OpenTelemetry Client" was added to refer collectively to the four components that are installed in the application: API, SDK, Instrumentation, and Plugins.

arminru:
@rakyll Could you please update your PR with the term that was agreed on in the meeting as stated by @tedsuo? Thanks!

rakyll:
I'll work on this when I have time, it'll require a pretty extensive change. Closing this for now.

PR #1520, opened 2021-03-09
Suggest the correct markdownlint CLI

rakyll:
So the developers can run `make markdownlint` to lint the files.


CONTRIBUTING.md
```diff
@@ -95,9 +95,9 @@ the box settings for this repository will be consistent.
 To check for style violations, use
 
 ```bash
-# Ruby and gem are required for mdl
-gem install mdl
-mdl -c .mdlrc .
+# npm is required to install markdownlint
+npm install -g markdownlint-cli
```
yurishkuro, line 99: I would suggest creating a package.json and installing locally instead of relying on global install.

rakyll: Developers are advanced enough to fallback to that option if they don't want to install global. This enables `make markdownlint` on a development machine which was the actual purpose of the change.

anuraaga: I don't think it has to block this change, but adding a docker image for markdownlint to the build-tools would be nice

jsuereth: +1 to the docker image idea.   Also, hope `make check` and `make fix` helps others :)

Does this tool have an auto-fix option?

jsuereth (approved):

rakyll:
Please feel free to follow up with a docker image and fix options.

reyang (approved):
LGTM.

CONTRIBUTING.md
```diff
@@ -95,9 +95,9 @@ the box settings for this repository will be consistent.
 To check for style violations, use
 
 ```bash
-# Ruby and gem are required for mdl
-gem install mdl
-mdl -c .mdlrc .
+# npm is required to install markdownlint
```
SergeyKanzhelev, line 98: nit

```suggestion
# npm is required to install markdownlint-cli
```

SergeyKanzhelev (approved):

anuraaga (approved):

arminru (approved):

CONTRIBUTING.md
bogdandrutu: We already have a make rule for this, what about `make install-markdownlint` or just `make install-tools`

rakyll:
PTAL, switched to `make install-markdownlint`.

arminru (approved):

jmacd:
(I was assigned this, and I was looking at the unresolved comment yesterday -- why it wasn't merged.)

PR #1530, opened 2021-03-10
Avoid using "infrastructure" terminology for all cloud

rakyll:
Infrastructure in cloud has a specific connotation for concepts
explained in Infrastructure-as-a-Service (IaaS). Avoid this
terminology because it's exclusive of some other services/platforms
that is represented by the "cloud" namespace.

Renaming "infrastructure_service" to "platform" also due to the new
key length limits we are considering and infrastructure_service is
a lengthy choice.


SergeyKanzhelev (approved):
good call

semantic_conventions/resource/cloud.yaml
```diff
@@ -42,7 +42,7 @@ groups:
         note: >
           AWS and Azure calls them availability zones or AZs.
         examples: ['us-central1-a', 'us-east-1c']
-      - id: infrastructure_service
+      - id: service
```
anuraaga, line 45: The word service comes up a lot in OTel and is very important in tracing so avoiding the same word would be ideal. But maybe there just isn't a better word - product? Would be good if anyone can help brainstorm.

arminru: I also think disambiguating it from the general "service" being monitored was the intention for choosing this name.
Therefore I prefer the current term better over just `service` but I agree with @rakyll's statements in the PR description. `product` is not perfect either but conveys rather well what this is about. Maybe someone has an even better idea 🙂

yurishkuro: Looking at the list of values, they all describe platforms on which "you can run your code". Definitely not "services".

How about `platform` or `execution_platform`?

arminru: `cloud.platform` would likely be mixed up with `cloud.provider` (AWS, Azure, GCP (Google Cloud **Platform**), ...).
`cloud.execution_platform` sounds alright, I wouldn't expect someone to specify `x64` here.

rakyll: I previously considered "platform" but I didn't think think platform is a good name either due to the connotation that platform is seen as a managed platform such as Heroku. Maybe I was overthinking. I can't think of a better name tbh.

"product" is probably the second best but it'd exclude some execution environments that wouldn't classify as a cloud product. For example, self managed knative is not a product.

Regarding to the use of service, the namespace helps ("cloud.service") but "service" is probably not a great solution because it excludes execution environments that are not a managed service.

rakyll:
PTAL

anuraaga (approved):

bogdandrutu (approved):

semantic_conventions/resource/cloud.yaml
```diff
@@ -42,7 +42,7 @@ groups:
         note: >
           Availability zones are called "zones" on Google Cloud.
         examples: ['us-east-1c']
-      - id: infrastructure_service
+      - id: platform
```
Oberon00, line 45: The only hits I get on google for "Cloud platform" is GCP as a whole...

Oberon00 (approved):

CHANGELOG.md
```diff
@@ -41,6 +41,7 @@ Google products under `cloud.infrastructure_service` ([#1496](https://github.com
 - `http.url` MUST NOT contain credentials ([#1502](https://github.com/open-telemetry/opentelemetry-specification/pull/1502))
 - Add `aws.eks.cluster.arn` to EKS specific semantic conventions ([#1484](https://github.com/open-telemetry/opentelemetry-specification/pull/1484))
 - Rename `zone` to `availability_zone` in `cloud` semantic conventions ([#1495](https://github.com/open-telemetry/opentelemetry-specification/pull/1495))
+- Rename `cloud.infrastructure_service` to `cloud.platform` ([#1530](https://github.com/open-telemetry/opentelemetry-specification/pull/1530))
```
bogdandrutu, line 44: Have we considered `cloud.paas`?

carlosalberto:
Ping @rakyll (it looks there are a pair of unaddressed comments)

CHANGELOG.md
rakyll: PaaS only means AppEngine, Lambda, Heroku not EC2, EKS, ECS, etc.

pmm-sumo:
BTW, since renaming is essentially a braking change, how should it work for specs v1.0+?

tigrannajaryan:
> BTW, since renaming is essentially a braking change, how should it work for specs v1.0+?

Hopefully once [Telemetry Schemas](https://github.com/open-telemetry/oteps/pull/152) are implemented renaming can be handled though them.

PR #1534, opened 2021-03-11
Define the fallback tracer name for invalid values.

### LINKED ISSUE #1472: Define the default value for null/empty instrumentation-name
Currently [TracerProvider.GetTracer()](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#get-a-tracer) is defined as:

> In case an invalid name (null or empty string) is specified, a working default Tracer implementation as a fallback is returned rather than returning null or throwing an exception.

However, different SIGs have different values, i.e.

* Java uses `unknown`
* JS uses the empty string
* Python uses `ERROR:MISSING MODULE NAME`

I'd suggest we define a SHOULD on this, so at least the default is (more or less) shared. Personally I'd go with `unknown` (although defaulting to "" probably would be helpful if we ever support a _default_ `Tracer`, i.e. `TracerProvider.getDefaultTracer()`)

carlosalberto:
Fixes #1472 

Keeps the original invalid value (to be handled by the backends instead), while asking implementations to log this as an error.

### DIFF
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -109,7 +109,7 @@ This API MUST accept the following parameters:
   In that scenario, the `name` denotes a module name or component name within that library
   or application.
   In case an invalid name (null or empty string) is specified, a working
-  default Tracer implementation MUST be returned as a fallback rather than returning
+  Tracer implementation MUST be returned as a fallback rather than returning
   null or throwing an exception, its `name` property SHOULD keep the original invalid value,
   and a message reporting that the specified value is invalid SHOULD be logged.
   A library, implementing the OpenTelemetry API *may* also ignore this name and
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -109,8 +109,10 @@ This API MUST accept the following parameters:
   In that scenario, the `name` denotes a module name or component name within that library
   or application.
   In case an invalid name (null or empty string) is specified, a working
-  default Tracer implementation as a fallback is returned rather than returning
-  null or throwing an exception.
+  default Tracer implementation as a fallback MUST be returned rather than returning
+  null or throwing an exception, and `name` SHOULD be the set to the
+  `<INVALID INSTRUMENTATION NAME PROVIDED>` literal, in order to signal the
+  specified value is invalid.
   A library, implementing the OpenTelemetry API *may* also ignore this name and
   return a default instance for all calls, if it does not support "named"
   functionality (e.g. an implementation which is not even observability-related).
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -111,7 +111,7 @@ This API MUST accept the following parameters:
   In case an invalid name (null or empty string) is specified, a working
   default Tracer implementation MUST be returned as a fallback rather than returning
   null or throwing an exception, and `name` SHOULD be the set to the
-  `<INVALID INSTRUMENTATION NAME PROVIDED>` literal, in order to signal the
+  "`<INVALID INSTRUMENTATION NAME PROVIDED>`" literal, in order to signal that the
   specified value is invalid.
   A library, implementing the OpenTelemetry API *may* also ignore this name and
   return a default instance for all calls, if it does not support "named"
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -110,8 +110,8 @@ This API MUST accept the following parameters:
   or application.
   In case an invalid name (null or empty string) is specified, a working
   default Tracer implementation MUST be returned as a fallback rather than returning
-  null or throwing an exception, and `name` SHOULD keep the original invalid value,
-  with a log message reporting that the specified value is invalid.
+  null or throwing an exception, its `name` property SHOULD keep the original invalid value,
+  and a message reporting that the specified value is invalid SHOULD be logged.
   A library, implementing the OpenTelemetry API *may* also ignore this name and
   return a default instance for all calls, if it does not support "named"
   functionality (e.g. an implementation which is not even observability-related).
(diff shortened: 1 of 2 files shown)


specification/trace/api.md
```diff
@@ -109,8 +109,10 @@ This API MUST accept the following parameters:
   In that scenario, the `name` denotes a module name or component name within that library
   or application.
   In case an invalid name (null or empty string) is specified, a working
-  default Tracer implementation as a fallback is returned rather than returning
-  null or throwing an exception.
+  default Tracer implementation as a fallback MUST be returned rather than returning
```
arminru, line 112:
```suggestion
  default Tracer implementation MUST be returned as a fallback rather than returning
```

specification/trace/api.md
```diff
@@ -109,8 +109,10 @@ This API MUST accept the following parameters:
...
+  null or throwing an exception, and `name` SHOULD be the set to the
+  `<INVALID INSTRUMENTATION NAME PROVIDED>` literal, in order to signal the
```
arminru, line 114:
```suggestion
  "`<INVALID INSTRUMENTATION NAME PROVIDED>`" literal, in order to signal that the
```

arminru (approved):
Please add a changelog entry. Thanks!

specification/trace/api.md
reyang, line 114: `name` is not exposed by Tracer https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#tracer, who would be consuming this value (how do we "signal")?

bogdandrutu (commented):
Personally I would prefer to use "" as a fallback and treat that as `invalid value`. That way we are consistent with our protocol as well.

specification/trace/api.md
arminru: Both the `SpanProcessor`s and the `SpanExporter`s receive a read/write or readable span representation, which provides access to the `InstrumentationLibrary` containing the `name` and `version` provided at tracer creation:
https://github.com/open-telemetry/opentelemetry-specification/blob/d89f446368307fc80c3fad76ad17d5fbef1059f9/specification/trace/sdk.md#L71-L76

yurishkuro:
I think it's worth comparing this with a missing service name, which is quite bad and we already have some fallback value for it. A missing/blank instrumentation library feels very much meh, i.e. a blank string would do just fine as a fallback with very little downside.

carlosalberto:
> I think it's worth comparing this with a missing service name

Indeed, I had considered this as well, and was wondering whether we should try to go with "unknown_name" or a similar one, to be in line with that a missing service name ends up doing ;)

reyang (approved):

Oberon00:
The now says that a missing service name is not an error (just an unknown service name), while a missing instrumentation name is an API usage error. Note that a missing service name naturally appears by doing nothing in particular while you have to decide to pass an empty string or null to GetTracer to get an invalid instrumentation name.

specification/trace/api.md
```diff
@@ -109,8 +109,10 @@ This API MUST accept the following parameters:
   In that scenario, the `name` denotes a module name or component name within that library
   or application.
   In case an invalid name (null or empty string) is specified, a working
-  default Tracer implementation as a fallback is returned rather than returning
-  null or throwing an exception.
+  default Tracer implementation MUST be returned as a fallback rather than returning
+  null or throwing an exception, and `name` SHOULD be the set to the
+  "`<INVALID INSTRUMENTATION NAME PROVIDED>`" literal, in order to signal that the
```
Oberon00, line 114:
```suggestion
  string `<INVALID INSTRUMENTATION NAME PROVIDED>`, in order to signal that the
```

Oberon00 (approved):

arminru:
@carlosalberto Please add a changelog entry that this fallback is now specified for SDK implementations.

bogdandrutu:
@Oberon00 @carlosalberto can you please document why empty string is not good? When it comes to receiving empty string from a source what does that mean?

I would vote for empty string since it means less unnecessary bytes on the wire and this case has to be treated by the receiver anyway.

carlosalberto:
From the discussion today: we want to allow the user passing `null` or empty strings, as this is something that can/should be handled by the backend (as backends must **very likely** already handle the null value case), while logging (even if only once) about the invalid `name` value.

carlosalberto:
@open-telemetry/specs-approvers Please re-review ;)

specification/trace/api.md
```diff
@@ -109,8 +109,9 @@ This API MUST accept the following parameters:
   In that scenario, the `name` denotes a module name or component name within that library
   or application.
   In case an invalid name (null or empty string) is specified, a working
-  default Tracer implementation as a fallback is returned rather than returning
-  null or throwing an exception.
+  default Tracer implementation MUST be returned as a fallback rather than returning
```
yurishkuro, line 112: can we remove the word "default"? It's not clear what it means, and might be confused with "default no-op implementation" that is returned by the API when no SDK is installed.

specification/trace/api.md
```diff
@@ -109,8 +109,9 @@ This API MUST accept the following parameters:
...
+  null or throwing an exception, and `name` SHOULD keep the original invalid value,
+  with a log message reporting that the specified value is invalid.
```
yurishkuro, line 114:
```suggestion
  null or throwing an exception, its `name` property SHOULD keep the original invalid value,
  and a message reporting that the specified value is invalid SHOULD be logged.
```

yurishkuro (approved):

SergeyKanzhelev (approved):

SergeyKanzhelev:
@carlosalberto pls resolve couple suggestions from @yurishkuro before merging. Thank you!

specification/trace/api.md
carlosalberto: Done, removed the `default` word ;)

yurishkuro (approved):

PR #1535, opened 2021-03-11
Release v1.1.0

bogdandrutu:
Per the decision we should do a new release of the specs every month to ensure we include all the new changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ release.
 
 ## v1.1.0 (2021-03-11)
 
-- Adds note that implementations can ignore links with invalid SpanContext([#1492](https://github.com/open-telemetry/opentelemetry-specification/pull/1492))
+- Implementations can ignore links with invalid SpanContext([#1492](https://github.com/open-telemetry/opentelemetry-specification/pull/1492))
 - Adds `none` as a possible value for OTEL_TRACES_EXPORTER and OTEL_METRICS_EXPORTER to disable export
   ([#1439](https://github.com/open-telemetry/opentelemetry-specification/pull/1439))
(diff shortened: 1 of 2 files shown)


cijothomas (approved):

carlosalberto (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ release.
 
 ## Unreleased
 
+## v1.1.0 (2021-03-11)
+
 - Adds note that implementations can ignore links with invalid SpanContext([#1492](https://github.com/open-telemetry/opentelemetry-specification/pull/1492))
```
yurishkuro, line 12:
```suggestion
- Implementations can ignore links with invalid SpanContext([#1492](https://github.com/open-telemetry/opentelemetry-specification/pull/1492))
```

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ release.
...
 - Adds `none` as a possible value for OTEL_TRACES_EXPORTER and OTEL_METRICS_EXPORTER to disable export
```
yurishkuro, line 13:
```suggestion
- Add `none` as a possible value for OTEL_TRACES_EXPORTER and OTEL_METRICS_EXPORTER to disable export
```

yurishkuro (commented):
I would suggest that if we want to do regular releases, then the changelog should be better structured, at minimum broken down into subsections for different signals & semantic conventions. Right how all items are in a single pile.

I would also suggest setting higher bar for the PR titles, which translate into changelog entries (imperative tense, etc. Cf. https://chris.beams.io/posts/git-commit/#seven-rules), so that they all read consistently.

bogdandrutu:
@yurishkuro I gave a try to split by signal and semantic conventions

yurishkuro (approved):
thanks!

PR #1539, opened 2021-03-12
Clarify the requirement about newly created Span

### LINKED ISSUE #1508: Clarification needed for Span Creation section of Tracer API
The [Span Creation](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.0.1/specification/trace/api.md#span-creation) says:

> Span creation MUST NOT set the newly created Span as the currently active Span by default, but this functionality MAY be offered additionally as a separate operation.

This requirement is more applicable to languages with implicit contexts but not suitable for languages with explicit contexts.

For some languages, like Go, we use explicit and immutable `Context` so the SDK returns a new `Context` with the new `Span` set as active. Users can choose which of the `Context` as the parent `Span` while using such `Context` instances to not cause an arbitrary `Span` becoming a parent.

Also, explicit set the currently active `Span` to `Context` seems wordy while the `Context` instance is explicit.
For instance:
```go
newCtx, span = CreateSpan(ctx, "span-name")
newCtx = SetActiveContext(newCtx, span)

newCtx2, span = CreateSpan(newCtx, "span-name")
newCtx2 = SetActiveContext(newCtx2, span)
```
Rather than:
```go
newCtx, span = CreateSpan(ctx, "span-name")

newCtx2, span = CreateSpan(newCtx, "span-name")
```

Prior conversations in [Slack](https://cloud-native.slack.com/archives/C01N7PP1THC/p1614737963034700)

XSAM:
The requirement that `Span` creation can't set the newly created `Span` in `Context` is not suitable for languages that use explicit `Context`, such as Go.

This corrects that ambiguity.

Resolves #1508

### DIFF
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -329,8 +329,8 @@ directly. All `Span`s MUST be created via a `Tracer`.
 There MUST NOT be any API for creating a `Span` other than with a [`Tracer`](#tracer).
 
 `Span` creation MUST NOT set the newly created `Span` as the
-active `Span` in the implicit `Context` by default, but this functionality MAY be offered additionally
-as a separate operation. Languages that use explicit `Context` are not suitable for this requirement.
+active `Span` in the [implicit `Context`](#context-interaction) by default, but this functionality MAY be offered additionally
+as a separate operation in languages with implicit context propagation.
 
 The API MUST accept the following parameters:
 
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -328,9 +328,9 @@ directly. All `Span`s MUST be created via a `Tracer`.
 
 There MUST NOT be any API for creating a `Span` other than with a [`Tracer`](#tracer).
 
-`Span` creation MUST NOT set the newly created `Span` as the currently
-active `Span` by default, but this functionality MAY be offered additionally
-as a separate operation.
+`Span` creation MUST NOT set the newly created `Span` as the
+active `Span` in the implicit `Context` by default, but this functionality MAY be offered additionally
+as a separate operation. Languages that use explicit `Context` are not suitable for this requirement.
 
 The API MUST accept the following parameters:
 
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -328,9 +328,10 @@ directly. All `Span`s MUST be created via a `Tracer`.
 
 There MUST NOT be any API for creating a `Span` other than with a [`Tracer`](#tracer).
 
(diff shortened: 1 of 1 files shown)


specification/trace/api.md
```diff
@@ -328,9 +328,9 @@ directly. All `Span`s MUST be created via a `Tracer`.
 
 There MUST NOT be any API for creating a `Span` other than with a [`Tracer`](#tracer).
 
-`Span` creation MUST NOT set the newly created `Span` as the currently
-active `Span` by default, but this functionality MAY be offered additionally
-as a separate operation.
+`Span` creation MUST NOT set the newly created `Span` as the
+active `Span` in the implicit `Context` by default, but this functionality MAY be offered additionally
```
arminru, line 332:
```suggestion
active `Span` in the [implicit `Context`](#context-interaction) by default, but this functionality MAY be offered additionally
```

specification/trace/api.md
```diff
@@ -328,9 +328,9 @@ directly. All `Span`s MUST be created via a `Tracer`.
...
+as a separate operation. Languages that use explicit `Context` are not suitable for this requirement.
```
arminru, line 333:
```suggestion
as a separate operation in languages with implicit context propagation.
```

arminru (approved):

bogdandrutu (approved):

specification/trace/api.md
```diff
@@ -328,9 +328,9 @@ directly. All `Span`s MUST be created via a `Tracer`.
 
 There MUST NOT be any API for creating a `Span` other than with a [`Tracer`](#tracer).
 
-`Span` creation MUST NOT set the newly created `Span` as the currently
-active `Span` by default, but this functionality MAY be offered additionally
-as a separate operation.
+`Span` creation MUST NOT set the newly created `Span` as the
+active `Span` in the [implicit `Context`](#context-interaction) by default, but this functionality MAY be offered additionally
+as a separate operation in languages with implicit context propagation.
```
carlosalberto, line 333:
```suggestion
In languages with implicit `Context` propagation, `Span` creation MUST NOT
set the newly created `Span` as the active `Span` in the
[current `Context`](#context-interaction) by default, but this functionality
MAY be offered additionally as a separate operation.
```

carlosalberto: I think that this reads better, by **starting** the paragraph mentioning this case applies only for languages with implicit context handling.

carlosalberto (approved):

PR #1547, opened 2021-03-15
Zipkin exporter: Switch code / description tag values.

anuraaga:
## Changes

Zipkin instrumentation sets `error` to a code type of value, for example the exception class name or the gRPC status code, and not the exception message of gRPC description. So we should avoid setting it to the status description for consistency with Zipkin instrumentation. As we only have one code corresponding with an error, I set a boolean instead but maybe it could be the word `ERROR` anyways.

Reference:
https://github.com/open-telemetry/opentelemetry-go/pull/1688#issuecomment-798968221

### DIFF
--- a/specification/trace/sdk_exporters/zipkin.md
+++ b/specification/trace/sdk_exporters/zipkin.md
@@ -144,12 +144,11 @@ The following table defines the OpenTelemetry `Status` to Zipkin `tags` mapping.
 
 | Status|Tag Key| Tag Value |
 |--|--|--|
-|Code | `otel.status_code` | Name of the code, either `OK` or `ERROR`. MUST NOT be set if the code is `UNSET`. |
-|Description| `error` | Description of the `Status`. MUST be set if the code is `ERROR`, use an empty string if Description has no value. MUST NOT be set for `OK` and `UNSET` codes. |
+|Code | `error` | `true` if the code is `ERROR`, unset otherwise. |
+|Description| `otel.status_description` | Description of the `Status`. MUST be set if the code is `ERROR`, use an empty string if Description has no value. MUST NOT be set for `OK` and `UNSET` codes. |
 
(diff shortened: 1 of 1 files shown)


carlosalberto (approved):

carlosalberto:
Please add a CHANGELOG entry as this changes the expected exporter output.

specification/trace/sdk_exporters/zipkin.md
```diff
@@ -144,12 +144,11 @@ The following table defines the OpenTelemetry `Status` to Zipkin `tags` mapping.
 
 | Status|Tag Key| Tag Value |
 |--|--|--|
-|Code | `otel.status_code` | Name of the code, either `OK` or `ERROR`. MUST NOT be set if the code is `UNSET`. |
-|Description| `error` | Description of the `Status`. MUST be set if the code is `ERROR`, use an empty string if Description has no value. MUST NOT be set for `OK` and `UNSET` codes. |
+|Code | `error` | `true` if the code is `ERROR`, unset otherwise. |
```
bogdandrutu, line 147: Can you add a link to the documentation of the `error`? Is error a boolean?

SergeyKanzhelev (approved):

specification/trace/sdk_exporters/zipkin.md
SergeyKanzhelev: I *think* this is from https://github.com/opentracing/specification/blob/master/semantic_conventions.md#user-content-span-tags-table:~:text=true%20if%20and%20only%20if%20the%20application%20considers%20the%20operation%20represented%20by%20the%20Span%20to%20have%20failed

anuraaga:
I discussed this with @jcchavezs and realized that this isn't quite correct either, we should use http / rpc code for example where available, otherwise description is still ok. Sorry for taking the time, let me close this for now and come up with a different idea for a PR.

PR #1550, opened 2021-03-16
Update semantic conventions to distinguish between int and double

### LINKED ISSUE #1096: Update YAML files for semantic conventions once supported by markdown generator
YAML model for attributes seems to only have number type. But we have int or double as possible attribute types.

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/common/common.md#attributes
https://github.com/open-telemetry/opentelemetry-specification/blob/master/semantic_conventions/syntax.md

Once the YAML syntax and MD generator are updated (https://github.com/open-telemetry/build-tools/issues/13), we need to update our YAML files here accordingly.

arminru:
This distinction was recently introduced in https://github.com/open-telemetry/build-tools/releases/tag/v0.3.0.
Note that the generic `number` type was dropped.

Fixes #1096.

### DIFF
--- a/semantic_conventions/syntax.md
+++ b/semantic_conventions/syntax.md
@@ -34,10 +34,12 @@ attributes ::= (id type brief examples | ref [brief] [examples]) [required] [not
 ref ::= id
 
 type ::= "string"
-     |   "number"
+     |   "int"
+     |   "double"
      |   "boolean"
      |   "string[]"
-     |   "number[]"
+     |   "int[]"
+     |   "double[]"
      |   "boolean[]"
      |   enum
 
@@ -95,10 +97,12 @@ An attribute is defined by:
 - `type`, either a string literal denoting the type or an enum definition (See later).
    The accepted strings literals are:
   * "string": String attributes.
-  * "number": Numeric attributes.
+  * "int": Integer attributes.
+  * "double": Double attributes.
   * "boolean": Boolean attributes.
   * "string[]": Array of strings attributes.
-  * "number[]": Array of numbers attributes.
+  * "int[]": Array of integer attributes.
+  * "double[]": Array of double attributes.
   * "boolean[]": Array of booleans attributes.
 - `ref`, optional string, reference an existing attribute, see later.
 - `required`, optional, specifies if the attribute is mandatory.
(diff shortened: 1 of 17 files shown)


arminru:
@anuraaga With this change you will be able to resolve https://github.com/open-telemetry/opentelemetry-specification/pull/1422#discussion_r573550539 since v0.3.0 allows you to define it as `double`.

carlosalberto (approved):

jsuereth (approved):

bogdandrutu (approved):

semantic_conventions/syntax.md
```diff
@@ -95,10 +97,12 @@ An attribute is defined by:
 - `type`, either a string literal denoting the type or an enum definition (See later).
    The accepted strings literals are:
   * "string": String attributes.
-  * "number": Numeric attributes.
+  * "int": Integer attributes.
+  * "double": Double attributes.
```
MrAlias, line 101: The naming here seems a bit odd to me. `int` encompasses all integer values (it does not make a statement about signed or unsigned or the precision level to store the value as), but `double` is short for double precision floating point number (it included a precision level as well as a number type).

Should there be an additional type to specify unsigned integer values?

Should `double` be replaced with the unspecified precision `float`?

arminru: The types used in our semantic conventions match the ones defined for attributes in general:
https://github.com/open-telemetry/opentelemetry-specification/blob/v1.1.0/specification/common/common.md#L19-L23
> The attribute value, which is either:
> - A primitive type: string, boolean, double precision floating point (IEEE 754-1985) or signed 64 bit integer.
> - An array of primitive type values. (...)

And consequently OTLP:
https://github.com/open-telemetry/opentelemetry-proto/blob/v0.7.0/opentelemetry/proto/common/v1/common.proto#L30-L37

arminru: @MrAlias I added a reference to that definition in 743dfca.

carlosalberto: @MrAlias Is this resolved by the latest change?

MrAlias: :eyes:

MrAlias (approved):

anuraaga (approved):

PR #1555, opened 2021-03-17
Clarify non-blocking requirement from span API End

### LINKED ISSUE #1522: Non-Blocking clarification for Span End method
[The specification states](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.0.1/specification/trace/api.md#end) states a "Span becomes non-recording by being ended". [The specification](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.0.1/specification/trace/api.md#isrecording) reiterates this in the `IsRecording` section by saying "[a]fter a Span is ended, it usually becomes non-recording and thus IsRecording SHOULD consequently return false for ended Spans." This means that the `End` and `IsRecorded` must share some state about the ended nature of a Span.

How this shared state is manage in implementations as it can lead to a race condition if it is accessed simultaneously. To specify how this should be handled [the specification state](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.0.1/specification/trace/api.md#concurrency) that "[a]ll methods of Span are safe to be called concurrently", but it also [states](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.0.1/specification/trace/api.md#end) the `End` method of a Span "MUST be non-blocking."

This seems to introduce a logical inconsistency. How can state be shared, be concurrent safe, and not block?

MrAlias:
Fixes #1522 

## Changes

[The specification states](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.0.1/specification/trace/api.md#end) states a "Span becomes non-recording by being ended". [The specification](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.0.1/specification/trace/api.md#isrecording) reiterates this in the `IsRecording` section by saying "[a]fter a Span is ended, it usually becomes non-recording and thus IsRecording SHOULD consequently return false for ended Spans." This means that the `End` and `IsRecorded` must share some state about the ended nature of a Span.

How this shared state is manage in implementations as it can lead to a race condition if it is accessed simultaneously. To specify how this should be handled [the specification state](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.0.1/specification/trace/api.md#concurrency) that "[a]ll methods of Span are safe to be called concurrently", but it also [states](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.0.1/specification/trace/api.md#end) the `End` method of a Span "MUST be non-blocking."

This introduces a logical inconsistency.

The term lock-free is also [interpreted](https://github.com/open-telemetry/opentelemetry-specification/issues/1522#issuecomment-796626444) as:

> you should not synchronously do I/O e.g. to export the span.

And as the comment goes on to state, "that's exactly what the built-in SimpleSpanProcessor does." 

This change updates the specification to clarify what "non-blocking" means:

- Do not preform blocking I/O on calling thread
- Try and provide a lock-free algorithm and require lock usage be minimized
- Scope to not include "non-production" SpanExporters and SpanProcessors


yurishkuro:
Putting aside the possible inconsistencies in the existing wording, what are you trying to achieve? To allow `span.End()` to do IO/flushing on the caller thread? I feel like "non-blocking" is an important requirement, but we could clarify that it specifically refers to doing IO, not acquiring locks.

Also, locks are implementation details, they are not required. The functionality of End and IsRecording could be implemented with lockless structures.

MrAlias:
> Putting aside the possible inconsistencies in the existing wording, what are you trying to achieve?

The end goal is to ensure the Go implementation of the specification is compliant with the specification. We currently have an implementation with locks at the application level that 

> To allow `span.End()` to do IO/flushing on the caller thread? I feel like "non-blocking" is an important requirement, but we could clarify that it specifically refers to doing IO, not acquiring locks.

That would be fine, but as was pointed out in the linked issue doing IO/flushing is "exactly what the built-in SimpleSpanProcessor does". I don't think this definition would be consistent with the rest of the specification for that reason either.

> Also, locks are implementation details, they are not required. The functionality of End and IsRecording could be implemented with lockless structures.

I'm not the most knowledgeable about the lockfree solution you mean, but based on my understanding that could mean that application locks have moved down to the OS or hardware level. Given the lack of a definition for what non-blocking means it could preclude even this.

Oberon00 (approved):

specification/trace/api.md
```diff
@@ -617,8 +617,6 @@ Parameters:
 - (Optional) Timestamp to explicitly set the end timestamp.
   If omitted, this MUST be treated equivalent to passing the current time.
 
-This API MUST be non-blocking.
```
yurishkuro, line 620: Suggest: The implementations of this API MUST NOT perform blocking I/O.

I actually wonder if there was another intent here, because the line refers to API not the implementation, which could be interpreted as "the signature of the method must be explicitly async". But I don't think it would be right for the `End` method specifically, it should be a plain function, which is different from things like `Tracer.ForceFlush`.

MrAlias: > Suggest: The implementations of this API MUST NOT perform blocking I/O.

How will we account for the SimpleSpanProcessor doing blocking I/O when a span `End` method is called?

yurishkuro: We could relax to SHOULD NOT. I am not familiar with SimpleSpanProcessor and why it is implemented as blocking, since it makes it unusable in production setting.

jkwatson: I don't think the SSP does any blocking I/O itself...it just calls the exporter (which indeed might, but that's not the SSP's concern).

iNikem:
[SpanProcessor spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#onendspan) says:

> OnEnd is called after a span is ended (i.e., the end timestamp is already set). This method MUST be called synchronously within the Span.End() API, therefore it should not block or throw an exception.

[SpanExporter spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#exportbatch) says:

> Export() MUST NOT block indefinitely

To me this means that per current spec the whole pipeline `Span.end -> SpanProcessor.onEnd -> SpanExporter.export` should not or even must not block for a prolonged period of time. There is no inconsistency here nor any need to allow `Span.end` to be blocking.

iNikem:
Whether `Span.end` and `Span.isRecording` need shared state or not is implementation detail. Whether that share state require blocking or not is also implementation detail which may vary from language to language. Thus I am against this change to the spec.

The only reason to drop current non-blocking requirement may be if some language absolutely CANNOT implement this portion of the spec. Does Go really not allow for multi-threaded non-blocking access to shared state? Or is it just current implementation that can be changed?

Oberon00:
> Export() MUST NOT block indefinitely

This requirement should be removed or clarified as well IMHO, as it says exactly nothing. As long as the method finishes at some point (even after a 5 minute timeout), the requirement is satisfied.

"MUST NOT block" is a much stricter requirement than "MUST NOT block indefinitely". Thus I still think that SimpleSpanProcessor violates the "MUST NOT block" requirement of the spec.

iNikem:
> This requirement should be removed or clarified as well IMHO

Agreed

specification/trace/api.md
MrAlias: Okay, how do we account for the exporter that might be called synchronously from the SimpleSpanProcessor that will perform blocking I/O when span `End` method is called? Pushing the problem down the chain doesn't remove it.

Should the API be specifying this if it is a concern of the implementations further downstream?

MrAlias:
> The only reason to drop current non-blocking requirement may be if some language absolutely CANNOT implement this portion of the spec. Does Go really not allow for multi-threaded non-blocking access to shared state? Or is it just current implementation that can be changed?

We can definitely engineer an algorithm differently, we would just need to know what the requirements are. It sounds like to some languages "non-blocking" has significance to some, but it is not clear to us and others what that means. Can locks be used? Can atomic operations be used? Can I/O happen?

Given the inconsistency with the rest of the specifications, I opened this PR to remove the confusion entirely. I think it seems appropriate as there is still disagreement in the community overall what "non-blocking" means.

MrAlias:
> "MUST NOT block" is a much stricter requirement than "MUST NOT block indefinitely". Thus I still think that SimpleSpanProcessor violates the "MUST NOT block" requirement of the spec.

Agreed. I think this is the point where the rest of the "non-blocking" requirements become superficial at best.

I also do not think we can switch to "MUST NOT block" at this point. That would be a backwards incompatible change to the specification as it would further restrict a requirement and violate our stability guarantee.

anuraaga (approved):
This language seems more useful for the BatchSpanProcessor than the API - there is really nothing `Span.end()` can do about the possibility of blocking in span processors / exporters. So it needs to be just as non-blocking as other methods like setAttribute. A text at the top of the doc applying to all methods could make sense to me but not on only the one `Span.end`.

yurishkuro (dismissed):
Blocking this against accidental merge. There's no consensus from what I can see, and this is a critical piece of the API which if we relax it we cannot make it more restrictive again.

MrAlias:
> Blocking this against accidental merge. There's no consensus from what I can see, and this is a critical piece of the API which if we relax it we cannot make it more restrictive again.

Just to make sure this doesn't become stale and wither away, @yurishkuro what needs to be done to merge this or resolve the associated issue? This resolution is something needed for the Go implementation to make a release candidate and a 1.0.0 release.

yurishkuro:
I think we need to keep the non-blocking wording, but clarify that it means the implementation must not perform blocking I/O on the calling thread.

specification/trace/api.md
Oberon00: In the end, this makes it possible for the user to configure the SDK in a way that violates the spec: The SimpleSpanProcessor may call the exporter synchronously, assuming the exporter will handle the export asynchronously. But the exporter may work synchronously because it assumes it is only called from a SpanProcessor that already calls export on a background thread (and synchronous I/O is generally faster (per operation) than asynchronous I/O).

Oberon00: Another use case that this MUST breaks, even in the "no blocking I/O" form is a debug use case where I want each span to be synchronously logged as soon as it is ended, so that e.g. spans written to the console are properly ordered relative to application output / logs written to the console.

Oberon00:
I think we should:
* Replace this API requirement with just a general "Note:" for all span APIs that they may be called in hot 
  application paths and therefore should be designed to complete fast, preferably immediately. Note that I/O and 
  locking should be minimized to the required minimum, etc.

  If an API-implementer actually thinks it is a good idea to make a synchronous HTTP request on each end call, would 
  they even read the spec? 😃 Or maybe they have a very good reason?
* Add a more specific requirement to the SDK's `End`, namely that it MUST ONLY record the end timestamp, mark the 
  span as ended and non-recording plus call the OnEnd of any span processors.

iNikem:
> * Replace this API requirement with just a general "Note:" for all span APIs that they may be called in hot
>   application paths and therefore should be designed to complete fast, preferably immediately. Note that I/O and
>   locking should be minimized to the required minimum, etc.

I support this proposal

MrAlias:
@anuraaga @Oberon00 @yurishkuro @iNikem I've updated based on the feedback. PTAL

yurishkuro (approved):

carlosalberto (approved):

specification/trace/api.md
```diff
@@ -617,7 +617,14 @@ Parameters:
...
+Expect this operation to be called in the "hot path" of production
```
Oberon00, line 620: This applies to all span methods, not only End.

specification/trace/api.md
```diff
@@ -617,7 +617,14 @@ Parameters:
...
+applications. It needs to be designed to complete fast, if not immediately.
+This operation itself MUST NOT perform blocking I/O on the calling thread.
```
Oberon00, line 622: I would not add this requirement. "Blocking" I/O e.g. to a file might be very fast. At most, I would make it a SHOULD.

specification/trace/api.md
```diff
@@ -617,7 +617,14 @@ Parameters:
...
+Any locking used MUST be minimized and SHOULD be removed entirely if
```
Oberon00, line 623: I would also make this a note. Also "MUST be minimized" is more or less a SHOULD

specification/trace/api.md
```diff
@@ -617,7 +617,14 @@ Parameters:
...
+possible. The downstream SpanProcessors and subsequent SpanExporters called
```
Oberon00, line 624:
```suggestion
possible. Some downstream SpanProcessors and subsequent SpanExporters called
```

specification/trace/api.md
MrAlias: @yurishkuro would you be okay with the change suggested here?

specification/trace/api.md
MrAlias: Agreed. I'm a little hesitant to increase the scope of the changes in this PR even further though. @Oberon00 do you think we need to restructure so these requirements apply to all operations of a span in this PR, or can we do that in a subsequent PR?

specification/trace/api.md
MrAlias: Agreed, I was having a hard time stating this. Based on your comment, what do you think of:

```suggestion
Any locking used needs be minimized and SHOULD be removed entirely if
```

That way we remove the ineffective requirement and make it clear what the suggestion is.

specification/trace/api.md
yurishkuro: I was ok with MUST NOT because it's qualified that the End() function itself must not do that, while exporters that might get called "should not".

specification/trace/api.md
Oberon00: If you like you can do it in two PRs (I will keep my approval here), it just seems odd to me to commit code that you know will 100% change (be moved).

reyang (approved):

specification/trace/api.md
carlosalberto: @Oberon00 Please fill an issue for a follow up if you think we need to make this a SHOULD instead. Otherwise merging as I don't want this PR to remain stale.

specification/trace/api.md
carlosalberto: @MrAlias please let's fill a follow up for @Oberon00's  suggestion ;)

PR #1556, opened 2021-03-17
Remove Included Propagators section from trace API spec

### LINKED ISSUE #1544: Is the Included Propagators section a duplicate of Propagators Distribution?
The [Included Propagators](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.0.1/specification/trace/api.md#included-propagators) section specifies that a W3C tracecontext propagator must be included in the API, however the [Propagators Distribution](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.0.1/specification/context/api-propagators.md#propagators-distribution) also mentions that a W3C tracecontext propagator must be included (though it says it MAY be included directly in the API).

This seems like duplicate specification at best and possibly even conflicting specification.

The duplication of this specification should be resolved by unifying on the location of this information. It seems [Propagators Distribution](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.0.1/specification/context/api-propagators.md#propagators-distribution) is the more comprehensive body of work so it seems a better fit there.

The requirement or option that the implementation be included in the API needs to be resolved. Easing a requirement is a backwards compatible change as opposed to making an option now a requirement. Therefore, it would seem the best course of action here is to resolve this by specifying it is an OPTION to be distributed as part of the OpenTelemetry API.

MrAlias:
This is a duplicate of the [Propagators Distribution](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.0.1/specification/context/api-propagators.md#propagators-distribution) of the context specification.

Fixes #1544


utpilla (approved):
LGTM.

iNikem (approved):

carlosalberto:
Not quite a duplicate, as the Propagators Distribution is a general note, while this one is about the API being **the place** where `TraceContext` lives. That being said, we may need to have better wording for it, in order to avoid confusion.

MrAlias:
> Not quite a duplicate, as the Propagators Distribution is a general note, while this one is about the API being **the place** where `TraceContext` lives. That being said, we may need to have better wording for it, in order to avoid confusion.

@carlosalberto I'm not sure I follow. From this section of the specification:

> The API layer or an extension package MUST include the following `Propagator`s

My understanding reading that is that the API _or_ an extension package is the place `TraceContext` lives.  Additionally this is restated in the Propagators Distribution section:

>The official list of propagators that MUST be maintained by the OpenTelemetry organization and MUST be distributed as OpenTelemetry extension packages [include] W3C TraceContext [which] MAY alternatively be distributed as part of the OpenTelemetry API.

This excerpt from the Propagators Distribution section contains normative requirements and options and doesn't seem to be a note.

cijothomas (approved):

carlosalberto:
@MrAlias You are definitely correct - the original intention of that paragraph in api.md (before the amendment in the Propagators Distribution doc), is that we **really** suggest SIGs to have the TraceContext (and Baggage) propagators right in the base API packages, even if distributing them as extension packages is allowed.

reyang (approved):

jsuereth (approved):

MrAlias:
@carlosalberto I'm still not 100% sure what your suggestion for a change here is?

You would definitely be the one to know what the original intention was there and I don't mean to step on toes. I'm just going off what the final result ended up being and what was released. Based on how both section of the specification are written today they appear to me to be duplicates and removing the less comprehensive of the two seems like a good way to remove redundancy and avoid future confusion and internal divergences.

specification/trace/api.md
```diff
@@ -748,15 +747,6 @@ be called concurrently.
...
 
-## Included Propagators
-
-The API layer or an extension package MUST include the following `Propagator`s:
-
-* A `TextMapPropagator` implementing the [W3C TraceContext Specification](https://www.w3.org/TR/trace-context/).
-
-See [Propagators Distribution](../context/api-propagators.md#propagators-distribution)
-for how propagators are to be distributed.
-
```
bogdandrutu, line 759: Should we keep a link just in case people start reading this and make sure they know about that?

bogdandrutu (approved):

specification/trace/api.md
MrAlias: SGTM

carlosalberto:
@MrAlias We just need a small note here regarding that we suggest (not even a SHOULD) that SIGs include this TracePropagator in the API package/artifact itself. Will do that myself to not stale this PR any longer. Thanks!

carlosalberto (approved):

PR #1533, opened 2021-03-11
README for metrics spec

reyang:
## Changes

1. Have a general README doc to guide readers through the metrics specs.
2. Provide context - e.g. design goal.

Eventually we want to move the metrics terms/concepts here, so the API/SDK spec can focus on the concrete classes/methods/requirements.

Related issues #

Related [oteps](https://github.com/open-telemetry/oteps):
* https://github.com/open-telemetry/oteps/blob/main/text/metrics/0146-metrics-prototype-scenarios.md

Issue #1366, opened 2021-01-22
Need clarity on scope of Instrument Name


victlu:
# Instruments with same namespace / names

> Meter implementations MUST return an error when multiple instruments are registered by the same name.

## Example

A service installs OTel SDK and uses a 3P library (instrumented with OTel).  The service is required to initializing the 3P library for each new transaction.  This causes the 3P library to register the instrument (with same namespace and instrument name).  Thus, resulting in failure except for 1st instance.

How should this situation be resolved?

## Expectation

When 3P library calls New-Instrument(Name) from multiple instances (in scope of one OTel SDK instance), I would expect the SAME instrument to be returned.  Thus, all instances of 3P library will add/record measurements to the singleton "namespace/name" instrument.

# Different instrument type with same name

> Metric instrument names belong to a namespace, established by the the associated Meter instance.

## Example

Create Counter and UpDownCounter and ValueRecorder with the same instrument name.  Are they the same instrument or are they 3 different instruments (differentiated by kind/group)?  Are Counter and UpDownCounter one or two distinct instruments (both are technically counters)?

## Expectation

Need to incorporate "Kind" (or groupings) of instrument when establishing distinctness of instruments.

# Implementation of "number-type" specific instruments

## Example

C# SDK (other SDKs as well) incorporates number-types (i.e. "Int64", "Double") into the API when creating an instrument.  (i.e. CreateInt64Counter(), CreateDoubleCounter()).  If these are given the same instrument name, are they distinct and different instruments?

## Expectation

Number-Type distinction should not be part of the API name.   The APIs (i.e. Add/Record) should accept different number-types as parameters (for languages that are sensitive to different number types).

Also, conversions/promotions of number-types needs to be specified (i.e. int64 can be promoted to double, etc...) when summing or summarizing or aggregating results.  This may be specific to SDK implementations.

jmacd:
These are all good questions. Given our recent discussions in the Metric working group(s), I'd like to try and re-cast these questions in terms of the protocol and not the API and SDK behavior. Suppose the collector is presented with more than one kind of OTLP data point for a metric with the same name. How should it treat that data? I believe the best approach is to treat these as distinct, i.e., Counter and UpDownCounter of the same name are different.

[...]

PR #1557, opened 2021-03-18
MeterProvider and Meter API spec

### LINKED ISSUE #1046: Metrics: Make duplicate instrument registration invalid

The OpenCensus Metric API made statements about when a duplicately-named instrument could be registered. This needs to be reconsidered in today's OTel API and SDK specification. Since we have added the concept of Named Meters, I believe we can remove any treatment of duplicate registration and make duplicate registration invalid behavior.

See this section in the current SDK specification: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/sdk.md#sdk-instrument-registration

This states that under no conditions can the same instrument name be registered in the same Named Meter. The assumption is that if you are a single instrumentation library, then you are able to arrange to construct each metric instrument that you will use once. If you are expecting to create a metric with the same name, you should use a different Named Meter.

It never made sense to support duplicate registration for asynchronous instruments, since it would imply some conflict in the number of callbacks. For synchronous instruments it we can devise rules to allow duplicate registration, but it's not clear that this is useful to the user.

Since this is already written into the current specification, we can close this issue if people agree.

### LINKED ISSUE #1113: Rename "Label" concept to "Attribute" for consistency with OpenTelemetry tracing

The resolution from https://github.com/open-telemetry/opentelemetry-specification/issues/948 is that "Label" should be renamed "Attribute" for consistency across OpenTelemetry. This has been discussed widely and there are no objections on the Metrics SIG to this renaming.

This renaming will introduce a breaking change for JSON-format exporters using OTLP v0.5, but will not introduce a binary-format incompatibility when this change is released as OTLP v0.6.

The specification will be updated to explain that although we use a single concept and term to describe attributes, sometimes we choose to limit their values to only strings. To address this across OpenTelemetry, we will specify how to handle the situation whenever requesting a string-valued attribute in a context where it would be prohibitively expensive to support. These cases are:

1. A list-valued attribute used as a metric attribute or as a resource attribute exported through a legacy metricexporter
2. A map-valued attribute used as a ... (same) ...

In both cases, I think a simple text-value replacement like "unrepresentable" would satisfy me.

**Additional context.**

Issue #948 contains a lengthy discussion of this topic. The terms "Attribute", "Label", "Tag", "Property", and "Association" are all very close in meaning, and we have come to agree that having more than one of these terms in use causes more trouble than it helps with.

### LINKED ISSUE #1433: Consider removing Async Instruments (Observers) from first Metric version
Doing async instruments properly seems rather complicated to me (e.g. #1432). If we specify synchronous instruments (especially BatchRecorder) properly, users should be able to implement a functionally equivalent solutions on their own (e.g. with their own background thread). (The only extra thing that might be needed for that is some "last-chance to collect" callback before every export.)

reyang:
Fixes #1046.
Fixes #1433.

## Changes

Follow up on #1533 and the Metrics API/SDK SIG.

* Reduced the scope of the metrics API spec by focusing on [OTEP146](https://github.com/open-telemetry/oteps/blob/main/text/metrics/0146-metrics-prototype-scenarios.md) scenario.
* Borrowed the idea from the stable version of Tracing API, organized the Metrics API spec in the similar flow.
* Documented the MeterProvider and Meter API.
* Added the initial skeleton of the Instrument API.

Related issues #1113 #1366

Related [oteps](https://github.com/open-telemetry/oteps)
* https://github.com/open-telemetry/oteps/blob/main/text/metrics/0146-metrics-prototype-scenarios.md

### DIFF
--- a/specification/metrics/new_api.md
+++ b/specification/metrics/new_api.md
@@ -52,19 +52,19 @@ metrics API:
     |
     +-- Meter(name='io.opentelemetry.runtime', version='1.0.0')
     |   |
-    |   +-- Instrument<Gauge, int>(name='cpython.gc', attributes=['generation'], unit='kB')
+    |   +-- instruments...
     |
     +-- Meter(name='io.opentelemetry.contrib.mongodb.client', version='2.3.0')
         |
         +-- Instrument<Counter, int>(name='client.exception', attributes=['type'], unit='1')
         |
-        +-- Instrument<Histogram, double>(name='client.duration', attributes=['net.peer.host', 'net.peer.port'], unit='ms')
+        +-- instruments...
 
 +-- MeterProvider(custom)
     |
     +-- Meter(name='bank.payment', version='23.3.5')
         |
-        +-- Instrument<Histogram, double>(name='transaction', attributes=['category', 'status'], unit='USD')
+        +-- instruments...
 ```
 
 ## MeterProvider
@@ -165,10 +165,18 @@ the name:
   Instruments under one Meter SHOULD NOT interfere with Instruments under
   another Meter.
 
-Instrument names MUST conform to the following syntax:
+Instrument names MUST conform to the following syntax (described using the
+[Augmented Backus-Naur Form](https://tools.ietf.org/html/rfc5234)):
+
+```abnf
+instrument-name = ALPHA 0*62 ("_" / "." / "-" / ALPHA / DIGIT)
+
+ALPHA = %x41-5A / %x61-7A; A-Z / a-z
+DIGIT = %x30-39 ; 0-9
+```
 
 * They are not null or empty strings.
-* They are case-insensitive.
+* They are case-insensitive, ASCII strings.
 * The first character must be an alphabetic character.
 * Subsequent characters must belong to the alphanumeric characters, '_', '.',
   and '-'.
--- a/specification/metrics/new_api.md
+++ b/specification/metrics/new_api.md
@@ -36,7 +36,7 @@ Table of Contents
 
 ## Overview
 
-The Metrics API consists of these main classes:
+The Metrics API consists of these main components:
 
 * [MeterProvider](#meterprovider) is the entry point of the API. It provides
   access to `Meters`.
--- a/specification/metrics/new_api.md
+++ b/specification/metrics/new_api.md
@@ -1,0 +1,205 @@
+# Metrics API
+
+**Status**: [Experimental](../document-status.md)
+
+**Owner:**
+
+* [Reiley Yang](https://github.com/reyang)
+
+**Domain Experts:**
+
+* [Bogdan Brutu](https://github.com/bogdandrutu)
+* [Josh Suereth](https://github.com/jsuereth)
+* [Joshua MacDonald](https://github.com/jmacd)
+
+Note: this specification is subject to major changes. To avoid thrusting
+language client maintainers, we don't recommend OpenTelemetry clients to start
+the implementation unless explicitly communicated via
+[OTEP](https://github.com/open-telemetry/oteps#opentelemetry-enhancement-proposal-otep).
+
+<details>
+<summary>
+Table of Contents
+</summary>
+
+* [Overview](#overview)
+* [MeterProvider](#meterprovider)
+  * [MeterProvider operations](#meterprovider-operations)
+* [Meter](#meter)
+  * [Meter operations](#meter-operations)
+* [Instrument](#instrument)
+  * [Counter](#counter)
+    * [Counter Creation](#counter-creation)
+* [Measurement](#measurement)
+
+</details>
+
+## Overview
+
+The Metrics API consists of these main classes:
+
+* [MeterProvider](#meterprovider) is the entry point of the API. It provides
+  access to `Meters`.
+* [Meter](#meter) is the class responsible for creating `Instruments`.
+* [Instrument](#instrument) is responsible for reporting
+  [Measurements](#measurement).
+
+Here is an example of the object hierarchy inside a process instrumented with the
+metrics API:
+
+```text
++-- MeterProvider(default)
+    |
+    +-- Meter(name='io.opentelemetry.runtime', version='1.0.0')
+    |   |
+    |   +-- Instrument<Gauge, int>(name='cpython.gc', attributes=['generation'], unit='kB')
+    |
+    +-- Meter(name='io.opentelemetry.contrib.mongodb.client', version='2.3.0')
+        |
+        +-- Instrument<Counter, int>(name='client.exception', attributes=['type'], unit='1')
+        |
+        +-- Instrument<Histogram, double>(name='client.duration', attributes=['net.peer.host', 'net.peer.port'], unit='ms')
+
++-- MeterProvider(custom)
+    |
+    +-- Meter(name='bank.payment', version='23.3.5')
+        |
+        +-- Instrument<Histogram, double>(name='transaction', attributes=['category', 'status'], unit='USD')
+```
+
+## MeterProvider
+
+`Meter`s can be accessed with a `MeterProvider`.
+
+In implementations of the API, the `MeterProvider` is expected to be the
+stateful object that holds any configuration.
+
+Normally, the `MeterProvider` is expected to be accessed from a central place.
+Thus, the API SHOULD provide a way to set/register and access a global default
+`MeterProvider`.
+
+Notwithstanding any global `MeterProvider`, some applications may want to or
+have to use multiple `MeterProvider` instances, e.g. to have different
+configuration for each, or because its easier with dependency injection
+frameworks. Thus, implementations of `MeterProvider` SHOULD allow creating an
+arbitrary number of `MeterProvider` instances.
+
+### MeterProvider operations
+
+The `MeterProvider` MUST provide the following functions:
+
+* Get a `Meter`
+
+#### Get a Meter
+
+This API MUST accept the following parameters:
+
+* `name` (required): This name must identify the [instrumentation
+  library](../overview.md#instrumentation-libraries) (e.g.
+  `io.opentelemetry.contrib.mongodb`). If an application or library has built-in
+  OpenTelemetry instrumentation, both [Instrumented
+  library](../glossary.md#instrumented-library) and [Instrumentation
+  library](../glossary.md#instrumentation-library) may refer to the same
+  library. In that scenario, the `name` denotes a module name or component name
+  within that library or application. In case an invalid name (null or empty
+  string) is specified, a working Meter implementation MUST be returned as a
+  fallback rather than returning null or throwing an exception, its `name`
+  property SHOULD keep the original invalid value, and a message reporting that
+  the specified value is invalid SHOULD be logged. A library, implementing the
+  OpenTelemetry API *may* also ignore this name and return a default instance
+  for all calls, if it does not support "named" functionality (e.g. an
+  implementation which is not even observability-related). A MeterProvider
+  could also return a no-op Meter here if application owners configure the SDK
+  to suppress telemetry produced by this library.
+* `version` (optional): Specifies the version of the instrumentation library
+  (e.g. `1.0.0`).
+
+It is unspecified whether or under which conditions the same or different
+`Meter` instances are returned from this functions.
+
+Implementations MUST NOT require users to repeatedly obtain a `Meter` again with
+the same name+version to pick up configuration changes. This can be achieved
+either by allowing to work with an outdated configuration or by ensuring that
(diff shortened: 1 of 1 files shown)


specification/metrics/api.md
```diff
@@ -17,1125 +17,204 @@ language client maintainers, we don't recommend OpenTelemetry clients to start
...
-```golang
-func (s *server) method(ctx context.Context) {
-    // ... other work
-
-    s.instruments.counter1.Add(ctx, 1,
-        kv.String("labelA", "..."),
-        kv.String("labelB", "..."),
-        )
-}
+The Metrics API consists of these main classes:
```
MrAlias, line 43: This is not programming language agnostic language. Please do not impose OoO paradigms in the specification, they are not implementable in all programming languages.

specification/metrics/api.md
```diff
@@ -17,1125 +17,204 @@ language client maintainers, we don't recommend OpenTelemetry clients to start
...
+
+* [MeterProvider](#meterprovider) is the entry point of the API. It provides
+  access to `Meters`.
+* [Meter](#meter) is the class responsible for creating `Instruments`.
+* [Instrument](#instrument) is used to refer to [Counter](#counter),
+  [Gauge](#gauge) and [Histogram](#histogram), which are the classes responsible
+  for reporting `Measurements`.
```
MrAlias, line 50: What happened to the synchronous/asynchronous instruments?

specification/metrics/api.md
cijothomas: If we are fixing this, we need to address other parts of the specs as well which use similar wording (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#tracing-api)

MrAlias: I'm would rather not increase the scope of this PR to fix the rest of the specification, but instead limit it to not include the same mistakes from the past.

bogdandrutu (dismissed):
This is a huge change on how we envisioned metrics API and how we wanted this to be. I am not clear about motivation and direction of this project.

bogdandrutu (commented):
Also I would encourage to create a new file for the moment "new_api.md" it is extremely hard to review since too many deletes and github does not help.

specification/metrics/api.md
reyang: I agree with "do not impose OoO paradigms in the specification".
The wording and structure are borrowed from the stable version of tracing spec so I think we shouldn't block on this?

specification/metrics/api.md
reyang: The instruments should cover both synchronous and asynchronous.

reyang:
> Also I would encourage to create a new file for the moment "new_api.md" it is extremely hard to review since too many deletes and github does not help.

Yes, we'll discuss this in today's Metrics API/SDK SIG meeting. Current options are:
1. use OTEP before making PR to spec repo.
2. create a separate doc (e.g. `new_api.md`).
3. change the current spec.

bogdandrutu:
> create a separate doc (e.g. new_api.md).

Asking because it is hard to read the new text, before merging we can discuss if we replace the old text, but cannot read what is the new text

jmacd (approved):

reyang:
Based on the discussion during the [Metrics API/SDK SIG meeting](https://docs.google.com/document/d/1-bCYkN-DWJq4jw1ybaDZYYmx-WAe6HnwfWbkm8d57v8/edit#heading=h.czd4en83sgxk), I've created a separate file (`new_api.md`) to make it easier for PR review, and removed gauge/histogram (which we will cover in separate PRs).

specification/metrics/new_api.md
```diff
@@ -0,0 +1,202 @@
...
+  types
+
+Instruments are associated with the Meter during creation, and are identified by
+the name:
+
+* Meter implementations MUST return an error when multiple Instruments are
+  registered under the same Meter instance using the same name.
+* Different Meters MUST be treated as separate namespaces. The names of the
+  Instruments under one Meter SHOULD NOT interfere with Instruments under
+  another Meter.
```
sirianni, line 166: You mentioned in Slack:
> if you use the same metric name and different instrumentation libraries, the intention is that they are the same metric

However, unless I'm misreading, I don't see this explicitly stated here.

specification/metrics/new_api.md
```diff
@@ -0,0 +1,202 @@
...
+configuration must be stored per-meter (such as disabling a certain meter), the
+meter could, for example, do a look-up with its name+version in a map in the
+`MeterProvider`, or the `MeterProvider` could maintain a registry of all
+returned `Meter`s and actively update their configuration if it changes.
+
+## Meter
+
+The meter is responsible for creating [Instruments](#instrument).
+
+Note that `Meter` should usually *not* be responsible for configuration. This
```
jsuereth, line 137: nit: weak language.

```
Note: `Meter` should not be responsible for this configuration.   This should be the responsibility of the `MeterProvider`.
```

You're already using SHOULD instead of MUST so don't weaken it any further.

jsuereth (approved):

specification/metrics/new_api.md
jsuereth: "are identified by their name".   Maybe that need some highlights/implciations.

specification/metrics/new_api.md
reyang: Updated the wording.

specification/metrics/new_api.md
```diff
@@ -0,0 +1,202 @@
...
+        |
+        +-- Instrument<Counter, int>(name='client.exception', attributes=['type'], unit='1')
+        |
+        +-- Instrument<Histogram, double>(name='client.duration', attributes=['net.peer.host', 'net.peer.port'], unit='ms')
+
++-- MeterProvider(custom)
+    |
+    +-- Meter(name='bank.payment', version='23.3.5')
+        |
+        +-- Instrument<Histogram, double>(name='transaction', attributes=['category', 'status'], unit='USD')
```
bogdandrutu, line 67: Can we remove references to `Histogram` :)

specification/metrics/new_api.md
```diff
@@ -0,0 +1,202 @@
...
+    |
+    +-- Meter(name='io.opentelemetry.runtime', version='1.0.0')
+    |   |
+    |   +-- Instrument<Gauge, int>(name='cpython.gc', attributes=['generation'], unit='kB')
+    |
+    +-- Meter(name='io.opentelemetry.contrib.mongodb.client', version='2.3.0')
+        |
+        +-- Instrument<Counter, int>(name='client.exception', attributes=['type'], unit='1')
+        |
+        +-- Instrument<Histogram, double>(name='client.duration', attributes=['net.peer.host', 'net.peer.port'], unit='ms')
```
bogdandrutu, line 61: Can we remove references to `Histogram` :)

specification/metrics/new_api.md
```diff
@@ -0,0 +1,202 @@
...
+
+Here is an example of the object hierarchy inside a process instrumented with the
+metrics API:
+
+```text
++-- MeterProvider(default)
+    |
+    +-- Meter(name='io.opentelemetry.runtime', version='1.0.0')
+    |   |
+    |   +-- Instrument<Gauge, int>(name='cpython.gc', attributes=['generation'], unit='kB')
```
bogdandrutu, line 55: Can we remove references to `Gauge` :)

specification/metrics/new_api.md
```diff
@@ -0,0 +1,202 @@
...
+Instruments are associated with the Meter during creation, and are identified by
+the name:
+
+* Meter implementations MUST return an error when multiple Instruments are
+  registered under the same Meter instance using the same name.
+* Different Meters MUST be treated as separate namespaces. The names of the
+  Instruments under one Meter SHOULD NOT interfere with Instruments under
+  another Meter.
+
+Instrument names MUST conform to the following syntax:
```
hdost, line 168: are we restricting to ASCII? or UTF-8?

specification/metrics/new_api.md
```diff
@@ -0,0 +1,202 @@
...
+
+* They are not null or empty strings.
+* They are case-insensitive.
+* The first character must be an alphabetic character.
```
hdost, line 172: Kinda goes with the other comment. is this "[a-zA-Z]" or are we allowing further alphabets?

specification/metrics/new_api.md
```diff
@@ -0,0 +1,202 @@
...
+* Subsequent characters must belong to the alphanumeric characters, '_', '.',
+  and '-'.
+* They can have a maximum length of 63 characters.
```
hdost, line 175: ASCII?

specification/metrics/new_api.md
hdost: > "are identified by their name". Maybe that need some highlights/implciations.

I'm assuming this is related to: #1113

specification/metrics/new_api.md
reyang: I've added RFC5234 ABNF, please check.

specification/metrics/new_api.md
reyang: Yes, I've added RFC5234 ABNF, please check.

specification/metrics/new_api.md
reyang: Removed.

specification/metrics/new_api.md
reyang: Removed.

specification/metrics/new_api.md
reyang: I've added RFC5234 ABNF, please check.

specification/metrics/api.md
jsuereth: We should open a bug to address it across both.  I agree not to impose OO.  Instead this should say three main "components".

@MrAlias Have you been updating the Trace spec when you find similar verbage?

bogdandrutu (approved):
Overall a good starting point

cijothomas (approved):

specification/metrics/api.md
reyang: I've fixed the wording here. For tracing spec, I've created issue #1569 for tracking.

specification/metrics/new_api.md
```diff
@@ -0,0 +1,210 @@
...
+Note: `Meter` SHOULD NOT be responsible for the configuration. This should be
+the responsibility of the `MeterProvider` instead.
+
+### Meter operations
+
+The `Meter` MUST provide functions to create new [Instruments](#instrument):
+
+* [Create a new Counter](#counter-creation) (see the section on `Counter`)
+
+## Instrument
```
noahfalk, line 146: Do we need to say anything about the lifetime of these objects? I'm fine if lifetime is left as the language implementer's choice, or if the spec wants to mandate something specific. I assume the simplest option is everything exists as long as its container exists and MeterProvider exists until the end of the process.

specification/metrics/new_api.md
```diff
@@ -0,0 +1,210 @@
...
+* [Instrument](#instrument) is responsible for reporting
+  [Measurements](#measurement).
+
+Here is an example of the object hierarchy inside a process instrumented with the
+metrics API:
+
+```text
++-- MeterProvider(default)
+    |
+    +-- Meter(name='io.opentelemetry.runtime', version='1.0.0')
```
noahfalk, line 53: I know .NET might be a bit unique here. It's fine if MeterProvider has an API to create a Meter but we were intending that .NET libraries will also be able to create Meter directly (similar to ActivitySource), and that we would recommend them to do so. Thus on .NET MeterProvider needs to be able to subscribe to a Meter that already exists. I don't care whether the spec calls it out, but I want to make sure this isn't taking anyone by surprise in the design process.

reyang: This seems to be the same approach as what `ActivitySource` is doing for trace/span, so I don't see a problem here.

specification/metrics/new_api.md
reyang: I think it is fine to **not** specify the lifetime here (similar like the trace spec).

specification/metrics/new_api.md
```diff
@@ -0,0 +1,210 @@
...
+
+**Owner:**
+
+* [Reiley Yang](https://github.com/reyang)
+
+**Domain Experts:**
+
+* [Bogdan Brutu](https://github.com/bogdandrutu)
+* [Josh Suereth](https://github.com/jsuereth)
+* [Joshua MacDonald](https://github.com/jmacd)
```
arminru, line 13: NB: This is questioned in #1503 and might be removed again.

specification/metrics/new_api.md
noahfalk: OK I'm assuming anything not specified is implicitly implementer's choice : )

PR #1561, opened 2021-03-18
Broaden criteria for populating Zipkin `remoteEndpoint`

### LINKED ISSUE #1560: Criteria for Otel to Zipkin remoteEndpoint transformation should be broadened
The current Otel <-> Zipkin [span mapping](https://github.com/open-telemetry/opentelemetry-specification/blob/b67faa4e0106cac194a837e3d419ae4fab9fc1c4/specification/trace/sdk_exporters/zipkin.md#otlp---zipkin) specifies that Zipkin's `remoteEndpoint` should be specified if Zipkin's `SpanKind` resolves to either `SpanKind.CLIENT` or `SpanKind.PRODUCER`, but omits `SpanKind.CONSUMER`.

>If Zipkin SpanKind resolves to either SpanKind.CLIENT or SpanKind.PRODUCER then the service SHOULD specify remote endpoint otherwise Zipkin won't treat the Span as a dependency. peer.service is the preferred attribute but is not always available. The following table lists the possible attributes for remoteEndpoint by preferred ranking:

Per discussion in https://github.com/open-telemetry/opentelemetry-specification/discussions/1546 , this should be broadened to omit any specific `SpanKind`, and instead require that `remoteEndpoint` be populated wherever possible regardless of `SpanKind`.

It remains a `SHOULD` because there will still be cases where connections from things such as Unix domain sockets will not necessarily have a way to map to `remoteEndpoint`.

evantorrie:
Fixes #1560

## Changes

Broadens the criteria for when to populate Zipkin's `remoteEndpoint` when converting from an `OpenTelemetry` span.

### DIFF
--- a/specification/trace/sdk_exporters/zipkin.md
+++ b/specification/trace/sdk_exporters/zipkin.md
@@ -90,7 +90,7 @@ OpenTelemetry Span's `InstrumentationLibrary` MUST be reported as `tags` to Zipk
 
 ### Remote endpoint
 
-#### OTLP -> Zipkin
+#### OpenTelemetry -> Zipkin
 
 Zipkin's `remoteEndpoint` SHOULD be populated to allow Zipkin to
 correctly treat the Span as a dependency regardless of `SpanKind`.
@@ -113,9 +113,9 @@ attributes for `remoteEndpoint` by preferred ranking:
 * `net.peer.ip` can be used by itself as `remoteEndpoint` but should be combined
   with `net.peer.port` if it is also present.
 
-#### Zipkin -> OTLP
+#### Zipkin -> OpenTelemetry
 
-When mapping from Zipkin to OTLP set `peer.service` tag from `remoteEndpoint`
+When mapping from Zipkin to OpenTelemetry set `peer.service` tag from `remoteEndpoint`
 unless there is a `peer.service` tag defined explicitly.
 
 ### Attribute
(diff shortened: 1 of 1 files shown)


specification/trace/sdk_exporters/zipkin.md
```diff
@@ -90,13 +90,13 @@ OpenTelemetry Span's `InstrumentationLibrary` MUST be reported as `tags` to Zipk
...
 
-#### OTLP -> Zipkin
+#### OpenTelemetry -> Zipkin
 
-If Zipkin `SpanKind` resolves to either `SpanKind.CLIENT` or `SpanKind.PRODUCER`
-then the service SHOULD specify remote endpoint otherwise Zipkin won't treat the
-Span as a dependency. `peer.service` is the preferred attribute but is not
-always available. The following table lists the possible attributes for
-`remoteEndpoint` by preferred ranking:
+Zipkin's `remoteEndpoint` SHOULD be populated to allow Zipkin to
```
anuraaga, line 95: Ah I should have read this in more detail when responding earlier.

An endpoint has three fields, ip, port, service.   Ip and port should always be specified when available. 

Service should only be specified on remote endpoint for client and producer based on this list of prioritized things.

carlosalberto: Ping @evantorrie

PR #1570, opened 2021-03-23
Clarify B3 requirements and configuration

### LINKED ISSUE #1562: B3 Propagation Config
There are essentially three ways to configure b3. The b3 spec defines b3 single and multi-header formats. OTel has its own [flavor of b3](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#b3-requirements) where single and multi-header formats should be extracted, but only single-header should be injected. 

The environment variable specification only defines configuration for two of the three formats mentioned above. It defines `b3` as b3 single-header, and `b3multi` as b3 multi-header, but does not provide a value for the OTel flavor. It might be too late to change this, but in a perfect world, I think the following values would make sense:

* `b3` - OTel flavor b3
* `b3multi` - multi-header b3
* `b3single` - single-header b3

Is this change possible or is there an alternative that we can consider?

mwear:
Fixes #1562

## Changes

This PR updates the spec as discussed in #1562. It clarifies that b3 propagators for single and multi-header formats should extract both formats. The B3 single propagator should give precedence to b3 single when extracting, and inject using the single format only. Similarly, the B3 multi propagator should give precedence to b3 multi extracting and inject using only the multi format.

### DIFF
--- a/specification/context/api-propagators.md
+++ b/specification/context/api-propagators.md
@@ -360,8 +360,9 @@ been established for B3 context propagation.
 When extracting B3, propagators:
 
 * MUST attempt to extract B3 encoded using single and multi-header
-  formats. The single-header variant takes precedence over
-  the multi-header version.
+  formats. When configured for B3 single, the single-header format will take
+  precendence, when configured for B3 multi, the multi-header will take
+  precendence.
 * MUST preserve a debug trace flag, if received, and propagate
   it with subsequent requests. Additionally, an OpenTelemetry implementation
   MUST set the sampled trace flag when the debug flag is set.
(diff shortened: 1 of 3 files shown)


specification/context/api-propagators.md
```diff
@@ -360,8 +360,9 @@ been established for B3 context propagation.
 When extracting B3, propagators:
 
 * MUST attempt to extract B3 encoded using single and multi-header
```
anuraaga, line 362: Can we clarify what `Propagator.fields()` is supposed to return? I think it's the injected headers

specification/context/api-propagators.md
```diff
@@ -360,8 +360,9 @@ been established for B3 context propagation.
...
-  formats. The single-header variant takes precedence over
-  the multi-header version.
+  formats. When configured for B3 single, the single-header format will take
```
yurishkuro, line 363: > When configured for B3 single, the single-header format will take precedence

what else could it do, if it's **configured** for single header?

yurishkuro (approved):

specification/context/api-propagators.md
mwear: I added some guidance. Let me know if it needs any changes.

anuraaga (approved):

carlosalberto (approved):

specification/context/api-propagators.md
```diff
@@ -360,8 +360,9 @@ been established for B3 context propagation.
...
+  precedence, when configured for B3 multi, the multi-header will take
+  precedence.
```
bogdandrutu, line 365: Do we want this change? I thought that we discussed that nothing changes on the extract side.

mwear: We discussed several options, but I think it makes sense to prefer the newer, single header format in either case.

iNikem (approved):

bogdandrutu (approved):

rakyll (approved):

PR #1574, opened 2021-03-23
First cut at single writer principle.

### LINKED ISSUE #1573: Metrics Datamodel: Write "Single-Writer" assumptions to DataModel specification.
- A timeseries instance should have a single writer
- Call out nuance of "Instrument" vs. "Metric" identity
- Call out implications / reasoning
- Call out nuance across metric-in-flight aggregations

jsuereth:
Fixes #1573

## Changes

Adds Single-Writer principle documentation to metrics data model specification


specification/metrics/datamodel.md
```diff
@@ -224,7 +224,45 @@ histograms.
 
 ## Single-Writer
 
-Pending
+All generated TimeSeries from OTLP must have one logical writer.  This means,
+conceptually, that any Timeseries created from the Protocol must have one
+originating source of truth.  In practical terms, this implies the following:
+
+- All metric data points produce by OTel SDKs must by globally unique.
```
aabmass, line 231: Writers might not be OTel SDKs, or did you mean that specifically?

specification/metrics/datamodel.md
```diff
@@ -224,7 +224,45 @@ histograms.
...
+  - `Resource` is expected to uniquely identify the source "service" or
+    "application".
+  - `Metric` name and `Attribute`s are expected to unqiuely identify a
+    timeseries generated by the resource.
```
aabmass, line 235: Copying my comment from [the doc](https://docs.google.com/document/d/1lwLZlQ2RttzZ60N9xhJuJlivs3m5kFjUEZgrfpaF09c/edit#): I believe the InstrumentationLibrary name/version should also be part of single writer requirement. For example, you have two instrumented http libraries in the same process (resource), both producing http.client.duration metric.

specification/metrics/datamodel.md
```diff
@@ -224,7 +224,45 @@ histograms.
...
+- Aggregations of metrics performed "in-motion" must only be written from a
+  single logical source.
+  __Note: This implies all aggregated metrics must reach one destination__.
```
aabmass, line 238: Its not super clear to me what this part means. By "single logical source", do you mean the thing doing the aggregation or the combination of all input resources to the aggregation?

specification/metrics/datamodel.md
jsuereth: I need to figure out a way to specify this.   In the event we "receive" non OTel SDK generated data, we need to provide enough semantics to the data to make it globally unique.  Let me think more about wording.

specification/metrics/datamodel.md
jsuereth: I actually almost added that, but I'm not 100% convinced.

TL;DR:  Do we expect backends to append Resource labels + InstrumentationLibrary-as-labels to gain unique timeseries?  For resource, almost certainly (and in Google Cloud we directly map this concept).   It's less certain to me that's true on Instrumentation library.

specification/metrics/datamodel.md
jsuereth: You need the output of your OTLP to look like all the metrics went to one source and were aggregated/written by one process.   Whether you ACTUALLY do that isn't important, but the logical assumption is there.

I.e. only one thing can write a Timeseries implies if I'm synthesizing new timeseries by aggregating, then I only do that in one place for the same timeseries, meaning we have simpler error handling + logic for dealing with errors.

specification/metrics/datamodel.md
```diff
@@ -224,7 +224,45 @@ histograms.
 
 ## Single-Writer
 
-Pending
+All generated TimeSeries from OTLP must have one logical writer.  This means,
```
rakyll, line 227: Is there any background context about the single-writer behavior and what it is trying to solve?

specification/metrics/datamodel.md
```diff
@@ -224,7 +224,45 @@ histograms.
...
+conceptually, that any Timeseries created from the Protocol must have one
+originating source of truth.  In practical terms, this implies the following:
```
rakyll, line 229: "Timeseries created from the Protocol"

What does this mean? Timeseries created by ingesting the protocol?

specification/metrics/datamodel.md
```diff
@@ -224,7 +224,45 @@ histograms.
...
+- All metric data points produce by OTel SDKs must by globally unique.
+  - `Resource` is expected to uniquely identify the source "service" or
+    "application".
+  - `Metric` name and `Attribute`s are expected to unqiuely identify a
+    timeseries generated by the resource.
+- Aggregations of metrics performed "in-motion" must only be written from a
+  single logical source.
+  __Note: This implies all aggregated metrics must reach one destination__.
+
+In real systems, particularly in misbehaving systems, there is in fact the
```
rakyll, line 240: We don't need to include any subjective statements in specs such as "real" vs "misbehaving" systems. From a reader perspective, it's not clear here what's the unwanted behavior. You can just explain the misbehaving case instead.

specification/metrics/datamodel.md
rakyll: Maybe saying something like "globally uniquely produced and free from duplicates" might be better wording.

specification/metrics/datamodel.md
```diff
@@ -224,7 +224,45 @@ histograms.
 
 ## Single-Writer
 
-Pending
+All generated TimeSeries from OTLP must have one logical writer.  This means,
+conceptually, that any Timeseries created from the Protocol must have one
+originating source of truth.  In practical terms, this implies the following:
+
+- All metric data points produce by OTel SDKs must by globally unique.
+  - `Resource` is expected to uniquely identify the source "service" or
```
rakyll, line 232: I'd give a link to the service attributes here and avoid mentioning "application" in quotes. There is no semantic convention family for applications.

specification/metrics/datamodel.md
```diff
@@ -224,7 +224,45 @@ histograms.
...
+    "application".
+  - `Metric` name and `Attribute`s are expected to unqiuely identify a
+    timeseries generated by the resource.
+- Aggregations of metrics performed "in-motion" must only be written from a
+  single logical source.
+  __Note: This implies all aggregated metrics must reach one destination__.
+
+In real systems, particularly in misbehaving systems, there is in fact the
+possibility of multiple writers sending data for the same metric and attribute
+set.  This requirement states that receivers SHOULD presume a single writer was
```
rakyll, line 242: This case is duplication and the elimination is deduplication, right? If I'm not wrong, would it make sense to stick to this terminology because it's already commonly known/used.

specification/metrics/datamodel.md
```diff
@@ -224,7 +224,45 @@ histograms.
...
+intended and eliminate overlap when converting OTLP into the Timeseries data
+model.
+
+### Overlap resolution
+
+When more than one process writes the same timeseries, OTLP data points may
+appear to overlap. This condition typically results from misconfiguration, but
+can also result from running identical processes (indicative of operating system
```
rakyll, line 250: Maybe we can give a concrete example here such as missing process attributes? https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/resource/process.yaml

specification/metrics/datamodel.md
```diff
@@ -224,7 +224,45 @@ histograms.
...
+or SDK bugs). When there are overlapping points, receivers SHOULD eliminate
```
rakyll, line 251: How does this work for Prometheus? Not deduping would be fine (and might be used for redundancy) because everything is cumulative/absolute. Do we care about this behavior for all or is it only for deltas?

specification/metrics/datamodel.md
jsuereth: Yeah, we discussed this extensively in the DataModel SiG today.  Going to reword this and provide a new concept "Metric Stream" which represents a TimeSeries-in-transit via OTLP.

Effectively we need a "thing" to identify that can be aggregated against or manipulated and has identity.   Effectively, most mappings from OTLP => TimeSeries data storage should have 1:1 conversion from Metric Stream => TimeSeries.

specification/metrics/datamodel.md
jsuereth: Like the phrasing, thanks @rakyll

specification/metrics/datamodel.md
jsuereth: Right, and service isn't broad enough to denote what we mean here IMO.  I'll just use service with a link as suggested, as the Resource debate belongs in a broader forum.

specification/metrics/datamodel.md
jsuereth: Agreed, reworded PTAL.

specification/metrics/datamodel.md
jsuereth: I'll let @jmacd comment here.   From what I understand this can cause issues in prometheus, but I wasn't able to get any specifics. I'll keep doing some investigation for a better answer here.

specification/metrics/datamodel.md
jsuereth: I added some background at the bottom of this section (this is a specification, so I wasn't sure where background belongs).

specification/metrics/datamodel.md
```diff
@@ -224,7 +238,53 @@ histograms.
...
+- All metric data streams produce by OTel SDKs must by globally uniquely produced
+  and free from duplicates.
+- Aggregations of metric streams must only be written from a single logical
+  source.
+  __Note: This implies aggregated metric streams must reach one destination__.
+
+In systems, there is the possibility of multiple writers sending data for the
+same metric stream (duplication).  This is considered an error state, or
+misbehaving system. Receivers SHOULD presume a single writer was intended and
+eliminate overlap / deduplicate.
```
yurishkuro, line 254: I am not sure I follow this. If I have N service instances each incrementing a counter with the exact same name and, for simplicity, no other dimensions, that is a perfectly valid use case in most of existing systems today, but it seems to contradict this section.

specification/metrics/datamodel.md
```diff
@@ -140,9 +140,9 @@ OpenTelemetry fragments metrics into three interacting models:
 
 - An Event model, representing how instrumentation reports metric data.
 - A TimeSeries model, representing how backends store metric data.
-- The *O*pen*T*e*L*emetry *P*rotocol (OTLP) data model representing how metrics
-  are manipulated and transmitted between the Event model and the TimeSeries
-  storage.
+- The *O*pen*T*e*L*emetry *P*rotocol (OTLP) data model representing how metric
+  data streams are manipulated and transmitted between the Event model and the
+  TimeSeries storage.
```
tigrannajaryan, line 145: Inconsistent capitalization of TimeSeries.

specification/metrics/datamodel.md
```diff
@@ -199,7 +199,20 @@ further development of the correspondence between these models.
 
 ### OpenTelemetry Protocol data model
 
-The OpenTelemetry data model for metrics includes four basic point kinds, all of
+The OpenTelmetry protocol data model is composed of Metric data streams.  These
```
tigrannajaryan, line 202:
```suggestion
The OpenTelmetry protocol data model is composed of Metric data streams. These
```

specification/metrics/datamodel.md
```diff
@@ -199,7 +199,20 @@ further development of the correspondence between these models.
...
+streams are in turned composed of metric data points.   Metric data streams
```
tigrannajaryan, line 203:
```suggestion
streams are in turn composed of metric data points. Metric data streams
```

specification/metrics/datamodel.md
```diff
@@ -199,7 +199,20 @@ further development of the correspondence between these models.
...
+can be converted directly into timeseries, and share the same identity
+characteristics for a timeseries.   A metric stream is identified by:
+
+- The originating `Resource`
+- The metric stream's `name`.
+- The attached `Attribute`s
+- The metric stream's point kind.
```
tigrannajaryan, line 210: Is the point kind really intended to be part of the identity? Is it possible to have Timeseries with the exact same set of the above 3 elements and only differing in the point kind? Do any/most metric backends that exist today support this notion?

specification/metrics/datamodel.md
```diff
@@ -224,7 +238,53 @@ histograms.
 
 ## Single-Writer
 
-Pending
+All metric data streams within OTLP must have one logical writer.  This means,
```
tigrannajaryan, line 241: It would be useful to explain how it is possible to end up with more than one logical writer, why is it a problem that needs a solution.

specification/metrics/datamodel.md
```diff
@@ -224,7 +238,53 @@ histograms.
...
+originating source of truth.  In practical terms, this implies the following:
+
+- All metric data streams produce by OTel SDKs must by globally uniquely produced
+  and free from duplicates.
+- Aggregations of metric streams must only be written from a single logical
+  source.
+  __Note: This implies aggregated metric streams must reach one destination__.
+
+In systems, there is the possibility of multiple writers sending data for the
+same metric stream (duplication).  This is considered an error state, or
```
tigrannajaryan, line 252: Are we talking about duplication of the same data points or different data points of the same Event source, data points which possibly originate at different times?

specification/metrics/datamodel.md
tigrannajaryan: @yurishkuro I think (and I may misunderstand too) what this intends to say is that the measurements of that counter should be done such that the resulting written Data Stream is done by a single entity which guarantees there are no duplicate measurements. The act of incrementing the counter in your example is the what this document calls an "Event", which is not limited to single writer. The act of measuring and producing "Data Stream" from such event is what is required to behave according to what is called a "single writer" in the doc.

@jsuereth it would be great to clarify the document. It is not easy to understand. Some examples would help. Expand it a bit.

specification/metrics/datamodel.md
```diff
@@ -224,7 +238,53 @@ histograms.
...
+well-behaved systems can perform metric stream manipulation without undesired
+degradation or loss of visibility.
+
+### Overlap resolution
+
+When more than one process writes the same metric data stream, OTLP data points
+may appear to overlap. This condition typically results from misconfiguration, but
+can also result from running identical processes (indicative of operating system
+or SDK bugs, like missing
+[process attribnutes](../resource/semantic_conventions/process.md)). When there
```
tigrannajaryan, line 273:
```suggestion
[process attributes](../resource/semantic_conventions/process.md)). When there
```

specification/metrics/datamodel.md
```diff
@@ -224,7 +238,53 @@ histograms.
...
+are overlapping points, receivers SHOULD eliminate points so that there are no
```
tigrannajaryan, line 274: It would be useful to define what exactly "overlapping" means.

specification/metrics/datamodel.md
```diff
@@ -224,7 +238,53 @@ histograms.
...
+overlaps. Which data to select in overlapping cases is not specified.
+
+### Overlap observability
+
+OpenTelemetry collectors SHOULD export telemetry on the appearance of
+overlapping points, so that the user can monitor for erroneous configurations.
```
tigrannajaryan, line 280:
```suggestion
OpenTelemetry collectors SHOULD export telemetry when they observe
overlapping points in data streams, so that the user can monitor for erroneous configurations.
```

specification/metrics/datamodel.md
```diff
@@ -224,7 +238,53 @@ histograms.
 
 ## Single-Writer
 
-Pending
+All metric data streams within OTLP must have one logical writer.  This means,
+conceptually, that any Timeseries created from the Protocol must have one
+originating source of truth.  In practical terms, this implies the following:
+
+- All metric data streams produce by OTel SDKs must by globally uniquely produced
```
tigrannajaryan, line 245: I am not sure I like this sweeping limitation. This prohibits use cases when multiple distributed producers may contribute delta data points that add up to a valid total counter. I heard about such uses cases, probably very rare, but it exists.

specification/metrics/datamodel.md
```diff
@@ -140,9 +140,9 @@ OpenTelemetry fragments metrics into three interacting models:
 
 - An Event model, representing how instrumentation reports metric data.
 - A TimeSeries model, representing how backends store metric data.
-- The *O*pen*T*e*L*emetry *P*rotocol (OTLP) data model representing how metrics
-  are manipulated and transmitted between the Event model and the TimeSeries
-  storage.
+- The *O*pen*T*e*L*emetry *P*rotocol (OTLP) data model representing how metric
+  data streams are manipulated and transmitted between the Event model and the
```
yurishkuro, line 144: what are "metric data streams"? Never defined till now.

specification/metrics/datamodel.md
jsuereth: We discussed this heavily in the SiG.  TL;DR: it's likely an error scenario/misbehaving system that would have the same timeseries with different point types.   For the purposes of OTel + Aggregation we do NOT unify them and leave that option to backends to determine the right thing to do.

This isn't about preventing backends from supporting it, but allowing them to do so and not forcing OTel to specify some kind of behavior that works with backends which do not.

tigrannajaryan: >This isn't about preventing backends from supporting it, but allowing them to do so and not forcing OTel to specify some kind of behavior that works with backends which do not.

On the other hand, if we explicitly declare that this is a valid situation OpenTelemetry-compliant metric sources may start emitting it causing problems for backends that don't expect to see such data. I think this is the case when we need to have a stance and that stance has a consequence for backends. We either explicitly prohibit Timeseries with identical (resource,name,attributes) and different data points or we explicitly allow it, which means backend have to support it otherwise they can't accept OpenTelemetry-compliant data.

I can see how allowing this can be valuable (e.g. I may export a gauge and a histogram for the same instrument with exact same name). My concern here is that if we allow this we may suddenly put vendors in a situation where they can't really accept this data.

Just to be clear: I don't know if this is really a problem for backends. If you know that it is commonly accepted in the industry to emit such timeseries, then I don't see a problem.

jsuereth: On thing I'm not sure I made clear (and it was crystal clear in the discussion).

1. The same 3 elements but differing point kind coming out of an SDK is considered an "error state" that should be handled by an SDK.
2. The treatment here (and discussion) should be about what happens if the *collector* (or downstream processor) receives two streams that only differ by point type.   The collector treats them as separate strings.
3. Backends have the option to issue an error *or* try to handle the scenario.   For the situation where the differences are floats vs. ints we think it's common to do conversion (and may plan to define our own "weak conformance" conversion for the collector).  We think the backend has the *best* chance of issuing a good error message notifying the user they have two independent metric sources reported as if they are the same with different metric types.

TL;DR; This *is* an error state, we just describe here how the "model" treats that error scenario so progress can be made and the best component has a chance to give the best error message.

jmacd (approved):

specification/metrics/datamodel.md
aabmass: @jmacd any thought on this?

specification/metrics/datamodel.md
```diff
@@ -34,16 +34,16 @@ Prometheus Remote Write protocol without loss of features or semantics, through
 well-defined translations of the data, including the ability to automatically
 remove attributes and lower histogram resolution.
 
-## Events → Data → Timeseries
+## Events → Data Stream → Timeseries
 
 The OTLP Metrics protocol is designed as a standard for transporting metric
 data. To describe the intended use of this data and the associated semantic
-meaning, OpenTelemetry metric data types will be linked into a framework
+meaning, OpenTelemetry metric data stream types will be linked into a framework
```
yurishkuro, line 41: > OpenTelemetry metric data stream

never defined, yet imbued with special meaning in this document

specification/metrics/datamodel.md
```diff
@@ -224,11 +238,109 @@ histograms.
 
 ## Single-Writer
 
-Pending
+All metric data streams within OTLP must have one logical writer.  This means,
+conceptually, that any Timeseries created from the Protocol must have one
+originating source of truth.  In practical terms, this implies the following:
+
+- All metric data streams produce by OTel SDKs must by globally uniquely
+  produced and free from duplicates.   All metric data streams can be uniquely
```
yurishkuro, line 246: > free from duplicates

What is considered a duplicate?

specification/metrics/datamodel.md
```diff
@@ -224,11 +238,109 @@ histograms.
...
+  identified in some way.
+- Aggregations of metric streams must only be written from a single logical
+  source.
+  __Note: This implies aggregated metric streams must reach one destination__.
```
yurishkuro, line 250: The part in bold kinda makes sense to me, but I don't see why this is a requirement of the protocol. The part about single source doesn't make sense to me.

I think protocol needs to support data in such a format that it can be repeatedly re-aggregated, if necessary.

specification/metrics/datamodel.md
jsuereth: It's defined on line 202, and yes this document imbues it with special meaning.   We decided in the metric data model SiG that we needed  a term to identify an "in-motion" stream of metric data that OpenTelemetry operates on to differentiate it from a Time Series generated from that stream.

Should I move the definition/description of metric data stream further above?  Do you need more motivation for why we want a different term here?

Put simply, a lot of your concerns/comments are around generated TimeSeries from open-telemetry, whereas this document outlines what's required for a Metric Data Stream.  See my follow on comments for examples where I think having this distinction helps opentelemetry AND backends.

specification/metrics/datamodel.md
```diff
@@ -199,7 +199,20 @@ further development of the correspondence between these models.
 
 ### OpenTelemetry Protocol data model
 
-The OpenTelemetry data model for metrics includes four basic point kinds, all of
+The OpenTelmetry protocol data model is composed of Metric data streams. These
+streams are in turn composed of metric data points. Metric data streams
+can be converted directly into Timeseries, and share the same identity
+characteristics for a Timeseries.   A metric stream is identified by:
+
+- The originating `Resource`
```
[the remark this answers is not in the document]
jsuereth: I think you're taking the view that a TimeSeries generated from OTLP doesn't need the resource.  That's completely fine, we're documenting what a Metric Data Stream is and how it's identified.  Backends can map these to timeseries however they wish (as is done today), albeit OTel should come with a recommended conversion.


What this specification means is that *within* OTLP you'll have a resource that's part of your identity (which is inherently true by protocol structure).

specification/metrics/datamodel.md
jsuereth: Added description above.

specification/metrics/datamodel.md
jsuereth: updated to call these Metric Data Streams.

specification/metrics/datamodel.md
jsuereth: > Our stance for the Collector so far has been that by default it is not in the business of interpreting data or "treating" it in any way.
> It only does so when explicitly configured by the user to do so, or when a particular non-OTLP format is used and a translation must happen.
>
> Do you have a different expectation from the Collector?

I think some of the built-in operations (label removal, aggregations, etc.) that are available in the baseline image should be clearly specified against the Data Model.   Yes, I agree users configure these things, but that doesn't mean we don't clearly outline what they do and ensure their "compatibility" in shape to lead to the least-possible issues with exporters. 

Specifically:  I think some generally-available collector operations should be specified in a way that they can be used without causing issues across Exporters.  If users provide their own processors that do crazier things outside this model, that's fine and outside any guarantees.

> Sounds good, makes sense to me. I may have missed it, but I do not see this in the text. Can we explicitly call out that this is an error state in the text?

Thought that I had, let me add that now.

specification/metrics/datamodel.md
jsuereth: I'm not sure how this requirement leads to data that can't be repeatedly re-aggregated.   Can you provide a counter example where it's an issue?  AFAIK you should be able to continually re-aggregate assuming:

1. All incoming streams are uniquely identified (or could be logically identified, e.g. StatsD incoming counters)
2. All reaggregations output as a new identity (e.g. different attributes or resource)

In practice that's actually a pretty easy requirement.  For backends where you drop resource, you're back to aggregated what looks like the same timeseries from multiple sources but within OTLP we can identify their separate sources.

specification/metrics/datamodel.md
jsuereth: This was from the @jmacd document. I *think* this is synonymous with "overlap", defined below.  I can update but want to check with Josh to make sure I haven't missed an important point here.

specification/metrics/datamodel.md
```diff
@@ -199,7 +199,24 @@ further development of the correspondence between these models.
...
+- The metric stream's `name`.
```
victlu, line 208: On Line 180 you have ` Metric name and description`.  I assume description does not play any part of the identity of a Time Series or OTLP Data Model Stream.

victlu: Since we're in this section.  Do you see "Unit" as a part of identity for data model stream? Or are the units already normalized prior?

specification/metrics/datamodel.md
```diff
@@ -224,11 +242,109 @@ histograms.
 
 ## Single-Writer
 
-Pending
+All metric data streams within OTLP must have one logical writer.  This means,
+conceptually, that any Timeseries created from the Protocol must have one
+originating source of truth.  In practical terms, this implies the following:
+
+- All metric data streams produce by OTel SDKs must by globally uniquely
```
reyang, line 249:
```suggestion
- All metric data streams produced by OTel SDKs must by globally uniquely
```

specification/metrics/datamodel.md
```diff
@@ -224,11 +242,109 @@ histograms.
...
+same resource.  In this case, metrics will be reported at inconsistent time
+intervals.  For metrics like cumulative sums, this could cause issues where
+pairs of points appear to reset the cumulative sum leading to unusable metrics.
+
+Multiple writers for a metric stream is considered an error state, or
+misbehaving system. Receivers SHOULD presume a single writer was intended and
+eliminate overlap / deduplicate.
+
+Note: Identity is an important concept in most metrics systems.  For example,
+[prometheus directlly calls out uniqueness](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#metric_relabel_configs):
```
reyang, line 269:
```suggestion
[Prometheus directly calls out uniqueness](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#metric_relabel_configs):
```

reyang (approved):
LGTM with couple minor suggestions.

specification/metrics/datamodel.md
jsuereth: Yeah, let me reword that, it should just be the metric name.

I don't see Unit as part of the data stream identity, at least currently.   I think units are a bit under-specified, and more "documentational" rather than "annotational" or "semantic" at this point.

Given this definition of metric identity, I'd consider unit-mismatch to be an "error" right now, something an SDK should not allow.   While I think you raise some good questions for consideration of X-SDK unit treatment and mixing of data streams with different units, I think that's out of scope for the current PR/discussion and we should be able to define something more robust in the future.

specification/metrics/datamodel.md
```diff
@@ -224,11 +242,109 @@ histograms.
...
+
+> Care must be taken with `labeldrop` and `labelkeep` to ensure that metrics
+> are still uniquely labeled once the labels are removed.
+
+For OTLP, the Single-Writer principle grants a way to reason over error
+scenarios and take corrective actions.  Additionally, it ensures that
+well-behaved systems can perform metric stream manipulation without undesired
+degradation or loss of visibility.
 
 ## Temporarily
```
maxedmands, line 279: Yikes, I think this should be 
```suggestion
## Temporality
```

specification/metrics/datamodel.md
```diff
@@ -224,11 +242,109 @@ histograms.
...
+Overlap occurs when more than one metric data point occurs for a data stream
+within a time window.   This is particularly problematic for data points meant
+to represent an entire time window, e.g. a Histogram reporting population
+density of collected metric data points for a time window.  If two of these show
+up with overlapping time windows, how do backends handle this situation?
+
+We define three principles for handling overlap:
+
+- Resolution (correction via dropping points)
+- Obersvability (allowing the data to flow to backends)
```
maxedmands, line 323:
```suggestion
- Observability (allowing the data to flow to backends)
```

specification/metrics/datamodel.md
jmacd: I agree with everything being said here. Different point kinds should be considered errors. I agree that the collector's default behavior should be to simply pass the data through. We are trying to lay out correctness conditions for aggregating data and want definitions that help aggregators do the Right Thing when these errors arise.

It seems to me that metric name, the resource, and metric attributes are first class identifiers for metric streams. The point kind and unit are somehow in a different category: these should be treated as distinct for the purposes of pass-through data processing, but should be treated as errors for aggregation.

I'm not sure we have the proper terminology to describe properties of a metric stream that are identifying but erroneous vs identifying and compatible. 

Also I think we may be distracted by trying to define what is and is not an error, where we were trying to define what is a stream and its single-writer property. We may have a set of individually valid streams with single writers, that when considered together form an erroneous condition. Example: when two streams have the same name and resource, but distinct point kind and label values. Each stream is valid, but we should never see a mixture of point kinds for a metric name, independent of how many streams it writes.

specification/metrics/datamodel.md
jmacd: I am thinking of "duplicate" points as points in a stream with identical start and end timestamps, whereas overlapping points are defined based on the temporality. Cumulative points overlap, in one sense, but two points are compatible when they share a start time. Two cumulative streams would have an overlap condition when multiple start times are mixed with overlapping end times, however. Gauge (instantaneous temporality) and Delta temporality points have different overlap rules.

Still, I think duplicate points are in a class of their own, particularly when we begin to talk about external labels as Prometheus defines them. Duplicate points may be expected, and when they are truly duplicates we typically discard all but one.

specification/metrics/datamodel.md
jmacd: We are trying to define a "single-writer" property so that we know when it is safe to re-aggregate data. There are ways to process metrics data that remove the single-writer property (e.g., horizontal scaling) and there are ways to restore that property (e.g., having a message-queue in place) to allow re-aggregation.

We are trying to ensure that metrics originate from a single-writer so that we know these transformations will work. For these rules to work, we have to be sure that multiple processes cannot somehow write to the same stream, because that would mean they can overwrite each other. Thus, as @jsuereth implies, we will generate new metric streams following re-aggregation.

jmacd (approved):

specification/metrics/datamodel.md
aabmass: Feel free to resolve this comment (I can't for some reason).

specification/metrics/datamodel.md
```diff
@@ -224,11 +242,109 @@ histograms.
...
+
+The use of delta temporality for metric sums is also common, exemplified by
+Statsd. There is a connection between OpenTelemetry tracing, in which a Span
+event commonly is translated into two metric events (a 1-count and a timing
+measurement). Delta temporality enables sampling and supports shifting the cost
+of cardinality outside of the process.
+
+## Overlap
+
+Overlap occurs when more than one metric data point occurs for a data stream
```
aabmass, line 314: IIUC, overlap is the violation of single-writer requirement. Just to link this together with the single writer section, can we call that out?

specification/metrics/datamodel.md
```diff
@@ -224,11 +242,109 @@ histograms.
...
+
+### Overlap resolution
+
+When more than one process writes the same metric data stream, OTLP data points
+may appear to overlap. This condition typically results from misconfiguration, but
+can also result from running identical processes (indicative of operating system
+or SDK bugs, like missing
+[process attributes](../resource/semantic_conventions/process.md)). When there
+are overlapping points, receivers SHOULD eliminate points so that there are no
+overlaps. Which data to select in overlapping cases is not specified.
```
aabmass, line 334: L333
> receivers SHOULD eliminate points so that there are no overlaps

and L338
> OpenTelemetry collectors SHOULD export telemetry when they observe overlapping points in data streams

Don't these conflict each other, or is "receivers" meant to mean backends specifically?

specification/metrics/datamodel.md
```diff
@@ -224,11 +242,109 @@ histograms.
...
+OpenTelemetry collectors SHOULD export telemetry when they observe overlapping
+points in data streams, so that the user can monitor for erroneous
+configurations.
+
+### Overlap interpolation
+
+When one process starts just as another exits, the appearance of overlapping
+points may be expected. In this case, OpenTelemetry collectors SHOULD modify
+points at the change-over using interpolation for Sum data points, to reduce
+gaps to zero width in these cases, without any overlap.
```
aabmass, line 347: As resource semantic conventions are specified with `service.instance.id`, I'm not sure if this is expected or a misconfiguration:
> [`service.namespace,service.name,service.instance.id` triplet MUST be globally unique](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.1.0/specification/resource/semantic_conventions/README.md#user-content-service:~:text=MUST%20be%20unique%20for%20each%20instance,service.namespace%2Cservice.name%2Cservice.instance.id%20triplet%20MUST%20be%20globally%20unique)

So maybe this falls under the advice above – "collectors SHOULD export telemetry when they observe overlapping
points in data streams, so that the user can monitor for erroneous
configurations"?

aabmass (approved):

specification/metrics/datamodel.md
victlu: Question: if we get multiple point kinds with same metric name, how would we know which stream is erroneous?  Do we need to register the point-kind + metric name prior?

victlu (approved):

specification/metrics/datamodel.md
jsuereth: @victlu It's likely impossible for 'open telemetry' to know which is erroneous.  a backend that supports schema'd metrics could use the currently registered metric schema to determine which one is 'stale' and ingest the correct one, but the key here is this is an "error' scenario.

SergeyKanzhelev:
@yurishkuro are you satisfied with answers?

SergeyKanzhelev (approved):

SergeyKanzhelev:
this PR reached the required number of approvals and the company diversity of approvals. Also it is active for enough time. Merging.

Issue #764, opened 2020-08-07
Exception semantic conventions and Go error return values


Oberon00:
See https://github.com/open-telemetry/opentelemetry-specification/pull/761#discussion_r466313064.

For Go errors, see https://blog.golang.org/go1.13-errors. Go errors are not exceptions, as the defining characteristic is that it has some kind of special throw/try/catch control flow that facilitates separating the "happy path" from the "error path". 

Currently we seem to implicitly encourage this in https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/exceptions.md#stacktrace-representation that Go should use [`runtime.debug.Stack`](https://golang.org/pkg/runtime/debug/#Stack) (a function that returns the current stack, not directly associated with error handling or error objects).

On the other hand, go also has panics https://blog.golang.org/defer-panic-and-recover which seem to match exceptions far better.

Rust may require similar treatment (it also has panics and a `Result` type), and some C++ applications (using Boost.Outcome, std::expected, etc).

I can think of two solutions:
1. Make clear that Go panics are fine to use with the exception conventions but Go errors are not.
2. Describe in more detail how Go errors are mapped to exceptions.

Oberon00:
CC @open-telemetry/go-approvers

Aneurysm9:
I think that it makes sense to treat panic/recover like throw/catch.  Errors are definitely not exceptions, though sometimes they are used similarly, even if only for the fact that they can be easily (even inadvertently) ignored, which stops propagation through the call stack.

tigrannajaryan:
This is a blocker for Go, but is this a blocker for the spec? Do we believe that while working on this we may find that Go requires something that makes current exception specification impossible to use?

Oberon00:
I don't think so. We might just discover that Go needs additional attributes or a different convention. E.g. maybe Go will use an event attribute "unwinding=false" to distinguish panics from error returns.

andrewhsu:
@open-telemetry/go-maintainers ptal, fyi this one was moved to `after-ga`, and respond if this is not desirable.

MrAlias:
[...]

PR #1576, opened 2021-03-23
Update spec-compliance-matrix RecordException for Go

MrAlias:
See https://github.com/open-telemetry/opentelemetry-specification/issues/764 for more detail.

Resolves https://github.com/open-telemetry/opentelemetry-go/issues/1595

### DIFF
--- a/spec-compliance-matrix.md
+++ b/spec-compliance-matrix.md
@@ -64,8 +64,8 @@ status of the feature is not known.
 | Add order preserved                                                                              |          | +  | +    | +  | +      | +    | +      | +   | +    | +   | +    | +     |
 | Safe for concurrent calls                                                                        |          | +  | +    | +  | +      | +    | +      | +   | +    | +   | +    | +     |
 | [Span exceptions](specification/trace/api.md#record-exception)                                   |          |    |      |    |        |      |        |     |      |     |      |       |
-| RecordException                                                                                  |          | -  | +    | +  | +      | +    | -      |     | +    | -   | +    | -     |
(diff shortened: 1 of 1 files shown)


tigrannajaryan:
Was there a decision made that Go won't have RecordException? Reading https://github.com/open-telemetry/opentelemetry-specification/issues/764 I cannot see a resolution.

carlosalberto (approved):

MrAlias:
@tigrannajaryan it is still an active discussion in https://github.com/open-telemetry/opentelemetry-go/issues/1491. We would like to provide idiomatic handling of errors and panics but it is still unclear how this will map to exceptions. We plan to update this if things change.

tigrannajaryan:
> @tigrannajaryan it is still an active discussion in [open-telemetry/opentelemetry-go#1491](https://github.com/open-telemetry/opentelemetry-go/issues/1491). We would like to provide idiomatic handling of errors and panics but it is still unclear how this will map to exceptions. We plan to update this if things change.

@MrAlias Should we wait for the outcome of that discussion before marking this N/A in case it turns out not to be N/A? Or do you already know that it is definitely going to be a different approach for Go, so this line item is N/A no matter what is eventually decided?

MrAlias:
> @MrAlias Should we wait for the outcome of that discussion before marking this N/A in case it turns out not to be N/A? Or do you already know that it is definitely going to be a different approach for Go, so this line item is N/A no matter what is eventually decided?

Sure, sounds good :+1:

PR #1578, opened 2021-03-25
Adding the Metrics Counter API

reyang:
## Changes

During the [03/18/2021 Metrics API/SDK SIG Meeting](https://docs.google.com/document/d/1-bCYkN-DWJq4jw1ybaDZYYmx-WAe6HnwfWbkm8d57v8/edit#heading=h.raqds2xgauu0), we've agreed to focus on `Counter` as the 1st step to tackle the instruments.

This will be discussed during the upcoming (3/25) Metrics API/SDK SIG meeting.

Related issues #


specification/metrics/new_api.md
```diff
@@ -184,8 +186,54 @@ DIGIT = %x30-39 ; 0-9
...
+desired, the client can decide the language idomatic name(s), for example
+`CreateUInt64Counter`, `CreateDoubleCounter`, `CreateCounter<UInt64>`,
+`CreateCounter<double>`.
+
+The API MUST accept the following parameters:
+
+* The `name` of the Instrument, following the [instrument naming
+  rule](instrument-naming-rule).
+* An optional `unit of measure`.
+* An optional `description`.
```
reyang, line 212: To be discussed during the 3/25 Metrics SIG meeting - do we want to put restriction/limitation on `description`? (maximum length, allowed characters, etc.)

specification/metrics/new_api.md
```diff
@@ -184,8 +186,54 @@ DIGIT = %x30-39 ; 0-9
...
+[`Meter`](#meter). This MAY be called `CreateCounter`. If strong type is
+desired, the client can decide the language idomatic name(s), for example
+`CreateUInt64Counter`, `CreateDoubleCounter`, `CreateCounter<UInt64>`,
+`CreateCounter<double>`.
+
+The API MUST accept the following parameters:
+
+* The `name` of the Instrument, following the [instrument naming
+  rule](instrument-naming-rule).
+* An optional `unit of measure`.
```
reyang, line 211: To be discussed during the 3/25 Metrics SIG meeting - do we want to API/SDK to have understanding of unit (e.g. any validation, constraints), or it is simply an opaque string?

From previous meeting notes:

Unit https://github.com/open-telemetry/opentelemetry-specification/issues/705
Josh - the unit is part of the unique name of instrument, the SDK will NOT understand unit
Bogdan - SDK don’t handle unit conversion, backend might want to do it
Josh - https://github.com/open-telemetry/opentelemetry-specification/pull/1177

jsuereth: My $.02 -

1. We should not do anything fancy with units right away
2. We should only discuss "what would prevent new features later"

I think this should be ok as-specified and there's all sorts of clever ways to make units more useful over time.

specification/metrics/new_api.md
jsuereth: If we tie this to OTLP, we're restricted to UTF-8 unless we provide Instrument Name => Metric Stream name conversions that go from ?? => UTF-8.

specification/metrics/new_api.md
```diff
@@ -184,8 +186,54 @@ DIGIT = %x30-39 ; 0-9
...
+##### Add
+
+Increment the Counter by a fixed amount.
+
+This API SHOULD NOT return a value (it MAY return a dummy value if required by
+certain programming languages or systems, for example `null`, `undefined`).
+
+Required parameters:
+
+* The [attributes](../common/common.md#attributes)
```
jsuereth, line 227: Shouldn't these be optional given the caveat below?

Either that or we need to rephrase the wording of the last paragraph in this section to make it more clear what to do here.

jmacd (approved):

specification/metrics/new_api.md
reyang: Discussed during the SIG meeting, we'll document that the unit is opaque string, we'll put 64 chars as the maximum size.

specification/metrics/new_api.md
reyang: Clarify description is a free form thing, it has no particular meaning (which means user should not take dependency).
We can link to the OTEP.

specification/metrics/new_api.md
```diff
@@ -184,8 +186,54 @@ DIGIT = %x30-39 ; 0-9
...
+`CreateUInt64Counter`, `CreateDoubleCounter`, `CreateCounter<UInt64>`,
+`CreateCounter<double>`.
+
+The API MUST accept the following parameters:
+
+* The `name` of the Instrument, following the [instrument naming
+  rule](#instrument-naming-rule).
+* An optional `unit of measure`.
+* An optional `description`.
+* An optional list of [`Attribute`](../common/common.md#attributes) names and
```
reyang, line 213: jmacd: suggest that we don't enforce the user to predefine everything (attribute names and values) in order to report data point.

reyang: noahfalk: default value, static values (prom-client)

reyang: Weyert: Maybe offer something like OTEL_STRICTNESS_LEVEL=strict | relaxed?
then the user could opt-in for more strictness were everything needs OTel be defined ahead of time

reyang: noahfalk: probably we can use "Hints"

specification/metrics/new_api.md
reyang: I've updated the PR:
1. clarified that description must be treated as opaque string from the API and SDK
2. required the API/SDK to support Unicode BMP at minimum
3. put a size limitation

specification/metrics/new_api.md
reyang: I've added a statement that "The API MUST allow callers to provide flexible attributes at invocation time rather than having to register all the possible attribute names during the instrument creation."

specification/metrics/new_api.md
reyang: Nice catch, I've fixed this and also added some descriptions and examples.

specification/metrics/new_api.md
```diff
@@ -182,10 +184,115 @@ DIGIT = %x30-39 ; 0-9
...
+items_sold_counter = meter.create_counter(name="item_sold", description="number of items sold", value_type=int)
+
+items_sold_counter.Add(2, {"item": "Tomato", "customer": "Tom"})
+items_sold_counter.Add(3, item="Tomato", customer="Jerry"})
+```
+
+```csharp
+// C#
+
+var counterItemsSold = meter.CreateCounter<UInt64>("item_sold", description="number of items sold");
```
bogdandrutu, line 281: In java we went with non-generic version to avoid boxing/unboxing. Is that a problem in C# as well?

specification/metrics/new_api.md
```diff
@@ -165,6 +165,8 @@ the name:
   Instruments under one Meter SHOULD NOT interfere with Instruments under
   another Meter.
 
+<a name="instrument-naming-rule"></a>
```
bogdandrutu, line 168: Why is this random html code here? why not markdown headers?

specification/metrics/new_api.md
```diff
@@ -182,10 +184,115 @@ DIGIT = %x30-39 ; 0-9
   and '-'.
 * They can have a maximum length of 63 characters.
 
+<a name="instrument-unit"></a>
```
bogdandrutu, line 187: same

specification/metrics/new_api.md
```diff
@@ -182,10 +184,115 @@ DIGIT = %x30-39 ; 0-9
...
+SHOULD be treated as an oqaque string from the API and SDK (e.g. the SDK is not
+expected to validate the unit of measurement, or perform the unit conversion).
+
+* If the `unit` is not provided or the `unit` is null, the API and SDK MUST make
+  sure that the behavior is the same as an empty `unit` string.
+* It MUST be case-sensitive (e.g. `kb` and `kB` are different units), ASCII
+  string.
+* It can have a maximum length of 63 characters.
+
+<a name="instrument-description"></a>
```
bogdandrutu, line 199: Same

bogdandrutu (commented):
If I were you I would split this into 2 PRs:
* One that clarifies instrument properties
* One that adds the Counter details

specification/metrics/new_api.md
```diff
@@ -182,10 +184,115 @@ DIGIT = %x30-39 ; 0-9
...
+```
+
+```csharp
+// C#
+
+var counterItemsSold = meter.CreateCounter<UInt64>("item_sold", description="number of items sold");
+
+counterItemsSold.Add(2, ("item", "Tomato"), ("customer", "Jerry"));
+
+readonly struct CashReceived
```
bogdandrutu, line 285: In general "money" is an interesting topic, and the reason to use metrics for billing/money counting may be controversial. I would avoid examples that include billing/money

specification/metrics/new_api.md
reyang: There are ways to avoid the boxing and other perf hits in C#.

specification/metrics/new_api.md
reyang: It is following the tracing spec https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span-data-members.

I _guess_ it helps to avoid bloating the ToC (we only need this to reference a section of a topic, instead of having to introduce a header).

specification/metrics/new_api.md
reyang: Same as explained [here](https://github.com/open-telemetry/opentelemetry-specification/pull/1578#discussion_r602578753)

specification/metrics/new_api.md
reyang: Same as explained [here](https://github.com/open-telemetry/opentelemetry-specification/pull/1578#discussion_r602578753)

specification/metrics/new_api.md
```diff
@@ -182,10 +184,115 @@ DIGIT = %x30-39 ; 0-9
   and '-'.
 * They can have a maximum length of 63 characters.
 
+<a name="instrument-unit"></a>
+
+The `unit` is an optional string provided by the author of the instrument. It
+SHOULD be treated as an oqaque string from the API and SDK (e.g. the SDK is not
```
cijothomas, line 190:
```suggestion
SHOULD be treated as an opaque string from the API and SDK (e.g. the SDK is not
```

specification/metrics/new_api.md
```diff
@@ -182,10 +184,115 @@ DIGIT = %x30-39 ; 0-9
...
+* If the `unit` is not provided or the `unit` is null, the API and SDK MUST make
+  sure that the behavior is the same as an empty `unit` string.
+* It MUST be case-sensitive (e.g. `kb` and `kB` are different units), ASCII
+  string.
+* It can have a maximum length of 63 characters.
+
+<a name="instrument-description"></a>
+
+The `description` is an optional free-form text provided by the author of the
+instrument. It MUST be treated as an oqaque string from the API and SDK.
```
cijothomas, line 202:
```suggestion
instrument. It MUST be treated as an opaque string from the API and SDK.
```

specification/metrics/new_api.md
reyang: Removed "money"

specification/metrics/new_api.md
```diff
@@ -182,10 +184,114 @@ DIGIT = %x30-39 ; 0-9
...
+
+The API MUST accept the following parameters:
+
+* The `name` of the Instrument, following the [instrument naming
+  rule](#instrument-naming-rule).
+* An optional `unit of measure`, following the [instrument unit
+  rule](#instrument-unit).
+* An optional `description`, following the [instrument description
+  rule](#instrument-description).
+* An optional list of [`Attribute`](../common/common.md#attributes) names and
```
cijothomas, line 242: could we add an example which shows how one can use the attribute names given at Counter creation time?

specification/metrics/new_api.md
noahfalk: .NET preserves exact type information for generics at runtime and the JIT compiler can take them into account when generating code. This avoids the boxing problem that Java runs into with it's generic code.

specification/metrics/new_api.md
```diff
@@ -182,10 +184,114 @@ DIGIT = %x30-39 ; 0-9
...
+counterItemsSold.Add(2, ("item", "Tomato"), ("customer", "Jerry"));
+
+readonly struct PowerConsumption
+{
+    string customer;
+};
+
+var counterPowerUsed = meter.CreateCounter<double, PowerConsumption>("power_consumption", unit="kWh");
+counterPowerUsed.Add(13.5, new PowerConsumption { customer = "Tom" });
+counterPowerUsed.Add(200, new PowerConsumption { customer = "Jerry" }, ("is_green_energy", true));
```
noahfalk, line 292: While I don't anticipate our .NET 6 implementation to include this capability it does do a good job indicating the kind of flexibility we'd like language implementors to have.

specification/metrics/new_api.md
bogdandrutu: I did not realize it is there :)) probably because I couldn't get a link while reading. Anyway not a big deal

specification/metrics/new_api.md
bogdandrutu: Smart :)

specification/metrics/new_api.md
reyang: I've added one example showing how this could be achieved.

cijothomas (approved):

specification/metrics/new_api.md
```diff
@@ -182,10 +184,126 @@ DIGIT = %x30-39 ; 0-9
...
 
+<a name="instrument-unit"></a>
+
+The `unit` is an optional string provided by the author of the instrument. It
+SHOULD be treated as an oqaque string from the API and SDK (e.g. the SDK is not
+expected to validate the unit of measurement, or perform the unit conversion).
+
+* If the `unit` is not provided or the `unit` is null, the API and SDK MUST make
+  sure that the behavior is the same as an empty `unit` string.
+* It MUST be case-sensitive (e.g. `kb` and `kB` are different units), ASCII
```
jsuereth, line 195: Non-blocking comment: Is there a reason this is ASCII and later the description is BMP?

specification/metrics/new_api.md
```diff
@@ -182,10 +184,126 @@ DIGIT = %x30-39 ; 0-9
...
+<a name="instrument-description"></a>
+
+The `description` is an optional free-form text provided by the author of the
+instrument. It MUST be treated as an oqaque string from the API and SDK.
+
+* If the `description` is not provided or the `description` is null, the API and
+  SDK MUST make sure that the behavior is the same as an empty `description`
+  string.
+* It MUST support [BMP (Unicode Plane
+  0)](https://en.wikipedia.org/wiki/Plane_(Unicode)#Basic_Multilingual_Plane).
```
jsuereth, line 208: Nit: For folks like myself, can you mention that BMP is basically *only* the first three bytes of UTF-8 (or "utf8mb3")

jsuereth (approved):
Meta Comment:

Strings in OTel Specification SHOULD be specified somewhere globally and you should just denote that things are strings, referencing that specification.   We have a lot off "string" references in the Trace Spec where this spec is far more explicit.

I think it's worth opening a bug to address that and not blocking this PR.  That said, not 100% happy with ASCII vs BMP and wanted to understand why.

specification/metrics/new_api.md
reyang: The `unit` is ASCII, later we mentioned that `description` is UTF-8.
I guess there is no UTF-8 `unit`? If we want to make both UTF-8, I am happy to make the change.

specification/metrics/new_api.md
reyang: I've updated the description.

specification/metrics/new_api.md
reyang: I've created an issue #1587.

specification/metrics/new_api.md
```diff
@@ -182,10 +184,127 @@ DIGIT = %x30-39 ; 0-9
...
+
+The API MUST accept the following parameters:
+
+* The `name` of the Instrument, following the [instrument naming
+  rule](#instrument-naming-rule).
+* An optional `unit of measure`, following the [instrument unit
+  rule](#instrument-unit).
+* An optional `description`, following the [instrument description
+  rule](#instrument-description).
+* An optional list of [`Attribute`](../common/common.md#attributes) names and
```
victlu, line 243: We should take an optional list of Attributes (instead of just name and type)?  The value can be the default value if it is not overridden in subsequent add-measurement calls.

noahfalk: I think these are two different features being proposed? Neither one is a replacement or a super-set of the other. If we are adding any attributes we will need to specify what their intended interpretation is. A few possibilities illustrating reasonable but widely different ways of interpreting a per-instrument attribute list:

1. Attribute name/types represent a pre-declaration of dimensions and the values for each dimension will provided when the measurement is captured (Prometheus model).
2. The attribute names represent optional hints about an default set of attribute values to aggregate over. If additional  dynamic attributes are specified with a measurement they are dropped unless configuration elsewhere overrides the hint.
3. The attributes represent constant attributes that will be appended to any attributes applied via other configuration or at the callsite. It is not intended for the user to provide a dynamic attribute at the measurement recording callsite with the same key. If there is a collision it could either be specified as undefined behavior or a well-defined error. This is similar to Prometheus [Static Labels](https://github.com/prometheus-net/prometheus-net#static-labels).
4. They represent default attributes to include with all measurements and there is some order of precedence that determines how values merge when an attribute key name specified elsewhere collides with a name provided in the defaults.
 
All of these can be added into the design independently of the others, ie there are 2^4 different subsets of these features and each of them is a valid design option. My current vote would be (1) no, (2) yes eventually, (3) maybe? (4) no.

specification/metrics/new_api.md
```diff
@@ -182,10 +184,127 @@ DIGIT = %x30-39 ; 0-9
...
+
+The `unit` is an optional string provided by the author of the instrument. It
+SHOULD be treated as an oqaque string from the API and SDK (e.g. the SDK is not
+expected to validate the unit of measurement, or perform the unit conversion).
+
+* If the `unit` is not provided or the `unit` is null, the API and SDK MUST make
+  sure that the behavior is the same as an empty `unit` string.
+* It MUST be case-sensitive (e.g. `kb` and `kB` are different units), ASCII
+  string.
+* It can have a maximum length of 63 characters.
```
lzchen, line 197: Maybe add a reasoning to why `63` was chosen?

specification/metrics/new_api.md
```diff
@@ -182,10 +184,127 @@ DIGIT = %x30-39 ; 0-9
...
+
+* If the `description` is not provided or the `description` is null, the API and
+  SDK MUST make sure that the behavior is the same as an empty `description`
+  string.
+* It MUST support [BMP (Unicode Plane
+  0)](https://en.wikipedia.org/wiki/Plane_(Unicode)#Basic_Multilingual_Plane),
+  which is basically only the first three bytes of UTF-8 (or `utf8mb3`).
+  Individual language client can decide if they want to support more Unicode
+  [Planes](https://en.wikipedia.org/wiki/Plane_(Unicode)).
+* It MUST support at least 1023 characters. Individual language client can
```
lzchen, line 212:
```suggestion
* It MUST support at least 1023 characters. Individual language clients can
```

specification/metrics/new_api.md
```diff
@@ -182,10 +184,127 @@ DIGIT = %x30-39 ; 0-9
...
+  decide if they want to support more.
+
 ### Counter
 
+`Counter` is a synchronous Instrument which supports non-negative increment.
```
lzchen, line 217:
```suggestion
`Counter` is a synchronous Instrument which supports non-negative increments.
```

specification/metrics/new_api.md
```diff
@@ -182,10 +184,127 @@ DIGIT = %x30-39 ; 0-9
...
+  rule](#instrument-description).
+* An optional list of [`Attribute`](../common/common.md#attributes) names and
+  types.
+
+Here are some examples that individual language client might consider:
+
+```python
+# Python
+
+exception_counter = meter.create_counter(name="exceptions", description="number of exceptions caught", value_type=int)
```
lzchen, line 251: Referring to what is mentioned above for desired strong types, should we define the types that are acceptable (`double` and int`)?

specification/metrics/new_api.md
```diff
@@ -182,10 +184,127 @@ DIGIT = %x30-39 ; 0-9
...
+
+This API SHOULD NOT return a value (it MAY return a dummy value if required by
+certain programming languages or systems, for example `null`, `undefined`).
+
+Required parameters:
+
+* Optional [attributes](../common/common.md#attributes).
+* The increment amount, which MUST be a non-negative numeric value.
+
+The client MAY decide to allow flexible
```
lzchen, line 282: What does `flexible` mean in this context?

lzchen (approved):

specification/metrics/new_api.md
reyang: The spec intentionally leaves such flexibility for the language clients. For example, it is totally fine if a certain language (e.g. JavaScript) decides to support float only, and use that for all the values (whether semantically it is an integer or not).

specification/metrics/new_api.md
reyang: This is out of scope for this PR, we can discuss this topic on the upcoming SIG meeting.

bogdandrutu (approved):

SergeyKanzhelev (approved):

specification/metrics/new_api.md
noahfalk: Ideally I'd say we should not include the "optional list of attribute names/types" in the spec at all until it is in scope to discuss their intended usage : ) However as long as it gets addressed in some follow up PR all is well.

PR #1581, opened 2021-03-26
Add aws.ecs.task.revision semantic convention

mx-psi:
## Changes

Add semantic convention for an ECS task definition revision, described in the [AWS ECS Task metadata endpoint documentation](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint-v4.html) as

> The revision of the Amazon ECS task definition for the task.

Follow up to open-telemetry/opentelemetry-collector-contrib#2814.


carlosalberto (approved):

carlosalberto:
@rakyll @anuraaga please take a look.

anuraaga (approved):

semantic_conventions/resource/cloud_provider/aws/ecs.yaml
```diff
@@ -35,4 +35,9 @@ groups:
         brief: >
           The task definition family this task definition is a member of.
         examples: ['opentelemetry-family']
+      - id: task.revision
+        type: string
+        brief: >
+          The revision for this task definition.
+        examples: ["8", "26"]
```
arminru, line 42: If these are integer counters that get incremented for each revision, please use the `integer` type.
If the revision could be non-integer as well (e.g. `"8b"`), please add an example to make this clear.
Thanks!

mx-psi: The endpoint returns a string but that string can be usually (maybe always?) parsed as an integer. [AWS' docs](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint-v4.html) only exemplify that the type returned is a string but they don't clarify if it's always an integer.  

I think it's better to use a string in case something like `8b` can be returned by the endpoint (either now or in the future), but I don't feel comfortable adding it as an example since I have not seen that happen in practice and no similar example exists in the docs.

Maybe @anuraaga or @rakyll can shed some light into this?

anuraaga: We should stick with string if that's the type in the API docs I think

arminru: Alright! If we can't tell if it could be any string other than one resembling an integer I'm fine with also keeping the examples as is.

arminru (approved):

bogdandrutu (approved):

Issue #1347, opened 2021-01-15
Semantic conventions for batch jobs


mateuszrzeszutek:
**What are you trying to achieve?**

I want to introduce some semantic conventions for batch jobs, since there are currently no conventions around that. I'm mostly interested in instrumenting Spring Batch applications.
There already is a Java JSR-352 spec that describes a batch job API, I was thinking of basing the trace spec on that - the concepts of job/step/chunk seem generic and language-agnostic enough (and there doesn't seem to be any other batch job specification).
Before diving into details, is there a place for this in the trace semantic conventions?

**Additional context.**

* [Spring Batch docs](https://docs.spring.io/spring-batch/docs/4.3.x/reference/html/domain.html#domainLanguageOfBatch)
* [JSR-352 spec](https://www.jcp.org/aboutJava/communityprocess/final/jsr352/index.html)

fbogsany:
Good timing: we discussed this in the Ruby SIG meeting earlier this week. Background job queues are common in Ruby web applications, with the dominant implementations being Resque and Sidekiq. So far, we've instrumented those systems using the messaging semantic conventions, but they're really not a great fit for background/batch jobs. See https://github.com/open-telemetry/opentelemetry-ruby/pull/547#issuecomment-758889459

Oberon00:
Wouldnt this be a case where you just use an INTERNAL span without any attributes? Not every span needs to conform to a semantic convention. What information do you want to have on batch job spans?

fbogsany:
> the concepts of job/step/chunk

^ this bit seems useful.

For Ruby batch job systems, relative to message systems:
1. "receiver" spans are not particularly useful
2. The job class name is usually much more interesting as part of the span name than the queue name (e.g. `MyJob enqueue` vs `default send`)
3. The `... enqueue` suffix is more in keeping with the domain language than `... send`
4. `destination_kind` is likely always `queue`, so probably isn't an interesting thing to specify (and certainly shouldn't be required).

Oberon00:
For the "job class name": There are semantic conventions for code locations, see https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/span-general.md#source-code-attributes.

fbogsany:
> For the "job class name": There are semantic conventions for code locations, see https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/span-general.md#source-code-attributes.

[...]

PR #1582, opened 2021-03-27
[RFC] [Semantic Convention] "Job" traces

johnnyshields:
This PR defines a semantic convention for "Job" traces.

A "Job" models a batch job or task, which can be either enqueued, scheduled, or run on-demand.

Examples of job-related systems are:
- [Sidekiq](https://github.com/mperham/sidekiq), instrumented [here](https://github.com/open-telemetry/opentelemetry-ruby/tree/main/instrumentation/sidekiq) using "Messaging" convention
- [Delayed Job](https://github.com/collectiveidea/delayed_job), instrumented [here](https://github.com/open-telemetry/opentelemetry-ruby/tree/main/instrumentation/delayed_job) using "Messaging" convention
- [Celery](https://github.com/celery/celery), instrumented [here](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation/opentelemetry-instrumentation-celery) using "Messaging" convention
- [Resque](https://github.com/resque/resque), issue [here](https://github.com/open-telemetry/opentelemetry-ruby/issues/250)
- [Rake](https://github.com/ruby/rake), issue [here](https://github.com/open-telemetry/opentelemetry-ruby/issues/259)
- [Apache Airflow](https://airflow.apache.org/), issue [here](https://github.com/apache/airflow/issues/12771)

The spec for "Job" would be somewhere between "Messaging" and "FAAS". It would cover both the producer and consumer aspects of running jobs.

As noted above, OTEL instrumentation libraries are using the "Messaging" convention to represent Jobs today. However "Messaging", which is intended for Kafka, RabbitMQ, etc. is not a great fit:
- Messaging systems focus on the delivery of messages irrespective of the message contents. Jobs performing work/processing based on the job instruction.
- In a Messaging system the categorization is done by the "queue" or "topic" on which messages are transmitted; in a trace viewing system (e.g. Datadog, Lightstep, etc.) I would want to see messages labelled according to their queue. In Job systems, the categorization is based on job name/type, and the queue is just for worker (consumer) resource allocation. Hence I would want traces labelled as "ChargePaymentJob", "RefundPaymentJob", etc. even though all such jobs are in the same "payments" queue.

FAAS is also not ideal, as it is typically for serverless providers such as AWS Lambda which the infrastructure running the job has been abstracted away, and the primary focus is on the "function".


johnnyshields:
@dazuma @robertlaurin @adamantike

iNikem:
I think we also need to do some modeling work. E.g. to answer the following questions:
- What `SpanKind` has the root span?
- If job executor reads a batch of tasks and splits it into chunks, should we have nested spans as well? Should they have the same semantic conventions? Should we have chunkId/itemId?
- How to decide when use `messaging` convention and when this new `job`? E.g. Kafka consumer listening a topic and then doing some job, is it messaging or job? Why?

johnnyshields:
> What SpanKind has the root span?

For DelayedJob, Rake, Airflow, etc. it would be `SpanKind.CONSUMER` for the process that performs jobs. Rake/Airflow run on-demand or by a timer aren't exactly "consumers" but it's the best fit out of all the existing span-kind values we have.

> If job executor reads a batch of tasks and splits it into chunks, should we have nested spans as well? Should they have the same semantic conventions? Should we have chunkId/itemId?

In practice, child jobs are typically enqueued as new jobs with new `job_ids`. It would be make more sense to do this as a `parent_job_id` attribute which would be common across all child jobs spawned from the original master job. Since each child is executed asynchronously in it's own process and thread they would not appear as "nested" with the parent. AFAIK DelayedJob/Resque/etc don't have such an attribute today but Airflow likely does as it is [DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph)-based.

> How to decide when use messaging convention and when this new job? E.g. Kafka consumer listening a topic and then doing some job, is it messaging or job? Why?

A process which listens/consumes from Kafka and then performs a job synchronously is probably traced as a `Messaging consumer` because it's Kafka listening library will be instrumented. Theoretically you could also add in a `Job` span but it would be redundant with the message consumer.

iNikem:
> In practice, child jobs are typically enqueued as new jobs with new `job_ids`. It would be make more sense to do this as a `parent_job_id` attribute which would be common across all child jobs spawned from the original master job. Since each child is executed asynchronously in it's own process and thread they would not appear as "nested" with the parent. AFAIK DelayedJob/Resque/etc don't have such an attribute today but Airflow likely does as it is [DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph)-based.

I am more familiar with https://spring.io/projects/spring-batch which, it seems to me, follows different approach, hence my questions.

iNikem:
Btw, @johnnyshields have you seen https://github.com/open-telemetry/opentelemetry-specification/issues/1347 ?

johnnyshields:
@iNikem no i had not seen that, funny because I searched a bit. I think it is roughly compatible with this one.

johnnyshields:
@mateuszrzeszutek

semantic_conventions/trace/job.yaml
```diff
@@ -0,0 +1,110 @@
...
+          type: string
+          required: always
+          brief: >
+            The type or class name of the job.
+          examples: ['SignupMailJob', 'ChargePaymentJob']
+        - id: job_id
+          type: string
+          required: always
+          brief: >
+            A value as an identifier for the job, represented as a string.
```
jsuereth, line 24: Nit:  Does this need to be globally unique to the process?

johnnyshields: This is typically the database ID of the persisted job object. In the case of SQL databases it may be an incrementing integer so it is not necessarily "globally unique".

semantic_conventions/trace/job.yaml
```diff
@@ -0,0 +1,110 @@
...
+      brief: >
+          This document defines the attributes used in
+          batch job queuing and processing systems.
+      attributes:
+        - id: system
+          type: string
+          required: always
+          brief: >
+            A string identifying the job processing system.
+          examples: ['sidekiq', 'resque', 'rake', 'celery', 'airflow']
```
mateuszrzeszutek, line 13: Isn't this roughly equivalent to the [instrumentation library](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/overview.md#instrumentation-libraries)?

semantic_conventions/trace/job.yaml
```diff
@@ -0,0 +1,110 @@
...
+        - id: job_id
+          type: string
+          required: always
+          brief: >
+            A value as an identifier for the job, represented as a string.
+          examples: 'b18fc2c2f887f61452a7c7c7c7048572'
+        - id: job_label
+          type: string
+          brief: >
+            A value assigned as a human-readable label for the job.
```
mateuszrzeszutek, line 29: Do you have an example of this in any existing lib? I just wonder what's the real use case for this here

semantic_conventions/trace/job.yaml
```diff
@@ -0,0 +1,110 @@
...
+          This document defines the attributes used in
+          batch job queuing and processing systems.
+      attributes:
+        - id: system
+          type: string
+          required: always
+          brief: >
+            A string identifying the job processing system.
+          examples: ['sidekiq', 'resque', 'rake', 'celery', 'airflow']
+        - id: job_name
```
mateuszrzeszutek, line 14: With the `job` prefix it'd look like `job.job_name`. How about just `name`? Same applies to other attributes.

semantic_conventions/trace/job.yaml
```diff
@@ -0,0 +1,110 @@
...
+        - id: format
+          type: string
+          brief: >
+            The name of the job format when enqueued for processing.
+          examples: ['resque']
+        - id: queue_name
+          type: string
+          brief: >
+            The processing queue to which the job belongs.
+          examples: ['mailers', 'payments']
```
mateuszrzeszutek, line 55: I'm not sure if I understand that correctly, but you want to use messaging conventions when a queue is used, right? Isn't this attribute a duplicate of `messaging.destination_kind=queue` and `messaging.destination`?

semantic_conventions/trace/job.yaml
```diff
@@ -0,0 +1,110 @@
...
+            [ISO 8601](https://www.iso.org/iso-8601-date-and-time-format.html)
+            format expressed in [UTC](https://www.w3.org/TR/NOTE-datetime).
+            For job attempts which are being retried, this should reflect when the current
+            attempt was scheduled (e.g. using a back-off timer.)
+          examples: "2021-03-23T13:47:06Z"
+        - ref: net.peer.name
+          note: >
+            This should be the IP/hostname of the process (or other network-level peer) used to enqueue or perform job.
+          required:
+            conditional: If available.
```
mateuszrzeszutek, line 84: Some batch/job libraries (spring-batch 😄 ) do not send the tasks over to other processes, maybe the condition should explicitly state that this is required only for those job systems that are somewhat distributed?

mateuszrzeszutek:
This roughly corresponds to the first two spans for spring-batch that I listed [here](https://github.com/open-telemetry/opentelemetry-specification/issues/1347#issuecomment-762878173).

* Is the CONSUMER/PRODUCER kind mandatory? This makes little sense for JSR-352 implementations (spring batch), as they do not concern themselves with queueing or scheduling, they just execute jobs in the current process (either synchronously or asynchronously). You can run trigger spring batch jobs with a HTTP call (button click on your website), and AFAIR this is what one of our clients does. Anyway, for job libraries that do not deal with queueing or messaging INTERNAL would be a better choice for the span kind.
* Is there any kind of result/status returned by at least one of those libs? Spring batch jobs return a user-defined `exit_status` string, which can be pretty much anything - still worth capturing.

> > If job executor reads a batch of tasks and splits it into chunks, should we have nested spans as well? Should they have the same semantic conventions? Should we have chunkId/itemId?
> 
> In practice, child jobs are typically enqueued as new jobs with new job_ids. It would be make more sense to do this as a parent_job_id attribute which would be common across all child jobs spawned from the original master job. Since each child is executed asynchronously in it's own process and thread they would not appear as "nested" with the parent. AFAIK DelayedJob/Resque/etc don't have such an attribute today but Airflow likely does as it is DAG-based.

This is rather different from what spring batch jobs look like. Spring batch jobs are hierarchical: every job is composed of steps, each step may be split into chunks (roughly equivalent to DB transactions), and each chunk may read, process and write several items. Everything in the same process.
Spring batch seems to be the only framework that structures its jobs this way though: I don't think that the general job spec should concern itself with steps and items, we should only model the job level here. Steps, chunks and items may be covered in a specific instrumentation spec (like the AWS one).

semantic_conventions/trace/job.yaml
```diff
@@ -0,0 +1,110 @@
...
+          examples: 3
+        - id: scheduled_run_time
+          type: string
+          brief: >
+            A string containing the time when the job was scheduled to run, specified in
+            [ISO 8601](https://www.iso.org/iso-8601-date-and-time-format.html)
+            format expressed in [UTC](https://www.w3.org/TR/NOTE-datetime).
+            For job attempts which are being retried, this should reflect when the current
+            attempt was scheduled (e.g. using a back-off timer.)
+          examples: "2021-03-23T13:47:06Z"
```
mateuszrzeszutek, line 79: I wonder if we need a small, separate semantic convention for scheduled operations: you can run a cron job (or use a `ScheduledExecutorService` in Java, or Quartz, or...) that runs any code, not necessarily a batch job or a FAAS function.

semantic_conventions/trace/job.yaml
mateuszrzeszutek: Hmm, do those `net.*` attributes apply only for queue/messaging job systems? In that case the messaging convention already defines them, there's no need to define them here again.

johnnyshields:
> Spring batch seems to be the only framework that structures its jobs this way though

I believe Apache Airflow is similar to Spring Batch in this regard. I guess there is a distinction between "Batch/Scheduled Jobs" and "Queued Jobs".

semantic_conventions/trace/job.yaml
fbogsany: Job systems often define multiple job queues for prioritization. E.g. we might have `default`, `low` and `payment` queues.

semantic_conventions/trace/job.yaml
johnnyshields: Sure, it would the job class name in DelayedJob / ActiveJob or the batch job name in Apache Airflow.

semantic_conventions/trace/job.yaml
```diff
@@ -0,0 +1,110 @@
...
+              - id: other
+                value: 'other'
+                brief: 'If none of the others apply'
+          brief: >
+            Type of the trigger on which the job is performed.
+        - id: format
+          type: string
+          brief: >
+            The name of the job format when enqueued for processing.
+          examples: ['resque']
```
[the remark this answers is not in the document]
johnnyshields: Yeah the serialization format. This node is probably not necessary.

semantic_conventions/trace/job.yaml
johnnyshields: @fbogsany depending on the job system, queues and priority can be two separate things; it is possible that jobs within the same queue could have different priority values. Note I have also added a `priority` field.

semantic_conventions/trace/job.yaml
weyert: Yes, in Hangfire for .NET you can use the `DisplayName`-attribute to define aa human-readable label, see this example: https://github.com/HangfireIO/Hangfire/blob/3c6d893023d5d7558d6098e80034865004b73786/samples/ConsoleSample/Services.cs#L88

semantic_conventions/trace/job.yaml
Oberon00, line 13: I think we can use an "open" enum here, like for DB: https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/trace/database.yaml#L12-L14

semantic_conventions/trace/job.yaml
Oberon00: Good explanation! I think this comment could and should be used as a "note".

semantic_conventions/trace/job.yaml
```diff
@@ -0,0 +1,110 @@
...
+          brief: >
+            The type or class name of the job.
+          examples: ['SignupMailJob', 'ChargePaymentJob']
+        - id: id
+          type: string
+          required: always
+          brief: >
+            A value as an identifier for the job, represented as a string.
+          examples: 'b18fc2c2f887f61452a7c7c7c7048572'
+        - id: label
```
Oberon00, line 26: I think `display_name` may be a better name, as "label" does not make it as clear that this is human readable. As mentioned elsewhere, that would also match the name that some libs use for this concept: https://github.com/open-telemetry/opentelemetry-specification/pull/1582#discussion_r607955035

semantic_conventions/trace/job.yaml
Oberon00: I thought Hangfire was a messaging system? Wouldn't you use the messaging semantic conventions rather than the job ones for it?

Oberon00 (commented):
Re: https://github.com/open-telemetry/opentelemetry-specification/pull/1582#issuecomment-808867406

>> How to decide when use messaging convention and when this new job? E.g. Kafka consumer listening a topic and then doing some job, is it messaging or job? Why?
>
> A process which listens/consumes from Kafka and then performs a job synchronously is probably traced as a `Messaging consumer` because it's Kafka listening library will be instrumented. Theoretically you could also add in a `Job` span but it would be redundant with the message consumer.

I also think this is an important question. We probably want to avoid cases where the sender uses messaging and the receiver job spans or vice versa. After looking at attributes I think that we should remove these:

* job.queue_name: messaging.destination
* net.peer.*: Also referenced by messaging.

If a job system uses queuing with a separate sender/producer of the jobs, messaging semantic conventions should be used additionally if further description of the queuing is desired.

Important editorial note: Please generate a markdown table in a new file under `specification/trace/semantic_conventions/` and add a link in the README in that directory. Alternatively, you can add it to the same markdown file as messaging and change the title.
Adding a markdown file also gives you room for additional general explanation text.

semantic_conventions/trace/job.yaml
johnnyshields: I agree

johnnyshields:
@Oberon00 I disagree that Job systems that have "queues" should use Messaging semantics. "Queue" in the context of a Jobs is usually an optional feature. For example, Ruby's Delayed Job will work all jobs in FIFO order, but it optionally allows you to set named "queues" and then assign worker resources to specific queues. Unlike a Messaging system where the queue/topic is the primary feature, in a Job system the queue is an optional/secondary attribute.

In addition, `net.peer.*` should still be used here because ultimately traces are sent from physical hosts, and we want to see which host sent the trace.

semantic_conventions/trace/job.yaml
ahayworth: It's worth noting that [ActiveJob in Rails](https://api.rubyonrails.org/classes/ActiveJob/Core.html) provides a `job_id` attribute as well, which could fit nicely into this attribute.

semantic_conventions/trace/job.yaml
johnnyshields: `net.peer.*` is the host which generated the trace, so it is needed here.

semantic_conventions/trace/job.yaml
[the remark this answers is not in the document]
johnnyshields: I have removed this in my latest version

semantic_conventions/trace/job.yaml
johnnyshields: Changed in my latest version

semantic_conventions/trace/job.yaml
johnnyshields: Added to description

semantic_conventions/trace/job.yaml
johnnyshields: I think this is the key question remaining for this RFC.

semantic_conventions/trace/job.yaml
johnnyshields: Great suggestion, have applied it.

johnnyshields:
Do not mark as stale please

jmacd:
This should not be considered stale! Sorry for the delays.

jmacd:
> A better litmus test is, "does work happen based on the contents/payload of the message".

:100:

jmacd:
Here are some questions, since this PR is labeled an RFC.

OTel tracing has a Link concept, but I have rarely seen it used. The big question in this discussion seems to be whether we should use Queuing conventions, or Job conventions.

- Why not use both conventions? The queue system is independent of the job system, it looks like.
- Can we Link these traces?

Suppose a system uses a request that contains a batch of work items. Now one queued message spawns several jobs. How are they linked?

Oberon00:
> A better litmus test is, "does work happen based on the contents/payload of the message".

Being a litmus test, it may work to exclude some messages, but I don't think this enough as a distinction. If no work happens based on the message, why send it anyway? So I suppose most messaging-system messages also pass the test.

ahayworth:

Is this PR actually stale? I've been watching it to see what happens. Do we feel that we have reached any kind of consensus on this proposal?

johnnyshields:
@ahayworth as a next step I need to write the documentation for this. I'd like to keep it open. Any feedback would be appreciated!

ahayworth:
@johnnyshields My overall feedback is "I like it" - is there documentation that you need help with?

johnnyshields:
@ahayworth yes, I need help with the markdown documentation for this. I have some work in progress which I'll try to get committed this weekend.

johnnyshields:
Please re-open. I should be able to get to the remaining steps this weekend.

miketzian:
Ping so that this doesn't become stale, this looks like a useful addition

jmacd:
We can't merge this without more reviews, and I believe there is more work to do.

ahayworth:
> @ahayworth yes, I need help with the markdown documentation for this. I have some work in progress which I'll try to get committed this weekend.

@johnnyshields I can help with the documentation if needed - is there a preferred way you'd like to collaborate?

johnnyshields:
@ahayworth I've naively committed all my work in progress on the docs. If you can just fork my PR and clean everything up that will be fine.

simaoribeiro:
Hi @johnnyshields , are you playing to continue this work in the future?
As I mentioned in this discussion https://github.com/open-telemetry/opentelemetry-specification/discussions/2170, my team is currently migrating from OpenTracing to OpenTelemetry and a job scheduler specification is something that is missing.
My only experience is using Hangfire, but I would gladly help push this

PR #1590, opened 2021-04-01
Add observable (async) counter to the metrics API

reyang:
Related to #1578.

## Changes

During the [03/25/2021 Metrics API/SDK SIG Meeting](https://docs.google.com/document/d/1-bCYkN-DWJq4jw1ybaDZYYmx-WAe6HnwfWbkm8d57v8/edit#heading=h.raqds2xgauu0), we've agreed to tackle an async instrument after `Counter` is done.

After this PR, the only outstanding issues for the metrics API spec would be:
* Figure out the remaining instruments
  * How many do we need
  * What should be the names
* Hint API (we've discussed during the SIG meeting that we will move forward to the SDK spec and come back to this topic after SDK spec is in shape)

This will be discussed during the upcoming (4/1) Metrics API/SDK SIG meeting.


specification/metrics/new_api.md
```diff
@@ -308,6 +312,87 @@ counterPowerUsed.Add(13.5, new PowerConsumption { customer = "Tom" });
...
+
+* The `name` of the Instrument, following the [instrument naming
+  rule](#instrument-naming-rule).
+* An optional `unit of measure`, following the [instrument unit
+  rule](#instrument-unit).
+* An optional `description`, following the [instrument description
+  rule](#instrument-description).
+* An optional list of [`Attribute`](../common/common.md#attributes) names and
+  types.
+* A `callback` function.
```
reyang, line 347: This is a bit different from the previous approach, for context please refer to [OTEP72](https://github.com/open-telemetry/oteps/blob/main/text/metrics/0072-metric-observer.md).

specification/metrics/new_api.md
```diff
@@ -308,6 +312,87 @@ counterPowerUsed.Add(13.5, new PowerConsumption { customer = "Tom" });
...
+  * This callback function will only be called when the Meter is being observed.
+  * This callback function will return the value(s) with optional
+    [`Attribute`](../common/common.md#attributes). Individual language client
+    can decide what is the idomatic approach (e.g. it could be a list, tuple,
+    generator, enumerator, etc.).
+  * Individual language client SHOULD define whether this callback function
+    needs to be reentrant safe / thread safe or not.
+* An optional `state`.
```
noahfalk, line 355: I'd suggest the spec require that there is some way to pass state to the callback, and then use idiomatic language conventions to determine whether that is an explicit parameter or something else. In .NET it is very easy to capture state with a delegate so there is little need to have another state parameter. Ie:
```C#
IAtomicClock clock = AtomicClock.Connect();
var obCaesiumOscillates = meter.CreateCounterObserver<UInt64>("caesium_oscillates", () => clock.GetCaesiumOscillates());
```

reyang: I've updated the description by saying "The API SHOULD provide some way to pass `state` to the callback".
I've kept the example as-is just to show that it is _one possible way_ of doing this (but not necessarily _the only way_ or _the best way_).

specification/metrics/new_api.md
victlu: What is the concern regarding OTEP72 pattern that was deficient? I assume getting the ObserverResult object with a set of well defined API would allow better futureproofing.

specification/metrics/new_api.md
```diff
@@ -308,6 +312,87 @@ counterPowerUsed.Add(13.5, new PowerConsumption { customer = "Tom" });
...
+// A simple scenario where only one value is reported
+
+interface IAtomicClock
+{
+    UInt64 GetCaesiumOscillates();
+}
+
+IAtomicClock clock = AtomicClock.Connect();
+
+var obCaesiumOscillates = meter.CreateCounterObserver<UInt64>("caesium_oscillates", clk => clk.GetCaesiumOscillates(), clock);
```
victlu, line 388: How do you deal with different return types? (i.e. returning one value vs your python example of returning 3 values) I'm not sure in C# if you can overload via callback function with only different return types.

specification/metrics/new_api.md
```diff
@@ -308,6 +312,87 @@ counterPowerUsed.Add(13.5, new PowerConsumption { customer = "Tom" });
...
+[`Meter`](#meter). This MAY be called `CreateObservableCounter`. If strong type
+is desired, the client can decide the language idomatic name(s), for example
+`CreateUInt64ObservableCounter`, `CreateDoubleObservableCounter`,
+`CreateObservableCounter<UInt64>`, `CreateObservableCounter<double>`.
+
+The API MUST accept the following parameters:
+
+* The `name` of the Instrument, following the [instrument naming
+  rule](#instrument-naming-rule).
+* An optional `unit of measure`, following the [instrument unit
```
victlu, line 341: Would be interested in seeing an example of what you think this looks like in the API call.

specification/metrics/new_api.md
```diff
@@ -308,6 +312,87 @@ counterPowerUsed.Add(13.5, new PowerConsumption { customer = "Tom" });
...
+  rule](#instrument-unit).
+* An optional `description`, following the [instrument description
+  rule](#instrument-description).
+* An optional list of [`Attribute`](../common/common.md#attributes) names and
```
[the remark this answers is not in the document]
reyang: Discussed during the [04/01/2021 Metrics API/SDK SIG Mtg](https://docs.google.com/document/d/1-bCYkN-DWJq4jw1ybaDZYYmx-WAe6HnwfWbkm8d57v8/edit#heading=h.bv4b9skwq3an) and the concluded that we'll remove it.

specification/metrics/new_api.md
reyang: The concern is that the ObserverResult might escape from the callback.

For example:

```
class YourClass {
  private static final Meter meter = ...;
  private static final ObserverDouble cpuLoad = ...;

  void init() {
    LabelSet labelSet = meter.createLabelSet("low_power", isLowPowerMode());
    cpuLoad.setCallback(
        new ObserverDouble.Callback<ObserverDouble.Result>() {
          @Override
          public void update(Result result) {
              ThreadPool.QueueTask(() => { result.Observe(getCPULoad(), labelSet); });
        });
  }
}
```

reyang: I _guess_ it might be a problem for clients that are async by default (e.g. Node.js).

reyang: Discussed during the [04/01/2021 Metrics API/SDK SIG Mtg](https://docs.google.com/document/d/1-bCYkN-DWJq4jw1ybaDZYYmx-WAe6HnwfWbkm8d57v8/edit#heading=h.bv4b9skwq3an) and concluded that we'll give language clients flexibility.

specification/metrics/new_api.md
reyang: I guess this is not a spec issue.

specification/metrics/new_api.md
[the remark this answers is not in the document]
reyang: Removed.

specification/metrics/new_api.md
reyang: I've updated the PR to reflect what we've concluded.

specification/metrics/new_api.md
```diff
@@ -29,7 +29,11 @@ Table of Contents
   * [Meter operations](#meter-operations)
 * [Instrument](#instrument)
   * [Counter](#counter)
-    * [Counter Creation](#counter-creation)
+    * [Counter creation](#counter-creation)
+    * [Counter operations](#counter-operations)
+  * [ObservableCounter](#observablecounter)
```
reyang, line 34: For folks who haven't got a chance to join the [04/01/2021 Metrics API/SDK SIG Mtg](https://docs.google.com/document/d/1-bCYkN-DWJq4jw1ybaDZYYmx-WAe6HnwfWbkm8d57v8/edit#heading=h.bv4b9skwq3an):

* Naming is a hard topic because it is very very subjective
* We want to avoid the previous name `SumObserver` since most of the users find it hard to mentally connect the concept with `Counter`, although English wise it might be more accurate.
* We've explored the word `async` and decided to avoid it, because async has many different meanings (e.g. Async I/O, async function in Python and C#, etc.)
* The general thinking is that if the sync instrument is called `Foo`, then the corresponding async instrument should be called `ObservableFoo`.

So far this name seems to be well accepted by the SIG members and the metrics experts. If you have strong feeling that we should pick a different name, please reply here and explain the pros/cons. Thanks!

specification/metrics/new_api.md
```diff
@@ -153,8 +157,6 @@ will have the following information:
   instruments, whether it is synchronous or asynchronous
 * An optional `unit of measure`
 * An optional `description`
-* An optional list of [`Attribute`](../common/common.md#attributes) names and
```
reyang, line 156: This is removed as a result of discussion during the [04/01/2021 Metrics API/SDK SIG Mtg](https://docs.google.com/document/d/1-bCYkN-DWJq4jw1ybaDZYYmx-WAe6HnwfWbkm8d57v8/edit#heading=h.bv4b9skwq3an). Please check the meeting notes.

specification/metrics/new_api.md
```diff
@@ -319,6 +409,14 @@ for the interaction between the API and SDK.
 * A value
 * [`Attributes`](../common/common.md#attributes)
 
+## Compatibility
```
reyang, line 412: @bogdandrutu I've added a small section here based on the discussion during the SIG meeting.
I think the ideal approach would be push this to the entire spec, but it might take some time. I can file an issue to track this and push for a resolution, please let me know your suggestion.

specification/metrics/new_api.md
```diff
@@ -319,6 +421,14 @@ for the interaction between the API and SDK.
...
+
+All the metrics components SHOULD allow new APIs to be added to existing
```
jsuereth, line 426: +1.  We may need to provide guidance here during implementations, but glad to see this called out.

specification/metrics/new_api.md
```diff
@@ -308,6 +308,108 @@ counterPowerUsed.Add(13.5, new PowerConsumption { customer = "Tom" });
...
 
+### ObservableCounter
+
+`ObservableCounter` is an asynchronous Instrument which reports
+[monotonically](https://wikipedia.org/wiki/Monotonic_function) increasing
+value(s) when the Meter is observed.
+
+Example uses for `ObservableCounter`:
+
+* [CPU time](https://wikipedia.org/wiki/CPU_time), which could be reported for
```
jsuereth, line 319: I think It''d be good to call out use cases for where "ObservableCounter" is *not* good for tracking CPU time.


E.g. in a  multi-tenant scenario where you want to track CPU usage / user, you'd actually want some kind of sycnhronous instrument that can report on request-level usage with an identified user.

Observable / Async metrics are really only useful for context-less metrics (those that cannot interact with other telemetry).   I'd like to call that out here for folks deciding between Async/Sync as I think Otel really shines in the Sync scenario w/ Context.

jsuereth (approved):
A few non-blocking comments.

specification/metrics/new_api.md
reyang: I think it is already covered. Please search for `unit="kWh"`.

specification/metrics/new_api.md
bogdandrutu: Please split this in a separate PR.

specification/metrics/new_api.md
bogdandrutu: Indeed, but this is tricky because you don't want users that report cpu_usage (not per request) to use the sync version and believe that is better :)

specification/metrics/new_api.md
```diff
@@ -308,6 +308,108 @@ counterPowerUsed.Add(13.5, new PowerConsumption { customer = "Tom" });
...
+* The `name` of the Instrument, following the [instrument naming
+  rule](#instrument-naming-rule).
+* An optional `unit of measure`, following the [instrument unit
+  rule](#instrument-unit).
+* An optional `description`, following the [instrument description
+  rule](#instrument-description).
+* A `callback` function.
+
+The `callback` function is responsible for reporting the
+[Measurement](#measurement)s. It will only be called when the Meter is being
```
bogdandrutu, line 344: "when the instrument is being observed" not the meter. or the metric associated with this instrument.

specification/metrics/new_api.md
```diff
@@ -308,6 +308,108 @@ counterPowerUsed.Add(13.5, new PowerConsumption { customer = "Tom" });
...
+some examples:
+
+* Return a list (or tuple, generator, enumerator, etc.) of `Measurement`s.
+* Use an observer argument to allow individual `Measurement`s to be reported.
+
+It is the responsibility of the [SDK](./README.md#sdk) to decide how to handle
+duplicates. For example, during the callback invocation if two measurements
+`value=1, attributes={pid:4 bitness:64}` and `value=2, attributes={pid:4,
+bitness:64}` are reported, the SDK can decide to drop the entire data, pick the
+last one, take all the data and add them together, or something else.
```
bogdandrutu, line 358: I think the API should say that duplicates are not allowed, but if that happens the [SDK] will decide how to handle that error case following the not throwing exception rule.

specification/metrics/new_api.md
```diff
@@ -308,6 +308,108 @@ counterPowerUsed.Add(13.5, new PowerConsumption { customer = "Tom" });
...
+        (10465,    ("pid", 880), ("bitness", 32)),
+    )
+
+page_faults_observable_counter = meter.create_observable_counter(name="PF", description="process page faults", pf_callback)
+```
+
+```python
+# Python
+
+def pf_callback(observer):
```
bogdandrutu, line 384: s/observer/observation?

observation.record()? or use `Result.observe` as we had before.

specification/metrics/new_api.md
bogdandrutu: As well as the other unrelated change, consider to split it in a separate PR.

specification/metrics/new_api.md
```diff
@@ -308,6 +308,108 @@ counterPowerUsed.Add(13.5, new PowerConsumption { customer = "Tom" });
 counterPowerUsed.Add(200, new PowerConsumption { customer = "Jerry" }, ("is_green_energy", true));
 ```
 
+### ObservableCounter
+
+`ObservableCounter` is an asynchronous Instrument which reports
+[monotonically](https://wikipedia.org/wiki/Monotonic_function) increasing
```
jonatan-ivanov, line 314: Would it make sense calling out that the `Counter` is monotonic too?

from
>`Counter` is a synchronous Instrument which supports non-negative increments.

to
>`Counter` is a synchronous Instrument which supports non-negative increments (monotonic).

bogdandrutu: The increments are not monotonic, but the sum of them is

reyang: I've create an issue #1593, will submit a separate PR.

specification/metrics/new_api.md
reyang: Good catch, fixed.

specification/metrics/new_api.md
reyang: I've updated the wording.

specification/metrics/new_api.md
reyang: Removed from this PR for now.

specification/metrics/new_api.md
reyang: Removed from this PR, will create a separate PR.

specification/metrics/new_api.md
reyang: Fixed. Use the `result` as what is in the [OTEP0072](https://github.com/open-telemetry/oteps/blob/main/text/metrics/0072-metric-observer.md).

specification/metrics/new_api.md
reyang: Separate PR created #1594.

specification/metrics/new_api.md
noahfalk: I am OK with 'ObservableFoo', but thinking upon our SIG discussion perhaps we'd be a little happier with the pattern 'FooObserver'?
**Pros:**
1. I think there was a valid complaint that 'ObservableCounter' isn't really a counter because it doesn't have counter operations on it like Add() or Increment(). 'CounterObserver' instead implies that the user has a counter (which they do) and this instrument is going to be observing the state of their counter (which it will).
2. "Foo" and "FooObserver" will sort alphabetically close together and one is a prefix of the other. This makes it likely they will appear close together in API listings and IDE auto-completion tools. This will aid users in discovering the types are related.
**Cons:** None I could come up with relative to the existing suggestions. In general I could imagine that users may not have an intuition about what an 'Observer' or 'Observable' instrument would do.

specification/metrics/new_api.md
noahfalk: A request (for now or later): Is there a good writeup anywhere that describes why it is important to distinguish monotonic and non-monotonic use-cases? I don't expect the spec to carry all the rationale inline but a link would be very useful.

I'll admit I am currently in the camp of not understanding the importance of it. Selfishly I'd love a link for myself but I am also guessing that I won't be the only one who has this question. I searched around and found some occasional mentions in the issues, but nothing that seemed intended to explain it to someone who wasn't already fairly well versed in the space.

specification/metrics/new_api.md
reyang: I was trying to avoid `FooObserver` since `Observer` serves different purpose in some languages, e.g. Java https://docs.oracle.com/javase/7/docs/api/java/util/Observer.html.

Want to see the feedback from other folks.

reyang: `CounterFunc` is another possible name that we could consider https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#CounterFunc.

specification/metrics/new_api.md
```diff
@@ -308,6 +312,109 @@ counterPowerUsed.Add(13.5, new PowerConsumption { customer = "Tom" });
...
+
+page_faults_observable_counter = meter.create_observable_counter(name="PF", description="process page faults", pf_callback)
+```
+
+```python
+# Python
+
+def pf_callback(result):
+    # Note: in the real world these would be retrieved from the operating system
+    result.Observe(8,        ("pid", 0),   ("bitness", 64))
```
victlu, line 391: I understand there is a difference between sync Counter vs async Observable Counter in terms of the numeric value we provide. For sync, we provide an incremental value (to be sum by sdk), but for async we provide the final sum value itself. In this latter case, how do we guarantee (or know when) the sum value is monotonic?

specification/metrics/new_api.md
reyang: The 2nd point from @noahfalk - "This makes it likely they will appear close together in API listings and IDE auto-completion tools. This will aid users in discovering the types are related." - sounds like a big plus to me.

@bogdandrutu @jmacd @jonatan-ivanov @jsuereth @open-telemetry/specs-metrics-approvers please give your vote here.

specification/metrics/new_api.md
victlu: One approach is to provide an result.Add() function thus avoiding the issue. However, it does require our users to calculate the delta for what they want to report.

reyang: I think the API doesn't enforce/guarantee it.

specification/metrics/new_api.md
jonatan-ivanov: @noahfalk I would love to see such a writeup too. :)
I think I can explain it from the high level.

Let's say you want a counter that is incremented by one for every incoming request. You have low traffic, so the counter is at 10 at the end of the first second, 20 at the second and by the clock hits 10 seconds, you processed 100 requests. If you plot this, you will only see a slope which is not very informative. Also, if your service runs long enough, your counter might overflow and you will see a huge drop. The same thing happens when you restart the server.  

Because of these, looking at the rate of increase (the first derivative of the slope by time) is more informative since you will see by how much your value increased instead of the values themselves, e.g.:
`0-10-20-30-40-50-60-70-80-90-100`
vs.
`10-10-10-10-10-10-10-10-10-10`

The monotonicity of a property makes it possible to the `rate` function to handle cases when the counter resets: overflow or server restart.

specification/metrics/new_api.md
jmacd: @reyang I appreciate your thoughts about the terminology.

Note that the Prometheus client API has an `Observe()` method for (synchronous) Histogram and Summary instruments, for recording observations.

The OTel-Go prototype API for the observable counter has a signature `func(observer ObserverResult)` where `ObserverResult` has an `Observe()` method for recording (async) observations.

This has me leaning toward the Prometheus style where an instrument `Foo` have asynchronous form `FooFunc`, an observable, where the function receives an Observer from the SDK for recording observations. It has the sorting property identified by @noahfalk.

specification/metrics/new_api.md
jmacd: Agreeing with @jonatan-ivanov, the reason we care about monotonicity is that when viewed as a timeseries, it is almost always the case that monotonic series should be viewed as a rate, whereas non-monotonic series should be viewed as their actual value. In both cases, it is well-defined to ask for the opposite view, just much less common.

specification/metrics/new_api.md
jmacd: +1 the API doesn't enforce this. The SDK and data model specifications will both talk about this situation.

specification/metrics/new_api.md
```diff
@@ -304,6 +308,109 @@ counterPowerUsed.Add(13.5, new PowerConsumption { customer = "Tom" });
...
+  rule](#instrument-naming-rule).
+* An optional `unit of measure`, following the [instrument unit
+  rule](#instrument-unit).
+* An optional `description`, following the [instrument description
+  rule](#instrument-description).
+* A `callback` function.
+
+The `callback` function is responsible for reporting the
+[Measurement](#measurement)s. It will only be called when the Meter is being
+observed. Individual language client SHOULD define whether this callback
```
jmacd, line 345: I agree this should be on the individual language client. Is it worth calling out these separate scenarios?
 
 - Independent SDKs are independently being called
 - One SDK has called a callback, and it's taking a very long time to return

specification/metrics/new_api.md
```diff
@@ -304,6 +308,109 @@ counterPowerUsed.Add(13.5, new PowerConsumption { customer = "Tom" });
...
+
+* Return a list (or tuple, generator, enumerator, etc.) of `Measurement`s.
+* Use an observer argument to allow individual `Measurement`s to be reported.
+
+Duplicates are not allowed. If it happens, the [SDK](./README.md#sdk) can decide
+how to handle it. For example, during the callback invocation if two
+measurements `value=1, attributes={pid:4 bitness:64}` and `value=2,
+attributes={pid:4, bitness:64}` are reported, the SDK can decide to drop the
+entire data, pick the last one, take all the data and add them together, or
+something else.
```
jmacd, line 359: In this case, the former API specification (written by me) did say something more firm, and it may be OK to leave this level of ambiguity in the API specification. The OTLP data model can in some ways address this, if we specify in the SDK that all observations observed in a single callback invocation have an identical timestamp. This requires that individual calls to `Observe()` do not record individual timestamps, but instead all the observations are recorded at once (logically). This allows the consumer of OTLP data to identify simultaneous observations from a single resource, which has certain uses. This supports counting exact cardinality because sets of counters are "coherent", they can be used to form ratios.

For example, you can: (1) take all the observations of a counter for a single resource and timestamp, (2) eliminate duplicates because they are not allowed by the API specification, (3) sum them to form a total, translate each counter value into a gauge representing its portion of the total. This is something async counters can do that is more difficult to do with synchronous counter instruments.

jmacd: Let me formally propose this. Add:

```
A single timestamp will be used per invocation of the callback.
```

This allows the SDK to simply ignore duplicates ("pass through") and a downstream OTLP consumer can detect duplication. I am a bit worried that the observed timeseries data will depend on which choice the SDK makes: "pass through" and "add them together" will output different results from "pick the last one".

If we follow this single-timestamp proposal and allow that a metrics processor can de-duplicate observations, then I'd prefer we do not allow "add them together" behavior. Another way of framing this, is that I want to require the aggregator to de-duplicate before adding together for the sake of correctness, and this can be stated as part of the data model impacting processors, not producers or exporters. Producers (especially SDKs) can decide to ignore duplicates, passing them through, or it can be configured to choose a single value, but it MUST not combine values.

I like to think of this in physical terms. In physics, it is not possible to observe two values of a measurement at the same point time and space. The API can safely ignore this as long as aggregators SHOULD de-duplicate before aggregating (IMO).

specification/metrics/new_api.md
noahfalk: Thanks! If the primary (and only?) goal is to indicate whether viewing by rate is the preferred aggregation, perhaps that is how we should be defining the instrument in the spec/user documentation?

specification/metrics/new_api.md
noahfalk: I'm happy with "CounterFunc". The name should also make sense to people coming from Micrometer which has the similar name "FunctionCounter." Our API review folks may request that we not abbreviate and use "CounterFunction", unless the abbreviation had particular importance. Does anyone consider the use of 'Func' over 'Function' important?

specification/metrics/new_api.md
```diff
@@ -304,6 +308,109 @@ counterPowerUsed.Add(13.5, new PowerConsumption { customer = "Tom" });
...
+* An optional `description`, following the [instrument description
+  rule](#instrument-description).
+* A `callback` function.
+
+The `callback` function is responsible for reporting the
+[Measurement](#measurement)s. It will only be called when the Meter is being
+observed. Individual language client SHOULD define whether this callback
+function needs to be reentrant safe / thread safe or not.
+
+Individual language client can decide what is the idomatic approach. Here are
```
aabmass, line 348:
```suggestion
Individual language client can decide what is the idiomatic approach. Here are
```

specification/metrics/new_api.md
aabmass, line 345:
```suggestion
observed. Individual language clients SHOULD define whether this callback
```

specification/metrics/new_api.md
reyang: Thanks @jmacd. I've updated this section, please take another look.

specification/metrics/new_api.md
aabmass: > I understand there is a difference between sync Counter vs async Observable Counter in terms of the numeric value we provide. For sync, we provide an incremental value (to be sum by sdk), but for async we provide the final sum value itself.

Can we call this point out specifically, maybe like a "calling convention" section? I've noticed a lot of confusion on this in the current spec e.g. https://github.com/open-telemetry/opentelemetry-js/issues/1564

aabmass (approved):

specification/metrics/new_api.md
reyang: No strong opinion but I do _slightly_ prefer `func` vs. `function`:
1. It seems `func` is a widely used thing, e.g. C99 `__func__`, .NET `System.Func<>` (also Prometheus .NET's `CounterFunc`).
2. It seems that `new FooFunc(func)` looks more consistent? `new FooFunction(func)` is a bit weird to me (and in many languages folks cannot use `new FooFunction(function)` since `function` is a reserved word.

@noahfalk do you think `CounterFunc` will introduce confusion with the .NET `System.Func<>`?

specification/metrics/new_api.md
reyang: Thanks for the catch! Fixed.

specification/metrics/new_api.md
jonatan-ivanov: I like Func/Function/Functional name more than `ObservableCounter`/`CounterObserver`.
I think Micrometer calls it [`FunctionCounter`](https://micrometer.io/docs/concepts#_function_tracking_counters) because it is a counter over a function (a special type of Counter).

noahfalk: > @noahfalk do you think CounterFunc will introduce confusion with the .NET System.Func<>?

I don't think it will. I was mostly just trying to predict possible feedback we'd hear from the API review team. They may have no issue with "CounterFunc" at all.

reyang: I took the name `CounterFunc` based on the discussions here.

hdost (approved):
Are any of the open comments actually un-resolved at this point?

specification/metrics/new_api.md
```diff
@@ -304,6 +308,110 @@ counterPowerUsed.Add(13.5, new PowerConsumption { customer = "Tom" });
...
+}
+
+IAtomicClock clock = AtomicClock.Connect();
+
+var obCaesiumOscillates = meter.CreateCounterFunc<UInt64>("caesium_oscillates", clk => clk.GetCaesiumOscillates(), clock);
+```
+
+#### CounterFunc operations
+
+`CounterFunc` is only intended for asynchronous scenario, it does not provide
```
victlu, line 412: CounterFunc provides an Observe (naming to be decided) operation.  This Observe operation is either the return value of the callback or calling a function on an argument passed to the callback.

victlu: Also, you may want to state that the measurements taken with this instrument represents a Sum and any statement regarding monotonicity as well.

reyang: Good point. I've updated this section.

specification/metrics/new_api.md
reyang: I've updated the wording here, please take another look.

specification/metrics/new_api.md
reyang: Let's follow up on #1593.

reyang:
> Are any of the open comments actually un-resolved at this point?

So far all the outstanding/blocking comments are resolved.

There are some open discussions (e.g. nice to have some high level document describing why we want to distinguish monotonic vs. non-monotonic, and providing examples/suggestions on when to use sync vs. async instrumentation), these are non-blocking and can be addressed in separate PRs.

specification/metrics/new_api.md
```diff
@@ -304,6 +308,115 @@ counterPowerUsed.Add(13.5, new PowerConsumption { customer = "Tom" });
...
+// A simple scenario where only one value is reported
+
+interface IAtomicClock
+{
+    UInt64 GetCaesiumOscillates();
+}
+
+IAtomicClock clock = AtomicClock.Connect();
+
+var obCaesiumOscillates = meter.CreateCounterFunc<UInt64>("caesium_oscillates", clk => clk.GetCaesiumOscillates(), clock);
```
cijothomas, line 411:
```suggestion
var obCaesiumOscillates = meter.CreateCounterFunc<UInt64>("caesium_oscillates", clk => clk.GetCaesiumOscillates(), state: clock);
```
This might make it more obvious that `clock` is *the* state.

specification/metrics/new_api.md
```diff
@@ -304,6 +308,115 @@ counterPowerUsed.Add(13.5, new PowerConsumption { customer = "Tom" });
...
+
+A single timestamp MUST be used per invocation of the callback. Duplicates are
+not allowed. If it happens, the [SDK](./README.md#sdk) can decide how to handle
+it. For example, during the callback invocation if two measurements `value=1,
+attributes={pid:4 bitness:64}` and `value=2, attributes={pid:4, bitness:64}` are
+reported, the SDK can decide to simply let them pass through (so the downstream
+consumer can handle duplication), drop the entire data, pick the last one, or
+something else.
+
+The API SHOULD provide some way to pass `state` to the callback. Individual
```
victlu, line 366: The term `state` seems to imply more. Maybe it's just... some way to pass a user provided argument to the callback.

specification/metrics/new_api.md
```diff
@@ -304,6 +308,115 @@ counterPowerUsed.Add(13.5, new PowerConsumption { customer = "Tom" });
...
+}
+
+IAtomicClock clock = AtomicClock.Connect();
+
+var obCaesiumOscillates = meter.CreateCounterFunc<UInt64>("caesium_oscillates", clk => clk.GetCaesiumOscillates(), state: clock);
+```
+
+#### CounterFunc operations
+
+`CounterFunc` is only intended for asynchronous scenario. The only operation is
```
victlu, line 416: Just minor nit. I thought it might read better if the details of the callback (and it's operation) would belong in this section rather than the section above.

specification/metrics/new_api.md
```diff
@@ -304,6 +308,115 @@ counterPowerUsed.Add(13.5, new PowerConsumption { customer = "Tom" });
...
+// A simple scenario where only one value is reported
+
+interface IAtomicClock
+{
+    UInt64 GetCaesiumOscillates();
+}
+
+IAtomicClock clock = AtomicClock.Connect();
+
+var obCaesiumOscillates = meter.CreateCounterFunc<UInt64>("caesium_oscillates", clk => clk.GetCaesiumOscillates(), state: clock);
```
victlu, line 411: It's not 100% clear that the callback will pass the clock as an argument to the lambda as clk.  Also, not sure how we know that clk is of type IAtomicClock from this snippet of code.

specification/metrics/new_api.md
```diff
@@ -304,6 +308,115 @@ counterPowerUsed.Add(13.5, new PowerConsumption { customer = "Tom" });
...
+* An optional `description`, following the [instrument description
+  rule](#instrument-description).
+* A `callback` function.
+
+The `callback` function is responsible for reporting the
+[Measurement](#measurement)s. It will only be called when the Meter is being
+observed. Individual language client SHOULD define whether this callback
+function needs to be reentrant safe / thread safe or not.
+
+The callback function SHOULD NOT take indefinite amount of time. If multiple
```
victlu, line 348: Perhaps the wording should be the callback function SHOULD return immediately with minimum delay.

specification/metrics/new_api.md
```diff
@@ -304,6 +308,115 @@ counterPowerUsed.Add(13.5, new PowerConsumption { customer = "Tom" });
...
+        (10465,    ("pid", 880), ("bitness", 32)),
+    )
+
+page_faults_counter_func = meter.create_counter_func(name="PF", description="process page faults", pf_callback)
+```
+
+```python
+# Python
+
+def pf_callback(result):
```
victlu, line 390: Maybe `result` should be renamed to be `observer`? So, you can have observer.Observe(...). This is just to stay in sync with your previous statement above.

```
Use an observer argument to allow individual `Measurement`s to be reported.
```

specification/metrics/new_api.md
```diff
@@ -304,6 +308,115 @@ counterPowerUsed.Add(13.5, new PowerConsumption { customer = "Tom" });
...
+independent SDKs coexist in a running process, they MUST invoke the callback
+function(s) independently.
+
+Individual language client can decide what is the idomatic approach. Here are
+some examples:
+
+* Return a list (or tuple, generator, enumerator, etc.) of `Measurement`s.
+* Use an observer argument to allow individual `Measurement`s to be reported.
+
+A single timestamp MUST be used per invocation of the callback. Duplicates are
```
noahfalk, line 358: It isn't immediately obvious how the timestamp relates to duplicates because the user never specifies a timestamp. Perhaps text like this:
```
User code is recommended not to provide more than one measurement with the same attributes in a single callback.
If this does occur the SDK gets to decide how to handle it. For example, ...
```

specification/metrics/new_api.md
noahfalk: I suspect the code would not work as written. What I suggest we do for C# is make use of the closure:
```C#
var obCaesiumOscillates = meter.CreateCounterFunc<UInt64>("caesium_oscillates", () => clock.GetCaesiumOscillates());
```

If we did want to demonstrate an alterative implementation that passes state as an opaque blob of extra data then the signature would probably be CreateCounterFunc\<UInt64>(string, Action\<object>, object) and the usage needs a cast:
```C#
var obCaesiumOscillates = meter.CreateCounterFunc<UInt64>("caesium_oscillates", clk => ((IAtomicClock)clk).GetCaesiumOscillates(), state: clock);
```

specification/metrics/new_api.md
```diff
@@ -304,6 +308,115 @@ counterPowerUsed.Add(13.5, new PowerConsumption { customer = "Tom" });
 counterPowerUsed.Add(200, new PowerConsumption { customer = "Jerry" }, ("is_green_energy", true));
 ```
 
+### CounterFunc
+
+`CounterFunc` is an asynchronous Instrument which reports
+[monotonically](https://wikipedia.org/wiki/Monotonic_function) increasing
```
noahfalk, line 314: Who do we believe has the onus to enforce that the values are monotonic? Is it undefined behavior if the callback returns a value that is smaller than the one it returned on the last invocation?

specification/metrics/new_api.md
victlu: I think the intension may be to avoid closure by passing in a user argument:

```
            Func<int> funcAsArg = () => {
                return 121;
            };

            var intcounterfunc = meter.CreateCounterFunc<int>(
                name: "intcounterfunc", 
                callback: (observer, arg) => {
                    if (arg is Func<int> func)
                    {
                        var val = func();
                        observer.Observe(val, ("location", "tere"), ("id", 221));
                    }
                }, 
                userarg: funcAsArg);
```

specification/metrics/new_api.md
```diff
@@ -304,6 +308,115 @@ counterPowerUsed.Add(13.5, new PowerConsumption { customer = "Tom" });
...
+
+* The `name` of the Instrument, following the [instrument naming
+  rule](#instrument-naming-rule).
+* An optional `unit of measure`, following the [instrument unit
+  rule](#instrument-unit).
+* An optional `description`, following the [instrument description
+  rule](#instrument-description).
+* A `callback` function.
+
+The `callback` function is responsible for reporting the
```
noahfalk, line 343: I think everyone discussing understands it, but I didn't notice any text in the spec that says what the measurement returned by the callback means. We should say that is the absolute value of the counter for a given attribute set at the time the callback is invoked. To determine the reported rate the counter is changing, the difference between successive measurements is used.

noahfalk (approved):
I added some suggestions, but I am also happy to accept the changes as-is and follow up on specific adjustments/questions in future targetted PRs.

specification/metrics/new_api.md
noahfalk: is the point of avoiding the closure to demonstate what the alternative looks like without a closure or suggesting that using the closure in this case is bad and shouldn't be used? No qualms with showing alternatives, but for this particular case in C# I haven't seen any particular reason we'd want to avoid it.

specification/metrics/new_api.md
reyang: Initially in this PR the argument was called `observer`, and later we changed this to `result` based on this comment https://github.com/open-telemetry/opentelemetry-specification/pull/1590#discussion_r606378200.

I think "report" is fine, all the data are reported, either synchronously or asynchronously.

specification/metrics/new_api.md
victlu: I guess I'll ask @reyang for his rational for asking for `state`.  Is it for performance reasons, or for context reasons?

specification/metrics/new_api.md
reyang: This looks similar to the original commit. I updated the wording based on [this feedback](https://github.com/open-telemetry/opentelemetry-specification/pull/1590#discussion_r607378920). I personally think what @jmacd suggested is reasonable. Although I do agree that the `timestamp` concept here is not explained. Let me see if I can find a balance here.

specification/metrics/new_api.md
jmacd: I would say something like the default SDK SHOULD enforce monotonicity, and that it is an error when a callback returns a value less than a previously returned value. 

> I can still make a mathematically well defined rate function that is sum_of_measurements / time and there is no requirement to only accept positive inputs

I take it you are considering whether we could drop the monotonicity idea. Why should we use separate instruments for monotonic and non-monotonic streams? There are legacy arguments that this information should be preserved, but I don't like to use those arguments. One good reason is simply to help the user. If you have a monotonic instrument, we can detect when you use it incorrectly.

specification/metrics/new_api.md
jmacd: (I read this as a form of context, not a performance concern.)

specification/metrics/new_api.md
jmacd: Since "timestamp" is implicit in the API, we may say something instead like: "the API must treat observations from a single callback as logically taking place at a single instant, such that when recorded, observations from a single callback MUST be reported with identical timestamps."

jmacd (approved):

specification/metrics/new_api.md
noahfalk: > I take it you are considering whether we could drop the monotonicity idea

Yeah, but very open to the notion that I may not have the full picture. What I have read so far in the conversation suggests that users don't care about monotonicity, they care whether the data is shown as a rate or as an absolute value. If its true that rate vs. absolute value is all they care about then I think pros/cons of enforcing monotonicity anyways would be:
pros: 
1. some users may misuse the API and we correctly gave them error that helped them fix their code
cons: 
1. some users might want to report a negative rate and they will be frustrated if the SDK prevents that usage
2. If we add more instruments to have separate monotonic and non-monotonic rate options this adds complexity to all users when choosing the appropriate instrument for their use case
3. There is a (tiny) perf cost to verify every measurement, probably only relevant in the synchronous instrument case.

specification/metrics/new_api.md
reyang: Please refer to this discussion https://github.com/open-telemetry/opentelemetry-specification/pull/1590#discussion_r605427802.
We've taken the `state` approach instead of `argument` since argument can be one way of passing in the state, but necessarily the only way.

specification/metrics/new_api.md
reyang: It is for context/state.

specification/metrics/new_api.md
reyang: `Indefinite` is the right term here, `immediate` is hard to define (e.g. if the callback needs to access a sensor over network).
It is the same term used across the spec repo, for example https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#exportbatch.

specification/metrics/new_api.md
reyang: I've updated the wording based on the feedback from @jmacd and @noahfalk. I believe the new wording would resolve the comments here.

specification/metrics/new_api.md
reyang: I've added a note on this.

specification/metrics/new_api.md
reyang: I've updated the code example.

cijothomas (approved):
LGTM

specification/metrics/new_api.md
reyang: We've covered this topic during the SIG meeting that monotonic vs. non-monotonic is important for the user experience, and it is also widely adopted by many well established metrics systems/libraries.

So here goes my suggestion:
1. Keep the current approach - have dedicated monotonic instruments (e.g. Counter/CounterFunc)
2. Clarify in the SDK spec whether we want to the SDK to enforce monotonicity or not (I think I agree with @jmacd that the default SDK SHOULD enforce monotonicity).
3. Use #1593 to track the further clarification - for example, we might want to explain these concepts and put some examples in the README file.

noahfalk: > We've covered this topic during the SIG meeting that monotonic vs. non-monotonic is important for the user experience

Perhaps I am being dense but that wasn't my understanding of the conversation. To me it felt like I put forth a question and an idea. Nobody responded saying they loved it nor did they respond saying its bad. Perhaps others thought it wasn't a good idea and were being polite, or wanted time to think it over, or I simply didn't understand the feedback : )

@jmacd, I know at the end of the SIG meeting you said you felt like you didn't have adequate opportunity to respond and mentioned some concern that the conversation was calling into question too much from the original spec. I wasn't clear whether you were refering to my comments or other comments/questions that were in the vicinity. Certainly my aim is not to push you (or anyone else) into a position you disagree with and I welcome your thoughts. So far I get the impression you aren't particularly excited about pivoting the definition of Counter to mean "displays as rate by default" with no enforcement of monotonicity. I don't know if anything I have said since was convincing or you still feel that being able to report negative measurements as errors is the most compelling concern at play so we should lay the issue to rest with that as the conclusion?

> So here goes my suggestion:

To me the issue appeared unresolved. I don't want to hold up the PR and I am fine accepting that Counter defined to be monotonic is the presumptive outcome at this point. I am hoping there will be a little more discussion in #1593, ideally considering the alternatives, but if not that then at least to clarify and document the rationale. Assuming that Counter is monotonic, having the SDK be the enforcement makes sense to me. Thanks all!

PR #1595, opened 2021-04-04
[semantic conventions] trivial: Amend Azure FaaS Description

Sherlouk:
Just a quick copy+paste fx - Azure currently has an incorrect description associated with it. This fixes it 😄

Might I propose a new analysis step in CI which detects when multiple entries have the same name or brief? I can't think of a case where that is expected behaviour and is typically a sign of copy, pasting and then forgetting.


carlosalberto (approved):

reyang (approved):
LGTM.

SergeyKanzhelev (approved):

SergeyKanzhelev:
> Might I propose a new analysis step in CI which detects when multiple entries have the same name or brief? I can't think of a case where that is expected behaviour and is typically a sign of copy, pasting and then forgetting.

Please file an issue or suggest a PR here: https://github.com/open-telemetry/build-tools/tree/main/semantic-conventions

SergeyKanzhelev:
trivial, merging

PR #1596, opened 2021-04-04
Add Device (Mobile) and OS Version Resource Semantic Attributes

### LINKED ISSUE #1499: Add Mobile specific Semantic conventions

I'd like to define mobile device/application specific semantic conventions.

**Additional context.**
The current semantic conventions are mostly container/service centric.
Mobile applications are an important part of the software stack, and need to be monitored just like any other link in the chain. There should be some predefined conventions around mobile specific platform details in the resource spec. 

Ideally the following should be added: 
- `device.id` : a unique identifier for the device (e.g.: vendor identifier for iOS)
- `device.model` : model of the device (.e.g `iPhone3,4`)

It would also be useful to have a parsable os.version

Sherlouk:
Fixes #1499

Hey folks 👋 

Apologies, I'm not fully clued up on the running process of adding new semantic attributes, only just starting to contribute to OpenTelemetry. I've created this pull request based on the suggestions of the issue above as well as my own beliefs - but I'm more hoping this will prompt a conversation. Please let me know if there's a better way to go about this.

## Changes

All changes are purely additive, no breaking changes.

### OS Additions

I've added an operating system name attribute, I believe this to be different enough to the existing `os.type` to justify its own attribute. While `os.type` will specify a more generalised "Linux" value (in this example), `os.name` can be the specific distro such as Ubuntu or Fedora. In a mobile world, from which I'm from, the `os.name` allows us to distinguish between multiple Darwin based Apple OS's such as iOS and watchOS.

I've also added an `os.version` value. This is intended to be machine readable and to follow the existing standards of what a version attribute should be.

### Device Additions

I've added a `device.model` value which is the machine readable model identifier of the device. Importantly this is not necessarily the same as the promotional or market/consumer friendly name - though this could be an optional extra attribute if people think it's suitable. The reason I've opted for this method is because on both iOS and Android it is not currently possible to get the market model name without large switch statements which become quickly out of date - it is more feasible to have mapping logic (through a processing pipeline) at the collector level or indeed at the visualisation layer.

I've also added a `device.id` value which represents a unique device identifier.

### DIFF
--- a/semantic_conventions/resource/device.yaml
+++ b/semantic_conventions/resource/device.yaml
@@ -1,0 +1,20 @@
+groups:
+  - id: device
+    prefix: device
+    brief: >
+        The device on which the process represented by this resource is running.
+    attributes:
+      - id: id
+        type: string
+        brief: 'A unique identifier representing the device'
+        note: >
+          For example, in an iOS application this might be the 
+          [vendor identifier](https://developer.apple.com/documentation/uikit/uidevice/1620059-identifierforvendor).
+        examples: ['2ab2916d-a51f-4ac8-80ee-45ac31a28092']
+      - id: model
+        type: string
+        brief: 'The model identifier for the device'
+        note: >
+          It's recommended this value represents a machine readable version of the model identifier rather than
+          the market or consumer-friendly name of the device.
+        examples: ['iPhone3,4', 'SM-G920F']
--- a/semantic_conventions/resource/device.yaml
+++ b/semantic_conventions/resource/device.yaml
@@ -8,11 +8,14 @@ groups:
         type: string
         brief: 'A unique identifier representing the device'
         note: >
-          For example, in an iOS application this might be the
-          [vendor identifier](https://developer.apple.com/documentation/uikit/uidevice/1620059-identifierforvendor).
-          Caution should be taken when storing personal data or anything which
-          can identify a user. GDPR and data protection laws may apply, ensure
-          you do your own due diligence.
+          The device identifier MUST only be defined using the values outlined below. This value is not an advertising
+          identifier and MUST NOT be used as such.
+          On iOS (Swift or Objective-C), this value MUST be equal to the [vendor identifier](https://developer.apple.com/documentation/uikit/uidevice/1620059-identifierforvendor).
+          On Android (Java or Kotlin), this value MUST be equal to the Firebase Installation ID or a globally unique
+          UUID which is persisted across sessions in your application. More information can be found [here](https://developer.android.com/training/articles/user-data-ids)
+          on best practices and exact implementation details.
+          Caution should be taken when storing personal data or anything which can identify a user. GDPR and
+          data protection laws may apply, ensure you do your own due diligence.
         examples: ['2ab2916d-a51f-4ac8-80ee-45ac31a28092']
       - id: model.identifier
         type: string
--- a/semantic_conventions/resource/device.yaml
+++ b/semantic_conventions/resource/device.yaml
@@ -8,13 +8,24 @@ groups:
         type: string
         brief: 'A unique identifier representing the device'
         note: >
-          For example, in an iOS application this might be the 
+          For example, in an iOS application this might be the
           [vendor identifier](https://developer.apple.com/documentation/uikit/uidevice/1620059-identifierforvendor).
+          Caution should be taken when storing personal data or anything which
+          can identify a user. GDPR and data protection laws may apply, ensure
+          you do your own due diligence.
         examples: ['2ab2916d-a51f-4ac8-80ee-45ac31a28092']
-      - id: model
+      - id: model.identifier
         type: string
         brief: 'The model identifier for the device'
         note: >
-          It's recommended this value represents a machine readable version of the model identifier rather than
-          the market or consumer-friendly name of the device.
+          It's recommended this value represents a machine readable version of
+          the model identifier rather than the market or consumer-friendly name
+          of the device.
         examples: ['iPhone3,4', 'SM-G920F']
+      - id: model.name
+        type: string
+        brief: 'The marketing name for the device model'
+        note: >
+          It's recommended this value represents a human readable version of the
+          device model rather than a machine readable alternative.
+        examples: ['iPhone 6s Plus', 'Samsung Galaxy S6']
--- a/semantic_conventions/resource/os.yaml
+++ b/semantic_conventions/resource/os.yaml
@@ -49,6 +49,16 @@ groups:
       - id: description
         type: string
         brief: >
-            Human readable (not intended to be parsed) OS version information,
-            like e.g. reported by `ver` or `lsb_release -a` commands.
+          Human readable (not intended to be parsed) OS version information,
+          like e.g. reported by `ver` or `lsb_release -a` commands.
         examples: ['Microsoft Windows [Version 10.0.18363.778]', 'Ubuntu 18.04.1 LTS']
+      - id: name
+        type: string
+        brief: 'Human readable operating system name.'
+        examples: ['iOS', 'Android', 'Ubuntu']
+      - id: version
+        type: string
+        brief: >
+          The version string of the operating system as defined in
+          [Version Attributes](../../resource/semantic_conventions/README.md#version-attributes).
+        examples: ['14.2.1', '18.04.1']
--- a/specification/resource/semantic_conventions/device.md
+++ b/specification/resource/semantic_conventions/device.md
@@ -1,0 +1,18 @@
+# Device
+
+**Status**: [Experimental](../../document-status.md)
+
+**type:** `device`
+
+**Description**: The device on which the process represented by this resource is running.
+
+<!-- semconv device -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `device.id` | string | A unique identifier representing the device [1] | `2ab2916d-a51f-4ac8-80ee-45ac31a28092` | No |
+| `device.model` | string | The model identifier for the device [2] | `iPhone3,4`; `SM-G920F` | No |
+
+**[1]:** For example, in an iOS application this might be the  [vendor identifier](https://developer.apple.com/documentation/uikit/uidevice/1620059-identifierforvendor).
+
+**[2]:** It's recommended this value represents a machine readable version of the model identifier rather than the market or consumer-friendly name of the device.
+<!-- endsemconv -->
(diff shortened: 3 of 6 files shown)


bryce-b:
As a bias iOS developer, I like these changes. OS version is a critical metric for mobile developers in understanding their customers' usage, and in understanding/debugging potential issues. Being able machine parse os version in mobile apps is pretty indispensable.

specification/resource/semantic_conventions/device.md
```diff
@@ -0,0 +1,18 @@
...
+
+**Description**: The device on which the process represented by this resource is running.
+
+<!-- semconv device -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `device.id` | string | A unique identifier representing the device [1] | `2ab2916d-a51f-4ac8-80ee-45ac31a28092` | No |
+| `device.model` | string | The model identifier for the device [2] | `iPhone3,4`; `SM-G920F` | No |
+
+**[1]:** For example, in an iOS application this might be the  [vendor identifier](https://developer.apple.com/documentation/uikit/uidevice/1620059-identifierforvendor).
```
reyang, line 15: If the intention is to use `device.id` as the "unique identifier", need to clarify if the scope of uniqueness (e.g. if user A sells iPhone to user B who reimaged the phone, would the id remain the same or not). Do we expect all the vendors to use the same algorithm/implementation so they would get the same id, or each vendor has freedom to implement their own version (which means a device might have different `device.id` for different vendors, and different devices might have the same `device.id` if they are processed by different vendors).

Sherlouk: Uniqueness as determined by the device, the strict criteria and determination of this changes on a vendor by vendor basis. In the iOS example I've given here, vendor identifier is unique per user per device per author of app and will be changed if the device is reset to factory/sold/reimaged/etc.

It's useful in order to understand the number of unique devices in a given time range that have used a given application. 

We do not expect every SDK/app to use the exact same strategy to define this value, for some apps a different device identifier might be more useful. For example some SDKs create a different device identifier whenever the app is reinstalled, this might be better for some people. We shouldn't enforce a specific format or approach to defining this value.

weyert: Do we want to highlight in these specs that sending these attributes can potentially personal data and the GDPR will apply? Same for potential other conventions were we send the ip address?

Sherlouk: I think that's a really fair and important point - I'm happy to update the wording here but might be worth a wider discussion on adding that as a general note for other attributes.

Do you have any specific proposal on the wording? I was thinking maybe something like:

```
Caution should be taken when storing personal data or anything which can identify a user. GDPR and data protection laws may apply, ensure you do your due diligence.
```

weyert: I am liking your wording. Let's see what other people think :)

semantic_conventions/resource/os.yaml
```diff
@@ -49,6 +49,16 @@ groups:
       - id: description
         type: string
         brief: >
-            Human readable (not intended to be parsed) OS version information,
-            like e.g. reported by `ver` or `lsb_release -a` commands.
+          Human readable (not intended to be parsed) OS version information,
+          like e.g. reported by `ver` or `lsb_release -a` commands.
         examples: ['Microsoft Windows [Version 10.0.18363.778]', 'Ubuntu 18.04.1 LTS']
+      - id: name
```
rakyll, line 55: I wonder if this would qualify as a "distro".

Sherlouk: I did consider this, but I personally thought this was a little too specific to the Linux OS. I would never really refer to iOS as a "distro" of Darwin, for example.  Do you disagree with that?

semantic_conventions/resource/device.yaml
```diff
@@ -0,0 +1,20 @@
...
+        brief: 'A unique identifier representing the device'
+        note: >
+          For example, in an iOS application this might be the 
+          [vendor identifier](https://developer.apple.com/documentation/uikit/uidevice/1620059-identifierforvendor).
+        examples: ['2ab2916d-a51f-4ac8-80ee-45ac31a28092']
+      - id: model
+        type: string
+        brief: 'The model identifier for the device'
+        note: >
+          It's recommended this value represents a machine readable version of the model identifier rather than
```
johnbley, line 18: I think it's fine to make a semantic distinction here between "machine-readable" and "human-readable".  What attribute name would then be used for the human-friendly translation? (i.e. if `device.model=iPhone13,1` then `???=iPhone 12 Mini`).  `device.model.name`?   Use `device.model.version` for machine-readable and `device.model.name` for human-readable?

Sherlouk: I can see where you're coming from, though perhaps `device.model.identifier` would be more appropriate? How do you feel about that?

I know speaking from a purely mobile perspective it would be uncommon for `device.model.name` to be set (see my PR comment for an explanation why) but this might be more useful for other audiences.

Sherlouk:
Updated to hopefully address some of the ideas mentioned above 🙂

johnbley (approved):

carlosalberto (approved):

semantic_conventions/resource/os.yaml
carlosalberto: Ping @rakyll

semantic_conventions/resource/device.yaml
```diff
@@ -0,0 +1,31 @@
...
+        The device on which the process represented by this resource is running.
+    attributes:
+      - id: id
+        type: string
+        brief: 'A unique identifier representing the device'
+        note: >
+          For example, in an iOS application this might be the
+          [vendor identifier](https://developer.apple.com/documentation/uikit/uidevice/1620059-identifierforvendor).
+          Caution should be taken when storing personal data or anything which
+          can identify a user. GDPR and data protection laws may apply, ensure
```
anuraaga, line 14: For reference, we have a user ID attribute

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#general-identity-attributes

It doesn't currently have any notes on PII but probably should

semantic_conventions/resource/device.yaml
```diff
@@ -0,0 +1,31 @@
...
+    prefix: device
+    brief: >
+        The device on which the process represented by this resource is running.
+    attributes:
+      - id: id
+        type: string
+        brief: 'A unique identifier representing the device'
+        note: >
+          For example, in an iOS application this might be the
+          [vendor identifier](https://developer.apple.com/documentation/uikit/uidevice/1620059-identifierforvendor).
```
anuraaga, line 12: The vendor identifier seems to be more useful for observability than ads identifier, which I suspect everyone agrees we don't want, so this seems ok to me. Still, the first thing I thought of when seeing `device.id` was that it's the ads identifier, so I wonder if there's a better name than just `id`. Can't come up with a suggestion though...

Considering how dangerous for privacy issues a bad implementation of this attribute could be (e.g., using the ads identifier), instead of loose wording like `For example`, maybe we can just outline the exact value that needs to be used on a set of devices. I'm guessing that means starting with 2, iOS and Android - and we add text that this id MUST NOT be filled on other devices until the specification has been added for how to fill it.

semantic_conventions/resource/device.yaml
Sherlouk: This was kind of alluded to before, but thanks for the specific example. I think there are a handful of values which **could** be implemented poorly, and could have an impact on GDPR and such due to it being PII.

I feel like more values aught to have some form of warning for due diligence.

Sherlouk:
Still very keen to get this merged, just unsure on next steps - can anybody advise? 🙂 Thanks

semantic_conventions/resource/device.yaml
carlosalberto: Ping @Sherlouk

carlosalberto:
@Sherlouk Maybe this?

> maybe we can just outline the exact value that needs to be used on a set of devices. I'm guessing that means starting with 2, iOS and Android - and we add text that this id MUST NOT be filled on other devices until the specification has been added for how to fill it.

carlosalberto:
@bryce-b mentioned he started trying out OTel Java with Android, so he may have an opinion about this matter.

semantic_conventions/resource/device.yaml
Sherlouk: I can definitely add specific examples of which platforms to use, and how to do it - but I'd need an example for how to do this on Android as I'm not actively sure on what the code would be.

If anybody has precise wording I'm happy to update 👍

weyert (approved):

semantic_conventions/resource/device.yaml
Sherlouk: Potential new wording:

```
The device identifier MUST only be defined using the values outlined below. This value is not an advertising identifier and MUST NOT be used as such.

On iOS (Swift or Objective-C), this value MUST be equal to the [vendor identifier](https://developer.apple.com/documentation/uikit/uidevice/1620059-identifierforvendor).

On Android (Java or Kotlin), this value MUST be equal to the Firebase Installation ID or a globally unique UUID which is persisted across sessions in your application. More information can be found [here](https://developer.android.com/training/articles/user-data-ids) on best practices and exact implementation details.

Caution should be taken when storing personal data or anything which can identify a user. GDPR and data protection laws may apply, ensure you do your own due diligence.
```

Sherlouk:
Attached new wording above, let me know if it's suitable and I'll update PR 👍

semantic_conventions/resource/device.yaml
carlosalberto: @Sherlouk Sounds good to me. Please apply this so we can merge this PR

Sherlouk: Updated 👍

semantic_conventions/resource/device.yaml
```diff
@@ -0,0 +1,34 @@
...
+          the model identifier rather than the market or consumer-friendly name
+          of the device.
+        examples: ['iPhone3,4', 'SM-G920F']
+      - id: model.name
+        type: string
+        brief: 'The marketing name for the device model'
+        note: >
+          It's recommended this value represents a human readable version of the
+          device model rather than a machine readable alternative.
+        examples: ['iPhone 6s Plus', 'Samsung Galaxy S6']
```
anuraaga, line 34: Just to confirm, are there APIs available on iOS and Android to provide the identifier and the name? If only one can be provided and the other would require a map of sort, there wouldn't be a need to include it in telemetry I guess

anuraaga (approved):

semantic_conventions/resource/device.yaml
Sherlouk: It all depends on the platform and application. The device name is unlikely (but not impossible) to be specified for iOS since it's difficult to maintain, but other platforms may make this a lot easier. The inverse can also be true. Both have their purpose, in my opinion.

carlosalberto:
We are ready to merge. We can iterate as we go. Could you please update the CHANGELOG entry to be in the current unreleased cycle @Sherlouk ? so sorry for that, just realized it now ;(

Sherlouk:
No worries, updated 👍

PR #1598, opened 2021-04-06
Have consistent formatting for semantic convention enums

### LINKED ISSUE #1519: Consistently format semantic convention enums
Semantic convention enums currently have inconsistent formatting. For example, `os.type` values are all capital (https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/resource/os.yaml#L10) whereas `cloud.infrastructure_service` values are underscored lowercase (https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/resource/cloud.yaml#L45).

With the new requirement of [autogenerating the semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/pull/1515), formatting in the YAML files is important for the code generator developers. Bless a format and consistently use it everywhere.

rakyll:
Currently, semantic convention enum values are not consistent.
For example, os.type values are all capitals whereas
cloud.infrastructure_service values are underscored lowercase.
This change improves the inconsistencies.

We also expect language implementations to autogenerate code
from enum values. Each language has their own conventions for
constant variable identifiers and we expect the consistent
formatting is going to help the language implementors.

Fixes #1519.


semantic_conventions/resource/os.yaml
```diff
@@ -11,38 +11,38 @@ groups:
...
+              value: 'linux'
               brief: "Linux"
-            - id: DARWIN
-              value: 'DARWIN'
+            - id: darwin
+              value: 'darwin'
               brief: "Apple Darwin"
-            - id: FREEBSD
-              value: 'FREEBSD'
+            - id: freebsd
```
rakyll, line 23: Not sure what to do with BSDs tbh. free_bsd or free_b_s_d would be impossibly difficult to read. That's why I'm keeping it simple here by keeping them as a single word.

reyang (approved):
LGTM.

carlosalberto (approved):

carlosalberto:
Probably we should also add this to some guidelines section, so future conventions stay uniform.

yurishkuro (approved):
I assume we already have scripts reading these YAML files. I suggest adding a check that all IDs are lowercase (and also ASCII, alphanum, no dashes, whichever rules we want to enforce). It's better than just a written convention, and checks should fail the build on violations.

carlosalberto:
cc @thisthat

rakyll:
Submitted https://github.com/open-telemetry/opentelemetry-specification/issues/1603 to add linting rules for enum values.

thisthat:
Adding such a check in our tool would be easy :) I am happy to work on that as soon as we agree on the rule set :)

semantic_conventions/trace/general.yaml
```diff
@@ -8,15 +8,15 @@ groups:
...
           members:
-            - id: IP.TCP
-              value: "IP.TCP"
-            - id: IP.UDP
-              value: "IP.UDP"
-            - id: IP
-              value: "IP"
+            - id: ip_tcp
+              value: "ip_tcp"
+            - id: ip_udp"
```
arminru, line 13:
```suggestion
            - id: ip_udp
```

codeboten (approved):

rakyll:
Ready to be merged now.

carlosalberto:
@alolita FYI, We have to decided to merge this PR right after we do (today) our monthly release.

Issue #1592, opened 2021-04-02
Log data model documentation terms: "key/value pair list" vs. "map"


jmm:
Hello,

The [Log Data Model](https://github.com/open-telemetry/opentelemetry-specification/blob/ef4612dd0b1f4266e6eca7177290ee2a5f087199/specification/logs/data-model.md) uses both the terms "key/value pair list" and "map", seemingly interchangeably, as in:

> Note: below we use type `any`, which can be a scalar value (number, string or boolean), or an array or map of values. Arbitrary deep nesting of values for arrays and maps is allowed (essentially allow to represent an equivalent of a JSON object).

> Field: `Attributes`
> Type: key/value pair list.

In terms of JSON or languages like Python, "key/value pair list" could be interpreted to mean an array (described as "list" in JSON prose) of 2-element arrays or list of tuples:

```json
[
  ["key", "value"]
]
```

So I think it would be good for clarity to use one consistent term -- whether it's "map", "object" (as in JSON), "dictionary", "hash", "record", or even "key/value pairs" (omitting "list") or "name/value pairs" (as in JSON prose) -- and clearly define it (e.g. the JSON serialization is object [unordered]).

I think "map" is fine as long as the definition is clear, but since the document seems to use "array" to refer to lists as JSON prose does, perhaps it makes sense to underpin the terminology with JSON and use "object" to refer to key/value pairs. The only current usages of "object" refer to JSON objects.

Thanks

reyang:
Discussed during the Friday issue triage meeting, the suggestion is to clarify in the spec that list / map is a language client decision, each language client can have freedom to decide what's the best/idiomatic/performant approach.

reyang:
@tigrannajaryan would you take this?

jmm:
Thanks @reyang.

So are you saying that current instances of "map" and "key/value pair list" would both be replaced by something like: "key/value collection", and at the client level that could be represented by, say, a map (a/k/a dictionary or object) and / or a list of key-value pairs?

To make it more concrete, let's say it's JavaScript, as far as the spec is concerned, something like `Attributes` could potentially be represented in either of the following ways at the client level?:

```js
someClient.log({
  Attributes: {
    key: "value",
  },
})

// or

someClient.log({
  Attributes: [
    ["key", "value"],
  ],
})
```

reyang:
[...]

PR #1604, opened 2021-04-07
Clarify "key/value pair list" vs "map" in Log Data Model

tigrannajaryan:
Parts of the data model documents were using the term "map", the others
were using the term "key/value pair list" without clearly telling why
and how they are the same.

This change clarifies and ensures that we refer to a map of key/value pairs
that can be represented in different way in different languages.

Resolves https://github.com/open-telemetry/opentelemetry-specification/issues/1592

### DIFF
--- a/specification/logs/data-model.md
+++ b/specification/logs/data-model.md
@@ -144,10 +144,10 @@ Note: below we use type `any`, which can be one of the following:
 
 - An array (a list) of `any` values,
 
-- A map of string keys to `any` values. The representation of the map of
-  key/values is language-dependent. Some languages may use built-in maps or
-  dictionaries, others may use a list of key/value pairs. It is required that
-  the implementation has traditional
+- A map of string keys to `any` values. This is denoted as `map<string, any>`
+  throughout this document. The representation of the map is language-dependent.
+  Some languages may use built-in maps or dictionaries, others may use a list of
+  key/value pairs. It is required that the implementation has traditional
   [map](https://en.wikipedia.org/wiki/Associative_array) semantics.
 
 Arbitrary deep nesting of values for arrays and maps is allowed (essentially
@@ -387,30 +387,27 @@ occurrence of the event coming from the same source. This field is optional.
 
 ### Field: `Resource`
 
-Type: map of key/value pairs.
+Type: `map<string, any>`.
 
 Description: Describes the source of the log, aka
-[resource](../overview.md#resources).
-"key" of each pair is a `string` and "value" is of `any` type. Multiple
-occurrences of events coming from the same event source can happen across time
-and they all have the same value of `Resource`. Can contain for example
-information about the application that emits the record or about the
-infrastructure where the application runs. Data formats that represent this data
-model may be designed in a manner that allows the `Resource` field to be
-recorded only once per batch of log records that come from the same source.
-SHOULD follow OpenTelemetry
+[resource](../overview.md#resources). Multiple occurrences of events coming from
+the same event source can happen across time and they all have the same value of
+`Resource`. Can contain for example information about the application that emits
+the record or about the infrastructure where the application runs. Data formats
+that represent this data model may be designed in a manner that allows the
+`Resource` field to be recorded only once per batch of log records that come
+from the same source. SHOULD follow OpenTelemetry
 [semantic conventions for Resources](../resource/semantic_conventions/README.md).
 This field is optional.
 
 ### Field: `Attributes`
 
-Type: map of key/value pairs.
+Type: `map<string, any>`.
 
-Description: Additional information about the specific event occurrence. "key"
-of each pair is a `string` and "value" is of `any` type. Unlike the `Resource`
-field, which is fixed for a particular source, `Attributes` can vary for each
-occurrence of the event coming from the same source. Can contain information
-about the request context (other than TraceId/SpanId). SHOULD follow
+Description: Additional information about the specific event occurrence. Unlike
+the `Resource` field, which is fixed for a particular source, `Attributes` can
+vary for each occurrence of the event coming from the same source. Can contain
+information about the request context (other than TraceId/SpanId). SHOULD follow
 OpenTelemetry
 [semantic conventions for Attributes](../trace/semantic_conventions/README.md).
 This field is optional.
@@ -642,13 +639,13 @@ Rest of SDIDs -> Attributes["syslog.*"]</td>
   </tr>
   <tr>
     <td>Dimensions</td>
-    <td>map of string to string</td>
+    <td>map&lt;string, string></td>
     <td>Helps to define the identity of the event source together with EventType and Category. Multiple occurrences of events coming from the same event source can happen across time and they all have the value of Dimensions. </td>
     <td>Resource</td>
   </tr>
   <tr>
     <td>Properties</td>
-    <td>map of string to any</td>
+    <td>map&lt;string, any></td>
     <td>Additional information about the specific event occurrence. Unlike Dimensions which are fixed for a particular event source, Properties can have different values for each occurrence of the event coming from the same event source.</td>
     <td>Attributes</td>
   </tr>
@@ -695,7 +692,7 @@ Rest of SDIDs -> Attributes["syslog.*"]</td>
   </tr>
   <tr>
     <td>fields</td>
-    <td>Map of any</td>
+    <td>map&lt;string, any></td>
     <td>Specifies a JSON object that contains explicit custom fields.</td>
     <td>Attributes</td>
   </tr>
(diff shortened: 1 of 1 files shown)


tigrannajaryan:
@jmm please review.

jmm:
@tigrannajaryan Thanks for working on this.

When I originally filed the issue, this is the outcome I was expecting. I thought that where it said "key/value pair list" it really meant map. But then I thought @reyang was saying the decision was that it would be up to each client what data structure(s) / format(s) they'd accept as input, possibly including lists for example. If you see my [example here](https://github.com/open-telemetry/opentelemetry-specification/issues/1592#issuecomment-812621875), I wouldn't consider the second version (an array of 2-element key/value arrays) to be consistent with the "map semantics" part of this description.

I wonder if it would be simplest to say that it's a client-dependent / language-dependent representation of a collection of `string` key and `any` value pairs that the client could serialize to a JSON object (not that it necessarily will).

To me, ability to represent the data as a JSON object seems like it covers the aspects that need to be consistent across clients and languages without prescribing what input they need to accept.

Each client will have to support and document the specific data structure(s) / format(s) it will accept as input in any case right? For example, in JavaScript both plain objects and Maps (with string keys) would satisfy the criteria of map semantics, but can't be used interchangeably without the client supporting it. And by the same token the client could choose to accept some other representation, like an array of 2-element key/value arrays, which it would be capable of converting to a plain object, Map, or JSON object.

---

In any case, I'd suggest considering some label that appears verbatim in all of the places it gets referenced -- for example, one of the following:

* KeyValueCollection

* map<string, any>

  This notation is used in the Google Cloud Logging mapping example.

* map of key/value pairs

So something like "KeyValueCollection" or "map<string, any>" could precede the definition on line 147 and then appear as the Type in the field definitions and that would make it really clear and searchable that they refer to the same thing.

---

There are also references to "key/value pair lists" in the Field Kinds section.

carlosalberto:
(There's also the situation regarding the rest of the specification, as we mainly use "list of key-value pairs")

tigrannajaryan:
> So something like "KeyValueCollection" or "map<string, any>" could precede the definition on line 147 and then appear as the Type in the field definitions and that would make it really clear and searchable that they refer to the same thing.

I like this suggestion. Updated. See if it is easier to understand now.

tigrannajaryan:
> (There's also the situation regarding the rest of the specification, as we mainly use "list of key-value pairs")

Good point. It may be worth extracting common definitions to another document and refer from here. I suggest we do it in a separate PR after we are happy with the re-wording I am doing here.

carlosalberto:
Doing it in a separate PR works great ;)

jmacd (approved):

jmm:
@tigrannajaryan  Thanks for the update, having that one definition of the type is great.

My only concern is whether this is a requirement you want to place:

> It is required that the implementation has traditional [map](https://en.wikipedia.org/wiki/Associative_array) semantics.

I had thought @reyang indicated a different preference. But also, this part seems inconsistent to me:

> others may use a list of key/value pairs

Maybe you have an example in mind of a language that would use a list of key/value pairs that has map semantics?

tigrannajaryan:
> @tigrannajaryan Thanks for the update, having that one definition of the type is great.
> 
> My only concern is whether this is a requirement you want to place:
> 
> > It is required that the implementation has traditional [map](https://en.wikipedia.org/wiki/Associative_array) semantics.
> 

Yes, I believe this is a requirement everywhere in Otel specs, the attributes are a map. This is not new.

> I had thought @reyang indicated a different preference. But also, this part seems inconsistent to me:
> 
> > others may use a list of key/value pairs
> 
> Maybe you have an example in mind of a language that would use a list of key/value pairs that has map semantics?

See Otel Collector's attributes map implementation for example. It is a key/value pair list where uniqueness of keys is enforced.

jmm:
@tigrannajaryan Ok thanks. I'm not sure where to start there (and I don't know Go), is it this?:

https://github.com/open-telemetry/opentelemetry-collector/blob/03c7bf98f7edf6e578aac303e4d8b767f117541e/consumer/pdata/common.go#L360

As I said, I don't know Go, but I think what I'm seeing there is a custom map interface (`AttributeMap`) wrapping some kind of list. And it also looks like Go has a map type, and it looks like `AttributeMap` is actually designed to consume those and populate the data into its internal list.

---

Let's say you're in Go and there are at least a few different ways you could represent attributes:

* A `map[string]Something`.

* An `AttributeMap` or something similar.

* Some kind of list (like whatever `orig` is in `AttributeMap`).

None of those seem to me like a "list of key/value pairs" that has map semantics. Whatever `orig` is doesn't have map semantics or it wouldn't need to be wrapped by `AttributeMap`, right? And the fact that `AttributeMap` is backed by a list is an implementation detail that whatever is consuming it shouldn't know about.

---

I think that highlights that what's relevant to this specification is the logical characteristics. So I feel like saying both of these things makes it harder to understand:

> may use a list of key/value pairs

> It is required that the implementation has traditional map semantics.

I think the clearest way to explain it would be to say something to the effect that logically it's an unordered `map<string, any>` with unique keys and it's outside the scope of this specification how a client or language would represent that. By way of illustration, it's something that _could_ be serialized as a JSON object with unique keys.

tigrannajaryan:
@jmm map definition reworded.

specification/logs/data-model.md
```diff
@@ -112,10 +112,8 @@ Value of type `any` can be one of the following:
 #### Type `map<string, any>`
 
 Value of type `map<string, any>` is a map of string keys to `any` values. The
-representation of the map is language-dependent. Some languages may use built-in
-maps or dictionaries, others may use a list of key/value pairs. It is required
-that the implementation has traditional
-[map](https://en.wikipedia.org/wiki/Associative_array) semantics.
+keys in the map are unique (duplicate keys are not allowed). The representation
+of the map is language-dependent.
```
jmacd, line 116: I agree with this rewording, and thanks @jmm for the rewording. 

@tigrannajaryan were you hoping to say something, not in the data model, but about the use of OTLP with protobuf and JSON transports, which both technically allow duplicate keys in the wire format of their built-in map<string, any> types?

I believe we are trying to say, _for the sake of the data model_, that when an OTLP map is interpreted in the form of a list, and there are duplicate values for a key present, that some kind of conflict has to be resolved? In this case, protobuf choses the last value among duplicates: no error. Should a valid OTLP JSON implementation be required to remove duplicates (in any order)? Are they allowed to return errors? See https://stackoverflow.com/questions/21832701/does-json-syntax-allow-duplicate-keys-in-an-object

jmacd (approved):

jmm:
Thanks @tigrannajaryan 🙌 and @jmacd.

That makes sense to me based on my understanding of the scope of this part of the spec.

specification/logs/data-model.md
tigrannajaryan: @jmacd yes, we should probably clarify that in OTLP (proto) repository.

tigrannajaryan: Added an issue to clarify in proto repo: https://github.com/open-telemetry/opentelemetry-proto/issues/296

zenmoto (approved):

carlosalberto (approved):

specification/logs/data-model.md
```diff
@@ -94,6 +94,30 @@ The Data Model aims to successfully represent 3 sorts of logs and events:
...
+#### Type `any`
+
+Value of type `any` can be one of the following:
+
+- A scalar value: number, string or boolean,
+
+- An array (a list) of `any` values,
+
+- A `map<string, any>`.
+
```
kumoroku, line 111: do we want to support binary here? I am thinking in particular of field `Body` which is on type `any`. I understand that we don't explicitly speak to it under [Field: Body](#field-body) but at least I personally have always considered for myself that the Body could just be a binary blob in some (likely legacy) cases.

jsuereth (approved):

specification/logs/data-model.md
tigrannajaryan: Yes, I think we should. Not just for logs, but generally for all Attributes in this spec. I submitted an issue to do it: https://github.com/open-telemetry/opentelemetry-specification/issues/1644

bogdandrutu (approved):

specification/logs/data-model.md
```diff
@@ -94,6 +94,32 @@ The Data Model aims to successfully represent 3 sorts of logs and events:
...
+#### Type `map<string, any>`
+
+Value of type `map<string, any>` is a map of string keys to `any` values. The
```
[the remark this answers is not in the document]
tigrannajaryan: I think map is well understood. If we see any confusion I will clarify.

PR #1614, opened 2021-04-13
Expand event-model description to more clearly delineate instrument v…

jsuereth:
…s. metric stream and some reasoning behind it.

Fixes ##1366

## Changes

Improve the documentation around the "Event Model" and the relationship of "Instruments" to "metrics"


specification/metrics/datamodel.md
```diff
@@ -146,26 +146,44 @@ OpenTelemetry fragments metrics into three interacting models:
...
-  instruments characterize a group of measurements. The numbers passed to adding
-  instruments define division, in the algebraic sense, while the numbers passed
-  to grouping instruments are generally not. Adding instrument values are always
-  parts of a sum, while grouping instrument values are individual measurements.
-- Monotonic vs. Non-Monotonic. The adding instruments are categorized by whether
-  the derivative of the quantity they express is non-negative. Monotonic
-  instruments are primarily useful for monitoring a rate value, whereas
-  non-monotonic instruments are primarily useful for monitoring a total value.
+The event model is where recording of data happens. Its foundation is made of
+[Instruments](api.md), which are used to record observations on data.  These raw
```
punya, line 150: "observations on data" is confusing. Does it mean the same thing as "raw events"? I would find it clearer if we stuck to events because the thing we are observing is generally some real-world thing like a process, a network, etc., not data.

specification/metrics/datamodel.md
```diff
@@ -146,26 +146,44 @@ OpenTelemetry fragments metrics into three interacting models:
...
+events are then transformed in some fashion before being sent to some other
+system.  OpenTelemetry metrics are designed such that the same instrument and
+events can be used in different ways to generate metric streams.
```
punya, line 153: :+1: It's great to call out the distinction. Is worth adding an example to make it even clearer?

specification/metrics/datamodel.md
```diff
@@ -146,26 +146,44 @@ OpenTelemetry fragments metrics into three interacting models:
...
+
+![Events → Streams](img/model-event-layer.png)
+
+While observation events could be reported directly to a backend, in practice
+this would be infeasible due to the shear volume of data used in observability
```
punya, line 158:
```suggestion
this would be infeasible due to the sheer volume of data used in observability
```

specification/metrics/datamodel.md
```diff
@@ -146,26 +146,44 @@ OpenTelemetry fragments metrics into three interacting models:
...
+systems, and the limited amount of network/cpu telemetry collection has
+available.  The best example of this is the Histogram metric where raw
```
punya, line 160:
```suggestion
systems, and the limited amount of network/cpu resources available for telemetry collection
purposes.  The best example of this is the Histogram metric where raw
```

specification/metrics/datamodel.md
```diff
@@ -146,26 +146,44 @@ OpenTelemetry fragments metrics into three interacting models:
...
+events are recorded in a compressed format rather than individual timeseries.
+
+while OpenTelemetry provides flexibility in how instruments can be transformed
```
punya, line 163:
```suggestion
While OpenTelemetry provides flexibility in how instruments can be transformed
```

specification/metrics/datamodel.md
```diff
@@ -146,26 +146,44 @@ OpenTelemetry fragments metrics into three interacting models:
...
+The [OpenTelemetry metric instruments](api.md) are designed around the
+following concerns:
+
+- Synchronous vs. Asynchronous collection.
+  - Synchronous instruments are those where an application/library records a
+    metric data point inline. In this scenario, OpenTelemetry *can* attach
+    context (e.g. baggage) to recorded data points.
+  - Asynchronous instruments are those where OpenTelemetry (not the application)
+    will execute a callback (or other similar mechanism) to pull data points
+    on demand.   This is generally done at regular intervals.
```
punya, line 176: Does this imply that data points used by async instruments can't have a context/baggage attached? E.g., if points to be collected in future are kept in a buffer, couldn't the context be attached at the time when they are put into the buffer?

specification/metrics/datamodel.md
```diff
@@ -146,26 +146,44 @@ OpenTelemetry fragments metrics into three interacting models:
...
+- Adding vs. Grouping aggregation.
+  - Adding instruments express a sum.  All points recorded via this instrument
+    are parts of a whole.
+  - Grouping instruments characterize a group of measurements.  All points
```
punya, line 180: Where does this leave things like quantiles?

specification/metrics/datamodel.md
```diff
@@ -146,26 +146,44 @@ OpenTelemetry fragments metrics into three interacting models:
...
+systems, and the limited amount of network/cpu telemetry collection has
+available.  The best example of this is the Histogram metric where raw
+events are recorded in a compressed format rather than individual timeseries.
+
+while OpenTelemetry provides flexibility in how instruments can be transformed
+into metric streams, the instruments are defined such that a reasonable default
+mapping can be provided.
+
+The [OpenTelemetry metric instruments](api.md) are designed around the
+following concerns:
```
punya, line 168: Is this supposed to be a full taxonomy of instruments? I.e. is it exhaustive? Or should we think of these as "think of these considerations to understand the design"?

specification/metrics/datamodel.md
jsuereth: More than the image, or is it worth diving into the image?

specification/metrics/datamodel.md
jsuereth: Reworded, PTAL.

specification/metrics/datamodel.md
jsuereth: I don't think so.   This is my reworking of @jmacd, so he can comment if this was meant to be exhaustive or descriptive.

specification/metrics/datamodel.md
jsuereth: Quantile is a form of grouping.

specification/metrics/datamodel.md
jsuereth: Yes, that is what we're saying.  

Given how baggage/context is propagated, you cannot  accurately attach context/baggage to async instrument recordings.  In fact, you can't determine if a counter (e.g.) needs to be reported against multiple people's context.

specification/metrics/datamodel.md
```diff
@@ -43,7 +43,7 @@ containing a higher-level model, about Metrics APIs and discrete input values,
 and a lower-level model, defining the Timeseries and discrete output values.
 The relationship between models is displayed in the diagram below.
 
-![Events  → Data → Timeseries Diagram](img/model-layers.png)
+![Events  → Streams → Timeseries Diagram](img/model-layers.png)
```
SergeyKanzhelev, line 46: does `total_latency` aggregation ever makes sense? It may not be the best idea to include example like this

SergeyKanzhelev, line 46: would it be possible to have layers named on the picture the same way as in comment  `Events → Streams → Timeseries`

specification/metrics/datamodel.md
```diff
@@ -146,26 +146,45 @@ OpenTelemetry fragments metrics into three interacting models:
...
-  the derivative of the quantity they express is non-negative. Monotonic
-  instruments are primarily useful for monitoring a rate value, whereas
-  non-monotonic instruments are primarily useful for monitoring a total value.
+The event model is where recording of data happens. Its foundation is made of
+[Instruments](api.md), which are used to record data observations via events.
+These raw events are then transformed in some fashion before being sent to some
+other system.  OpenTelemetry metrics are designed such that the same instrument
+and events can be used in different ways to generate metric streams.
+
+![Events → Streams](img/model-event-layer.png)
```
SergeyKanzhelev, line 155: does `total_latency` aggregation ever makes sense? It may not be the best idea to include example like this

SergeyKanzhelev (approved):
lgtm except the example on the picture. I suggest to fix it before merging

specification/metrics/datamodel.md
jsuereth: done

specification/metrics/datamodel.md
jsuereth: I do want an example that shows how a single instrument could turn into all of Sum, Histogram + Gauge.

While I can synthesize hair-brained scenarios where I think "total latency" might make sense (e.g. ridiculous statistics like How many days have users waited for our website in aggregate), You're right it's not a super useful example.  Going to take a day to brainstorm and looking for other ideas on this example.

specification/metrics/datamodel.md
jmacd: Somewhat related, I do not care to add APIs for asynchronous histogram observations (a.k.a. "GaugeHistogram", https://github.com/open-telemetry/opentelemetry-proto/issues/274), which are structurally equivalent to what you have when you buffer a number of histogram observations, so I want to clarify that I am opposed to buffered synchronous histogram measurements _and_ I am opposed to multi-valued asynchronous histogram measurements.

I prefer not to admit GaugeHistogram because it allows multiple numbers to cross the API boundary in a single event. I would prefer the user iterate through their histogram and make observations one by one, using an _anonymous_ label; I would then argue for metrics SDKs to erase anonymous labels by combining measurements into a histogram.  This is another example of the general rule: don't try to outperform the metric SDK: it should be able to construct a histogram with fewer copies than the application would.

specification/metrics/datamodel.md
jmacd: I still believe this is a useful concept, but I am worried we've lost the connection to the API design by now. There's something about how Gauge and Histogram instruments are the same semantics with different default aggregations: both group individual measurements.

The point is that Gauge and Histogram inputs are semantically different than Sum inputs, because of individuality. This is meant to help the user choose between Counter and Histogram, for example. @reyang I mean to follow up on last week's API/SDK SIG meeting, in which we discussed some of this.

specification/metrics/datamodel.md
```diff
@@ -146,26 +146,45 @@ OpenTelemetry fragments metrics into three interacting models:
...
+    context (e.g. baggage) to recorded data points.
+  - Asynchronous instruments are those where OpenTelemetry (not the application)
+    will execute a callback (or other similar mechanism) to pull data points
+    on demand.   This is generally done at regular intervals.
+- Adding vs. Grouping aggregation.
+  - Adding instruments express a sum.  All points recorded via this instrument
+    are parts of a whole.
+  - Grouping instruments characterize a group of measurements.  All points
+    recorded via this instrument are individual measurements.
+- Monotonic vs. Non-Monotonic (adding instruments only). These instruments are
```
jmacd, line 183: I worry that there are conflicting definitions here as compared with the suggestions I made above. I suspect it would be best to remove the discussion of monotonicity here.

specification/metrics/datamodel.md
jsuereth: The goal of this is NOT to define what instruments the API uses, but show the flexibility of the model to adapt to differing instruments.  

I'm thinking about doing the following:
- Focus on adapting instruments (OTel + non-otel) into the data model
- Focus on *general* concepts that may be used (quantiles, grouping, etc.)
- Direct to the API specification for specific instruments / instrument names.

This specification should be about WHAT metrics mean, not what instruments are available in the API.

specification/metrics/datamodel.md
jsuereth: i'm not sure I read much actionable feedback on current wording here.

+1 to directly specifying "don't buffer measurements" as a principle.  I highly agree with that, not sure if we should take a stand in this document or in the API.

specification/metrics/datamodel.md
```diff
@@ -146,26 +146,45 @@ OpenTelemetry fragments metrics into three interacting models:
...
+observability systems, and the limited amount of network/cpu telemetry
+collection resources available for telemetry collection purposes. The best
+example of this is the Histogram metric where raw events are recorded in a
+compressed format rather than individual timeseries.
+
+While OpenTelemetry provides flexibility in how instruments can be transformed
+into metric streams, the instruments are defined such that a reasonable default
+mapping can be provided.
+
+The [OpenTelemetry metric instruments](api.md) are designed around the
```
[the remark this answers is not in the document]
jsuereth: Done.   We'll evaluate later if we think it's needed.

specification/metrics/datamodel.md
jsuereth: Ok, this is removed from this PR given your suggestion to separate out that discussion.

reyang (approved):
LGTM.

bogdandrutu (approved):

SergeyKanzhelev:
@jmacd do you want more changes in PR or merge as-is for now?

PR #1617, opened 2021-04-14
Skeleton for the remaining metrics instruments

reyang:
Related to #1578 and #1590.

## Changes

So far we've made good progress on the `Counter` and `CounterFunc` APIs in #1578 and #1590. This PR tries to propose the names for the remaining instruments. Note that I intentionally left several `TODO`s just to help us focusing on the skeleton, once this PR got merged, I will send a follow-up PR to fill in the real meat.

When deciding which Instrument to use, here goes the decision tree:

* I want to count (taking a delta value) something => Is the value monotonically increasing?
  * YES => `Counter`
  * NO => `UpDownCounter`
* I want to measure something (by absolute value) => Do I need to add up the values?
  * NO => `GaugeFunc`
  * YES => Is the value monotonically increasing?
    * YES => `CounterFunc`
    * NO => `UpDownCounterFunc`
* I want to record something, and the statistics about this thing are likely to be meaningful => `Histogram`
  * I want to time something (e.g. duration) => `Histogram` (or `Timer` if we have it as a specialized version of `Histogram`)

Another way of making decision on Instrument:
* Sync (have access to the context)
  * Would statistics about this thing are likely to be meaningful?
    * YES => `Histogram`
    * NO => Is the value monotonically increasing?
      * YES => `Counter`
      * NO => `UpDownCounter`
* Async (have **no** access to the context)
  * Is the value additive (link to wikipedia) or not?
    * YES => Is the value monotonically increasing?
       * YES => `CounterFunc`
       * NO => `UpDownCounterFunc`
    * NO => `GaugeFunc`

After this PR, the only outstanding issues for the metrics API spec would be:
* Hint API (we've discussed during the SIG meeting that we will move forward to the SDK spec and come back to this topic after SDK spec is in shape)


specification/metrics/new_api.md
```diff
@@ -34,6 +34,15 @@ Table of Contents
   * [CounterFunc](#counterfunc)
     * [CounterFunc creation](#counterfunc-creation)
     * [CounterFunc operations](#counterfunc-operations)
+  * [Gauge](#gauge)
+    * [Gauge creation](#gauge-creation)
+    * [Gauge operations](#gauge-operations)
+  * [GaugeFunc](#gaugefunc)
+    * [GaugeFunc creation](#gaugefunc-creation)
+    * [GaugeFunc operations](#gaugefunc-operations)
+  * [Distribution](#distribution)
```
reyang, line 43: Regarding the name "Distribution", I found the following options by comparing several existing metrics APIs:
1. Histogram - I would not recommend this name since it suggests a particular use case, although I don't have a very strong opinion.
2. Summary - I personally think this shares the same problem with Histogram, and it could be confusing to folks who came from Prometheus world.
3. ValueRecorder - This was in the original API spec, I wonder if this name is way "too generic"?

specification/metrics/new_api.md
```diff
@@ -425,6 +438,78 @@ var obCaesiumOscillates = meter.CreateCounterFunc<UInt64>("caesium_oscillates",
 provided by the `callback`, which is registered during the [CounterFunc
 creation](#counterfunc-creation).
 
+### Gauge
```
reyang, line 441: We've talked about the name "gauge" in the SIG meetings. Here are my notes:
1. "Gauge" is a very confusing term, there are many metrics systems using the word "gauge", but each of them has slightly different meaning (e.g. Prometheus Gauge is different from Micrometer Gauge).
2. We've already used the term "gauge" in the [Data Model Specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#timeseries-model), and it is for timeseries model rather than event model.

specification/metrics/new_api.md
```diff
@@ -425,6 +438,78 @@ var obCaesiumOscillates = meter.CreateCounterFunc<UInt64>("caesium_oscillates",
...
+* the number of active requests
+* the number of items in a queue
+
+#### Gauge creation
+
+TODO
+
+#### Gauge operations
+
+##### Add
```
reyang, line 459: I want to get feedback here:
1. I think it is fine to provide both `Add` (which reports delta) and `Set` (which reports the absolute value) in the sync instrument.
2. I think we should only provide `Set` in the async instrument. If a callback is provided to multiple async instruments (which is something we might want to discourage?), the semantic of the delta value would be very murky.

specification/metrics/new_api.md
victlu: Just throwing this out here... I am not opinionated on this at this time.

Wiki have these [definitions](https://en.wikipedia.org/wiki/Gauge_(instrument))
- `is a device used to make measurements or in order to **display** certain dimensional information`.
- `a device that displays the measurement of a monitored system by the use of a needle or pointer that moves along a calibrated scale`

In the physical world, these gauges always include an actual **display** for immediate use. In our world, that is not exactly how it works. Thus, maybe we are dealing with just the "sensor" part without the display?

Wiki has a list of [sensors](https://en.wikipedia.org/wiki/List_of_sensors). They include terms like *phone, *monitor, *meter, *sensor, *detector, *gauge, *scope, etc...

I'm thinking maybe a sensor is more accurate to what our code does.

specification/metrics/new_api.md
jsuereth: So does "add" make sense in a Gauge setting?   If you're going to be reporting "countable" things where adding works, why not use Counter?

specification/metrics/new_api.md
jsuereth: So, I'm going to invert your naming question and ask:

"Should I be able to determine what metric-type is generated in OTLP based on the instrument?"

From a practical standpoint, we could JUST have value-recorder + value-func and all semantics get tied via other means.    i.e. the actual "instrument" that pulls data is just grabbing a point and pushing it somewhere.    "Histogram", "Distribution", etc. are all semantics around aggregation/meaning of these points.

I agree with you that Histogram/Summary are overloaded terms with too much meaning.


Effectively, in my mind, we have the following:


Instrument = The way users give us data points
Metric (in OTLP) = Aggregation Semantics + Data Point

from my perspective, whatever is chosen here should pass this litmus test;
- Users know (or have control over) what aggregation semantics + metric types are generated from an instrument
- Users aren't confused over behavior of terms we use in the event our usage heavily diverges from common usage
- Users have a clear idea *which instrument to use* when they are looking at adding a new metric.

From that lens, I think taking your 'example metrics' for each instrument and trying to design a decision tree for when to pick one vs. another would be enlightening.   I can help with that if you want.

specification/metrics/new_api.md
```diff
@@ -425,6 +438,78 @@ var obCaesiumOscillates = meter.CreateCounterFunc<UInt64>("caesium_oscillates",
...
+
+TODO
+
+##### Set
```
jsuereth, line 463: nit:  Why not "Record" here (similar to distribution)

specification/metrics/new_api.md
reyang: > So, I'm going to invert your naming question and ask:
> 
> "Should I be able to determine what metric-type is generated in OTLP based on the instrument?"

My answer is "based on the instrument, we'll give the default metric type, and you can change the type if there is a need". This means one instrument might result in three metric streams in OTLP, one as histogram, another as exemplar, and the 3rd as summary.

> from my perspective, whatever is chosen here should pass this litmus test;
> 
> * Users know (or have control over) what aggregation semantics + metric types are generated from an instrument
> * Users aren't confused over behavior of terms we use in the event our usage heavily diverges from common usage
> * Users have a clear idea _which instrument to use_ when they are looking at adding a new metric.
>
> From that lens, I think taking your 'example metrics' for each instrument and trying to design a decision tree for when to pick one vs. another would be enlightening. I can help with that if you want.

Totally agree, this is the "decision tree" in my mind:
1. Decide whether it is sync or async. If sync, go to 2; If async goto 4.
2. Decide if it is monotonically increasing or not, if yes use Counter, if no go to 3.
3. Decide if the intention is to get the last value, or statistics (histogram, summary, etc.). If the goal is to get the last value, use Gauge, otherwise use Distribution.
4. Decide if the value in monotonically increasing or not, if yes use CounterFunc, if not use GaugeFunc.

I'll pick your brain on Slack.

specification/metrics/new_api.md
reyang: Caveat, here the `Add` could take positive and/or negative value, while `Counter.Add` only takes non-negative values (might need to find a better name since it is very confusing).

Stepping back, I think the 1st question is - do we need a function to report delta (whether it is called Add, or Inc/Dec can be determined later)? or we only take the absolute value?

My current thinking is that we need delta, otherwise if users get delta value from the source, they will have to maintain some state and calculate the absolute value (which means they will allocate memory for each combination of attributes, do some interlocked operation, essentially implementing their own aggregation).

The 2nd question is - do we want to provide *both* delta and absolute function? do we want to provide both in the same instrument, or create two separate instruments? Looking at Prometheus client, I tend to have one instrument that provides both absolute and delta value.

specification/metrics/new_api.md
reyang: This is related to [this comment](https://github.com/open-telemetry/opentelemetry-specification/pull/1617#discussion_r613565957).

If we just need one function which takes the absolute value, then `Set` and `Record` are both good to me.
If we need `Add`, having `Set` seems more natural? (e.g. in Prom-client https://prometheus.github.io/client_java/io/prometheus/client/Gauge.html#set-double-)

specification/metrics/new_api.md
```diff
@@ -425,6 +438,78 @@ var obCaesiumOscillates = meter.CreateCounterFunc<UInt64>("caesium_oscillates",
 provided by the `callback`, which is registered during the [CounterFunc
 creation](#counterfunc-creation).
 
+### Gauge
+
+`Gauge` is a synchronous Instrument which can be used for values that go up
+and/or down. It is intended for monitoring things with natural upper bounds.
+
+Note: never gauge something you can count with [Counter](#counter)!
```
Aneurysm9, line 446: Why should a `Gauge` not be used with something that can be counted with a `Counter`?

specification/metrics/new_api.md
Aneurysm9: For synchronous use I think there is definitely value in providing a means to adjust a gauge with value deltas.  In both of the examples given above, outstanding requests and queue length, I could see it being significantly easier to increment/decrement at the start and end of a request or in push()/pop() than to separately keep track of the cumulative value.

The scenarios where `Set` would be valuable with a synchronous instrument are less apparent to me, but I don't think that means it is not worth providing.

jmacd: I am opposed to the `Add()` operation on a Gauge. If the instrument supports `Add()` it should be one of the Counters. Historical uses of `Add()` with Gauges should convert to UpDownCounter (IMO).

specification/metrics/new_api.md
jmacd: `Set()` has a lot of historical precedent, I would say.

specification/metrics/new_api.md
jmacd: The rationale behind my opinion is that in historical systems, both Statsd and Prometheus, to increment or decrement a Gauge has meant to maintain an in-memory copy of the value of the instrument. This forces an implementation to do something stateful, in other words. My goal is that each instrument has a single kind of event that can be aggregated, semantically, into one of the OTLP points.

If Gauge instruments are allowed to support both Set() and Add() APIs, then the observer of the stream has to deal with two kinds of OTLP point, some Gauges and some Counters, at which point the timestamp has to be used to reconstruct the value at a given point in time. There is not an "intrinsic decomposable aggregation function", in other words.

specification/metrics/new_api.md
victlu: Just curious why is Sync vs Async the first decision a user will choose?

Just thinking out loud, the sync/async aspect is dependent on the specific language being used and any applicable programming model currently at play.

Again, just thinking out loud, I would personally decides lossly in this order.
- Do I want a Time Series or a Statistic
- Time Series
  - I have access to absolute and/or delta number values
    - Absolute Value
      - I can provide the # in both sync and/or async depending on when I have enough context.
      - Do I also want to see rate time series? Do I also want statistics?
    - Delta Values
      - I can provide the # in both sync and/or async depending on when I have enough context.
      - Do I also want statistics?
- Statistic
  - Do I want a Summary or hi-def histograms
  - Summary
    - I can provide the # in both sync and/or async depending on when I have enough context.
    - Do I want to highlight any events/points?
  - Histogram/Distribution
    - I can provide the # in both sync and/or async depending on when I have enough context.
    - Do I want to highlight any events/points?

jsuereth: Comments from our slack discussion:

- I think "Monotonically increasing" is a Prometheus concern.   We *can* represent counters as Sums that are not monotonically increasing (or cumulative).  However, i agree we shouldn't confuse users on what the semantics are (e.g. OTel vs. Prometheus vs. Micrometer)
- (as you said) there's a spectrum between "Instrument records raw points, views provide aggregation into metrics" and "every instrument is 1:1 with a metric type" that we're trying to live between.    I'm a fan of pushing the boundary closer to "instrument record raw points" and making sure the "view' (or hint) API is really rich, clear and easy.  In practice, this is likely a matter of taste + guesswork :)
- i'm more concerned about Sync API as Async suffers from two things:
  1. It can't participate in `Context` one of the major "special sauce' of OpenTelemetry API
  2. I think it's likely that Sync API is used *more often* by users for custom metrics above built-in instrumentation (i.e. I think Async API is one we build for our own purposes, e.g. grabbing JVM GC metrics in our instrumentation offerings).
- in my mind "add" relates to DeltaSum / DeltaCounter vs. "set" relates to CumulativeSum / CumulativeCounter.   Is it worth exposing that level of granularity in the instrument name?

specification/metrics/new_api.md
```diff
@@ -34,6 +34,15 @@ Table of Contents
   * [CounterFunc](#counterfunc)
     * [CounterFunc creation](#counterfunc-creation)
     * [CounterFunc operations](#counterfunc-operations)
+  * [Gauge](#gauge)
```
bogdandrutu, line 37: My 2 cents here:
I would avoid mixing increment/decrement operations which make sense only for a "sum", with "Gauge" (set operation). So I would call this an `UpDownCounter` that has only add/dec (or just `add` but allows negative values) operations, but no `set`. Also I've failed to get an answer for a use-case of a `set` operation for a sync instrument, maybe worth documenting the use-case or add a link if I missed that.

In my mind we can have:
1. Counter + CounterFunc (callback based version of the Counter, however we decide to name it) - the result is a monotonic sum
2. UpDownCounter + UpDownCounterFunc (callback based version of the UpDownCounter, however we decide to name it) - the result is a non-monotonic sum
3. Sync instrument for reporting values that "grouping" makes the most sense for them. We can call it `Distribution` (this is a noun which does not fit very well for an Instrument), consider (Grouper, ValueRecorder etc.). For the moment no clear use-case for an async version of this.
4. Helper sync instrument for reporting "durations". Call it `Timer`? For the moment no clear use-case for an async version of this.
5. Async instrument to report "independent" values (e.g. room temperature). Call it `Gauge` (ValueObserver)?

From data-model perspective it is important to distinguish between a non-monotonic sum and a "gauge" (a.k.a value that we don't know how to re-aggregate/merge). The example I had in mind is if users report "cpu usage" as a metric with one label "state" with possible values ["used", "free", "inactive"] and we want to compute a new "memory usage" with no labels, excluding "state=free" we will be able to do that because we know summing the value from state=used + state=inactive is correct, if this was a gauge we cannot do that automatically, we need to ask the user what aggregation to apply when merging points.

I do understand that this requirement leaks some internal details into the API, but without user telling us via a "contract" that this represents a non-monotonic sum, we cannot infer that information.

specification/metrics/new_api.md
```diff
@@ -425,6 +438,78 @@ var obCaesiumOscillates = meter.CreateCounterFunc<UInt64>("caesium_oscillates",
...
+
+`Gauge` is a synchronous Instrument which can be used for values that go up
+and/or down. It is intended for monitoring things with natural upper bounds.
+
+Note: never gauge something you can count with [Counter](#counter)!
+
+Example uses for `Gauge`:
+
+* the number of active requests
+* the number of items in a queue
```
aabmass, line 451: Just throwing something out there – all the use cases for **synchronous** Gauge/UpDownCounter I've seen involve increment/decrement events to keep a single count. You could achieve the same thing with two Counters, one for increments and one for decrements, and the total value is `increments - decrements`. In which case, do we need the Gauge/UpDownCounter instrument at all?

A pro of two monotonic counters is you can see what happens between between aggregation intervals better, for example
- At time t<sub>2</sub>, the queue size is 5
- Between t<sub>1</sub> and t<sub>2</sub>, there were 10000 increments and 9995 decrements

I saw noticed this originally in https://github.com/open-telemetry/opentelemetry-specification/pull/1076#discussion_r508118388 for database semantic conventions, where Justin proposed `db.connections.taken` and `db.connections.returned` for keeping count of redis connection pool usage.

> The more detailed metrics allow you to see things like connection churn. With them, a user could differentiate between an application making millions of tiny database operations (and ending a harvest period with 4/5 connections in use) and an application with four very long and slow queries in progress (also ending the harvest period with 4 connections in use).

specification/metrics/new_api.md
reyang: Based on the feedback I collected from the SIG meeting, comments, Slack discussions, I've made the changes with these principles in mind:
1. For async instruments, we only support "Report(absolute value)" rather than supporting delta value. The main considerations are - 1) it is uncommon to have async callback to report delta value 2) if a callback is registered to multiple places, the delta value would by super messy
2. For sync instruments, we never mix absolute and delta in a single instrument - a sync instrument can only take one approach (e.g. distribution takes the absolute value, counter takes the delta).

specification/metrics/new_api.md
reyang: @jmacd @jsuereth after struggling with the other instruments and their operations, now I updated the PR with the hope that we use *one consistent* operation name for all absolute values - anything that takes an absolute value will be called "Report" (or "Record" - please let me know which one you prefer, I personally prefer "Report" but don't have strong opinion).

specification/metrics/new_api.md
reyang: Thanks for calling it out.
The initial thinking was to mimic Prometheus by not having separate instruments for Gauge (e.g. CPU fan speed - taking the speed from multiple CPU fans and sum them up makes no sense) and UpDownCounter (e.g. heap size - taking the heap size from multiple processes and sum them up makes sense).
I've updated the PR to split the concepts based on another [comment thread](https://github.com/open-telemetry/opentelemetry-specification/pull/1617#discussion_r614486788), please take another look.

specification/metrics/new_api.md
reyang: I've addressed most parts of this comment, the only thing remaining (I believe) is the Timer.
I think Timer is not a blocker, it is more like a "nice to have" layer on top of distribution.

The PR is updated, please take another look.

specification/metrics/new_api.md
reyang: If anyone wants to explore the "sensor" name please comment here.

I personally think "sensor" is a very broadly used term, and it is not found from the famous metrics systems, so I would avoid introducing it to OpenTelemetry.

specification/metrics/new_api.md
reyang: > IMO Histogram/Distribution/Summary don't make good instrument names, those all sound like a kind of aggregation to me.

Completely agree, this is the name that we (folks in the metrics API SIG) struggle the most.

> The OTel API doesn't know anything about aggregations, the SDK is responsible for that.

I do think the OTel metrics API has a goal to understand the *default handling* and *default aggregation* for the most common scenarios. The SDK has a goal to *take over the control* but we don't expect folks use the SDK to provide all configurations/hints in order to make things work. Please refer to the ["developer X requirements" in OTEP146](https://github.com/open-telemetry/oteps/blob/main/text/metrics/0146-metrics-prototype-scenarios.md#scenario-2-http-server).

jkwatson:
Just my opinion, but rather than `*Func` as the "name" of the async version of an instrument, I wonder if we can just call them "Asynchronous *". So, rather than "CounterFunc", we can just say "Asynchronous Counter".

specification/metrics/new_api.md
```diff
@@ -397,9 +413,9 @@ page_faults_counter_func = meter.create_counter_func(name="PF", description="pro
 
 def pf_callback(result):
     # Note: in the real world these would be retrieved from the operating system
-    result.Observe(8,        ("pid", 0),   ("bitness", 64))
-    result.Observe(37741921, ("pid", 4),   ("bitness", 64))
-    result.Observe(10465,    ("pid", 880), ("bitness", 32))
+    result.Report(8,        ("pid", 0),   ("bitness", 64))
+    result.Report(37741921, ("pid", 4),   ("bitness", 64))
+    result.Report(10465,    ("pid", 880), ("bitness", 32))
```
bogdandrutu, line 418: If you are planning to use `Report` as the action on the distribution I would stick with `Observe` here to ensure we don't create confusions.

bogdandrutu:
@jkwatson I think there is an open item for discussion about renaming the `Func`. I would say that it is separate item, but I do like your proposal.

specification/metrics/new_api.md
bogdandrutu: If you are planning to use `Record` as the action on the distribution I would try to avoid using the same one for async instruments to avoid confusion.

reyang:
> Just my opinion, but rather than `*Func` as the "name" of the async version of an instrument, I wonder if we can just call them "Asynchronous *". So, rather than "CounterFunc", we can just say "Asynchronous Counter".

I'll send separate PRs about this based on the options we've discussed yesterday.

specification/metrics/new_api.md
reyang: How does this sound?
* Use "Observe(absolute value)" for async instruments.
* Use "Report(absolute value)" for Distribution (and Timer, if we introduce it later).

specification/metrics/new_api.md
```diff
@@ -425,6 +441,104 @@ var obCaesiumOscillates = meter.CreateCounterFunc<UInt64>("caesium_oscillates",
...
+
+#### Distribution operations
+
+##### Report
+
+TODO
+
+### GaugeFunc
+
+`GaugeFunc` is an asynchronous Instrument which reports non-additive value(s)
```
victlu, line 467: Why is a Gauge not needed?

`Gauge` is a synchronous Instrument which reports non-additive values(s)

The example would be to report the room temperature when I finished an expensive operation. I have "context" as I'm associating my operation to the room temperature. If I only used GaugeFunc with a callback, I cannot provide the correct timing context as well as operation context.

reyang: Would you explain why the room temperature is associated with operation context? Or consider finding another scenario.

victlu: What about changing room temperature to CPU temperature. I want to record the CPU temperature after I run a set of benchmark tests.

bogdandrutu: do you? I think this is kind of an artificial use-case. If you really want that you can save the value and return it in the func for the moment.

Also probably at that point you want the distribution of temperatures after multiple runs.

specification/metrics/new_api.md
jonatan-ivanov: To me, a `Gauge` is naturally async and it has a `set` method. A synchronous scenario for a `Gauge` could be having a synchronous process and recording the last processed `something` in a gauge.

e.g.:
```java
public updateCar(long id) {
    carRepository.update(id, ...);
    lastUpddatedCar.set(id);
}
```

In micrometer, a `Gauge` is always async but it can give you the backing state object that you can update:

```java
AtomicLong lastUpdatedCar = registry.gauge("numberGauge", new AtomicLong(-1));
lastUpddatedCar.set(id);
```

jonatan-ivanov: Could somebody please help me to understand the nature of `UpDownCounter`? I suppose it will not publish a rate but it will publish the value just like a `Gauge` would, right?

If so, is it really needed?
There are two examples in the docs: "the number of active requests" and "the number of items in a queue". To me, both seem like a "`Gauge`-scenario", more than that the first one is an example for synchronously updating a `Gauge` in synchronous environments.

Let me give you an example for both:

```java
AtomicInteger activeRequests = registry.gauge("numberGauge", new AtomicInteger(0));

public Response process(Request request) {
    activeRequests.increment();
    ...
    return response;
}
```
and

```
Queue<String> queue = new BlockingQueue();
...
Gauge.builder("queueSize", queue, Queue::size).register(registry);
```
or
```
Queue<String> queue = registry.gauge("queueSize", tags, new BlockingQueue<>(), Queue::size);
```

My ultimate question is, do we need all of these? Can't we just get along with a single Gauge like the above that can cover all of these scenarios?

specification/metrics/new_api.md
reyang: I've updated the PR description and explained the "decision tree", please take a look.

specification/metrics/new_api.md
reyang: I discussed with @jonatan-ivanov offline, here goes the summary why we need `UpDownCounter` (the name can be changed, but we do need such thing as a separate instrument):
1. We need a sync instrument (which has access to the context/baggage) which could take delta value AND is not monotonic. We cannot squeeze this into an async instrument since we do not want to have any delta value in async instrument (we want async instrument to take absolute value instead).
2. This instrument by nature (since it takes delta) would allow values to be added across dimensions (take the heap size of each worker process id, you will get the total heap size of all worker processes).
3. This instrument does not fit Counter or Distribution.
4. Calling it Gauge is probably a bad idea, since Gauge doesn't imply additive property.

@aabmass brought up and idea that we might be able to achieve this by having two Counters https://github.com/open-telemetry/opentelemetry-specification/pull/1617#discussion_r615003910. I want to see more feedback to his comment.

specification/metrics/new_api.md
reyang: Let me revert this change, so we can focus on the names of instrument (rather than the names of the operations) in this PR.

I'll send a separate PR to address the operation and names (currently many are left as TODOs).

reyang: Reverted the change here.

specification/metrics/new_api.md
victlu: I'm struggling to understand why this scenario and/or pattern is artificial or unwanted. The pattern involves doing an operation (where I have my context) and I want to report a value (that is not additive).

i.e.
- Report my CPU Temperature (value: Temp) after a benchmark test (context: test_name)
- After processing an incoming request (context: Type of request) I want to report a status code (value: 1=success/2=failure/#=etc.)

Are we saying I should use the sync Distribution instrument for these?

> If you really want that you can save the value and return it in the func for the moment.

This is true. However, current design requires the callback to report for same time. This presents a problem trying to report multiple operations from one (async) collection period.

specification/metrics/new_api.md
```diff
@@ -425,6 +441,104 @@ var obCaesiumOscillates = meter.CreateCounterFunc<UInt64>("caesium_oscillates",
...
+`GaugeFunc` is only intended for asynchronous scenario. The only operation is
+provided by the `callback`, which is registered during the [GaugeFunc
+creation](#gaugefunc-creation).
+
+### UpDownCounter
+
+`UpDownCounter` is a synchronous Instrument which supports increments and
+decrements.
+
+Note: if the value grows
```
jsuereth, line 496: +1 to all these call-outs when to use other instruments.

specification/metrics/new_api.md
jsuereth: I realize the content of the "which instrument to pick" is effectively in the document, but i'd love if we could move it into one location.  In this PR or another is fine, but IMO the "I want to measure something, what do I pick" is the most important user-story to hit first.  Outlining these instruments, in the vein of "do we have the right set of instruments" is answered via that question.

Also, I'd love to link directly to that section from the DataModel specification.

jsuereth (approved):
Approve with two caveats:

- I don't like Histogram of "Func" names, but don't have an alternative to propose you haven't already considered *or* that doesn't violate some other principle for the API.
- I'd like the decision tree we discussed (and user story of "What instrument to pick") to be highlighted in one spot

specification/metrics/new_api.md
reyang: I will create a separate PR to update the README file, tracked by #1632.

specification/metrics/new_api.md
victlu: I like the double accounting! Maybe we can incorporate this into one instrument? The add() always monotonically add to one internal state, while the dec() always monotonically add to a second internal state.

specification/metrics/new_api.md
Aneurysm9: > IMO Histogram/Distribution/Summary don't make good instrument names, those all sound like a kind of aggregation to me.

I would second this as well as @jsuereth's point that all we really need is a `ValueRecorder` and everything else is syntactic sugar, but I understand the desire to have instruments that encapsulate some of the aggregation semantics to use where it makes sense.  

I wonder whether we can/should lean into that more and say that `ValueRecorder` is the fundamental instrument type and it has a default aggregation that produces a distribution/summary/histogram.  All other instrument types are shorthand for different bundles of default aggregation and interfaces.  A `Gauge` is a `ValueRecorder` that must have a `LastValue` aggregator and therefore won't expose increment/decrement/record but only set operations.  An `UpDownCounter` is a `ValueRecorder` that has a `Sum` aggregator attached and therefore doesn't expose a set operation, only increment/decrement/record.

specification/metrics/new_api.md
```diff
@@ -425,6 +441,104 @@ var obCaesiumOscillates = meter.CreateCounterFunc<UInt64>("caesium_oscillates",
...
+[monotonically](https://wikipedia.org/wiki/Monotonic_function), use
+[Counter](#counter) instead.
+
+Example uses for `UpDownCounter`:
+
+* the number of active requests
+* the number of items in a queue
```
noahfalk, line 503: As an example, lets say I have an app runs on a bunch of machines and each machine has a bunch of queues with incoming customer traffic (I'm guessing this is a common situation). I know that if any of these queues gets particularly large that is a bad sign, implying high latency and maybe failure in the near future because my service is receiving requests faster than it can process them. In my app I create an UpDownCounter and use the queue name as a label. Then in my monitoring dashboard I want to set up alerts and graphs that show me the maximum queue size (or maybe the 99 percentile of queue size) across my fleet. Can I do that or should I have chosen a different instrument for this use case?

I think the underlying question may go more to the data model, but without answering it I don't know what to expect from the API. The data model refers to data streams having aggregate functions that work across both temporal and spatial dimensions. Does that mean the aggregation is the same no matter which dimension I am aggregating across? In this example I picked UpDownCounter because it is summing over time but across the queue dimension I don't want a sum, I want a maximum or a percentile. It wasn't clear if there was any way to express this. 

I've probably got various follow up questions, but I need to figure this one out first : )

specification/metrics/new_api.md
noahfalk: > * I'd like the decision tree we discussed (and user story of "What instrument to pick") to be highlighted in one spot

Can I propose we go further and attempt to write out a proposal for what the documentation text on the OTel website will be? Hopefully it should feel like the instruments and names we are choosing make the story we tell in the documentation concise, consistent and easily understood. If we can't make it sound that way I'd take it as a warning sign. I'm glad to take the first stab at this or collaborate with someone else.

noahfalk:
> [@bogdandrutu] consider Grouper, ValueRecorder

I'll admit initially I wasn't a fan of ValueRecorder but it has grown on me while "Distribution" feels worse over time (and Reiley using 'Distribution' in the first place may be my fault, I used that name in the .NET metric prototype code). Our other names fit the naming pattern of being an instrument (a Counter counts, a Gauge gauges) whereas "Distribution" doesn't describe an instrument. 

I agree that ValueRecorder is quite generic, but framed correctly that might be exactly what we need to tell a good story. For example the docs might say something like:

> 1. If you are timing something, use a Timer
> 2. If you are counting something that only goes up, use a Counter
> 3. If you are counting something that could go up or down, use UpDownCounter
> 4. If you have a value that is fluctuating outside your control but you can read
>     the current value on demand, use a Gauge
> 5. For anything else, use ValueRecorder. This instrument is intentionally generic so that
>    it can be flexibly used for many different scenarios. You can optionally configure a default
>    aggregation and end-users doing monitoring can always specify an alternative way
>    of analyzing the recorded numbers if they need it.

specification/metrics/new_api.md
reyang: Here is my take:
1. `UpDownCounter` is the proper instrument since "number of items" by nature allows Sum. The "number of items" can be added across temporal and spatial dimensions. Although depending on the scenario one might prefer other aggregations (e.g. "the total number of items in all queues" might make more sense if the queues are homogeneous, and will make less sense if each queue is serving a totally different purpose).
2. My answer would be yes - one should be able to customize the consumption experience and do alerting based on maximum queue size (or 99 percentile queue size) without having to use a different instrument.

noahfalk: Thanks Reiley! This leads me to a few follow ups:

1. My understanding of the data model spec is that there is a single aggregation function for data streams across spatial and time dimensions whereas your answer implies there can be different aggregation functions for different dimensions. Am I misreading the spec or if not how would this configuration work?
2. I would guess that it is rare to find a scenario where summing is the desirable spatial re-aggregation function for every label dimension. For example summing over a machine name label or a process instance label implies all the alert levels would have to be reconfigured whenever the number of machines/processes in the distributed system changes. Does that match your expectations or am I not considering the right scenarios?
3.  Following (2), if the default configuration for an UpDownCounter is that it always sums when aggregated over any label and nearly all scenarios have labels that should not be summed over, this implies nearly everyone will need to change the configuration away from the default for their UpDownCounters. If GaugeFunc and UpDownCounterFunc only differ based on their default spatial aggregation behavior and most use-cases won't be able to use those defaults then for practical purposes is it fair to say that UpDownCounterFunc and GaugeFunc don't have any differences that are likely to impact the typical scenarios that would make use of them? (I'm not implying that redundancy automatically needs to be eliminated, I just want to understand what are the practical differences, if any)

specification/metrics/new_api.md
noahfalk: My two cents : )
From the point of view of the ops team or the developer tasked with diagnosing an issue, this two counter view seems like it would be helpful sometimes. If I was setting monitoring alerts I'd probably still want to set my alert based on the size of the queue, but then when I was investigating an issue or just trying understand some general behavioral characteristic of the app I might want to know separately how many queue size changes happened or what was the distribution of the size deltas.
From the point of view of the developer whose job it is to write the instrumentation, I'd like to write the minimal amount of code that will be able to flexibly handle different monitoring needs, hopefully without me needing to predict all the use cases in advance.

Putting those two perspectives together the design it leads me to would be:
1. The developer writing the code uses a single UpDownCounter instrument in their code
2. Either by default configuration or more likely by explicit configuration in a given monitoring environment it should be possible to derive and report the values of those two independent time series even though the developer did not explicitly create two instruments in the code.

I think this may be the same suggestion you were making @victlu?

specification/metrics/new_api.md
```diff
@@ -425,6 +441,104 @@ var obCaesiumOscillates = meter.CreateCounterFunc<UInt64>("caesium_oscillates",
...
+
+#### UpDownCounter creation
+
+TODO
+
+#### UpDownCounter operations
+
+##### Add
```
noahfalk, line 511: If I want to go downwards do I call Add() with a negative value? That sounds fine to me but just confirming.

specification/metrics/new_api.md
jmacd: I looked up "instrument" and one of the definitions is "measuring device used to gauge". I've come to terms with "gauge" in the sense that was mentioned in last week's SIG meeting--it's an instrument of last resort, when either a Histogram or Counter are for some reason not appropriate, meaning for cases where you do not know the aggregation being applied.

"gauge" is a great term for cases where there is not a more specific instrument, and it's a great term for when there's not a better term. "Gauge" has several meanings in English--inspecific. Gauge instruments have many kinds of aggregation--inspecific.

specification/metrics/new_api.md
jmacd: > one consistent operation name for all absolute values

I like this! So, something like:

Counter: Add()
UpDownCounter: Add()

Gauge: Record()
Histogram: Record()

ObservableCounter: Observe()
ObservableUpDownCounter: Observe()
ObservableGauge: Observe()

specification/metrics/new_api.md
jmacd: @reyang this is :100:

specification/metrics/new_api.md
jmacd: Here's another thread on this double-counting idea.

https://github.com/open-telemetry/opentelemetry-specification/pull/1618#discussion_r616230383

I think of `UpDownCounter` as an instrument that could be semantically configured to output two monotonics. I proposed a special attribute label `positivity` but I'm sure we could find something better.

specification/metrics/new_api.md
jmacd: I will admit that I have trouble finding many real-life examples of a synchronous gauge. The most common use of the term, I think comes from reading the speedometer of a car. This reading is instantaneous, you didn't make any request to your dashboard.

For real-life examples, I'm thinking about reading an electrical meter. You are an energy company and need to send people or RPCs around to each meter to get a reading of the current usage. As this request makes it's way into the meter, it has context: which part of the grid it's on, where's the substation, who's taking the measurement, and so on. When it reaches the meter, it might see Voltage (a true Gauge) and kWh used (UpDownSum) and want to Record them.

I've introduced a conflict--we're talking about adding a synchronous Gauge (to set Voltage) but we have not been not talking about adding a synchronous UpDownSum instrument, for synchronously making cumulative observations (discussed as long ago as https://github.com/open-telemetry/oteps/pull/88), and I hope that we don't.

So why is it important to offer synchronous true Gauge? Maybe because of historical precedent. However, it OTLP is going to embrace its Non-Monotonic Cumulative Sum point, which is often confused with Gauge, then we better have a good answer as to why we're not introducing a Synchronous Non-Monotonic Cumulative Sum instrument. This would be the instrument used by the meter reader at my house, since I have solar panels. 😀

specification/metrics/new_api.md
jmacd: There can be performance implications of using `UpDownCounter` in the way described here. Although we think of `UpDownCounter` as a way to encode non-monotonic deltas semantically, they are almost always used with in-process aggregation, thus exported in cumulative form.

The consumer of a stream of non-monotonic deltas is able to calculate a smoothed rate over windows of time, but has to start reading and keeping state itself from beginning of time in order to extract actual queue sizes in these examples.

The reason users will sometimes prefer `UpDownCounter` to its asynchronous form is that the Metric SDK can be configured to compute virtual queue sizes along multiple sets of dimensions, whereas an asynchronous callback will likely only be able to report on actual queue sizes.

I occasionally talk about a "Stateless" exporter configuration for OTLP exporters and OTel SDKs that outputs deltas instead of cumulative sums, thus allowing the SDK to release memory. It makes sense to have a stateless option for all the instruments _except_ `UpDownCounter` because of the preceding argument--they're almost always useful aggregated from the beginning of time, and it SHOULD be the caller's responsibility if possible to maintain that state. Still, I'd like it to be possible for a stateless SDK to output `UpDownCounter` aggregations with delta temporality, provided the consumer can maintain lifetime state on behalf of the process.

jmacd:
I'm not sure I like `Distribution`. I would prefer `Histogram`.

I believe our goal is to choose instrument names that sound like instruments. This may be a stretch, but I want to write that "a Histogram is a statistical instrument for measuring a distribution".

specification/metrics/new_api.md
noahfalk: @jmacd - Thanks! Your answer makes me wonder a few things:

1. Your answer felt detailed, but our current guidance/decision tree does not contain that level of detail. Are you worried that users applying a simple decision tree such as "Queue size is good to monitor with UpDownCounter" will have a bad outcome or you think the they will still be reasonably served by the simple advice?

2.  Thinking more about the requirement for the SDK to maintain state, this worries me about recommending people to use UpDownCounter in scenarios like this. I know that the common/simple scenario for monitoring involves enabling the SDK at process startup and having it track state for the lifetime of the service but if possible I'd like the API to support other scenarios such as:
  - Ops personel use a monitoring dashboard to reconfigure collection policies dynamically. New configuration is transmitted over the network to a monitored app which applies it without restarting.
  - Ad-hoc tools similar to perfmon on Windows or top on Linux want to dynamically start listening to a metric for a period of time, then stop.

These scenarios are places where ideally the SDK does not pay the cost of maintaining the state up-front, but it can enabled on demand. However since it is impossible to recover the cummulative state of an UpDownCounter the SDK is forced to either track everything just in case or fail any request to start tracking on demand. However if the developer had used an instrument where they provided the size of the queue directly rather than its delta (sync sum, sync Gauge, async sum, async Gauge) there would be no problem enabling these metrics on demand. As the developer I'd expect it is just as easy for me to record total queue size as it is to record the delta in queue size so if one of these options gives me more capabilities than the other then picking UpDownCounter sounds like it would be a worse choice and I'll never want to pick it.

(Also I still think the follow up questions in my 2nd post on this thread remain unanswered, apologies if I am being dense and this was intended to answer them as apparently I didn't understand that connection)

noahfalk:
> I'm not sure I like `Distribution`. I would prefer `Histogram`.
> 
> I believe our goal is to choose instrument names that sound like instruments. This may be a stretch, but I want to write that "a Histogram is a statistical instrument for measuring a distribution".

IMO, it is a stretch : ) If I read that sentence in the OTel docs this is what goes through my head: "I've never thought of Histogram this way before and it doesn't feel like an intuitive definition, but I guess the OTel folks are making their own novel re-definition of the term and I'll do my best to roll with it"

specification/metrics/new_api.md
jsuereth: +1 I like the consistent naming on the method.

specification/metrics/new_api.md
reyang: Yes. I'll add explanation to this section in a separate PR.

reyang:
We've discussed this PR during the [04/22/2021 Metrics API/SDK Spec SIG Meeting](https://docs.google.com/document/d/1-bCYkN-DWJq4jw1ybaDZYYmx-WAe6HnwfWbkm8d57v8/edit#heading=h.nq6d5s1aqo2h).

Here goes the conclusion:
1. The only blocking thing is the name of "Distribution" instrument, we've collected feedback during the meeting and "Histogram" turned out to be a clear winner. As a result I'm going to update this PR to use "Histogram" as the instrument name.
2. There are other comments which are not blocking this PR, we'll follow up in separate PRs.

specification/metrics/new_api.md
reyang: I've updated the PR based on the discussion during the SIG meeting.

For "ObservableCounter", I'll address that in a follow up PR based on the [comment](https://github.com/open-telemetry/opentelemetry-specification/pull/1617#issuecomment-821454608) and what we've concluded during the SIG meeting.

**We use “Asynchronous Counter” in the spec and encourage people to use “ObservableCounter” if they need a type name (and call out that it is highly discouraged to use a different name unless there is a strong reason).**

specification/metrics/new_api.md
reyang: This has been discussed during the [04/22 SIG meeting](https://docs.google.com/document/d/1-bCYkN-DWJq4jw1ybaDZYYmx-WAe6HnwfWbkm8d57v8/edit#) and the conclusion is "go ahead with the current approach (UpDownCounter)". Here goes the main takeaways:

* reyang: worried about different configurations (e.g. dimensions) on two instruments
* jmacd: joining two things could be hard, which seems to be against the API goal (which is to make things simple for the user)


We might want to explore the idea brought by @jmacd in [this comment](https://github.com/open-telemetry/opentelemetry-specification/pull/1617#discussion_r618125415).

jmacd (approved):

SergeyKanzhelev (approved):
with the @reyang comment lgtm

SergeyKanzhelev:
Policy https://github.com/open-telemetry/opentelemetry-specification/blob/main/CONTRIBUTING.md#how-to-get-your-pr-merged is satisfied except the time from the last significant change. I will merge tomorrow evening PT if no objections will be posted.

noahfalk:
I'm OK with the PR, but I do still have significant concerns that UpDownCounter's requirements to maintain state forever make it incompatible with scenarios where the monitoring user wants to capture data ad-hoc or do dynamic reconfiguration. If nothing changed on that aspect of the design (or in my understanding of the issue) my guess is that .NET would implement the instrument but never recommend using it.

aabmass (approved):

PR #1618, opened 2021-04-15
Add Sum description in data model specification

### LINKED ISSUE #1566: How to rebuild from deltas?
An earlier discussion about rebuilding from deltas raised a question: https://github.com/open-telemetry/wg-prometheus/pull/25#discussion_r595010522.

Currently, the delta data points don't have sequence numbers. There is no way to identify duplicates or missing data points. Generally speaking, we don't know how to rebuild from deltas without sequence numbers. The spec/data model should address this issue.

cc @RichiH @jmacd

jsuereth:
Fixes #1566

## Changes

- Adds specification details around "Sum" and the semantics encoding in OTLP for sum
- Clarifies "OTLP => Timeseries" mapping for Sum, Gauge + Histogram
- Make sure "Summary" is described as legacy support
- Outline algorithm for Delta=>Cumulative transformation, including limitations for accuracy.


specification/metrics/datamodel.md
```diff
@@ -199,28 +199,88 @@ further development of the correspondence between these models.
...
+translated into a Timeseries Counter, when the sum is monotonic, or
+a Gauge when the sum is not monotonic.
 
-The two points that become Gauges in the Timeseries model are distinguished by
-their built in aggregate function, meaning they define re-aggregation
-differently. Sum points combine using addition, while Gauge points combine into
-histograms.
+![Stream → Timeseries](img/model-layers-stream.png)
+
+Speciffically, in OpenTelemetetry Sums always have an aggregate fungation where
```
sirianni, line 221: typos
```suggestion
Specifically, in OpenTelemetry Sums always have an aggregate function where
```

specification/metrics/datamodel.md
```diff
@@ -199,28 +199,88 @@ further development of the correspondence between these models.
...
+### Histogram
+
+Pending
+
+### Summary (Legacy)
+
+[Summary](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L244)
+metric data points convey quantile summaries, e.g. What is the 99-th percentile
+latency of my HTTP server.  Unlike other point types in OpenTelemetry, Summary
+points cannot be merged in a meaningful way.   This point type is not
```
jack-berg, line 281: What do you mean by "merged in a meaningful way"?

If I have summary points for 1 minute windows, I can merge them and produce summary points for 5 minute windows.

specification/metrics/datamodel.md
```diff
@@ -199,28 +199,88 @@ further development of the correspondence between these models.
...
+  - A time window (start, end) time for which the Sum was calculated.
+
+The aggregation temporality is used to understand the context in which the sum
+was calculated. When the aggregation temporality is "delta", we expect to have
+no overlap in time windows for metric streams, e.g.
+
+![Delta Sum](img/model-delta-sum.png)
+
+Contrast with cumulative aggregation temporality where we expect to report the
+full sum since 'start' (where usually start means a process/application start):
```
punya, line 254: Does OTLP include a way to report the start time for a cumulative metric?

specification/metrics/datamodel.md
jsuereth: this specifically related to percentiles iIUC.  The verbage is copied from the existing proto: https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L238

I missed the word "always" here, will fix.  Would you like a description of scenarios where quantiles cannot be merged?

specification/metrics/datamodel.md
```diff
@@ -242,6 +302,66 @@ Pending
...
+## Stream Manipulations
+
+Pending introduction.
+
+### Sums: Delta-to-Cumulative
+
+While OpenTelemetry (and some metric backends) allows both Delta and Cumulative
+sums to be reported, the timeseries model we target does not support delta
+counters.  To this end, converting from delta to cumulative needs to be defined
+so that backends can use this mechanism.
```
punya, line 314: Is it worth pointing out that there isn't a strict definition of correctness for the conversion from delta to cumulative, and that the algorithm outlined below is informational rather than normative?

specification/metrics/datamodel.md
jsuereth: Yes.   All Sum data points have start + end times.

specification/metrics/datamodel.md
```diff
@@ -199,28 +199,88 @@ further development of the correspondence between these models.
...
+
+### Sums
+
+[Sum](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L202)s
+in OTLP consist of the following:
+
+- An *Aggregation Temporality* of delta or cumulative.
+- A flag denoting whether the Sum is
+  [monotonic](https://en.wikipedia.org/wiki/Monotonic_function). In the case of
+  metrics, this usually means the sum is always increasing.
```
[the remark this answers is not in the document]
jsuereth: Nominally, yes.  We currently do NOT limit OTLP sums to be non-negative.

there's the classic "how do we calculate rate of people entering a building by watching folks enter and leave".  You have one monotonic sum for people leaving and one for people entering and you take their difference (or sum them if "people leaving" is monotonically negative).

Practically, I don't think ANYONE uses the 'negative number' part of the monotonic definition.

specification/metrics/datamodel.md
```diff
@@ -199,28 +199,88 @@ further development of the correspondence between these models.
...
+- A set of data points, each containing:
+  - A different set of Attribute name-value pairs.
+  - A time window (start, end) time for which the Sum was calculated.
```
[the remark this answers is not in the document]
jsuereth: Yes, let me copy more out of the proto description here.

specification/metrics/datamodel.md
```diff
@@ -199,28 +199,88 @@ further development of the correspondence between these models.
...
+![Stream → Timeseries](img/model-layers-stream.png)
+
+Specifically, in OpenTelemetry Sums always have an aggregate function where
+you can combine via addition. So, for non-monotonic sums in OpenTelemetry we
+can aggregate (naturally) via addition.  In the timeseries model, you cannot
+assume that any particular Gauge is a sum, so the default aggregation would not
+be addition.
+
+In addition to the core point kinds used in OTLP, there are also data types
+designed for conmpatibility with existing metric formats.
```
reyang, line 228:
```suggestion
designed for compatibility with existing metric formats.
```

specification/metrics/datamodel.md
jack-berg: > Would you like a description of scenarios where quantiles cannot be merged?
I'm not an approver or anything - was just scrolling through this PR because I'm following metrics API developments, and was just surprised to see summary metrics marked as legacy.

specification/metrics/datamodel.md
```diff
@@ -358,6 +420,66 @@ Pending
...
+
+#### Sums: detecting alignment issues
+
+When the next delta sum reported for a given metric stream does not align with
+where we expect it, one of several things could have occurred:
+
+- the process reporting metrics was rebooted, leading to a new reporting
+  interval for the metric.
+- A bug or misconfiguration leads to multiple processes reporting the same
+  metric stream rather than uniquely identifying via resource and attributes.
```
jmacd, line 469: May we / can we call this out as a violation of the single-writer principle?

specification/metrics/datamodel.md
```diff
@@ -358,6 +420,66 @@ Pending
...
+    - Add the current value to the cumulative counter
+    - Output a new cumulative point with the original start time and current
+      last seen time and count.
+  - if the next point does NOT align with the expected next-time window, then
+    reset the counter following the same steps performed as if the current point
+    was the first point seen.
+
+For comparison, see the simple logic used in
+[statsd sums](https://github.com/statsd/statsd/blob/master/stats.js#L281)
+where all points are added, and lost points are ignored.
```
jmacd, line 459: I'd remove the reference to statds because is does not use timestamps. We could expand (in another PR) with an algorithm for converting un-timestamped deltas in which we _read-modify-write_ the last-seen time at each point in the sequence.

While this certainly works, no one should do it. The algorithm should accumulate over short intervals, more likely.

specification/metrics/datamodel.md
```diff
@@ -358,6 +420,66 @@ Pending
...
+  metric stream rather than uniquely identifying via resource and attributes.
+- There was a lost data point, or dropped information.
+
+In all of these scenarios we do our best to give any cumulative metric knowledge
+that some data was lost, and reset the counter.
+
+We detect alignment via two mechanisms:
+
+- If the incoming delta time interval has significant overlap with the previous
+  time interval, we cannot accurately combine its sum and reset the cumulative
```
jmacd, line 478: Can this be restated as "we must assume a violation of the single-writer principle"?

jmacd (approved):

specification/metrics/datamodel.md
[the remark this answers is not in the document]
jmacd: Above, where I wrote "not less than the previous value.", I think I mean "not less than the previous timestamped value in the stream, as written by a the single writer".

specification/metrics/datamodel.md
jsuereth: done.

specification/metrics/datamodel.md
jsuereth: True.  I am trying to call out to other implementations and expectations.  SPECIFICALLY i think it's important to note that we HAVE timestamps.  Let me update the verbage and ptal.

specification/metrics/datamodel.md
jsuereth: done

specification/metrics/datamodel.md
jsuereth: done.

specification/metrics/datamodel.md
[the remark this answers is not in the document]
jsuereth: Fixed.

reyang (approved):
LGTM.

specification/metrics/datamodel.md
```diff
@@ -216,29 +216,96 @@ __Note: The same `Resource`, `name` and `Attribute`s but differing point kind
...
+  - For delta monotonic sums, this means the reader should expect non-negative
+    values.
+  - For cumulative monotonic sums, this means the reader should expect values
+    that are not less than the previous value.
+- A set of data points, each containing:
+  - An independent set of Attribute name-value pairs.
+  - A time window (of `(start, end]`) time for which the Sum was calculated.
+    - The time interval is inclusive of the end time.
+    - Times are specified in Value is UNIX Epoch time in nanoseconds since
+      `00:00:00 UTC on 1 January`
```
victlu, line 270: `00:00:00 UTC on 1 January 1970`

specification/metrics/datamodel.md
```diff
@@ -358,6 +425,78 @@ Pending
...
+    - Output a new cumulative point with the original start time and current
+      last seen time and count.
+  - if the next point does NOT align with the expected next-time window, then
+    reset the counter following the same steps performed as if the current point
+    was the first point seen.
+
+#### Sums: detecting alignment issues
+
+When the next delta sum reported for a given metric stream does not align with
+where we expect it, one of several things could have occurred:
```
victlu, line 468: Do we account for `out-of-order` delta points?

If we have P0 and P1 which are aligned and sent to the same destination. Due to potential routing delays, we may get P1 and then P0. Hopefully, we have some buffering to allow P0 to arrive late. Otherwise, we will treat P0 as a drop (cause it's late), or a reset (what happens in this case?), or cause a future P2 to be treated as a restart.

victlu: As a follow up question... If a producer was to generate P0, P1, P2, P3, etc... and sent to one destination. But some delay occurred and we get P1, P2, **P0**, P3... How is this handled?

jmacd:
@open-telemetry/specs-metrics-approvers please take a look.

specification/metrics/datamodel.md
jsuereth: I don't explicitly call it out, but here's my thoughts:

- This scenario is highly unlikely for "common" usage of metrics, specifically if we collect metrics every minute it means there'd have to be an incredibly long delay in export to get out of order points.  Not impossible, but not a very high probability.
- When this scenario inevitably happens, we have already reset the cumulative counter when we received the newer point prior to the previous.  This means we ALREADY have published a more up-to-date cumulative sum that resets the counter.   

A few thoughts and AIs:

- [ ] Prometheus (and other metric backends) doesn't allow out-of-order timestamps to be written.  It's likely worth specifying our stance on whether we allow out-of-order points to be considered "ok" or not.   Implicitly given the stance that the timeseries model is idealized-PRW, I'd say right now the answer would be "no out of order timestamps".  I'll open a slack thread to discuss this and see if we need a formal discussion.
- [ ] I have *not* accounted for what to do with time windows that fall wholly *behind* the last reported time.  I'll expand the algorithm to describe what to do in this scenario.   TL;DR: we have two unlikely failure scenarios we need to think through there:   Clock issues + out-of-order networking.  Both highly unlikely in common practice (>=1m reporting interval), but each with diagnosis challenges.

jmacd: I believe our intention is that OTLP supports out-of-order delivery in general. Your question @victlu refers to **late-arriving** data. I believe that when re-aggregating data in this way, there are two options: delay reporting to allow sufficient time for data to arrive (and refuse to accept data that is too old); allow reporting of data before 100% of it arrives, and report late-arriving copies of the data.

Supporting late-arriving metric data has not been discussed widely in OTel, although the question came up here, https://github.com/open-telemetry/wg-prometheus/issues/46#issuecomment-824306687, and I think we can do this without changing the protocol. The re-aggregator that is aware of late-arriving data may re-report points with the same `StartTimeUnixNano`, `TimeUnixNano`, `Attributes`, and `Resource` and some kind of external label to indicate temporal replication. See also https://github.com/open-telemetry/wg-prometheus/issues/35.

jmacd (approved):

bogdandrutu (approved):

jsuereth:
@jmacd This should be ready to go now.

victlu (approved):

jmacd:
(Technically, we have not enough @open-telemetry/specs-metrics-approvers.)

PR #1630, opened 2021-04-20
Added http.page_title attribute

kkruk-sumo:
This new attribute could be potentially used by [@opentelemetry/instrumentation-document-load](https://www.npmjs.com/package/@opentelemetry/instrumentation-document-load) (PR in progress) to help with diagnosing websites where page URL is not enough (mostly Single Page Applications where URL stay the same, but title changes).


Oberon00:
Conceptually, I think this should not be an HTTP attribute, since it is not really related to HTTP but more to HTML or maybe "web pages". E.g. if you load a `file://` URL (or `data:text/html`), no HTTP is involved but you can still have a page title.

jsuereth:
I think the use case mentioned (Single-Page-Applications) is a great one.

However, I see a few questions:

1. Is the client hitting the same URL to grab in the new content all the time?  (i.e. shouldn't the span be related to the HTTP endpoint you hit to pull in data and not the HTTP location of the initial application load?)
2. Is title-as-the-best-identifying-measure a common thing among single-page web applications?

In other words, should this be standardized, or is it worth prototyping this and writing up an experience report on how to do distributed tracing in single-page apps before we commit to a semantic convention?   I'm not concerned with adding page-title as an optional attribute but I am concerned your stated use case may still not be solved with this change.

yurishkuro:
Agree with @jsuereth 's points. We need to think about prescribing a model that could be used to describe single-page apps before jumping to add semantic conventions.

kkruk-sumo:
I'm going to close this PR. Thank you all for the discussion.

PR #1637, opened 2021-04-22
Clarify that 64 bit integer numbers are decimal strings in OTLP/JSON

tigrannajaryan:
Resolves https://github.com/open-telemetry/opentelemetry-proto/issues/268

This behavior is already the documented behavior for JSON Protobuf,
see https://developers.google.com/protocol-buffers/docs/proto3#json:

>JSON value will be a decimal string. Either numbers or strings are accepted.

### DIFF
--- a/specification/protocol/otlp.md
+++ b/specification/protocol/otlp.md
@@ -363,6 +363,11 @@ they are not base64-encoded like it is defined in the standard
 The hex encoding is used for `trace_id` and `span_id` fields in all OTLP
 Protobuf messages, e.g. the `Span`, `Link`, `LogRecord`, etc. messages.
 
+Note that according to [Protobuf specs](
+https://developers.google.com/protocol-buffers/docs/proto3#json) 64-bit integer
+numbers in JSON-encoded payloads are encoded as decimal strings, and either
+numbers or strings are accepted when decoding.
+
 #### OTLP/HTTP Response
 
 Response body MUST be the appropriate serialized Protobuf message (see below for


obecny:
In collector exporter the number will be sent as `intValue` for numbers between `-2147483648 to 2147483647`;
Other numbers will be sent as `doubleValue` 
https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-exporter-collector/src/transform.ts#L93

When we send data as proto then the whole data is being converted using `protobufjs` package which reads the proto file definitions and converts it automatically.

yurishkuro (approved):

tigrannajaryan:
> In collector exporter the number will be sent as `intValue` for numbers between `-2147483648 to 2147483647`;
> Other numbers will be sent as `doubleValue`
> https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-exporter-collector/src/transform.ts#L93
> 

@obecny Neither int32 nor double are good enough for timestamps (which is one of the most common uses cases for 64 bit integers in proto). int32 does not have enough range, double does not have enough precision. I do not know what are the options to deal with 64 bit numbers in JS and I understand it is painful but unfortunately this is the current reality of the proto.

This PR does not change that at all. This is just a clarification of how 64 bit numbers (which we have plenty in OTLP) work in Protobuf JSON encoding.

jsuereth (approved):

reyang (approved):
LGTM.

obecny (approved):

bogdandrutu (approved):

PR #1640, opened 2021-04-23
Clarify usage of "otel." attribute namespace

tigrannajaryan:
I noticed developers adding their own attributes to this namespace
without going through the specification. We need to regulate this
namespace through the specification, just like we do it for other
semantic conventions.

### DIFF
--- a/specification/common/attribute-and-label-naming.md
+++ b/specification/common/attribute-and-label-naming.md
@@ -130,3 +130,17 @@ It is recommended to limit names to printable Basic Latin characters
 (more precisely to
 [U+0021 .. U+007E](https://en.wikipedia.org/wiki/Basic_Latin_(Unicode_block)#Table_of_characters)
 subset of Unicode code points).
+
+## otel.* Namespace
+
+Attribute and label names that start with `otel.` are reserved to be defined by
+OpenTelemetry specification. These are typically used to express OpenTelemetry
+concepts in formats that don't have a corresponding concept.
+
+For example, the `otel.library.name` attribute is used to record the
+instrumentation library name, which is an OpenTelemetry concept that is natively
+represented in OTLP, but does not have an equivalent in other telemetry formats
+and protocols.
+
+Any additions to the `otel.*` namespace MUST be approved as part of
+OpenTelemetry specification.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ release.
 - Fix the inconsistent formatting of semantic convention enums ([#1598](https://github.com/open-telemetry/opentelemetry-specification/pull/1598/))
 - Add details for filling resource for AWS Lambda([#1610](https://github.com/open-telemetry/opentelemetry-specification/pull/1610))
 - Add already specified `messaging.rabbitmq.routing_key` span attribute key to the respective YAML file.
+- Clarify usage of "otel." attribute namespace ([#1640](https://github.com/open-telemetry/opentelemetry-specification/pull/1640)).
 
 ### Compatibility
 


yurishkuro (approved):
No objections, but what practical impact (aside from being a warning) do you foresee here? Do we need some prescription of how to handle the non-spec attributes that start with `otel.`?

specification/common/attribute-and-label-naming.md
```diff
@@ -130,3 +130,16 @@ It is recommended to limit names to printable Basic Latin characters
...
 [U+0021 .. U+007E](https://en.wikipedia.org/wiki/Basic_Latin_(Unicode_block)#Table_of_characters)
 subset of Unicode code points).
+
+## otel.* Namespace
+
+Attribute names that start with `otel.` are reserved to be defined by
+OpenTelemetry specification. These are typically used to express OpenTelemetry
+concepts in formats that don't have a corresponding concept.
+
+For example `otel.library.name` attribute is used to record the instrumentation
```
arminru, line 140:
```suggestion
For example, the `otel.library.name` attribute is used to record the instrumentation
```

specification/common/attribute-and-label-naming.md
```diff
@@ -130,3 +130,16 @@ It is recommended to limit names to printable Basic Latin characters
...
+library name, which is an OpenTelemetry concept that is natively represented in
+OTLP, but does not have an equivalent in other telemetry formats and protocols.
+
+Any additions to `otel.*` namespace should be approved as part of OpenTelemetry
```
arminru, line 144:
```suggestion
Any additions to the `otel.*` namespace should be approved as part of OpenTelemetry
```

specification/common/attribute-and-label-naming.md
```diff
@@ -130,3 +130,16 @@ It is recommended to limit names to printable Basic Latin characters
 (more precisely to
 [U+0021 .. U+007E](https://en.wikipedia.org/wiki/Basic_Latin_(Unicode_block)#Table_of_characters)
 subset of Unicode code points).
+
+## otel.* Namespace
+
+Attribute names that start with `otel.` are reserved to be defined by
```
arminru, line 136:
```suggestion
Attribute and label names that start with `otel.` are reserved to be defined by
```

arminru (approved):
Sounds reasonable to me but I also wonder about the practical implications as Yuri already stated.
Also, please mention this change in the changelog.

specification/common/attribute-and-label-naming.md
Oberon00, line 140: Nitpick: I wonder if you can call that an attribute. This string will only appear in the Jager/Zipkin exporters at which point its debatable if this is an attribute or another kind of key/value pair.

specification/common/attribute-and-label-naming.md
Oberon00:
```suggestion
Any additions to the `otel.*` namespace MUST be approved as part of OpenTelemetry
```

Oberon00 (approved):

Oberon00:
> what practical impact (aside from being a warning) do you foresee here?

We might define an optional collector filter that discards/warn about unknown `otel` attributes? Although that might be useful only for testing and not an actual production deployment, as it would hamper forward compatibility.

carlosalberto (approved):

carlosalberto:
Looks good to me, once the feedback has been addressed ;)

bogdandrutu (approved):
Overall LGTM, please resolve the comments :)

tigrannajaryan:
> No objections, but what practical impact (aside from being a warning) do you foresee here? 

No other impact. Just a warning to prevent people from polluting "otel." namespace.

>Do we need some prescription of how to handle the non-spec attributes that start with `otel.`?

We do prescribe behavior for each individual attribute (e.g. `otel.library.name` mentioned in the text). I do not think we need a  generic prescription.

One more thing we can do is to refactor the spec a bit and have all `otel.*` attributes listed and defined in one place (currently we have duplicated info about `otel.library.*` for Jaeger and Zipkin exporter, with exactly the same prescribed behavior). This refactoring can be done in a separate PR (not directly related to this PR).

specification/common/attribute-and-label-naming.md
tigrannajaryan: It is supposed to be recorded in the same map (tags in Jaeger vocabulary) where we record attributes. So, it occupies the same namespace of attribute names. I am not sure if there is a better term for this, I am open to suggestions.

carlosalberto:
@tigrannajaryan I think we are ready to merge this. Could you add a CHANGELOG entry before we do so?

tigrannajaryan:
> @tigrannajaryan I think we are ready to merge this. Could you add a CHANGELOG entry before we do so?

Done.

PR #699, opened 2020-07-13
Adding OTLP Exporter

codeboten:
This PR adds configuration and retry behaviour specification for the OTLP exporter. Most of the configuration options are based off of the opentelemetry-collector configuration and the environment variables proposed are based on #666 

Fixes #681

PR #1642, opened 2021-04-23
update OTLP exporter transient errors table

paivagustavo:
## Changes

The transient errors table in the `protocol/exporter.md` specify `Permission Denied` and `Unauthenticated` errors as retryable, but they're defined as _permanent_ error in the [otep-35](https://github.com/open-telemetry/oteps/blob/main/text/0035-opentelemetry-protocol.md#export-response) and in the `protocol/otlp.md` file.

### Related issues

This was found while implementing [retries in the go sdk](https://github.com/open-telemetry/opentelemetry-go/pull/1832#discussion_r618961367).


specification/protocol/exporter.md
```diff
@@ -74,14 +74,12 @@ Transient errors are errors which expect the backend to recover. The following s
 | ---------------- | ----------- |
 | 1  | Cancelled |
 | 4  | Deadline Exceeded |
-| 7  | Permission Denied |
 | 8  | Resource Exhausted |
 | 10 | Aborted |
 | 10 | Out of Range |
 | 14 | Unavailable |
 | 15 | Data Loss |
-| 16 | Unauthenticated |
```
arminru, line 83: We'll need to have a discussion if this change is considered breaking as this is declared [stable](https://github.com/open-telemetry/opentelemetry-specification/blob/d0d2b489f15c9af585c63783434785b4e705b19d/specification/document-status.md) already.
In theory, one could rely on exports being retried in these cases although I think it's unlikely.

arminru (commented):
In general, the spec is usually more up to date than the respective OTEP. OTEPs are used for the initial discussion but as details go into the spec, changes applied there are usually not reflected back in the OTEP. @tigrannajaryan PTAL if this deviation is intended or not.

It sounds reasonable to remove them, however, as I don't think we would be able to recover from missing or misconfigured credentials unless authentication was turned off or permissions for the configured credentials were changed during runtime of the exporter. I'm not sure if this is likely enough so that consistently retrying is worth it.

tigrannajaryan:
[specification/protocol/exporter.md](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md) is incorrect. [specification/protocol/otlp.md](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md) is correct. I think a mistake happened when https://github.com/open-telemetry/opentelemetry-specification/pull/699 added specification/protocol/exporter.md file to the spec.

The retryable codes are defined by OTEP 0065: https://github.com/open-telemetry/oteps/pull/65. Permission Denied and Unauthenticated were pointed out by @yurishkuro that they should be non-retryable and I agreed and that's how OTEP 0065 was approved.

I would consider this a typo in `specification/protocol/exporter.md` that needs to be fixed. `specification/protocol/otlp.md` is the canonical information for OTLP and it is corerct.

We should probably remove the list of codes from `specification/protocol/exporter.md` and just refer to `specification/protocol/otlp.md` as the source of truth.

IMO, this is not a breaking change in any way, just a typo correction in the spec.

arminru:
> I would consider this a typo in `specification/protocol/exporter.md` that needs to be fixed. `specification/protocol/otlp.md` is the canonical information for OTLP and it is corerct.

I agree.

> We should probably remove the list of codes from `specification/protocol/exporter.md` and just refer to `specification/protocol/otlp.md` as the source of truth.

Yes, that would be cleaner. @paivagustavo could you adapt the PR accordingly? Thanks!

tigrannajaryan:
Let's make sure we also have this change listed in CHANGELOG so that anyone who implemented it incorrectly can fix. I believe it should be considered an implementation bug, no a breaking behavior change.

specification/protocol/exporter.md
```diff
@@ -59,27 +59,7 @@ The `OTEL_EXPORTER_OTLP_HEADERS`, `OTEL_EXPORTER_OTLP_TRACES_HEADERS`, `OTEL_EXP
...
-[Transient errors](#transient-errors) MUST be handled with a retry strategy. This retry strategy MUST implement an exponential back-off with jitter to avoid overwhelming the destination until the network is restored or the destination has recovered.
-
-## Transient Errors
-
-Transient errors are errors which expect the backend to recover. The following status codes are defined as transient errors:
-
-| HTTP Status Code | Description |
-| ---------------- | ----------- |
-| 408 | Request Timeout |
-| 5xx | Server Errors |
```
arminru, line 71: The referenced section in `otlp.md#failures` actually does not specify this.
Could you a statement there pointing out that 408 and 5xx are considered retryable or otherwise leave this table here?

paivagustavo: Right, I've added a statement that point to that info. But I think It would be better if we have a full table just like the `gRPC`.

arminru: This would be ideal yes. The current state is enough for this bugfix PR, however.

arminru (approved):

tigrannajaryan (approved):

carlosalberto (approved):

PR #1606, opened 2021-04-07
added JSON RPC specific attributes

Rast1234:
Tracing attributes for [JSON RPC protocol](https://www.jsonrpc.org/specification).

## Changes

Added attributes:

```
rpc.jsonrpc.version
rpc.jsonrpc.method
rpc.jsonrpc.id
rpc.jsonrpc.error_code
rpc.jsonrpc.error_message
```

Not sure if `null` values are OK for attributes, maybe need to change `rpc.jsonrpc.id` a bit. Also not sure if `error_message` is OK because it is arbitrary (but short!) text, can impact cardinality.

Related issue: #1601

Issue #1601, opened 2021-04-06
Please guide how to properly extend attributes for not-yet-documented protocols


Rast1234:
**What are you trying to achieve?**

I am implementing tracing library for Json Rpc 2.0. It would be convenient to have some additional attributes as GRPC does: `rpc.grpc.status_code`. However, it is not clear to me how to extend attributes properly. I can think of two approaches:

1. Use existing attrs and extend them:
  * `rpc.system = jsonrpc`. **Question**: is this attribute intended to be used for any sort of RPC? Since i can't find closed list of allowed values...
  *  `rpc.jsonrpc.something_protocol_related = "blah"` because GRPC does the same

2. Make completely separate attrs like `com.my_organization.jsonrpc.something_protocol_related` since Json Rpc is not standartized here (yet) to avoid collisions in future

I followed docs about [semantic conventions ](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/trace/semantic_conventions) and [naming](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-and-label-naming.md) and i see the pattern how attributes extend for specific protocols. However i'm not sure if extensibility by nesting is intended for developers outside of opentelemetry project. **Question here is:** should we treat attributes mentioned in these docs as *read-only allowed things* or can we extend them?

Oberon00:
> `rpc.system = jsonrpc`. **Question:** is this attribute intended to be used for any sort of RPC? Since i can't find closed list of allowed values...

Good point, that should probably be formatted as an (open) enum. It would be intended to be used for such a case.

> should we treat attributes mentioned in these docs as read-only allowed things or can we extend them?

Most "enum"-like attributes explicitly allow using "a custom value", and rpc.system is IMHO supposed to be one of them.

Oberon00:
For adding new attributes, I would use a custom prefix. E.g. for your example I would use `my_organization.rpc.jsonrpc.something_protocol_related` if you do not intend to contribute semantic conventions for Json Rpc to OpenTelemetry. However if we are talking about that JSON RPC https://en.wikipedia.org/wiki/JSON-RPC, then I think it would be best to create a PR with your suggested semantic conventions. In the meantime, if you already need them, use the version with the prefix internally or be prepared for them to change during PR review.

Rast1234:
Thanks!

 > Most "enum"-like attributes explicitly allow using "a custom value"
[...]

PR #1643, opened 2021-04-24
added JSON RPC specific attributes

Rast1234:
This is second attempt: PR #1606 was automatically closed by bot.

Tracing attributes for [JSON RPC protocol](https://www.jsonrpc.org/specification).

## Changes

Added attributes:

```
rpc.jsonrpc.version
rpc.jsonrpc.method
rpc.jsonrpc.id
rpc.jsonrpc.error_code
rpc.jsonrpc.error_message
```

Not sure if `null` values are OK for attributes, maybe need to change `rpc.jsonrpc.id` a bit. Also not sure if `error_message` is OK because it is arbitrary (but short!) text, can impact cardinality.

Related issue: #1601


semantic_conventions/trace/rpc.yaml
```diff
@@ -78,3 +78,33 @@ groups:
           required: always
           brief: "The [numeric status code](https://github.com/grpc/grpc/blob/v1.33.2/doc/statuscodes.md) of the gRPC request."
           examples: [0, 1, 16]
+    - id: rpc.jsonrpc
+      prefix: rpc.jsonrpc
+      extends: rpc
+      brief: 'Tech-specific attributes for JSON RPC.'
```
arminru, line 84:
```suggestion
      brief: 'Tech-specific attributes for [JSON RPC](https://www.jsonrpc.org/).'
```

semantic_conventions/trace/rpc.yaml
```diff
@@ -78,3 +78,33 @@ groups:
...
+        - id: version
+          type: string
+          required:
+            conditional: 'If missing, it is assumed to be "2.0".'
+          brief: "Protocol version as in `jsonrpc` property of request/response. Since first protocol version does not specify this, value can be omitted entirely."
+          examples: ['2.0', '1.0']
+        - id: method
+          type: string
+          required: always
+          brief: "`method` property from request. Unlike `rpc.method`, this may not relate to the actual method being called."
```
arminru, line 95: Could you add a note to help distinguish this better (also for telemetry consumers)? Can one still derive the proper value for `rpc.method` (= actual method being called) from the client-side or would this only be available on the server?

semantic_conventions/trace/rpc.yaml
```diff
@@ -78,3 +78,33 @@ groups:
           required: always
           brief: "The [numeric status code](https://github.com/grpc/grpc/blob/v1.33.2/doc/statuscodes.md) of the gRPC request."
           examples: [0, 1, 16]
+    - id: rpc.jsonrpc
```
arminru, line 81: Please make sure to reference this new group in `specification/trace/semantic_conventions/rpc.md` and run the markdown generator as documented here: https://github.com/open-telemetry/opentelemetry-specification/tree/main/semantic_conventions

arminru, line 81: Please add an entry in `CHANGELOG.md`.

semantic_conventions/trace/rpc.yaml
```diff
@@ -78,3 +78,33 @@ groups:
...
+        - id: error_code
+          type: int
+          required:
+            conditional: 'If missing, response is assumed to be successful.'
+          brief: "`error.code` property of response if it is an error response."
+          examples: [-32700, 100]
+        - id: error_message
+          type: string
+          brief: "`error.message` property of response if it is an error response."
+          examples: ['Parse error', 'User already exists']
```
arminru, line 110: Is there always a 1:1 mapping to `error_code` where the message would not provide any additional information but could rather be looked up in the JSON-RPC spec given the code, or can one add additional, custom error codes themselves?

semantic_conventions/trace/rpc.yaml
```diff
@@ -78,3 +78,33 @@ groups:
...
+          examples: ['2.0', '1.0']
+        - id: method
+          type: string
+          required: always
+          brief: "`method` property from request. Unlike `rpc.method`, this may not relate to the actual method being called."
+          examples: ['users.create', 'get_users']
+        - id: id
+          type: string
+          brief: "`id` property of request. Since protocol allows id to be int, string, `null` or missing (for notifications), value is expected to be cast to string for simplicity. Omit if this is a notification."
+          examples: ['10', 'request-7', '']
```
arminru, line 100: I wonder if this would be generic enough to be added on the parent-level (e.g., as `rpc.call_id` or `rpc.request_id`) so we don't have overlaps/duplicates later if this is added for other protocols/frameworks as well. @open-telemetry/specs-approvers @open-telemetry/specs-trace-approvers any opinions?

semantic_conventions/trace/rpc.yaml
```diff
@@ -78,3 +78,33 @@ groups:
...
+          required:
+            conditional: 'If missing, it is assumed to be "2.0".'
+          brief: "Protocol version as in `jsonrpc` property of request/response. Since first protocol version does not specify this, value can be omitted entirely."
+          examples: ['2.0', '1.0']
+        - id: method
+          type: string
+          required: always
+          brief: "`method` property from request. Unlike `rpc.method`, this may not relate to the actual method being called."
+          examples: ['users.create', 'get_users']
+        - id: id
```
arminru, line 97: `id` is quite ambiguous (could be an identifier for the caller, server, method, request, session, ...). How about `request_id` since this is the `id` property of the `request` object?

semantic_conventions/trace/rpc.yaml
Rast1234: agreed, added a note on that

semantic_conventions/trace/rpc.yaml
Rast1234: done

semantic_conventions/trace/rpc.yaml
Rast1234: Uh, my bad, `id` is also used it response and shuold be the same as in request (if it existed).

semantic_conventions/trace/rpc.yaml
Rast1234: Totally forgot the actual human-readable document! Updated and generated table. Also updated TOC with `markdown-toc` (not done by Makefile)

Rast1234:
Fixed and replied where i could

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ release.
 - Clarify non-blocking requirement from span API End. ([#1555](https://github.com/open-telemetry/opentelemetry-specification/pull/1555))
 - Remove the Included Propagators section from trace API specification that was a duplicate of the Propagators Distribution of the context specification. ([#1556](https://github.com/open-telemetry/opentelemetry-specification/pull/1556))
 - Remove the Baggage API propagator notes that conflict with the API Propagators Operations section and fix [#1526](https://github.com/open-telemetry/opentelemetry-specification/issues/1526). ([#1575](https://github.com/open-telemetry/opentelemetry-specification/pull/1575))
+- Add JSON RPC specific conventions ([#1643](https://github.com/open-telemetry/opentelemetry-specification/pull/1643)).
```
arminru, line 24: There is a separate section for semantic conventions below.

specification/trace/semantic_conventions/rpc.md
```diff
@@ -164,3 +166,21 @@ one for sent messages and one for received message. This way we guarantee that
 the values will be consistent between different implementations. In case of
 unary calls only one sent and one received message will be recorded for both
 client and server spans.
+
+## JSON RPC
+
+Conventions specific to [JSON RPC](https://www.jsonrpc.org/), namely `2.0`.
```
arminru, line 172: 1.x is also supported, right?

semantic_conventions/trace/rpc.yaml
arminru: In the JSON-RPC spec it says that for also for the response "It MUST be the same as the value of the id member in the Request Object." (https://www.jsonrpc.org/specification#response_object).
So `request_id` should be a good and unambiguous name here, right?

semantic_conventions/trace/rpc.yaml
arminru: I see. In this case, preserving the message could be helpful.
We could consider adding a "_SHOULD NOT be set for error codes defined in the specification_" but it's probably not worth the special handling since the messages are indeed quite concise and not an entire stack trace or something like that.

arminru (commented):
Thanks!
Please ignore the failing docfx check for now, it's just Chocolatey currently being down (https://status.chocolatey.org).

semantic_conventions/trace/rpc.yaml
```diff
@@ -78,3 +78,33 @@ groups:
...
+          brief: "Protocol version as in `jsonrpc` property of request/response. Since first protocol version does not specify this, value can be omitted entirely."
+          examples: ['2.0', '1.0']
+        - id: method
+          type: string
+          required: always
+          brief: "`method` property from request. Unlike `rpc.method`, this may not relate to the actual method being called. Useful for client-side traces since client does not know what will be called on the server."
+          examples: ['users.create', 'get_users']
+        - id: id
+          type: string
+          brief: "`id` property of request or response. Since protocol allows id to be int, string, `null` or missing (for notifications), value is expected to be cast to string for simplicity. Omit if this is a notification."
```
arminru, line 99: > Not sure if `null` values are OK for attributes, maybe need to change `rpc.jsonrpc.id` a bit.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attributes says:

> Attribute values of `null` are not valid and attempting to set a `null` value is undefined behavior.

You could change it to either be omitted, passed as an empty string or the string `"null"` in this case.

semantic_conventions/trace/rpc.yaml
Rast1234: Yes, stack trace and other rich information is to be stored in `error.data` property.

semantic_conventions/trace/rpc.yaml
Rast1234: For JSON RPC, yes, `request_id` is OK. Not sure about RPCs. Should we wait for response from people you tagged in another thread?

specification/trace/semantic_conventions/rpc.md
Rast1234: Yes but by default people think about 2.0 that's why i wrote that. Is it misleading?

CHANGELOG.md
Rast1234: fixed

semantic_conventions/trace/rpc.yaml
Rast1234: done

specification/trace/semantic_conventions/rpc.md
arminru: I think it is since one can still specify values for 1.0. Does it make any difference for the attributes except for `rpc.jsonrpc.version`?

semantic_conventions/trace/rpc.yaml
arminru: Currently your attribute is in the JSON RPC scope anyway but we can also wait for the outcome of the other thread and leave this one open for now.

specification/trace/semantic_conventions/rpc.md
Rast1234: As far as i know V1 major difference aside "version" is that it does not specify `error` structure so we can't get code or message. So this convention works for V1 IMO

semantic_conventions/trace/rpc.yaml
arminru: Not sure if we get an opinion on the other thread, so let's go with `rpc.jsonrpc.request_id` for now.
gRPC and Java RMI, for example, don't have such an ID, so it's probably rather JSON-RPC-specific anyway.

specification/trace/semantic_conventions/rpc.md
arminru: As per https://www.jsonrpc.org/specification_v1 it does have an error property but doesn't specify it further so it seems like any non-null value can be used to signal an error - this would be hard to spec anyway. If someone is using v1, they can just set the [span status](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.2.0/specification/trace/api.md#set-status) to `Error` to signal that (i.e., if `error != null`).
I'd say let's just remove the "_, namely 2.0_" then as it can be used for both.

semantic_conventions/trace/rpc.yaml
```diff
@@ -78,3 +78,33 @@ groups:
...
+    - id: rpc.jsonrpc
+      prefix: rpc.jsonrpc
+      extends: rpc
+      brief: 'Tech-specific attributes for [JSON RPC](https://www.jsonrpc.org/).'
+      attributes:
+        - id: version
+          type: string
+          required:
+            conditional: 'If missing, it is assumed to be "2.0".'
+          brief: "Protocol version as in `jsonrpc` property of request/response. Since first protocol version does not specify this, value can be omitted entirely."
```
arminru, line 90:
```suggestion
          required:
            conditional: 'If missing, it is assumed to be "1.0".'
          brief: "Protocol version as in `jsonrpc` property of request/response. Since JSON-RPC 1.0 does not specify this, the value can be omitted."
```

I just noticed that this is not consistent. If it is missing, `1.0` would have to be assumed, right?

arminru (approved):

carlosalberto (approved):

PR #1646, opened 2021-04-26
Add detail about StartTimeUnixNano, gaps and resets

jmacd:
## Changes

Adds detail about when to use `StartTimeUnixNano` and handling of unknown start-time resets.

Related issues:
https://github.com/open-telemetry/opentelemetry-proto/issues/289
https://github.com/open-telemetry/opentelemetry-proto/issues/292


reyang (approved):
LGTM.

specification/metrics/datamodel.md
```diff
@@ -282,14 +282,24 @@ degradation or loss of visibility.
...
-measurement was taken and when it was reported as being still the last value.
-
-The notion of temporality refers to a configuration choice made in the system
-as a whole, indicating whether reported values incorporate previous
-measurements, or not.
+Every OTLP metric data point has two associated timestamps.  The
+first, mandatory timestamp is the one associated with the observation,
+the moment when the measurement became current or took effect, and is
+referred to as `TimeUnixNano`.  The second, optional timestamp is used
+to indicate when a sequence of points is unbroken, and is referred to as
```
jsuereth, line 289: Nit:  "The second, optional, timestamp is used to detect missing points within a metric stream."

specification/metrics/datamodel.md
```diff
@@ -282,14 +282,24 @@ degradation or loss of visibility.
...
+`StartTimeUnixNano`.
+
+The second timestamp is strongly recommended for Sum, Histogram, an
+Summary points, as it is necessary to correctly interpret the rate
+from an OTLP stream, in a manner that is aware of restarts.  The use
+of `StartTimeUnixNano` to indicate the start of an unbroken sequence
+of points means it can also be used to encode implicit gaps in
+the stream.
+
+The notion of temporality refers to the way additive quanties are
```
jsuereth, line 299: This likely needs to be the first paragraph of this section.  If you've already described what delta + cumulative mean, it's a lot easier to introduce the timestamps and use these terms in the description.

specification/metrics/datamodel.md
```diff
@@ -313,13 +323,76 @@ event commonly is translated into two metric events (a 1-count and a timing
...
+points in the same sequence.  For the initial points in an unbroken
+sequence:
+
+- When `StartTimeUnixNano` is less than `TimeUnixNano`, a new unbroken sequence of observations begins with a "true" reset at a known start time. The zero value is implicit, it is not necessary to record the starting point.
+- When `StartTimeUnixNano` equals `TimeUnixNano`, a new unbroken sequence of observations begins with a reset at an unknown start time. The initial observed value is recorded to indicate that an unbroken sequence of observations resumes. These points have zero duration, and indicate that nothing is known about previously-reported points and that data may have been lost.
+
+For subsequent points in an unbroken sequence:
+
+- For points with delta aggregation temporality, the `StartTimeUnixNano` of each point matches the `TimeUnixNano` of the preceding point
+- Otherwise, the `StartTimeUnixNano` of each point matches the `StartTimeUnixNano` of the initial observation.
```
jsuereth, line 343: I like this definition for cumulative start time. It allows some good flexibility.

specification/metrics/datamodel.md
```diff
@@ -313,13 +323,76 @@ event commonly is translated into two metric events (a 1-count and a timing
...
+To correctly compute the rate contribution of the first point in a
+unbroken sequence requires knowing whether it is the first point.
+Unknown start-time reset points appear with `TimeUnixNano` equal to
+the `StartTimeUnixNano` of a stream of points, in which case the rate
+contribution of the first point is considered zero.  An earlier
+sequence of observations is expected to have reported the same
+cumulative state prior to a gap in observations.
+
+The presence or absence of a point with `TimeUnixNano` equal to the
+`StartTimeUnixNano` indicates how to count rate contribution frmo the
```
jsuereth, line 366: s/frmo/from/

specification/metrics/datamodel.md
```diff
@@ -313,13 +323,76 @@ event commonly is translated into two metric events (a 1-count and a timing
...
+first point in a sequence.  If the first point in an unknown
+start-time reset sequence is lost, the consumer of this data is might
```
jsuereth, line 368: s/is//

jsuereth (commented):
Generally looks good, just a bit of wordsmithing nits.

specification/metrics/datamodel.md
```diff
@@ -313,13 +323,76 @@ event commonly is translated into two metric events (a 1-count and a timing
...
+overcount the rate contribution of the second point, as it then appears
+like a "true" reset.
```
victlu, line 370: I'm trying to understand...

> If the first point in an unknown start-time reset sequence is lost, the consumer of this data is might
> overcount the rate contribution of the second point, as it then appears like a "true" reset.

In cumulative stream, each data point (that is not an unknown start-time reset point) has t0, t1, and a sum. I would think that's enough to calculate rate for that data point. Thus, we can get the rate contribution of the second point regardless of losing the 1st unknown start-time reset point.

I would also think the issue is more related to our ability to correctly calculating the true cumulative sum. If we lose the 1st data point, as it should contain the initial observed value, then we cannot correctly assume the sum in the second data point represent true cumulative sum (as we'll need a reference from our algorithm to "assume" a previous known sum)

specification/metrics/datamodel.md
```diff
@@ -282,14 +282,24 @@ degradation or loss of visibility.
...
-as a whole, indicating whether reported values incorporate previous
-measurements, or not.
+Every OTLP metric data point has two associated timestamps.  The
+first, mandatory timestamp is the one associated with the observation,
+the moment when the measurement became current or took effect, and is
+referred to as `TimeUnixNano`.  The second, optional timestamp is used
+to indicate when a sequence of points is unbroken, and is referred to as
+`StartTimeUnixNano`.
+
+The second timestamp is strongly recommended for Sum, Histogram, an
```
victlu, line 292: s/an/and

specification/metrics/datamodel.md
jsuereth: You only have enough information to calculate rate if you assume all counters start at 0.

That's not necessarily true for things like memory usage.

victlu: True.

So, does it make sense to include the observed value at `StartTimeUnixNano` for each data point? As it stand currently, we are asking for a 1st data point with the initial observed value anyways. If the observed value at `StartTimeUnixNano` is included in every data point, we will not run into issues with lost data points.

jmacd: @victlu you are understanding this right :joy:

> If the observed value at StartTimeUnixNano is included in every data point

I have thought of proposing more detail on how problems of this nature can be solved, and one of them is to apply the start time of the process as a resource attribute. If cumulative values reset off of the start time, a new semantic-conventional metric attribute could be introduced. (This topic is as much about Resource attributes as it is about StartTimeUnixNano.)

> In that case, the observed value will be an optional value in each data point. And this situation will be equivalent to us losing the 1st data point.

That's right. Remember, this only impacts data imported from sources without start times--the desired remedy is to get true start times! Either the approach mentioned above (add a resource) or the one mentioned below ("inserting true reset points") apply in this situation, meaning we can avoid over-counting by always under-counting.

specification/metrics/datamodel.md
jmacd: Done!

specification/metrics/datamodel.md
jmacd: Fixed (as you suggested below).

victlu (approved):

jsuereth (approved):

Issue #1297, opened 2020-12-18
Metrics: Requirements for safe attribute removal


jmacd:
**What are you trying to achieve?**

Make it safe to remove labels from OTLP metrics in a way that
preserves their semantics.  We have identified two scenarios where it
is not safe to simply strip labels away from metrics data.

1. When all identifying resource attributes have been removed from a
CUMULATIVE metric, the resulting timeseries has overlapping time
intervals (i.e., their `StartTimeUnixNanos` and `TimeUnixNanos`
overlap).
2. When `SumObserver` and `UpDownSumObserver` data points are stripped
of labels that were used to subdivide an observed sum, the resulting
points lose meaning.

An example for both is provided below.

### CUMULATIVE points and the loss of unique labels

For the first item, OTLP has not specified how to interpret CUMULATIVE
data in this scenario.  For reference, current OTLP Metrics Sum data
point defines `AggregationTemporality` as follows:

```
  // DELTA is an AggregationTemporality for a metric aggregator which reports
  // changes since last report time. Successive metrics contain aggregation of
  // values from continuous and non-overlapping intervals.
  //
  // The values for a DELTA metric are based only on the time interval
  // associated with one measurement cycle. There is no dependency on
  // previous measurements like is the case for CUMULATIVE metrics.
```

and:

```
  // CUMULATIVE is an AggregationTemporality for a metric aggregator which
  // reports changes since a fixed start time. This means that current values
  // of a CUMULATIVE metric depend on all previous measurements since the
  // start time. Because of this, the sender is required to retain this state
  // in some form. If this state is lost or invalidated, the CUMULATIVE metric
  // values MUST be reset and a new fixed start time following the last
  // reported measurement time sent MUST be used.
```

Simply translating overlapping CUMULATIVE data into Prometheus results in incorrect interpretation
because timeseries are not meant to have overlapping points (e.g.,
https://github.com/open-telemetry/opentelemetry-collector/issues/2216).
For this reason, we should specify that CUMULATIVE timeseries MUST
always retain at least one uniquely identifying resource attribute.

[This point suggests that `service.instance.id` should be a required
resource attribute.](https://github.com/open-telemetry/opentelemetry-specification/issues/1034): https://github.com/open-telemetry/opentelemetry-specification/issues/1034

We should also specify how consumers are expected to handle the
situation where, because of a mis-configuration, data is presented with some
[...]

PR #1648, opened 2021-04-26
Add detail about metric staleness

jmacd:
## Changes

Adds information about staleness markers using Prometheus-style NaN values.

Related issues #1297


specification/metrics/datamodel.md
```diff
@@ -350,17 +350,86 @@ points may be expected. In this case, OpenTelemetry collectors SHOULD modify
 points at the change-over using interpolation for Sum data points, to reduce
 gaps to zero width in these cases, without any overlap.
 
-## Resources
+## Staleness
 
-Pending
+Implementations have implicit and explicit means of detecting when a
```
jsuereth, line 355: `<putting on my author hat>`
This paragraph struggles in introducing the concept and transitioning to the meat of the technical details.

Suggestion:

> Staleness is a property of a metric stream.   When a metric stream that previously produced data is no longer > producing new (valid) data, the stream is considered "stale", i.e. we know the stream existed, but we're not
> sure whether or not to expect data going forward.

> Consumers and Producers of metric streams have two mechanisms to denote "staleness" of a metric stream:  > An implicit mechanism that leverages expectation on when metric points should arrive, and an explicit 
> mechanism where staleness is denoted directly in metric data points.


I think the helps motivate the next to sections/details a bit better.  I'm not certain if you want to outline the use-case of staleness better.
`</removing the author hat>`

specification/metrics/datamodel.md
```diff
@@ -350,17 +350,86 @@ points may be expected. In this case, OpenTelemetry collectors SHOULD modify
...
+stream ceases being written.  Consumers of metric streams may wish to
+know when a stream is defined by an invalid value, a condition known
+as "staleness".
+
+### Explicit Staleness
+
+The explicit mechanism, characteristic of the Prometheus Remote-Write
```
jsuereth, line 362: nit: "The explicit staleness detection mechanism"

specification/metrics/datamodel.md
```diff
@@ -350,17 +350,86 @@ points may be expected. In this case, OpenTelemetry collectors SHOULD modify
...
+protocol, uses IEEE 754 floating point NaN values to encode invalid
+measurements with an explicit timestamp.
+
+NaN values can be used to indicate an explicit gap in an metric stream
```
jsuereth, line 366: Nit:  I'd rework this wording to point at the actual OTLP types.


> Nan values can be used to indicate an explicit gap in a metric stream.
> - For Gauge and Sum streams, place NaN in `NumberDataPoint.as_double`
> - For Histogram streams, place NaN in `HistogramDataPoint.sum`
> - For Summary streams, place NaN in `SummaryDataPoint.sum`

specification/metrics/datamodel.md
```diff
@@ -350,17 +350,86 @@ points may be expected. In this case, OpenTelemetry collectors SHOULD modify
...
+These observations can be included in a sequence of measurements
+without being considered resets, meaning they do not interact with the
+interpretation of `StartTimeUnixNano`.  NaN-valued points are
+considered explicitly invalid point values.
+
+### Implicit Staleness
+
+The implicit mechanism, characteristic of systems that push metric
+data, have a time-to-live (TTL) associated with metric streams.  In
+these systems, data is expected to arrive more often than a specified
```
jsuereth, line 379: Is it worth denoting that this tends to be associated with "push-based" metric protocols?

specification/metrics/datamodel.md
```diff
@@ -350,17 +350,86 @@ points may be expected. In this case, OpenTelemetry collectors SHOULD modify
...
+considered stale.
+
+Implementations that perform a staleness calculation can be configured
+to issue an explicit staleness marker, or they can simply forget those
+streams.
 
 ## Temporal Alignment
 
-Pending
+Temporal alignment describes how to modify a stream of metric data
```
jsuereth, line 389: I think this may need an image to help outline it better.  I'm happy to provide one if you want.

specification/metrics/datamodel.md
```diff
@@ -350,17 +350,86 @@ points may be expected. In this case, OpenTelemetry collectors SHOULD modify
...
+with arbitary `StartTimeUnixNano` and `TimeUnixNano` timestamps into a
+stream with selected timestamps, typically into a stream of
+constant-duration points.  Temporal alignment is defined on the basis
+of point kind and is performed in a way that respects the original
+aggregation.
+
+Temporal alignment is performed after overlapping data points have
```
jsuereth, line 396: One question I have around temporal-alignment *and* streaming metric data is what kind of "look-behind" we need on the stream.  I.e. how much buffering do we need/plan to have to perform something like temporal alignment.

Does it make sense to outline the streaming use case specifically?  E.g. my (naive) brain keeps coming back to the notion that it may make sense to define a "simple" temporal alignment that:

- Keeps *one* TTL of data at any given time.
- Performs alignment within the "current window" and drops points that fall entirely before the "current" window after aligning them. (not sure I used good english here)
- A heuristic for when to stop collecting/aligning for a given bucket where the next piece of data is fired out.

That's likely way to detailed for just outlining the concept and ensuring implementations know where and how to use the data in the data model, but wanted to run the idea by you.

specification/metrics/datamodel.md
```diff
@@ -350,17 +350,86 @@ points may be expected. In this case, OpenTelemetry collectors SHOULD modify
...
+### Sum and Histogram points with delta aggregation temporality
+
+For delta aggregation temporality, Sum values, Histogram bucket
+counts, and the Histogram sum and count fields allocate their value
+proportionally into the output, based on the fraction of the input
+time range covered by the output time range.
+
+### Sum and Histogram points with cumulative aggregation temporality
+
+For cumulative aggregation temporality, Sum values, Histogram bucket
```
jsuereth, line 416: I think this is good enough to demonstrate feasibility.

Couldn't you just use the same fractional contribution as done in the delta aggregation, except you'd *also* preserve the previous time-window values and aggregate those in?

specification/metrics/datamodel.md
```diff
@@ -350,17 +350,86 @@ points may be expected. In this case, OpenTelemetry collectors SHOULD modify
...
+For Gauge points, temporal alignment outputs the input value with the
+most recent `TimeUnixNano` prior to the output `TimeUnixNano`.
 
-## External Labels
+### Summary points
 
-Pending
+For Summary points, temporal alignment applies the logic for Histogram
+sum and count fields to the Summary sum and count fields, and it uses
+the logic for Gauge values to its `ValueAtQuantile` field.
```
jsuereth, line 432: (curiousity, non-blocking) Is this how it's done in prometheus?  I mean I can't see any other alternatives here.

specification/metrics/datamodel.md
```diff
@@ -350,17 +350,86 @@ points may be expected. In this case, OpenTelemetry collectors SHOULD modify
...
 gaps to zero width in these cases, without any overlap.
 
-## Resources
+## Staleness
 
-Pending
+Implementations have implicit and explicit means of detecting when a
+stream ceases being written.  Consumers of metric streams may wish to
+know when a stream is defined by an invalid value, a condition known
+as "staleness".
```
victlu, line 358: Just a commentary...

I always thought of these as two different concepts. Invalid values are just that, and has no association with staleness. 

In my mind, staleness has to do with timeliness of observed value (even if it's invalid) and I think of it as "freshness". Thus, we provide, in our UI, a freshness timestamp of the last known value. And, we leave it to our users to apply their own "TTL" values for each metric (on UI side). This helps when they see "invalid" values that is fresh as of 1 minute ago. When we showed no value or invalid value only or no freshness timestamp, our users constantly ask if the telemetry system is up or down.

specification/metrics/datamodel.md
```diff
@@ -350,17 +350,86 @@ points may be expected. In this case, OpenTelemetry collectors SHOULD modify
...
+
+### Explicit Staleness
+
+The explicit mechanism, characteristic of the Prometheus Remote-Write
+protocol, uses IEEE 754 floating point NaN values to encode invalid
```
brian-brazil, line 363: Not quite, it's not any NaN but rather one particular bit pattern which happens to be a NaN. This is distinct from a NaN that can be scrapped directly, which will have a different bit pattern.

specification/metrics/datamodel.md
jmacd: I tried to reword this sentence after your feedback and the feedback above.

specification/metrics/datamodel.md
jmacd: 👍 I've written a recommendation for "any IEEE 754 floating point NaN value" to indicate invalid values. I want to just use what IEEE says about NaN values as the definition for what these mean as observations.

specification/metrics/datamodel.md
jmacd: I've removed the content about temporal alignment. I will preserve your feedback.

specification/metrics/datamodel.md
jmacd: Yes.

jmacd:
@open-telemetry/specs-metrics-approvers PTAL.

specification/metrics/datamodel.md
brian-brazil: That's still incorrect, as written you're saying that any NaN means stale and that NaNs are never valid values.

MrAlias (approved):

specification/metrics/datamodel.md
```diff
@@ -567,17 +567,62 @@ points may be expected. In this case, OpenTelemetry collectors SHOULD modify
...
+data point to be independent of the `Attributes`, `StartTimeUnixNanos`
+and `TimeUnixNanos` fields.
+
+The recommended mechanism for active monitors to convey invalid
+measurements, characteristic of the Prometheus Remote-Write protocol,
+is to encode any IEEE 754 floating point NaN value:
+
+- For `Sum` data points, use the `NumberDataPoint.as_double` field
+- For `Gauge` data points, use the `NumberDataPoint.as_double` field
+- For `Histogram` data points, use the `HistogramDataPoint.count` field
```
jsuereth, line 606: `HistogramDataPoint.count` is `fixed64` which is an integer type and NaN is unrepresentable here.

Similar concern for Summary count.

jsuereth (changes requested):

jsuereth:
I'd like to talk about this at Today's data-model SiG.

I have some open questions:

1. Should NaN be used to represent staleness?
2. How do we represent staleness in Histogram / Summary where we have an aggregation of data with lots of optional points?

This PR answers:

1. Yes
2. Use `count` as it's required (unfortunately, it's also not a double).

I know 2 is a problem, not sure I agree with 1, want to talk that through.

specification/metrics/datamodel.md
```diff
@@ -567,17 +567,62 @@ points may be expected. In this case, OpenTelemetry collectors SHOULD modify
...
+data point to be independent of the `Attributes`, `StartTimeUnixNanos`
+and `TimeUnixNanos` fields.
+
+The recommended mechanism for active monitors to convey invalid
+measurements, characteristic of the Prometheus Remote-Write protocol,
+is to encode any IEEE 754 floating point NaN value:
+
+- For `Sum` data points, use the `NumberDataPoint.as_double` field
+- For `Gauge` data points, use the `NumberDataPoint.as_double` field
+- For `Histogram` data points, use the `HistogramDataPoint.sum` field
```
bogdandrutu, line 606: Why not `Count`? Seems more natural than `Sum`.

jmacd:
We have discussed replacing this PR with a data model proposal that takes advantage of optional bits to encode staleness and potentially last-data records. @victlu will begin working on a prototype. :tada:

specification/metrics/datamodel.md
jsuereth: count is an integer :(

PR #1649, opened 2021-04-27
Add detail about reaggregation for safe attribute removal

### LINKED ISSUE #1297: Metrics: Requirements for safe attribute removal

jmacd:
Fixes #1297 

## Changes

Adds information about how to re-aggregate metric streams for safe semantic attribute removal.
Depends on: #1618 #1646 #1648.

### DIFF
--- a/specification/metrics/datamodel.md
+++ b/specification/metrics/datamodel.md
@@ -362,6 +362,46 @@ Pending
 
 Pending
 
+## Reaggregation
+
+Reaggregation may be used to correctly combine multiple metric streams
+into a single stream, according to their semantics.  Reaggregation can
+be used to remove attributes from metric data points and resources and
+to lower temporal resolution.
+
+Reaggregation has already been described in specific examples, for
+example, transforming delta into cumulative aggregation temporality,
+and the process of inserting true reset points in cumulative streams.
+
+Reaggregation should be applied after resolving overlaps, otherwise
+overlapping points will be reaggregated together incorrectly.
+
+After resolving overlaps, the first step in reaggregation is to
+perform temporal alignment on the desired streams to the desired
+output resolution.
+
+Following overlap resolution and temporal alignment, the next step is
+to remove unwanted attributes.  At this point a number of overlapping
+streams map onto one output stream, and a number of
+temporally-realigned input points have been computed mapping onto one
+output point.  Next, apply the natural aggregation function
+corresponding to the stream point kind, yielding output point values.
+
+### Identifying reaggregated streams
+
+When a new output stream is produced by the reaggregation process, the
+reaggregation processor is responsible for remembering the state of
+its input streams (for correct ovelrap handling) and its start time.
+
+Points written to the output stream use `StartTimeUnixNano` to convey
+an unbroken sequence of observations of the aggregated streams.  When
+the reaggregating processor forgets a stream because of the staleness
+of its inpouts, it should be sure to use a new `StartTimeUnixNano`
+if the stream is ever resumed.
+
+Reaggregating processors are responsible for ensuring that their
+outputs do not overlap.
+
 ## Footnotes
 
 <a name="otlpdatapointfn">[1]</a>: OTLP supports data point kinds that do not
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ release.
 
 ### Metrics
 
+- Add detail about reaggregation for safe attribute removal and to lower temporal resolution. ()
+
 ### Logs
 
 ### Semantic Conventions


specification/metrics/datamodel.md
```diff
@@ -362,6 +362,46 @@ Pending
...
+to lower temporal resolution.
+
+Reaggregation has already been described in specific examples, for
+example, transforming delta into cumulative aggregation temporality,
+and the process of inserting true reset points in cumulative streams.
+
+Reaggregation should be applied after resolving overlaps, otherwise
+overlapping points will be reaggregated together incorrectly.
+
+After resolving overlaps, the first step in reaggregation is to
```
jsuereth, line 379: Nit: Can we have an example of what this looks like / means?  I can do a drawing if you'd like :)

specification/metrics/datamodel.md
```diff
@@ -362,6 +362,46 @@ Pending
...
+### Identifying reaggregated streams
+
+When a new output stream is produced by the reaggregation process, the
+reaggregation processor is responsible for remembering the state of
+its input streams (for correct ovelrap handling) and its start time.
+
+Points written to the output stream use `StartTimeUnixNano` to convey
+an unbroken sequence of observations of the aggregated streams.  When
+the reaggregating processor forgets a stream because of the staleness
+of its inpouts, it should be sure to use a new `StartTimeUnixNano`
```
jsuereth, line 399: nit:  s/inpouts/inputs/

However, it makes me think `inpout` is a way better term than `inout`.  CORBA was wrong.

specification/metrics/datamodel.md
```diff
@@ -362,6 +362,46 @@ Pending
...
+if the stream is ever resumed.
+
+Reaggregating processors are responsible for ensuring that their
```
jsuereth, line 402: I assume this just means time overlap.

Do we want to call out "stream overlap" i.e. understanding whether or not the new metric stream being output from an aggregation processor overrides/conflicts with another stream?

I think implicitly we've already defined that as an "error scenario".

jsuereth (commented):
Would like to see a better example/image, but not a hard ask.  Otherwise a few questions about whether we need to more clearly call out additional identity things.

I think you nailed the important points already, so consider this a non-blocking ask for changes.

specification/metrics/datamodel.md
```diff
@@ -362,6 +362,46 @@ Pending
...
+streams map onto one output stream, and a number of
+temporally-realigned input points have been computed mapping onto one
+output point.  Next, apply the natural aggregation function
+corresponding to the stream point kind, yielding output point values.
+
+### Identifying reaggregated streams
+
+When a new output stream is produced by the reaggregation process, the
+reaggregation processor is responsible for remembering the state of
+its input streams (for correct ovelrap handling) and its start time.
```
victlu, line 394: s/ovelrap/overlap

specification/metrics/datamodel.md
```diff
@@ -362,6 +362,46 @@ Pending
...
+
+Reaggregation has already been described in specific examples, for
+example, transforming delta into cumulative aggregation temporality,
+and the process of inserting true reset points in cumulative streams.
+
+Reaggregation should be applied after resolving overlaps, otherwise
+overlapping points will be reaggregated together incorrectly.
+
+After resolving overlaps, the first step in reaggregation is to
+perform temporal alignment on the desired streams to the desired
```
victlu, line 380: When we are doing temporal alignment, don't we need to account for attribute combinations to know how to proportionally allocate to the correct temporal slot? If so, then, that seems like we need to do safe attribute removal in that stage as well?

Thus, don't we want to do attribute removal first, and then temporal alignment?

specification/metrics/datamodel.md
```diff
@@ -362,6 +362,46 @@ Pending
...
+output resolution.
+
+Following overlap resolution and temporal alignment, the next step is
+to remove unwanted attributes.  At this point a number of overlapping
+streams map onto one output stream, and a number of
+temporally-realigned input points have been computed mapping onto one
+output point.  Next, apply the natural aggregation function
+corresponding to the stream point kind, yielding output point values.
```
reyang, line 388: Not a blocker for this PR, I want to get more perspectives regarding "the natural aggregation function corresponding to the stream point kind" - do we think there will always be a natural aggregation across all dimensions, or it varies depending on which dimensions and combinations we are dealing with?

For example, it makes sense to see the total voltage on the Toyota 4Runner is 12 Volts (6 cells, each 2.0V), but it doesn't make sense to say the total voltage across a 4Runner and F150 is 23.4 Volts:

| voltage (Volts) | the battery cell | make | model | year |
| - | - | - | - | - |
| 2.0 | 1 | Toyota | 4Runner | 2021 |
| 2.0 | 2 | Toyota | 4Runner | 2021 |
| 2.0 | 3 | Toyota | 4Runner | 2021 |
| 2.0 | 4 | Toyota | 4Runner | 2021 |
| 2.0 | 5 | Toyota | 4Runner | 2021 |
| 2.0 | 6 | Toyota | 4Runner | 2021 |
| 1.8 | 1 | Ford | F-150 | 2017 |
| 2.0 | 2 | Ford | F-150 | 2017 |
| 2.0 | 3 | Ford | F-150 | 2017 |
| 1.6 | 4 | Ford | F-150 | 2017 |
| 2.0 | 5 | Ford | F-150 | 2017 |
| 2.0 | 6 | Ford | F-150 | 2017 |

jsuereth: While I think we have a goal of having a natural aggregation function across dimensions (and that's fundamentally why we have Gauge, Sum, Histogram and Aggregation Temporality, etc.), I think you call out a great point that effectively there are things you can do with data that don't make any sense.

I'm viewing our data types the same I do as programming data types:   They are useful, up to a point.   If we figure out an elegant way to express all possible operations, great.  If we only define *some* useful limited operations and semantics for 80% of cases, I think we're ok.

My answer to your question is:   "Are the natural aggregations we've defined useful in MOST (~80%) of scenarios, or do we find we have a lot of caveats?".  I'm still leaning towards "these are useful most of the time".

reyang: I don't have enough understanding of the usage space.
But so far my gut feeling is *"80% time it will be useful, and 20% time it will be confusing"* and that's why I asked the question and also clarified that it is not a blocker.

jmacd:
I will return to this after #1648 and a new PR about temporal alignment merge first.

Issue #1360, opened 2021-01-21
The need for more granularity/clarity in CLIENT span conventions


jkwatson:
**What are you trying to achieve?**

Currently, the span specification reads, when describing span kinds:

```
CLIENT Indicates that the span describes a synchronous request to some remote service. This span is the parent of a remote SERVER span and waits for its response.
```

This seems fine on the surface, but when you get into the details of actually writing instrumentation (and auto-instrumentation in particular), things get murky very quickly. Multiple layers of an application-framework stack can (and do) consider themselves logically as CLIENTs, even if none of them is actually opening up TCP connections and writing bytes to the wire.

For example, let's take a typical java HTTP call. Here are some possible spans that might want to be created by auto-instrumentation:

* JAX-RS HTTP client library makes a GET call
  * Apache HTTP client library makes a GET call
    * Java TCP stack opens up a socket
       * Java TCP stack requests that the OS makes a DNS lookup
       * Java TLS implementation negotiates TLS with the server 

(details here might be off, but I think the idea is clear).

So, which of these operations should create a CLIENT span? If library instrumentation for JAX-RS decides it should be a CLIENT, how can the Apache HTTP client library instrumentation create spans at all, given that a CLIENT has already been created (although it can't tell that, since the parent span is write-only, and the `kind` is immutable after the span has been created!).

At the moment, instrumentation does not have a way to deal with this situation, with the spec as-written. We should solve this problem and provide clarity on how instrumentation should be written in order to handle scenarios like this one.

I will also note, that it becomes even more confusing and difficult when dealing with databases that implement their protocol over HTTP (elastic, for example). Now we have database CLIENT spans that really need HTTP CLIENT spans below them, with all the additional complexity of observing the underlying network stack, as decsribed above.

jkwatson:
Note: see this discussion in the java-instrumentation repository:  https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1822

Oberon00:
Probably we have to live with that. It is impossible for the parent to know if there will be a child, as maybe the child is not instrumented.

>  This span is the parent of a remote SERVER span and waits for its response.

This is not true anyway, since the remote end might not be instrumented at all. Probably it should relaxed to say that it can also be an indirect parent.

reyang:
Related to #110

blumamir:
[...]

Issue #620, opened 2020-05-26
Sampling decison is too late to gain much performance


Oberon00:
# Problem description
Instrumentations are encouraged to provide as many attributes as possible already when starting the span, to allow the sampler to make a better decision. However, collecting these attributes can already be expensive. There is one prominent case where samplers won't even look at these attributes, namely if the sampler respects parent decisions.

# Proposal

I think the spec should contain a mechanism for instrumentations to ask the sampler earlier, before even starting to collect expensive attributes. Two ideas:

1. Languages that use two-step span creation (via a SpanBuilder) could provide a "isRecordable" already on the SpanBuilder that calls the sampler with the information collected so far. A problem with this is that the span name may also be expensive to compute, but that is fixable by making it optional and providing a setName on the SpanBuilder.
2. All languages could provide a isRecordable method on the tracer with some arguments, probably only the parent SpanContext.

# Discussion

A difference to current calls to the sampler with both of these is that the new calls are (and should be!) made before a span ID is computed (whether the trace ID should be filled in from the parent context in cases where there is one is debatable). Nevertheless, I think the existing sampler interface is enough to support these use cases if samplers are written a bit defensively (the probability sampler which does look at the trace ID might be an exception, it would have to return true for invalid trace IDs since it cannot know yet).

EDIT: One issue with calling Sampler.shouldSample twice is that it may lose performance in the sampled case, especially if the sampler adds / computes attributes for its decision. This could be fixed by adding a new method to the sampler so the tracer knows any attributes won't be used. This can also be done backwards-compatible, as the method can be easily substituted with "return true" or "return shouldSample().isSampled()" for samplers that don't support it.

Another advantage of these methods would be that they could also support explicit tracing suppression (as per #530).

Oberon00:
Note that this feature my probably be possible to bolt-on in a way that is API-compatible (on the instrumentation side at least), but resolving this before 1.0 may lead to a cleaner design. For example, we could have completely separate interfaces for purely parent-based sampling and "root sampling". #610 already does this mostly but behind the same interface.

cijothomas:
> Instrumentations are encouraged to provide as many attributes as possible already when starting the span, to allow the sampler to make a better decision.

Where is this suggested? I couldn't find this in specs - My assumption was that, its entirely up to the libraries to decide how much attributes (or none at all) to pass initially (available for Samplers), and then check span.IsRecording before adding more and more (expensive) tags.

arminru:
[...]

Issue #526, opened 2020-03-24
SpanKind with layers


pauldraper:
Network communication is inherently layered, whether with formal OSI layering or more ad-hoc composition.

For example, a typical client-server HTTP call probably will involve the following operations:

```
Client: GetUser
    Client HTTP: GET /users/123
        Client DNS: users.example.com
        Client TCP: 10.0.5.1:15642 -> 189.10.14.10:80
            Server TCP: 10.0.5.1:15642 -> 189.10.14.10:80
                Server HTTP: GET /users/123
                    Server: GetUser
```

These aren't always fixed; the outer RPC layer could decide to retry. Or the HTTP layer could encounter redirects. Nor is are the layers used necessarily known to each other. For example, HTTP2/TCP could be replaced with HTTP3/QUIC.

The same is true of other architectures like message queues, e.g. SQS/HTTP/DNS+TCP. And some libraries/frameworks build even further on top of those.

If I am reading the [SpanKind](https://github.com/open-telemetry/opentelemetry-specification/blob/23dbc05986eec68741f82c79174df85a4379836e/specification/api-tracing.md#spankind) correctly, the semantics have only to do with the parent-child relationships.

How is this intended to work when communication is inherently layered?

Should the RPC instrumenter use SpanKind.CLIENT/SpanKind.SERVER, event though the lower layers may be instrumented?

Should instrumentations have options of what type of SpanKinds to use, and then the user decides which layers are instrumented?

Should RPC instrumentation also use CLIENT/SERVER, and the user just needs to avoid instrumenting lower layers when they are being used by RPC?

---

If we really want this sort feature, there should be optional span attribute msg.kind which is "sender" or "receiver" and msg.protocol which is "tcp" or "http" or "grpc" or "mongodb", and the backend matches them up.

Or, propagate the other span ids in the context inter-process so it can be linked.

Or propagate a flag in the context indicating the middle of a client/server span that prevents lower layers from adding spans.

(Though to be honest, I'm not seeing quite seeing the utility of a two-sided span. It seems tricky to work with and seems to heavily weight towards at-most-once communication paradigms.)

blumamir:
I am experiencing another problem related to span kind, and was happy to find this great issue
In my case, the app produces network spans from 2 sources: 
1. From an instrumented library (in my case, aws-sdk ) which creates the chain aws-sdk -> HTTP -> tls -> net, dns. The tls/net/dns spans are of type "internal" according to the specification.
[...]

Issue #173, opened 2019-07-04
Q: Way to ignore healthcheck traces when using automatic tracer across all languages?


naseemkullah:
Hi,

Will there be a way to ignore healthchecks when using the automatic tracer?

Ideally maybe only trace unusually slow healthchecks?

z1c0:
Hi,
not sure I understand the question correctly. Do you think of a way to exclude e.g. specific URLs? Why only for automatic tracer?

naseemkullah:
Yes it would be nice to exclude specific URLs in the automatic tracer, and also not in the automatic tracer, but just want to make sure this exclusion feature is configurable in the automatic tracer.

discostu105:
I agree that URL exclusions would make sense. We currently have that feature in Dynatrace for exactly that purpose.

In OpenTelementry, the broader question is:

At what level does such configuration need to be applied?
 * API (config is exposed via API level)
 * Tracer-impl (every tracer-implementation can have it's own config system)
 ** Should the OTel SDK have such a config system?
 * Exporter (it's up to the exporter to have a config)

I think config should _not_ be exposed via API. Would add much more complexity. With healthchecks specifically however, it could make sense that a user gives a "hint" about "unimportant" requests (such as healthchecks), so that the Tracer-impl can decide on it's own to exclude them or not.

mtwo:
+1 to this issue. Some OpenCensus implementations had a list of paths that wouldn't get sampled, which were often used to ensure that health checks, exporters, and other telemetry providers weren't generating traces. This was incredibly useful.

tedsuo:
This feels like an SDK detail, but it would be great if there was a way to do this externally: a config file for example.

naseemkullah:
https://github.com/open-telemetry/opentelemetry-js/issues/585 might be of interest

pavolloffay:
Has been any decision made on this?  I have started looking into this in the context of https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1060

From the thread I cannot figure out how the skip URL pattern should work. There are two approaches
1. exclude a span from being created
2. create non-recording (not sampled) span, but respect upstream sampling decision 

The first approach might result in broken traces (e.g. a health check might have an upstream parent or it calls any downstream service/instrumentation). The second approach is more consistent.

Is there a way to instruct the tracer to create non-recording span?

pavolloffay:
from @iNikem 

[...]

PR #1653, opened 2021-04-27
Add Suppress Tracing context key

### LINKED ISSUE #530: Use Context to stop tracing
There are circumstances for instrumented actions where:

1. The action is frequent and of low interest: a healthcheck, polling a message queue, etc.
2. An OpenTracing exporter uses libraries that themselves may be instrumented (risking infinite tracing).
3. If the current layer (e.g. RPC) happens to offer sufficiently detailed tracing, lower HTTP/DNS/TCP/UDP layers do not need to be traced when invoked with this library. This need is heighten by the use of server/client spans, which AFAIK needs to remove spans in between them. (#526).

(1) has come up in #173 

(2) has been a problem in opentelemetry-js (https://github.com/open-telemetry/opentelemetry-js/issues/332) HTTP-based exporters, since the Node.js stdlib is instrumented globally. The solution in that has was adding a special HTTP header `x-opentelemetry-outgoing-request` headers, that the http instrumentation ignores. In essence it uses HTTP headers as a poor-man's context API. (This may not scale to other APIs.)

---

It would be nice to cut off all automatic tracing "below this current scope" by setting a context key. The default tracer implementation would create no-op spans if the context disables spans.

EDIT: Case (1) may call for sampling, rather than full disabling.

dyladan:
Fixes #530 

As discussed in the specification SIG today, this adds a predefined context key for suppressing instrumentation.


specification/context/context.md
```diff
@@ -125,3 +127,26 @@ The API MUST accept the following parameters:
...
+The following are predefined keys that may modify the behavior of multiple signals.
+For each key, there MUST be corresponding methods to set and get its value.
+The key itself MUST NOT be exposed publicly.
+
+### Suppress Instrumentation
+
+In some cases it may be useful to temporarily disable instrumentation.
+For example, this may be used to prevent span exports from being traced and exported,
+or by an instrumentation which wraps a lower-level package which may also be
+instrumented in order to prevent duplicate spans.
```
yurishkuro, line 142: These use cases don't exactly justify this all-or-nothing hammer, rather they point to the need of more fine-grained control. In my company we have a notion of "observability flags", which is a bitmap with different flags controlling different aspects of instrumentation behavior. For example, one of the bits controls verbosity of tracing by allowing spans that represent service calls, but disabling "internal" spans. Should we consider something like that instead of a single Boolean? I am especially skeptical about controlling both metrics and traces with a single flag.

dyladan: For now, I think only tracing really _needs_ to be disabled, but if propagation at least is not disabled the context will be propagated to tracing backends on the actual export call which is a little awkward.

specification/context/context.md
```diff
@@ -125,3 +127,26 @@ The API MUST accept the following parameters:
...
+
+If this key is set to `true`, the following behavior is changed:
+
+- Any `Span` created by `StartSpan` MUST be a non-recording `Span`.
+- `Propagator#Inject` MUST NOT modify the carrier.
```
yurishkuro, line 147: this is a bit ironic since it means the very same new flag won't be passed to downstream

dyladan: This is not a traceflag meant to be propagated downstream. It is meant to prevent in-process instrumentation. The motivating usecase was an exporter that used http causing an infinite loop if the http library used was instrumented. The exporter should be able to suppress instrumentation when it sends data to the tracing backend.

carlosalberto:
@open-telemetry/specs-approvers Take a look at this please.

specification/context/context.md
```diff
@@ -125,3 +127,26 @@ The API MUST accept the following parameters:
...
+- `Propagator#Extract` MUST return the `Context` passed to it as an argument unchanged.
+- `Metrics` MUST NOT be recorded.
+
+If this key is not set it is assumed to be `false`.
+The `get` method for this key MAY return `false` if the key is unset.
```
Oberon00, line 152:
```suggestion
The `get` method for this key MUST return `false` if the key is unset.
```

Or is there anything else it can return?

Oberon00 (dismissed):

specification/context/context.md
yurishkuro: > I think only tracing really needs to be disabled

Yet this is not what the text says, metrics are also disabled.

> but this "hammer" actually serves quite a useful purpose IMO

Yet this purpose is not documented. The use cases that are listed do not need the "hammer".

specification/context/context.md
yurishkuro: Exporter affecting instrumentation is a complete abstraction leak. If the desire is to disable instrumentation within a single process, this can be done by using default impl of the API, which is no-op.

yurishkuro (dismissed):
I am -1 on this change in the current form, it feels rushed to me (lacks extensibility), without a use case clearly documented

yurishkuro:
> The issue and general solution design here are quite old, see #530 

#530 is mostly about stopping tracing, not stopping all instrumentation. This change would make a lot more sense to me if it was about stopping tracing only, but even then a single boolean is a crude design that cannot be extended further.

specification/context/context.md
jmacd: I would put this kind of functionality in an extension package.

In the future, I would prefer that OTel consider specifying a mechanism for setting the current Tracer and Meter in the Context. Then, the desired functionality in this PR can be replaced by updates to the Context-local Tracer and Meter.

This idea was prototyped in https://github.com/open-telemetry/oteps/issues/78

dyladan: > > I think only tracing really needs to be disabled
> 
> Yet this is not what the text says, metrics are also disabled.

I was proposing a change to the text for your feedback.

> > but this "hammer" actually serves quite a useful purpose IMO
> 
> Yet this purpose is not documented. The use cases that are listed do not need the "hammer".

Again, with the proposed change it would be tracing only. What would you think of that?

Why did you not speak up in the spec call? We talked about this for 40 minutes.

specification/context/context.md
dyladan: > Exporter affecting instrumentation is a complete abstraction leak. If the desire is to disable instrumentation within a single process, this can be done by using default impl of the API, which is no-op.

But how do you signal to the instrumentation not to trace the request? The mechanics of how the instrumentation handles that is easy, but knowing which requests to ignore requires some sort of signal.

dyladan:
> > The issue and general solution design here are quite old, see #530
> 
> #530 is mostly about stopping tracing, not stopping all instrumentation. This change would make a lot more sense to me if it was about stopping tracing only, but even then a single boolean is a crude design that cannot be extended further.

I don't understand the need to be able to extend this key in the future. The whole point of having context with opaque keys was to be able to create keys that mean different things. We can make one for tracing and in the future if we want another signal disabled (or all signals) we can just create a new key.

specification/context/context.md
dyladan: No I just thought to leave it up to language maintainers. Some may prefer to know the difference between `false` and `unset`.

dyladan:
@yurishkuro modified to only suppress tracing by creating non-recording spans and preventing injection. Extract and metrics removed from the wording.

blumamir:
I want to add that this mechanism is being used by me in few instrumentation libraries and proven to be very handy.

It is useful to reduce the visual and performance overhead of collecting non-interesting underlying implementation details of high-level libraries such as:
- the HTTP rpc part of aws-sdk requests.
- the HTTP part of elastic-search requests.
- the database driver spans where an ORM is being used (suppress `mongodb` driver spans for operations generated by `mongoose`, suppress `mysql` driver spans for operations generated from `typeorm`, etc).

You can read more about it [here](https://github.com/aspecto-io/opentelemetry-ext-js/blob/master/doc/conventions.md#suppressinternalinstrumentation)

iNikem:
I probably miss something fundamental here, but why #530 cannot be solved by using Samplers? Especially point 1, of health checks and such.

yurishkuro:
> I probably miss something fundamental here, but why #530 cannot be solved by using Samplers? Especially point 1, of health checks and such.

+1

> I don't understand the need to be able to extend this key in the future. The whole point of having context with opaque keys was to be able to create keys that mean different things. We can make one for tracing and in the future if we want another signal disabled (or all signals) we can just create a new key.

My problem is not with the opaque key, but with growing the surface of the API. For comparison, the OpenTracing API had **6 methods in total**: inject, extract, start, finish, setTag, addLog. Here we're proposing two new methods to the public API surface to account for a relatively obscure use case. And we're not even making it extensible so that the cognitive overhead of wider API surface could be amortized for more use cases in the future (back to my original https://github.com/open-telemetry/opentelemetry-specification/pull/1653#discussion_r621544532).

dyladan:
> I probably miss something fundamental here, but why #530 cannot be solved by using Samplers? Especially point 1, of health checks and such.

haha i think i might also be missing something because I can't see how samplers could fix this issue in an automatic way without using context. how does the exporter/spanprocessor signal to the sampler that this is an export span that shouldn't be sampled?

dyladan:
> > I probably miss something fundamental here, but why #530 cannot be solved by using Samplers? Especially point 1, of health checks and such.
> 
> +1
> 
> > I don't understand the need to be able to extend this key in the future. The whole point of having context with opaque keys was to be able to create keys that mean different things. We can make one for tracing and in the future if we want another signal disabled (or all signals) we can just create a new key.
> 
> My problem is not with the opaque key, but with growing the surface of the API. For comparison, the OpenTracing API had **6 methods in total**: inject, extract, start, finish, setTag, addLog. Here we're proposing two new methods to the public API surface to account for a relatively obscure use case. And we're not even making it extensible so that the cognitive overhead of wider API surface could be amortized for more use cases in the future (back to my original [#1653 (comment)](https://github.com/open-telemetry/opentelemetry-specification/pull/1653#discussion_r621544532)).

specification/context/context.md
```diff
@@ -125,3 +127,26 @@ The API MUST accept the following parameters:
 
 The API MAY return a value used to check whether the operation
 was successful or not.
+
+## Predefined Keys
```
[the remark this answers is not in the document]
yurishkuro: I would move it to tracing spec.

specification/context/context.md
```diff
@@ -125,3 +127,24 @@ The API MUST accept the following parameters:
...
+
+In some cases it may be useful to temporarily suppress tracing.
+For example, this may be used to prevent span exports from being traced and exported,
+or by an instrumentation which wraps a lower-level package which may also be
+instrumented in order to prevent duplicate spans.
+
+If this key is set to `true`, the following behavior is changed:
+
+- Any `Span` created by `StartSpan` MUST be a non-recording `Span`.
+- `Propagator#Inject` MUST NOT modify the carrier.
```
yurishkuro, line 147: I would prefer another mechanism for suppressing injections, like the Span itself returning invalid context. This way the new setting only needs to be checked in one place, by the tracer when creating a new span.

yurishkuro (commented):
> i think i might also be missing something because I can't see how samplers could fix this issue in an automatic way without using context.

When creating a context used for export, the exporter starts a span with a name or some other attribute (like instrumentation library, as in #1658) which is recognized by the sampler as "do not trace". While this creates somewhat of a coupling between the exporter and which sampler can be used with the SDK, it does not require any public API changes.

dyladan:
> > i think i might also be missing something because I can't see how samplers could fix this issue in an automatic way without using context.
> 
> When creating a context used for export, the exporter starts a span with a name or some other attribute (like instrumentation library, as in #1658) which is recognized by the sampler as "do not trace". While this creates somewhat of a coupling between the exporter and which sampler can be used with the SDK, it does not require any public API changes.

Hmm that makes sense. I think the context is a better mechanism for it though, even if the sampler is where the context key is eventually read and acted on.

To me that seems quite error prone if users implement any custom components. If a user implements a custom sampler they would need to know to look for this special flag and they would need to know to set it with a custom exporter.

specification/context/context.md
dyladan: AFAIK invalid context already suppresses injections

specification/context/context.md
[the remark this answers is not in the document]
dyladan: Moved

cijothomas:
OpenTelemetry .NET uses this same mechanism already for 2 things:

1. To prevent live loops in Exporters when the exporter itself can be using a instrumented library.  All built-in exporters use it, and we suggest custom exporters to also use this https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/docs/trace/extending-the-sdk/MyExporter.cs#L33-L35

2. For Instrumentations for library which wraps another library that is also instrumented. For example, the GrpcClient .NET is built on top on HttpClient. Users have the ability to suppress the spans from underlying httpclient. The instrumentation for GrpcClient uses this feature to suppress spans from httpclient. (https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.GrpcNetClient/GrpcClientInstrumentationOptions.cs#L30).

dyladan:
> OpenTelemetry .NET uses this same mechanism already for 2 things:
> 
> 1. To prevent live loops in Exporters when the exporter itself can be using a instrumented library.  All built-in exporters use it, and we suggest custom exporters to also use this [open-telemetry/opentelemetry-dotnet@`main`/docs/trace/extending-the-sdk/MyExporter.cs#L33-L35](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/docs/trace/extending-the-sdk/MyExporter.cs#L33-L35)
> 2. For Instrumentations for library which wraps another library that is also instrumented. For example, the GrpcClient .NET is built on top on HttpClient. Users have the ability to suppress the spans from underlying httpclient. The instrumentation for GrpcClient uses this feature to suppress spans from httpclient. ([open-telemetry/opentelemetry-dotnet@`main`/src/OpenTelemetry.Instrumentation.GrpcNetClient/GrpcClientInstrumentationOptions.cs#L30](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.GrpcNetClient/GrpcClientInstrumentationOptions.cs#L30)).

I've heard that there are multiple languages with such implementations. In addition to obviously JS, Python also has this mechanism.

yurishkuro (approved):

specification/trace/api.md
```diff
@@ -14,6 +14,7 @@ Table of Contents
 * [TracerProvider](#tracerprovider)
   * [TracerProvider operations](#tracerprovider-operations)
 * [Context Interaction](#context-interaction)
+  * [Suppress Tracing](#suppress-tracing)
```
reyang, line 17: Do we want individual flags for different signals (e.g. SuppressTracing, SuppressMetrics), or it should be one single flag (e.g. SuppressInstrumentation)?

iNikem:
Now I sense that we have blurred the line about what problem do we try to solve. Because I think different problems ask for potentially different solutions.

First, the problem described in point 1 in #530, about ignoring health checks and other unimportant traces, is exactly what Sampler is meant for. Application developer, who knows what endpoints are not interesting, provides a custom Sampler to ignore them. The main use-case for Sampler, IMOs. This is user-facing problem with already existing user-facing solution.

Second, avoiding loops in SDK, e.g. in exporters. Do we think this is user-facing problem or the one that SDK authors have to solve? Or even auto-instrumentation authors? I assume that all and every SDK has already solved this problem in one way or another, do we need updating API for this?

Second and a half. Avoiding loops in exports for me is a problem similar to the one of suppressing nested CLIENT spans. Should the solutions to these two problems be related as well?

iNikem:
I really want to stress this point of my previous comment: who exactly has still unsolved problem number 2 from #530?

> An OpenTelemetry exporter uses libraries that themselves may be instrumented (risking infinite tracing).

specification/trace/api.md
dyladan: haha the original implementation suppressed everything. It was changed to only suppress tracing based on feedback.

yurishkuro:
> An OpenTelemetry exporter uses libraries that themselves may be instrumented (risking infinite tracing).

the way I understand it is yes, most SDKs probably already solved  that problem, but in a hacky way where the instrumentation of the underlying library is using a secret handshake (special header or context entry) that is understood by the SDK. This creates a fragile coupling since the library instrumentation should be only concerned with OTEL API.

However, using a secret handshake in the form of a context entry may avoid this issue, only creating a coupling between exporter and tracer, while leaving the library instrumentation agnostic to the handshake mechanism. This can be done entirely within SDK, without exposing to the API level.

dyladan:
> Now I sense that we have blurred the line about what problem do we try to solve. Because I think different problems ask for potentially different solutions.
> 
> First, the problem described in point 1 in #530, about ignoring health checks and other unimportant traces, is exactly what Sampler is meant for. Application developer, who knows what endpoints are not interesting, provides a custom Sampler to ignore them. The main use-case for Sampler, IMOs. This is user-facing problem with already existing user-facing solution.

I think you're correct the sampler is the right place to solve this. In most cases it is "too late" to set a context property by the time a user gets a request handler anyway and the span is already generated. The sampler is the only thing early enough to fix this for incoming requests.

> Second, avoiding loops in SDK, e.g. in exporters. Do we think this is user-facing problem or the one that SDK authors have to solve? Or even auto-instrumentation authors? I assume that all and every SDK has already solved this problem in one way or another, do we need updating API for this?
> 
> Second and a half. Avoiding loops in exports for me is a problem similar to the one of suppressing nested CLIENT spans. Should the solutions to these two problems be related as well?

The solution provided here solves both of these things. I do see them as related. In both cases you are trying to suppress the span. Instrumentation should ONLY use the API in order to ensure that they should be used with third party SDKs. In JS we have enforced this quite strictly for our instrumentations and we have no interaction with the SDK not mediated by the API.

Based on his response a minute ago above it seems at least Yuri agrees about the API/SDK separation.

>However, using a secret handshake in the form of a context entry may avoid this issue, only creating a coupling between exporter and tracer, while leaving the library instrumentation agnostic to the handshake mechanism. This can be done entirely within SDK, without exposing to the API level.

Yes, but this then makes it so that instrumentations cannot suppress the underlying transports. For instance, the mongoose instrumentation cannot suppress spans created by the mongodb driver and the aws-sdk cannot suppress spans created by the http library.

iNikem:
> Yes, but this then makes it so that instrumentations cannot suppress the underlying transports. For instance, the mongoose instrumentation cannot suppress spans created by the mongodb driver and the aws-sdk cannot suppress spans created by the http library.

I am not sure that suppressing underlying spans is a right thing to do. E.g. take a look at #1360. I am pretty sure that in most cases we still want information from the underlying transport libraries. `aws-sdk` and `http` being the perfect example. But depending on the existence of the higher level instrumentation transport may want to produce e.g. INTERNAL spans instead of CLIENT or events instead of spans. Also read https://docs.google.com/document/d/1QvWnpYWg-Qj04Sx5NJGfp5rcqb7sMAUgoFM5XmQIf_0/edit#heading=h.111kkbj3xhha

blumamir:
> I am not sure that suppressing underlying spans is a right thing to do. E.g. take a look at #1360. I am pretty sure that in most cases we still want information from the underlying transport libraries

In my opinion, the transport spans can be interesting to see in some cases (debugging low-level issues, performance issues, etc), but most of the time they just add noise to the trace and distract the attention from the logical operation. Allowing the user to turn this instrumentation feature on and off via config option can be very handy to support a wide range of users with different needs.

Before this mechanism was introduced, we had to "clean" the trace in the backend which is possible but more tedious.

specification/trace/sdk.md
```diff
@@ -183,6 +183,8 @@ When asked to create a Span, the SDK MUST act as if doing the following in order
    A non-recording span MAY be implemented using the same mechanism as when a
    `Span` is created without an SDK installed or as described in
    [wrapping a SpanContext in a Span](api.md#wrapping-a-spancontext-in-a-span).
+   If the [`SuppressTracing`](./api.md#suppress-tracing) `Context` flag is set,
+   the newly created `Span` should be a non-recording `Span`.
```
owais, line 187: Can there be a situation where it might be desired to suppress tracing but then re-enable it later in the same trace? For example a chain of 5 middlewares (all instrumented by a common instrumentation) where a user suppresses tracing in middleware 2 and then enable it again in middleware 4. If we generate non-recording spans, the spans from middleware 4 and  5 will not will not be children of the span generated by middleware 2. Not recording any spans when tracing is suppressed instead would allow this case to generate correct traces.

This could also cause issues with context propagation. If tracing is suppressed on a code path that end up making a network call and there is user/instrumentation code that inject trace context into the outgoing call, it'll end up either not injecting context or injecting non-recording span's span ID depending on the implementation of non-recording span.

May be there are other such cases.

If we were to actually suppress tracing, i.e, not record any spans at all instead of non-recording spans, both of these cases would work. Neither solution is perfect  but not recording anything would probably result in traces that are "less broken".

dyladan: >Can there be a situation where it might be desired to suppress tracing but then re-enable it later in the same trace?
>
>If we generate non-recording spans, the spans from middleware 4 and 5 will not will not be children of the span generated by middleware 2. Not recording any spans when tracing is suppressed instead would allow this case to generate correct traces

Sure I think that's a valid usecase. It's definitely not solved by this though for the reasons you stated. It's not easy to solve either because this puts the burden on the caller to not create spans. If an instrumentation doesn't check the context key before calling `startSpan`, then it can't be suppressed.

>This could also cause issues with context propagation. If tracing is suppressed on a code path that end up making a network call and there is user/instrumentation code that inject trace context into the outgoing call, it'll end up either not injecting context or injecting non-recording span's span ID depending on the implementation of non-recording span.

Yes, it was my intention to suppress injection. For the usecases described, I think this is the desired behavior.

>If we were to actually suppress tracing, i.e, not record any spans at all instead of non-recording spans, both of these cases would work. Neither solution is perfect but not recording anything would probably result in traces that are "less broken".

The spec states `startSpan` MUST return something implementing the Span interface, so I don't really see any option other than returning a non-recording span. Open to ideas if you have them.

owais: I wonder if context API could act like a stack in this case. Suppressing instrumentation could store a reference to the active span and undoing it later would re-activate that span again. That'd solve both the cases but it probably bring up more issues and complicates implementation too much.

yurishkuro: context API already works like a stack since context is immutable, you need a new copy for nested scope

owais: > Sure I think that's a valid usecase. It's definitely not solved by this though for the reasons you stated. It's not easy to solve either because this puts the burden on the caller to not create spans. If an instrumentation doesn't check the context key before calling startSpan, then it can't be suppressed.

True but if we make it part of the spec or semantic conventions then it's very likely they that they will.

> context API already works like a stack since context is immutable, you need a new copy for nested scope

Yea, I meant when suppression is turned off, it could "pop" all the contexts pushing while the suppression was on, resulting in the last known non-suppressed span becoming active again. In reality suppressing tracing could just store a reffernece to the active span on the context with a different key and once suppression is disabled, it could activate that span if found in current context.

weyert:
Interesting issue, this might help solving my long standing issue at work were a lot of spans are created for RPC and HTTP by instrumentation libraries spans while I don't really care about those in some contexts, when calling a specific function/action but do care about in all other cases.

For example, I am using Datastore which uses RPC for it calls means I am having spans for dns (for the host), RPC connect, my database client span, my database sql builder span, etc. In this case I want to suppress the RPC connect, NET connect, DNS connect  spans but only in one case were I doing many calls to Datastore or when I am calling many underlaying REST services.

Oberon00:
After reading all the feedback on samplers, I think this could indeed also work: We provide, in the SDK, a Sampler `SuppressingSampler(innerSampler)` that checks the context key and a corresponding **SDK** method to set the flag on the context. Then we increase only the SDK's API surface but not the API's API surface. The major disadvantage though is that performance will be worse because initial span attributes will still be collected but that is #620.

Oberon00:
> Interesting issue, this might help solving my long standing issue at work were a lot of spans are created for RPC and HTTP by instrumentation libraries spans

That use case might be better solved by providing samplers access to a readable parent span (I think this already works in some languages with some casting, since the sampler already gets the full parent context)

specification/trace/sdk.md
dyladan: >True but if we make it part of the spec or semantic conventions then it's very likely they that they will.

In my experience this is not true. Instrumentations created by SIGs will, but the community is very unlikely to read the spec that closely.

dyladan:
I would like to avoid instrumentations taking on the SDK as a dependency since this is essentially the primary reason for the strict API/SDK split. If a library author builds opentelemetry into their library they should only have to depend on the API

pauldraper:
> Here we're proposing two new methods to the public API surface to account for a relatively obscure use case

To quote #530 

> There are circumstances for instrumented actions where:
>
> 1. The action is frequent and of low interest: a healthcheck, polling a message queue, etc.
> 2. An OpenTracing exporter uses libraries that themselves may be instrumented (risking infinite tracing).
> 3. If the current layer (e.g. RPC) happens to offer sufficiently detailed tracing, lower HTTP/DNS/TCP/UDP layers do not need to be traced when invoked with this library. This need is heighten by the use of server/client spans, which AFAIK needs to remove spans in between them. (#526).
>
> (1) has come up in #173
>
> (2) has been a problem in opentelemetry-js (open-telemetry/opentelemetry-js#332) HTTP-based exporters, since the Node.js stdlib is instrumented globally. The solution in that has was adding a special HTTP header x-opentelemetry-outgoing-request headers, that the http instrumentation ignores. In essence it uses HTTP headers as a poor-man's context API. (This may not scale to other APIs.)

Please address alternative solutions to these issues.

The problem is so non-obscure that Python and Ruby implementations independently created this solution over a year ago.

dyladan:
@pauldraper Are you asking me to address alternatives?

pauldraper:
Sorry, @yurishkuro

dyladan:
Discussed this in the spec SIG today. This PR boils down to 3 use-cases

1. Prevent infinite loop with exporters
2. Allow instrumentations to suppress lower level instrumentations
3. Suppress frequent noise like polling and health checks

### Prevent Exporter Loops

This can be solved in the SDK only with no API change and would not require this API to exist.

### Instrumentations Suppress Lower Levels

This is a special case of a larger question: how should we handle cases where an instrumentation wants to be the "final" span? This PR suggests that they be suppressed and not traced, but there may be other solutions to this problem.

### Suppress Noisy Spans

In some ways this is an extension of (2), but was not my original target in this PR and is possibly better solved with other mechanisms like smarter sampling.

## Takeaway

For now, the specification SIG suggested that JS SIG should move this functionality out of the API into some sort of API extensions package. Instrumentations can rely on this package which our SDK will respect, but third party SDKs will be under no official obligation to respect.

I am going to leave this open because I still believe the mechanism has value and I have not yet seen a proposal for solving (2) that is obviously better than this. The linked issue is currently one of our oldest open issues and having multiple languages already having implemented this on the SDK level seems to me to show the value in this mechanism.

pauldraper:
> This can be solved in the SDK only with no API change and would not require this API to exist.

How?

Example situation:

gRPC module uses the http module.

The gRPC module is instrumented. The http module is instrumented.

The gRPC instrumentation creates a client span. The http module creates a client span.

Client spans are now doubled up which AFAIK is a problem for systems with two-sided spans like Zipkin.

To prevent doubled-up client spans, the gRPC instrumentation needs a way to tell the http instrumentation to not trace.

---

This is a very common situation.

Am I missing something?

jkwatson:
> > This can be solved in the SDK only with no API change and would not require this API to exist.
> 
> How?
> 
> Example situation:
> 
> gRPC module uses the http module.
> 
> The gRPC module is instrumented. The http module is instrumented.
> 
> The gRPC instrumentation creates a client span. The http module creates a client span.
> 
> Client spans are now doubled up which AFAIK is a problem for systems with two-sided spans like Zipkin.
> 
> To prevent doubled-up client spans, the gRPC instrumentation (which uses only the API) needs a way to tell the http instrumentation (which uses only the API) to not trace.
> 
> This is a very common situation.
> 
> Am I missing something?

Your example isn't an exporter loop, which is where you snipped your comment from.

Oberon00:
> To prevent doubled-up client spans, the gRPC instrumentation (which uses only the API) needs a way to tell the http instrumentation (which uses only the API) to not trace.

This is only one possible solution. Another theoretically (not currently) possible solution is that the SDK user / application owner configures the SDK to suppress any local child spans of spans with type=CLIENT and having an `rpc.system=grpc` attribute. Or any CLIENT spans that are children of another CLIENT should be suppressed.

IMHO, having the grpc instrumentation decide whether or not the child spans should be suppressed is wrong. The HTTP child spans are potentially useful, e.g., if you have a specific HTTP-related issue you want to debug.

pauldraper:
> Your example isn't an exporter loop, which is where you snipped your comment from.

Ah, yes. My bad. Regardless, I would like to understand the group's answer to that.

> IMHO, having the grpc instrumentation decide whether or not the child spans should be suppressed is wrong. The HTTP child spans are potentially useful, e.g., if you have a specific HTTP-related issue you want to debug.

FWIW I agree, I think client/server spans are terrible. Zipkin two-sided spans are terrible. Protocols are inherently nested and Otel seems to ignore that (#526).

But, as the approach stands today, it seems that double-client spans are Bad.

carlosalberto:
@dyladan As the related issue will still exists, are you ok with closing this PR?

dyladan:
I don't mind if we close this issue as long as we eventually solve this use-case. The issue has been around for over a year with no movement

carlosalberto:
Oh definitely, let's keep #530 open until we _finally_ solve it for all SIGs in an uniform way. Thanks!

PR #1654, opened 2021-04-27
Get Tracer operation should use an empty string if the specified `name` is null

### LINKED ISSUE #1619: Consider reverting/amending #1534 (removes fallback tracer name)
#1534 changed the spec to require null tracer names to be kept intact. Unfortunately, this ends up introducing a pretty significant behavioral breaking change to the Java SDK interfaces for exporters. It would require our `InstrumentationLibraryInfo` to return a `nullable` name attribute, which means that all exporters (both ours and any 3rd party) now have to guard against a null attribute, or risk throwing NullPointerExceptions in the exporter.

See this Java PR for the impact within the java repository itself: https://github.com/open-telemetry/opentelemetry-java/pull/3153

I think that we should amend this part of the specification to note that nulls should be converted into empty strings, or we should revert that change and specify an actual fallback value.

carlosalberto:
Fixes #1619 - as discussed previously in a previous SIG call.

The `Get Tracer` operation, upon receiving an invalid value, should set the `name` to an empty string, instead of allowing null values to pass through.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ release.
 
 ### Traces
 
-- `Get Tracer` should use an emptry string if the specified `name` is null. ([#1654](https://github.com/open-telemetry/opentelemetry-specification/pull/1654))
+- `Get Tracer` should use an empty string if the specified `name` is null. ([#1654](https://github.com/open-telemetry/opentelemetry-specification/pull/1654))
 
 ### Metrics
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ release.
 
 ### Traces
 
+- `Get Tracer` should use an emptry string if the specified `name` is null. ([#1654](https://github.com/open-telemetry/opentelemetry-specification/pull/1654))
+
 ### Metrics
 
 ### Logs
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -110,7 +110,7 @@ This API MUST accept the following parameters:
   or application.
   In case an invalid name (null or empty string) is specified, a working
   Tracer implementation MUST be returned as a fallback rather than returning
-  null or throwing an exception, its `name` property SHOULD keep the original invalid value,
+  null or throwing an exception, its `name` property SHOULD be set to an **empty** string,
   and a message reporting that the specified value is invalid SHOULD be logged.
   A library, implementing the OpenTelemetry API *may* also ignore this name and
   return a default instance for all calls, if it does not support "named"


yurishkuro (approved):

cijothomas (approved):

jmacd (approved):

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ release.
 
 ### Traces
 
+- `Get Tracer` should use an emptry string if the specified `name` is null. ([#1654](https://github.com/open-telemetry/opentelemetry-specification/pull/1654))
```
jkwatson, line 17:
```suggestion
- `Get Tracer` should use an empty string if the specified `name` is null. ([#1654](https://github.com/open-telemetry/opentelemetry-specification/pull/1654))
```

jkwatson (approved):

anuraaga (approved):

specification/trace/api.md
```diff
@@ -110,7 +110,7 @@ This API MUST accept the following parameters:
   or application.
   In case an invalid name (null or empty string) is specified, a working
   Tracer implementation MUST be returned as a fallback rather than returning
-  null or throwing an exception, its `name` property SHOULD keep the original invalid value,
+  null or throwing an exception, its `name` property SHOULD be set to an **empty** string,
```
ThomsonTan, line 113: Should `TracerProvider::GetTracer(null)` and `TracerProvider::GetTracer("")` return the same `Tracer` with `""` as name?

iNikem (approved):

specification/trace/api.md
Oberon00: Identity of the returned tracer is not relevant to the spec, as the only per-tracer state is the (immutable) Instrumentation Library name and version.

Oberon00 (approved):

arminru (approved):

specification/trace/api.md
ThomsonTan: Thanks. I am still a little bit unclear about "its name property SHOULD be set to an empty string. Does it mean the name property of the tracer should be set to **empty** string even if `null` is passed to `GetTracer`, but keep the name stored in Instrumentation Library as `null` which is the original value passed to `GetTracer`?

carlosalberto: Probably we need a further clarification, but the idea is that `Get Tracer (null)` essentially becomes a call to `Get Tracer ("")`.

ThomsonTan: `Get Tracer(null)` becomes a call to `Get Tracer("")` is very clear to me. I think further clarification will be helpful to guide implementations.

Issue #1298, opened 2020-12-18
Adding resource attributes post-creation (e.g. via auto-discovery)


vovencij:
**What are you trying to achieve?**

I would like to have a possibility to add resource attributes via some auto-discovery mechanism, where attribute values are not known at the Resource creation time, but rather become known during application initialization. In case of auto-instrumentation Otel SDK is created very early in process runtime. Some useful information about the process may not be accessible yet. 

An example may be a process of discovering an app server name and its version (https://github.com/open-telemetry/opentelemetry-specification/issues/1143) by instrumentation, rather than setting these values via environment. Even without instrumentation, discovering and setting a version of an application (or whatever component that makes sense at the `Resource` level) at runtime is much easier than obtaining it before an application starts (as is needed for setting attributes via environment).

Oberon00:
I think there are several possible routes we can go here:

* Allow adding attributes to resources at any time (or rather, merging new resources with Tracer/MeterProvider at any time). Removing or changing attributes should IMHO not be allowed.
* Add a second "dynamic" resource that can be changed at any time, keeping the original resource as an actually immutable potion. This may be useful for backends/caching/etc. 
* There may be another kind of resource required, for things like pod labels, IP addresses, that can actually change during runtime. But this is not really related to this issue.
* Alternatively, report things that don't work well as span attributes, as suggested in https://github.com/open-telemetry/opentelemetry-specification/pull/1293#discussion_r544580934 and as proposed in https://github.com/open-telemetry/opentelemetry-specification/pull/1265

Oberon00:
Note, this was also briefly discussed in https://github.com/open-telemetry/opentelemetry-specification/pull/515#discussion_r393752660

anuraaga:
I've always wished `Resource` is an interface to allow implementing a dynamic one. Some things are resolved lazily, or change very rarely, but not never, being able to use `Resource` for them would be nice.

Oberon00:
Let's try to not get too much into implementation details of how to allow changing resources (interface / struct / sync / async / auto-detected / explicit) at first and maybe discuss change in semantics and impact on backends/collectors/maybe exporters.

I think this has the potential to break assumptions and optimizations. The resource is supposed to capture identification information about the entity (runtime/process/container/...) producing telemetry. It would be very annoying if the backends identifies the resource as belonging to different entities over the lifetime of a telemetry producer just because some identification information is missing.

[...]

Issue #734, opened 2020-07-24
Should there be a "prepared span" construct in the API?


yurishkuro:
Many span attributes are "static", i.e. they never change for the given class of spans. Span name is one such attribute, but there are many others from semantic conventions, e.g. `http.method` or `db.type`. Right now the trace API does not provide any facilities for optimization, requiring setting all those attributes on the spans over and over, on the critical path of the requests.

A possible solution to this is the concept of prepared span, somewhat similar to "bound instrument" in the metrics API, or "prepared statement" in the database APIs, e.g.
```
func NewDBMiddleware(traceProvider, dbDriver) *Middleware {
    tracer := traceProvider.GetTracer("me, the middleware")
    return &Middleware{
        preparedSpan: tracer.PrepareSpan(
          "db-query", 
          Attribute{"db.type", "MySQL"},
          Attribute{"db.connection", dbDriver.Connection.String()},
       )
    }
}

func (m *Middleware) Execute(query, next) {
    span := m.preparedSpan.Start()
    span.SetAttribute("db.statement", query)
    defer span.FinishI()

    next(query)
}
``` 

Preparing the span allows the SDK to optimize how the span is stored in memory and exported to the backend, e.g. static attributes can be either encoded more efficiently or not sent with each span at all.

Oberon00:
Note that we had (still have) the same problem with a hypothetical `http.app` attribute (#335) where @SergeyKanzhelev also noted that it would be the same most of the time in https://github.com/open-telemetry/opentelemetry-specification/pull/263#discussion_r329281769. Which lead me to create #274 to allow resources as span attributes. This would solve the same problem, but instead of preparing a span and then finishing it, you would prepare a resource and the set it as a resource attribute.

@jmacd has a more complex proposal to solve this problem at open-telemetry/oteps#78. @tigrannajaryan also suggested there should be some way to have more than one resource per span in #579.

Oberon00:
@open-telemetry/specs-approvers @open-telemetry/specs-trace-approvers Should we create some label like "shared-span-data" to track all these issues?

bogdandrutu:
When I read this I immediately thought about the Builder being possible to be reused (because it is kind of that intermediate state). Not sure if that is the right approach but would like us to think about the relationship between this proposal and the current Builder (in some languages that use this pattern).

yurishkuro:
[...]

Issue #579, opened 2020-05-04
Define a way to specify more than one Resource in telemetry data


tigrannajaryan:
Currently it is possible to refer to a Resource in spans and metrics. The Resource indicates where the signal was generated.

However, we may need a way to reference the second (or third, etc) resources.

For example an application A (Resource 1) may report a metric such as "request count" and it may be the number of requests that application A made to application B (Resource 2). Thus, this requires that the metric references multiple resources.

See discussion context here: https://github.com/open-telemetry/oteps/pull/97#discussion_r418747530

Oberon00:
See also issue on HTTP app reporting (or per span service resource depending how you look at it): https://github.com/open-telemetry/opentelemetry-specification/issues/335
Also related: https://github.com/open-telemetry/opentelemetry-specification/issues/274 "Allow resources as span attributes"

jmacd:
Related content here: https://github.com/open-telemetry/oteps/issues/78

To support resources as an API-level concept, as in OTEP 78, requires protocol support for "more than one Resource" as you say.

tigrannajaryan:
One new use case for this is eBPF. Network monitoring collector needs to record telemetry about 2 entities communicating, each entity can be one of the Resource types that we record using semantic conventions (e.g. network traffic between 2 pods). It is currently not possible to do since there can be only one set of semantic conventions per type of entity recorded in the Resource. Even if we could record 2 sets of conventions in the Resource it doesn't seem to the right approach, since the Resource is supposed to be the indicating the origin of the telemetry.

duxing:
@tigrannajaryan totally agree with what you said. any application that has network IO to other entities (service, cache, db, etc) should really reflect the traces with 2 entities per communication. one for the client (self, source of network io) and one for the client's observation of server entity (destination of network io).

danielgblanco:
@tigrannajaryan we're assuming you're the sponsor for this, but please do let us know if that's not the case. Thanks.

Issue #508, opened 2020-03-10
Change Processor and Exporter interface to pass resource


rghetia:
In current [resources](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/sdk-resource.md) specification, resource is associated with a Trace Provider. Semantically, that makes sense as well. However, there is problem retrieving resource in a SpanProcessor. SpanProcessor has no reference to TraceProvider to which it is registered. The problem is only in async path.

In Go (and most likely other languages as well), when span ends, the data flow is as follows.
Sync Path: Tracer (ctx1) -->Simple Span Processor (ctx1, SpanData) --> Exporter (ctx1, SpanData).
Async Path: Tracer (ctx1)-->Batched Span Processor (ctx1, SpanData)  --> Exporter (ctx2, []SpanData)

In sync path, there are no issues. Resources can be added to SpanData and Exporter can send the resource to its backend after remapping the resource if necessary.

In async path, however, it is not efficient to duplicate the same resource with all the SpanData. After all we have common [resource](https://github.com/open-telemetry/opentelemetry-proto/blob/674ae1c8703f16925d59603591c23cad0a90b8fd/opentelemetry/proto/trace/v1/trace.proto#L31) for a batch of spans in proto for a reason. How do we achieve the efficiency?

- Group all SpanData on BSP by resources. It requires
  - [OnEnd(SpanData)](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/sdk-tracing.md#onendspan) interface to change to onEnd(Resource, SpanData) to contain reference to the Resource. Batcher can simply keep one reference for all the accumulated SpanData. 
  - [Export(Batch)](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/sdk-tracing.md#exportbatch) must be changed to Export(Resourc, Batch)  to contain reference to the Resource. 
  - Restrict BSP to be associated only with one Trace Provider (current spec doesn't restrict associating BSP with multiple Trace Provider).
    - If it must be allowed to associate with Multiple Provider then use reference to Resource (assuming same resource object will be provided with each onEnd() call) to group SpanData by Resource. Then call Export(Resource, []SpanData) for each group.

Alternatively, 
- Create resource at Exporter.

bogdandrutu:
In Java SpanData contains the Resource, but that can change if we decide to implement this. Also same discussion should apply for InstrumentationLibrary.

SergeyKanzhelev:
[...]

PR #1658, opened 2021-04-28
Add Resource to ShouldSample.

### LINKED ISSUE #1588: Instrumentation library info and Resources are not available to Samplers
ShouldSample receives as input all the known information about the Span to be created, see https://github.com/open-telemetry/opentelemetry-specification/blob/v1.1.0/specification/trace/sdk.md#shouldsample, *except* the instrumentation library info and the Resources.

Suppressing unwanted spans depending on the instrumentation library name was one of the original motivations of introducing it in the first place, see https://github.com/open-telemetry/oteps/blob/main/text/0016-named-tracers.md#solution

> [...] a Sampler could be implemented that discards Spans or Metrics from certain libraries

Adding an overload of ShouldSample that accepts these additional arguments (or an object with these additional properties) and providing a default implementation that forwards to the currently specified implementation should be possible at least in C++, Java, Python.

Oberon00:
Fixes #1588

## Changes

- Add Resource as Sampler argument.


iNikem (approved):

specification/trace/sdk.md
```diff
@@ -206,6 +206,11 @@ Returns the sampling Decision for a `Span` to be created.
 * Collection of links that will be associated with the `Span` to be created.
   Typically useful for batch operations, see
   [Links Between Spans](../overview.md#links-between-spans).
+* `Resource` of the `Span` to be created.
```
iNikem, line 209: How can we make this change in backward compatible way?

Oberon00: In Java it is trivial to do so by adding the new method with extended parameters as a default method to the interface, and the implementation calls the old method, ignoring the additional parameters. The same works for other languages that allow default implementations in interfaces (only C++ comes to mind).

A solution that will work in all object-oriented languages is to have a new interface (Sampler2) and have the `SetSampler` (or equivalent) API accept either the new or the old interface.

carlosalberto (approved):

carlosalberto:
@open-telemetry/specs-approvers please review this.

Worth mentioning this change again in the next maintainers meeting, and also worth having a prototype for a language that does not support default implementation/parameters.

specification/trace/sdk.md
jkwatson: I think in Java we would want to make *both* methods have default implementations, since otherwise those who only wanted to use the new method would have to also implement the old-one as a no-op, which would be a pretty annoying dev experience. But, that will also make it somewhat confusing to Sampler implementors about what they're supposed to do. I guess documentation is the key.

yurishkuro: Passing the resource, which is not supposed to change during Sampler's life cycle, is a constant unnecessary performance overhead. Why can't it be given to sampler during sampler construction?

specification/trace/sdk.md
```diff
@@ -206,6 +206,11 @@ Returns the sampling Decision for a `Span` to be created.
...
+* `InstrumentationLibrary` (name and version) of the `Span` to be created.
```
yurishkuro, line 210: Instrumentation library is given to tracer on construction, and presumable included in the spans this tracer creates. I think it's better to make it retrievable from the span rather than passing it to sampler as argument (which adds constant perf overhead).

Oberon00: >  I think it's better to make it retrievable from the span rather than passing it to sampler as argument (which adds constant perf overhead).

This is not a "rather than". We need to do both. The span does not exist when ShouldSample is called.

That said two ideas how this could be implemented instead:

1. Pass a `ReadableSpanBuilder` to the sampler, that should cover all existing parameters.
2. Instead of passing Resource and InstrumentationLibrary separately, pass a `ReadableTracer` to ShouldSample.

I'd like to leave this choice up to the SIGs.

Oberon00: We had this discussion with exporters already and I think the conclusion was that an Exporter must in general be prepared for each span to have a different resource, because the same exporter might be attached to different SpanProcessors and the same SpanProcessor to different TracerProviders with different resources. I think the same (minus the intermediate SpanProcessor) applies to the sampler.

I would have been all for adding a restriction to the spec to forbid this, but it might be too late now.

> which adds constant perf overhead

That is really neglegible. Depending on how you implement it, it might boil down to a single assembly instruction and a single pointer stack space.

specification/trace/sdk.md
Oberon00: We had this discussion with exporters already and I think the conclusion was that an Exporter must in general be prepared for each span to have a different resource, because the same exporter might be attached to different SpanProcessors and the same SpanProcessor to different TracerProviders with different resources. I think the same (minus the intermediate SpanProcessor) applies to the sampler.

I would have been all for adding a restriction to the spec to forbid this, but it might be too late now.

> which adds constant perf overhead

That is really neglegible. Depending on how you implement it, it might boil down to a single assembly instruction and a single pointer stack space.

iNikem (approved):

spec-compliance-matrix.md
```diff
@@ -74,6 +74,7 @@ formats is required. Implementing more than one format is optional.
 | [Sampling](specification/trace/sdk.md#sampling)                                                  |          |    |      |    |        |      |        |     |      |     |      |       |
 | Allow samplers to modify tracestate                                                              |          | +  | +    |    | +      | +    | +      |     | +    |     | -    | +     |
 | ShouldSample gets full parent Context                                                            |          | +  | +    | +  | +      | +    | +      |     | +    | +   | -    | +     |
+| ShouldSample gets Resources and InstrumentationLibrary                                           |          |    |      |    |        |      |        |     |      |     |      |       |
```
bogdandrutu, line 77: I would like to be able to implement this differently. Maybe we can actually keep ShouldSample the same and apply the same stricture as "Tracer" for "Sampler":

* Have a "SamplerProvider" with a "getSampler(InstrumentationLibrary).
* Have a "SamplerProviderFactory" that has a "getSamplerProvider(Resource)"

The reason to do this is because then Sampler can be optimized to pre-compute things for Resource/InstrumentationLibrary.

Is this reasonable?

bogdandrutu (changes requested):

spec-compliance-matrix.md
bogdandrutu: Also the need for Resource is very unclear since a Sampler is installed in a Provider which has immutable Resource, any filtering/sampling logic that applies to Resource can be evaluated before installing

Oberon00: > Also the need for Resource is very unclear since a Sampler is installed in a Provider which has immutable Resource

This is the same concern that @yurishkuro had and I answered it at https://github.com/open-telemetry/opentelemetry-specification/pull/1658#discussion_r627143176

Oberon00: That SamplerProviderFactory + SamplerProvider design sounds rather complex to me. If we are concerned about sampling performance we should start with #620 😃 Returning an AlwaysOff sampler for certain instrumentationlibraries from the SamplerProvider might be a usecase.

Regarding backwards compatibility, I don't see this as an advantage, as you can have the same backwards-compat behavior by adding a SamplerV2 interface as alternative to the current sampler.

I think it's a reasonable alternative, but I don't like it very much, as introducing two new classes seems more complex than two new parameters.

specification/trace/sdk.md
bogdandrutu: The exporter is very different in the way that it does not run on the critical path, hence the small overhead is negligible compare with this. I would say that indeed consistency would be good, but we need to be careful with the overhead in this case.

spec-compliance-matrix.md
bogdandrutu: If we don't expose the Resource at all in the Sampler.ShouldSample, samplers that need resource will be forced to evaluate that at the beginning when available during construction of the Sdk provider. So I think we can at least do a good thing and not add the Resource, maybe just the Instrumentation Library since that is a smaller object.

specification/trace/sdk.md
Oberon00: One more point: Currently, it should be fine in most language implementations to call GetTracer each time when creating a span. Some languages may even require this, as calling GetTracer in a static initializer or similar may be executed before the SDK is configured and yield you a no-op tracer. With this change, we would make GetTracer call a "plugin interface", so the call can become arbitrarily expensive. That cost would then be recovered because StartSpan is cheaper by that, but still, it would have some more implications.

bogdandrutu:
@Oberon00 let's split this into 2 PRs to allow progress:
1. that adds the InstrumentationLibrary which we've already agreed to support sampling by instrumentation library, then when we find a very good use-case for Resource we can discuss, before that point as pointed in the issue an implementation can be done anyway.
2. The second PR will be when we have good use-case for Resource and prove that the alternative does not work.

iNikem:
> @Oberon00 let's split this into 2 PRs to allow progress:
> 
> 1. that adds the InstrumentationLibrary which we've already agreed to support sampling by instrumentation library, then when we find a very good use-case for Resource we can discuss, before that point as pointed in the issue an implementation can be done anyway.
> 2. The second PR will be when we have good use-case for Resource and prove that the alternative does not work.

@Oberon00 Do you agree with this suggestion?

specification/trace/sdk.md
Oberon00: @yurishkuro This was now merged without any opposition on #1850, so I'm resolving this.

spec-compliance-matrix.md
Oberon00: @bogdandrutu  We now went the simpler route in #1850, so I think we can do the same for Resources. Please re-open the conversation if you disagree.

specification/trace/sdk.md
Oberon00: Backwards-compatibility is not more of an issue with Resources than with InstrumentationLibrary, which was merged in #1850.

Oberon00:
I re-opened this PR since we are "breaking" compatibility anyway through #1850. @anuraaga said he had a use case for resources with the X-Ray sampler: https://github.com/open-telemetry/opentelemetry-specification/issues/1588#issuecomment-893245998

anuraaga (approved):
Thanks @Oberon00. I agree that there is no conceptual difference between InstrumentationLibrary and Resource in terms of this method signature. Also, there should be no real extra "cost" to adding two parameters to the method vs just one in a single release.

For reference, Resource is associated with a `TracerProvider`
> When used with distributed tracing, a resource can be associated with the TracerProvider when the TracerProvider is created. That association cannot be changed later. When associated with a TracerProvider, all Spans produced by any Tracer from the provider MUST be associated with this Resource.

And Sampler is part of TracerProvider

> Configuration (i.e., SpanProcessors, IdGenerator, SpanLimits and Sampler) MUST be managed solely by the TracerProvider and it MUST provide some way to configure all of them that are implemented in the SDK, at least when creating or initializing it.

I think this means it is the TracerProvider's job to provide the Resource information to the Sampler - it is the holder of all this configuration. And in the end, any alternative results in an awkward user experience

```java
Resource resource = createResource();
AwsXrayRemoteSampler sampler = AwsXrayRemoteSampler.create(resource);
TracerProvider tracerProvider = SdkTracerProvider.builder()
  .setResource(resource)
  // Passed resource above, why also pass to sampler? What if it's out of sync? etc
  .setSampler(sampler)
  .build();
```

iNikem:
> I think this means it is the TracerProvider's job to provide the Resource information to the Sampler

How exactly should this happen?

Oberon00:
I think @anuraaga means that the tracerprovider (indirectly) calls ShouldSample with the resource as argument in the end.

trask (approved):

Oberon00:
@bogdandrutu Can you please check if your "Request changes" still applies?

arminru (approved):

jkwatson:
I'm against this change. Anyone who wants to write a sampler that uses the Resource can provide a constructor (or whatever is language-equivalent) for that Sampler and it can use the Resource however it wants. I'd rather not add more stuff to the sampler API that we need, and I don't see how this is a necessary addition.

jkwatson:
> > Resource can provide a constructor (or whatever is language-equivalent) for that Sampler and it can use the Resource however it wants
> 
> I think this creates an awkward user experience - the SDK already has a resource registered, registering it again for the Sampler, but not for the export pipeline, adds some tedium and isn't symmetric. Given the way the APIs are structured, I'd expect Sampler and Exporter to act similarly (just use the Resource I passed into the SDK). This is particularly true given for many users they won't even create a Resource - some sort of autoconfiguration would set it up. It's unnecessary to expose this autoconfigured Resource to the user just to register a Sampler.

I don't deny this might be a bit awkward, but I also know that users can use this technique today, with no spec changes necessary to implement it. Do we have more than one use-case for this?

anuraaga:
> Do we have more than one use-case for this?

Hmm - it's hard to word it, but I think that there indeed may only be one precise use case. It is the concept of "sampling by attribute" - the attributes for a span are both what's set in its builder and its Resource I think. So we could never consider attribute-based sampling complete without the Resource. We currently have a PoC of attribute-based sampling for Java

https://github.com/open-telemetry/opentelemetry-java-contrib/pull/70

which I would suspect to be spec'd in some form some day. Currently this sampler can't look at Resource but it seems like it should be able to.

The attribute sampler would likely be a primitive type used by more complex sampling - does each of them count as a use case? Then we'd probably have several :P Providing the primitive itself (the ability to sample on a span's resource attributes) seems fundamentally important and worth making less awkward for the user. I don't know if I'd want to add a API addition twice for it, which is why hoping to bundle this into the same release as the InstrumentationLibrary update - in that case, it seems like a minor addition with a good gain in functionality.

Oberon00:
I think a good argument just came to my mind:
If we pass the resource in the sampler constructor, we close the door to allowing multiple different resources per TracerProvider. Allowing different resources e.g. per Tracer is something that has been discussed in the past e.g. in #579 and related discussions in #335, #734. Passing the resource in the constructor may also decrease implementation flexibility for post-creation resources / resource attributes (#1298).

iNikem:
Just raising my voice that we definitely should not close the door for #1298

Oberon00:
@bogdandrutu As you still have "Changes Requested", what is your take on this?

carlosalberto:
> I'm against this change. Anyone who wants to write a sampler that uses the Resource can provide a constructor (or whatever is language-equivalent) for that Sampler and it can use the Resource however it wants. I'd rather not add more stuff to the sampler API that we need, and I don't see how this is a necessary addition.

This (written by @jkwatson ) was the _general_ feeling during a SIG call a few weeks ago (hence the comment), namely: do not add stuff to the API at all, unless absolutely necessary. 

@bogdandrutu would you mind answering the latest's answers on this topic?

Oberon00:
I agree that adding stuff to the API (the SDK's plugin API in this case) should be minimized generally. But one goal behind this is IMHO keeping the implementation flexible, allowing later changes, by not exposing too many internals.
But in this case, by *not* providing the Resource, we are forcing sampler developers that need the resource to make the assumption that the resource is something static that will always be the same for all spans that the sampler processes.
Crucially, this *contradicts* the fact that the readable span ("SpanData") has the Resource as a public property, which would allow the assumption that different Spans provided to the same SpanProcessor (exporter) can have different resources.
In effect, exporters need to go through the trouble of preparing for distinct resource objects across a batch or different batches, while samplers (configured on the same level as spanprocessors) have to be coded with the limitation that all spans have the same resource. It follows, that at least one of the following is currently broken:
* A sampler that needs a Resource
* Different resources within the same TracerProvider (see https://github.com/open-telemetry/opentelemetry-specification/pull/1658#issuecomment-912297271)
* Reusing the same sampler instance across different TracerProviders

I think the last one doesn't matter much (though we seem to support it for spanprocessors/exporters). But the first and second one are a problem.

bogdandrutu:
@Oberon00 I think exposing the Resource in the SpanData was a mistake that hurts a lot the performance having to do the mapping for OTLP. So I would not repeat that mistake, also the reasoning was because SpanProcessor can be shared across multiple TraceProviders which in practice (never seen this case) could have been avoided and share exporter if needed.

So I think as explained before unless we prove with clear real code example that this does not work we should not do it.

bogdandrutu:
Also the example and the motivation that an object is shared between multiple instances is not a good reason to add unnecessary APIs that will limit performance, see https://github.com/open-telemetry/opentelemetry-specification/pull/1658#pullrequestreview-727210068

bogdandrutu:
> Hmm - it's hard to word it, but I think that there indeed may only be one precise use case. It is the concept of "sampling by attribute" - the attributes for a span are both what's set in its builder and its Resource I think. So we could never consider attribute-based sampling complete without the Resource. We currently have a PoC of attribute-based sampling for Java

> open-telemetry/opentelemetry-java-contrib#70

> which I would suspect to be spec'd in some form some day. Currently this sampler can't look at Resource but it seems like it should be able to.

@anuraaga please give me a real example when sampling by resource attributes is useful in practice.

I've seen extremely limited number of Applications with more than one Provider (personally 0, but let's say they exists) and usually in all these cases I expect that the attributes that are different in the Resource to be the "service" (or at least that was the motivation before to support multiple Providers) and if that is the case you can simply just ignore the resource attributes and set different Samplers per service instance of the Provider.

Oberon00:
> I think exposing the Resource in the SpanData was a mistake that hurts a lot the performance having to do the mapping for OTLP. So I would not repeat that mistake, also the reasoning was because SpanProcessor can be shared across multiple TraceProviders which in practice (never seen this case) could have been avoided and share exporter if needed.

I agree with that actually. I think exporters and span processors and samplers should only ever used with a single TracerProvider (span processor). There is also #508 open which would change how resources are passed (probably we need to close that one as its too late).

But now we have the Resource in the SpanData and I feel we already have reached consensus that #1298 is worthwhile to implement. If we pass the resource to the sampler, we can just start passing the new resource with the new attributes.
If the resource is given to the sampler outside the SDK's control, we now need to have a callback like OnResourceChanged. And the sampler developers would need to implement that callback to update their resource. Probably the callback could also fire at initial resource creation... Then we have the same features for samplers that this PR would provide plus much more flexibility but also some complexity.

@anuraaga What do you think of adding an OnResourceChanged callback once we implement #1298? Would that improve usability?

bogdandrutu:
@Oberon00 if we were to do a breaking change (since this one is) I would suggest to better investigate the path where all of these components are associated with only one Resource and maybe have a hook when the "Resource" changes to support late Resource resolution :)

This is my thought, I would better invest my time in doing something that we know is more performant and can improve things than adding more "simple" but not optimal changes.

Oberon00:
> if we were to do a breaking change (since this one is)

We have already "broken" this interface for 1.7 because we added the InstrumentationLibrary, which is why we are discussing this PR now.

jmacd:
Could we resolve this PR with an additional statement that language SIGs MAY provide optimized Sampler interfaces that bind certain arguments such as Resource and InstrumentationLibrary when they are known to be constant? That way we allow this PR in but @bogdandrutu's reservations are met, and we don't block #1298 (because I wrote "when they are known to be constant", any optimized API the language SIGs create will have to contend with 1298 when the time comes).

Oberon00:
How could that optimized interface look like?

jmacd:
There is already language in the SDK specification that offers SDKs a lot of freedom in how TraceID and SpanIDs are generated, I'm referring to "When asked to create a Span, the SDK MUST act as if doing the following in order:"

Step 2 is "Query the `Sampler`'s `ShouldSample` method".

I imagine the specification just saying what the act of calling `ShouldSample` does, then we shouldn't have to specify any alternate APIs. The optimized Sampler API MUST act the same as the non-optimized Sampler API, that's all.

In Go, `ShouldSample` takes a struct with some possibly-constant fields. The optimized Sampler API would provide a factory method for binding constant fields which returns the former Sampler implementation.

Oberon00:
I think @bogdandrutu has switched arguments over the course of the review: Originally, the suggestion was to add a more complex family of interfaces in order to gain some performance. Now, the main issue with this PR seems to be that we should keep the API minimal and not add unneeded things.
I think adding more implementation options would actually be worse in that regard.

jmacd:
I believe the underlying problem is that we do not have a data model for Resources.

Oberon00:
> I believe the underlying problem is that we do not have a data model for Resources.

Hmm, we do have a data model for the resources themselves IMHO, but the current solution of how they are connected to the actual telemetry is unsatisfying. E.g. you pass it to a tracer provider and separately to a meter?

jmacd:
> > I believe the underlying problem is that we do not have a data model for Resources.

> Hmm, we do have a data model for the resources themselves IMHO

I apologize, I over-stated that. I meant to say the data model we have is not strong enough, but that's a topic we should continue discussing in #1298. The resource SDK spec says:

```
A Resource is an immutable representation of the entity producing telemetry as Attributes.
```

We haven't made any requirements about the "entity" described there, and it means everyone has a different opinion about what **this** PR means.

> the current solution of how they are connected to the actual telemetry is unsatisfying

Yes. Also, as with #1298, the resource data model is not strong enough to accommodate Prometheus and OpenMetrics (https://github.com/open-telemetry/opentelemetry-specification/issues/1298#issuecomment-775749864).

In #1588 you wrote:

> Suppressing unwanted spans depending on the instrumentation library name 

Let me suggest that the problem is you're trying to use a Sampler API to suppress Spans. When DynaTrace originally introduced the Named Tracer and instrumentation library concept, it was stated that an SDK may decide to suppress spans based on their library. I.e., it's the `Tracer`'s responsibility to suppress spans.

I've made statements about how treating Sampler as a means of filtering spans is causing confusion which is an obstacle for probability sampling here: https://github.com/open-telemetry/opentelemetry-specification/issues/1844#issuecomment-920216604

If your goal is to suppress spans based on instrumentation library, that should happen before the Sampler is called. If your goal is to suppress spans based on resource, you can likewise do that before the Sampler is called. If your goal is to setup sampling rules that are applied on a per-instrumentation library basis, we should be specifying a View Configuration mechanism for spans that will configure a per-library Sampler (for example).

Based on this above reasoning, I'm on the fence about whether InstrumentationLibrary belongs in the sampler API at all. The reason InstrumentationLibrary is different than Resource, and why it was easier to accept, is that InstrumentationLibrary is at least a variable across the "entity producing telemetry", whereas Resource is not.

jmacd (commented):
I'm removing my approval, now that I've understood the motivation better. If the goal is to suppress spans based on Resource (or Instrumentation Library), better to do that before the Sampler is called.

Oberon00:
> If your goal is to suppress span ... better to do that before the Sampler is called.

That's not really possible, the sampler is the only hook that is exposed for that purpose. Replacing the TracerProvider with a custom implementation is usually not practical (one reason being that Resource APIs are only defined in the default SDK).

anuraaga:
> If the goal is to suppress spans based on Resource (or Instrumentation Library), better to do that before the Sampler is called.

At least for me the goal is to change sample rate based on service name, or k8s pod name, or hostname, type of resource attributes. Often this is something like seeing metrics report failure on one service and wanting to crank up the sampling rate for that service temporarily.

This works when sampling configuration is managed centrally, for example with X-ray remote sampling, or even if it just reads a file from etcd.

So attributes in the resource seem useful to have available when configuring sampling. I guess instrumentation library may be more geared to suppression, not sure but not resource.

iNikem:
> > If your goal is to suppress span ... better to do that before the Sampler is called.
> 
> That's not really possible, the sampler is the only hook that is exposed for that purpose. Replacing the TracerProvider with a custom implementation is usually not practical (one reason being that Resource APIs are only defined in the "official" SDK).

@jmacd I have to agree with @Oberon00 here: where else can we suppress spans (and how) if not by calling Sampler and returning `SamplingDecision.DROP`?

anuraaga:
> What about instead ask the remote service to tell your config and pass the Resource then?

This seems like a nice design, especially for an OTel configuration service API. AFAIK current remote sampling APIs can't accept a Resource so this approach doesn't seem to support them well. And I think for medium size teams with a dozen or so microservices on on-prem hosts, a single file pushed through Ansible or such seems reasonable - I've done similar for experiment configs before, not specifically sampling. I hope OTel doesn't just assume these are bad use cases.

A span has many attributes, some of them happen to be on the span, others on the Resource. It seems inconsistent to only provide one of these when sampling though.

Oberon00:
I will close this PR if/when #1942 is merged.

Oberon00:
Closing due to merge of #1942.

jmacd:
> It seems inconsistent to only provide one of these when sampling though

We've made an argument that performance suffers when every Sampler invocation considers the resource or instrumentation library. 

It sounds like you're asking for a standard mechanism. In the Go SDK, what you're asking for is two new interfaces:

```
type SDKSampler interface {
  SamplerForResource(resource.Resource) InstrumentationLibrarySampler
}

type InstrumentationLibrarySampler interface {
  SamplerForLibrary(instrumentation.Library) Sampler
}
```

Since all of the machinery implied by these two new interfaces is part of setting up and configuring an SDK, it falls outside the scope of OTel's current specification. You could build these and use them, and have what you want. 

Should OpenTelemetry offer these interfaces? Users want configurable tracer SDKs, so I think the answer is yes.

For your simple application of a file containing resource and library parameters for sampling configuration, you can: 
(1) parse the intended resource attribute to sampler settings map
(2) inspect the resource of the current SDK
(3) apply all resource attributes to the predicates used in the mapping
(4) some of the sampler settings mapped in your resource file are unreachable, their predicates are already false if resources do not match
(5) configure only the non-false predicates as the `SDKSampler`

the `SamplerForLibrary` method does roughly the same, evaluate any remaining predicates with the instrumentation library details, now the returned `Sampler` does exactly what you want and uses the current API -- the decisions about instrumentation library and resource are pre-bound by the time a Span arrives, and no further computation has to be done for any of the resource or library predicates.

Issue #274, opened 2019-09-30
Allow resources as span attributes


Oberon00:
See https://github.com/open-telemetry/opentelemetry-specification/pull/263#discussion_r329281769.

For some types of spans (e.g. HTTP server) there is potentially a lot of information that can be different from span to span, but there is only a low number of combinations for each process. For example, it would make sense to have `http.app_info` attribute the value of which is a resource that contains information such as an application root, server name, application name, server port (see linked PR).

bogdandrutu:
Please clarify what an "Application" is. Is that a "http.route"?

Oberon00:
See my new comments: https://github.com/open-telemetry/opentelemetry-specification/pull/263#discussion_r329726240

Oberon00:
Some clarifications (from https://github.com/open-telemetry/opentelemetry-specification/pull/263#discussion_r336907328):

> Basically the problem with the app attributes is that in many scenarios (admittedly, probably most) they will have the same values every time they are set. However, I think there are scenarios in which there are multiple apps within the same process, and that's why I disagree with your statement that these attributes "clearly" belong to resources. In these cases we can't use the resources concept which is process-wide, because we have no way to indicate which app the span belongs too.
>
> My goal with #274 would be to have just a single attribute "http.app_info" or similar the value of which is a (reference to a) process-wide app resource. Then you can use the process-wide resources to enumerate all the apps in the process and still know to which app the span belongs. Logically, I think that the app attributes would be perfectly fine as-is, but from a performance-perspective I can see that having them grouped together or transferred as a resource reference can have advantages.
>
> Of course that attribute would not be inherited, just like a string attribute is not inherited.

carlosalberto:
Hey @Oberon00 what's the state of this? can we close it? If not, what remains to be done?

Oberon00:
Well, https://github.com/open-telemetry/opentelemetry-specification/pull/421/ and open-telemetry/oteps#66 could both be helpful when designing a solution, but nothing has been done explicitly to solve this issue and the initial description still applies.

I think of this primarily as a potential solution to #335, which I would still love to be solved.

jmacd:
Related proposal: https://github.com/open-telemetry/oteps/issues/78

carlosalberto:
Moving this to the next milestone.

tsloughter:
This sounds more like having complex values for attributes and not necessarily resources?

[...]

PR #1661, opened 2021-04-29
Write description for Gauge metric type in DataModel Specification

jsuereth:


specification/metrics/datamodel.md
```diff
@@ -296,7 +296,31 @@ best tradeoff for their use case.
...
+  - A timestamp when the value was sampled (`time_unix_nano`)
+  - (optional) A timestamp (`start_time_unix_nano`) which has [TBD semantics](https://github.com/open-telemetry/opentelemetry-proto/pull/295).
+
+In OTLP, a Gauge stream represents the last-sampled event for a given time
+window.
+
+![Gauge](img/model-gauge.png)
+
+In this example, we can see an underlying timeseries we are sampling with our
+Gauage.  While the event model *can* sample more than once for a given metric
```
punya, line 314:
```suggestion
Gauge.  While the event model *can* sample more than once for a given metric
```

specification/metrics/datamodel.md
```diff
@@ -296,7 +296,31 @@ best tradeoff for their use case.
...
+reporting interval, only the last value is reported in the metric stream via
+OTLP.
+
+Gauges do not provide an aggregation semantic, instead "last sample value" is
+used when performing operations like temporal alignment or adjusting resolution.
+
+Gauages can be aggregated through transformation into histograms, or other
```
punya, line 321:
```suggestion
Gauges can be aggregated through transformation into histograms, or other
```

specification/metrics/datamodel.md
```diff
@@ -296,7 +296,31 @@ best tradeoff for their use case.
...
+
+In this example, we can see an underlying timeseries we are sampling with our
+Gauge.  While the event model *can* sample more than once for a given metric
+reporting interval, only the last value is reported in the metric stream via
+OTLP.
+
+Gauges do not provide an aggregation semantic, instead "last sample value" is
+used when performing operations like temporal alignment or adjusting resolution.
+
+Gauges can be aggregated through transformation into histograms, or other
```
kjordy, line 321: It looks like the AggregationTemporality 
 of a [Histogram](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L225-L231) expects either cumulative or delta changes. Does this mean non-cumulative histograms (for example: a distribution of memory usage across many containers) are not supported? Are there examples were a gauge stream would want to be transformed it into a cumulative or delta histogram?

specification/metrics/datamodel.md
```diff
@@ -296,7 +296,31 @@ best tradeoff for their use case.
...
+A [Gauge](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L174)
+in OTLP represents a sampled value at a given time.  A Gauge stream consists of:
+
+- A set of data points, each containing:
+  - An independent set of Attribute name-value pairs.
+  - A value sampled from some observable timeseries (e.g. current memory usage)
+  - A timestamp when the value was sampled (`time_unix_nano`)
+  - (optional) A timestamp (`start_time_unix_nano`) which has [TBD semantics](https://github.com/open-telemetry/opentelemetry-proto/pull/295).
+
+In OTLP, a Gauge stream represents the last-sampled event for a given time
```
Oberon00, line 308: I thought a stream was more than one event?

specification/metrics/datamodel.md
```diff
@@ -296,7 +296,31 @@ best tradeoff for their use case.
 
 ### Gauge
 
-Pending
+A [Gauge](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L174)
+in OTLP represents a sampled value at a given time.  A Gauge stream consists of:
+
+- A set of data points, each containing:
+  - An independent set of Attribute name-value pairs.
+  - A value sampled from some observable timeseries (e.g. current memory usage)
```
kenfinnigan, line 304: I could be missing context, but must a Gauge be an observation of a timeseries, or can it be "a sampled value"?

specification/metrics/datamodel.md
jsuereth: Ah, right.  This should be a "point in a Gauge stream".

specification/metrics/datamodel.md
jsuereth: The assumption there is it is a "sampled value" for some existing thing.  I can remove the word timeseries and find a better word for what I mean.

specification/metrics/datamodel.md
ebullient: This is where I need @jmacd and @reyang to sort me out. If I understood correctly, a Gauge is not a measurement that can be sensibly accumulated/aggregated across dimensions.

Values that can be accumulated/aggregated would come from UpDownCounters instead.

e.g. you would use a gauge to measure temperature (where you don't then add up all of the temperature measurements to some meaningful total), but you could use an UpDownCounter to observe memory usage (and then aggregate across lots of instances to get the total memory used).

@jmacd and @reyang .. please correct me. ;)

reyang: Gauges are not accumulated/aggregated across dimensions. The user might configure the SDK to force it to be treated as additive but normally we don't expect to see this (e.g. adding the CPU fan speed / temperature across multiple servers wouldn't make much sense).

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/new_api.md#asynchronous-gauge

reyang: Coming back to this PR, the wording here still sounds accurate to me.
For example, people can configure the SDK to report Gauges as Histograms - e.g. the distribution of CPU temperature, with buckets [RED - melting down, YELLO - hot, GREEN - normal].

reyang (approved):
LGTM.

specification/metrics/datamodel.md
jmacd: @kjordy see #274

jmacd: @kjordy see https://github.com/open-telemetry/opentelemetry-proto/issues/274

specification/metrics/datamodel.md
jmacd: I agree that using "timeseries" maybe adds confusion. I'm comfortable removing timeseries, but the best word I have to replace it with is unfortunately "gauge".

I've called this a "computed value", but that was a while ago.  I'll call it a  "non-sum pre-aggregated value", which isn't very catchy.

I like to think of Gauges as having unknown scale. They are not necessarily interval scale or ratio scale. (Delta sums are always ratio scale. Cumulative sums are always interval scale. See https://en.wikipedia.org/wiki/Level_of_measurement)

jmacd (approved):

specification/metrics/datamodel.md
jsuereth: @jmacd Given your definition, I'd say "nominal", possibly "ordinal" for Gauage.  However, I think (while more precise) this doesn't aide in the level of understanding I'm hoping to create here (something more for folks new to metrics to understand nuance and the more refined language further down).

If you'd like, it may be worth doing a Level-of-Measurement discussion and denoting more mathematical specifics throughout this document.  I'd like to leave that for a separate PR though.

specification/metrics/datamodel.md
jsuereth: @kjordy I think you're right to call out this line as a "Why would I want to histogram gauges".   Specifically, unlike most expected histogram, Gauges => Histogram would have a large loss of data compared to sending Histograms.  Specifically, since you gauges are only expected to report last value, converting to a histogram to aggregate across multiple processes/services is odd (but doable).

I agree this needs a better use case description.   I'm calling it out as a possibility for now, but I fully agree we should justify its use.

kenfinnigan (approved):
Thanks @jsuereth

ebullient (approved):

ebullient (approved):

Issue #1656, opened 2021-04-27
Standardize how to record DroppedAttributeCount in protocols that have no equivalent


tigrannajaryan:
For [Jaeger](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/jaeger.md) we suggest to use `otel.event.dropped_attributes_count` tag for Events (Jaeger Logs). We say nothing about how to record the value of DroppedAttributeCount for Spans.

For [Zipkin](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/zipkin.md) this is not defined (the spec says TBD).

I believe we need one standard way to record DroppedAttributeCount in protocols that have no equivalent and do it the same way for Spans, SpanEvents, Logs, Records.

I suggest that we always use `otel.dropped_attributes_count` attribute name to represent DroppedAttributeCount field of Spans, SpanEvents, Logs, Records when the equivalent does not exist in the protocol.

reyang:
@reyang  to follow up with @carlosalberto and create milestones for tracking.

codeboten:
@tigrannajaryan it's unclear to me what generates these dropped attributes/events in the first place. Is that defined somewhere in the specification?

tigrannajaryan:
> @tigrannajaryan it's unclear to me what generates these dropped attributes/events in the first place. Is that defined somewhere in the specification?
> 
> To be clear, I'm asking what would be causing this value to be non-zero

I believe they should be populated by SDKs when the limits are hit: https://github.com/open-telemetry/opentelemetry-specification/blob/5a19b53d71e967659517c02a69b801381d29bf1e/specification/trace/sdk.md#span-limits

It is not defined explicitly in the spec, worth adding.

codeboten:
Thank you for clarifying! I can add this to the spec.

PR #1662, opened 2021-04-29
Refactor non-OTLP Format Mapping and Fix Dropped Attribute Count bug

tigrannajaryan:
Fixes https://github.com/open-telemetry/opentelemetry-specification/issues/1656

The documents that describe transformations to Jaeger and Zipkin contained
duplicate information about how to record InstrumentationLibrary.

They also contained incomplete and incorrect information about how to record
dropped attribute count.

This change extracts common transformation rules into a separate document (non-otlp.md)
and references this document where appropriate.

We also define otel.event.dropped_attributes_count as the only way to record
dropped attribute count, regardless of the entity for which it is recorded.
Previously in Jaeger format we specified that for Events we must use
otel.event.dropped_attributes_count and did not specify at all what to use for Spans.

I believe using the same uniform key for Spans and Events is correct.

I consider this a specification bug and do not believe it should be treated as
a breaking change.

Also delete the Zipkin mapping for LocalChildSpanCount since there is no such
OpenTelemetry field anymore.

### DIFF
--- a/specification/trace/sdk_exporters/jaeger.md
+++ b/specification/trace/sdk_exporters/jaeger.md
@@ -3,6 +3,10 @@
 **Status**: [Stable](../../document-status.md)
 
 This document defines the transformation between OpenTelemetry and Jaeger Spans.
+The generic transformation [rules specified here](non-otlp.md) also apply. If a
+particular generic transformation rule and the rule in this document contradict
+then the rule in this document MUST be used.
+
 Jaeger accepts spans in two formats:
 
 * Thrift `Batch`, defined in [jaeger-idl/.../jaeger.thrift](https://github.com/jaegertracing/jaeger-idl/blob/master/thrift/jaeger.thrift), accepted via UDP or HTTP
@@ -23,12 +27,15 @@ and Jaeger.
 | Span.TraceId             | Span.traceIdLow/High | Span.trace_id | See [IDs](#ids)     |
 | Span.ParentId            | Span.parentSpanId | as SpanReference | See [Parent ID](#parent-id)     |
 | Span.SpanId              | Span.spanId       | Span.span_id     |      |
+| Span.TraceState          | TBD               | TBD              |      |
 | Span.Name                | Span.operationName | Span.operation_name |  |
 | Span.Kind                | Span.tags["span.kind"] | same | See [SpanKind](#spankind) for values mapping |
 | Span.StartTime           | Span.startTime | Span.start_time | See [Unit of time](#unit-of-time) |
 | Span.EndTime             | Span.duration | same | Calculated as EndTime - StartTime. See also [Unit of time](#unit-of-time) |
 | Span.Attributes          | Span.tags | same | See [Attributes](#attributes) for data types for the mapping.            |
+| Span.DroppedAttributesCount| Add to Span.tags | same | See [Dropped Attributes Count](non-otlp.md#dropped-attributes-count) for tag name to use. |
 | Span.Events              | Span.logs | same | See [Events](#events) for the mapping format. |
+| Span.DroppedEventsCount  | Add to Span.tags | same | See [Dropped Events Count](non-otlp.md#dropped-events-count) for tag name to use. |
 | Span.Links               | Span.references | same | See [Links](#links) |
 | Span.Status              | Add to Span.tags | same | See [Status](#status) for tag names to use. |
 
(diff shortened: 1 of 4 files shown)


specification/trace/sdk_exporters/non-otlp.md
```diff
@@ -0,0 +1,54 @@
...
+|--|--|--|
+|Code | `otel.status_code` | Name of the code, either `OK` or `ERROR`. MUST NOT be set if the code is `UNSET`. |
+|Description | `otel.status_description` | Description of the `Status` if it has a value otherwise not set. |
+
+### Dropped Attribute Count
+
+OpenTelemetry Span's dropped attribute count MUST be reported as a key-value
+pair associated with the Span. Similarly, Span Event's dropped attribute count
+MUST be reported as a key-value pair associated with the Span Event. In both
+cases the key name MUST be `otel.dropped_attribute_count`.
```
nrcventura, line 52: Should the name be `otel.dropped_attributes_count` to more closely match the fields on the span and event?

specification/trace/sdk_exporters/non-otlp.md
```diff
@@ -0,0 +1,54 @@
...
+
+The following table defines the OpenTelemetry `Status`'s mapping to Span's
+key-value pairs:
+
+|OpenTelemetry Status Field|non-OTLP Key|non-OTLP Value|
+|--|--|--|
+|Code | `otel.status_code` | Name of the code, either `OK` or `ERROR`. MUST NOT be set if the code is `UNSET`. |
+|Description | `otel.status_description` | Description of the `Status` if it has a value otherwise not set. |
+
+### Dropped Attribute Count
```
nrcventura, line 47: Should there be another section for dropped events count, or can this section cover both of those counts?

specification/trace/sdk_exporters/jaeger.md
```diff
@@ -3,6 +3,10 @@
 **Status**: [Stable](../../document-status.md)
 
 This document defines the transformation between OpenTelemetry and Jaeger Spans.
+The generic transformation [rules specified here](non-otlp.md) also apply. If a
```
nrcventura, line 6: With this link in place, should the section at the end of this document that refers to the dropped counts on events be updated too?

specification/trace/sdk_exporters/jaeger.md
```diff
@@ -118,18 +114,6 @@ In Jaeger Proto format the timestamps and durations MUST be represented
 with nanosecond precision using `google.protobuf.Timestamp` and
 `google.protobuf.Duration` types.
 
-### Status
```
yurishkuro, line 121: I would keep the section with a reference to non-otlp.md for the main transformation, and then make the following section "Error flag" a sub-section of Status

yurishkuro (approved):

specification/trace/sdk_exporters/non-otlp.md
tigrannajaryan: Yes, you are right.

specification/trace/sdk_exporters/non-otlp.md
tigrannajaryan: We need another section, I will add.

specification/trace/sdk_exporters/jaeger.md
tigrannajaryan: Yes, good catch.

specification/trace/sdk_exporters/non-otlp.md
tigrannajaryan: Fixed.

specification/trace/sdk_exporters/non-otlp.md
tigrannajaryan: Added.

specification/trace/sdk_exporters/jaeger.md
tigrannajaryan: Fixed.

specification/trace/sdk_exporters/jaeger.md
tigrannajaryan: Done.

nrcventura (approved):

carlosalberto (approved):

carlosalberto:
cc @anuraaga

specification/trace/sdk_exporters/non-otlp.md
```diff
@@ -0,0 +1,61 @@
...
+
+OpenTelemetry `InstrumentationLibrary`'s fields MUST be reported as key-value
+pairs associated with the Span using the following mapping:
+
+| OpenTelemetry InstrumentationLibrary Field | non-OTLP Key |
+| ------------------- | --- |
+| `InstrumentationLibrary.name`|`otel.library.name`|
+| `InstrumentationLibrary.version`|`otel.library.version`|
+
+### Span Status
```
yurishkuro, line 34: What about adding `Span.TraceState -> span.tags["trace_state"]` to common section as well ?

tigrannajaryan: Yes, let's do in a separate PR. I want this refactoring/clarification to be merged since it is blocking a Collector PR.

PR #1666, opened 2021-04-30
Introduce Schema URL in Tracer and Meter API

tigrannajaryan:
This adds to the API the changes proposed in OTEP 0152:
https://github.com/open-telemetry/oteps/blob/main/text/0152-telemetry-schemas.md

I will submit corresponding changes to the proto repository separately.

### DIFF
--- a/specification/metrics/api.md
+++ b/specification/metrics/api.md
@@ -133,22 +133,28 @@ This API MUST accept the following parameters:
   suppress telemetry produced by this library.
 * `version` (optional): Specifies the version of the instrumentation library
   (e.g. `1.0.0`).
+* [since 1.4.0] `schema_url` (optional): Specifies the Schema URL that should be
+  recorded in the emitted telemetry.
 
 It is unspecified whether or under which conditions the same or different
 `Meter` instances are returned from this functions.
 
 Implementations MUST NOT require users to repeatedly obtain a `Meter` again with
-the same name+version to pick up configuration changes. This can be achieved
-either by allowing to work with an outdated configuration or by ensuring that
-new configuration applies also to previously returned `Meter`s.
+the same name+version+schema_url to pick up configuration changes. This can be
+achieved either by allowing to work with an outdated configuration or by
+ensuring that new configuration applies also to previously returned `Meter`s.
 
 Note: This could, for example, be implemented by storing any mutable
 configuration in the `MeterProvider` and having `Meter` implementation objects
 have a reference to the `MeterProvider` from which they were obtained. If
 configuration must be stored per-meter (such as disabling a certain meter), the
-meter could, for example, do a look-up with its name+version in a map in the
-`MeterProvider`, or the `MeterProvider` could maintain a registry of all
+meter could, for example, do a look-up with its name+version+schema_url in a map
+in the `MeterProvider`, or the `MeterProvider` could maintain a registry of all
 returned `Meter`s and actively update their configuration if it changes.
+
+The effect of associating a Schema URL with a `Meter` MUST be that the telemetry
+emitted using the `Meter` will be associated with the Schema URL, provided that
+the emitted data format is capable of representing such association.
 
 ## Meter
 
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -118,22 +118,29 @@ This API MUST accept the following parameters:
   A TracerProvider could also return a no-op Tracer here if application owners configure
   the SDK to suppress telemetry produced by this library.
 - `version` (optional): Specifies the version of the instrumentation library (e.g. `1.0.0`).
-
+- [since 1.4.0] `schema_url` (optional): Specifies the Schema URL that should be
+  recorded in the emitted telemetry.
+  
 It is unspecified whether or under which conditions the same or different
 `Tracer` instances are returned from this functions.
 
 Implementations MUST NOT require users to repeatedly obtain a `Tracer` again
-with the same name+version to pick up configuration changes.
+with the same name+version+schema_url to pick up configuration changes.
 This can be achieved either by allowing to work with an outdated configuration or
 by ensuring that new configuration applies also to previously returned `Tracer`s.
 
 Note: This could, for example, be implemented by storing any mutable
 configuration in the `TracerProvider` and having `Tracer` implementation objects
-have a reference to the `TracerProvider` from which they were obtained.
-If configuration must be stored per-tracer (such as disabling a certain tracer),
-the tracer could, for example, do a look-up with its name+version in a map in
-the `TracerProvider`, or the `TracerProvider` could maintain a registry of all
-returned `Tracer`s and actively update their configuration if it changes.
+have a reference to the `TracerProvider` from which they were obtained. If
+configuration must be stored per-tracer (such as disabling a certain tracer),
+the tracer could, for example, do a look-up with its name+version+schema_url in
+a map in the `TracerProvider`, or the `TracerProvider` could maintain a registry
+of all returned `Tracer`s and actively update their configuration if it changes.
+
+The effect of associating a Schema URL with a `Tracer` MUST be that the
+telemetry emitted using the `Tracer` will be associated with the Schema URL,
+provided that the emitted data format is capable of representing such
+association.
 
 ## Context Interaction
 
--- a/specification/versioning-and-stability.md
+++ b/specification/versioning-and-stability.md
@@ -99,6 +99,26 @@ All existing API calls MUST continue to compile and function against all future 
 
 Languages which ship binary artifacts SHOULD offer [ABI compatibility](glossary.md#abi-compatibility) for API packages.
 
+##### Extending Existing API Calls
+
+An existing API call MAY be extended without incrementing the major version
+number if the particular language allows to do it in a backward-compatible
+manner.
+
+To add a new parameter to an existing API call depending on the language several
+approaches are possible:
+
+- Add a new optional parameter to existing methods. This may not be the right
+  approach for languages where ABI stability is part of our guarantees since it
+  likely breaks the ABI.
+
+- Add a method overload that allows passing a different set of parameters, that
+  include the new parameter. This is likely the preferred approach for languages
+  where method overloads are possible.
+
+There may be other ways to extend existing APIs in non-breaking manner. Language
+maintainers SHOULD choose the idiomatic way for their language.
+
 #### SDK Stability
 
 Public portions of SDK packages MUST remain backwards compatible.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ release.
 
 - `Get Tracer` should use an empty string if the specified `name` is null. ([#1654](https://github.com/open-telemetry/opentelemetry-specification/pull/1654))
 - Clarify how to record dropped attribute count in non-OTLP formats. ([#1662](https://github.com/open-telemetry/opentelemetry-specification/pull/1662))
+- Add schema_url support to `Tracer`. ([#1666](https://github.com/open-telemetry/opentelemetry-specification/pull/1666))
 
 ### Metrics
 
@@ -52,6 +53,7 @@ release.
 - Remove the "Func" name, use "Asynchronous" and "Observable". ([#1645](https://github.com/open-telemetry/opentelemetry-specification/pull/1645))
 - Add details to UpDownCounter API. ([#1665](https://github.com/open-telemetry/opentelemetry-specification/pull/1665))
 - Add details to Histogram API. ([#1657](https://github.com/open-telemetry/opentelemetry-specification/pull/1657))
+- Add schema_url support to `Meter`. ([#1666](https://github.com/open-telemetry/opentelemetry-specification/pull/1666))
 
 ### Logs
 
--- a/spec-compliance-matrix.md
+++ b/spec-compliance-matrix.md
@@ -19,6 +19,7 @@ formats is required. Implementing more than one format is optional.
 | [TracerProvider](specification/trace/api.md#tracerprovider-operations)                           |          |    |      |    |        |      |        |     |      |     |      |       |
 | Create TracerProvider                                                                            |          | +  | +    | +  | +      | +    | +      | +   | +    | +   | +    | +     |
 | Get a Tracer                                                                                     |          | +  | +    | +  | +      | +    | +      | +   | +    | +   | +    | +     |
+| Get a Tracer with schema_url                                                                     |          |    |      |    |        |      |        |     |      |     |      |       |
 | Safe for concurrent calls                                                                        |          | +  | +    | +  | +      | +    | +      | +   | +    | +   | +    | +     |
 | Shutdown (SDK only required)                                                                     |          | +  | +    | +  | +      | +    | -      |     | +    | +   | +    | +     |
 | ForceFlush (SDK only required)                                                                   |          | [-][go1606]  | +    | -  | +      | +    | -      |     | +    | -   | +    | +     |


specification/trace/api.md
```diff
@@ -135,6 +137,39 @@ the tracer could, for example, do a look-up with its name+version in a map in
...
+
+The effect of associating a Schema URL with a `Tracer` SHOULD be that the
+`schema_url` in the `InstrumentationLibrarySpans` message for all the telemetry
+emitted with the associated `Tracer` is populated with the provided Schema URL
+value (when using OTLP exporter).
+
+If the `Tracer` is not associated with a Schema URL then the exporter MUST leave
+the `schema_url` field in OTLP messages unset, in which case the
+application-wide Schema URL [will apply](sdk.md#application-wide-schema-url).
+
```
jkwatson, line 172: I don't think that the API spec should have any reference to OTLP, since that's 100% an SDK concern.

specification/metrics/api.md
```diff
@@ -375,6 +377,39 @@ other libraries.  The name of the `Meter` is explicitly not intended
...
+
+The effect of associating a Schema URL with a `Meter` SHOULD be that the
+`schema_url` in the `InstrumentationLibraryMetrics` message for all the
+telemetry emitted with the associated `Meter` is populated with the provided
+Schema URL value (when using OTLP exporter).
+
+If the `Meter` is not associated with a Schema URL then the exporter MUST leave
+the `schema_url` field in OTLP messages unset, in which case the
+application-wide Schema URL will apply (link to SDK TBD).
+
```
jkwatson, line 412: I don't think the API spec should reference OTLP, since that's an SDK concern.

specification/metrics/new_api.md
```diff
@@ -150,6 +152,39 @@ meter could, for example, do a look-up with its name+version in a map in the
...
+  telemetry is emitted using the instance.
+
+There may be other ways to modify the Get Meter API to allow the association
+with a Schema URL. Language maintainers SHOULD choose the idiomatic way for
+their language.
+
+The effect of associating a Schema URL with a `Meter` SHOULD be that the
+`schema_url` in the `InstrumentationLibraryMetrics` message for all the
+telemetry emitted with the associated `Meter` is populated with the provided
+Schema URL value (when using OTLP exporter).
```
zenmoto, line 182: Would it be better to say Schema URL value is supplied to the exporter to make it non OTLP-specific?

specification/metrics/api.md
reyang: +1

specification/trace/api.md
reyang: +100 💯

specification/metrics/new_api.md
jkwatson: No, the API doesn't know anything about exporters, either. We need a way to say this without referencing any SDK behavior or components.

jkwatson: Perhaps something like "the downstream telemetry will have this url associated with it, to identify what version of the schema has been used in generating the data".

specification/trace/api.md
tigrannajaryan: I rewrote this part, please take a look.

specification/metrics/api.md
tigrannajaryan: Rewritten.

specification/metrics/new_api.md
tigrannajaryan: Rewritten.

specification/metrics/api.md
```diff
@@ -375,6 +377,36 @@ other libraries.  The name of the `Meter` is explicitly not intended
 to be used as part of the instrument name, as that would prevent
 instrumentation libraries from capturing metrics by the same name.
 
+Note: `schema_url` optional parameter is added since spec 1.3.0. This change in
```
Oberon00, line 380: I don't think this should be placed here. This is a general guidance that would apply anytime we add a new parameter to an existing method. Maybe add it to the libary guidelines or versioning documents?

tigrannajaryan: Do you mean it would be good to extract these recommendations about how to introduce a new parameter to existing API in a separate common place in the spec?

tigrannajaryan: Done in https://github.com/open-telemetry/opentelemetry-specification/commit/903997be9aa2d750064230b5b6a676e9362b7bbb

carlosalberto (approved):

iNikem (approved):

specification/metrics/api.md
```diff
@@ -150,6 +152,9 @@ meter could, for example, do a look-up with its name+version in a map in the
 `MeterProvider`, or the `MeterProvider` could maintain a registry of all
 returned `Meter`s and actively update their configuration if it changes.
 
+The effect of associating a Schema URL with a `Meter` SHOULD be that the the
```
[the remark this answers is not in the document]
tigrannajaryan: This is for the case when it is not possible to make such association (e.g. exporting using a protocol that has no way to record such association).

specification/trace/api.md
```diff
@@ -118,7 +118,9 @@ This API MUST accept the following parameters:
   A TracerProvider could also return a no-op Tracer here if application owners configure
   the SDK to suppress telemetry produced by this library.
 - `version` (optional): Specifies the version of the instrumentation library (e.g. `1.0.0`).
-
+- [since 1.4.0] `schema_url` (optional): Specifies the Schema URL that should be
```
reyang, line 121: How would this work with the global default tracer?

specification/resource/sdk.md
```diff
@@ -70,6 +74,8 @@ The resulting resource MUST have all attributes that are on any of the two input
 If a key exists on both the old and updating resource, the value of the updating
 resource MUST be picked (even if the updated value is empty).
 
+The resulting resource will have the `schema_url` of the updating resource.
```
reyang, line 77: Would be great to provide an example of how `schema_url` would look like.

specification/trace/api.md
tigrannajaryan: I think the global tracer should not have any schema URL associated with it. I think it is best to require instrumentation to explicitly specify the schema they are using by calling using obtaining a Tracer, which is what we do in this PR.

It is possible that we can also have a default schema URL associated with the global default, but I am not totally sure it won't result in a wrong schema URL in the emitted telemetry due to instrumentation having no idea about it and disrespecting the concept of telemetry schema. If we think that it is a non-issue this can be added as an API/SDK requirement in a future PR (and possibly also have a way for the end usr to configure the default schema URL for global tracer).

specification/metrics/api.md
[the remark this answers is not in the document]
yurishkuro: I would phrase it as MUST with a caveat "provided that the export protocol supports schema".

To put it differently, if the protocol does support schema, what would be the reason for SDK to *not* include it?

SHOULD in the API spec make things unreliable for the consumer.

specification/trace/api.md
jmacd: I am looking forward to the use of a process-wide schema to indicate the schema for job and instance, to indicate how the process' Resource attributes can be used to look up secondary attributes via service discovery. If data originates from Prometheus, it's key attributes are `job` and `instance`. If data originates from an OTel SDK, it's key atttributes may be different. This can be decided in the future!

specification/versioning-and-stability.md
```diff
@@ -99,6 +99,26 @@ All existing API calls MUST continue to compile and function against all future
 
 Languages which ship binary artifacts SHOULD offer [ABI compatibility](glossary.md#abi-compatibility) for API packages.
 
+##### Extending Existing API Calls
+
+An existing API call MAY be extended without incrementing the major version
```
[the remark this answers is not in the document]
tigrannajaryan: I know that Go SDK did that at least for some APIs: for example it allows arbitrary new options to be added to Get Tracer API: https://github.com/open-telemetry/opentelemetry-go/blob/1e3fa3a3e01511c22cc24978379af13e5534bc2a/trace/trace.go#L667 which is precisely what I needed for this PR.

specification/metrics/api.md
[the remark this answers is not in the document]
tigrannajaryan: Done. Re-worded to the following:
>The effect of associating a Schema URL with a `Meter` MUST be that the telemetry
emitted using the `Meter` will be associated with the Schema URL, provided that
the emitted data format is capable of representing such association.

jkwatson:
This seems good to me. I'm slightly concerned that this will get merged and released with spec v1.4.0 before we have an actual schema URL to use. Should we publish a schema before we merge this, so that language maintainers will have something to use?

tigrannajaryan:
> This seems good to me. I'm slightly concerned that this will get merged and released with spec v1.4.0 before we have an actual schema URL to use. Should we publish a schema before we merge this, so that language maintainers will have something to use?

@jkwatson this is a good question and is is the next I want to tackle.

The _content_ for the fist schema we publish is easy (see OTEP). However, I wanted to discuss what we want to choose as OpenTelemetry's Schema URL. I was initially thinking about https://opentelemetry.io/schemas/1.4.0 but I am not sure there is a good way to publish at that URL from a file that is located in this repo since I don't know how opentelemetry.io is published. (And I think it is important for the schema file source content to be located in this repo where it can be modified simultaneously with semantic conventions).

Alternatively the OpenTelemetry Schema URL can point directly to a raw file in this repo, but it is both longer and more difficult (impossible?) to relocate if needed in the future.

I think we have plenty of time to make that location decision before 1.4.0 spec release, but if the implementation of the decision is complicated, then you may be right, we may not be able to publish it before the spec release.

We can delay merging this PR by a couple days and I can figure out the publishing if we feel that's what we want.

bogdandrutu (commented):
Partial review

specification/resource/sdk.md
bogdandrutu, line 77: This may be error prone, I think if you have 2 detectors running on different schema, instead of picking one of them as the winner better to return an error in that case.

tigrannajaryan: That's a good point. Let me remove resource/sdk.md modification from this PR and submit a separate PR for that.

tigrannajaryan: Done.

specification/resource/sdk.md
```diff
@@ -56,6 +56,10 @@ object. A factory method is recommended to enable support for cached objects.
 Required parameters:
 
 - [`Attributes`](../common/common.md#attributes)
+- [since 1.4.0] `schema_url` (optional): Specifies the Schema URL that should be
+  recorded in the emitted resource. If the `schema_url` parameter is unspecified
+  then the emitted resource will have the Schema URL that matches the
+  specification version that the SDK implements.
```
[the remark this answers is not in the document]
tigrannajaryan: Removed from this PR. Will create a new PR for Resource changes.

specification/metrics/api.md
```diff
@@ -133,6 +133,8 @@ This API MUST accept the following parameters:
   suppress telemetry produced by this library.
 * `version` (optional): Specifies the version of the instrumentation library
   (e.g. `1.0.0`).
+* [since 1.4.0] `schema_url` (optional): Specifies the Schema URL that should be
+  recorded in the emitted telemetry.
```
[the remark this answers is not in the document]
tigrannajaryan: In that case the schema URL is empty / unknown / unspecified.

specification/trace/api.md
```diff
@@ -118,7 +118,9 @@ This API MUST accept the following parameters:
...
+  recorded in the emitted telemetry.
+
```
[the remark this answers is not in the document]
tigrannajaryan: This is a good point, but I am not sure what is the right answer. We may want to do what you suggest or we want to keep it as is, so that it is not possible to instantiate 2 different Tracers for the same library but which use different schemas. I incline towards this later approach, since presumably each library should choose one schema and stick to it.
What do you think?

tigrannajaryan:
> This seems good to me. I'm slightly concerned that this will get merged and released with spec v1.4.0 before we have an actual schema URL to use. Should we publish a schema before we merge this, so that language maintainers will have something to use?

@jkwatson I briefly discussed with @austinlparker and it appears it is fairly easy to publish schema files from this repo (which will be the source of truth for schemas) to the website where the schema files will be downloadable as `https://opentelemetry.io/schemas/<version>`. 

Here is the draft plan: https://github.com/open-telemetry/opentelemetry.io/issues/577

I think we are good to move forward with this PR.

We can add further checks and controls that prevents mistakes (e.g. ensures that the schema is always published when making a release or that if the semantic conventions is changed there is a corresponding entry in the schema file). This can be done gradually, it is not a blocker.

tigrannajaryan:
This is ready for the final pass by commenters who did not yet approve and can be merged tomorrow if there are no new comments.

specification/trace/api.md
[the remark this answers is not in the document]
carlosalberto: That's actually an important item @tigrannajaryan - my impression is that we _should_ allow different schemas but discourage users to do this.

specification/trace/api.md
```diff
@@ -173,6 +180,13 @@ The `Tracer` MUST provide functions to:
 
 - [Create a new `Span`](#span-creation) (see the section on `Span`)
 
+The `Tracer` MAY provide a function to:
+
+- `SetSchema(schema_url)`. This is an optional method that may exist in
```
carlosalberto, line 185: Do we have a language that may need this? I'd leave this out if possible ;)

(if we _really_ need to keep this, we should only use the first specified value and ignore the rest)

specification/trace/api.md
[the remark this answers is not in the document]
jkwatson: I think that figuring this out *should* block this PR. We need to understand if schema URL is identifying information for a Tracer instance or not, and that's pretty crucial for implementors to know.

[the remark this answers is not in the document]
tigrannajaryan: I created an issue to decide this https://github.com/open-telemetry/opentelemetry-specification/issues/1695

specification/trace/api.md
tigrannajaryan: I was suspecting Go may need this, but no, it uses the Options idiom to allow adding new parameters in backward compatible manner. Java,C++,C# should not need this, they allow function overloads.

I am guessing it should be fine for Python, PHP, Ruby and JS too, since AFAIK they allow optional parameters, but I don't know what the current implementations of Get Trace/Meter do. @open-telemetry/python-maintainers @open-telemetry/javascript-maintainers @open-telemetry/php-maintainers @open-telemetry/ruby-maintainers can you help answer this? Are we able to add an optional parameter to Get Trace/Get Meter methods in a backward compatible manner?

I don't know how things work in other languages (Swift? Erlang? Rust?) @open-telemetry/swift-mainteiners @open-telemetry/erlang-maintainers @open-telemetry/rust-approvers  can you chime in?

cijothomas (approved):

specification/trace/api.md
codeboten: Optional parameters will work fine for Python.

TommyCpp: Rust doesn't support a Golang alike optional parameters but we can provide multiple functions to get tracers or maybe use the builder pattern. We haven't GA yet so breaking changes is also OK.

yurishkuro: Go doesn't support optional parameters either. The common approach to API stability in Go is to accept parameters in the form of a struct (passed by value to avoid allocations). New fields can be added to the struct at any time without breaking compatibility. There's also functional options pattern, but it has worse discoverability than a struct.

TommyCpp: Sorry I was thinking about Java. If the expectation here is to maintain backward compatibility for getting tracer API in case there are more fields/configurations in the future then we will have to revisit the current get tracer API. The builder pattern 
has been applied in many configurations within Rust SDK and exporters so the builder pattern may be a choice here.

tsloughter: No issue in Erlang.

carlosalberto: @MrAlias Could you confirm this is not an issue in Go?

tigrannajaryan: @carlosalberto I can confirm there is no issue in Go. I already have a draft PR that we discussed with @MrAlias and @Aneurysm9: https://github.com/open-telemetry/opentelemetry-go/pull/1889
(Go uses the Options pattern for obtaining the Tracer, so any new options can be added without breaking the API).

carlosalberto: @tigrannajaryan Thanks!

carlosalberto: ( @TommyCpp Let's discuss it separately your concern for Rust, as I personally think it's a bad idea to have the builder pattern for this specific case ;) )

tigrannajaryan: OK, unless someone objects I am going to remove this paragraph from this PR for now. If we find that any language needs it we can add it.

specification/trace/api.md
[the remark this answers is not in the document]
tigrannajaryan: All right, I made schema_url explicitly a part of the Tracer/Meter identity in https://github.com/open-telemetry/opentelemetry-specification/commit/ed15f3cc98abedf5a1d328e53d3fc64884be5c87

tigrannajaryan:
I believe this now addresses all comments. Please have another look.

tigrannajaryan:
This is now ready for merging. I squashed all commits into one. Will merge tomorrow if I don't hear any objections.

Issue #1647, opened 2021-04-26
Add semantic convention for describing network connectivity conditions


bryce-b:
**What are you trying to achieve?**
On mobile devices, the state of network connectivity is very important. Determining the severity of a network issue can be greatly impacted by the networking state a mobile device is operating under. 

I would like to add several semantic conventions to the spec to describe network conditions on a mobile device. To get the conversation started here a some proposed attributes:

`network.effectiveType` 
This can describe whether the device is connected via wifi, 2G, 3G, 4G, 5G.

`network.carrier`
Describes the cellular carrier if applicable. e.g. : `Sprint`

`network.isoCode`
2 character country code per [ISO 3166-1 standard](https://en.wikipedia.org/wiki/ISO_3166-1). e.g.: `DE` for Germany

`network.mobileCode` 
The mobile network code (MNC) of a cell service provider.  e.g.: `050` 
Though this is a numeric code, it's probably wise to transmit as a string to preserve leading zeros.

`network.countryCode`
The mobile country code (MCC). e.g.: `312` 
This may also have leading zeros and should be transmitted as a string.

The MCC & MNC can be used to uniquely identify a mobile network operator. All three codes are valuable, as one particular MCC may service multiple countries.

For more information on these values : https://en.wikipedia.org/wiki/Mobile_country_code

Oberon00:
These seem potentially useful to me.

I think you are already suggesting that but just to add a rationale: Adding these per span would make sense as a device may have multiple connections. Also, resources being immutable or maybe soon append-only (https://github.com/open-telemetry/opentelemetry-specification/issues/1298#issuecomment-827398011) may not really fit that.


A few notes on the names in general:

1. We have a general naming guide at https://github.com/open-telemetry/opentelemetry-specification/blob/v1.2.0/specification/common/attribute-and-label-naming.md which mandates the use of snake_case, so you would use `network.effective_type`, etc. instead of `network.effectiveType`.
2. We already have namespaces `net`, `net.peer` and `net.host`, see https://github.com/open-telemetry/opentelemetry-specification/blob/v1.2.0/specification/trace/semantic_conventions/span-general.md#general-network-connection-attributes. I believe these attributes would best be put under `net.host` or maybe `net.host.mobile` / `net.host.cellular`.


And on particular names:

3. effectiveType: Why "effective"?
4. isoCode: That's too generic IMHO. Maybe iso_country_code
[...]

PR #1669, opened 2021-05-04
Added mobile carrier info related semantic conventions to span spec

bryce-b:
Fixes #
Adds semantic conventions defining mobile carrier data defined in [#1647](https://github.com/open-telemetry/opentelemetry-specification/issues/1647)

## Changes
- adds `net.host.carrier.*`
- adds `net.host.connection_type`

Related issues #
#1647 : `Add semantic convention for describing network connectivity conditions`

### DIFF
--- a/specification/trace/semantic_conventions/span-general.md
+++ b/specification/trace/semantic_conventions/span-general.md
@@ -41,7 +41,25 @@ the `net.peer.*` properties of a client are equal to the `net.host.*` properties
 | `net.host.ip` | string | Like `net.peer.ip` but for the host IP. Useful in case of a multi-IP host. | `192.168.0.1` | No |
 | `net.host.port` | int | Like `net.peer.port` but for the host port. | `35555` | No |
 | `net.host.name` | string | Local hostname or similar, see note below. | `localhost` | No |
+| `net.host.connection_type` | string | The current network connection type in common terms | `wifi` | No |
+| `net.host.carrier.name` | string | Name of the mobile carrier. | `sprint` | No |
+| `net.host.carrier.mcc` | string | [Mobile country code](https://en.wikipedia.org/wiki/Mobile_country_code) | `310` | No | 
+| `net.host.carrier.mnc` | string | [Mobile network code](https://en.wikipedia.org/wiki/Mobile_country_code) | `001` | No |
+| `net.host.carrier.icc` | string | [ISO country code (two character)](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) | `DE` | No |
 
+`net.host.connection_type` SHOULD be one of the following : 
+
+| Value | Description |
+| --- | --- |
+| `wifi` | connected via wifi | 
+| `1G` |  connected via a first gen cellular technology, such as `AMPS` or equivalent | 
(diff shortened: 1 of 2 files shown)


specification/trace/semantic_conventions/span-general.md
```diff
@@ -41,7 +41,25 @@ the `net.peer.*` properties of a client are equal to the `net.host.*` properties
 | `net.host.ip` | string | Like `net.peer.ip` but for the host IP. Useful in case of a multi-IP host. | `192.168.0.1` | No |
 | `net.host.port` | int | Like `net.peer.port` but for the host port. | `35555` | No |
 | `net.host.name` | string | Local hostname or similar, see note below. | `localhost` | No |
-
+| `net.host.connection_type` | string | The current network connection type in common terms | `wifi` | No |
+| `net.host.carrier.name` | string | Name of the mobile carrier. | `sprint` | No |
+| `net.host.carrier.mcc` | string | [Mobile country code](https://en.wikipedia.org/wiki/Mobile_country_code) | `310` | No | 
+| `net.host.carrier.mnc` | string | [Mobile network code](https://en.wikipedia.org/wiki/Mobile_country_code) | `001` | No |
+| `net.host.carrier.icc` | string | [ISO country code (two character)](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) | `DE` | No |
```
iNikem, line 48: These are not "general network connection attributes", but are very specific to mobile clients.

Oberon00: I think the "general" in the title is more about "low-level networking" as opposed to "protocol-specific attributes".

In any case, I think if we make them span attributes (which seems to be the best currently available option) then putting them under `net.host` is sensible.

Also `net.host.connection_type` would make sense for other clients as well.

specification/trace/semantic_conventions/span-general.md
```diff
@@ -41,7 +41,25 @@ the `net.peer.*` properties of a client are equal to the `net.host.*` properties
...
+
+`net.host.connection_type` SHOULD be one of the following : 
+
+| Value | Description |
+| --- | --- |
+| `wifi` | connected via wifi |
```
Oberon00, line 54: What about adding a `wired` value?

Oberon00, line 54: How about also adding a `wired` value?

Oberon00 (commented):
Semantic conventions are mostly edited in the YAML sources from which the markdown is then generated. See https://github.com/open-telemetry/opentelemetry-specification/tree/main/semantic_conventions

specification/trace/semantic_conventions/span-general.md
bryce-b: sure!

nachoBonafonte (approved):

carlosalberto:
Ping @bryce-b

specification/trace/semantic_conventions/span-general.md
bryce-b: @iNikem do you think these should be placed in a different semantic convention document, specifically for mobile?

jmacd:
Please re-open if you would like to resume this work.

PR #1671, opened 2021-05-05
Release v1.3.0

arminru:
Monthly spec release, also adds missing changelog items.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,22 @@ release.
 
 ### Context
 
+### Traces
+
+### Metrics
+
+### Logs
+
+### Semantic Conventions
+
+### Compatibility
+
+### OpenTelemetry Protocol
+
+## v1.3.0 (2021-05-05)
+
+### Context
+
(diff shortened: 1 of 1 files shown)


carlosalberto (approved):

CHANGELOG.md
```diff
@@ -9,6 +9,22 @@ release.
...
+
+### Compatibility
+
+### OpenTelemetry Protocol
+
+## v1.3.0 (2021-05-05)
+
+### Context
+
 - Add `OTEL_EXPORTER_JAEGER_TIMEOUT` environment variable. ([#1612](https://github.com/open-telemetry/opentelemetry-specification/pull/1612))
```
Oberon00, line 28: Why is this under `Context`? Are the `Context` and `v1.3.0` heading swapped?

arminru: It was there before already but the diff makes it look as if the heading was just introduced. It should indeed not go there, however, so good catch. I'll add a separate heading for SDK env vars.

carlosalberto: Mmm, I think the entries were originally put in the incorrect section.

jmacd (approved):

PR #1672, opened 2021-05-05
Rename the metrics new_api.md to api.md and updated links

reyang:
This is the 2nd part of the change, the 1st part is covered by #1674.

Previously we've [decided](https://github.com/open-telemetry/opentelemetry-specification/pull/1557#issuecomment-802494722) to make progress on the metrics API spec by creating a separate `new_api.md` doc just to make PR review easier.

Now that the metrics API spec is getting in shape, we've discussed in the [Apr. 29th Metrics API/SDK spec SIG meeting](https://docs.google.com/document/d/1-bCYkN-DWJq4jw1ybaDZYYmx-WAe6HnwfWbkm8d57v8/edit#heading=h.7r6jflku0q8c) that we will rename it to `api.md` in early May.

## Changes

* Renamed `new_api.md` to `api.md`.
* Removed `old_api.md`.
* Made necessary updates to the links.

Related [oteps](https://github.com/open-telemetry/oteps)
* https://github.com/open-telemetry/oteps/blob/main/text/metrics/0146-metrics-prototype-scenarios.md


jmacd (approved):

cijothomas (approved):

bogdandrutu (approved):

jsuereth (approved):

arminru (commented):
@reyang I think if you do it in two commits (not sure if renaming is enough or if an explicit `git mv` is necessary), it should keep the file history intact. This way, one would also be able to find the respective PRs including context for each line using `git blame`.

reyang:
> @reyang I think if you do it in two commits (not sure if renaming is enough or if an explicit `git mv` is necessary), it should keep the file history intact. This way, one would also be able to find the respective PRs including context for each line using `git blame`.

@arminru thanks for calling this out. Please take a look at [this](https://github.com/reyang/opentelemetry-specification/blame/reyang/flips-metrics-old-new-apis/specification/metrics/api.md) which shows the git blame is smart enough to show all the history on the new API doc.

Issue #1116, opened 2020-10-20
Document Metrics Processor APIs and SDK requirements


as-polyakov:
(moving from https://github.com/open-telemetry/opentelemetry-java/issues/1810)
**Is your feature request related to a problem? Please describe.**
Similar to SpanProcessor, we need a hook for metrics which would allow users to intercept metrics lifecycle eventы with their own logic.

**Describe the solution you'd like**

```
public interface MetricsProcessor {

  /**
   * Called when bind() method is called. Allows to manipulate labels which this
   * instrument is bound to. Particular use case includes enriching lables and/or adding more labels
   * depending on the Context
   * @param ctx context of the operation
   * @param descriptor instrument descriptor
   * @param labels immutable labels. When processors are chained output labels of the previous one is passed as
   * an input to the next one. Last labels returned by a chain of processors are used for bind() operation.
   * @return labels to be used as an input to the next processor in chain or bind() operation if this is the last processor
   */
    Labels onLabelsBound(Context ctx, InstrumentDescriptor descriptor, Labels labels);

  /**
   *  Called when .record() method is called. Allows to manipulate recorded value. When chained input of the next
   *  call is the output of the previous call. Final output is recorded
   * @param ctx context of the operation
   * @param descriptor instrument descriptor
   * @param labels immutable labels. When processors are chained output labels of the previous one is passed as
   * @param value recorded value
   * @return value to be used as an input to the next processor in chain or record() operation if this is the last processor
   */

    long onLongMeasurement(Context ctx, InstrumentDescriptor descriptor, Labels labels, long value);

    double onDoubleMeasurement(Context ctx, InstrumentDescriptor descriptor, Labels labels, double value);
}
```

**Describe alternatives you've considered**
Main problem I am trying to address - is to be able to add certain labels to metrics depending on Context (i.e. dynamically). One alternative can be to just do that and make it a standalone feature aka "sticky labels". Had a simple PR for this [here](https://github.com/parallelstream/opentelemetry-java/pull/1). However MetricsProcessor is a more generic solution which can address other use cases.
As for the implementation, instead of using immutable Labels we can pass ReadWriteLables object, this would make it follow the same patter as in SpanProcessor

**Additional context**
There's a problem with this approach - statically bound instruments will not be able to dynamically add labels during record operation. Since we only intercept onLabelsBound() in case one statically binds an instrument and then uses that to record multiple measurements with different contexts, only first bind call will be intercepted

**Example PR**
[...]

PR #1673, opened 2021-05-06
Metrics SDK spec skeleton

reyang:
## Changes

Added the metrics SDK specification skeleton

Related issues #1116

### DIFF
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -22,7 +22,6 @@ the implementation unless explicitly communicated via
 Table of Contents
 </summary>
 
-* [Overview](#overview)
 * [MeterProvider](#meterprovider)
 * [MeasurementProcessor](#measurementprocessor)
 * [MetricProcessor](#metricprocessor)
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -54,7 +54,7 @@ TODO:
 
 In addition to things listed above, if the `Measurement` is reported by a
 synchronous `Instrument` (e.g. [Counter](./api.md#counter)),
-`MeasurementProcessor` MUST provide access to:
+`MeasurementProcessor` MUST have access to:
 
 * [Baggage](../baggage/api.md)
 * [Context](../context/context.md)
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -42,12 +42,12 @@ TODO:
 
 ## MeasurementProcessor
 
-`MeasurementProcessor` is an interface which allows hooks when a Measurement is
-recorded by an Instrument.
+`MeasurementProcessor` is an interface which allows hooks when a
+[Measurement](./api.md#measurement) is recorded by an
+[Instrument](./api.md#instrument).
 
-`MeasurementProcessor` has access to the [raw
-measurements](./datamodel.md#event-model) and the corresponding Instrument. It
-could also access the [Baggage](../baggage/api.md) and
+`MeasurementProcessor` has access to the Measurement and the corresponding
+Instrument. It could also access the [Baggage](../baggage/api.md) and
 [Context](../context/context.md) if the Baggage/Context are available.
 
 ```text
@@ -105,8 +105,14 @@ Metric Exporter has access to the [pre-aggregated metrics
 data](./datamodel.md#timeseries-model).
 
 There could be multiple [Push Metric Exporters](#push-metric-exporter) or [Pull
-Metric Exporters](#pull-metric-exporter) or even a mixture of both configured on a
-given MeterProvider.
+Metric Exporters](#pull-metric-exporter) or even a mixture of both configured on
+a given `MeterProvider`. Different exporters can run at different schedule, for
+example:
+
+* Exporter A is a push exporter which sends data every 1 minute.
+* Exporter B is a push exporter which sends data every 5 seconds.
+* Exporter C is a pull exporter which reacts to a scraper over HTTP.
+* Exporter D is a pull exporter which reacts to another scraper over named pipe.
 
 ### Push Metric Exporter
 
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -93,17 +93,17 @@ The following diagram shows `MetricProcessor`'s relationship to other components
 in the SDK:
 
 ```text
-+-----------------+  +-----------------+  +-----------------------+
-|                 |  |                 |  |                       |
-> In-memory state |--> MetricProcessor |--> MetricExporter (push) |--> Another process
-|                 |  |                 |  |                       |
-+-----------------+  +-----------------+  +-----------------------+
++-----------------+            +-----------------+            +-----------------------+
+|                 | Metrics... |                 | Metrics... |                       |
+> In-memory state | -----------> MetricProcessor | -----------> MetricExporter (push) |--> Another process
+|                 |            |                 |            |                       |
++-----------------+            +-----------------+            +-----------------------+
 
-+-----------------+  +-----------------+  +-----------------------+
-|                 |  |                 |  |                       |
-> In-memory state |--> MetricProcessor |--> MetricExporter (pull) |--> Another process (scraper)
-|                 |  |                 |  |                       |
-+-----------------+  +-----------------+  +-----------------------+
++-----------------+            +-----------------+            +-----------------------+
+|                 | Metrics... |                 | Metrics... |                       |
+> In-memory state |------------> MetricProcessor |------------> MetricExporter (pull) |--> Another process (scraper)
+|                 |            |                 |            |                       |
++-----------------+            +-----------------+            +-----------------------+
 ```
 
 ## MetricExporter
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -46,9 +46,24 @@ TODO:
 [Measurement](./api.md#measurement) is recorded by an
 [Instrument](./api.md#instrument).
 
-`MeasurementProcessor` has access to the Measurement and the corresponding
-Instrument. It could also access the [Baggage](../baggage/api.md) and
-[Context](../context/context.md) if the Baggage/Context are available.
+`MeasurementProcessor` MUST provide access to:
+
+* The `Measurement`
+* The `Instrument`, which is used to report the `Measurement`
+* The `Resource`, which is associated with the `MeterProvider`
+
+In addition to things listed above, if the `Measurement` is reported by a
+synchronous `Instrument` (e.g. [Counter](./api.md#counter)),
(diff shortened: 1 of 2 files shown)


specification/metrics/sdk.md
```diff
@@ -0,0 +1,122 @@
...
+TODO:
+
+* Allow multiple pipelines (processors, exporters).
+* Configure "Views".
+* Configure timing (related to [issue
+  1432](https://github.com/open-telemetry/opentelemetry-specification/issues/1432)).
+
+## MeasurementProcessor
+
+`MeasurementProcessor` is an interface which allows hooks when a Measurement is
```
cijothomas, line 45: Are Aggregators expected to be implemented as `MeasurementProcessor`?

specification/metrics/sdk.md
```diff
@@ -0,0 +1,122 @@
...
+`MeasurementProcessor` has access to the [raw
+measurements](./datamodel.md#event-model) and the corresponding Instrument. It
+could also access the [Baggage](../baggage/api.md) and
+[Context](../context/context.md) if the Baggage/Context are available.
+
+```text
++------------------+
+| MeterProvider    |                 +----------------------+
+|   Meter A        | Measurements... |                      | Metrics... +-----------------+
+|     Instrument X |-----------------> MeasurementProcessor +------------>                 |
```
reyang, line 57: @jkwatson mentioned during the SIG meeting that we need a way to make composite processors.

specification/metrics/sdk.md
```diff
@@ -0,0 +1,122 @@
...
+of telemetry data.
+
+The goal of the interface is to minimize burden of implementation for
+protocol-dependent telemetry exporters. The protocol exporter is expected to be
+primarily a simple telemetry data encoder and transmitter.
+
+Metric Exporter has access to the [pre-aggregated metrics
+data](./datamodel.md#timeseries-model).
+
+There could be multiple [Push Metric Exporters](#push-metric-exporter) or [Pull
```
reyang, line 107: @jkwatson need to call out that multiple push exporters can have different frequency.

specification/metrics/sdk.md
reyang: At this stage we can implement aggregators as SDK internal components that can be used in a MeasurementProcessor, and we have better understanding, we can decide if we want to expose Aggregator as an interface.

specification/metrics/sdk.md
hdost: Related to this

There are a few different aspects of concern for me as it relates to both the MetricProcessors and Measurement Processors.

I have a few questions:
Would we want to allow for "parallel processes" on a singular stream?  Or more clearly maybe is there a possibility that a split may occur where two separate processors operate on a measurement/metric to then join back later to be combined.
Are separate measurements coming in allowed to be joined into a single metric?

specification/metrics/sdk.md
```diff
@@ -0,0 +1,128 @@
...
+|     Instrument Z |                                                     | In-memory state |
+|     ...          |                 +----------------------+            |                 |
+|     ...          | Measurements... |                      | Metrics... |                 |
+|     ...          |-----------------> MeasurementProcessor +------------>                 |
+|     ...          |                 |                      |            +-----------------+
+|     ...          |                 +----------------------+
++------------------+
+```
+
+## MetricProcessor
```
victlu, line 69: @reyang 

Your current pipeline is...

[API/SDK] -> `measurement` -> [1. MeasurementProcessor] -> [2. MetricProcessor <-> `[**In-Memory-State**]`] -> `metric` -> [3. MetricExporter]

Maybe we can fully expand to...

[API/SDK] -> `measurement` -> [1. MeasurementProcessor]* -> `measurement` -> [2. MetricProcessor <-> `[**In-Memory-State**]`] -> [MetricCollector] -> `metric` -> **[MetricEnricher]** -> `metric` -> [3. MetricExporter]

MeasurementProcessor: Allows adjustment to attributes. Can be chained/composed.

MetricCollector: Collect aggregated metrics from In-Memory-State. This is likely build-in/internal to SDK.

MetricEnricher: Allows custom adjustment to metrics. Can be chained/composed.

specification/metrics/sdk.md
```diff
@@ -0,0 +1,129 @@
...
+
+* Allow multiple pipelines (processors, exporters).
+* Configure "Views".
+* Configure timing (related to [issue
+  1432](https://github.com/open-telemetry/opentelemetry-specification/issues/1432)).
+
+## MeasurementProcessor
+
+`MeasurementProcessor` is an interface which allows hooks when a
+[Measurement](./api.md#measurement) is recorded by an
```
jsuereth, line 46: One thing that I *think* needs to be called out here.

If we want the SDK to support [Exemplars](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L664), then we need a different definition of Measurement that optionally includes "Context" (or active trace/span + attributes) here.

I'm not sure if we've discussed exemplars or where they fit and if the API should expose the notion or they belong only in SDK / MeasurementProcessor but I *do* think we need them to be possible in the SDK and that means making sure they fit this scaffold.

reyang: The current description is like "MeasurementProcessor will be triggered when Measurements are reported, and the MeasurementProcessor will have access to Baggage/Context if they are available (e.g. in a synchronous instrument scenario)".

I think this is sufficient to allow the SDK to capture the current span + attributes.

I can try to add more clarification here and probably come up with one example.

specification/metrics/sdk.md
reyang: This is something I intended to avoid 😃

I think the goal is to start from a small set of essential (ideally must-have) LEGO pieces, and depending on what we've learned while building the SDKs we can decide to introduce more if there is a strong reason.

specification/metrics/sdk.md
reyang: I've updated the description, please take another look.

specification/metrics/sdk.md
```diff
@@ -0,0 +1,144 @@
...
+active span](../trace/api.md#context-interaction)).
+
+```text
++------------------+
+| MeterProvider    |                 +----------------------+
+|   Meter A        | Measurements... |                      | Metrics... +-----------------+
+|     Instrument X |-----------------> MeasurementProcessor +------------>                 |
+|     Instrument Y +                 |                      |            |                 |
+|   Meter B        |                 +----------------------+            |                 |
+|     Instrument Z |                                                     | In-memory state |
```
reyang, line 75: @jkwatson and @noahfalk suggested that we put separate boxes for "In-memory state" just to reflect the logical flow (and clarify that people can optimize things).

reyang: I've updated the diagrams based on the discussion during the Metrics API/SDK Spec SIG.

specification/metrics/sdk.md
reyang: Inside the SDK, the data will not join back.

It is possible that the data get forked and sent via multiple exporters (e.g. one exporter push at every 1 second, another export push every 1 minute), and the exporters send data to the same collector/backend. In this case the data might join back and cause problem.

jkwatson (approved):
Thanks! Looks like a good start!

cijothomas (approved):

specification/metrics/sdk.md
```diff
@@ -0,0 +1,144 @@
...
+
+`MetricExporter` defines the interface that protocol-specific exporters must
+implement so that they can be plugged into OpenTelemetry SDK and support sending
+of telemetry data.
+
+The goal of the interface is to minimize burden of implementation for
+protocol-dependent telemetry exporters. The protocol exporter is expected to be
+primarily a simple telemetry data encoder and transmitter.
+
+Metric Exporter has access to the [pre-aggregated metrics
```
victlu, line 119: Can you clarify a bit on how Metric Exporter gain access to pre-aggregated metrics? Is it that it has access to the "in-memory state" or is it passed those via the MetricProcessor?

reyang: I've updated the diagram to clarify that exporters don't have direct access to the "in-memory-state", and it receives (whether pulls from, or being pushed) pre-aggregated metrics from the processor.

reyang: The "input" should be a batch of "exportable representation", similar to the tracing SDK spec https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#exportbatch. These are the meats that we'll need to add in subsequent PRs.

jmacd (approved):

specification/metrics/sdk.md
jmacd: @hdost The data model specifies what "overlapping streams" look like. You could consider the parallel streams of your example intentionally overlapping points.

We may want to explain that the kind of parallel data you are describing should be renamed as a new metric name. I do not want to have to specify anything about parallel streams of data that are aggregated differently but carry duplicate meaning, for example.

specification/metrics/sdk.md
jmacd: +1 on avoiding this kind of strictness until there's a strong reason (thinking about mistakes I've made) 🤣

arminru:
@reyang Changelog please :-)

reyang:
@arminru I've struggling if I should add changelog right now or later (e.g. when we remove the banner "language clients should not start the implementation unless we explicitly communicated").

If we put the changelog now, it might be too noisy? (I'm thinking that we have one changelog entry at late May saying "Added the experimental version of Metrics SDK Spec (a long list of PRs...)")

Please let me know your suggestion.

victlu (approved):

jsuereth (approved):

specification/metrics/sdk.md
```diff
@@ -0,0 +1,144 @@
...
+* Configure timing (related to [issue
+  1432](https://github.com/open-telemetry/opentelemetry-specification/issues/1432)).
+
+## MeasurementProcessor
+
+`MeasurementProcessor` is an interface which allows hooks when a
+[Measurement](./api.md#measurement) is recorded by an
+[Instrument](./api.md#instrument).
+
+`MeasurementProcessor` MUST provide access to:
```
Aneurysm9, line 49:
```suggestion
`MeasurementProcessor` MUST be provided access to:
```

Is this what's meant here?  It seems like these (as well as baggage, context, etc.) are values that would be provided _to_ the`MeasurementProcessor` at the time a `Measurement` is taken.  It's not clear to me how the `MeasurementProcessor` would provide these values.

reyang: Thanks for the catch! I've reworded it to `MUST have access to`.

SergeyKanzhelev:
Looks like the PR is OK to merge - enough approvals, cool off period completed, and comments were addressed.

PR #1677, opened 2021-05-07
Add `OTEL_SERVICE_NAME` environment variable

### LINKED ISSUE #709: Add OTEL_SERVICE_NAME env variable

To simplify the configuration of auto-instrumentations via environment variables, especially for first time users, add new variable `OTEL_SERVICE_NAME` as a way to configure service name for all exporters. As was pointed in https://github.com/open-telemetry/opentelemetry-specification/pull/666#discussion_r445607065, configuring service name via resources may be confusing for new users.

**Additional context.**
This issue assumes that #666 is merged and we have environment variables recommendation in spec.

iNikem:
Closes #709

## Changes

Add `OTEL_SERVICE_NAME` environment variable to SDK configuration


pellared (approved):

specification/sdk-environment-variables.md
```diff
@@ -32,6 +32,7 @@ For example, the value `12000` indicates 12000 milliseconds, i.e., 12 seconds.
 | Name                     | Description                                       | Default                           | Notes                               |
 | ------------------------ | ------------------------------------------------- | --------------------------------- | ----------------------------------- |
 | OTEL_RESOURCE_ATTRIBUTES | Key-value pairs to be used as resource attributes |                                   | See [Resource SDK](./resource/sdk.md#specifying-resource-information-via-an-environment-variable) for more details. |
+| OTEL_SERVICE_NAME        | Value of the [`service.name`](./resource/semantic_conventions/README.md#service) resource attribute | | If both this and `OTEL_RESOURCE_ATTRIBUTES` are provided then `OTEL_SERVICE_NAME` is [merged into](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#merge) `OTEL_RESOURCE_ATTRIBUTES`
```
arminru, line 35: I think the fact that both are merged into `OTEL_RESOURCE_ATTRIBUTES` is an implementation detail that we don't have to expose to users.
More important would be to specify which one is taking precedence if `service.name` is also provided in `OTEL_RESOURCE_ATTRIBUTES`.

arminru, line 35: The description "_Value of the [`service.name`](./resource/semantic_conventions/README.md#service) resource attribute_" could be interpreted as "_you have to specify the same value here_" or even "_the value defined for `service.name` is exposed in this env var_".

How about this?
> Defines a default value for the [`service.name`](./resource/semantic_conventions/README.md#service) resource attribute

arminru, line 35:
```suggestion
| OTEL_SERVICE_NAME        | Value of the [`service.name`](./resource/semantic_conventions/README.md#service) resource attribute | | If both this and `OTEL_RESOURCE_ATTRIBUTES` are provided then `OTEL_SERVICE_NAME` is [merged into](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#merge) `OTEL_RESOURCE_ATTRIBUTES` |
```

arminru (commented):
@SergeyKanzhelev You seemed to have a strong opinion on #709, so please take a look at this proposal.

specification/sdk-environment-variables.md
pellared: For me "merged into" means that `OTEL_SERVICE_NAME` overrides the value from `OTEL_RESOURCE_ATTRIBUTES`.

iNikem: I agree with the assessment, but I am not sure about "default value". I usually interpret "default" as "value used unless user provided their opinion". Here is exactly user's opinion :) Maybe just

> Sets the value of `service.name` resource attribute?

iNikem: I can do it this way, sure :)

arminru: The "default" wording would stress that values provided by application/instrumentation in code take precedence over the env var but this is already stated in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#specifying-resource-information-via-an-environment-variable which is linked in the line above anyway. Your proposal sounds good as well!

specification/sdk-environment-variables.md
```diff
@@ -32,6 +32,7 @@ For example, the value `12000` indicates 12000 milliseconds, i.e., 12 seconds.
 | Name                     | Description                                       | Default                           | Notes                               |
 | ------------------------ | ------------------------------------------------- | --------------------------------- | ----------------------------------- |
 | OTEL_RESOURCE_ATTRIBUTES | Key-value pairs to be used as resource attributes |                                   | See [Resource SDK](./resource/sdk.md#specifying-resource-information-via-an-environment-variable) for more details. |
+| OTEL_SERVICE_NAME        | Sets the value of the [`service.name`](./resource/semantic_conventions/README.md#service) resource attribute | | If `service.name` is also provided in `OTEL_RESOURCE_ATTRIBUTES` then `OTEL_SERVICE_NAME` takes precedence. |
```
arminru, line 35:
```suggestion
| OTEL_SERVICE_NAME        | Sets the value of the [`service.name`](./resource/semantic_conventions/README.md#service) resource attribute | | If `service.name` is also provided in `OTEL_RESOURCE_ATTRIBUTES`, then `OTEL_SERVICE_NAME` takes precedence. |
```

owais (approved):

specification/sdk-environment-variables.md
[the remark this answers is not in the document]
iNikem: I usually prefer to keep PR scope small, if possible. This PR solves one specific issue. I am happy to file a follow-up one, if you feel this is needed :)

jkwatson (approved):

MrAlias (approved):

specification/sdk-environment-variables.md
```diff
@@ -32,6 +32,7 @@ For example, the value `12000` indicates 12000 milliseconds, i.e., 12 seconds.
 | Name                     | Description                                       | Default                           | Notes                               |
 | ------------------------ | ------------------------------------------------- | --------------------------------- | ----------------------------------- |
 | OTEL_RESOURCE_ATTRIBUTES | Key-value pairs to be used as resource attributes |                                   | See [Resource SDK](./resource/sdk.md#specifying-resource-information-via-an-environment-variable) for more details. |
+| OTEL_SERVICE_NAME        | Sets the value of the [`service.name`](./resource/semantic_conventions/README.md#service) resource attribute | | If `service.name` is also provided in `OTEL_RESOURCE_ATTRIBUTES`, then `OTEL_SERVICE_NAME` takes precedence. |
```
SergeyKanzhelev, line 35: somewhere needs to be a note on what wins - environment variable or code.

SergeyKanzhelev (changes requested):
I am ok with the suggestion, but before merging, let's clarify whether environment variable should take precedence over the value set in code.

SergeyKanzhelev (approved):

specification/sdk-environment-variables.md
SergeyKanzhelev: yep, sorry

yurishkuro (approved):

anuraaga (approved):

CHANGELOG.md
```diff
@@ -23,6 +23,8 @@ release.
 
 ### SDK Configuration
 
+- Add `OTEL_SERVICE_NAME` environment variable. ([#1677](https://github.com/open-telemetry/opentelemetry-specification/pull/1677))
```
bogdandrutu, line 26: I would add NAMESPACE and INSTANCE as well. Service name should not be treated special than the others.

bogdandrutu (dismissed):
Question that is not answer:

Why service name is treated differently than any other resource attribute?

CHANGELOG.md
iNikem: I usually prefer to keep PR scope small, if possible. This PR solves one specific issue. I am happy to file a follow-up one, if you feel this is needed :)

iNikem:
> Question that is not answer:
> 
> Why service name is treated differently than any other resource attribute?

1. It is the only de-facto "required" resource attribute
2. The majority of vendors already have similar variable (https://github.com/open-telemetry/opentelemetry-specification/issues/709#issuecomment-813260663), so this is what end-users already expect
3. This PR does not preclude any other PR which adds support for other resource attributes.

specification/sdk-environment-variables.md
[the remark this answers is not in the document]
arminru: That's reasonable. If we add it, it would ideally be in the same spec version, however, so SDK maintainers/implementers can tackle this tightly coupled feature at once. The next release is scheduled for early June, so we should be good here, as the follow-up should be as uncontroversial as this PR.

iNikem:
@bogdandrutu as discussed during spec SIG meeting, I tried to stress the significance of this specific attribute. Please check my changes in `README.md`

CHANGELOG.md
Oberon00: I have never heard of anyone using namespace (but maybe it just works so well that there are no questions?). instance.id has is currently IMHO too vaguely defined to be useful, as discussed in #1034. So the service name is IMHO definitely the most useful one.

Oberon00 (approved):

specification/sdk-environment-variables.md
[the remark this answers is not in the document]
Oberon00: Same as https://github.com/open-telemetry/opentelemetry-specification/pull/1677#discussion_r629718065: I don't think the service version is used often enough to warrant it's own attribute. We did receive user requests for service.name AFAIK, but not for any others.

CHANGELOG.md
Oberon00: (If exporter specs are updated to include service.namespace in their reported service, which maybe they should, then adding a special envvar for service.namespace might make sense)

specification/resource/semantic_conventions/README.md
```diff
@@ -36,9 +36,20 @@ Attributes are grouped logically by the type of the concept that they described.
 
 Certain attribute groups in this document have a **Required** column. For these groups if any attribute from the particular group is present in the Resource then all attributes that are marked as Required MUST be also present in the Resource. However it is also valid if the entire attribute group is omitted (i.e. none of the attributes from the particular group are present even though some of them are marked as Required in this document).
 
+## Attributes with Special Handling
+
+Given their significance some resource attributes receive a special treatment as described below.
```
arminru, line 41: Nit: This sounds quite colloquial (at least for me as a non-native speaker). How about this?
```suggestion
Given their significance some resource attributes are treated specifically as described below.
```

carlosalberto:
Ping @bogdandrutu

CHANGELOG.md
bogdandrutu: > If exporter specs are updated to include service.namespace in their reported service, which maybe they should, then adding a special envvar for service.namespace might make sense

I think the "service name" as defined right now is unique within a namespace, and the fact that Jaeger/Zipkin don't have this concept made us use just "service name" which is wrong. 

@Oberon00 I agree that exporters that only have a "service name" concept should use actually "{service.namespace}/{service.name" as the name.

bogdandrutu: https://github.com/open-telemetry/opentelemetry-specification/pull/1710

SergeyKanzhelev: https://github.com/open-telemetry/opentelemetry-specification/blob/d001d5d25ea11a28bd4c4ec2652e2db1a41fc064/specification/trace/sdk_exporters/zipkin.md#:~:text=note%2C%20the%20attribute%20service.namespace%20must%20not%20be%20used%20for%20the%20zipkin%20service%20name%20and%20should%20be%20sent%20as%20a%20zipkin%20tag. 

> Note, the attribute service.namespace MUST NOT be used for the Zipkin service name and should be sent as a Zipkin tag.

https://github.com/open-telemetry/opentelemetry-specification/blob/b46bcab5fb709381f1fd52096a19541370c7d1b3/specification/resource/semantic_conventions/README.md#:~:text=Note,same%20string

> Note: service.namespace and service.name are not intended to be concatenated for the purpose of forming a single globally unique name for the service. For example the following 2 sets of attributes actually describe 2 different services (despite the fact that the concatenation would result in the same string):

SergeyKanzhelev: We can change that, but it was intentionally not suppose to be concatenated to avoid the need to parse the namespace out of the name

bogdandrutu: I do understand that, and I've just added an open question to my PR. I think in case of one value string like Jaeger/Zipkin we have to guarantee the uniqueness somehow, otherwise we will have conflicts. For example 2 service names named "kafka" run by 2 teams/namespaces "a" and "b" will be seen as same service by these backends.

SergeyKanzhelev: happy to change that. Given that namespace is empty in many cases now it is not such a breaking change

iNikem: Does this discussion/related PR block this PR from merging?

trask (approved):

carlosalberto:
@bogdandrutu  ;)

iNikem:
@open-telemetry/technical-committee Anybody feeling confident enough to merge this?

carlosalberto:
@bogdandrutu Thanks for going on with the compromise, really appreciated. Merging (we will see in the long run how things unfold! ;) )

Issue #780, opened 2020-07-15
bytes / Byte (uint8_t) arrays support


maxgolov:
Current spec / protocol allows the following types:
```
bool
int32_t
int64_t
uint32_t
uint64_t
double
string (UTF-8, no UTF-16)
```

and

`array` of above primitive types.

Proposal is to **add support for '8-bit byte array' event properties for Traces and Logs**:

**Practical reasons**

- existing / legacy telemetry flows could already allow support for byte arrays.

- `8-bit byte array` or `bin` type is available in MsgPack protocol.

- same type known in [protobuf](https://developers.google.com/protocol-buffers/docs/proto?csw=1#scalar) as `bytes` 

- Most IoT protocols usually operate on 8-bit byte arrays or octet streams, allowing fields be either `uint8_t` or array / sequence of `uint8_t`, or the entire payload be 8-bit / octet stream.

- **UUID** or **GUID** type is best represented on SDK API surface and on wire as 16-bytes rather than 36+-character string, resulting in at least x2 times better performance for native (C/C++) code, as well as more compact net-bytes on wire. Although it may be possible to represent it as array of `int32_t` - it'd be a bit unnatural to do it.

- although it may be possible to convert `byte array` to base64-encoded `string` representation, this results in an extra encoding/decoding in SDK, elevated memory and CPU pressure, as well as type information (such as it's a binary blob, not a string) - is inherently lost.

Is it possible to add an **optional** provision to the protocol, that Platform SDKs that may require **bin** or *octet stream* or *8-bit byte array* - **may** implement it on API surface?

**Use-cases**

Examples of exporters that would benefit from having `byte array` in the protocol spec:
- ETW - Event Tracing for Windows supports both GUID (16-bytes) and other arbitrary Byte arrays.
- Other telemetry systems previously using byte arrays flow for more compact binary data representation.

Not having this type natively supported by OpenTelemetry would impede the adoption of OT API / SDK by customers currently relying on octet / binary streams in their data flows. A reference implementation can be provided based on Open Telemetry C++ SDK to illustrate the use-cases, as well as perf implications of not having native support for byte arrays in API/SDK/proto/specification.

maxgolov:
I would like to have conditional provisions for this type implemented in OpenTelemetry C++ SDK. Disabled by default. But still there. Have ability to experiment with it and enable it later on.

[...]

PR #1678, opened 2021-05-07
Add bytes (byte array) support to Log Data Model

tigrannajaryan:
Contributes to https://github.com/open-telemetry/opentelemetry-specification/issues/780

OTLP already supports bytes as a data type for Any value, see:
https://github.com/open-telemetry/opentelemetry-proto/blob/de4fc37940d39370194fb774e634ca408dacd865/opentelemetry/proto/common/v1/common.proto#L37

Byte arrays are an important case for unparsed, unstrusctured log data, so we are
formally adding them as a supported data type to the Log Data Model.

TODO: consider adding bytes to Trace API too.

### DIFF
--- a/specification/logs/data-model.md
+++ b/specification/logs/data-model.md
@@ -105,6 +105,8 @@ Value of type `any` can be one of the following:
 
 - A scalar value: number, string or boolean,
 
+- A byte array,
+
 - An array (a list) of `any` values,
 
 - A `map<string, any>`.


jmacd (approved):

kumoroku (approved):

maxgolov:
C++ SDK already has a draft support for byte arrays on Tracing API surface. Yay!

tigrannajaryan:
> C++ SDK already has a draft support for byte arrays on Tracing API surface. Yay!

@maxgolov if you are interested it may be worth for you to bring this up and advocate for adding it to the tracing API spec (in a backwards compatible way). Use cases, examples can help convince that we need it.

PR #1685, opened 2021-05-10
Add clarifications to SetStatus

### LINKED ISSUE #1584: Clarification required for Span.setStatus
One specific situation is currently absent from https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-status. Namely, imagine we have a web application running with Otel auto-instrumentation. Instrumentation creates a SERVER span for an incoming request. During this request, some exception is thrown, which sets response code to 400. But framework provides an exception callback, during which application developer retrieves the SERVER span from the currently active context and sets its status to `OK`, based on some application logic.

Now, when auto-instrumentation is going to end this SERVER span, then based on http semantic convention it MUST set spans status to `ERROR`. And spec referenced above allows it to do so. But I think the intention was to allow application developer or operator to override any automatic decision by manually setting span status to `OK`. Should this mean that auto-instrumentation MUST NOT override span status `OK`? Please note also, that currently there is no way to read span status from Otel API.

iNikem:
Fixes #1584

## Changes

Add the requirement to "freeze" span status after setting it to `Ok`.


jmacd (approved):

jsuereth (approved):

yurishkuro:
I think this solution introduced an internal contradiction. The argument against getters for status was that in streaming impl "current" status is undefined, yet this change explicitly requires the SDK to know about the previously set status.

iNikem:
> I think this solution introduced an internal contradiction. The argument against getters for status was that in streaming impl "current" status is undefined, yet this change explicitly requires the SDK to know about the previously set status.

It does not actually. I use "SHOULD", not "MUST". It is Ok for SDK not to fulfil this recommendation if they cannot. We already have sort of "backup" for this in [SetStatus API](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-status):

> Analysis tools SHOULD respond to an Ok status by suppressing any errors they would otherwise generate

yurishkuro:
> I use "SHOULD", not "MUST". It is Ok for SDK not to fulfil this recommendation if they cannot.

That doesn't help much. The point of API is portability across different SDKs, and here the semantic behavior of status could change via SDK without any changes to instrumentation. In general, I'd prefer to avoid "SHOULD" in the API as much as possible, since it indicates a feature you cannot rely on.

Aneurysm9 (approved):

specification/trace/api.md
```diff
@@ -552,6 +552,8 @@ The Span interface MUST provide:
 
 The status code SHOULD remain unset, except for the following circumstances:
```
SergeyKanzhelev, line 553: this sentence says **status code** should remain unset. Which may imply that description should be set, even for `Ok` status.

But this interpretation conflicts with the newly introduced rule that forbids the `Unset` status code.

Perhaps this sentence can be improved and **status code** replaced by Status. Maybe even explicitly say - do not try to set a description for `Ok` case

specification/trace/api.md
```diff
@@ -564,6 +566,9 @@ status code as `Unset` unless there is an error, as described above.
 
 Application developers and Operators may set the status code to `Ok`.
 
+When span status is set to `Ok` it SHOULD be considered final and any further
```
SergeyKanzhelev, line 569: I didn't get from reading the issue description why @tedsuo 's recommendation to forbid overriding anything except `Unset` wouldn't work. What if customer set status to `Error` with the proper description. Would instrumentation's `Description` make more sense than what customer set? Based on .NET POC https://github.com/open-telemetry/opentelemetry-dotnet/pull/2032 the auto instrumentation's Description wins, which probably is not desired.

There is also slight chance that customer want to set `Error` for `200 OK` response. Like [here](https://softwareengineering.stackexchange.com/questions/305250/should-i-use-http-status-codes-to-describe-application-level-events). I understand that based on current spec, instrumentation will not attempt to set status code. But will it override the `Description`?

SergeyKanzhelev (changes requested):
please see my comments. Sorry for being late for the issue discussion

specification/trace/api.md
iNikem: > do not try to set a description for Ok case

This is already forbidden by the existing specification:

> `Description` MUST only be used with the `Error` `StatusCode` value

and

> `Description` MUST be IGNORED for `StatusCode`
  `Ok` & `Unset` values

specification/trace/api.md
iNikem: > why @tedsuo 's recommendation to forbid overriding anything except Unset wouldn't work

Because application developer or operator must be able to override `Error` with `Ok`.

We don't have API to set status description without setting code. I always interpreted all rules in such a way that code+description are a whole atomic unit. If setting code should be ignored then description should not be changed as well. @SergeyKanzhelev  do you think this is not so obvious and requires explicit statement?

SergeyKanzhelev: > Because application developer or operator must be able to override `Error` with `Ok`.

So the scenario is that automatic instrumentation set if to `Error` and AFTER this customer wants to override it? Would the automatic instrumentation run last in most cases?

 > code+description are a whole atomic unit.

Totally. but than if we allow to override `Error` with another `Error` , we may be loosing an important `Description` that is coming from the user's code.

iNikem: > So the scenario is that automatic instrumentation set if to `Error` and AFTER this customer wants to override it? Would the automatic instrumentation run last in most cases?

The order does not matter. Any sequence of calls which has at least one `Ok` will result in `Ok`. Or I did not understand your question?

> Totally. but than if we allow to override Error with another Error , we may be loosing an important Description that is coming from the user's code.

Yes, two calls to `SetStatus` with `Error` code will result in the first one being ignored. This was in the spec even before this PR:

> Only the value of the last call will be recorded, and implementations are free to ignore previous calls.

SergeyKanzhelev: > > So the scenario is that automatic instrumentation set if to `Error` and AFTER this customer wants to override it? Would the automatic instrumentation run last in most cases?
> 
> The order does not matter. Any sequence of calls which has at least one `Ok` will result in `Ok`. Or I did not understand your question?

My question is whether the intention is to make user signal to be the most important one or the intention to make `Ok` to be a definitive `Ok`. If user signal is more important, I wonder if there ever be a situation when user signal comes AFTER auto-instrumented call. If not, we can just rely on `Unset`.

> > Totally. but than if we allow to override Error with another Error , we may be loosing an important Description that is coming from the user's code.
> 
> Yes, two calls to `SetStatus` with `Error` code will result in the first one being ignored. This was in the spec even before this PR:

I believe the user signal (user's provided `Descritption` is more important that one from auto instrumentation). So I think overriding `Error` with another `Error` is not ideal. Thus if the answer for the previous question is that user code is almost always comes before auto-instrumentation, we can just rely on `Unset` and solve the problem of overriding `Description`

yurishkuro:
After reading @SergeyKanzhelev 's comments, I am -1 on this change, because it tries to solve a multi-dimensional problem (is user intent more important, or is OK status more important) with a one-dimensional solution. I think this MUST NOT be enforced via API/SDK in any way, and the most recent call to SetStatus controls the value. The instrumentations SHOULD provide hooks to override span status.

SergeyKanzhelev:
> After reading @SergeyKanzhelev 's comments, I am -1 on this change, because it tries to solve a multi-dimensional problem (is user intent more important, or is OK status more important) with a one-dimensional solution. I think this MUST NOT be enforced via API/SDK in any way, and the most recent call to SetStatus controls the value. The instrumentations SHOULD provide hooks to override span status.

There is a second dimension solving the problem. Which is "Unset" is the same as "Ok". So I think the stated problem of auto-instrumentation can be solved by disallowing override value that was set explicitly already (assuming we are adding "state" to API to preserve the set/unset status). I am trying to understand the drawbacks and real life examples when last value should win.

yurishkuro:
My fear is that putting these non-trivial rules of status precedence is pretty easy to get wrong by not considering other possible use cases. The original ticket describes a very specific use case which has a very specific solution - provide a hook/mechanism in the framework to allow user to override the status.

iNikem:
> provide a hook/mechanism in the framework to allow user to override the status.

I don't understand this solution. What is "framework" here? Why should user use any addition hook, if there is already a totally obvious way for him to achieve their goal: to call `SetStatus` on a span.

iNikem:
And sorry for my "broken recording", but to me this situation again seems like "perfect is the enemy of good". Does this PR moves a spec into a right direction? Even a tiny bit? Even if this is not an ideal final solution? Does it prevents further steps in the direction of that ideal solution?

specification/trace/api.md
iNikem: > the intention to make Ok to be a definitive Ok

This one. This is a goal for spec clarification in this PR.

yurishkuro:
> I don't understand this solution. What is "framework" here?

The instrumentation library is the "framework". What if I don't like the span name that it assigns, or any other behavior, not just how it sets the status? Are we going to start building state machines for every property remembering "who set what"?

> Why should user use any addition hook, if there is already a totally obvious way for him to achieve their goal: to call SetStatus on a span.

Back in the days when I worked with derivatives trading UI we had each data element annotated with an extra metadata "set by user / set by default". If we must to support "user value overrides default value" that is the proper way of modeling it rather than encoding "who set this" meaning into the _values_ of the properties. But I don't believe it's necessary for OTEL because we can simply get away with "latest wins" behavior. If I, as a user of instrumentation library, do not like the defaults that instrumentation assigns, the library provides a hook for me to override the values as the last step of working with the span. That is the generic solution that solves the current issue without creating any complexities or making difficult to solve future use cases that we're not considering now.

Note that this whole approach of setting things on the span won't even work for client-side spans which are usually hidden inside the instrumentation library, so the hook is _the only_ way to affect those.

> And sorry for my "broken recording", but to me this situation again seems like "perfect is the enemy of good". Does this PR move a spec into a right direction? Even a tiny bit? Even if this is not an ideal final solution? Does it prevent further steps in the direction of that ideal solution?

In my opinion it moves the spec in the wrong direction. Every use case can be solved with an extra if-else branching, which could be an improvement for that specific use case but an overall detriment to the whole. Increasing complexity is never a good thing.

iNikem:
>  Increasing complexity is never a good thing.

And you think that requiring ALL instrumentation libraries, even written by not Otel maintainers, to provide such hooks is less complexity?? Comparing to one totally hidden inside SDK?

SergeyKanzhelev:
We already have complexity in the spec from the wording of "`Unset == Ok` and one should never set `Ok` unless it is very important to indicate that this is a final 'Ok'". 

I think it is unrealistic to ask libraries to provide this hook. Especially when we talk about byte code instrumentation. I also strongly against complicating logic by preferring `Ok` over `Error` and allowing to to override `Error` description. 

So we have a problem at hand - need to allow customer to override the Status of span that is collected via code injection. I think we can do either:

- Create a getter for `Status`.
- Set the simpler rule - `Unset` can be overridden, nothing else can.

iNikem:
> * Set the simpler rule - `Unset` can be overridden, nothing else can.

The spirit of the current spec is that user must be able to override anything with their `Ok`

iNikem:
Current specification leads to actual bugs for our users, as described in #1584. I think that this proposal solves that problem and still preserves the opportunity for future improvements. Rejecting this PR without any alternative suggestions (expressed as PRs) will leave our users without any solution.

HaloFour:
In my opinion the "hook" approach, while simpler on the spec, would be much more cumbersome to implement and work with.  How would a developer tie into such a hook?  Where would that live?  How far removed would it be from the logic actually trying to determine the appropriate status of the span?  What information would such a hook have to work with?  And that doesn't even touch on how such a hook might be implemented in one of multiple possible instrumentation libraries or whether they can be implemented in a consistent manner.

As a developer any such hook that I would write would, at best, seek to implement the logic as described in this proposal: that instrumentation doesn't override the status on a span where the status has already been already set.  Whether that be accomplished via first-wins or some means to detect if the status has already been set doesn't matter to me so much, although the latter makes more sense to me.  But it seems like a common enough goal that facilitating it through a side channel would be, in my opinion, too much of a burden on developers who need a simple escape hatch.

carlosalberto:
> In my opinion the "hook" approach, while simpler on the spec, would be much more cumbersome to implement and work with.

I agree with this. It was even discussed that the related prototype added enough complexity to mark as undesirable. The current approach, even imperfect, is a good step in the right direction IMHO.

yurishkuro:
@jkwatson

> The backend in a hypothetical streaming SDK can handle this case easily by ignoring any updates to the status after an OK has been received. The backend has state just like an in-process SDK has state. It does not require the in-process streaming implementation to keep that state like a 'get' would.

Streaming SDKs are not hypothetical, I work with a bunch. And the transports delivering events to the backend are much more expensive (if available at all, in multi-zone scenarios) if they must guarantee the order of delivery. So in practice this change would require the SDK to keep the state.

jkwatson:
> @jkwatson
> 
> 
> 
> > The backend in a hypothetical streaming SDK can handle this case easily by ignoring any updates to the status after an OK has been received. The backend has state just like an in-process SDK has state. It does not require the in-process streaming implementation to keep that state like a 'get' would.
> 
> 
> 
> Streaming SDKs are not hypothetical, I work with a bunch. And the transports delivering events to the backend are much more expensive (if available at all, in multi-zone scenarios) if they must guarantee the order of delivery. So in practice this change would require the SDK to keep the state.

Interesting. I'm not an expert here, and we're ranging far afield, but isn't it pretty cheap to include a Lamport clock field in your payloads to guarantee eventual ordering consistency?

yurishkuro:
The simplest solution is to have a sequence number (local to the span) for each event, but it boils down to the same "stateful" requirement & synchronization overhead.

jkwatson:
> The simplest solution is to have a sequence number (local to the span) for each event, but it boils down to the same "stateful" requirement & synchronization overhead. 

In your streaming SDKs, how do you deal with the same attribute being set more than once, then?

jkwatson:
@yurishkuro I don't think your backend would need to care about ordering with this proposal. If it *ever* sees an OK, no matter what the order it arrived in, then you can ignore everything else. 

So, if the true local order was "ERROR", "OK", "ERROR", but it arrived as "OK", "ERROR", "ERROR", or "ERROR", "ERROR", "OK", then any way you resolve it, you end up with "OK", which is the correct answer.

yurishkuro:
And what happens when there is a user who says "I don't like that default instrumentation is setting the status to OK, I really want to set it to ERROR"?

This is why I am opposed to this change - a SetStatus method should just record the value, not have surprising behavior where sometimes it works, sometimes it doesn't, based on an internal state that nobody can access. If we genuinely require _priorities_ attached to status values, then priorities should be modeled explicitly. E.g. we could introduce SetDefaultStatus method to be used by instrumentation, and let the end user call SetStatus which will always override default status, regardless of the actual value.

jkwatson:
> And what happens when there is a user who says "I don't like that default instrumentation is setting the status to OK, I really want to set it to ERROR"?
> 
> This is why I am opposed to this change - a SetStatus method should just record the value, not have surprising behavior where sometimes it works, sometimes it doesn't, based on an internal state that nobody can access. If we genuinely require _priorities_ attached to status values, then priorities should be modeled explicitly. E.g. we could introduce SetDefaultStatus method to be used by instrumentation, and let the end user call SetStatus which will always override default status, regardless of the actual value.

Instrumentation should never set the status to OK. This has been clarified before, I believe.

iNikem:
> then priorities should be modeled explicitly

@yurishkuro We have already did that. Priorities are modeled already in the spec: `Ok` overwrites anything else, `Error` overwrites `Unset`. It's already written in the spec since we reduced 26 statuses to just 3. What else is needed, in your opinion, @yurishkuro ?

iNikem:
> E.g. we could introduce SetDefaultStatus method to be used by instrumentation, and let the end user call SetStatus

How do you imagine to have these 2 methods and be 100% sure that the end-user will call the intended one?

yurishkuro:
> We have already did that. Priorities are modeled already in the spec: Ok overwrites anything else, Error overwrites Unset. 

ok, you convinced me. I would make this ^ very explicit right where status codes are introduced, instead of spreading it around in text. Also make it explicit that this priority order must always exist, meaning if for some reason we decide to add a 4th status code in the future, it MUST be slotted somewhere with a well defined priority.

carlosalberto:
@SergeyKanzhelev Anything pending to resolve?

iNikem:
@yurishkuro I added explicit mention of priorities: https://github.com/open-telemetry/opentelemetry-specification/pull/1685/commits/7fd0fcc9f4fbf03e2335a570e22d162b9a067817. PTAL and approve.

jkwatson (approved):

yurishkuro (approved):

carlosalberto:
Ping @SergeyKanzhelev - this PR has resolved a few issues and has more approvals. Please re-review.

SergeyKanzhelev:
the logic os `SetStatus` still has issue of customer provided error description will be overriden by the automatic instrumentation's description. But this PR does not address this question or makes things worse.

SergeyKanzhelev (approved):

PR #1699, opened 2021-05-13
add note about reporting dropped attributes/events/links

### LINKED ISSUE #1698: No mention of dropped attributes/events/links counts in SDK spec
The exporter spec defines how to handle DroppedAttributesCount and DroppedEventsCount (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/non-otlp.md#dropped-attributes-count) at export time, but it isn't clear from the SDK spec that these should be reported.

codeboten:
Fixes #1698 

## Changes

Clarify that dropped counts should be reported from the SDK. Would love feedback from reviewers on whether or not this change should include a change to the matrix.

Related issues #1656

### DIFF
--- a/specification/trace/sdk.md
+++ b/specification/trace/sdk.md
@@ -356,6 +356,9 @@ There SHOULD be a log emitted to indicate to the user that an attribute, event,
 or link was discarded due to such a limit. To prevent excessive logging, the log
 should not be emitted once per span, or per discarded attribute, event, or links.
 
+Attributes, events and links dropped due to collection limits MUST be reported at export
+time.
+
 ## Id Generators
 
 The SDK MUST by default randomly generate both the `TraceId` and the `SpanId`.
--- a/specification/trace/sdk.md
+++ b/specification/trace/sdk.md
@@ -92,6 +92,10 @@ Thus, the SDK specification defines sets of possible requirements for
   (some languages might implement this by having an end timestamp of `null`,
   others might have an explicit `hasEnded` boolean).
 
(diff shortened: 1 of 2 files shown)


specification/trace/sdk.md
```diff
@@ -356,6 +356,10 @@ There SHOULD be a log emitted to indicate to the user that an attribute, event,
 or link was discarded due to such a limit. To prevent excessive logging, the log
 should not be emitted once per span, or per discarded attribute, event, or links.
 
+Counts for attributes, events and links dropped due to collection limits MUST be reported
```
Oberon00, line 359: I think this MUST is too much, as it depends on the protocol. What we should say is that the counts MUST be available for exporters. I think this would best fit in the "readable span" wording, somewhere around https://github.com/open-telemetry/opentelemetry-specification/pull/1699/files#diff-ace73543d690260e873d774428419c23895283b6c31177d405816b154d1f5aa0R91

codeboten: Thanks @Oberon00, I updated the note, PTAL.

Oberon00: I think it would be best to move this wording to the "readable span" definition, somewhere around line 91, as this will probably be the part where this wording has to be implemented. But semantically, the change looks good to me!

carlosalberto (approved):

specification/trace/sdk.md
carlosalberto: @codeboten Could you update this part according to @Oberon00 ?

codeboten: Done!

Issue #1432, opened 2021-02-11
Metric (observer) collections should be independent from metric export


Oberon00:
(the metric spec is still very unfinished, so this might already be included in one of the more general metric spec issues)

Consider this use case:
- I want to collect JVM memory metrics (e.g. currently committed memory bytes) every 10 seconds
- I want to export this (and all other metrics) aggregated (min/max/avg) every six collections i.e. every minute.

I think the spec currently assumes (or assumed before it was deleted) that metric collection and export happens in the same step, but this seems overly restrictive.

What is more, it would make sense to have each observer instrument run potentially on it's own schedule. E.g. some metrics might profit more from more granularity than others, some might be expensive to retrieve, others cheap.

(This paragraph might warrant it's own issue:) A good implementation of this would also ensure that collections whose intervals are a proper fraction or multiple of the export internval do not in-deterministically move between exports. For example, in the above JVM use case, an implementation that does not do any synchronization between export and collection will have a race between every 6th collection of metrics and export which are both supposed to happen at T=60. Such an implementation would sometimes send an aggregate of only 5 and then the next time up to 7 collections. This might be undesirable and should be avoidable without much performance overhead.

bogdandrutu:
@Oberon00 who is "I" in this example "instrumentation author"? "application owner"?

Oberon00:
@bogdandrutu very good question actually 🤔 The two "I" might not even be the same. For collection interval, we are interested in:

1. How fluctuating is this metric actually (CPU usage vs number of loaded classes) -- Usually the instrumentation author has a good grasp on this, but the application owner might want to influence this e.g. to debug specific problems.
2. How expensive is it to collect the metric -- Usually the instrumentation author has a good estimate here too, but it might depend on the deployment situation, etc. (e.g. what if we are deployed to a machine with 256 cores -- collecting CPU usage might become unexpectedly expensive)
3. How interesting is this metric for the use case of the application owner

For export interval, the instrumentation author has no clue. I think this is more for the application owner to decide (that would be a typical environment variable setting). The backend's current/usual load factor could also be a factor here, as you can't export so often that you overwhelm it.

A recommended minimum export interval can also be calculated as the greatest common divisor of all collection intervals by the SDK which knows all registered observers. However, that would often be 1, if we do not provide some guidance on selecting the intervals.

[...]

PR #1703, opened 2021-05-14
Add details to Asynchronous Gauge API

reyang:
Follow up #1617.

## Changes

Added details to the Asynchronous Gauge instrument. In #1617 these were left as TODO.

### DIFF
--- a/specification/metrics/api.md
+++ b/specification/metrics/api.md
@@ -619,7 +619,7 @@ def cpu_frequency_callback():
         (0.56, ("cpu", 1), ("core", 1)),
     )
 
-meter.create_observable_counter(
+meter.create_observable_gauge(
     name="cpu.frequency",
     description="the real-time CPU clock speed",
     callback=cpu_frequency_callback,
--- a/specification/metrics/api.md
+++ b/specification/metrics/api.md
@@ -637,7 +637,7 @@ def cpu_frequency_callback(result):
     result.Observe(0.57, ("cpu", 1), ("core", 0))
     result.Observe(0.56, ("cpu", 1), ("core", 1))
 
-meter.create_observable_counter(
+meter.create_observable_gauge(
     name="cpu.frequency",
     description="the real-time CPU clock speed",
     callback=cpu_frequency_callback,
--- a/specification/metrics/api.md
+++ b/specification/metrics/api.md
@@ -551,7 +551,107 @@ Example uses for Asynchronous Gauge:
 
 #### Asynchronous Gauge creation
 
-TODO
+There MUST NOT be any API for creating an Asynchronous Gauge other than with a
+[`Meter`](#meter). This MAY be called `CreateObservableGauge`. If strong type is
+desired, the client can decide the language idiomatic name(s), for example
+`CreateUInt64ObservableGauge`, `CreateDoubleObservableGauge`,
+`CreateObservableGauge<UInt64>`, `CreateObservableGauge<double>`.
+
+It is highly recommended that implementations use the name `ObservableGauge`
+(or any language idiomatic variation, e.g. `observable_gauge`) unless there is
+a strong reason not to do so. Please note that the name has nothing to do with
+[asynchronous
+pattern](https://en.wikipedia.org/wiki/Asynchronous_method_invocation) and
+[observer pattern](https://en.wikipedia.org/wiki/Observer_pattern).
+
+The API MUST accept the following parameters:
+
+* The `name` of the Instrument, following the [instrument naming
+  rule](#instrument-naming-rule).
+* An optional `unit of measure`, following the [instrument unit
+  rule](#instrument-unit).
+* An optional `description`, following the [instrument description
+  rule](#instrument-description).
+* A `callback` function.
+
+The `callback` function is responsible for reporting the
+[Measurement](#measurement)s. It will only be called when the Meter is being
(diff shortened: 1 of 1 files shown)


specification/metrics/api.md
```diff
@@ -551,7 +551,107 @@ Example uses for Asynchronous Gauge:
...
+def cpu_frequency_callback():
+    # Note: in the real world these would be retrieved from the operating system
+    return (
+        (3.38, ("cpu", 0), ("core", 0)),
+        (3.51, ("cpu", 0), ("core", 1)),
+        (0.57, ("cpu", 1), ("core", 0)),
+        (0.56, ("cpu", 1), ("core", 1)),
+    )
+
+meter.create_observable_counter(
```
cijothomas, line 622:
```suggestion
meter.create_observable_gauge(
```

specification/metrics/api.md
```diff
@@ -551,7 +551,107 @@ Example uses for Asynchronous Gauge:
...
+# Python
+
+def cpu_frequency_callback(result):
+    # Note: in the real world these would be retrieved from the operating system
+    result.Observe(3.38, ("cpu", 0), ("core", 0))
+    result.Observe(3.51, ("cpu", 0), ("core", 1))
+    result.Observe(0.57, ("cpu", 1), ("core", 0))
+    result.Observe(0.56, ("cpu", 1), ("core", 1))
+
+meter.create_observable_counter(
```
cijothomas, line 640:
```suggestion
meter.create_observable_gauge(
```

specification/metrics/api.md
```diff
@@ -551,7 +551,107 @@ Example uses for Asynchronous Gauge:
...
+    unit="GHz",
+    value_type=float)
+```
+
+```csharp
+// C#
+
+// A simple scenario where only one value is reported
+
+meter.CreateObservableCounter<double>("temperature", () => sensor.GetTemperature());
```
cijothomas, line 653:
```suggestion
meter.CreateObservableGauge<double>("temperature", () => sensor.GetTemperature());
```

cijothomas (approved):

bogdandrutu (approved):

specification/metrics/api.md
```diff
@@ -551,7 +551,107 @@ Example uses for Asynchronous Gauge:
...
+* An optional `description`, following the [instrument description
+  rule](#instrument-description).
+* A `callback` function.
+
+The `callback` function is responsible for reporting the
+[Measurement](#measurement)s. It will only be called when the Meter is being
+observed. Individual language client SHOULD define whether this callback
+function needs to be reentrant safe / thread safe or not.
+
+The callback function SHOULD NOT take indefinite amount of time. If multiple
```
jsuereth, line 582: Are there any rules about how many times per-exporter this is called?

specification/metrics/api.md
```diff
@@ -551,7 +551,107 @@ Example uses for Asynchronous Gauge:
...
+independent SDKs coexist in a running process, they MUST invoke the callback
+function(s) independently.
+
+Individual language client can decide what is the idiomatic approach. Here are
+some examples:
+
+* Return a list (or tuple, generator, enumerator, etc.) of `Measurement`s.
+* Use an observer argument to allow individual `Measurement`s to be reported.
```
jsuereth, line 590: i think "observer argument" is a little "loose" here.  It may make sense to provide an example in one langauge of what this means and denote languages are free to use the best paradigm to provide this.

specification/metrics/api.md
bogdandrutu: It should be called once per export cycle. That is my understanding.

specification/metrics/api.md
reyang: A Python example is provided at Ln:633 in this PR. On Ln:586 we've mentioned that "language client can decide what is the idiomatic approach".

specification/metrics/api.md
reyang: I think we don't put restriction in the API spec. The SDK spec might put additional requirements/explanation.
I would say by default it follows what @bogdandrutu mentioned.
In some situation, the user might decide to collect the data at a higher frequency but report it on lower frequency (e.g. configure the SDK to gather the data from callbacks every 5 seconds, but only export the data every one minute - refer to #1432).

bogdandrutu: But even if the user wants to do that, they need to schedule their own timer then. The valid SDK implementation should call the callback once every export cycle I think

bogdandrutu: Maybe call it "collection" cycle for the moment. Until we define if collection cycle == exporter cycle.

specification/metrics/api.md
jmacd:
```suggestion
* Use an observer result argument to allow individual `Measurement`s to be reported.
```

jmacd (approved):

specification/metrics/api.md
reyang: Thanks @jmacd. I've created #1715, will update the other APIs in a separate PR.

SergeyKanzhelev (approved):

SergeyKanzhelev:
@jsuereth please comment if you still has concerns. Otherwise I will merge in a few hours

SergeyKanzhelev:
@jsuereth if there are outstanding comments, let's resolve them as issues

PR #1710, opened 2021-05-18
Use service.namespace in Jager's service name if present

bogdandrutu:
OpenQuestion: There is a note in the https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions#service about not concatenating, but I think (🤔) for cases where only one value is passed like Jaeger it would be important to concatenate them.

### DIFF
--- a/specification/trace/sdk_exporters/jaeger.md
+++ b/specification/trace/sdk_exporters/jaeger.md
@@ -51,10 +51,14 @@ OpenTelemetry resources MUST be mapped to Jaeger's `Span.Process` tags. Multiple
 single process and exporters need to handle this case accordingly.
 
 Critically, Jaeger backend depends on `Span.Process.ServiceName` to identify the service
-that produced the spans. That field MUST be populated from the `service.name` attribute
-of the [`service` resource](../../resource/semantic_conventions/README.md#service).
-If no `service.name` is contained in a Span's Resource, that field MUST be populated from the
+that produced the spans. This field MUST be populated from the `service.namespace`
+and `service.name` attributes of the [`service` resource](../../resource/semantic_conventions/README.md#service):
+
+* If `service.name` is not present, Jaeger's field MUST be populated from the
 [default](../../resource/sdk.md#sdk-provided-resource-attributes) `Resource`.
+* If `service.namespace` is not present, Jaeger's field MUST be populated from the `service.name` attribute.
+* If both `service.namespace` and `service.name` attribute are present then that field
+MUST be populated as `{service.namespace} + "/" + {service.name}`
 
 ### IDs
 


bogdandrutu:
Some extra comments related to this https://github.com/open-telemetry/opentelemetry-specification/pull/1677#discussion_r634523040

Oberon00:
Good point, but I think we should change Zipkin at the same time (there is no reason for this to apply only to Jaeger). Actually, I just noticed we have a document https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/non-otlp.md, probably this should be moved there.

The semantic convention should also be changed to mention the preferred separator in cases where concatenation is needed.

specification/trace/sdk_exporters/jaeger.md
```diff
@@ -51,10 +51,14 @@ OpenTelemetry resources MUST be mapped to Jaeger's `Span.Process` tags. Multiple
...
 Critically, Jaeger backend depends on `Span.Process.ServiceName` to identify the service
-that produced the spans. That field MUST be populated from the `service.name` attribute
-of the [`service` resource](../../resource/semantic_conventions/README.md#service).
-If no `service.name` is contained in a Span's Resource, that field MUST be populated from the
+that produced the spans. This field MUST be populated from the `service.namespace`
+and `service.name` attributes of the [`service` resource](../../resource/semantic_conventions/README.md#service):
+
+* If `service.name` is not present, Jaeger's field MUST be populated from the
 [default](../../resource/sdk.md#sdk-provided-resource-attributes) `Resource`.
+* If `service.namespace` is not present, Jaeger's field MUST be populated from the `service.name` attribute.
```
Oberon00, line 59: These points are a bit hard to read. IMHO a better structure would be:

* If service.name and service.namespace, ...
* Otherwise if just service.name, ...
* Otherwise (irrespective of service.namespace), ...

Also, the current wording leaves the case where service.namespace is present but service.name is not undefined. With the suggestion above it would be ignored but using `{service.namespace}/{default service name}` might make sense.

specification/trace/sdk_exporters/jaeger.md
```diff
@@ -51,10 +51,14 @@ OpenTelemetry resources MUST be mapped to Jaeger's `Span.Process` tags. Multiple
...
+* If both `service.namespace` and `service.name` attribute are present then that field
+MUST be populated as `{service.namespace} + "/" + {service.name}`
```
tigrannajaryan, line 61: I think this is not a bad idea but would be good to hear from Jaeger experts. Maybe there is a commonly accepted way to model service namespaces or teams this in Jaeger? @yurishkuro do you know?

yurishkuro: Jaeger backend does not support the notion of namespaces. Does any backend do? What is the specific functionality that applies to namespace only and not the FQ service name?

Back at Uber we had several thousand services and did not come across a need for a namespace as a first class entity, although various teams did use certain prefixes for service names like jaeger_indexer, jaeger_ingester.

Regarding the concatenation, using slash makes many Jaeger URLs quite ugly since slash must be URL-encoded.

SergeyKanzhelev (changes requested):
Let's change it for Zipkin and also for resource labels description here: https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions#service

this kind of inconsistency is bothersome

PR #1716, opened 2021-05-20
Fix a few typos in the datamodel specification

pirgeo:
## Changes

- Fix a few typos
- Update the references pointing to metric point types in the proto and fix the proto version in the reference. I am not sure if there is a more elegant way of referencing the proto at a specific line while still pointing at the latest proto version. Doing it this way, the reference will always point at the right line. On the flip side, the references have to be updated every time a new version of the proto is released. Is there a preferred way to do this?

### DIFF
--- a/specification/metrics/datamodel.md
+++ b/specification/metrics/datamodel.md
@@ -227,9 +227,9 @@ same kind. <sup>[1](#otlpdatapointfn)</sup>
 
 The basic point kinds are:
 
-1. [Sum](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L200)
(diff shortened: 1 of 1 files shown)


arminru (approved):
Thanks!
I think pointing to proto@v0.9.0 is fine to have stable references, let's see what the @open-telemetry/specs-metrics-approvers say.

tigrannajaryan:
> Thanks!
> I think pointing to proto@v0.9.0 is fine to have stable references, let's see what the @open-telemetry/specs-metrics-approvers say.

I think referring to a specific version is a safer approach. If we point to main and it changes we need to update the references anyway. If updating the references is a requirement to keep the spec correct then I would rather link to specific versions and ensure it is never broken, just maybe outdated if a newer version is released (which I think is less bad).

tigrannajaryan (approved):

jsuereth (approved):

reyang (approved):
LGTM.

PR #1727, opened 2021-05-27
Add guidelines on using log Body vs Attributes (#1613)

### LINKED ISSUE #1613: Distinction between structured `Body` and `Attributes`
Is there a clear distinction between what belongs in a structured `Body` vs. in `Attributes`? If there's not, it seems to make it less predictable where data is expected, and how to map between other models.

A top-level "message" string seems to be pretty common among other log data models. In this model, if you want to have a top-level message string and structured data describing the event, are you supposed to put the message in `Body` and other data that would otherwise be in `Body` in `Attributes`? I don't see anything about special-casing a property of `Body` as a top-level message.

From the mapping perspective, considering the Elastic Common Schema (ECS) example, it doesn't include all ECS fields, but `message` is the only one shown as mapping to `Body`. `Body` seems like a logical (perhaps the most logical) place for fields like `error.message` or `event.id`. So the fact that fields like `error.message` are shown as mapping into `Attributes` makes it seem that there's not a clear distinction, and that the mapping could be ambiguous depending on whether `message` is populated.

(BTW, unlike most of the rest of the document, the ECS example refers to "body" and "attributes" rather than "Body" and "Attributes", which I assume is a typo or holdover from a previous version.)

---

`Attributes` is documented as:

> SHOULD follow OpenTelemetry semantic conventions for Attributes.

If I understand correctly, that means that a property of `Attributes` that has the name of a well-known attribute should have the meaning and data type defined for that attribute, but meanwhile `Attributes` can also include arbitrary custom attributes.

That also seems to have implications for placement of data within `Body` vs. `Attributes`, because whereas no semantic conventions apply to `Body`, if a property gets bumped from `Body` into `Attributes`, then it may be that conventions are supposed to apply that wouldn't otherwise.

pmm-sumo:
Fixes #1613

## Changes

Provides guidelines on using `Body` vs `Attributes` in Logs Data Model

### DIFF
--- a/specification/logs/data-model.md
+++ b/specification/logs/data-model.md
@@ -23,6 +23,7 @@
   * [Field: `Body`](#field-body)
   * [Field: `Resource`](#field-resource)
   * [Field: `Attributes`](#field-attributes)
+  * [Guidelines on using `Body` vs `Attributes`](#guidelines-on-using-body-vs-attributes)
 * [Example Log Records](#example-log-records)
 * [Appendix A. Example Mappings](#appendix-a-example-mappings)
   * [RFC5424 Syslog](#rfc5424-syslog)
@@ -422,6 +423,35 @@ information about the request context (other than TraceId/SpanId). SHOULD follow
 OpenTelemetry
 [semantic conventions for Attributes](../trace/semantic_conventions/README.md).
 This field is optional.
+
+### Guidelines on using `Body` vs `Attributes`
+
+If the log record contains one or more "pieces" of data that may fit either
+in `Body` or `Attributes` but do not fit the description of the other top-level
+fields (`Timestamp`, `Severity`, etc) then follow these guidelines to decide how
+to record these pieces of data in the `Body` and `Attribute` fields:
+
+1. If the log record contains a single raw byte sequence that is not associated
+   with any particular key (such that we cannot think of it as a key/value pair)
+   put it in the `Body`.
+2. If the log record contains a single raw Unicode character string that is not
+   associated with any particular key then put it in the `Body`.
+3. If the log record contains a single JSON-like array value or a primitive value
+   that is not associated with any particular key then put it in the Body.
+4. If the log record contains a single JSON-like object with key/value pairs it may
+   be recorded either in the Body or in the Attributes. The choice in this case
+   should be based on whether we are interested in individual key/value pairs
+   (e.g. each key/value has a particular meaning, possibly independently from the
+   meaning of the other key/value pairs), in which case recording them in the
+   `Attributes` would be preferable. Otherwise if we are more interested in just
+   recording the object as a whole and getting it delivered then the `Body` would
+   be preferable.
+5. If the log record contains a piece of data that is associated with a key then
+   record that piece as a key/value pair in the `Attributes`.
+
+Note: if there is more than one piece of data that matches the rules 1-4 then we
+cannot record then in the Body, we have to come up with some keys and record each
+piece of data as a key/value pair in the Attributes.
 
 ## Example Log Records
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@ release.
 
 ### Logs
 
+- Adds guidelines on using `Body` vs `Attributes` in Logs Data Model.
+  ([#1727](https://github.com/open-telemetry/opentelemetry-specification/pull/1727))
+
 ### Resource
 
 - Add schema_url support to `Resource`. ([#1692](https://github.com/open-telemetry/opentelemetry-specification/pull/1692))


tigrannajaryan (approved):

specification/logs/data-model.md
```diff
@@ -423,6 +424,35 @@ OpenTelemetry
 [semantic conventions for Attributes](../trace/semantic_conventions/README.md).
 This field is optional.
 
+### Guidelines on using `Body` vs `Attributes`
+
+If the log record contains one or more "pieces" of data that may fit either
```
yurishkuro, line 429: Sorry I am late to the party, but I feel the description below makes things waaay more complicated that it needs to. This is also the first time in this document that the term "log record" is used - what is that referring to?

The original issue seem to boil down to a question: if the log message (in the sense of classical unstructured logging APIs) contains structured data (like JSON), should it be recorded in the Body or broken down into Attributes?  I would contend that it should be kept in the Body, _because it comes from unstructured logging API_. If, however, a structured logging API is used, then the question becomes moot. One exception is when the logging API is too structured to the point where it does not have a "message" part, in which case a pre-agreed attribute name might be used to extract the value from the Attribute and move into Body.

tigrannajaryan: > This is also the first time in this document that the term "log record" is used - what is that referring to?

"log record" is used through the data model document. It refers to a single log item, a single event.

>Sorry I am late to the party, but I feel the description below makes things waaay more complicated that it needs to.

I am primarily responsible for the complication :-) Can you please comment on the specific items which you feel are excessive or can be deleted/rephrased? I would also want to have more succinct guidelines but I failed to come up with something shorter that is also not too ambiguous.

>if the log message (in the sense of classical unstructured logging APIs) contains structured data (like JSON), should it be recorded in the Body or broken down into Attributes?

This is one other cases and point 4 attempts to help with that. I believe I heard people being confused in some other cases as well, hence points 1-3.

yurishkuro: > "log record" is used through the data model document. It refers to a single log item, a single event.

log records in the examples below (L466) are already highly structured, I am not sure they pose the same challenge/confusion. And the way "log record" is introduced in L125 refers to OTEL data model, whereas this new section uses "log record" to refer to _something else_ that is being translated into OTEL model, unless I am missing something.

tigrannajaryan: Our glossary says:
>Log Record - A recording of an event. Typically the record includes a timestamp indicating when the event happened as well as other data that describes what happened, where it happened, etc.

I do not think it describes exclusively an Otel log record, but is a generic definition of what this spec thinks a log record is. I may be wrong, so if you feel we need to use a different wording to refer to non-Otel log records then maybe we need one, however I am struggling with finding such wording. Maybe we can say "non-Otel log data"?

yurishkuro: I think my point is that non-OTEL log record are either fully structured, like the ones from Zap, or partially structured, like those from log4j (which have certain structured fields like ts & log level, but free form text message). And the rule is:
  * a structured field
    * if it maps to OTEL model directly, it goes into dedicated OTEL field
    * otherwise it goes to Attributes
  * an unstructured blob goes to Body

pmm-sumo: The guidelines come largely to address the questions that were being raised recently. I think the problem can be roughly expressed as:  "since `Body` can be structured, in what circumstances should I put structured data there (and not into `Attributes`)?"

The recommendation we provide is largely like the above (use `Attributes`) except the case when one is "more interested in just  recording the object as a whole and getting it delivered" when "the `Body` would be preferable". This gives good amount of flexibility, which I think is a good thing since there are so many logs use-cases 

If we would not want to allow that, then another question would arise, which might be actually harder to answer: which types should be allowed for `Body` (and why)?

yurishkuro: > "since Body can be structured, in what circumstances should I put structured data there (and not into Attributes)?"

(a) I don't believe this PR answers that, (b) I don't believe it's the place of the spec to actually try to answer that. I dislike that we're discussing some abstract situations with "log records" without specifying how those records are produced. Structured logging starts with structured API. If the API is not structured (like log4j), then I consider the Body to be unstructured, regardless of whether the user encoded a structured JSON in there. Even if the logging backend (like ES) decides to parse that JSON, it's not the business of the protocol / data model to deal with that.

pmm-sumo: That's a fair argument. I wonder if we should enforce that the only allowed type for `Body` is `string` (since this is what this implies, If I read it correctly?) I like the flexibility `Body` provides via `any`, but then maybe it's not really needed? 

The conclusion, until now, was to leave it to the user. Perhaps the only case when `Body` was filled with anything other than `string` is via [filelogreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/filelogreceiver). We actually had a discussion on that [some time ago](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/2851#issuecomment-809547738)

tigrannajaryan: >That's a fair argument. I wonder if we should enforce that the only allowed type for Body is string (since this is what this implies, If I read it correctly?) I like the flexibility Body provides via any, but then maybe it's not really needed?

At the minimum I think we need `bytes` too. However, what do we do with an arbitrary JSON payload coming from some log source if we only allow `string` and `bytes`? I think `any` is useful.

yurishkuro: > However, what do we do with an arbitrary JSON payload coming from some log source

imo that's not the concern of the protocol. If the process generating the OTLP log record knows that it's a JSON payload AND wants to deconstruct that payload into structured representation using `any`, they can do that, but writing JSON as a string is also ok, they shouldn't be forced into expensive deconstruction.

tigrannajaryan: > > However, what do we do with an arbitrary JSON payload coming from some log source
> 
> imo that's not the concern of the protocol. If the process generating the OTLP log record knows that it's a JSON payload AND wants to deconstruct that payload into structured representation using `any`, they can do that, but writing JSON as a string is also ok, they shouldn't be forced into expensive deconstruction.

I agree. I am saying it is useful to have the option to deconstruct and represent as structured data instead of JSON string (especially if it is already in a structured form in-memory). No forcing, I agree.

yurishkuro: well, the current PR, if not exactly "forcing", is trying to prescribe what the component should do in this case, and that's why I think it's an over-reach.

tigrannajaryan: @yurishkuro do you think there is a better place for this sort of guidelines? We get the question once in a while so it would be useful to have the guidelines somewhere in Otel.

yurishkuro: I am not in favor of the type of guidance as currently expressed in the PR to be in the spec at all. It's not a question of where to put it.

I have re-read the original ticket #1613 and I fail to see what the confusion is in the first place. So I don't know what problem we're trying to address with this PR.

pmm-sumo:
I opened another issue where I bring the idea of limiting the types allowed for `Body`: https://github.com/open-telemetry/opentelemetry-specification/issues/1752

specification/logs/data-model.md
tigrannajaryan: Here is another example of the confusion: https://cloud-native.slack.com/archives/CJFCJHG4Q/p1627987975028000

Issue #1426, opened 2021-02-10
Inconsistency of configuring secured OTLP export


puckpuck:
**What are you trying to achieve?**

Configure the SDK from multiple languages to send traces to a secured GRPC endpoint.

**Additional context.**

The specification, and docs for the Java Instrumentation agent indicates to specify the OTLP endpoint with the `http://` or `https://` scheme. This is odd because the communication is done using GRPC by default, in which case the scheme should follow the `grpc://` or `grpcs://` format.

To add further confusion, the JS SDK, will return a DNS error if the endpoint has the `http://` or `https://` scheme. 

Configuring the GRPC export options for secured communication is also inconsistent across languages.  In Java you set a useTLS flag to true, but other languages like Go or Js require a nil or empty ssl credentials object to be created and passed in.

I'm hoping to get 2 things accomplished.
1. Allow for the use of a proper scheme for OTLP export (grpc rather than http)
2. Make specifying secured usage of OTLP easy and consistent across languages

alanwest:
Thanks for raising this @puckpuck.

> 1. Allow for the use of a proper scheme for OTLP export (grpc rather than http)

To my knowledge, there is no such `grpc://` or `gprcs://` scheme. gRPC is often performed over HTTP/2 but can be over other transports like unix domain sockets.

> 2. Make specifying secured usage of OTLP easy and consistent across languages

Agreed. I've noted the inconsistencies as well [here](https://github.com/open-telemetry/opentelemetry-collector/issues/2539#issuecomment-812696546) in the context of the collector's OTLP exporter.

puckpuck:
The confusion of using `http://` to define the use of GRPC is because OTLP also supports HTTP transport on a different port by default in the Collector.  There are also some SDKs that only support HTTP transport and not GRPC (Browser JS). Using a scheme here would be clear in which transport is being used for OTLP.

Though I don't believe there is an industry standard, I have certainly seen multiple cloud providers use the `grpc://` syntax when describing an API endpoint that can receive data in both http and grpc forms.  Here is Google's for creating a GRPC based API endpoint discussing the use of a `grpc://` syntax: https://cloud.google.com/endpoints/docs/grpc-service-config/reference/rpc/google.api#backendrule

obecny:
in js we are removing the protocol from the link when using `@opentelemetry/exporter-collector-grpc`. This means it will work in both cases whether you provide http(s) or not

PR #1729, opened 2021-05-27
Split out OTLP gRPC vs. HTTP endpoint configuration

alanwest:
Here's a can of worms some folks may have hoped would just go away 😄.

## Background

The specification of the endpoint configuration for the OTLP exporter has caused some confusion among users due to inconsistencies with the implementations of the language SDKs as well as the collector.

### Related issues
* https://github.com/open-telemetry/opentelemetry-specification/issues/1426
* https://github.com/open-telemetry/opentelemetry-collector/issues/2539

### Prior context

In a prior version of the specification, the endpoint configuration did not mandate it be a URL with scheme included. There also used to be an `insecure` configuration flag which was removed. See #1234 where these changes were made. It seems that these changes were originally intended for the benefit of making the configuration of OTLP/HTTP exporters more natural, yet the discussion on this PR seems to suggest that folks accepted that the changes would equally apply to OTLP/gRPC exporters.

### Existing implementations

I have studied a few of the OTLP/gRPC exporter implementations. Please correct me on any of the observations I have made - I am not an expert in all of these languages.

#### .NET and Java SDKs
* Requires a scheme in the endpoint configuration.
* Requires the scheme be `http` or `https`.
* Uses scheme to infer secure vs. insecure channel.
* Does not have an `insecure` configuration option.
* References: [.NET](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpTraceExporter.cs#L80) and [Java](https://github.com/open-telemetry/opentelemetry-java/blob/43f1ecf7ede49104c2aac1c9e5448b7cf7859ce7/exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporterBuilder.java#L88)

#### Go SDK and Collector
* A scheme is not permitted in the endpoint configuration. An error is thrown to the effect of `address https://some-host:4317/: too many colons in address`
  * Go has implemented handling of a scheme specifically for configuration from the `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable. See https://github.com/open-telemetry/opentelemetry-go/issues/1885.
* Has the `insecure` configuration option no longer present in the specification.

#### JS/Node
* Allows for a scheme in the endpoint configuration.
* Requires the scheme, if present, be `http`, `https`, `grpc`, or `grpcs`. [Code here](https://github.com/open-telemetry/opentelemetry-js/blob/7fa4ff70d1aad5875cb14c6bff99c838228447bb/packages/opentelemetry-exporter-collector-grpc/src/util.ts#L117-L121)
* The scheme has no bearing on whether a secure or insecure channel will be used. Instead it appears that this is configured by instantiating and using a [`ChannelCredentials`](https://grpc.github.io/grpc/node/grpc.credentials.html). The `ChannelCredentials` are used [here](https://github.com/open-telemetry/opentelemetry-js/blob/7fa4ff70d1aad5875cb14c6bff99c838228447bb/packages/opentelemetry-exporter-collector-grpc/src/util.ts#L36-L37).

### Input from Collector SIG

Based on a conversation on this topic with folks (@tigrannajaryan @bogdandrutu @alolita) during the 5/26 collector SIG meeting, I am drafting this PR to reopen a discussion for how (and if) we wish to resolve these inconsistencies.

To summarize the discussion from 5/26 collector SIG meeting:

  > If the collector adopts support for a scheme being present in the endpoint configuration, then people would prefer that this be implemented in a  backwards compatible way that would maintain support for a scheme-less endpoint.

## Changes

In an attempt to address 1) the desire for backwards compatibility and 2) the confusion brought on from the inconsistencies in the various implementations, here's the idea I'm toying with:

1. Use different language for the OTLP/gRPC and OTLP/HTTP exporters for the endpoint configuration.
2. Re-introduce the `insecure` flag only for use by the OTLP/gRPC exporter and only when a scheme is not present in the endpoint configuration.


tigrannajaryan (approved):
Thanks @alanwest 

Please also add a changelog entry.

I believe this is a non-breaking change since it makes Endpoint processing more _permissive_. Every endpoint value that was acceptable previously is still acceptable. Some values which were previously not accepted and would result in an error will be now accepted. I think we can treat this as a non-breaking change.

specification/protocol/exporter.md
```diff
@@ -10,7 +10,9 @@ The following configuration options MUST be available to configure the OTLP expo
...
+| Endpoint (OTLP/gRPC) | Target to which the exporter is going to send spans or metrics. The endpoint MUST accept a URL with or without a scheme. If it includes a scheme (http or https), a scheme of https indicates a secure connection. | `https://localhost:4317` | `OTEL_EXPORTER_OTLP_ENDPOINT` `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` |
+| Insecure             | Whether to enable client transport security for the exporter's `grpc` connection. Only applies to OTLP/gRPC when a scheme is not included in the `endpoint`. OTLP/HTTP always uses the scheme provided for the endpoint. | `false` | `OTEL_EXPORTER_OTLP_INSECURE` `OTEL_EXPORTER_OTLP_SPAN_INSECURE` `OTEL_EXPORTER_OTLP_METRIC_INSECURE` |
```
anuraaga, line 15: Does this mean SDKs all have to support this variable? I'd prefer not to support it - we already have a method for picking secure / insecure in the endpoint scheme, having another way of doing the same thing is extra documentation / support overhead (not to mention bringing back this problematic word `insecure` which security scanners don't like much). If the idea is to allow either (no-scheme url + insecure flag) or (scheme url) than that's ok but don't see that mentioned in this wording.

anuraaga: Another point is that it's always been quite mysterious when `host:port` defaults to TLS - I don't think that matches most users' expectations, it's just a security push by us I think. The schemed URLs solved this problem nicely, if allowing non-schemed URLs I really recommend the default being insecure, I think it matches most users' expectations (even `curl google.com` will hit HTTP, not HTTPS, this is what people think IMO).

iNikem: > like to maintain the ability to configure without a scheme. Go SDK also still supports the `insecure` config setting

Do they want to configure without scheme but still use TLS?

alanwest: > Do they want to configure without scheme but still use TLS?

Correct, TLS is the default in Go.

Go looks something like this
```go
        driver := otlpgrpc.NewDriver(
		otlpgrpc.WithInsecure(), // This option is required if you do not want TLS otherwise TLS is the default
		otlpgrpc.WithEndpoint("localhost:4317"), // To my knowledge, including a scheme here does not work
	)
	exp, err := otlp.NewExporter(ctx, driver)
```

On the other hand, Go does support a scheme when configuration is done via the `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable. The scheme `http` implies the `WithInsecure` option.

alanwest: > SDKs will always have extra knobs compared to what's in the spec. Usually they're language specific but it doesn't seem terrible if one SDK supports insecure flag if they want to.

This stance seems reasonable.

Perhaps at a minimum do you think it makes sense to say for OTLP/gRPC:

1) SDKs MUST support endpoint with configured with a scheme and the scheme `https` indicates use TLS.
2) SDKs MAY support an endpoint without a scheme but if they do then an additional `insecure` configuration is required.

First point effectively maintains status quo. Second point enables the flexibility for SDKs and requires a level of consistency when adopting this.

anuraaga: Yup I have no problem with text related to an `insecure` flag if it's optional.

alanwest: I've clarified things so that implementing the `insecure` flag is optional if the SDK does not allow for an endpoint to be configured without a scheme.

anuraaga (approved):

carlosalberto:
I feel like this needs more eyes. @open-telemetry/specs-approvers please review.

carlosalberto:
@open-telemetry/specs-approvers Please do review/approve ;)

jmacd (approved):

specification/protocol/exporter.md
```diff
@@ -10,7 +10,9 @@ The following configuration options MUST be available to configure the OTLP expo
...
+| Endpoint (OTLP/gRPC) | Target to which the exporter is going to send spans or metrics. The endpoint MUST accept a URL with a scheme and MAY accept a URL without a scheme. If it includes a scheme (http or https), a scheme of https indicates a secure connection. If it does not include a scheme then the `insecure` configuration option is used to determine whether a secure connection should be used. | `https://localhost:4317` | `OTEL_EXPORTER_OTLP_ENDPOINT` `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` |
```
Aneurysm9, line 14: For gRPC it is my understanding that the scheme is used to determine the address resolution method and not the transport protocol since the transport protocol is always HTTP/2, which requires TLS unless explicitly configured otherwise with and indication that insecure connections are acceptable.

[Valid scheme values for gRPC](https://github.com/grpc/grpc/blob/master/doc/naming.md) are `dns`, `unix`, `unix-abstract`, `ipv4`, and `ipv6`.  `dns` is the default value if no scheme is specified.

Using `http` and `https` as a shorthand for `dns-with-insecure` and `dns` may be acceptable, but we should be very explicit about what we're doing and how client implementations should treat otherwise invalid gRPC endpoint names.  We should also make clear that the other gRPC-supported schemes are valid.

specification/protocol/exporter.md
```diff
@@ -10,7 +10,9 @@ The following configuration options MUST be available to configure the OTLP expo
...
+| Endpoint (OTLP/HTTP) | Target to which the exporter is going to send spans or metrics. The endpoint MUST be a valid URL with scheme (http or https) and host, and MAY contain a port and path. A scheme of https indicates a secure connection. When using `OTEL_EXPORTER_OTLP_ENDPOINT`, exporters SHOULD follow the collector convention of appending the version and signal to the path (e.g. `v1/traces` or `v1/metrics`). The per-signal endpoint configuration options take precedence and can be used to override this behavior. See the [OTLP Specification][otlphttp-req] for more details. | `https://localhost:4317` | `OTEL_EXPORTER_OTLP_ENDPOINT` `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` |
```
iNikem, line 13:
```suggestion
| Endpoint (OTLP/HTTP) | Target to which the exporter is going to send spans or metrics. The endpoint MUST be a valid URL with scheme (http or https) and host, and MAY contain a port and path. A scheme of https indicates a secure connection. When using `OTEL_EXPORTER_OTLP_ENDPOINT`, exporters SHOULD follow the collector convention of appending the version and signal to the path (e.g. `v1/traces` or `v1/metrics`), if not present already. The per-signal endpoint configuration options take precedence and can be used to override this behavior. See the [OTLP Specification][otlphttp-req] for more details. | `https://localhost:4317` | `OTEL_EXPORTER_OTLP_ENDPOINT` `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` |
```

specification/protocol/exporter.md
iNikem: This is very important point IMO. If gRPC already assigns some meaning to schema part of their URLs then won't it be confusing if we offer completely different schemas with completely different meaning for "gRPC URLs"?

iNikem: @alanwest You use the same environment variables for both /HTTP and /gRPC options. How should SDK  understand which protocol to use?

alanwest: > You use the same environment variables for both /HTTP and /gRPC options. How should SDK understand which protocol to use?

I left them the same since it has always been the case that the same environment variables are to be used for both HTTP and gRPC exporters.

To my knowledge, some languages only support gRPC at the moment. For languages that support exporting via both gRPC and HTTP it seems common that these are shipped as two distinct components.

I do not know the history here, but the specification does mention a [`OTEL_EXPORTER_OTLP_PROTOCOL` environment variable](https://github.com/open-telemetry/opentelemetry-specification/blob/2a712f7626cb62b21ae210ef3175612e7744f285/specification/protocol/exporter.md#specify-protocol) with the note "We reserve the following environment variables for configuration of protocols in the future".

I imagine that the `OTEL_EXPORTER_OTLP_PROTOCOL` variable would be necessary if a language SDK decided to develop a single component that could be configured to export either by HTTP or gRPC.

alanwest: > Using `http` and `https` as a shorthand for `dns-with-insecure` and `dns` may be acceptable, but we should be very explicit about what we're doing and how client implementations should treat otherwise invalid gRPC endpoint names. We should also make clear that the other gRPC-supported schemes are valid.

Agreed there's some room to make this more clear - I took a stab at new wording, but I'm open to any suggestions for making this any more clear. See [this commit](https://github.com/open-telemetry/opentelemetry-specification/pull/1729/commits/ecabd5580f62f315dca3c24da0bbccedfb6b8d58).

I've attempted to keep things as client implementation agnostic as possible yet striving to be very clear. Client implementations differ in the schemes they support. As a point of contrast to the [valid scheme values for gRPC](https://github.com/grpc/grpc/blob/master/doc/naming.md) you pointed out, the .NET gRPC library, for example, requires including a scheme of `http` or `https` - that is, it doesn't really have a notion of defaulting to the `dns` scheme and in fact if you explicitly add `dns://` it doesn't work.

That said, I think it's safe to require OTel SDKs to require the ability to declare `http(s)` - I think the risk is low that we would stomp on the semantics of any specific client implementation.

Aneurysm9: This is much better, I think.  That said, the handling of `insecure` still feels weird.  None of the gRPC standard schemes convey any information about whether TLS enforcement is necessary.  It seems wrong to say that `insecure` should only be honored when no scheme is present.  The address resolution scheme and TLS enforcement are distinct concerns for gRPC (and even for HTTP where a service may initiate TLS with an untrusted certificate and the user wants to ignore that).

alanwest: > That said, the handling of insecure still feels weird.

Yep agreed, excellent callout. I've cleaned up the wording for the `insecure` option [here](https://github.com/open-telemetry/opentelemetry-specification/pull/1729/commits/3ca03ae902fc1ba864a883225b39ccadb8ab015e).

I've attempted again to remain as flexible and non-prescriptive as possible given the variety of client implementations some of which do not have a need for a separate insecure setting.

tigrannajaryan:
@alanwest please resolve the conflicts.

carlosalberto:
This has been around for some time now, and no more reviews/issues were reporting, hence merging.

PR #1733, opened 2021-05-28
Add clarification about async vs. sync instruments

reyang:
## Changes

* Editorial change - added clarification and examples about sync vs. async instruments.

### DIFF
--- a/specification/metrics/api.md
+++ b/specification/metrics/api.md
@@ -242,12 +242,17 @@ instrument. It MUST be treated as an opaque string from the API and SDK.
 Instruments can be categorized based on whether they are synchronous or
 asynchronous:
 
+<a name="synchronous-instrument"></a>
+
 * Synchronous instruments (e.g. [Counter](#counter)) are meant to be invoked by
   the user (e.g. a library developer or an application developer). For example,
   an HTTP client could use a Counter to record the number of bytes it has
   received. [Measurements](#measurement) recorded by synchronous instruments can
   be associated with [Baggage](../baggage/api.md) and
   [Context](../context/context.md).
+
+<a name="asynchronous-instrument"></a>
+
 * Asynchronous instruments (e.g. [Asynchronous Gauge](#asynchronous-Gauge)) give
   the user a way to register callback function, and the callback function will
   only be invoked upon collection. For example, an embedded software could use
@@ -263,7 +268,8 @@ pattern](https://en.wikipedia.org/wiki/Asynchronous_method_invocation).
 
 ### Counter
 
-`Counter` is a synchronous Instrument which supports non-negative increments.
+`Counter` is a [synchronous Instrument](#synchronous-instrument) which supports
+non-negative increments.
 
 Example uses for `Counter`:
 
@@ -354,9 +360,9 @@ counterPowerUsed.Add(200, new PowerConsumption { customer = "Jerry" }, ("is_gree
 
 ### Asynchronous Counter
 
-Asynchronous Counter is an asynchronous Instrument which reports
-[monotonically](https://wikipedia.org/wiki/Monotonic_function) increasing
-value(s) when the instrument is being observed.
+Asynchronous Counter is an [asynchronous Instrument](#asynchronous-instrument)
+which reports [monotonically](https://wikipedia.org/wiki/Monotonic_function)
+increasing value(s) when the instrument is being observed.
 
 Example uses for Asynchronous Counter:
 
@@ -478,9 +484,9 @@ operation is provided by the `callback`, which is registered during the
 
 ### Histogram
 
-`Histogram` is a synchronous Instrument which can be used to report arbitrary
-values that are likely to be statistically meaningful. It is intended for
-statistics such as histograms, summaries, and percentile.
+`Histogram` is a [synchronous Instrument](#synchronous-instrument) which can be
+used to report arbitrary values that are likely to be statistically meaningful.
+It is intended for statistics such as histograms, summaries, and percentile.
 
 Example uses for `Histogram`:
 
@@ -562,10 +568,10 @@ httpServerDuration.Record(100, new HttpRequestAttributes { method = "GET", schem
 
 ### Asynchronous Gauge
 
-Asynchronous Gauge is an asynchronous Instrument which reports non-additive
-value(s) (_e.g. the room temperature - it makes no sense to report the
-temperature value from multiple rooms and sum them up_) when the instrument is
-being observed.
+Asynchronous Gauge is an [asynchronous Instrument](#asynchronous-instrument)
+which reports non-additive value(s) (_e.g. the room temperature - it makes no
+sense to report the temperature value from multiple rooms and sum them up_) when
+the instrument is being observed.
 
 Note: if the values are additive (_e.g. the process heap size - it makes sense
 to report the heap size from multiple processes and sum them up, so we get the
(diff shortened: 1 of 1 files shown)


specification/metrics/api.md
```diff
@@ -239,6 +239,28 @@ instrument. It MUST be treated as an opaque string from the API and SDK.
 * It MUST support at least 1023 characters. Individual language clients can
   decide if they want to support more.
 
+Instruments can be categorized based on whether they are synchronous or
+asynchronous:
+
+* Synchronous instruments (e.g. [Counter](#counter)) are meant to be invoked by
```
jkwatson, line 245:
```suggestion
* Synchronous instruments (e.g. [Counter](#counter)) are meant to be invoked directly by
```

specification/metrics/api.md
```diff
@@ -239,6 +239,28 @@ instrument. It MUST be treated as an opaque string from the API and SDK.
...
+  the user (e.g. a library developer or an application developer). For example,
+  an HTTP client could use a Counter to record the number of bytes it has
+  received. [Measurements](#measurement) recorded by synchronous instruments can
+  be associated with [Baggage](../baggage/api.md) and
+  [Context](../context/context.md).
+* Asynchronous instruments (e.g. [Asynchronous Gauge](#asynchronous-Gauge)) give
+  the user a way to register callback function, and the callback function will
+  only be invoked upon collection. For example, an embedded software could use
```
jkwatson, line 258:
```suggestion
  only be invoked upon collection. For example, a piece of embedded software could use
```

specification/metrics/api.md
```diff
@@ -239,6 +239,28 @@ instrument. It MUST be treated as an opaque string from the API and SDK.
...
+  an asynhronous gauge to collect the temperature from a sensor every 15
+  seconds, which means the callback function will only be invoked every 15
+  seconds. [Measurements](#measurement) recorded by asynchronous instruments
+  cannot be associated with [Baggage](../baggage/api.md) and
+  [Context](../context/context.md).
+
+Please note that the term _synchronous_ and _asynchronous_ have nothing to do
+with [asynchronous
```
jkwatson, line 266:
```suggestion
with the [asynchronous
```

specification/metrics/api.md
reyang: Taken. Thanks!

specification/metrics/api.md
reyang: Updated.

specification/metrics/api.md
reyang: Updated, thanks!

jsuereth (approved):

specification/metrics/api.md
```diff
@@ -239,9 +239,37 @@ instrument. It MUST be treated as an opaque string from the API and SDK.
 * It MUST support at least 1023 characters. Individual language clients can
   decide if they want to support more.
 
+Instruments can be categorized based on whether they are synchronous or
+asynchronous:
+
+<a name="synchronous-instrument"></a>
+
+* Synchronous instruments (e.g. [Counter](#counter)) are meant to be invoked
```
jsuereth, line 247: Nit:  `meant to be invoked inline with application/business processing logic, within the scope of [Context](../context/context.md).`

Baggage  (as specififed) is available via a context. I think you can call it out below as an additional thing, but the key here is;
- The logic is inline with the mainline processing of the application (not sheddable when under load)
- The logic can have automagical attachment of context if desired (and all the fun contexty things).

jmacd (approved):

specification/metrics/api.md
```diff
@@ -239,9 +239,37 @@ instrument. It MUST be treated as an opaque string from the API and SDK.
...
+  inline with application/business processing logic. For example, an HTTP client
+  could use a Counter to record the number of bytes it has received.
+  [Measurements](#measurement) recorded by synchronous instruments can be
+  associated with [Baggage](../baggage/api.md) and
+  [Context](../context/context.md).
```
jkwatson, line 252: this reads a little funny. Baggage is in the context, as is the current span. If you're going to call out Baggage, I would also call out the Span.

specification/metrics/api.md
```diff
@@ -239,9 +239,37 @@ instrument. It MUST be treated as an opaque string from the API and SDK.
...
+
+<a name="asynchronous-instrument"></a>
+
+* Asynchronous instruments (e.g. [Asynchronous Gauge](#asynchronous-gauge)) give
+  the user a way to register callback function, and the callback function will
+  only be invoked upon collection. For example, a piece of embedded software
+  could use an asynhronous gauge to collect the temperature from a sensor every
```
jkwatson, line 259:
```suggestion
  could use an asynchronous gauge to collect the temperature from a sensor every
```

specification/metrics/api.md
reyang: This is a good callout. I agree and I've removed Baggage here (the API spec).

In the SDK spec we will need to call out the difference between Baggage and Context (e.g. the SDK user might specify that he/she wants dimension "a" to come from the Baggage, and dimension "b" to come from the Context).

specification/metrics/api.md
reyang: Please take another look, I believe this has been addressed.

specification/metrics/api.md
jkwatson: I don't think we'd want anything to come directly from the Context via the SDK. Unless it's a span attribute, of course. The keys to objects in the Context should be private, so the Context couldn't access them.

reyang: > The keys to objects in the Context should be private, so the SDK couldn't access them.

+1

The longer version would be: sync instruments can be associated with `Context`, which means the user could configure additional dimensions that come from the Baggage, Span (including span attributes, name/kind/etc.), and anything that we added later to the Context that can be publicly accessed.

I think the key here is to clarify that async instruments are "contextless", so I picked the word "can be associated" instead "have access to". Sounds reasonable?

jkwatson: Perfect, thanks!

jkwatson (approved):
👍🏼

carlosalberto (approved):

carlosalberto:
Let's merge this on Monday prior to the 1.4.0 release.

specification/metrics/api.md
```diff
@@ -224,9 +224,35 @@ instrument. It MUST be treated as an opaque string from the API and SDK.
 * It MUST support at least 1023 characters. Individual language clients can
   decide if they want to support more.
 
+Instruments can be categorized based on whether they are synchronous or
+asynchronous:
+
+<a name="synchronous-instrument"></a>
```
bogdandrutu, line 230: Why html tags like this? Should use markdown headers if needed.

bogdandrutu (approved):

specification/metrics/api.md
carlosalberto: Ping @reyang

reyang: It's the same as other tags in this doc (and other spec doc such as the tracing one):
1. we don't want them to be in the top level ToC (since they interrupt the ToC flow)
2. we want to have an anchor

IIRC @bogdandrutu you've asked the same question before?
e.g. https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-naming-rule

https://github.com/open-telemetry/opentelemetry-specification/pull/1578#discussion_r602575017

carlosalberto: @bogdandrutu Please fill an issue to follow up this if this is still a problem, etc. Thanks!

PR #1748, opened 2021-06-09
Remove rpc.jsonrpc.method, clarify rpc.method instead.

Oberon00:
(no corresponding issue)

## Changes

Remove `rpc.jsonrpc.method`, clarify that `rpc.method` should be used instead.

## Related

* PR that introduced JSON RPC conventions: #1643 (@Rast1234)

This change was just released in 1.4, but semantic conventions are still experimental thus a breaking change like this may be acceptable.

### DIFF
--- a/semantic_conventions/trace/rpc.yaml
+++ b/semantic_conventions/trace/rpc.yaml
@@ -116,5 +116,8 @@ groups:
           type: string
           brief: "`error.message` property of response if it is an error response."
           examples: ['Parse error', 'User already exists']
-      constraints:
-        - any_of: ['rpc.method']
+        - ref: rpc.method
+          required: always
+          note: >
+           This is always required for jsonrpc. See the note in the general
+           RPC conventions for more information.
--- a/semantic_conventions/trace/rpc.yaml
+++ b/semantic_conventions/trace/rpc.yaml
@@ -17,7 +17,9 @@ groups:
             This is the logical name of the service from the RPC interface perspective,
             which can be different from the name of any implementing class.
             The `code.namespace` attribute may be used to store the latter
-            (despite the attribute name, it may include a class name).
+            (despite the attribute name, it may include a class name;
+            e.g., class with method actually executing the call on the server side,
+            RPC client stub class on the client side).
           examples: "myservice.EchoService"
         - id: method
           type: string
@@ -27,7 +29,9 @@ groups:
           note: >
             This is the logical name of the method from the RPC interface perspective,
             which can be different from the name of any implementing method/function.
-            The `code.function` attribute may be used to store the latter.
+            The `code.function` attribute may be used to store the latter
+            (e.g., method actually executing the call on the server side,
+            RPC client stub method on the client side).
           examples: "exampleMethod"
         - ref: net.peer.ip
         - ref: net.peer.name
--- a/semantic_conventions/trace/rpc.yaml
+++ b/semantic_conventions/trace/rpc.yaml
@@ -12,13 +12,22 @@ groups:
           type: string
           required:
             conditional: "No, but recommended"
-          brief: 'The full name of the service being called, including its package name, if applicable.'
+          brief: 'The full (logical) name of the service being called, including its package name, if applicable.'
+          note: >
+            This is the logical name of the service from the RPC interface perspective,
+            which can be different from the name of any implementing class.
+            The `code.namespace` attribute may be used to store the latter
+            (despite the attribute name, it may include a class name).
           examples: "myservice.EchoService"
         - id: method
           type: string
           required:
             conditional: "No, but recommended"
-          brief: 'The name of the method being called, must be equal to the $method part in the span name.'
+          brief: 'The name of the (logical) method being called, must be equal to the $method part in the span name.'
+          note: >
+            This is the logical name of the method from the RPC interface perspective,
+            which can be different from the name of any implementing method/function.
+            The `code.function` attribute may be used to store the latter.
           examples: "exampleMethod"
         - ref: net.peer.ip
         - ref: net.peer.name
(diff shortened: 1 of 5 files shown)


Rast1234:
So you changed the purpose of `rpc.method` to be logical, not physical method/function name on the server? Great!

Oberon00:
> So you changed the purpose of rpc.method to be logical, not physical method/function name on the server

In my mind, this was already the case before, but the wording was not very clear.

carlosalberto:
@Rast1234 Just to double confirm: are you ok with this change?

Rast1234:
Yes!

semantic_conventions/trace/rpc.yaml
```diff
@@ -108,3 +112,5 @@ groups:
           type: string
           brief: "`error.message` property of response if it is an error response."
           examples: ['Parse error', 'User already exists']
+      constraints:
+        - any_of: ['rpc.method']
```
arminru, line 116: I think you could also include/reference it and make it required (unconditionally). The current markdown output of this still looks a bit odd.

semantic_conventions/trace/rpc.yaml
```diff
@@ -12,13 +12,22 @@ groups:
           type: string
           required:
             conditional: "No, but recommended"
-          brief: 'The full name of the service being called, including its package name, if applicable.'
+          brief: 'The full (logical) name of the service being called, including its package name, if applicable.'
+          note: >
+            This is the logical name of the service from the RPC interface perspective,
+            which can be different from the name of any implementing class.
+            The `code.namespace` attribute may be used to store the latter
```
arminru, line 19:
```suggestion
            The `code.namespace` attribute may be used on the server to store the latter
```

semantic_conventions/trace/rpc.yaml
```diff
@@ -12,13 +12,22 @@ groups:
...
         - id: method
           type: string
           required:
             conditional: "No, but recommended"
-          brief: 'The name of the method being called, must be equal to the $method part in the span name.'
+          brief: 'The name of the (logical) method being called, must be equal to the $method part in the span name.'
+          note: >
+            This is the logical name of the method from the RPC interface perspective,
+            which can be different from the name of any implementing method/function.
+            The `code.function` attribute may be used to store the latter.
```
arminru, line 30:
```suggestion
            The `code.function` attribute may be used on the server to store the latter.
```

arminru (approved):
Thanks! I think explicitly declaring this as the logical names and referencing the `code.*` attributes makes sense in any case, not only for JSON-RPC.

carlosalberto (approved):

carlosalberto:
LGTM once the feedback has been addressed.

SergeyKanzhelev (approved):

semantic_conventions/trace/rpc.yaml
Oberon00: Hmm, the implementing method on the client *can* make sense, as many RPC frameworks generate client stub classes for you. Then you would store the name of the client stub class at the client side. Could use extra wording though. WDYT?

semantic_conventions/trace/rpc.yaml
Oberon00: Yeah, there is https://github.com/open-telemetry/build-tools/issues/33 for a related case.

Oberon00:
@arminru
> makes sense in any case, not only for JSON-RPC.
Not sure if this was meant as a suggestion or statement, but that's what this PR does already?

semantic_conventions/trace/rpc.yaml
arminru: Sounds reasonable, let's go with suggesting both then.

semantic_conventions/trace/rpc.yaml
Oberon00: b6fd5ce

semantic_conventions/trace/rpc.yaml
Oberon00: ce3512b

semantic_conventions/trace/rpc.yaml
Oberon00: ce3512b

arminru:
> > makes sense in any case, not only for JSON-RPC.
> 
> Not sure if this was meant as a suggestion or statement, but that's what this PR does already?

@Oberon00 Oh I was just saying that I like the change and that I agree refining the semantics beyond just solving it for JSON-RPC (as done in the PR) makes sense.

arminru (approved):

carlosalberto (approved):

carlosalberto:
@Oberon00 Please update your PR so we can merge it.

jkwatson:
This should include an update to the schema, correct? Our schema does support re-writing attributes from old to new, I believe.

Oberon00:
@jkwatson 

> This should include an update to the schema, correct? Our schema does support re-writing attributes from old to new, I believe.

Hmm, probably, unless we consider schema updates part of the work of cutting a release and do it all at once at the end.
In that case we rename an attribute to an already existing attribute. I'm not sure what should happen if the existing rpc.method attribute is already filled out. Based on what I think was the misunderstanding about rpc.method that caused rpc.jsonrpc.method, I think that rpc.method should be overwritten with rpc.jsonrpc.method.

That being said, I would appreciate help with writing the schema, as there are no examples to draw from yet? @tigrannajaryan, can you assist?

carlosalberto:
Ping @tigrannajaryan

tigrannajaryan:
Assuming this is really semantically an attribute rename this schema file should do it:
```yaml
file_format: 1.0.0
schema_url: https://opentelemetry.io/schemas/1.5.0
versions:
  1.5.0:
    spans:
      changes:
        # Sequence of translations to apply to convert the schema from a prior version
        # to this version. The order in this sequence is important. Translations are
        # applied from top to bottom in the listed order.
        - rename_attributes:
            # Rename attributes of all spans, regardless of span name.
            # The keys are the old attribute name used prior to this version, the values are
            # the new attribute name starting from this version.
            attribute_map:
              rpc.jsonrpc.method: rpc.method
```
The file should be named `schemas/1.5.0` (no extension in the file name).

Note, I did not review the PR thoroughly, so I am not completely sure it is actually a renaming and not a deletion of an old attribute and introduction of a new unrelated attribute.

P.S. I have put together a small tool to do some sanity checks of schema files: https://github.com/tigrannajaryan/telemetry-schema/blob/main/cmd/scheck/main.go

Oberon00:
> Note, I did not review the PR thoroughly, so I am not completely sure it is actually a renaming and not a deletion of an old attribute and introduction of a new unrelated attribute.

This PR is a deletion of a new attribute because it can already by represented by an existing attribute (new -> existing). So it's not exactly a renaming (new1 -> new2)

tigrannajaryan:
> What would `rename_attributes` do if both the new and old attribute are set?

The behavior is undefined in the OTEP right now. We can define it but I would say if there is an expectation that both may be set then it does not sound like renaming and seems to be something more complicated. I don't think `rename_attributes` can handle that.

tigrannajaryan:
> So then is it fine to merge this PR without a schema entry?

Yes, I think so. Since the semconv is not yet stable it is fine. In the future if something similar to this happens we will need to figure out how we can describe the change in the schema file.

PR #1130, opened 2020-10-22
Add an option to limit length of values of attributes and metric values

jtmalinowski:
Docs are based on existing Java implementation https://github.com/open-telemetry/opentelemetry-java/pull/1484

Fixes #1133 

Messaging about truncation was extracted to #1787 

Metrics support extracted to #1830

PR #1751, opened 2021-06-09
Adding environment variables for event and link attribute limits

### LINKED ISSUE #1750: Span link and event count limits should have corresponding environment variables

Environment variables are defined for setting span attribute limits, but not span link and event attribute limits. These should exist to be consistent.

codeboten:
Fixes #1750 

## Changes

Adding environment variable definitions for span event and link attributes.

### DIFF
--- a/spec-compliance-matrix.md
+++ b/spec-compliance-matrix.md
@@ -143,8 +143,8 @@ Note: Support for environment variables is optional.
 |OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT               | - | +  |   | +    | +  | -    |   | +  | - |    |     |
 |OTEL_SPAN_EVENT_COUNT_LIMIT                   | - | +  |   | +    | +  | -    |   | +  | - |    |     |
 |OTEL_SPAN_LINK_COUNT_LIMIT                    | - | +  |   | +    | +  | -    |   | +  | - |    |     |
-|OTEL_SPAN_ATTRIBUTE_PER_EVENT_COUNT_LIMIT     |   |    |   |      |    |      |   |    |   |    |     |
-|OTEL_SPAN_ATTRIBUTE_PER_LINK_COUNT_LIMIT      |   |    |   |      |    |      |   |    |   |    |     |
+|OTEL_EVENT_ATTRIBUTE_COUNT_LIMIT              |   |    |   |      |    |      |   |    |   |    |     |
+|OTEL_LINK_ATTRIBUTE_COUNT_LIMIT               |   |    |   |      |    |      |   |    |   |    |     |
 |OTEL_TRACES_SAMPLER                           | - | +  |   | +    | +  | +    |   | -  | - |    |     |
 |OTEL_TRACES_SAMPLER_ARG                       | - | +  |   | +    | +  | +    |   | -  | - |    |     |
 
--- a/spec-compliance-matrix.md
+++ b/spec-compliance-matrix.md
@@ -143,6 +143,8 @@ Note: Support for environment variables is optional.
 |OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT               | - | +  |   | +    | +  | -    |   | +  | - |    |     |
 |OTEL_SPAN_EVENT_COUNT_LIMIT                   | - | +  |   | +    | +  | -    |   | +  | - |    |     |
 |OTEL_SPAN_LINK_COUNT_LIMIT                    | - | +  |   | +    | +  | -    |   | +  | - |    |     |
+|OTEL_SPAN_ATTRIBUTE_PER_EVENT_COUNT_LIMIT     |   |    |   |      |    |      |   |    |   |    |     |
+|OTEL_SPAN_ATTRIBUTE_PER_LINK_COUNT_LIMIT      |   |    |   |      |    |      |   |    |   |    |     |
 |OTEL_TRACES_SAMPLER                           | - | +  |   | +    | +  | +    |   | -  | - |    |     |
 |OTEL_TRACES_SAMPLER_ARG                       | - | +  |   | +    | +  | +    |   | -  | - |    |     |
 
--- a/specification/sdk-environment-variables.md
+++ b/specification/sdk-environment-variables.md
@@ -78,11 +78,13 @@ Depending on the value of `OTEL_TRACES_SAMPLER`, `OTEL_TRACES_SAMPLER_ARG` may b
 
 See the SDK [Span Limits](trace/sdk.md#span-limits) section for the definition of the limits.
 
-| Name                            | Description                          | Default | Notes |
-| ------------------------------- | ------------------------------------ | ------- | ----- |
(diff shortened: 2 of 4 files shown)


spec-compliance-matrix.md
```diff
@@ -143,6 +143,8 @@ Note: Support for environment variables is optional.
 |OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT               | - | +  |   | +    | +  | -    |   | +  | - |    |     |
 |OTEL_SPAN_EVENT_COUNT_LIMIT                   | - | +  |   | +    | +  | -    |   | +  | - |    |     |
 |OTEL_SPAN_LINK_COUNT_LIMIT                    | - | +  |   | +    | +  | -    |   | +  | - |    |     |
+|OTEL_SPAN_ATTRIBUTE_PER_EVENT_COUNT_LIMIT     |   |    |   |      |    |      |   |    |   |    |     |
```
iNikem, line 146: Are these "span attributes" or "event/link attributes"?

specification/trace/sdk.md
```diff
@@ -342,6 +342,10 @@ public final class SpanLimits {
 
   public int getAttributeCountLimit();
 
+  public int getAttributePerEventCountLimit();
+
+  public int getAttributePerLinkCountLimit();
```
iNikem, line 347:
```suggestion
  public int getAttributeCountPerEventLimit();

  public int getAttributeCountPerLinkLimit();
```

spec-compliance-matrix.md
codeboten: Would you rename it `OTEL_SPAN_EVENT_ATTRIBUTE_COUNT_LIMIT`?

codeboten: Or just `OTEL_EVENT_ATTRIBUTE_COUNT_LIMIT`?

iNikem: > Or just `OTEL_EVENT_ATTRIBUTE_COUNT_LIMIT`?

I like this one

codeboten: Updated, PTAL

iNikem (approved):

carlosalberto (approved):

carlosalberto:
@open-telemetry/specs-approvers please review/approve ;)

tigrannajaryan:
Somewhat meta, but I am a bit worried about the proliferation of the env variables. I wonder if this is desirable or we should only have basic capabilities controlled by env variables and the rest requires writing actual code? Taken to the extreme we can start coding the entire SDK behavior in the env variables. Is there a good way to think about what should be an env variable and what shouldn't?

codeboten:
> Is there a good way to think about what should be an env variable and what shouldn't?

I agree, it would be great to have guidelines on this. Specifically around this PR, I think there should be equivalent variables for all available limits configuration, otherwise i could imagine this being confusing as a user. Why would some limits be configurable via env variables vs others not.

I suspect these will become less important once there's a configuration format available for otel components.

iNikem:
We cannot forbid language SIGs to provide configuration via environment variables. And as soon as one SDK has them, it is a good idea to standardise them across languages. But then, as @codeboten mentioned, it will be confusing to do that only for a subset of configuration options.

carlosalberto:
I agree with @tigrannajaryan's point (hence my initial idea of having a single env var to handle *all* the attribute limits), but for this case I think it does make sense to add them right now.

Hopefully the new Instrumentation group will be able to work out further details like this (as they are part of the configuration area). cc @tedsuo

carlosalberto:
Also, unless we have some opposition, let's merge this by early Monday, so allow more time for people to review/approve this.

specification/sdk-environment-variables.md
```diff
@@ -78,11 +78,13 @@ Depending on the value of `OTEL_TRACES_SAMPLER`, `OTEL_TRACES_SAMPLER_ARG` may b
...
-| ------------------------------- | ------------------------------------ | ------- | ----- |
-| OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT | Maximum allowed span attribute count | 128     |       |
-| OTEL_SPAN_EVENT_COUNT_LIMIT     | Maximum allowed span event count     | 128     |       |
-| OTEL_SPAN_LINK_COUNT_LIMIT      | Maximum allowed span link count      | 128     |       |
+| Name                             | Description                                    | Default | Notes |
+| -------------------------------- | ---------------------------------------------- | ------- | ----- |
+| OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT  | Maximum allowed span attribute count           | 128     |       |
+| OTEL_SPAN_EVENT_COUNT_LIMIT      | Maximum allowed span event count               | 128     |       |
+| OTEL_SPAN_LINK_COUNT_LIMIT       | Maximum allowed span link count                | 128     |       |
+| OTEL_EVENT_ATTRIBUTE_COUNT_LIMIT | Maximum allowed attribute per span event count | 128     |       |
```
jtmalinowski, line 86: Prefix with `OTEL_SPAN_` for consistency? Otherwise the name makes it seem like it applies to all events.

carlosalberto: https://github.com/open-telemetry/opentelemetry-specification/pull/1751#discussion_r651358547

jkwatson (approved):

tigrannajaryan:
https://github.com/open-telemetry/opentelemetry-specification/pull/1130 seems to be a very similar PR for different things to limit.

My initial worry stands: there seems to be a proliferation of env variables. Is there a point where this becomes unmanageable? Are we aiming at having full behavior configuration via env variables? Taken to the extreme we need a Turing-complete language and supply programs in env variables that modify the default behavior of SDK. I am not convinced this is the right direction.

Perhaps we need to support a config file and have one env variable that points to the config file to use?

tigrannajaryan:
> #1130 seems to be a very similar PR for different things to limit.
> 
> My initial worry stands: there seems to be a proliferation of env variables. Is there a point where this becomes unmanageable? Are we aiming at having full behavior configuration via env variables? Taken to the extreme we need a Turing-complete language and supply programs in env variables that modify the default behavior of SDK. I am not convinced this is the right direction.
> 
> Perhaps we need to support a config file and have one env variable that points to the config file to use?

@open-telemetry/technical-committee and @open-telemetry/specs-approvers: I would like your opinion before this PR gets merged.

I am not against this particular PR, but would like to understand if we need some best practices around no-code configuration and whether what we do with env variables is acceptable long term.

jmacd (approved):

specification/sdk-environment-variables.md
jtmalinowski: Surprising, but okay if it's decided already.

jtmalinowski (approved):

carlosalberto:
@tigrannajaryan we decided during the last SIG call to go ahead with these two aforementioned PRs, while creating #1773  to track the configuration dilemma. Let me know if that works for you ;)

tigrannajaryan:
> @tigrannajaryan we decided during the last SIG call to go ahead with these two aforementioned PRs, while creating #1773 to track the configuration dilemma. Let me know if that works for you ;)

SGTM.

PR #1754, opened 2021-06-10
Editorial change regarding metrics unit

reyang:
## Changes

Duration Python prototype I noticed that folks were struggling if the parameter name should be called `unit` or `unit_of_measure`. Trying to make this editorial change so it helps to encourage `unit`.


yurishkuro (approved):

jkwatson (approved):

arminru (approved):

arminru:
Thanks! With this change it is now consistent with the existing usage of `unit` in the description and examples.

PR #1759, opened 2021-06-14
Mobile spec additions

bryce-b:
Follow up of https://github.com/open-telemetry/opentelemetry-specification/pull/1669
I updated the yaml files & verified the various make checks pass. 
## Changes

- adds net.host.carrier.*
- adds net.host.connection_type

Related issues #

#1647 Add semantic convention for describing network connectivity conditions


bryce-b:
how can I run `make table-check` locally? I'm not getting a clear line number of which table is causing problems with the build error.

bryce-b:
hm, it seems that `make markdownlint` and `make table-check` are at odds.

bryce-b:
I need some assistance resolving this error in the semantic-conventions check. It oddly seems to resolve if a space is arbitrarily added in 7e7a188, but that breaks markdownlint. Is this actually an issue between the yaml spec & the markdown file? How can I view what the tool is generating from the yaml?

semantic_conventions/trace/general.yaml
```diff
@@ -60,6 +60,44 @@ groups:
         type: string
         brief: 'Local hostname or similar, see note below.'
         examples: 'localhost'
+      - id: host.connection_type
+        type:
+          allow_custom_values: false
```
jkwatson, line 65: I think we should allow custom values for this, for future extensibility.

bryce-b: 👍

arminru:
Hi @bryce-b!

> I need some assistance resolving this error in the semantic-conventions check. It oddly seems to resolve if a space is arbitrarily added in [7e7a188](https://github.com/open-telemetry/opentelemetry-specification/commit/7e7a1884dad1673a0e5a9a1a7f45a429e988b28a), but that breaks markdownlint. Is this actually an issue between the yaml spec & the markdown file? How can I view what the tool is generating from the yaml?

The tools' output is done right in-place if you run it as advised in the README, so you should see it right away and commit it together with the YAML change. You'll also have to re-run the tool now anyway since ba64493 introduced wording that is actually a bit different from what the Markdown generator would add, then it should pass again.

specification/trace/semantic_conventions/span-general.md
```diff
@@ -56,6 +61,18 @@ the `net.peer.*` properties of a client are equal to the `net.host.*` properties
...
+| `wifi` | wifi |
+| `wired` | wired |
+| `1G` | 1G |
+| `2G` | 2G |
+| `3G` | 3G |
+| `4G` | 4G |
+| `5G` | 5G |
+| `unavailable` | unavailable |
 <!-- endsemconv -->
 For `Unix` and `pipe`, since the connection goes over the file system instead of being directly to a known peer, `net.peer.name` is the only attribute that usually makes sense (see description of `net.peer.name` below).
```
arminru, line 77:
```suggestion
<!-- endsemconv -->

For `Unix` and `pipe`, since the connection goes over the file system instead of being directly to a known peer, `net.peer.name` is the only attribute that usually makes sense (see description of `net.peer.name` below).
```
I think the markdown linter will also expect a blank line in between here.

specification/trace/semantic_conventions/span-general.md
```diff
@@ -56,6 +61,18 @@ the `net.peer.*` properties of a client are equal to the `net.host.*` properties
 
 **[1]:** Signals that there is only in-process communication not using a "real" network protocol in cases where network attributes would normally be expected. Usually all other network attributes can be left out in that case.
 
+`net.host.connection_type` MAY be one of the following:
```
arminru, line 64: It seems as if this wording was added manually. Please run the markdown generator as described here:
https://github.com/open-telemetry/opentelemetry-specification/tree/main/semantic_conventions#generation
It will take the updated YAML and update the MD file accordingly, so you can commit and push both. The CI is verifying that the MD file matches the expected output based on the YAML.

bryce-b: Ah, I didn't grok what this originally meant. I ran it with my manually added changes and it didn't complain. 😆 I came at this PR backwards: adding the markdown first, then the yaml description!

bryce-b:
I've already made this comment, but it's in an outdated review comment chain. 

I deleted all the table markdown I manually added, and let the `table-generation` add them. (7af45c0) It looks much better, but it seems like the table-generation is adding an additional space at line 64 that is making markdownlint unhappy.
./specification/trace/semantic_conventions/span-general.md:64 MD012/no-multiple-blanks Multiple consecutive blank lines [Expected: 1; Actual: 2]

Oberon00:
Unfortunately, this looks like a bug in the semantic convention generator, when more than one enum is added and there is a (foot)note in one. 😞

bryce-b:
@Oberon00 phew! I'm glad it wasn't me, haha

specification/trace/semantic_conventions/span-general.md
```diff
@@ -56,7 +61,21 @@ the `net.peer.*` properties of a client are equal to the `net.host.*` properties
 
 **[1]:** Signals that there is only in-process communication not using a "real" network protocol in cases where network attributes would normally be expected. Usually all other network attributes can be left out in that case.
 
+
+`net.host.connection_type` MUST be one of the following or, if none of the listed values apply, a custom value:
```
jkwatson, line 65: @bryce-b on Android, depending on the API level, and whether the app has access to the `TelephonyManager`, you might only be able to get that the network is CELLULAR vs. WIFI. We should have a value in these allowed types that denotes a generic cellular connection, without the sub-type. Alternatively, these conventions could have a coarse-grained "WIFI/CELLULAR/WIRED" attributes, and a more fine-grained attribute for the sub-type of connection.

bryce-b: @jkwatson do you think there should be a new attribute for the connect subtype in that case? e.g. `connection_type` can be `wifi`, `cell`, `wired`, `unavailable`, and add `connection_subtype` for `3G`, `4G`, etc, and allow for other subtypes that might describe details on the other connect types if warrented?

jkwatson: I think that might be a little better, yes, since (I think) we can always get the coarse-grained information, but maybe not the fine-grained. But, I'm just starting to learn about this stuff, so I'm happy to defer to folks who have more experience with mobile than I do.

bryce-b: I think it's a fine idea. Should the attribute names be updated to not use an underscore? `_`. maybe something like `connection.type` & `connection.subtype`?

jkwatson: I like those. While we're on the topic, "host" feels weird to me for a mobile device. What would you think about using `net.device.connection.type` and `net.device.connection.subtype` ?

bryce-b: I would prefer to use `device`  vs `host` in a lot of areas... might be outside the scope of this PR though. Would it be trivial to update the spec to allow for device and host to be interchangeable, or would that require a whole host of changes?

jkwatson: 🔕 

I think if we also define lat/long location info for mobile, device is going to make a lot more sense than `host`. :)

specification/trace/semantic_conventions/span-general.md
```diff
@@ -56,7 +61,21 @@ the `net.peer.*` properties of a client are equal to the `net.host.*` properties
...
+
+| Value  | Description |
+|---|---|
+| `wifi` | wifi |
+| `wired` | wired |
+| `1G` | 1G |
```
maxgolov, line 71: What is the meaning of these values 1G..5G?

These bear very little semantics / practical sense, unless you also provide a function manually bins the most common "actual" network types into one of these proposed here. I don't think this list is adequate - since it may also not be necessarily representative of actual connection speed.

[List for Android as a reference](https://android.googlesource.com/platform/frameworks/base.git/+/master/telephony/java/android/telephony/TelephonyManager.java#2782) ...

How do you handle the case when user is connected to WiFi and LTE at the same time, what is going to be reported on Span? Why `wifi` is taking separate line (having at least 5 generations of different speeds). But for mobile connection - we are trying to add 5 separate (fairly artificial) distinct enumeration values?

maxgolov:
I think, as a minimum, the concepts that need to be gathered are:
- network cost
- device id
- manufacturer: make, model
- OS arch type
- possibly a vector of device features: hasKeyboard, hasMouse, hasLTE or WiFi only, hasPhone-alike characteristics, e.g. Dialer which may be a feature even on device that has WiFi only.
- many (most) devices in developing nations are dual-SIM ; you have to account for the fact that network is a vector, not a single value

specification/trace/semantic_conventions/span-general.md
bryce-b: [Mobile radio technologies](https://en.wikipedia.org/wiki/Mobile_radio_telephone) can be generally categorized into one of these values. Due to the esoteric nature of these [radio names](https://developer.apple.com/documentation/coretelephony/cttelephonynetworkinfo/radio_access_technology_constants) (i.e. not user friendly) I thought it would be better to provide naming that can be grokked at a glance. I'm not going to lie, this structure is very iOS centric, and will likely need to be extended when Otel has more of an Android presence. Due to that, the structure is based on what data is available on an iOS device, hence `wifi` is not broken down in to any further.  I think that allowing non-listed values can solve for this problem, as well as iterating on the spec.

I'm not too concerned about devices connected with both wifi and cell--the device will use one of them to to make network transactions, and the goal of adding this spec change is allow for a connection type to be attributed to networking spans as a way to provide additional understanding around response times on mobile devices: e.g.: response times broken-down by connection_type will help users discover when their slow rates may be due to devices connected via 1G or if there is something actionable in their stack.

specification/trace/semantic_conventions/span-general.md
bryce-b: I'll make a ticket to capture this proposition :)

specification/trace/semantic_conventions/span-general.md
jkwatson: I do agree with @maxgolov about these values. On Android, we get the values pointed to in the doc above (although, as mentioned, only if the application has permissions to use the TelephonyManager). We'd want a very clear mapping between these values, or just make this field free-form and let each OS provide the values it provides.

jkwatson:
> I think, as a minimum, the concepts that need to be gathered are:
> 
> * network cost
> * device id
> * manufacturer: make, model
> * OS arch type
> * possibly a vector of device features: hasKeyboard, hasMouse, hasLTE or WiFi only, hasPhone-alike characteristics, e.g. Dialer which may be a feature even on device that has WiFi only.
> * many (most) devices in developing nations are dual-SIM ; you have to account for the fact that network is a vector, not a single value
> * for any field added - there has to be a separate column, whether this field potentially needs a privacy review. For all of these privacy-sensitive fields - these should be definitely all marked as _optional_. But mere fact of collecting provider name and network id, deserves a separate asterisk / notion of a potential field that needs a privacy review.

I think it's ok to start simple and just pick one thing to specify, rather than trying to boil the ocean and specify every possible thing in a single PR. Let's keep this PR focussed on just describing the network that a mobile device is using for data communications, if that's ok.

specification/trace/semantic_conventions/span-general.md
bryce-b: @jkwatson I can add the mapping explicitly to the spec, it will be easy to do now that I have the Android `TelephonyProtoEnums`.

bryce-b: Additionally, we could revert back to my original proposal from the issue #1647 where I referred to the generation categories as the effective connection type, and add another field for the explicit radio name, e.g: `connection.type: 1xRTT , connection.effective_type: 3G`

bryce-b: Updated the spec with details about specific radios

specification/trace/semantic_conventions/span-general.md
bryce-b: https://github.com/open-telemetry/opentelemetry-specification/issues/1765

specification/trace/semantic_conventions/span-general.md
```diff
@@ -41,6 +41,12 @@ the `net.peer.*` properties of a client are equal to the `net.host.*` properties
 | `net.host.ip` | string | Like `net.peer.ip` but for the host IP. Useful in case of a multi-IP host. | `192.168.0.1` | No |
 | `net.host.port` | int | Like `net.peer.port` but for the host port. | `35555` | No |
 | `net.host.name` | string | Local hostname or similar, see note below. | `localhost` | No |
+| `net.host.connection.type` | string | The internet connection type currently being used by the host. | `wifi` | No |
+| `net.host.connection.subtype` | string | This describes more details regarding the connection.type. It may be the type of cell connection, but it could be used for describing details about a wifi connection. | `4G` | No |
+| `net.host.carrier.name` | string | The name of the mobile carrier. | `sprint` | No |
```
maxgolov, line 46: How does this work for device that is dual-SIM?

specification/trace/semantic_conventions/span-general.md
```diff
@@ -41,6 +41,12 @@ the `net.peer.*` properties of a client are equal to the `net.host.*` properties
 | `net.host.ip` | string | Like `net.peer.ip` but for the host IP. Useful in case of a multi-IP host. | `192.168.0.1` | No |
 | `net.host.port` | int | Like `net.peer.port` but for the host port. | `35555` | No |
 | `net.host.name` | string | Local hostname or similar, see note below. | `localhost` | No |
+| `net.host.connection.type` | string | The internet connection type currently being used by the host. | `wifi` | No |
```
maxgolov, line 44: How does it work for device that is simultaneously connected to LTE and WiFi? In Android, a network can have multiple transports at the same time. An example of this would be a VPN operating over both Wi-Fi and cellular networks at once. You can't degrade this to a single connection type.

maxgolov:
> Let's keep this PR focussed on just describing the network that a mobile device is using for data communications, if that's ok.

I believe there are technical issues with this PR, even if it's scoped purely on describing the network state:
- dual-sim devices.
- multiple interface devices.
- explicit capture of any possible privacy implications of capturing network name and ID, alongside with a note that these fields must be optional; never mandatory.
- **network cost** - is much more critical and relevant to capture than network connection speed / network type, e.g. you don't want users to pay for telemetry.

Oberon00:
> dual-sim devices.
> multiple interface devices.

This is, at least semantically, not a problem as the PR proposes span attributes, not resource attributes. Presumably, every span involves only a single network interface that is actually used for the traced operation.

specification/trace/semantic_conventions/span-general.md
bryce-b: I'm not totally familiar with how dual-sim and multiple transports work. It might be best to have them split up into multiple spans and denote which portions of them are associated with which transport method.

bryce-b:
I think changes are wrapped up at this point, unless something else major comes up. Last thing I'm waiting on is https://github.com/open-telemetry/build-tools/issues/47

Oberon00:
@bryce-b It might be fastest if you try to fix it yourself, the affected code should be somewhere around here https://github.com/open-telemetry/build-tools/blob/main/semantic-conventions/src/opentelemetry/semconv/templating/markdown/__init__.py#L243

jmacd (approved):

jmacd:
@maxgolov I read your concerns above and want to acknowledge them. I believe I am correct in saying that most of the issues you raised are about instrumentation, not semantic conventions. The users of these semantic conventions will have to pay attention to privacy, for example. I am under the impression that there is a new SIG that will form for client-side instrumentation, which at this time exists as a discussion in the [#otel-client-side-telemetry room of the CNCF slack](https://cloud-native.slack.com/archives/C0239SYARD2).

semantic_conventions/trace/general.yaml
```diff
@@ -60,6 +60,62 @@ groups:
...
+            - id: cell
+              value: "cell"
+            - id: unavailable
+              value: "unavailable"
+        brief: 'The internet connection type currently being used by the host.'
+        examples: 'wifi'
+      - id: host.connection.subtype
+        type:
+          allow_custom_values: true
+          members:
```
jkwatson, line 80: do we also want to put in the common "named" ones here? `LTE`, `EDGE`, etc?

bryce-b: `EDGE` and `LTE` are covered under `2G` and `4G` respectively... I'm not against adding familiar cellular networks, but I'm not sure I'm the best person to enumerate the most familiar ones. Maybe we can add a provision to the spec saying: "if you want to add the specific network to `host.connection.subtype` go right ahead. Since it is flag as allowing custom values.

jkwatson: As @maxgolov commented..there's no easy way to map from the values we get in Android to these values, so we should either 

a) enumerate everything we know, from both iOS and Android or 
b) maybe not have this be specified at all at this point. 

Putting in only iOS-friendly values makes it difficult for Android to know what to do.

bryce-b: I think there might be a little cross talk here: android and iOS phones use the same cell networks, and the cell technology the phone is available through the os. I think the issue @maxgolov was getting at is these values are only available if certain os permissions are allowed in the app.

bryce-b: Just for some clarification: here are the same values enumerated in the Android `TelephonyManager`
https://android.googlesource.com/platform/frameworks/base.git/+/master/telephony/java/android/telephony/TelephonyManager.java#2782
```
private static final @NetworkType int[] NETWORK_TYPES = {
--
  | NETWORK_TYPE_GPRS,
  | NETWORK_TYPE_EDGE,
  | NETWORK_TYPE_UMTS,
  | NETWORK_TYPE_CDMA,
  | NETWORK_TYPE_EVDO_0,
  | NETWORK_TYPE_EVDO_A,
  | NETWORK_TYPE_1xRTT,
  | NETWORK_TYPE_HSDPA,
  | NETWORK_TYPE_HSUPA,
  | NETWORK_TYPE_HSPA,
  | NETWORK_TYPE_IDEN,
  | NETWORK_TYPE_EVDO_B,
  | NETWORK_TYPE_LTE,
  | NETWORK_TYPE_EHRPD,
  | NETWORK_TYPE_HSPAP,
  | NETWORK_TYPE_GSM,
  | NETWORK_TYPE_TD_SCDMA,
  | NETWORK_TYPE_IWLAN,
  | NETWORK_TYPE_LTE_CA,
  | NETWORK_TYPE_NR
  | };
```

jkwatson: What values does iOS give you for the cellular network type?

bryce-b: Here are the iOS values : https://developer.apple.com/documentation/coretelephony/cttelephonynetworkinfo/radio_access_technology_constants

carlosalberto (approved):

semantic_conventions/trace/general.yaml
jkwatson: oh! Well, given that, I think we should *not* use 1G, 2G, etc, but instead use the actual network type names (EDGE, UMTS, LTE, NR, etc).

bryce-b: I'll make an update shortly

jkwatson (approved):

semantic_conventions/trace/general.yaml
```diff
@@ -60,6 +60,84 @@ groups:
...
+            - id: wifi
+              value: "wifi"
+            - id: wired
+              value: "wired"
+            - id: cell
+              value: "cell"
+            - id: unavailable
+              value: "unavailable"
```
jkwatson, line 74: does this mean that the network type isn't known or that there is no network available?

bryce-b: due to vagueness in the OS network status API it could mean either, cross referencing other details about the trace could reveal which it is: e.g. the transaction results in a `200 OK` means the os wasn't able to get details about the connection where as a network error such as `NSURLErrorNotConnectedToInternet` means the network connection is unavailable. But these details are only available after making a network request

jkwatson: I think on Android we can (at least sometimes) know for sure that this is no network available. I think we should distinguish between "no network" and "can't know if there is or isn't a network" with two separate values here.

Maybe "unavailable" and "unknown"?

jmacd:
This has only one approval since `net.host.connection.subtype` changed. @open-telemetry/specs-approvers please take another look.

nachoBonafonte (commented):
This is a needed PR for mobile tracking

bryce-b:
@carlosalberto could you re-approve, since there were a few changes since your original approval?

carlosalberto (approved):

bryce-b:
@jmacd how are we looking for merge?

jmacd:
Thanks @bryce-b!

PR #1761, opened 2021-06-15
Add Network conntrack metrics

emilgelman:
## Changes

Suggestion to add two network statistic metrics for Linux connection tracking, taken from https://manpages.debian.org/testing/conntrack/conntrack.8.en.html

Implementation PR:
https://github.com/open-telemetry/opentelemetry-collector/pull/3370

### DIFF
--- a/specification/metrics/README.md
+++ b/specification/metrics/README.md
@@ -12,8 +12,8 @@
 * [Josh Suereth](https://github.com/jsuereth)
 * [Joshua MacDonald](https://github.com/jmacd)
 
-Note: this specification is subject to major changes. To avoid thrusting
-language client maintainers, we don't recommend OpenTelemetry clients to start
+Note: this specification is subject to major changes. To ensure the best use of
+language client maintainers time, we don't recommend OpenTelemetry clients to start
 the implementation unless explicitly communicated via
 [OTEP](https://github.com/open-telemetry/oteps#opentelemetry-enhancement-proposal-otep).
 


specification/metrics/semantic_conventions/system-metrics.md
```diff
@@ -110,19 +110,21 @@ perf counter (similar for Writes)
...
+|                                                                                                                                      |                |                   |            | direction | transmit, receive                                                                              |
+| system.network.connections                           |                                                                               | {connections}  | UpDownSumObserver | Int64      | device    | (identifier)                                                                                   |
+|                                                      |                                                                               |                |                   |            | protocol  | tcp, udp, [etc.](https://en.wikipedia.org/wiki/Transport_layer#Protocols)                      |
+| system.conntrack.count<sup>[3](#linux)</sup>         | Number of currently allocated flow entries                                    | {flow entries} | UpDownSumObserver | Int64      |           |                                                                                                |
```
jmacd, line 125: I was initially confused because there are similar but different "conntrack" metrics in the Prometheus server, so these names may confuse others. Can we find a more descriptive name, perhaps `system.connections.count`?

specification/metrics/semantic_conventions/system-metrics.md
```diff
@@ -110,19 +110,21 @@ perf counter (similar for Writes)
...
+|                                                      |                                                                               |                |                   |            |           |                                                                                                |
+| system.conntrack.max<sup>[3](#linux)</sup>           | Size of connection tracking table                                             | {flow entries} | UpDownSumObserver | Int64      |           |                                                                                                |
```
jmacd, line 127: Can this value be calculated from the history of the `.count` metric above? If yes, this might become a view definition and we probably do not need to report it. Otherwise, I'd like to find a name other than `.max` to describe the size of a table here.

emilgelman: @jmacd Thank you for the review.
I was trying to stay consistent with the naming convention from 
https://www.kernel.org/doc/Documentation/networking/nf_conntrack-sysctl.txt

The value cannot be calculated from count, and I agree calling it `max` is a bit confusing.

What do you think about:
`system.connections.table.count`
`system.connections.table.size`

reyang:
Heads up, we're removing "labels" and using "attributes" everywhere #1775.

@jtmalinowski FYI.

specification/metrics/semantic_conventions/system-metrics.md
jmacd: How about the `.usage` and `.limit` pattern, i.e., 
`system.connections.table.usage`
`system.connections.table.limit`
See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/README.md#general-metric-semantic-conventions

PR #1762, opened 2021-06-15
Clarify the limit on the instrument unit

### LINKED ISSUE #1746: confusing limit on unit size
The spec says that a unit string "can have a maximum length of 63 characters. The number 63 is chosen to allow the unit strings (including the \0 terminator on certain language runtimes) to be stored and compared as 8-bytes integers when performance is critical."

You can only fit 8 (ASCII) characters into 8 bytes, so I'm not sure what the intent here is (eight 8-byte integers?), or if this is just a mistake.

Also, I see elsewhere (#705) that units should follow the [UCUM standard](https://ucum.org/ucum.html), but I don't see any length restrictions there.

reyang:
Fixes #1746.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ release.
 
 ### Metrics
 
+- Clarify the limit on the instrument unit.
+  ([#1762](https://github.com/open-telemetry/opentelemetry-specification/pull/1762))
+
 ### Logs
 
 ### Resource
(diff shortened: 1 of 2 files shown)


jsuereth (approved):

cijothomas (approved):

jmacd (approved):

MrAlias (approved):

weyert (approved):

carlosalberto:
Can this be merged? It has more than enough approvals already.

PR #1770, opened 2021-06-21
Add span.TracerProvider() method definition to trace API

Aneurysm9:
## Changes

Defines a method on `Span` for accessing the `TracerProvider` used to create the `Span`.  This minimizes the configuration overhead of instrumentation libraries and can simplify the handling of multiple `TracerProvider` configurations by enabling instrumentation to simply follow the lead of the `Span` that is currently active in the context.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ release.
 
 ### Traces
 
-- Add `span.TracerProvider()` method to API. (#TBD)
+- Add `span.TracerProvider()` method to API. ([#1770](https://github.com/open-telemetry/opentelemetry-specification/pull/1770))
 
 ### Metrics
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ release.
 
 ### Traces
 
+- Add `span.TracerProvider()` method to API. (#TBD)
+
 ### Metrics
 
 ### Logs
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -665,6 +665,17 @@ Note: `RecordException` may be seen as a variant of `AddEvent` with
 additional exception-specific parameters and all other parameters being optional
 (because they have defaults from the exception semantic convention).
 
+#### Get TracerProvider
+
+To simplify the configuration of instrumented libraries that may not rely on a
+global `TracerProvider` instance, a `Span` SHOULD provide a method to retrieve
+the `TracerProvider` instance that was used to create the `Tracer` that created
+the `Span`.  If this method is provided but accessing the `TracerProvider`
+instance is not possible, then a `Span` MUST return a no-op `TracerProvider`.
+
+The returned value MUST be the same for the entire Span lifetime. This method
+SHOULD be called `TracerProvider`.
+
 ### Span lifetime
 
 Span lifetime represents the process of recording the start and the end
--- a/spec-compliance-matrix.md
+++ b/spec-compliance-matrix.md
@@ -52,6 +52,7 @@ formats is required. Implementing more than one format is optional.
 | events collection size limit                                                                     |          | +  | +    | +  | +      | +    | -      |     | +    | -   | -    | +     |
 | attribute collection size limit                                                                  |          | +  | +    | +  | +      | +    | -      |     | +    | -   | -    | +     |
 | links collection size limit                                                                      |          | +  | +    | +  | +      | +    | -      |     | +    | -   | -    | +     |
+| get TracerProvider                                                                               |          | +  |      |    |        |      |        |     |      |     |      |       |
 | [Span attributes](specification/trace/api.md#set-attributes)                                     |          |    |      |    |        |      |        |     |      |     |      |       |
 | SetAttribute                                                                                     |          | +  | +    | +  | +      | +    | +      | +   | +    | +   | +    | +     |
 | Set order preserved                                                                              | X        | +  | -    | +  | +      | +    | +      | +   | +    | +   | +    | +     |


MrAlias (approved):

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ release.
 
 ### Traces
 
+- Add `span.TracerProvider()` method to API. (#TBD)
```
MrAlias, line 14:
```suggestion
- Add `span.TracerProvider()` method to API. ([#1770](https://github.com/open-telemetry/opentelemetry-specification/pull/1770))
```

yurishkuro:
Reasons for NOT doing this:
  1. it's a breaking change to the Span interface
  2. it encourages dependency on global state for something that should be a local dependency
  3. it indirectly changes how the instrumentation starts traces (e.g. it could be starting traces now even when there is no parent span, but won't be able to do that after this change)

Aneurysm9:
> Reasons for NOT doing this:
> 
>     1. it's a breaking change to the Span interface

It is an additive, optional change to an API interface and we have agreed that additive changes to the API are permissible and not breaking.  
 
>     2. it encourages dependency on global state for something that should be a local dependency

It does the opposite of this.  It avoids taking a dependency on the global `TracerProvider` instance and allows multiple `TracerProviders` to be used concurrently without the instrumentation needing to be aware that it is happening or when to use which `TracerProvider`.  An instrumentation author can simply extract a `Span` from a `Context`, get the `TracerProvider` from that `Span`, create a `Tracer` from that `TracerProvider` and then start a `Span` that will be processed with the same processing pipeline as the parent `Span`.

>     3. it indirectly changes how the instrumentation starts traces (e.g. it could be starting traces now even when there is no parent span, but won't be able to do that after this change)

Again, this is an optional, additive change to the API.  Nothing that could be done before cannot be done after this change.  For instrumentation that expects to start traces or receive remote `SpanContext` from the propagation system and start a new span in the current process, they will absolutely still need to be configured explicitly with a `TracerProvider` or utilize the global `TracerProvider`.  For instrumentation that expects spans to already exist in the process and to be carried by a `Context`, this provides a new option for them to discover the appropriate `TracerProvider` to use.

tigrannajaryan:
> An instrumentation author can simply extract a `Span` from a `Context`, get the `TracerProvider` from that `Span`, create a `Tracer` from that `TracerProvider` and then start a `Span` that will be processed with the same processing pipeline as the parent `Span`.

@Aneurysm9 Can you give an example use case when starting a new span by the instrumentation is needed?

Aneurysm9:
> > An instrumentation author can simply extract a `Span` from a `Context`, get the `TracerProvider` from that `Span`, create a `Tracer` from that `TracerProvider` and then start a `Span` that will be processed with the same processing pipeline as the parent `Span`.
> 
> @Aneurysm9 Can you give an example use case when starting a new span by the instrumentation is needed?

One example is the [`otelhttp` client helpers](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/net/http/otelhttp/client.go) that currently use a default transport that uses the global `TracerProvider`.  This is not useful in an application where there are multiple `TracerProvider`s active.  Perhaps more importantly, a library author instrumenting their library cannot know when they are in that situation, so they cannot ever use these convenience functions.  Indeed, even constructing an `otelhttp.Transport` of their own is perilous in that situation as the instrumentation would then need some way to track which `otelhttp.Transport` to use for which request.  

Instead, if the `TracerProvider` is available from the `context.Context` carried by the `http.Request`, the `otelhttp.Transport` could be rewritten to utilize the `TracerProvider` associated with the `Span` in each request as it is processed.  It could also still be configured to use an explicitly specified `TracerProvider` or the global `TracerProvider` should a user wish to start a trace from the `otelhttp.Transport` by making a request with no active `Span`.

specification/trace/api.md
```diff
@@ -665,6 +665,17 @@ Note: `RecordException` may be seen as a variant of `AddEvent` with
 additional exception-specific parameters and all other parameters being optional
 (because they have defaults from the exception semantic convention).
 
+#### Get TracerProvider
+
+To simplify the configuration of instrumented libraries that may not rely on a
+global `TracerProvider` instance, a `Span` SHOULD provide a method to retrieve
+the `TracerProvider` instance that was used to create the `Tracer` that created
```
reyang, line 672: For no-op span, what would be the TracerProvider returned? (for example, the sampler decided NOT to create a span, then a no-op span is returned, do we expect it to return the actual TracerProvider or we want something like NULL?)

reyang:
Based on the discussion in the 6/22/2021 Specification SIG meeting, it seems there are two different ways of looking at this PR:
1. It is a Golang specific thing - OpenTracing Go offers such feature, and in order to have the compatibility story, we need this as one-off.
2. It is a common requirement (@carlosalberto mentioned that folks using OpenTracing had similar asks - although OpenTracing Python didn't provide it).

specification/trace/api.md
Aneurysm9: I don't think that an SDK should return a no-op span when a sampler has decided to not sample, it should instead return a span that is not recording.  The span should still return the `TracerProvider` that was used and the sampler used by that `TracerProvider` should provide consistent results for all spans in the trace.

In the case of an actual no-op span it should still return the `TracerProvider` that created it, to the extent possible, or a no-op `TracerProvider` if it is not.  I would expect that that would be the same thing in most cases anyways since I would not expect an SDK `TracerProvider` to construct no-op spans.

bogdandrutu:
Question: How do we create a consistent API for metrics? In the future maybe for logs?

MrAlias:
> Question: How do we create a consistent API for metrics? In the future maybe for logs?

We talked in the Go SIG a bit about this with @jmacd. There is no definite conclusion, but the general sentiment seems to be that the metric signal pipeline is different enough that it would likely not need to worry about this type of instrumentation. Metric instruments are usually singularly defined and there isn't an inheritance issue related to a metric event like there is with span parent/child relationships. As for logs, we would probably pass a logger of some type in the context.

Overall though, this question should not be a blocker for this PR. Not sure if this question was posed as such, but requiring we "boil the ocean" here and have a solution for all potential signals will only serve to halt all progress on this PR.

PR #1775, opened 2021-06-23
Rename Metrics labels to attributes

### LINKED ISSUE #1774: Metrics docs inconsistently use "attributes" and "labels"

Currently in Metrics docs "attributes" and "labels" are used inconsitently to describe the same entity, ie. dimensions of measurements. For example https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md uses the term "attributes" and doesn't mention "labels", but https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/README.md uses "labels" and doesn't mention "attributes". My understanding is that we want to use "attributes" everywhere to describe the dimensions, except when we're talking about a specific protocol which uses a different term.

**Additional context.**

We [briefly discussed that in Slack](https://cloud-native.slack.com/archives/C01N7PP1THC/p1623685519017200) and I'm planning to send a PR which will perform the rename where it's appropriate.

jtmalinowski:
Fixes #1774 

## Changes

- all instances of `label` where it is used to describe a dimension in Metrics were renamed to `attribute`


specification/common/attribute-and-label-naming.md
```diff
@@ -1,146 +1,3 @@
 # Attribute and Label Naming
```
reyang, line 1: Curious - is this intended for folks who keep the external link, so that they won't get HTTP 404?

reyang: If the answer is YES, I think it is not a concern since the common practice is to refer to a persistent link (a link that tights to a tag/commit-hash).

specification/common/attribute-naming.md
```diff
@@ -0,0 +1,144 @@
+# Attribute Naming
+
+**Status**: [Experimental](../document-status.md)
```
reyang, line 3: I would recommend that we rename the file, so the change history is preserved.

specification/library-layout.md
```diff
@@ -36,7 +36,7 @@ This directory describes the Metrics API that can be used to record application
 ### [/baggage](baggage/api.md)
 
 This directory describes the Baggage API that can be used to manage context propagation
-and metrics-related labeling.
+and metrics-related attribution.
```
reyang, line 39: Not being a native English speaker, this sounds weird to me.

Perhaps "and providing contextual attributes to other telemetry signals, for example, providing extra metrics dimension"?

Let's get feedback from more folks.

specification/common/attribute-and-label-naming.md
jtmalinowski: the answer is yes, fixed

reyang (approved):
LGTM.

carlosalberto:
@open-telemetry/specs-metrics-approvers Can we get one more approval (or reviews)?

specification/library-layout.md
bogdandrutu: @jmacd @jsuereth any feedback?

MrAlias: This reads fine to me if we are trying to say the Baggage API can be used to describe things that were the cause of metric events. Though, I suspect we are instead trying to just talk about the attributes of those metric events, and not necessarily the Baggage API describing the cause of the events. If so, I would recommend the following. 

```suggestion
and metric event attributes.
```

MrAlias (approved):

carlosalberto:
>  I suspect we are instead trying to just talk about the attributes of those metric events, and not necessarily the Baggage API describing the cause of the events.

You are correct ;) Thanks for the suggestion!

Issue #1778, opened 2021-06-28
Issues with FaaS semantic conventions (too AWS-specific, faas.id/version unclear)


Oberon00:
The current FaaS conventions were developed with AWS Lambda in mind and have several issues. All in all, I think they are not GA-ready yet (good thing all semantic conventions are still experimental). At least some of
- Google Cloud Functions
- Google Cloud Run
- Google Firebase Function (?)
- Azure Functions

should also be well-supported by them (except if someone knows that the market share of one of them insignificant?), or they should be moved to an `aws` (sub-)namespace.

Some issues:
- The `faas.id` attribute is very nice but it is not clear to me how it applies to other providers. The current note on what to use for Azure (FunctionDirectory) seems wrong, we probably want something like https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftweb. For Google Cloud, there seems to be https://cloud.google.com/iam/docs/full-resource-names though I couldn't find how function IDs in particular would look like.
- Even just looking at AWS, the relationship between `faas.id`, `faas.version` and `faas.invoked_id` is not very clear, if you consider that you can call an AWS Lambda by invoking `arn:aws:lambda:us-east-1:12345:function:somefunction:somealias` which may invoke the function version `arn:aws:lambda:us-east-1:12345:function:somefunction:12`. Probably `faas.invoked_id` at the client side should contain the ARN ending in `:somefunction:somealias`, but what should `faas.id` contain? And wouldn't it be nice to have the `faas.invoked_arn` also on the server side? Note that the function name including alias cannot be used as `faas.id` since the same running instance could be invoked with multiple aliases and `faas.id` is a resource attribute, so there can only be one.
- faas.version seems to exist in AWS Lambda and Google Cloud Run ([revision](https://cloud.google.com/run/docs/managing/revisions)), but not Google Cloud Functions or Azure Functions ([deployment slots](https://docs.microsoft.com/en-us/azure/azure-functions/functions-deployment-slots) are something different, looks like the combination of an alias and a mutable version when compared to AWS Lambda)
- faas.instance, faas.execution: Unclear. These would need per-provider specification on what it is. For example, I could imagine using the request ID on AWS Lambda for faas.execution would make sense.

CC:
- @thisthat (as the original creator of the FaaS conventions)
- @anuraaga (as AWS engineer)
- @reyang (for Azure)
[...]

PR #1736, opened 2021-05-31
Add FaaS metrics semantic conventions

skonto:
Fixes #1013

## Changes
Adds metrics and metric labels for functions

I am proposing also changing the `faas` prefix to `func` as not all environments are of that kind.
Do I need to discuss that change in a WG meeting?

/cc @kolanos @kenfinnigan @jmacd

PR #1781, opened 2021-06-29
Rework FaaS semantic conventions.

### LINKED ISSUE #1280: Introduce a new resource attribute for AWS Lambda function alias

I propose adding a new optional resource attribute that will store the AWS Lambda function alias (if it's used at all), named `aws.lambda.alias`.

**Additional context.**

* [AWS Lambda alias docs](https://docs.aws.amazon.com/lambda/latest/dg/configuration-aliases.html)

Oberon00:
Fixes #1280 (through introduction of aws.lambda.invoked_arn span attribute, @mateuszrzeszutek please check if that is OK).
Addresses #1778.

## Changes

- More clearly describe some FaaS attributes, in particular ID-related ones.
- Add `aws.lambda.invoked_arn` and clarify that alias must be stripped from `faas.id` resource attribute for AWS Lambda.
- Describe what IDs to use for Azure and GCP (Note: It might be impractical to collect these IDs client-side).

CC @skonto

### DIFF
--- a/semantic_conventions/resource/faas.yaml
+++ b/semantic_conventions/resource/faas.yaml
@@ -21,7 +21,8 @@ groups:
           Take care not to use the "invoked ARN" directly but replace any
           [alias suffix](https://docs.aws.amazon.com/lambda/latest/dg/configuration-aliases.html) with the resolved function version, as the same runtime instance may be invokable with multiple
           different aliases.
-          * **GCP:** The [URI of the resource](https://cloud.google.com/iam/docs/full-resource-names), * **Azure:** The [Fully Qualified Resource ID](https://docs.microsoft.com/en-us/rest/api/resources/resources/get-by-id).
+          * **GCP:** The [URI of the resource](https://cloud.google.com/iam/docs/full-resource-names)
+          * ***Azure:** The [Fully Qualified Resource ID](https://docs.microsoft.com/en-us/rest/api/resources/resources/get-by-id).
 
           On some providers, it may not be possible to determine the full ID at startup,
           which is why this field cannot be made required. For example, on AWS the account ID
@@ -33,7 +34,7 @@ groups:
         type: string
         brief: The immutable version of the function being executed.
         note: |
-          Depending on the cloud provider, use:
+          Depending on the cloud provider and platform, use:
 
           * **AWS Lambda:** The [function version](https://docs.aws.amazon.com/lambda/latest/dg/configuration-versions.html)
             (an integer represented as a decimal string).
--- a/semantic_conventions/resource/faas.yaml
+++ b/semantic_conventions/resource/faas.yaml
@@ -9,6 +9,12 @@ groups:
         required: always
         brief: >
           The name of the single function that this runtime instance executes.
+        note:
+          This is the name of the function as configured/deployed on the FaaS
+          platform and is usually different from the name of the callback
+          function (which may be stored in the
+          [`code.namespace`/`code.function`](../../trace/semantic_conventions/span-general.md#source-code-attributes)
+          span attributes).
         examples: ['my-function']
       - id: id
         type: string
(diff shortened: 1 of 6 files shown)


semantic_conventions/resource/faas.yaml
```diff
@@ -8,32 +8,50 @@ groups:
         type: string
         required: always
```
carlosalberto, line 9: Not required? The paragraph below seems to indicate we may not be able to make it required.

Oberon00: The paragraph below is about the `id` for which I removed the `required`. But just the name should be available anywhere.

semantic_conventions/resource/faas.yaml
```diff
@@ -8,32 +8,56 @@ groups:
...
-          [ARN](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html)
-          value, in GCP to the URI of the resource, and in Azure to the
-          [FunctionDirectory](https://github.com/Azure/azure-functions-host/wiki/Retrieving-information-about-the-currently-running-function)
-          field.
+          The unique ID of the single function that this runtime instance executes.
+        note: |
+          Depending on the cloud provider, use:
+
+          * **AWS Lambda:** The function [ARN](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html).
+          Take care not to use the "invoked ARN" directly but replace any
```
anuraaga, line 27: Do you know if this is possible?

Oberon00: The string replacement or that "the  same runtime instance may be invokable with multiple different aliases"? Yes, I verified both. You can easily produce that scenario with the Lambda console UI by creating two different aliases for `$LATEST` and then you can test both and you will see that the logs end up in the same logstream without restart.

You can even test it without creating an extra alias. E.g. using `aws lambda invoke --function-name 'arn:aws:lambda:us-east-1:12345:function:myfunction:$LATEST` will end up with a different invoked ARN than just `aws lambda invoke --function-name 'arn:aws:lambda:us-east-1:12345:function:myfunction`, but the same invoked version & log stream.

thisthat (approved):

anuraaga (approved):

carlosalberto:
@mateuszrzeszutek Please review whenever you have cycles ;)

skonto:
Thanks @Oberon00 looks good to me. I will adjust #1736 as needed.

specification/trace/semantic_conventions/instrumentation/aws-lambda.md
```diff
@@ -35,9 +35,30 @@ Lambda `Context`.
...
 The following attributes SHOULD be set:
 
-- [`faas.execution`](../faas.md) - The value of the AWS Request ID, which is always available through an accessor on the Lambda `Context`
-- [`faas.id`](../../../resource/semantic_conventions/faas.md) - The value of the invocation ARN for the function, which is always available through an accessor on the Lambda `Context`
-- [`cloud.account.id`](../../../resource/semantic_conventions/cloud.md) - In some languages, this is available as an accessor on the Lambda `Context`. Otherwise, it can be parsed from the value of `faas.id` as the fifth item when splitting on `:`
+- [`faas.execution`][faas] - The value of the AWS Request ID, which is always available through an accessor on the Lambda `Context`.
+- [`faas.id`][faasres] - The value of the invocation ARN
+  for the function, which is always available through an accessor on the
+  Lambda `Context`, modified as follows: Discard all parts beyond the seventh (when split on `:`;
+  the seventh part is the function name) and append the [`faas.version`][faasres], separated by a colon.
```
mateuszrzeszutek, line 42: 👍

mateuszrzeszutek (approved):
LGTM 👍

carlosalberto:
@Oberon00 Please update your branch so we can merge this 👍

PR #1790, opened 2021-07-05
More explanation for OTLP transport protocols

### LINKED ISSUE #1764: Confusing transport protocols
The [spec currently says](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#specify-protocol):

> OTLP has more than one transport protocol it can support, e.g. grpc, http/json, http/protobuf

That is all that is said about this. Is this list exhaustive? MUST every SDK support them all? What exactly do they mean? E.g. what is the difference between gRPC and http/protobuf when gRPC already sends protobuf encoded data over HTTP.

iNikem:
Fixes #1764

## Changes

More explanation for OTLP transport protocols. Also clearly state that the support for one of the `grpc` or `http/protobuf` is required and `http/json` is not enough.


specification/protocol/exporter.md
```diff
@@ -41,9 +41,15 @@ export OTEL_EXPORTER_OTLP_METRICS_ENDPOINT=https://collector.example.com/v1/metr
 
 ### Specify Protocol
 
-Currently, OTLP has more than one transport protocol it can support, e.g.
-`grpc`,  `http/json`, `http/protobuf`.   As of 1.0 of the specification, there
-*is no specified default, or configuration via environment variables*.  We
+Currently, OTLP supports the following transport protocols:
+
+- `grpc` for protobuf-encoded data over HTTP/2 connection using [gRPC framework](https://grpc.io/)
```
anuraaga, line 46: How about for protobuf-encoded data using grpc wire format over HTTP/2

specification/protocol/exporter.md
```diff
@@ -41,9 +41,15 @@ export OTEL_EXPORTER_OTLP_METRICS_ENDPOINT=https://collector.example.com/v1/metr
...
+- `http/protobuf` for protobuf-encoded data over HTTP/1.1 connection
```
anuraaga, line 47: We could leave out 1.1 I guess. The main point here is lack of grpc wire format, otherwise it works with both 2.0 and 1.1. More specifically, we want to avoid implementations specifically disabling HTTP/2 for this protocol since there's no benefit to doing so.

specification/protocol/exporter.md
iNikem: Is there such thing as "gRPC wire format"? In addition to "protobuf over HTTP/2"?

specification/protocol/exporter.md
iNikem: But at the same time I think it is beneficial to stress that HTTP/1.1 still has to be supported.

specification/protocol/exporter.md
anuraaga: Yeah notably it adds a byte of metadata and a length prefix before each proto message in the payload, even for unary.

https://www.github.com/grpc/grpc/tree/master/doc%2FPROTOCOL-HTTP2.md

anuraaga: By the way here is code that implements only the wire format with nothing about protobuf in case it's easier to understand.

https://github.com/line/armeria/blob/master/grpc-protocol/src/main/java/com/linecorp/armeria/client/grpc/protocol/UnaryGrpcClient.java

It's actually not _that hard_ to implement unary grpc without the grpc library and it works fine on http/1. It could be an alternative to the http/protobuf support.

specification/protocol/exporter.md
anuraaga: Then 1.1+? Though I think it's fairly standard for the term "HTTP" to imply supporting at least 1.1 in practice.

specification/protocol/exporter.md
iNikem: Done

specification/protocol/exporter.md
iNikem: Ok, agreed :)

anuraaga (approved):

specification/protocol/exporter.md
```diff
@@ -41,9 +41,15 @@ export OTEL_EXPORTER_OTLP_METRICS_ENDPOINT=https://collector.example.com/v1/metr
...
-Currently, OTLP has more than one transport protocol it can support, e.g.
-`grpc`,  `http/json`, `http/protobuf`.   As of 1.0 of the specification, there
-*is no specified default, or configuration via environment variables*.  We
+Currently, OTLP supports the following transport protocols:
+
+- `grpc` for protobuf-encoded data using gRPC wire format over HTTP/2 connection
+- `http/protobuf` for protobuf-encoded data over HTTP connection
+- `http/json` for JSON-encoded data over HTTP connection
+
+SDKs MUST support either `grpc` or `http/protobuf` and SHOULD support both. They also MAY support `http/json`.
```
yurishkuro, line 50: I haven't seen a definitive analysis in the ticket of why one of protobuf-based protocols is a MUST, can you elaborate?

Specifically, in the past we've always had issues with supporting Thrift in the browser - I don't know the current state of protobuf for browser, but according to https://github.com/open-telemetry/opentelemetry-specification/issues/1764#issuecomment-866805714 it's a similar story, JSON is preferred.

iNikem: https://github.com/open-telemetry/opentelemetry-specification/issues/1764#issuecomment-866114683 and https://github.com/open-telemetry/opentelemetry-specification/issues/1764#issuecomment-868540940

It was also discussed during the last maintainers meeting (https://youtu.be/ZqQ4sbOsvLs?t=887) and this decision was made there.

anuraaga: Perhaps instead of SDKs, Languages is more obvious? I suppose the intent is that JS supports at least protobuf for node users, and if protobuf is prohibitive in a browser it can also support JSON for that specific use case.

carlosalberto (approved):

tigrannajaryan (approved):

carlosalberto:
I think we are ready to go. @iNikem would you mind adding a Changelog entry? Thanks!

iNikem:
> I think we are ready to go. @iNikem would you mind adding a Changelog entry? Thanks!

@carlosalberto done.

jmacd (approved):

PR #1792, opened 2021-07-05
Update instrument names

### LINKED ISSUE #1650: HTTP active_requests metric should be LongUpDownCounter?
I'm looking at https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md#http-server and it says the instrument to use is UpDownSumObserver. Should it be `LongUpDownCounter` since we would be counting up or down as requests start and stop, not observing a static value?

### LINKED ISSUE #1788: Update metrics semantics_conventions for new instrument names

The semantics conventions developed for metrics use old instrument names, for example [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/system-metrics.md).

ValueRecorder should be replaced by Histogram, ValueObserver should be replaced by Gauge, UpDownSumObserver becomes UpDownCounter, SumObserver becomes Counter. 

**Additional context.**

The semantics conventions do not need to state whether an instrument is asynchronous or not, this is not a meaningful distinction in the data model. There's no need to call out "asynchronous" anywhere in the semantics conventions.

pirgeo:
Fixes #1788
Fixes #1650 

## Changes

* Updated the Instruments that should be used with metrics specified in the semantic conventions.
* Reformatted the markdown tables.


arminru (approved):
LGTM!

specification/metrics/semantic_conventions/system-metrics.md
```diff
@@ -31,45 +31,45 @@ instruments not explicitly defined in the specification.
...
 **Description:** System level memory metrics. This does not include [paging/swap
 memory](#systempaging---pagingswap-metrics).
 
-| Name                      | Description | Units | Instrument Type   | Value Type | Attribute Key | Attribute Values         |
-| ------------------------- | ----------- | ----- | ----------------- | ---------- | ------------- | ------------------------ |
-| system.memory.usage       |             | By    | UpDownSumObserver | Int64      | state         | used, free, cached, etc. |
-| system.memory.utilization |             | 1     | ValueObserver     | Double     | state         | used, free, cached, etc. |
+| Name                      | Description | Units | Instrument Type | Value Type | Attribute Key | Attribute Values         |
+| ------------------------- | ----------- | ----- | --------------- | ---------- | ------------- | ------------------------ |
+| system.memory.usage       |             | By    | UpDownCounter   | Int64      | state         | used, free, cached, etc. |
```
jsuereth, line 46: `UpDownSumObserver` should no be called either `Asynchronous UpDownCounter` or `ObservableUpDownCounter`, see: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#asynchronous-updowncounter


Similar comment for remaining sum-observer types in this PR.

jsuereth (changes requested):
Glad to see someone doing this!

Think maybe the difference between `ObservableCounter` and `CounterObserver` may not be clear in the new API specification, but basically that's the only change that happened to those.

specification/metrics/semantic_conventions/system-metrics.md
pirgeo: I wasn't sure about that since #1788 explicitly calls out that synchronicity is not of consequence here. Happy to change it, though. Would that also mean updating all the `Counter`s to `ObservableCounter`s, or is this only relevant to the `UpDownCounter`s?

arminru: @jmacd Please take a look 🙂

arminru: @jmacd friendly ping since you specified this on #1788.

Issue #1085, opened 2020-10-11
Questions about span relationships in messaging semantic convention


iNikem:
## Problem description

I noticed an inconsistency in how [messaging semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md) describe parent-child relationships between producers and consumers.

[First example](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md#topic-with-multiple-consumers) says that  consumers should have message producer as their parent. But [second example](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md#batch-receiving) says that

>the propagated trace and span IDs are not known when the receiving span is started

and thus we don't have parent-child relation between consumer spans and producers. Also there is no indication what parent should "Receive" span have.

And then [third example](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md#batch-processing) contradicts it again and tell receiving spans to have message producers as their parent. Why second example couldn't have parents and this one can? And how shall backends handle such convention which essentially say "everything can happen"?

### Problem with old remote parent

First I would like to start a discussion about having message producer as a parent span for message consumers in an async case. Although it certainly does make sense in the majority of cases of streaming or event-based platforms, I wonder if this makes sense in situations, when we can reasonably expect to process old messages. Like in messages reprocessing in event-source systems or lambda-architectures (or kappa). If there is a gap of several days (or weeks) between consumer span and producer span, does it still make sense to have producer span as a parent?

### Pull vs push consumption

I see two different ways to consume messages. Push-based consumption uses some kind of listener mechanism and focuses on message processing. Message retrieval is done by framework and in general of no interest to us. Pull-based consumption has an explicit step of "requesting" or "receiving" the message and then processing it.

Despite the problem with old remote parents, I think that in push-based systems the only sensible relationship between message producing and its processing is the parent-child one. Message processing is done in the context of some retrieve/process cycle of the framework which we don't want to see as the parent.

In pull-based system the question "what span should be the parent of whom" is not that clear to me. Somewhere in _my application's code_ there is something similar to the following pseudo-code:
```
var messages = queue.pollNext()
for m in messages {
  process(m)
}
```

[...]

PR #1027, opened 2020-09-28
Resolve semantic inconsistencies for non traditional messaging

kenfinnigan:
Fixes #977

## Changes

Update tracing semantic conventions for messaging systems to better reflect "non-traditional" approaches such as Apache Kafka.

Include recommended semantic attributes for Apache Kafka

Issue #65, opened 2019-04-22
Sync and Async children (FOLLOWS_FROM)


tedsuo:
In OpenTracing, we have `CHILD_OF` and `FOLLOWS_FROM`. In the new project, we are considering whether to include this concept as a flag on the `SpanBuilder` option when setting the span parent. The new naming is proposed to be `sync` and `async` children, to make the relationship more clear.

Reference PR: https://github.com/bogdandrutu/openconsensus/pull/130

Questions:
* Do we still want this at all? It can be useful for critical path and other types of trace analysis.
* Do we also need an `unknown` flag as well?

rochdev:
In Node this concept is very important and is core to how context propagation works in the runtime itself. For this purpose specifically, we call them `execution` and `trigger`. In Node you can only have one of each because it's limited to function calls, but from a tracing perspective it makes sense that you could follow from multiple different operations.

Let me give an example specific to Node at the language level:

```js
const promise = new Promise((resolve, reject) => {
  resolve() // execution ID here it 2
})

// execution ID here is 1
promise.then(() => {
  // here, execution ID is 1 and trigger ID is 2
})
```

The reasoning for the above is that `resolve()` is what triggered the execution of the callback, but the callback was actually registered when `then()` was called, so that's its execution parent.

For a case like this, we could then say that the callback is running as a `ChildOf` the context where `then()` was called, and `FollowsFrom` the context where `resolve()` was called.

It's possible I got this completely wrong and that `ChildOf/FollowsFrom` has nothing to do with the relationship described above. I think the best person to explain the real meaning of these is probably @tedsuo.

In general, I think the different wordings proposed in this thread make sense, but they don't necessarily map 1:1 with each other.

yurishkuro:
The problem here is slightly bigger than what to call these. Kudos @rochdev for thinking OpenTracing got it right, but it didn't, not quite (cf. [this blog post][1]). The most fundamental question in analyzing the graph of events is the Lamport's `happens-before` relationship. In the OpenTracing span model the following holds:
```
parent.start  happens-before child.start
```
[...]

PR #1799, opened 2021-07-07
Update Kafka messaging example

trask:
I think(?) this was just an oversight in the example, and "Span Prod2" should parent "Span Proc1" (and no need for a link to "Span Prod1").

### DIFF
--- a/specification/trace/semantic_conventions/messaging.md
+++ b/specification/trace/semantic_conventions/messaging.md
@@ -258,8 +258,8 @@ Process CB:                           | Span Rcv2 |
 | Field or Attribute | Span Prod1 | Span Rcv1 | Span Proc1 | Span Prod2 | Span Rcv2
 |-|-|-|-|-|-|
 | Span name | `"T1 send"` | `"T1 receive"` | `"T1 process"` | `"T2 send"` | `"T2 receive`" |
-| Parent |  | Span Prod1 | Span Rcv1 |  | Span Prod2 |
-| Links |  |  | | Span Prod1 |  |
+| Parent |  | Span Prod1 | Span Rcv1 | Span Proc1 | Span Prod2 |
+| Links |  |  | | |  |
 | SpanKind | `PRODUCER` | `CONSUMER` | `CONSUMER` | `PRODUCER` | `CONSUMER` |
 | Status | `Ok` | `Ok` | `Ok` | `Ok` | `Ok` |
 | `peer.service` | `"myKafka"` |  |  | `"myKafka"` |  |


specification/trace/semantic_conventions/messaging.md
```diff
@@ -258,8 +258,8 @@ Process CB:                           | Span Rcv2 |
 | Field or Attribute | Span Prod1 | Span Rcv1 | Span Proc1 | Span Prod2 | Span Rcv2
 |-|-|-|-|-|-|
 | Span name | `"T1 send"` | `"T1 receive"` | `"T1 process"` | `"T2 send"` | `"T2 receive`" |
-| Parent |  | Span Prod1 | Span Rcv1 |  | Span Prod2 |
-| Links |  |  | | Span Prod1 |  |
+| Parent |  | Span Prod1 | Span Rcv1 | Span Proc1 | Span Prod2 |
+| Links |  |  | | |  |
```
arminru, line 262: I already made a comment about this in the original PR at https://github.com/open-telemetry/opentelemetry-specification/pull/1027#discussion_r498093519 where @kenfinnigan then explained why a parent/child relationship might not always be possible or applicable.
Maybe we understand this problem better now that the first Kafka instrumentations based on these semantic conventions were already developed.
CC @anuraaga who might have an opinion here.

kenfinnigan: My views have evolved somewhat, but I think it's also colored by https://github.com/open-telemetry/opentelemetry-specification/issues/65 not being resolved, and how the spans look when visualized.

While I still think it makes sense for producing a new message to have no parent on the "process" of the existing one, it really messes with the visualization aspect as it gives the impression, as I mentioned in https://github.com/open-telemetry/opentelemetry-specification/issues/1085#issuecomment-875041382, that the production of a new message is a child of the incoming span.

I'm ok with this change as it makes the visualization clearer. Maybe it is revisited when there is a concept of "follows from".

kenfinnigan:
Thanks @trask for doing this.

+1 from me

trask:
> "Span Proc1" isn't the bit that produces the new message, it occurs in frameworks after the processing of the message is complete

ah, I was assuming "Span Proc1" represents the entire processing of the message, e.g. a `@JmsListener`

kenfinnigan:
I think it still represents the entirety of the processing for a specific message, as in `@JmsListener`.

However, the creation of another message out after processing a message, at least in reactive messaging cases we have with Kafka and AMQP, is a separate step that runs once the message processing is complete.

In essence, we create a reactive stream of "receive the incoming message" -> "process message" -> "create new outgoing message", with each step a separate component in the reactive stream

trask:
@kenfinnigan just to clarify, it sounds like you agree with this change for non-reactive messaging flows? If so, maybe you could propose a separate example for reactive messaging flows?

kenfinnigan:
The proposed change is to the Kafka example, which would be applicable to reactive messaging flows.

I don't remember enough about JMS to say whether what I think fits for the Kafka example is applicable to JMS and a Broker situation, or if a separate one is needed.

trask:
oh, sorry, I think I confused things by mentioning `@JmsListener` above.

how would you model `@KafkaListener`?

kenfinnigan:
I'm not super familiar with `@KafkaListener`, but from an initial look it seems `@KafkaListener` on its own would be the same as the Kafka example up to "Span Proc1" as the listener on its own just processes a message.

If a method had `@KafkaListener` and `@SendTo` present, it would be exactly the same as the Kafka example in that the method processing the message will return another message, but it's the framework that actually sends the new outgoing message to Kafka, and not the method processing the incoming message.

I might be wrong, but that's what it appears would be the case with an initial look

trask:
I see, I was assuming the example was depicting the processing code sending the downstream message directly, as opposed to returning the message to be sent by the framework

kenfinnigan:
@trask you mean the example for Apache Kafka? If so, I think I was the one that originally added it and it was based on how messaging is usually handled with reactive streams in that the communication with Kafka is handled by the framework.

After reviewing some traces the other day I realized what I'd proposed as the convention was a bit off, and should be adjusted

trask:
> @trask you mean the example for Apache Kafka? If so, I think I was the one that originally added it and it was based on how messaging is usually handled with reactive streams in that the communication with Kafka is handled by the framework.

thx, makes sense, I'll close this PR then since that was not my interpretation

PR #1802, opened 2021-07-07
Add 1.5.0 Schema File

tigrannajaryan:
No schema changes since 1.4.0, so 1.5.0 section is empty in the file.

### DIFF
--- a/schemas/1.5.0
+++ b/schemas/1.5.0
@@ -1,0 +1,5 @@
+file_format: 1.0.0
+schema_url: https://opentelemetry.io/schemas/1.5.0
+versions:
+  1.5.0:
+  1.4.0:


schemas/1.5.0
```diff
@@ -0,0 +1,5 @@
+file_format: 1.0.0
+schema_url: https://opentelemetry.io/schemas/1.5.0
+versions:
+  1.5.0:
+  1.4.0:
```
arminru, line 5: Why does this list both versions?

tigrannajaryan: I think it is good to explicitly indicate there were no changes neither in 1.4.0 nor in 1.5.0.

arminru: Will this list always be added to, i.e., will 1.7.0 include all changes made since 1.4.0 (= the initial version)? I thought one would have to walk through every single schema file B+C+D to transform from a version A to the desired version D.

tigrannajaryan: >Will this list always be added to, i.e., will 1.7.0 include all changes made since 1.4.0 (= the initial version)? 

Changes are additive, so 1.7.0 _section_ will not list changes made in 1.4.0, but 1.7.0 _file_ will have sections for 1.4.0, 1.5.0, 1.6.0 and 1.7.0.

> I thought one would have to walk through every single schema file B+C+D to transform from a version A to the desired version D.

No, the schema file is supposed to contain all transformations for previous versions since they are additive. One schema file is enough to transform between any two versions N and M, we do not require you to obtain all schema files between versions N and M.

arminru (approved):

carlosalberto (approved):

jsuereth (approved):

PR #1804, opened 2021-07-07
Add details for Aggregator in Metric spec

victlu:
Add into Metric spec details for defining a Metric Aggregator.


specification/metrics/sdk.md
```diff
@@ -81,6 +81,158 @@ active span](../trace/api.md#context-interaction)).
...
+aggregator/s.
+
+SDK MUST provide aggregators to support the default configured aggregator
+per instrument kind. e.g. Counter instruments default to a "Sum" aggregator
+configured for monotonic values.
+
+### Last Value Aggregator
+
+The Last Value Aggregator collect data for the [Gauge Metric Point](./datamodel.md#gauge)
+and is default aggregator for the following instruments.
```
reyang, line 141:
```suggestion
and is default aggregator for the following instruments:
```

specification/metrics/sdk.md
```diff
@@ -81,6 +81,158 @@ active span](../trace/api.md#context-interaction)).
...
+
+* [Asynchronous Gauge](./api.md#asynchronous-gauge) instrument.
+
+Last Value Aggregator maintains the following in memory:
+
+* Last Value (latest Measurement given, avoid any time comparison.)
```
reyang, line 147: I think we also need to maintain the "timestamp when the value was sampled".

specification/metrics/sdk.md
```diff
@@ -81,6 +81,158 @@ active span](../trace/api.md#context-interaction)).
...
+### Histogram Aggregator
+
+The Histogram Aggregator collect data for the [Histogram Metric Point](./datamodel.md#histogram)
+and is default aggregator for the following instruments.
+
+* [Histogram](./api.md#histogram) instrument.
+
+The Histogram Aggregator MUST be configurable to support different Temporality.
+
+Histogram Aggregator maintains the following in memory:
```
reyang, line 176: I guess we can copy from https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#histogram (the section under "A set of data points, each containing")?

reyang:
Do we intend to cover MIN and MAX?

reyang:
Some high level questions:
1. Is `Aggregator` intended to be exposed by the SDK as a type/class? (my answer would be NO, we don't need such type/class, we probably want some enum/flags that we can consume while defining the View).
2. If the answer to 1) is yes, do we intend to provide extensibility (e.g. one could build their own Aggregator)? (my answer would be eventually YES, for the first stable release probably NO)
3. If the answer to 1) is no, what's the purpose or it is just the SDK internal details?
4. Do we allow multiple aggregators to co-exist for the same event(s)? For example, if I have _temperature_, could I have Histogram, Min/Max and Average at the same time? If the answer is yes, do we end up with one stream or multiple streams? (my answer would be - we'll have multiple streams)

specification/metrics/sdk.md
```diff
@@ -81,6 +81,158 @@ active span](../trace/api.md#context-interaction)).
...
+
+An `Aggregator` MUST provide an interface to "aggregate" [Measurement](./api.md#measurement)
+data into its `In-Memory State`.
+
+An `Aggregator` MUST provide an interface to "collect" [Pre-Aggregated Metric](./datamodel.md#timeseries-model)
+data from its `In-Memory State`. The collect call should be treated as a
+stateful action and may reset in-memory state data. e.g. Resetting the time window
+for delta temporality instruments.
+
+An `Aggregator` MUST have read/write access to memory storage (`In-Memory
```
jsuereth, line 117: If an aggregator is *given* state be some other owner, is that ok?

E.g.  can an aggregator pass back some state in its callback that someone else stores and restores to it, or does the Aggregator HAVE to be stateful?

specification/metrics/sdk.md
```diff
@@ -81,6 +81,158 @@ active span](../trace/api.md#context-interaction)).
...
+State`) where it can store/retrieve/manage its own internal state.
+
+Note: SDK SHOULD provide configuration and control for memory availability to
+optimize usage. Aggregators MUST log errors when memory limits are reached
+and/or data lost occurs due to memory mitigation strategy. (e.g dropping high
+cardinality attributes).
```
jsuereth, line 123: This `MUST` could be a really hard reach depending on language.

E.g. in Java do you expect aggregators to throw try-catch around everything to catch OOM?

More importantly, this implies aggregators need to *pre-allocate* log messages to fire when OOM so they don't re-cause an OOM in GC languages.


Second - The implicit notion we can drop high cardinality attributes due to memory limitations.   I *think* there are global SDK limits on attributes length that can be placed.  This implies we may also do something like that for high-cardinality metrics.   Should that be called out somewhere?

specification/metrics/sdk.md
```diff
@@ -81,6 +81,158 @@ active span](../trace/api.md#context-interaction)).
...
+
+An `Aggregator` MUST have access to or given the [View](./sdk.md)
+configuration so it can be properly configure. e.g. Configure a Sum aggregator
```
jsuereth, line 126: I think the opposite should be true.   a `View` determines how an `Aggregator` is configured, but an `Aggregator` should be naive of `View`s.

specification/metrics/sdk.md
```diff
@@ -81,6 +81,158 @@ active span](../trace/api.md#context-interaction)).
...
+for Monoticity and Temporality.
+
+SDK MUST route measurements to the configured aggregator/s. e.g. SDK
```
jsuereth, line 129: Can you comment on the interaction between "MeasurementProcessor" and "Aggregator" here?

In my prototype I have "MeasurementProcessor" responsible for setting up the channel of `Measurements -> Aggregators`.

specification/metrics/sdk.md
```diff
@@ -81,6 +81,158 @@ active span](../trace/api.md#context-interaction)).
...
+
+* [Counter](./api.md#counter) instrument.
+* [Asynchronous Counter](./api.md#asynchronous-counter) instrument.
+* [UpDownCounter](./api.md#updowncounter) instrument.
+* [Asynchrounous UpDownCounter](./api.md#asynchronous-updowncounter) instrument.
+
+The Sum Aggregator MUST be configurable to support different Monoticity and/or
+Temporality.
+
+Sum Aggregator maintains the following in memory:
```
jsuereth, line 162: I've found that the memory footprint is actually highly different between Synchronus + Asynchronous instruments.

For the Java implementation, I fragmented Aggregator into a "stateless" component and I have separate storage mechanisms for  Synchronous + Asynchronous instruments that leverage an Aggregator for *what* to store.

The important point here, though, is I found the need to store a few different things.

## Synchronous Instruments.
- Delta-Accumulations per-exporter (see Reiley's fun multi-exporter design)
- Last-Reported Cumulative sums per-exporter (if CUMULATIVE sum)
- Last-Reported Timestamp per-exporter
- Application-start timestamp

## Asycnhronous Instruments
- Last collected CUMMULATIVE sum (for DELTA export)
- Last reported timestamp per-exporter

specification/metrics/sdk.md
```diff
@@ -81,6 +81,158 @@ active span](../trace/api.md#context-interaction)).
...
+The Histogram Aggregator collect data for the [Histogram Metric Point](./datamodel.md#histogram)
+and is default aggregator for the following instruments.
+
+* [Histogram](./api.md#histogram) instrument.
+
+The Histogram Aggregator MUST be configurable to support different Temporality.
+
+Histogram Aggregator maintains the following in memory:
+
+* Time window (e.g. start time, end time)
```
jsuereth, line 178: Where does `end time` come from (i.e. how do we have a  windo where)?

specification/metrics/sdk.md
```diff
@@ -81,6 +81,158 @@ active span](../trace/api.md#context-interaction)).
 +------------------+
 ```
 
+### Aggregator
+
+An `Aggregator` is a type of `MeasurementProcessor` and computes "aggregate"
+data from [Measurements](./api.md#measurement) and its `In-Memory State` into
+[Pre-Aggregated Metric](./datamodel.md#timeseries-model).
```
jkwatson, line 88:
```suggestion
[Pre-Aggregated Metric](./datamodel.md#timeseries-model)s.
```

specification/metrics/sdk.md
jkwatson, line 147:
```suggestion
* Last Value (latest Measurement given, avoiding any time comparison.)
```

jkwatson:
I was surprised not to see any mention of attributes/labels in here, and how an Aggregator needs to handle them. Is this something that is just so obvious that it doesn't need mentioning, or was this missed in writing up the details?

specification/metrics/sdk.md
```diff
@@ -81,6 +81,158 @@ active span](../trace/api.md#context-interaction)).
...
+                 +---------------------+
+```
+
+An `Aggregator` MUST provide an interface to "aggregate" [Measurement](./api.md#measurement)
+data into its `In-Memory State`.
+
+An `Aggregator` MUST provide an interface to "collect" [Pre-Aggregated Metric](./datamodel.md#timeseries-model)
+data from its `In-Memory State`. The collect call should be treated as a
+stateful action and may reset in-memory state data. e.g. Resetting the time window
+for delta temporality instruments.
```
ocelotl, line 115: [This](https://en.wikipedia.org/wiki/Aggregate_function#Decomposable_aggregate_functions) article mentions that there are 3 "kinds" of aggregate functions:

1. Some functions compute the aggregate by computing the aggregate for subsets, and then aggregating these aggregates, for example `COUNT`, `MAX`, `MIN`, and `SUM`.
2. Some functions compute the aggregate by computing auxiliary numbers for subsets, aggregating these auxiliary numbers, and finally computing the overall aggregate at the end, for example `AVERAGE` (aggregate with `SUM` and `COUNT`, then aggregate by dividing the former with the latter) and `RANGE` (aggregate with `MIN` and `MAX`, then aggregate by subtracting the former from the latter).
3. Some functions must compute the aggregate by computing the whole data set at once, for example `DISTINCT COUNT`, `MEDIAN`, and `MODE`.

Let's consider that aggregators use one function that can be of one of the kinds defined above. Aggregators that use a function of kind 1 can aggregate with one single value, aggregators that use a function of kind 2 can only aggregate with 2 or more values and aggregators that use a function of kind 3 can only aggregate using all the values.

Should the aggregator interface reflect this? Maybe we can have 3 kinds of aggregators each with different, matching interfaces?

specification/metrics/sdk.md
ocelotl: Not all aggregators have to be stateful. An aggregator that uses `MIN` does not need to be stateful, but an aggregator that uses `DISTINCT COUNT` does in the sense that the former can perform aggregation with every single value that is being aggregated and the latter needs to have all the values before aggregating.

specification/metrics/sdk.md
victlu: Yup, that's what we should have (unless I missed a field).

specification/metrics/sdk.md
victlu: I agree also, and that was what I intended. But, definitely not what I wrote. I reworded so PTAL.

victlu:
> Do we intend to cover MIN and MAX?

An `Aggregator` is intended to target a [Metric Point](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#metric-points) in Metrics Data Model. Thus, an Aggregator internally (`in-memory state`) stores a composite of data required to generate a Metric Point (Sums, Gauge, Histogram).

Currently, our OTLP protocol and Metric Points do not require MIN and/or MAX. ("Legacy" Summary Metric Point has that concept but wrapped as Quantiles).

specification/metrics/sdk.md
victlu: An `Aggregator` is intended to target a [Metric Point](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#metric-points) in Metrics Data Model. Thus, an Aggregator internally (`in-memory state`) stores a composite of data required to generate a Metric Point (Sums, Gauge, Histogram).

At the Metric Data Model level, each Metric Point (Sum, Gauge, Histogram) are intended to support decomposable aggregation functions. Merging and combining of Metric Time Series is primarily effort downstream from the API/SDK.

specification/metrics/sdk.md
```diff
@@ -81,6 +81,158 @@ active span](../trace/api.md#context-interaction)).
...
+* Sum (sum of Measurements per Monoticity and Temporality configuration.)
+
+### Histogram Aggregator
+
+The Histogram Aggregator collect data for the [Histogram Metric Point](./datamodel.md#histogram)
+and is default aggregator for the following instruments.
+
+* [Histogram](./api.md#histogram) instrument.
+
+The Histogram Aggregator MUST be configurable to support different Temporality.
```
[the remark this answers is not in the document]
victlu: > I was surprised not to see any mention of attributes/labels in here, and how an Aggregator needs to handle them.

Handling of Attributes is indeed a central concept. I did have a few mention and examples of how to address this. But, it likely needs it's own section to clarify. In the meantime, I did add a statement regarding SDK needing to route/map measurements to aggregators. It needs to be more flushed out in spec-level details.

In summary, A `MeasurementProcessor` ("Aggregator MeasurementProcessor") will route a measurement to zero or more pre-configured `Aggregators`. This Processor supports a mapping scheme which maps Attributes to Key/Value pair combinations. And for support of multiple exporters, mapping scheme which map per collection frequency.

At the last SIG meeting, I asked where this "routing" task should reside. As part of SDK (which I currently have spec) or part of a `MeasurmentProcessor` (call it the "Aggregator" measurement processor?). This is unfortunately confusing between a "Aggregator MeasurementProcessor" and an `Aggregator`.

Aggregator `MeasurmentProcessor` = concern with taking a measurement and breaking it up to zero or more instances.
`Aggregator` = concern with how to "aggregate" (sum/count/bucket/etc...) values.

victlu:
> I was surprised not to see any mention of attributes/labels in here, and how an Aggregator needs to handle them.

Handling of Attributes is indeed a central concept. I did have a few mention and examples of how to address this. But, it likely needs it's own section to clarify. In the meantime, I did add a statement regarding SDK needing to route/map measurements to aggregators. It needs to be more flushed out in spec-level details.

In summary, A `MeasurementProcessor` ("Aggregator MeasurementProcessor") will route a measurement to zero or more pre-configured `Aggregators`. This Processor supports a mapping scheme which maps Attributes to Key/Value pair combinations. And for support of multiple exporters, mapping scheme which map per collection frequency.

At the last SIG meeting, I asked where this "routing" task should reside. As part of SDK (which I currently have spec) or part of a `MeasurmentProcessor` (call it the "Aggregator" measurement processor?). This is unfortunately confusing between a "Aggregator MeasurementProcessor" and an `Aggregator`.

Aggregator `MeasurmentProcessor` = concern with taking a measurement and breaking it up to zero or more instances.
`Aggregator` = concern with how to "aggregate" (sum/count/bucket/etc...) values.

specification/metrics/sdk.md
victlu: Please see my comments below.  Also I've updated specs. PTAL

specification/metrics/sdk.md
victlu: I suspect most `Aggregators` will need some state. Even for `MIN`, we need to know the latest min to update min.

specification/metrics/sdk.md
```diff
@@ -81,6 +81,179 @@ active span](../trace/api.md#context-interaction)).
...
+stateful action and may reset in-memory state data. e.g. Resetting the time window
+for delta temporality instruments.
+
+An `Aggregator` MUST have read/write access to memory storage (`In-Memory
+State`) where it can store/retrieve/manage its own internal state.
+
+Note: SDK SHOULD provide configuration and control for memory availability to
+optimize usage. Aggregators MUST log errors when memory limits are reached
+and/or data lost occurs due to memory mitigation strategy. e.g dropping high
+cardinality attributes.
```
jkwatson, line 123: I don't think we should include this paragraph. It doesn't provide any direction on what/how it should be accomplished, and thus won't be helpful for SDK implementors who are reading this spec and working on implementations.

specification/metrics/sdk.md
victlu: > If an aggregator is given state be some other owner, is that ok?

The spec's callout of `In-Memory State` somewhat complicates this `Aggregator` spec. An `Aggregator` internally has in-memory states, and does not necessarily need to be called out (this naturally maps to a "class" and private member fields in most languages).

Given my description above, `In-Memory State` and `Aggregator` is the same object. Thus, an Aggregator can never be given someone else's state. And management and scope is controlled by different instances of Aggregator (e.g. Per instrument).

victlu:
> 1. Is `Aggregator` intended to be exposed by the SDK as a type/class? (my answer would be NO, we don't need such type/class, we probably want some enum/flags that we can consume while defining the View).

I think this is more of a style question. In the end, we need to expose the concept of an aggregation and potentially any configuration (e.g. Monotonic and/or Cumulative/Delta and/or Bucket scaling).

## Enum

```
  .AddView(..., ENUM_AGGREGATOR_SUM_MONOTONIC_DELTA);
```

pro:
- Only need to expose one ENUM type
- If we support file based configs (i.e. json/xml), ENUMs are easier to encode.

con:
- Need to enumerate all possible configurations upfront.
- Need to update ENUM if we extend in future.

## Class/Type

```
  .AddView(..., new SumAggregator(monotonic=true, temporality=DELTA);
```

pro:
- Allows for finer control (e.g. new Histogram(bucket_boundaries=new int[] { 1, 10, 100 } ;)
- Potential for easy extensions in future.

con:
- Need to expose several classes/types.

victlu:
> 4. Do we allow multiple aggregators to co-exist for the same event(s)? For example, if I have _temperature_, could I have Histogram, Min/Max and Average at the same time? If the answer is yes, do we end up with one stream or multiple streams? (my answer would be - we'll have multiple streams)

I also think it would be valuable to allow one measurement to generate multiple metrics. (i.e. a Temperature Instrument can produce a Gauge timeseries and a Histogram timeseries and a Summary timeseries.

An `Aggregator` as defined here, does not care, as each instance of an aggregator represent one timeseries. The SDK is poised to route a measurement to zero or more aggregators as configured by the View API.

specification/metrics/sdk.md
jsuereth: So in the Java SDK prototype, we actually have Aggregators that are stateless: https://github.com/jsuereth/opentelemetry-java/blob/wip-metrics-api/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/Aggregator.java

They defer all state to the following:

- For Async instruments it is expected that someone else stores metrics on behalf of the aggregator and just uses the aggregator to determine WHAT to store, not how.
- For Performance reasons, Aggregators  define a "SynchronousStorageHandle".   This is a highly concurrency friendly storage to directly aggregate synchronous measurements when-reported (vs. using a disrupter/queue).

victlu: In your prototype, who then actually manages the in-memory state? And what relationship does that component have with other components in terms of our SDK concepts?

specification/metrics/sdk.md
jsuereth: This is an important point though.  Should we tie *view* aggregators to Meric Data Model?  The Metric Data Model is meant to provide a *default* aggregation semantic.  It could be:

`MAX` -> Gauge
`MIN` -> Gauge
`COUNT` -> Sum
`SUM` -> Sum
`MEDIAN` -> Gauge
etc.


I.e. I'm not convinced we should ONLY support "LastValue", "Sum" and "HIstogram Distribution" as our only aggregations, but I think they're a good first step.

specification/metrics/sdk.md
jmacd: In earlier drafts of the SDK specification, and in my OTel-Go prototype, I tried to maintain another requirement for the measurement processor about it's ability to release memory. I was state that the measurement processor MUST be able to forget (i.e., "expunge") records that are not in use to avoid perpetual buildup of memory, especially when high-cardinality metrics are combined with delta aggregation temporality. 

I assumed that each Aggregator instance would have some fixed memory requirement, the question is not how much memory each takes, but how many of them there are. In other words, I would address the state question here by allowing the measurement processor to forget Aggregators.

specification/metrics/sdk.md
jmacd: :+1:

specification/metrics/sdk.md
```diff
@@ -81,6 +81,179 @@ active span](../trace/api.md#context-interaction)).
...
+                      |                                        +---------+-------+  |
+                      |                                                  |          |
+                      |                                              [Metrics]      |
+                      |                                                  |          |
+                      |                                        +---------V-------+  |
+                      |                                        | MetricExporter  |=====> OTLP Collector
+                      |                                        +-----------------+  |
+                      +-------------------------------------------------------------+
+```
+
```
jmacd, line 256: This may take more than a github comment thread to answer, but I'd like all of your opinions on the OTel-Go prototype, which looks about the same, but from reading the above it sounds different. The OTel-Go "Accumulator" (like "Measurement processor") computes only deltas and has a mechanism for forgetting Aggregators. The "Processor" (like "Metric processor") handles delta-to-cumulative conversion. 

This means that in a cumulative setup, there are really two copies of the data. One being maintained by the measurement processor (as a delta, if in use), and one being maintained by the metric processor (as a cumulative).

jmacd (approved):

specification/metrics/sdk.md
```diff
@@ -81,6 +81,179 @@ active span](../trace/api.md#context-interaction)).
...
+and is default aggregator for the following instruments:
+
+* [Histogram](./api.md#histogram) instrument.
+
+The Histogram Aggregator MUST be configurable to support different Temporality
+and bucket Boundary Values.
+
+Histogram Aggregator maintains the following in memory:
+
+* Start Time<sup>2</sup>
```
cijothomas, line 196: EndTime as well needed?

specification/metrics/sdk.md
```diff
@@ -81,6 +81,179 @@ active span](../trace/api.md#context-interaction)).
...
+
+\[1\]: Data from latest Measurement given, avoiding any time comparison.
+
+### Sum Aggregator
+
+The Sum Aggregator collect data for the [Sum Metric Point](./datamodel.md#sums)
+and is default aggregator for the following instruments:
+
+* [Counter](./api.md#counter) instrument.
+* [Asynchronous Counter](./api.md#asynchronous-counter) instrument.
```
cijothomas, line 172: it sounds incorrect if both sync and async versions of Counter defaults to the same aggregator, as sync version of counter instrument gives delta, whereas async version gives cumulative.

jmacd: On the other hand, the Aggregation being performed (i.e., addition, computing a sum of the inputs) is the same.

specification/metrics/sdk.md
jmacd: EndTime should be captured somehow during Collect, I think.

jmacd (approved):

specification/metrics/sdk.md
jsuereth: @victlu To answer your quesiton, the Measurement processor in my prototype is responsible for configuring the Storage and assocaited an Aggregator to it.  The Storage is its own component, configured based on Instrument + View configuration.  The aggregator just calculates bits and passes them to the storage.

specification/metrics/sdk.md
```diff
@@ -81,6 +81,179 @@ active span](../trace/api.md#context-interaction)).
...
+
+SDK MUST instantiate and configure aggregator/s based on [View](./sdk.md)
+configuration. e.g. Create a Sum aggregator with monotonic values and delta
+temporality (per mapping/routing).
+
+SDK MUST map/route measurements, based on [View](./sdk.md) configuration, to the
+configured aggregator/s. Each map/route has zero or more instances of
+aggregators.
+
+e.g. [View](./sdk.md) configuration specify which Attribute Keys to include,
```
jsuereth, line 133: The phrasing (specifically verb `route`) implies a "push" or active control flow.

What about asynchronous measurements and aggregation?   When does the pull happen, and how does the aggregator support this?


Side note, one reason we divorced aggregator from storage is this.  In the Java prototype async insturments can be polled on demand and aggregation is run.   for Synchronous instruments, aggregation happens as measurements come in.

specification/metrics/sdk.md
```diff
@@ -81,6 +81,179 @@ active span](../trace/api.md#context-interaction)).
...
+
+**Note:** An `Aggregator` instance is scoped to "aggregate" values that result
+in one metric timeseries (see [Metrics Data Model](./datamodel.md)). The SDK
+(via the [View](./sdk.md) configuration) allows Measurements to map/route into
+zero or more resulting metric timeseries (aka Aggregators). One map/route scheme
+involve expanding measurement `Attributes` to combinations of keys and values.
+The SDK will facilitate this map/route scheme. Thus, SDK will create and
+configure aggregators and route measurments to these aggregator instances.
+
+SDK MUST provide aggregators to support the default configured aggregator
```
jsuereth, line 148: Can you explicitly provide a table of these mappings rather than just a .e.g.?  I'm not sure we've captured this (appropriately) as a MUST specification anywhere else.

specification/metrics/sdk.md
jsuereth: @jmacd I actually went a different approach to this as well.

- Synchronous insturments 
  - have an impelemntation of "storage" that allows for forgetting "storage handles" when attributes disappear.
  - Can ask an Aggregator for a new high-concurrent handle for storing synchronous measurements.
  - Always report "DELTA".
  -  Will join with past value if the aggregator informs Storage to preserve last collected value.
- Asynchronous instruments
  - Only pull measuremnets -> Aggregator on demand.
  - Remember past value *IF* aggregator informs storage it's necessary (e.g. DELTA aggregation)
  
  
  From what I can tell, this is similar to what you did, but the "MetricsProcessor" actually receives "as close to OTLP" data as possible.
  
  I think it makes a ton of sense to allow (internally) Synchronous = Delta accumulation, Asynchronous = Cumulative accumulation and allow adapters.
  
Alll this said, from a *Specification* standpoint, I don't think any of that has to be specified.   The real question is, have we achieved the same semantics *and* are those semantics specified?  If users configure the same way and get the same data, the implementation shouldn't matter.


So my question:

- Does your metric processor expose the DELTAs as DELTAs?  Do we need to "hide" your Delta => Cumulative converter or do we want to specify that behavior or leave it language specific?
- Do we need to specify storage requirements in memory for Aggregators, or just what values they produce when configured?

specification/metrics/sdk.md
victlu: Given multiple likes, I'll remove this.

specification/metrics/sdk.md
victlu: I'm wondering if we leave the concept of In-Memory State to be implementation details. Thus, we only have concept of an Aggregator at the spec level.

specification/metrics/sdk.md
cijothomas: Could we just say - aggregator has access to the raw measurements?

specification/metrics/sdk.md
victlu: I took this out of spec.

specification/metrics/sdk.md
```diff
@@ -81,6 +81,158 @@ active span](../trace/api.md#context-interaction)).
...
+data from its `In-Memory State`. The collect call should be treated as a
+stateful action and may reset in-memory state data. e.g. Resetting the time window
+for delta temporality instruments.
+
+An `Aggregator` MUST have read/write access to memory storage (`In-Memory
+State`) where it can store/retrieve/manage its own internal state.
+
+Note: SDK SHOULD provide configuration and control for memory availability to
+optimize usage. Aggregators MUST log errors when memory limits are reached
+and/or data lost occurs due to memory mitigation strategy. (e.g dropping high
```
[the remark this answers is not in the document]
victlu: I took this out of spec.

specification/metrics/sdk.md
```diff
@@ -81,6 +81,179 @@ active span](../trace/api.md#context-interaction)).
...
+and/or data lost occurs due to memory mitigation strategy. e.g dropping high
+cardinality attributes.
+
+SDK MUST instantiate and configure aggregator/s based on [View](./sdk.md)
+configuration. e.g. Create a Sum aggregator with monotonic values and delta
+temporality (per mapping/routing).
+
+SDK MUST map/route measurements, based on [View](./sdk.md) configuration, to the
+configured aggregator/s. Each map/route has zero or more instances of
+aggregators.
```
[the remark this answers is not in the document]
victlu: I reworded and simplified. PTAL

specification/metrics/sdk.md
victlu: I reworded and simplified. PTAL

specification/metrics/sdk.md
victlu: I reworded and simplified. PTAL

specification/metrics/sdk.md
victlu: I like to keep the implementation details out of the spec if possible. I reworded several sections to allow for more abstract wording that hopefully covers the implementation details. PTAL

specification/metrics/sdk.md
victlu: I think we need to discuss the "scope" of an Aggregator. Is it limited to one function, or is it a proxy for a Metric Point?

Let's discuss in SIG meeting.

specification/metrics/sdk.md
victlu: I think these are details for each Aggregator implementation. (At least I'm currently not seeing us defining memory structure that is shared?)

I'm thinking we create an Aggregator (with parameters i.e. Temporality, Monotonicity) instance, and it can manage the appropriate in-memory states.

specification/metrics/sdk.md
victlu: I reworded the spec to avoid having to deal with implementation details. PTAL.

specification/metrics/sdk.md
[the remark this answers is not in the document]
victlu: I think you have a good point. I think each language implementation can tackle this.

specification/metrics/sdk.md
```diff
@@ -81,6 +81,126 @@ active span](../trace/api.md#context-interaction)).
...
+The Gauge Aggregator collect data for the [Gauge Metric Point](./datamodel.md#gauge).
+
+Gauge Aggregator maintains the following state:
+
+* Last Timestamp<sup>2</sup>
+* Last Value<sup>2</sup>
+
+\[2\]: Data from latest Measurement given, avoiding any time comparison.
+
+### HistogramAggregator
```
jsuereth, line 189: Again, I think we should focus on inputs (configuration or defaults) and outputs rather than internal state.

Here's what I'd have for Histogram Aggregator:

```
### ExplicitBucketHistogramAggregator
The ExplicitBucketHistogramAggregator collects measurements and produces the [Histogram Metric Point](./datamodel.md#histogram).

The ExplicitBucketHistogramAggregator MUST allow specification of histogram bucket boundaries.

The ExplicitBucketHistogramAggregator MUST compute a count per-bucket of how many measurements fall within the bucket.

The ExplicitBucketHistogramAggregator MUST compute count of all measurements seen.

The ExplicitBucketHistogramAggregator SHOULD compute a sum of all measurements seen.
- The sum MUST default to monotonic for `Histogram` or `Counter` instruments.
- The sum MUST default to nonmontonic for `UpDownCounter` instruments.
- The sum MUST be calculated with the natural `+` operation for the numeric type.
```

jsuereth (changes requested):

specification/metrics/sdk.md
victlu: I reworded. PTAL.

specification/metrics/sdk.md
```diff
@@ -81,6 +81,126 @@ active span](../trace/api.md#context-interaction)).
...
+internal state.
+
+The SDK MUST provide the following Aggregators to support the [Metric Points](./datamodel.md#metric-points)
+in the [Metrics Data Model](./datamodel.md).
+
+- [Sum Aggregator](#SumAggregator)
+- [Gauge Aggregator](#GaugeAggregator)
+- [Histogram Aggregator](#HistogramAggregator)
+
+### SumAggregator
```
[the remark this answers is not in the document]
victlu: I reworded. PTAL.

specification/metrics/sdk.md
```diff
@@ -81,6 +81,126 @@ active span](../trace/api.md#context-interaction)).
...
+| Instrument Kind | Default Aggregator | Monotonic | Temporality | Notes |
+| --- | --- | --- | --- | --- |
+| [Counter](./api.md#counter) | [Sum Aggregator](#SumAggregator) | Monotonic | Delta | |
+| [Asynchronous Counter](./api.md#asynchronous-counter) | [Sum Aggregator](#SumAggregator) | Monotonic | Cumulative | |
+| [UpDownCounter](./api.md#updowncounter) | [Sum Aggregator](#SumAggregator) | Non-Monotonic | Delta | |
+| [Asynchrounous UpDownCounter](./api.md#asynchronous-updowncounter) | [Sum Aggregator](#SumAggregator) | Non-Monotonic | Cumulative | |
+| [Asynchronous Gauge](./api.md#asynchronous-gauge) | [Gauge Aggregator](#GaugeAggregator) | n/a | Cumulative | |
+| [Histogram](./api.md#histogram) | [Histogram Aggregator](#HistogramAggregator) | n/a | Delta | TBD Boundaries|
+
+### Aggregator
```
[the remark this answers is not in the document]
victlu: I reworded. PTAL.

specification/metrics/sdk.md
```diff
@@ -81,6 +81,126 @@ active span](../trace/api.md#context-interaction)).
...
+and MUST be configurable to support Monotonicity and/or Temporality.
+
+Sum Aggregator maintains the following state:
+
+* Start Time<sup>1</sup>
+* Sum (sum of Measurements)
+
+\[1\]: Start Time is exclusive (aka not inclusive) of the provided time.
+
+### GaugeAggregator
```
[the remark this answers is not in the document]
victlu: I reworded. PTAL.

specification/metrics/sdk.md
```diff
@@ -81,6 +81,126 @@ active span](../trace/api.md#context-interaction)).
...
+                 |        |    ...       |   |
+                 |        |    "Collect"--------[Pre-Aggregated Metrics]-->
+                 |        +--------------+   |
+                 |                           |
+                 +---------------------------+
+```
+
+#### Default Instrument to Aggregator Mapping
+
+| Instrument Kind | Default Aggregator | Monotonic | Temporality | Notes |
```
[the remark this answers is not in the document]
victlu: I reworded. PTAL.

specification/metrics/sdk.md
victlu: I reworded. PTAL.

specification/metrics/sdk.md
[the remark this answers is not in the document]
victlu: Please reopen if this is still a concern.

specification/metrics/sdk.md
victlu: I reworded. PTAL.

specification/metrics/sdk.md
victlu: I reworded. PTAL.

specification/metrics/sdk.md
```diff
@@ -81,6 +81,148 @@ active span](../trace/api.md#context-interaction)).
...
+- [Sum Aggregator](#SumAggregator)
+- [Last Value Aggregator](#LastValueAggregator) (For Gauge Metric Point)
+- [Explicit Bucket Histogram Aggregator](#ExplicitBucketHistogramAggregator)
+
+#### SumAggregator
+
+The Sum Aggregator collect data for the
+[Sum Metric Point](./datamodel.md#sums)
+and reports the arithmetic sum of seen `Measurements`.
+This Aggregator may be configured for monoticity and temporality.
```
pirgeo, line 173:
```suggestion
This Aggregator may be configured for monotonicity and temporality.
```

specification/metrics/sdk.md
```diff
@@ -81,6 +81,148 @@ active span](../trace/api.md#context-interaction)).
...
+and reports the last seen `Measurement`.
+
+A [Default Aggregator](#DefaultAggregators) MUST be provided.
+
+#### ExplicitBucketHistogramAggregator
+
+The Explicit Bucket Histogram Aggregator collect data for the
+[Histogram Metric Point](./datamodel.md#histogram)
+and reports a histogram with explicit bucket boundaries of seen `Measurements`.
+This Aggregator may be configured for custom bucket boundaries, monoticity, and
```
pirgeo, line 190:
```suggestion
This Aggregator may be configured for custom bucket boundaries, monotonicity, and
```

specification/metrics/sdk.md
```diff
@@ -248,6 +248,151 @@ active span](../trace/api.md#context-interaction)).
...
+| Instrument Kind | Default Aggregator | Monotonic | Temporality | Notes |
+| --- | --- | --- | --- | --- |
+| [Counter](./api.md#counter) | [Sum Aggregator](#SumAggregator) | Monotonic | Delta | |
+| [Asynchronous Counter](./api.md#asynchronous-counter) | [Sum Aggregator](#SumAggregator) | Monotonic | Cumulative | |
+| [UpDownCounter](./api.md#updowncounter) | [Sum Aggregator](#SumAggregator) | Non-Monotonic | Delta | |
+| [Asynchrounous UpDownCounter](./api.md#asynchronous-updowncounter) | [Sum Aggregator](#SumAggregator) | Non-Monotonic | Cumulative | |
+| [Asynchronous Gauge](./api.md#asynchronous-gauge) | [Last Value Aggregator](#LastValueAggregator) | Non-Monotonic | Cumulative | |
+| [Histogram](./api.md#histogram) | [Explicit Bucket Histogram Aggregator](#ExplicitBucketHistogramAggregator) | Monotonic | Delta | Default Bucket Boundaries<sup>1</sup> |
+
+\[1\]: Default Bucket Boundaries: (-&infin;, 0], (0, 5.0], (5.0, 10.0], (10.0, 25.0], (25.0, 50.0], (50.0, 75.0], (75.0, 100.0], (100.0, 250.0], (250.0, 500.0], (500.0, 1000.0], (1000.0, +&infin;)
```
weyert, line 394: Do we want to specify a default unit of measure for this?

victlu:
After discussion in SIG meeting, a new PR is created #1842 to continue with this spec.

PR #1807, opened 2021-07-08
Update trace example for Kafka messaging

kenfinnigan:
Minor update to the Kafka messaging example for tracing, as after seeing a real-life example of this situation it didn't look right. I now realize it makes more sense for the second producer to be a child of the receiver, as opposed to no parent and a link to the first producer.

### DIFF
--- a/specification/trace/semantic_conventions/messaging.md
+++ b/specification/trace/semantic_conventions/messaging.md
@@ -263,8 +263,8 @@ Process CB:                           | Span Rcv2 |
 | Field or Attribute | Span Prod1 | Span Rcv1 | Span Proc1 | Span Prod2 | Span Rcv2
 |-|-|-|-|-|-|
 | Span name | `"T1 send"` | `"T1 receive"` | `"T1 process"` | `"T2 send"` | `"T2 receive`" |
-| Parent |  | Span Prod1 | Span Rcv1 |  | Span Prod2 |
-| Links |  |  | | Span Prod1 |  |
+| Parent |  | Span Prod1 | Span Rcv1 | Span Rcv1 | Span Prod2 |
+| Links |  |  | |  |  |
 | SpanKind | `PRODUCER` | `CONSUMER` | `CONSUMER` | `PRODUCER` | `CONSUMER` |
 | Status | `Ok` | `Ok` | `Ok` | `Ok` | `Ok` |
 | `peer.service` | `"myKafka"` |  |  | `"myKafka"` |  |
--- a/specification/trace/semantic_conventions/messaging.md
+++ b/specification/trace/semantic_conventions/messaging.md
@@ -253,7 +253,7 @@ One process, CA, receives the message and publishes a new message to a topic T2 
 Frameworks such as Quarkus and Spring Boot separate processing of a received message from producing subsequent messages out.
 For this reason, receiving (Span Rcv1) is the parent of both processing (Span Proc1) and producing a new message (Span Prod2).
 The span representing message receiving (Span Rcv1) should not set `messaging.operation` to `receive`,
-as the work is hidden away by frameworks such as Quarkus and Spring Boot.
+as it does not only receive the message but also converts the input message to something suitable for the processing operation to consume and creates the output message from the result of processing.
 
 ```
 Process P:  | Span Prod1 |
--- a/specification/trace/semantic_conventions/messaging.md
+++ b/specification/trace/semantic_conventions/messaging.md
@@ -245,10 +245,15 @@ Process CB:                 | Span CB1 |
 | `messaging.operation` |  | `"process"` | `"process"` |
 | `messaging.message_id` | `"a1"` | `"a1"`| `"a1"` |
 
-### Apache Kafka Example
+### Apache Kafka with Quarkus or Spring Boot Example
 
 Given is a process P, that publishes a message to a topic T1 on Apache Kafka.
 One process, CA, receives the message and publishes a new message to a topic T2 that is then received and processed by CB.
+
+Frameworks such as Quarkus and Spring Boot separate processing of a received message from producing subsequent messages out.
+For this reason, receiving (Span Rcv1) is the parent of both processing (Span Proc1) and producing a new message (Span Prod2).
+The span representing message receiving (Span Rcv1) should not set `messaging.operation` to `receive`,
+as the work is hidden away by frameworks such as Quarkus and Spring Boot.
 
 ```
 Process P:  | Span Prod1 |
--- a/specification/trace/semantic_conventions/messaging.md
+++ b/specification/trace/semantic_conventions/messaging.md
@@ -272,7 +272,7 @@ Process CB:                           | Span Rcv2 |
 | `messaging.system` | `"kafka"` | `"kafka"` | `"kafka"` | `"kafka"` | `"kafka"` |
 | `messaging.destination` | `"T1"` | `"T1"` | `"T1"` | `"T2"` | `"T2"` |
 | `messaging.destination_kind` | `"topic"` | `"topic"` | `"topic"` | `"topic"` | `"topic"` |
-| `messaging.operation` |  | `"receive"` | `"process"` |  | `"receive"` |
+| `messaging.operation` |  |  | `"process"` |  | `"receive"` |
 | `messaging.kafka.message_key` | `"myKey"` | `"myKey"` | `"myKey"` | `"anotherKey"` | `"anotherKey"` |
 | `messaging.kafka.consumer_group` |  | `"my-group"` | `"my-group"` |  | `"another-group"` |
 | `messaging.kafka.client_id` |  | `"5"` | `"5"` | `"5"` | `"8"` |
--- a/specification/trace/semantic_conventions/messaging.md
+++ b/specification/trace/semantic_conventions/messaging.md
@@ -21,7 +21,7 @@
     + [Apache Kafka](#apache-kafka)
 - [Examples](#examples)
   * [Topic with multiple consumers](#topic-with-multiple-consumers)
-  * [Apache Kafka Example](#apache-kafka-example)
+  * [Apache Kafka with Quarkus or Spring Boot Example](#apache-kafka-with-quarkus-or-spring-boot-example)
   * [Batch receiving](#batch-receiving)
   * [Batch processing](#batch-processing)
 


arminru:
Related to #1799.

specification/trace/semantic_conventions/messaging.md
```diff
@@ -263,8 +263,8 @@ Process CB:                           | Span Rcv2 |
 | Field or Attribute | Span Prod1 | Span Rcv1 | Span Proc1 | Span Prod2 | Span Rcv2
 |-|-|-|-|-|-|
 | Span name | `"T1 send"` | `"T1 receive"` | `"T1 process"` | `"T2 send"` | `"T2 receive`" |
-| Parent |  | Span Prod1 | Span Rcv1 |  | Span Prod2 |
-| Links |  |  | | Span Prod1 |  |
+| Parent |  | Span Prod1 | Span Rcv1 | Span Rcv1 | Span Prod2 |
```
Oberon00, line 266: #1799's suggestion of

```suggestion
| Parent |  | Span Prod1 | Span Rcv1 | Span Proc1| Span Prod2 |
```

seems to make more sense:

Your suggestion here is:
* Prod1
  * Rcv1
    * Proc1
    * Prod2
      * Rcv2

Assuming the second message is produced when processing the first, and not as part of just receiving the first message, #1799 seems intuitively correct and also a natural outcome as long as the processing span sets its Span as "current" as long as it is working:

* Prod1
  * Rcv1
    * Proc1
      * Prod2
    	* Rcv2

kenfinnigan: That's where the issue is, for most message processing with Quarkus, Spring Boot, likely others, when the library/framework handles the production of a new message, the method doing the processing has already been completed.

It's also possible for post method processing evaluation by the library/framework to determine the outcome is to negatively acknowledge the consumed message and not produce a new outgoing message.

From this perspective it's more natural to say producing a new message is a subsequent step after processing an incoming message, as opposed to a direct child of the processing.

Oberon00: > for most message processing with Quarkus, Spring Boot, likely others, when the library/framework handles the production of a new message, the method doing the processing has already been completed

Interesting, can you maybe give me a link to some example code? I would have expected the processing method simply does something like call `myMessagingSystemConnection.connectToQueue("foo").send(myMessage)`.

Oberon00: From what you said, maybe we need to expand the definition of message receiving to also optionally include processing by a framework. Or we should do the actual business logic processing in a nested processing span /or even an INTERNAL span)

kenfinnigan: For Quarkus there is https://quarkus.io/guides/kafka#the-price-converter as an example of a processing business method that has no control over creating a PRODUCER span for `@Outgoing`.

For Spring there is https://docs.spring.io/spring-kafka/reference/html/#reply-message

It's possible I've misinterpreted the current spec for message tracing as well. I understood the operation name "receive" to just be related to receiving the message from some other system, and "process" was the business logic that acted on the received message. Maybe I'm wrong in my understanding.

If actual business logic was inside an INTERNAL span, what would a "process" operation relate to?

Oberon00: >  what would a "process" operation relate to?

Dispatching the previously received message to the actual business method, parsing function parameters out of the message, converting the function return value to another message, putting this other message in a queue (in a child produce span).

To be honest, the messaging semantic conventions were not written with such high-level frameworks in mind, they were written with a more low-level API like JMS in mind.

In the linked Quarkus example, you could argue that in this case, the business logic is not really dealing with messages anymore, it gets a POJO as input and returns a POJO and you could call it from a unit test without setting up or mocking any message system.

In the case of such a framework I would go with "process" as operation for both the thing the framwork does (where receiving is just one part, or even done further outside; read the description of the receive operation in the semantic conventions).
I think for such high level frameworks we could want a new operation kind of maybe "dispatch".

So you would have:

* Prod1
  * Dispatch1
    * Process
    * Prod2
      * Dispatch2

I don't know if we can even have a receive separate from the whole dispatching process.

kenfinnigan: > > what would a "process" operation relate to?
> 
> Dispatching the previously received message to the actual business method, parsing function parameters out of the message, converting the function return value to another message, putting this other message in a queue (in a child produce span).
> 
> To be honest, the messaging semantic conventions were not written with such high-level frameworks in mind, they were written with a more low-level API like JMS in mind.

Is it a difference between high-level frameworks and JMS, or between JMS and reactive messaging and event-driven approaches?

> 
> In the linked Quarkus example, you could argue that in this case, the business logic is not really dealing with messages anymore, it gets a POJO as input and returns a POJO and you could call it from a unit test without setting up or mocking any message system.
> 

Apologies, I should've explained that though the Quarkus example I linked utilizes the actual payload of the message as parameter and return values, it could also have used `Message` types to interact with more than just the payload.

> In the case of such a framework I would go with "process" as operation for both the thing the framwork does (where receiving is just one part, or even done further outside; read the description of the receive operation in the semantic conventions).
> I think for such high level frameworks we could want a new operation kind of maybe "dispatch".
> 
> So you would have:
> 
> * Prod1
>   
>   * Dispatch1
>     
>     * Process
>     * Prod2
>       
>       * Dispatch2

If I understand what you're suggesting, something like Quarkus would do both "Prod1" and "Dispatch1", but something like JMS would have the application code doing "Prod1" and the framework putting the message into a queue creating "Dispatch1"?

If so, I wonder whether "produce" is the correct terminology for what appears to mean "create" instead. It would be concerning if the process of "producing a message to a topic" for Kafka created a span called "Dispatch", as it doesn't semantically align with Kafka's meaning, and other like-minded messaging systems as well.

> 
> I don't know if we can even have a receive separate from the whole dispatching process.

Can you elaborate on what you mean?

kenfinnigan:
Any additional questions/comments?

specification/trace/semantic_conventions/messaging.md
Oberon00: >> I don't know if we can even have a receive separate from the whole dispatching process.
>
> Can you elaborate on what you mean?

I mean that in some frameworks it may be difficult to instrument just receiving the span, while not also covering the call into the business code.

> If I understand what you're suggesting, something like Quarkus would do both "Prod1" and "Dispatch1", but something like JMS would have the application code doing "Prod1" and the framework putting the message into a queue creating "Dispatch1"?
>
> If so, I wonder whether "produce" is the correct terminology for what appears to mean "create" instead. 

I think in JMS you would not typically have any "dispatch" nodes, you would just have "process nodes" (the callbacks) and the "procuce nodes" would just cover API calls like "send" <https://docs.oracle.com/javaee/7/api/javax/jms/MessageProducer.html#send-javax.jms.Message->

kenfinnigan: I see what you mean now.

However, as this is updating a specific example for tracing with Apache Kafka, I don't see how things being different with JMS comes into play.

Maybe the conventions need a JMS specific example?

kenfinnigan:
Any feedback @arminru?

kenfinnigan:
Any comments or feedback?

specification/trace/semantic_conventions/messaging.md
Oberon00: > I don't see how things being different with JMS comes into play.

Weren't we discussing the difference between receive/consume and the hypothetical dispatch span?

So not directly related to this example, only insofar that I don't think the Kafka case can be modeled well at all with the current semantic conventions.

kenfinnigan: Yes, sorry. Thinking on it some more I'm not sure the possible dispatch span really adds anything, as the framework handling of the receive, executing a method with the message, and possibly producing a new outgoing message, is fairly "thin" and adding further spans in that slice can have large impacts on the performance of the reactive stream.

I still think the following is the best representation of messaging with Kafka:

- Prod1
  - Rcv1
    - Proc1
    - Prod2
      - Rcv2

However, if there's further work required to define a more appropriate span representation of messaging with Kafka, an alternative is the removal of the example completely until it's rectified.

My main goal is to fix the current example, as it's definitely wrong. Whether that's fixing it how I've described, or removing it for later re-evaluation I don't mind

Oberon00: >  and adding further spans in that slice can have large impacts on the performance 

I did not suggest to add additional spans but to _replace_ the receive span with a dispatch span.
Receiving should only be receiving, not the follow-up processing. A receive span as defined currently cannot even have the producer as parent because more or less by definition it starts before we can read anything from the message, as it tracks the time between asking the messaging system for a new message and getting a response. Anything that already has the message as input is a "process" span, not "receive" according to the current definitions.

Oberon00: Oh, actually I remembered that wrong, the spec says:

> The receive span is [to] be used to track the time used for receiving the message(s), whereas the process span(s) track the time for processing the message(s). Note that one or multiple Spans with messaging.operation = process may often be the children of a Span with messaging.operation = receive. The distinction between receiving and processing of messages is not always of particular interest or sometimes hidden away in a framework (see the Message consumption section above) and therefore the attribute can be left out. 

So actually, I think using "Recv1" as name is fine here, but it would maybe be better to leave out the messaging.operation entirely.

kenfinnigan: I think that's reasonable, let me update the example to not have messaging.operation set

kenfinnigan:
@Oberon00 With the example updated to remove `message.operation`, is it aligned with what you were thinking or needs further work?

I'm on PTO next week so wanted to see if this can be resolved before

Oberon00:
I think the example data looks fine, but you need to add some explanation summarizing the outcome of our discussion, for why the example (parent/child, messaging.operation) is the way it is now.
Also, the example should maybe be renamed to "Apache Kafka with Quarkus or Spring Boot Example"

kenfinnigan:
@Oberon00 Let me know how that sounds, and whether it needs some wordsmithing

Thanks for all the discussion and feedback

specification/trace/semantic_conventions/messaging.md
```diff
@@ -245,11 +245,16 @@ Process CB:                 | Span CB1 |
...
-### Apache Kafka Example
+### Apache Kafka with Quarkus or Spring Boot Example
 
 Given is a process P, that publishes a message to a topic T1 on Apache Kafka.
 One process, CA, receives the message and publishes a new message to a topic T2 that is then received and processed by CB.
 
+Frameworks such as Quarkus and Spring Boot separate processing of a received message from producing subsequent messages out.
+For this reason, receiving (Span Rcv1) is the parent of both processing (Span Proc1) and producing a new message (Span Prod2).
+The span representing message receiving (Span Rcv1) should not set `messaging.operation` to `receive`,
+as the work is hidden away by frameworks such as Quarkus and Spring Boot.
```
Oberon00, line 256: For me, the reasoning would be "as it does not only receive the message but also converts the input message to something suitable for the processing operation to consume and creates the output message from the result of processing."

Oberon00 (approved):

specification/trace/semantic_conventions/messaging.md
kenfinnigan: Made the update to this as it's clearer than what I had. Thanks

kenfinnigan:
Could someone else kindly review, and hopefully approve, this?

carlosalberto (approved):

PR #1810, opened 2021-07-09
Add `consumer_id` to identify the consumer

### LINKED ISSUE #1796: Unique consumer identifier span attribute for messaging

Currently, the semantic conventions for [messaging](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md) don't define a unique identifier for a consumer. There is a unique identifier specifically for Kafka, but a messaging-wide attribute would be preferable.

Without such an identifier, reprocessing the same message will result in identical span details, potentially all the way down to runtime and process details on the Resource.

What did you expect to see?

The Kafka messaging semantic conventions define `messaging.kafka.consumer_group` and `messaging.kafka.client_id` to uniquely identify a consumer of a message, but there is nothing at the `messaging.` level for it. This causes any messaging system other than Kafka to not have a unique way to identify the id of the consumer processing a message.

My proposal would be to add `messaging.consumer_id` which is only required when kind is `CONSUMER`. We can define it to be constructed from the existing Kafka attributes, for Kafka, in a format of "{messaging.kafka.consumer_group}" or "{messaging.kafka.consumer_group} - {messaging.kafka.client_id}" when the client id is also known.

For brokers such as RabbitMQ and Artemis, I believe it would represent just a Client ID of the consumer, as that is sufficiently unique.

I'm happy to work on a PR for this change.

kenfinnigan:
Resolves #1796

## Changes

Adds `consumer_id` as a label for messaging spans to, as best as possible, uniquely identify the consumer of a message.

This is especially beneficial when the same message is "replayed" to different consumers, ensuring the spans are differentiated between the usages.


semantic_conventions/trace/messaging.yaml
```diff
@@ -57,6 +57,14 @@ groups:
             The [conversation ID](#conversations) identifying the conversation to which the message belongs,
             represented as a string. Sometimes called "Correlation ID".
           examples: 'MyConversationId'
+        - id: consumer_id
```
iNikem, line 60: Should this be expanded to `client_id` to include producerId as well?

carlosalberto (approved):

semantic_conventions/trace/messaging.yaml
Oberon00: Maybe we should consider that we currently have the used queue/topic/... as `destination`/`destination_kind`. Which also is not named perfectly, but the other way round.
I think client_id would make sense here, even if it only applicable for consumers in Kafka. But no strong opinion.

Oberon00:
@kenfinnigan Please also add a Changelog entry for the new attribute.

semantic_conventions/trace/messaging.yaml
Oberon00: Of course, the description should be updated accordingly to also allow a "sender ID".

kenfinnigan:
@Oberon00 I will update the changelog, but I'm not sure what to do regarding the suggested change to `client_id` as there's already approval for the change "as is"?

semantic_conventions/trace/messaging.yaml
kenfinnigan: Does a "sender ID" provide any differentiation that isn't already present?

My understanding is a single message can only be sent from a single sender or producer, which enables the process detail attributes to be used for uniquely identifying where the message came from.

With consumption, the same message can be consumed by many different consumers, either in near real-time or significantly delayed. It's for this reason I'd only focused on the consumption side for a unique identifier as it's possible to use the same process/service to consume the message and they are identified differently in terms of Kafka, or other brokers.

Oberon00:
> as there's already approval for the change "as is"?

There is only one approval yet, it's no big deal if you still change it (if you agree with @iNikem's and my comments).

Just re-request the review from @carlosalberto afterwards.

semantic_conventions/trace/messaging.yaml
Oberon00: > which enables the process detail attributes to be used for uniquely identifying where the message came from.

That makes sense, but doesn't the same apply, conceptually, to the consumer side? Or is it more of a "consumption" ID than a "consumer" ID?

semantic_conventions/trace/messaging.yaml
```diff
@@ -57,6 +57,14 @@ groups:
...
+          type: string
+          brief: >
+            The identifier for the consumer receiving a message. For Kafka, set it to
+            {`messaging.kafka.consumer_group` - `messaging.kafka.client_id`}, if both are present, or only
```
Oberon00, line 64: How should this be read? Should I put a hyphen surrounded by spaced in between?

Maybe more clear if that is the case:

```suggestion
            `{messaging.kafka.consumer_group} -  {messaging.kafka.client_id}`, if both are present, or only
```

semantic_conventions/trace/messaging.yaml
kenfinnigan: I think there is a difference in that the consumer could be the same physical service, with Resource identifiers, but if the configuration was changed to read messages from the earliest available, such as with Kafka, the different connection to Kafka to read the messages has a different ID associated with it and processes the same messages again.

I'm not sure I'm explaining myself very well

semantic_conventions/trace/messaging.yaml
kenfinnigan: I was trying to think of a good way to represent it, I like this better, thanks

Oberon00 (approved):

semantic_conventions/trace/messaging.yaml
Oberon00: I *think* I understand what you mean, but again, the producer process could also renew its connection and maybe behave slightly different from then on.

But: I just noticed that this is for the Kafka-specific semantic conventions. If we don't want to apply this attribute generally, then I think it makes sense to use "consumer_id" if that is how the concept is called for Kafka. If similar enough concepts also exists for other messaging systems we might want to reconsider though.

kenfinnigan: A producer could, but it would not be the same message it's sending to a broker, it would be a different set of messages starting from the point where it changed.

It codifies the individual values that are already present for Kafka, but is also applicable to Artemis/RabbitMQ, and I imagine most other brokers, where there is a unique id for the client consuming the message

Oberon00: > It codifies the individual values that are already present for Kafka, but is also applicable to Artemis/RabbitMQ, and I imagine most other brokers, where there is a unique id for the client consuming the message

Then I suggest moving this to the `messaging.consumer` convention.

kenfinnigan: I don't know what you mean by that.

Do you mean rename the field to just `consumer`?

Oberon00: This is currently in the general messaging semantic conventions that apply to both consumers and producers. But I assume that consumer_id only applies to consumer spans, not producer spans. So I'd put in the `messaging.consumer` semantic convention group, a bit down below in the YAML file (https://github.com/open-telemetry/opentelemetry-specification/pull/1810/files#diff-6ea24809c25647f7e2f067dfc8fa080983ab936471e9eda83b26f190181bdeb4R104)

kenfinnigan: Thanks! I see what you mean now

kenfinnigan: Would the attribute change to `id`, or it stays what it currently is and becomes `messaging.consumer.consumer_id`?

kenfinnigan: I think I've figured it out, so will update and get feedback.

Thanks for all the pointers

Oberon00: The messaging.consumer semantic convention still only has "messaging." as prefix, so the final string to use as key in the semantic conventions will still be just "messaging.consumer_id".

kenfinnigan:
I think this is good, but would appreciate a re-review to confirm

semantic_conventions/trace/messaging.yaml
```diff
@@ -112,6 +112,14 @@ groups:
           [Operation names](#operation-names) section above.
           If the operation is "send", this attribute MUST NOT be set, since the
           operation can be inferred from the span kind in that case.
+      - id: consumer_id
+        type: string
+        brief: >
+          The identifier for the consumer receiving a message. For Kafka, set it to
+          `{messaging.kafka.consumer_group} - {messaging.kafka.client_id}`, if both are present, or only
```
Oberon00, line 119: The only thing I'm not sure about in this PR is this format. Isn't there some standard "fully qualified client ID" format for Kafka?

Oberon00 (approved):

semantic_conventions/trace/messaging.yaml
kenfinnigan: I don't believe so, as https://jaceklaskowski.gitbooks.io/apache-kafka/content/kafka-properties-client-id.html describes how it's an optional identifier passed to the broker, so it could be anything

Oberon00: I meant the combined format, where we now put " - " in-between. I wondered if there is some existing notation to combine consumer group and client_id into a single identifier string.

kenfinnigan: Not that I've found unfortunately

kenfinnigan: They must be unique within a Kafka Broker, but they're not required. It's possible for a consumer to only have the consumer group id set and not the client id.

kenfinnigan:
Any other comments on this, or is it good?

kenfinnigan:
Any feedback/comments? I'd like to have this merged soon if possible so I can implement it in our frameworks

kenfinnigan:
Not sure why these checks failed, it almost appears they didn't even run.

Is there a problem with CI checks at present?

Oberon00:
I re-ran the CI.

kenfinnigan:
Thanks @Oberon00, seems ok now

kenfinnigan:
Is there additional feedback on this?

kenfinnigan:
Any additional concerns with this, or can it be merged?

kenfinnigan:
@Oberon00 Is this ok to be merged or is there more needed?

kenfinnigan:
I'd like to see this completed soon, what are the next steps for this to happen?

Oberon00:
I can't merge it. There needs to be one more approval and then a maintainer (= TC member) needs to merge it.

kenfinnigan:
@Oberon00 Ok. What can I do to help that process along?

carlosalberto (approved):

kenfinnigan:
Thanks @carlosalberto and @Oberon00!

PR #1814, opened 2021-07-13
Lint YAML files (mainly semantic conventions).

Oberon00:
Run yamllint (https://github.com/adrienverge/yamllint) over all yaml files.


Oberon00:
It even generates annotations on the diff page:

image

.github/workflows/checks.yaml
```diff
@@ -19,6 +19,20 @@ jobs:
...
+    - name: check out code
+      uses: actions/checkout@v2
+
+    - uses: actions/setup-python@v2
+
+    - name: install yamllint
+      run: make install-yamllint
+
+    - name: run yamllint
+      run: yamllint . -f github
```
Oberon00, line 34: Not using `make` here because of the `-f github` special option.

.yamllint
```diff
@@ -0,0 +1,15 @@
+extends: default
+
+rules:
+  document-start: disable
+  octal-values: enable
+  truthy:
+    allowed-values: ['true', 'false', 'on']  # 'on' for GH action trigger
```
arminru, line 7: I don't quite understand. Is the top level `on:` property for the Github actions considered a boolean value?

arminru (approved):

.yamllint
Oberon00: I'm not sure what a proper YAML parser should do, but at least yamllint interprets it that way.

Oberon00: It seems to be correct: https://makandracards.com/makandra/24809-yaml-keys-like-yes-or-no-evaluate-to-true-and-false Probably GitHub special-cases this when interpreting the workflow definitions.

SergeyKanzhelev:
Should it be mentioned in CONTRIBUTING.md? Maybe tools that can be used to auto-format?

SergeyKanzhelev (approved):
Would be great to update contributing.md as well

bogdandrutu (approved):

Oberon00:
> Maybe tools that can be used to auto-format?

I was looking for such a tool but couldn't quickly find one, at least not one that worked on Windows where I tried.

Oberon00:
> Should it be mentioned in CONTRIBUTING.md? Maybe tools that can be used to auto-format?

Good point, it would also be helpful to have a Make target that executes all linters. But maybe in a follow-up PR?

Also we should consider making a few more checks required, at least the old semantic convention check. I think this must be done by an admin in the repo settings.

carlosalberto (approved):

bogdandrutu:
@arminru since you have magic power to upgrade this branch, please upgrade it then merge the PR 🗡️

arminru:
Merged and also added `yamllint` as a required check for main.

PR #1821, opened 2021-07-20
Add distro name and version to Resource.

### LINKED ISSUE #1786: Add a identifier for agent/distro
Currently the OTel SDKs will provide some information such as `opentelemetry.sdk.name` and `opentelemetry.language.version`, which is very helpful. Now, I'm wondering if we could add something (optional) to identify the actual so-called distros (Amazon, Lightstep, Splunk, NewRelic, etc), whether it's an agent (Java, DotNet or Python) or a custom Collector instance.

A pair of names that come to mind: `opentelemetry.agent` or `opentelemetry.distro.name`?

carlosalberto:
Fixes #1786 

Added as well an entry in the glossary for "OpenTelemetry Distro". One interesting thing @trask mentioned was deprecation of `telemetry.auto.version`.

### DIFF
--- a/semantic_conventions/resource/telemetry.yaml
+++ b/semantic_conventions/resource/telemetry.yaml
@@ -48,7 +48,7 @@ groups:
       - id: distro.name
         type: string
         brief: >
-          The name string of the distribution, if used.
+          The name of the distribution, if applicable.
         examples: ["acme"]
       - id: distro.version
         type: string
--- a/semantic_conventions/resource/telemetry.yaml
+++ b/semantic_conventions/resource/telemetry.yaml
@@ -45,3 +45,13 @@ groups:
         brief: >
           The version string of the auto instrumentation agent, if used.
         examples: ["1.2.3"]
+      - id: distro.name
+        type: string
+        brief: >
+          The name string of the distribution, if used.
+        examples: ["acme"]
+      - id: distro.version
+        type: string
+        brief: >
+          The version string of the distribution, if used.
+        examples: ["1.2.3"]
--- a/specification/glossary.md
+++ b/specification/glossary.md
@@ -27,6 +27,7 @@ Some other fundamental terms are documented in the [overview document](overview.
   * [Instrumented Library](#instrumented-library)
   * [Instrumentation Library](#instrumentation-library)
   * [Tracer Name / Meter Name](#tracer-name--meter-name)
+  * [OpenTelemetry Distro](#opentelemetry-distro)
 - [Logs](#logs)
   * [Log Record](#log-record)
   * [Log](#log)
@@ -152,6 +153,14 @@ Synonyms: *Instrumenting Library*.
 This refers to the `name` and (optional) `version` arguments specified when
 creating a new `Tracer` or `Meter` (see [Obtaining a Tracer](trace/api.md#tracerprovider)/[Obtaining a Meter](metrics/api.md#meterprovider)).
 The name/version pair identifies the [Instrumentation Library](#instrumentation-library).
+
+### OpenTelemetry Distro
+
+This refers to any pre-configured OpenTelemetry compliant client or library,
+commonly associated with a specific working group, such as a telemetry vendor.
+OpenTelemetry Distros exist in order to encapsulate setup and configuration,
+and they may include a specific set of instrumentation libraries, plugins,
+samplers and other additional components.
 
 ## Logs
 
--- a/specification/resource/semantic_conventions/README.md
+++ b/specification/resource/semantic_conventions/README.md
@@ -111,6 +111,8 @@ The identifier SHOULD be stable across different versions of an implementation.
 | `telemetry.sdk.language` | string | The language of the telemetry SDK. | `cpp` | No |
 | `telemetry.sdk.version` | string | The version string of the telemetry SDK. | `1.2.3` | No |
 | `telemetry.auto.version` | string | The version string of the auto instrumentation agent, if used. | `1.2.3` | No |
+| `telemetry.distro.name` | string | The name string of the distribution, if used. | `acme` | No |
+| `telemetry.distro.version` | string | The version string of the distribution, if used. | `1.2.3` | No |
 
 `telemetry.sdk.language` MUST be one of the following or, if none of the listed values apply, a custom value:
 


jsuereth (approved):

iNikem:
> Fixes #1786
> 
> Added as well an entry in the glossary for "OpenTelemetry Distro". One interesting thing @trask mentioned was deprecation of `telemetry.auto.version`.

Unless we want to call auto-instrumentation agent, such as https://github.com/open-telemetry/opentelemetry-java-instrumentation a "distro", I don't think we can deprecated `telemetry.auto.version`. Moreover, at least in Java, we totally expect that vendor distros will be based on Otel Java Instrumentation, thus both versions are of interest.

specification/glossary.md
```diff
@@ -153,6 +154,14 @@ This refers to the `name` and (optional) `version` arguments specified when
 creating a new `Tracer` or `Meter` (see [Obtaining a Tracer](trace/api.md#tracerprovider)/[Obtaining a Meter](metrics/api.md#meterprovider)).
 The name/version pair identifies the [Instrumentation Library](#instrumentation-library).
 
+### OpenTelemetry Distro
```
iNikem, line 157: Do we want to call it "distro" or "distribution"?

carlosalberto: Probably we can actually call them "Distribution" formally, and "distro" as both informal/resource term.

trask:
> > Fixes #1786
> > Added as well an entry in the glossary for "OpenTelemetry Distro". One interesting thing @trask mentioned was deprecation of `telemetry.auto.version`.
> 
> Unless we want to call auto-instrumentation agent, such as https://github.com/open-telemetry/opentelemetry-java-instrumentation a "distro", I don't think we can deprecated `telemetry.auto.version`. Moreover, at least in Java, we totally expect that vendor distros will be based on Otel Java Instrumentation, thus both versions are of interest.

If vendors build their own SDK based on OTel SDK, how should they populate `telemetry.sdk.version`?

bogdandrutu (changes requested):
Wait?!?!?

Why is the `telemetry.sdk` not good for this?

```
| `telemetry.sdk.name` | string | The name of the telemetry SDK as defined above. | `opentelemetry` | No |
| `telemetry.sdk.language` | string | The language of the telemetry SDK. | `cpp` | No |
| `telemetry.sdk.version` | string | The version string of the telemetry SDK. | `1.2.3` | No |
```

iNikem:
> Wait?!?!?
> 
> Why is the `telemetry.sdk` not good for this?
> 
> ```
> | `telemetry.sdk.name` | string | The name of the telemetry SDK as defined above. | `opentelemetry` | No |
> | `telemetry.sdk.language` | string | The language of the telemetry SDK. | `cpp` | No |
> | `telemetry.sdk.version` | string | The version string of the telemetry SDK. | `1.2.3` | No |
> ```

If we replace sdk name and version with vendor distribution name and version, then SDK version is not available "in band". Meaning that if some backend is actually interested in the SDK version used by that distribution, they cannot find that data in the telemetry itself. They have to fetch that data somehow separately and join.

Second, let's take an example of vendor distribution based on Otel auto-intrumentation agent, e.g. Java. We have vendor distro version V, which is based on otel-java-instrumentation version A, which used SDK version S: V->A->S. If we continue to use `telemetry.auto.version` to record the version A, then we will have all of a sudden that V replaces S in `telemetry.sdk.version` and A is still separate. Sounds like A is more important than V and S, which is not true, IMO. I mean, if we disregard my first objection above, then we should have just 1 version recorded.

For the record: I support this PR and recording of all 3 versions right inside telemetry.

jmacd (approved):
"Distro" is fine with me (I'm not the language police).  The word has been in common use for decades in the this industry.

specification/glossary.md
tedsuo: +1 for distro, we've been using that shorthand everywhere.

tedsuo (approved):
This is a helpful addition! LGTM.

carlosalberto:
From yesterday's meeting: @bogdandrutu is afraid we will end up adding too many items to `Resource`, and that we could potentially _override_ the `telemetry.sdk` fields instead, as a potential alternative (to which, for the record, I disagree with). This is because once a vendor knows about its own version (say, `acme,version=1.1`), the vendor will in turn know what version of the SDK (and language) is being used.

If somebody has additional details to add, please do so. Otherwise I will close this issue by EOW ;)

iNikem:
If we close this PR, how the original issue will be solved? The only other proposed alternative is to reuse `telemetry.sdk.*` and at least two people are against it (Carlos and I).

specification/glossary.md
Oberon00: Normally, we use full names in resource, e.g. `rt` is a common abbreviation for `runtime` but we still choose the latter.
I think there was even a general discussion about this on some issue/PR quite some time ago, but I can't find it and https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-naming.md does not mention this.

So I'm weakly in favor of `distribution` instead of the abbreviated `distro`.

Oberon00:
> @bogdandrutu is afraid we will end up adding too many items to Resource

I think we already have a very large number of resource conventions, so if you set all applicable ones, you will probably have a few kB of resource info.

The distro would not be set by default by the SDK, right? Only if there actually is a distro?

semantic_conventions/resource/telemetry.yaml
```diff
@@ -45,3 +45,13 @@ groups:
         brief: >
           The version string of the auto instrumentation agent, if used.
         examples: ["1.2.3"]
+      - id: distro.name
+        type: string
+        brief: >
+          The name string of the distribution, if used.
```
Oberon00, line 51: The "if used" is unclear.

We need to define here what constitutes a "distribution" and what not. E.g. is opentelemetry-java, as uploaded to Maven Central, with all the BOM files, etc., a distribution? Or is only the agent a distribution?

It seems you added a glossary entry, which is good. Please link to it.

And nitpicking some more: "used" is still the wrong word here, probably should use

```suggestion
          The name of the distribution, if applicable.
```

carlosalberto:
> Only if there actually is a distro?

Only if there's a distro, i.e. a vanilla agent will **not** set this.

> The only other proposed alternative is to reuse telemetry.sdk.* and at least two people are against it (Carlos and I).

Good point. Maybe keep this issue open? And then we can discuss in a way to solve this, one way or another.

cc @trask

Oberon00:
> Only if there's a distro, i.e. a vanilla agent will not set this.

That's very confusing to me. I would view the agent as a distro of the SDK+API.

semantic_conventions/resource/telemetry.yaml
Oberon00: Unresolving, there is more to this comment than just this suggestion (and the suggestion also applies to distro.version)

tsloughter:
What happened with this?

I could have swore I had run into `telemetry.distro.version` in the spec but when I searched the repo I only found this issue.

The SDK name/version can't be used because distros aren't the same thing. A distro can be just a wrapper that configures and sets up the SDK, and could technically even work with multiple SDKs.

PR #1824, opened 2021-07-22
fix: add net.transport to http semcov

Flarna:
## Changes

http semconv requires net.peer.name/net.peer.port. But port doesn't exist
for pipe/socket based transport.

Add requirement for net.transport and make net.peer.port required for IP
only.

### DIFF
--- a/semantic_conventions/trace/http.yaml
+++ b/semantic_conventions/trace/http.yaml
@@ -105,7 +105,7 @@ groups:
       - any_of:
           - [http.url]
           - [http.scheme, http.host, http.target]
-          - [http.scheme, net.peer.transport, net.peer.name, net.peer.port, http.target]
+          - [http.scheme, net.transport, net.peer.name, net.peer.port, http.target]
           - [http.scheme, net.peer.ip, net.peer.port, http.target]
 
   - id: http.server
--- a/semantic_conventions/trace/http.yaml
+++ b/semantic_conventions/trace/http.yaml
@@ -105,7 +105,7 @@ groups:
       - any_of:
           - [http.url]
           - [http.scheme, http.host, http.target]
-          - [http.scheme, net.peer.name, net.peer.port, http.target]
+          - [http.scheme, net.peer.transport, net.peer.name, net.peer.port (IP only), http.target]
           - [http.scheme, net.peer.ip, net.peer.port, http.target]
 
   - id: http.server
@@ -140,6 +140,6 @@ groups:
     constraints:
       - any_of:
           - [http.scheme, http.host, http.target]
-          - [http.scheme, http.server_name, net.host.port, http.target]
-          - [http.scheme, net.host.name, net.host.port, http.target]
+          - [http.scheme, http.server_name, net.host.port (ip only), http.target]
+          - [http.scheme, net.host.name, net.host.port (ip only), http.target]
           - [http.url]
--- a/semantic_conventions/trace/http.yaml
+++ b/semantic_conventions/trace/http.yaml
@@ -105,7 +105,8 @@ groups:
       - any_of:
           - [http.url]
           - [http.scheme, http.host, http.target]
-          - [http.scheme, net.transport, net.peer.name, net.peer.port, http.target]
+          - [http.scheme, net.peer.name, net.peer.port, http.target]
+          - [http.scheme, net.peer.name, net.transport, http.target]
           - [http.scheme, net.peer.ip, net.peer.port, http.target]
 
   - id: http.server
--- a/semantic_conventions/trace/http.yaml
+++ b/semantic_conventions/trace/http.yaml
@@ -105,7 +105,7 @@ groups:
       - any_of:
           - [http.url]
           - [http.scheme, http.host, http.target]
-          - [http.scheme, net.peer.transport, net.peer.name, net.peer.port (IP only), http.target]
+          - [http.scheme, net.peer.transport, net.peer.name, net.peer.port, http.target]
           - [http.scheme, net.peer.ip, net.peer.port, http.target]
 
   - id: http.server
@@ -140,6 +140,6 @@ groups:
     constraints:
       - any_of:
           - [http.scheme, http.host, http.target]
-          - [http.scheme, http.server_name, net.host.port (ip only), http.target]
-          - [http.scheme, net.host.name, net.host.port (ip only), http.target]
+          - [http.scheme, http.server_name, net.host.port, http.target]
+          - [http.scheme, net.host.name, net.host.port, http.target]
           - [http.url]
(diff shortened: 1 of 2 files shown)


Oberon00:
> http semconv requires net.peer.name/net.peer.port. But port doesn't exist
for pipe/socket based transport.

net.peer.name/net.peer.port is more or less only a fallback if http.url and http.host are unavailable. I'd say that in the case of transport via pipe you'd need to set one of these.

Flarna:
As far as I know there is no url representation available for socket based connections.
host header and pipe/socket name are two different concerns and there is no need that they are the same. You can have a bunch of servers running on the same host listening on different sockets which all use `localhost` with no port so all of them would look the same. But `net.peer.name` would show the difference.

Flarna:
Spec requires also to set both if they are different which is likely here: `In that case it is strongly recommended to set the `net.peer.name` attribute in addition to `http.host`.`

semantic_conventions/trace/http.yaml
```diff
@@ -105,7 +105,7 @@ groups:
       - any_of:
           - [http.url]
           - [http.scheme, http.host, http.target]
-          - [http.scheme, net.peer.name, net.peer.port, http.target]
+          - [http.scheme, net.transport, net.peer.name, net.peer.port, http.target]
```
Oberon00, line 108:
```suggestion
          - [http.scheme, net.peer.name, net.peer.port, http.target]
          - [http.scheme, net.peer.name, net.transport, http.target]
```

Otherwise we still require the port in any case.

specification/trace/semantic_conventions/http.md
```diff
@@ -111,6 +111,8 @@ from the `net.peer.name`
 used to look up the `net.peer.ip` that is actually connected to.
 In that case it is strongly recommended to set the `net.peer.name` attribute in addition to `http.host`.
 
+Note that `net.peer.port` is only applicable for ip transports and therefore only required if IP transport is used.
```
Oberon00, line 114:
```suggestion
If a transport not based on TCP is used, and no http.host is available, then (in absence of `http.host` and `http.url`) leaving out `net.peer.port` is allowed, but `net.transport` must be set instead.
```

Oberon00:
For HTTP/1.1 and higher, a Host header (or `:authority` field in HTTP 2) is always required. But for HTTP/1.0 requests, it is not and we should probably be able to use these.
In that case, no URL (at least not with `http` or `https` scheme) can be associated with the request which probably means most backends won't be able to analyze it properly.

specification/trace/semantic_conventions/http.md
```diff
@@ -194,6 +196,8 @@ For example, `http.server_name` has shown great value in practice, as bogus HTTP
 
 It is strongly recommended to set `http.server_name` to allow associating requests with some logical server entity.
 
+Note that `net.host.port` is only applicable for ip transports and therefore only required if IP transport is used.
```
iNikem, line 199:
```suggestion
Note that `net.host.port` is only applicable for IP transports and therefore only required if IP transport is used.
```

specification/trace/semantic_conventions/http.md
Flarna: I used a different wording as I don't want to make transport describing entries depend on http ones.

specification/trace/semantic_conventions/http.md
```diff
@@ -194,6 +197,8 @@ For example, `http.server_name` has shown great value in practice, as bogus HTTP
 
 It is strongly recommended to set `http.server_name` to allow associating requests with some logical server entity.
 
+Note that `net.host.port` is only applicable for IP transports and therefore only required if IP transport is used.
```
Oberon00, line 200: I would use the same wording as above

specification/trace/semantic_conventions/http.md
```diff
@@ -111,6 +112,8 @@ from the `net.peer.name`
 used to look up the `net.peer.ip` that is actually connected to.
 In that case it is strongly recommended to set the `net.peer.name` attribute in addition to `http.host`.
 
+If transport is not based on IP `net.peer.port` is not applicable. Instead `net.transport` must be set.
```
Oberon00, line 115:
```suggestion
If transport is not based on IP `net.peer.port` is not applicable. Instead `net.transport` must be set (if neither `http.url` nor `http.host` are set).
```

Flarna: As said I would avoid to mix http concerns with transport concerns. Please note that you need `net.transport` to correctly interpret `net.peer.name` if it is not guranteed to be IP.

specification/trace/semantic_conventions/http.md
Oberon00: But without that, it reads like transport is always required if a non-TCP transport is used. Which would be OK for me, but then we'd need to add a it as a `ref:` attribute to the common HTTP attributes with `required: conditional: "If different from ip_tcp"`.

specification/trace/semantic_conventions/http.md
Oberon00: This document is about the HTTP semantic conventions here though. The general net semantic conventions are already described to that effect in <https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#general-network-connection-attributes>

In any case, it needs to be clear if, for HTTP spans, net.transport is required only if http.host/http.url is not set or for any non-TCP transport. Currently this is unclear to me.

specification/trace/semantic_conventions/http.md
Flarna: Sorry I don't understand what you mean. `net.transport` is as requires as `net.peer.name` and it is mentioned in the yaml in the same way. So why is there a `ref:` needed for the one but not for the other?

specification/trace/semantic_conventions/http.md
Flarna: But this is unrelated to this PR. Even before this PR it was not mentioned anywhere that e.g. `net.peer.name` and `net.peer.port` is not required if http.host/http.url is present.

Except if listing them as a separate line is enough then this is also done now.

Flarna:
I removed the changes for server side as I'm currently only interested in client side and server side can be done in a separate PR.

I removed also the extra explanation sentence as it seems it caused more confusions then clarifications. Users can select a line and fill the attributes as needed. So it's additionally allowed now to set `net.transport`/`net.peer.name`.

Oberon00:
I would not split up server and client side into two PRs, as in the meantime we would have differences between client and server side without an logical explanation. Unless you expect there are some conceptual differences for the server side?

Oberon00:
And I think a short explanation is still needed, especially since IMHO we should forbid specifying just `[http.scheme, net.peer.name, net.transport, http.target]` without `net.peer.port` if `net.transport` is `ip_tcp`.

Flarna:
I doubt I'm able to write this down in an acceptable way in a reasonable timeframe.

specification/trace/semantic_conventions/http.md
Oberon00: The `ref` would be needed if we want to add the `required: conditional:` I suggested.

PR #1825, opened 2021-07-22
Move syntax.md to build-tools, improve semconv README.

Oberon00:
Follow-up to https://github.com/open-telemetry/build-tools/pull/53: We don't need the syntax.md in two different places, if we already have a README with a link, just point the link to build-tools.

Also add a helpful warning that those who want to read semantic conventions confusingly should not look into `/semantic_conventions/` but `/specification/<signal>/semantic_conventions/`.

No CHANGELOG entry, as this is only a editorial change.


yurishkuro:
Not sure I agree with the move of the syntax.md file. It's a documentation file about how to write specs, which applies to files that exist in the specs, so why should be in a different repo? How is it even related to "tools", which are an implementation detail of the spec build process.

Oberon00:
> How is it even related to "tools", which are an implementation detail of the spec build process?

The syntax.md file documents the input format of the tool that generates the markdown output of the spec. IMHO the whole semantic_conventions/ top-level directory is as much an implementation detail of the spec as the semantic conventions generator tool.

The reason for the move was described in open-telemetry/build-tools#53, and it's a very practical one: As soon as you change the input format of the tool, you need to update syntax.md. If you have both in the same repository, you can update them in one PR and the changes to syntax.md also serve as documentation to the changes in the implementing Python code.

Oberon00:
Since we have the (now required) semantic-conventions check as part of every PR, you can't even write anything in the YAML that the tool doesn't support, even if it would maybe make sense conceptually.

yurishkuro (approved):
fair enough

jsuereth (approved):

arminru (approved):

Oberon00:
Heads-up: I created a PR to update syntax.md in the build tools, since it was not up to date with some recent additions to the tool. I also documented features that were there from the beginning but are not currently used in the spec (with a warning): https://github.com/open-telemetry/build-tools/pull/56

carlosalberto (approved):

PR #1831, opened 2021-07-27
[semantic conventions] add alibaba cloud as a cloud provider

legendecas:
## Changes

Add Alibaba Cloud as one of the cloud providers. So that SDK implementations with the generated definition can start working on detector support, etc.

Alibaba Cloud has ~10% of the global market share, ranking third in the world and first in Asia-Pacific.

Related issues https://github.com/open-telemetry/opentelemetry-js/pull/2378


Oberon00:
Thank you for the PR!
I think you probably also want to update the faas.invoked_provider trace semantic convention.

semantic_conventions/resource/cloud.yaml
```diff
@@ -8,6 +8,9 @@ groups:
         type:
           allow_custom_values: true
           members:
+            - id: Aliyun
```
Oberon00, line 11: Just confirming: Is "Aliyun" the the name that users of Alibaba Cloud would be most familiar with? Because I have only heard "Alibaba Cloud" used for it.

legendecas: Thanks for picking this up. I confirmed with the naming guideline, 'Aliyun' is widely used in legacy conditions. 'Alibaba Cloud' should be used instead.

specification/resource/semantic_conventions/README.md
```diff
@@ -167,6 +167,7 @@ resources can only be defined for providers listed as a valid `cloud.provider` i
 [Cloud](./cloud.md) and below. Provider-specific attributes all reside in the `cloud_provider` directory.
 Valid cloud providers are:
 
+- [Alibaba Cloud](https://www.alibabacloud.com/) (`aliyun`)
```
arminru, line 170:
```suggestion
- [Alibaba Cloud](https://www.alibabacloud.com/) (`alibaba_cloud`)
```

arminru (approved):
Thanks! Please also add an entry in CHANGELOG.md for your addition.

specification/resource/semantic_conventions/README.md
Oberon00: (probably not for this PR, but I feel that we should just mention the `cloud.provider` enum here, to not list all cloud providers a third time)

Oberon00 (approved):

carlosalberto (approved):

legendecas:
Just pushed to trigger unresponsive EasyCLA to re-check.

PR #1832, opened 2021-07-27
Add jsuereth to tc review list.

jsuereth:
See https://github.com/open-telemetry/community/pull/781

Time for me to be responsible :)


.github/auto_assign.yml
```diff
@@ -14,6 +14,7 @@ assigneeGroups:
         - bogdandrutu
         - carlosalberto
         - jmacd
+        - jsuereth
```
arminru, line 17: There's also https://github.com/open-telemetry/opentelemetry-specification/blob/main/.github/auto_assign_issue.yml for even more responsibility 🙂 
Maybe rename `auto_assign.yml` -> `auto_assign_pr.yml` while at it to make it clearer.

jsuereth: done!

jsuereth: QQ: How do I update the config that looks for this file on github?

arminru: That's just the config file for the workflow, so you'll also have to change that one accordingly:
https://github.com/open-telemetry/opentelemetry-specification/blob/342255820f6affd9e21409a8567242b979efc8c0/.github/workflows/auto-assign-tc-members.yml#L13

arminru: While at this again (sorry for the scope creep 😅) you could also rename the workflow files accordingly (e.g. `auto-assign-issue-workflow.yml`and `auto-assign-pr-workflow.yml` or rename the config files to `*-config.yml`) to clear this up as well.

jsuereth: done!

arminru: I think both files (config + workflow) having the same name but just being in different folders makes it a bit hard to grasp but we're not touching these files too frequently, so it's probably fine.

yurishkuro (approved):

Issue #1835, opened 2021-07-28
The default network port is 4317 for both OTLP/gRPC and OTLP/HTTP


ThomsonTan:
**What are you trying to achieve?**

The default port for both [OTLP/gRPC](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlpgrpc-default-port) and [OTLP/HTTP](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlpgrpc-default-port) is 4317. This could be cause some inconvenience for service which host both endpoints (like [OTel Collector](https://github.com/open-telemetry/opentelemetry-collector)), as it then requires one of the endpoint to be customer configured.

**Additional context.**

Add any other context about the problem here. If you followed an existing documentation, please share the link to it.

iNikem:
Duplicate of https://github.com/open-telemetry/opentelemetry-specification/issues/1816#issuecomment-884253345 ?

Issue #1816, opened 2021-07-14
Collector and Specification http port do not match


dyladan:
The specification says the default OTLP/HTTP port is 4317 [[link](https://github.com/open-telemetry/opentelemetry-specification/blob/faae76d4c936143e5a163ea4ab45b007486cda4f/specification/protocol/otlp.md#otlphttp-default-port)]

The collector itself defaults to port 55681 [[code link](https://github.com/open-telemetry/opentelemetry-collector/blob/7e7ee8fb3a4b9535788ce484bfca1d27786433ab/receiver/otlpreceiver/factory.go#L34)] [[docs link](https://github.com/open-telemetry/opentelemetry-collector/blob/7e7ee8fb3a4b9535788ce484bfca1d27786433ab/receiver/otlpreceiver)]

dyladan:
Should I open this in the collector?

lonewolf3739:
Related issue on collector repo https://github.com/open-telemetry/opentelemetry-collector/issues/3617

carlosalberto:
So currently the Spec indeed mentions 4317 is the default port for **both** gRPC and http - however (as @jmacd mentioned in the last Collector SIG call, IIRC), we at Lightstep had a few problems running different protocols in the very same port. Wondering _how bad_ would it be to change that now?

E.g. @bogdandrutu suggested to go with 4318 for http, which I'd support.

anuraaga:
Doesn't seem like 4318 has been claimed by OTel yet

https://www.speedguide.net/port.php?port=4318

Unlike e.g. 4317

https://www.speedguide.net/port.php?port=4317

Not sure how that happens but probably would need it if using 4318.

Oberon00:
> Not sure how that happens but probably would need it if using 4318.

This is described in #1148

tigrannajaryan:
Some historical context: we wanted to make grpc and http use the same port and we had an open issue in the Collector to do so: https://github.com/open-telemetry/opentelemetry-collector/issues/1256
The conclusion appears to be that there are technical hurdles that make it unfeasible: https://github.com/open-telemetry/opentelemetry-collector/issues/1256#issuecomment-880048833

Because of that we need to keep grpc and http ports separate. I believe this means we need to change the spec to say that otlp/http uses port 4318 and also submit for port registration with IANA like we did previously with 4317.

carlosalberto:
[...]

PR #1839, opened 2021-07-30
Change OTLP/HTTP port from 4317 to 4318

tigrannajaryan:
Related to https://github.com/open-telemetry/opentelemetry-specification/issues/1816
Fixes https://github.com/open-telemetry/opentelemetry-specification/issues/1835

Some historical context: we wanted to make grpc and http use the same port and we
had an open issue in the Collector to do so:
https://github.com/open-telemetry/opentelemetry-collector/issues/1256

The conclusion appears to be that there are technical hurdles that make it unfeasible:
https://github.com/open-telemetry/opentelemetry-collector/issues/1256#issuecomment-880048833

Because of that we need to keep grpc and http ports separate. I believe this means
we need to change the spec to say that otlp/http uses port 4318. Once this PR is merged
we will also neeed to submit for port 4318 registration with IANA like we did previously
with 4317 (https://github.com/open-telemetry/opentelemetry-specification/issues/1148#issuecomment-725067404).


iNikem (approved):

carlosalberto (approved):

jsuereth (approved):

codeboten (approved):

reyang (approved):

dyladan:
The collector currently uses 55681 as the default port. Will the collector make the breaking change to use this port?

codeboten:
> The collector currently uses 55681 as the default port. Will the collector make the breaking change to use this port?

@dyladan you can see the change here: https://github.com/open-telemetry/opentelemetry-specification/issues/1816

The collector will support both ports for some time.

tigrannajaryan:
> The collector currently uses 55681 as the default port. Will the collector make the breaking change to use this port?

Here is the Collector PR that supports both ports for receiving: https://github.com/open-telemetry/opentelemetry-collector/pull/3743

Issue #1782, opened 2021-06-29
How do I know which resource attributes are important enough to identify a Prometheus Metrics Stream?


jsuereth:
**What are you trying to achieve?**

In OpenTelemetry, we consider `Resource` and its attributes part of a [metric stream's identity](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#opentelemetry-protocol-data-model).  When exporting, this means resource labels help define the timeseries for that metric.

Prometheus (and many metrics databases) begin to have issues with high-cardinality metric streams (Where there are many unique key-value pairs for a given metric of the same name).  For example, [this presentation](https://promcon.io/2019-munich/slides/containing-your-cardinality.pdf) from 2019 covers a lot of great advice for how to avoid high-cardinality, when it is acceptable and rules of thumb.

Specifically, I'm looking at the following advice:

> Does the label help someone woken up at 2am identify the problem?
> Is there a predictable number of values for the label?
> If the answer to the above questions is no, consider having the label on an event log instead of in Prometheus.

When I look at OpenTelemetry and exporting metrics to prometheus, I'd like to support the following scenarios:
- If a user leverages [OTEL_RESOURCE_ATTRIBUTES](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#general-sdk-configuration), can I ensure these are part of the label-set used in Prometheus metrics?  E.g. if a user is adding `service.name` or `my.custom.app.label` and expect it to apply to all telemetry for correlation, can I ensure exported metrics have this?
- When exporting to Prometheus and leveraging `Resource` detection for various environments, should I be providing *all* the added labels?  E.g. look at [k8s](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/k8s.md) vs. [process](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/process.md)
  - `k8s` appears to be a minimal set of labels for various k8s resources and would be highly useful as labels.
  - `process` includes command line arguments, and could be highly cardinal / less important in a monitoring database vs. an event database.

**Thoughts**
This leads me to believe there may be TWO use cases for resource labels within opentelemetry:

1. High fidelity query-able metadata
2. Highly-Identifying targeted labels for time-series databases with high-cardinality restrictions.

I see a few ways we can work towards solving this:

[...]

PR #1841, opened 2021-08-02
Add `k8s.pod.ip` Resource Attribute

owais:
`k8s.pod.ip` can be used to uniquely identify a pod at any given moment of time. For example, OpenTelemetry collector's k8sprocessor uses it to make automatic correlation happen.


Oberon00:
How specific to k8s or could it also be added to the general host semantic conventions at <https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/host.md> as just `host.ip`?

Is there only ever a single k8s host IP, or should it be an array? Can the IP change (not supported by resources)?

specification/resource/semantic_conventions/k8s.md
```diff
@@ -69,6 +69,7 @@ containers on your cluster.
 |---|---|---|---|---|
 | `k8s.pod.uid` | string | The UID of the Pod. | `275ecb36-5aa8-4c2a-9c47-d8bb681b9aff` | No |
 | `k8s.pod.name` | string | The name of the Pod. | `opentelemetry-pod-autoconf` | No |
+| `k8s.pod.ip` | string | The IP address assigned to the Pod. | `172.17.0.5` | No |
```
Oberon00, line 72: How specific to k8s or could it also be added to the general host semantic conventions at <https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/host.md> as just `host.ip`?

Is there only ever a single k8s host IP, or should it be an array? Can the IP change (not supported by resources)?

owais: Does a host (as described by the semantic conventions) map to a k8s pod (I think it does) or more to a k8s node/host? I think we can add it to the host section.

> Is there only ever a single k8s host IP, or should it be an array? Can the IP change (not supported by resources)? Does the pod only have an IPv4 or also v6?

Generally, the value is extracted by a service that receives telemetry data from a pod/host such as the Otel collector so in practice it is always a single IP. IP can change but in that case, I think we should treat it as a separate resource for overall simplicity. I don't think we should assume it can be any valid IP (4 and 6).

Oberon00: > Does a host (as described by the semantic conventions) map to a k8s pod (I think it does) or more to a k8s node/host? I think we can add it to the host section.

Good point, this is currently a bit unclear to me anyway. We may have an k8s pod that runs inside a k8s node which is a VM which runs on a physical host, all of which might have different "host" information. I assume we always want the "innermost" host info, but I don't think this is stated anywhere.

Oberon00: > Generally, the value is extracted by a service that receives telemetry data from a pod/host such as the Otel collector so in practice it is always a single IP. IP can change but in that case, I think we should treat it as a separate resource for overall simplicity.

Good idea. In that case I think you should add a note that this is is not something reported by the agent. Also, since who reports the IP matters (a lot, if there is something like a NAT in-between), I would suggest to not use `k8s.pod.ip`, but `telemetry.received_from_ip` (or maybe a new group `telemetry.received_from` would make sense if there are similar attributes that only make sense to add on the collector, in which case it would become `telemetry.received_from.ip`).

Oberon00: >  I don't think we should assume it can be any valid IP (4 and 6).

Did you mean "I **do** think"?

owais: :) I meant to say it can be either and we shouldn't make assumptions that it can only be IPv4 for example

jmacd: I take it that the `k8s.pod.ip` used here is also the same default that Prometheus would use in a k8s environment as the default `instance` label for the k8s resource.

> k8sprocessor uses it to make automatic correlation happen

I believe this is specific to k8s, but I also recognize a general pattern for resource discovery. With the right kind of schema definition, we could probably encode which resource attributes are uniquely identifying to support more general kinds of autocorrelation. There are at least two kinds of identifying attribute, those which are unique in time and those which are truly unique. Any unique-in-time resource attribute should ensure that the Single-writer requirement is met for reporting metrics streams, for example; this is an open issue: https://github.com/open-telemetry/opentelemetry-specification/issues/1782

jmacd: > Is there only ever a single k8s host IP, or should it be an array?

I believe we are interested in the primary IP, which is well defined and intended for use in this sort of autocorrelation scheme. Another attribute could be used to list all the assigned IP addresses, but I would not find that useful--I would prefer to known the entity's identifiers and look up the assigned IP addresses.

jmacd (approved):

specification/resource/semantic_conventions/k8s.md
Oberon00: > I believe we are interested in the primary IP, which is well defined

Then this needs to be stated, preferably with a link to the Kubernetes documentation.

owais: @Oberon00 AFAICT, k8s only assigns a single IP to a pod. In some cases a server might see the wrong source IP depending on how the traffic is router but in most cases with in-cluster communication, the correct (and only) source IP can be identified.



> Each Pod has a single IP address assigned from the Pod CIDR range of its node. This IP address is shared by all containers running within the Pod, and connects them to other Pods running in the cluster.

This is from [GKE](https://cloud.google.com/kubernetes-engine/docs/concepts/network-overview
) but I believe it applies to stock k8s distribution as well.



https://kubernetes.io/docs/tutorials/services/source-ip/

tigrannajaryan (approved):

specification/resource/semantic_conventions/k8s.md
Oberon00: If I understood your earlier comments correctly, you wanted to set this semantic attribute on the collector or backend to the sender IP. If I understood @jmacd correctly, this should be set to a "well defined" "primary" IP (by whom? How do you know which is the primary)?

This leaves too many questions open, especially considering that none of the discussion has yet made it back to the markdown 🙂 I like the idea of the attribute, but think it needs to be described more precisely before merging.

Oberon00 (changes requested):
I'm requesting changes here

specification/resource/semantic_conventions/k8s.md
owais: > you wanted to set this semantic attribute on the collector or backend to the sender IP.

That is one of the use cases, yes, but I think resource semantic conventions should not prescribe that to be the only way to set this attribute. SDKs could set this from within the pod using a resource detector, collector could retrieve the pod IP via k8s API, collector would use source IP from an incoming requests, APM backends could independently correlate Pod metadata with incoming telemetry data, etc. So I don't think the semantic conventions should tie this attribute to source IP detection by an agent/collector.

>  If I understood @jmacd correctly, this should be set to a "well defined" "primary" IP (by whom? How do you know which is the primary)?

If I understand correctly, a pod can have multiple IPs if it has multiple network interfaces attached. If the IP is detected by a collector then it'd see the IP assigned by the network interface used by the pod to communicate with the collector. I'm not sure if K8S defines what a "primary IP" is. May be it can be deduced by figuring out the default network interface but I think it'd make things too complex.

So, in order to simplify things, I think the attribute should represent **any** IP that was assigned to the pod at a specific time just like the Pod UID.

We already describe the conditions under which names and UIDs are unique. I expanded it to include IPs.

//cc @jmacd

Oberon00: "any IP"  sounds fine to me, but then shouldn't this be just host.ip? Or is there still anything k8s-specific?

owais: That could work but then why have `k8s.pod.name/uid` instead of just using `host.name/id`? I think there is some value in having a k8s specific attribute. At least its presence can help identify the environment for example.

Oberon00: The name and UID *is* something k8s-specific: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/. Maybe there might be other UIDs assigned to a host, but the POD UID is a particular one.

owais: I see what you mean but I'm still not sure about using `host.ip` to represent pod. It sounds a bit ambiguous to me as it could mean k8s node IP or underlying VM IP. For example, collector has a `hostmetricsreceiver` which is generally used to collect metrics about a k8s node instead of a pod. Using `host.*` attribute to represent pod metadata could be confusing. Perhaps we could use `host.ip` in combination with `host.type=k8s.pod` but then the same resource could possibly have metadata about the node or underlying compute instance (ec2) with host prefix.

Oberon00: >  It sounds a bit ambiguous to me as it could mean k8s node IP or underlying VM IP

I agree this is an issue, this applies to all the host attributes and should best be solved in a separate PR (e.g. by adding wording that it should always describe the innermost host). While k8s.pod.ip solves the issue, I don't think it is very elegant.

@jmacd @tigrannajaryan As approvers of the PR, what do you think?

tigrannajaryan:
@owais please address the feedback.

specification/resource/semantic_conventions/k8s.md
Oberon00: Ah, so there *is* a definition of "POD IP" vs "some IP address of the POD" and the POD can have at most two (primary?) IPs (one v4, one v6). That's very useful input, thanks @seh!

Then I concur with the suggestion to have "primary" in the name, even if we allow an array of up to two addresses.

jmacd: Thank you @seh. I support adding "primary" to the name as well.

owais: What do all you think about having two attributes? `k8s.pod.ipv4` and `k8s.pod.ipv6`? If I read that comment correctly, it seems pod can have only one of each ipv4 and ipv6. The reason I'm a bit reluctant on using primary is that one of the use cases for this is the k8s processor which just reads the source IP of the sender pod. Could the collector see data coming in from the "secondary" IP based on network configuration? If so, in that case we'll end up recording the non-primary IP as the primary one. All functionality would still work as far as k8s processor is concerned but the metadata we record could be misleading.

owais: I think this attribute can be useful outside of k8s processor like use cases as well but thinking if a generic `data_source_ip` attribute makes more sense to solve more specific use cases for all platforms.

Oberon00: > What do all you think about having two attributes? k8s.pod.ipv4 and k8s.pod.ipv6?

I was thinking about that as well, but since one of these is in the pod_ip field and the other is only in the pod_ips array in that data structure liked by @seh, I would say that in doubt order matters.

So I would prefer a single attribute `k8s.pod.primary_ips` of type string array. Maybe `assigned_ips` instead of primary IPs.



>  if a generic data_source_ip attribute makes more sense to solve more specific use cases for all platforms

I agree to use a separate attribute for that use case, see my previous comment here:
https://github.com/open-telemetry/opentelemetry-specification/pull/1841#discussion_r680830356

seh: To clarify, while a pod can have more than one IP address, at present it can have at most one IPv4 address and one IPv6 address. It cannot have two or more IPv4 addresses, and cannot have two or more IPv6 addresses. [The validation asserts](https://sourcegraph.com/github.com/kubernetes/kubernetes@master/-/blob/pkg/apis/core/validation/validation.go?L3386-L3403) that the pod either has one IP address or it's dual-stacked and has one of each kind.

Given that, perhaps we can have something like _k8s.pod.primary_ip_address_ and _k8s.pod.secondary_ip_address_, where the latter is only occasionally populated, but if it is, the former must be populated too.

The Kubernetes API allows more IP addresses by type (the "status.podIPs" field being an array), but constrains the use and interpretation by the comments. I take the comments as binding.

Also, though not documented yet, the downward API _does_ allow a pod author to [capture the "status.podIPs" field](https://sourcegraph.com/github.com/kubernetes/kubernetes@master/-/blob/pkg/apis/core/pods/helpers.go?L92).

seh: One more detail, speaking of the downward API: Since the values injected as either environment variables or file content are of type string, Kubernetes has to flatten the "status.podIPs" array in order to inject it.

It turns out that Kubernetes [first sorts the bare value of the "status.podIPs" array in order to place the IP address with the same family as the hosting node's IP address, then supplies the array as a comma-separated string](https://sourcegraph.com/github.com/kubernetes/kubernetes@dd2d12f6dc0e654c15d5db57a5f9f6ba61192726/-/blob/pkg/kubelet/kubelet_pods.go?L822-L842). It's interesting that the downward API does not trust the prescribed relationship between the "status.podIP" and "status.podIPs" fields; not only does it not assume that the former is the first element of the latter, but it also may flip the latter around, such that what the downward IP presents does not match the pod's status fields.

PR #1851, opened 2021-08-05
Describe falling back to string mangling for AWS SDK RPC specifiers.

anuraaga:
## Changes

Not all AWS SDKs provide a way to access the service and operation names of an AWS API. We will work internally to ensure at least newer ones do, the models themselves that define AWS APIs include the data in the form of [cloudFormationName](https://awslabs.github.io/smithy/1.0/spec/aws/aws-core.html#cloudformationname).

In the meantime, we should allow instrumentation to generate a reasonable based on what it's got.

Related issues https://github.com/aspecto-io/opentelemetry-ext-js/issues/160


semantic_conventions/trace/instrumentation/aws-sdk.yml
```diff
@@ -16,12 +16,19 @@ groups:
         examples:
           - aws-api
       - ref: rpc.service
-        brief: "The name of the service to which a request is made, as returned by the AWS SDK."
+        brief: >
+          The name of the service to which a request is made, as returned by the AWS SDK. If the SDK does not
```
Oberon00, line 20: Consider moving part of this to the `note`. Also I would like to have (in the `note`) a more fully described example that states the "input" client interface name in addition to the resulting value.

Same for `rpc.method`.

carlosalberto (approved):

semantic_conventions/trace/instrumentation/aws-sdk.yml
anuraaga: Thanks, it seems to have fixed a bad `note` that was there too

semantic_conventions/trace/instrumentation/aws-sdk.yml
```diff
@@ -16,10 +16,11 @@ groups:
...
       - ref: rpc.service
-        brief: >
-          The name of the service to which a request is made, as returned by the AWS SDK. If the SDK does not
-          provide a away to retrieve a name, the name of the SDK's client interface for a service SHOULD be
-          used, removing the suffix `Client` if present, resulting in a PascalCase name with no spaces.
+        brief: "The name of the service to which a request is made, as returned by the AWS SDK."
+        note: >
+          If the SDK does not provide a away to retrieve a name, the name of the SDK's client interface for a
+          service SHOULD be used, removing the suffix `Client` if present, resulting in a PascalCase name with
+          no spaces. For example, `DynamoDBClient` would be `DynamoDB`.
```
Oberon00, line 23:
```suggestion
          no spaces. For example, `DynamoDBClient` would become `DynamoDB`.
```

semantic_conventions/trace/instrumentation/aws-sdk.yml
```diff
@@ -29,6 +30,10 @@ groups:
           The name of the operation corresponding to the request, as returned by the AWS SDK. If the SDK does
           not provide a way to retrieve a name, the name of the command SHOULD be used, removing the suffix
           `Command` if present, resulting in a PascalCase name with no spaces.
+        note: >
+          If the SDK does not provide a way to retrieve a name, the name of the command SHOULD be used,
+          removing the suffix `Command` if present, resulting in a PascalCase name with no spaces. For example,
+          `GetItemCommand` would be `GetItem`.
```
Oberon00, line 36:
```suggestion
          `getItemCommand` would become `getItem`.
```

Is that correct? Or do we want case normalization too?

Issue #366, opened 2019-11-26
Mark local root spans in distributed parts of a trace.


gbbr:
I'm proposing to make the [`same_process_as_parent_span` property](https://github.com/census-instrumentation/opencensus-proto/blob/d4d9f63/src/opencensus/proto/trace/v1/trace.proto#L309) of the span prototype part of the specification, as this can turn up to be useful in many scenarios.

It doesn't have to have the same form or the same name, but as long as the information is there, that would be great.

Flarna:
I think such information should be available via resources. It may be hard to detect if the parent span is from same process or not. Consider a HTTP request to your own application or you use some message queue and send messages to yourself. In both cases it's hard to distinguish if the sender was in process or not as you parse the tag on the message (e.g. `traceparent` HTTP header) to get the parent span id.

jmacd:
@Flarna I agree. It would be appropriate if there were a semantic convention that uniquely identified the process, but the are only TODOs around this: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-resource-semantic-conventions.md

gbbr:
That's a good point to consider. I'm wondering if that is a case we should worry about. It certainly seems very edge and possibly indicative of a sub optimal architecture, but forgive me if my assumption is wrong. Nevertheless it is a use case to consider.

What I'm actually hoping to obtain here is a flag which indicates if the parent is in another part of a distributed trace, similar to what existed in Opencensus (see https://github.com/census-instrumentation/opencensus-specs/issues/155), be that the same process or not. This would apply in the described scenario too.

jmacd:
I believe in the current specification, the goal of `SpanKind` is to convey this information.  If you use `SpanKind=CLIENT` you are a remote child, if you use `SpanKind=SERVER` you are a remote parent. 

There are several active discussions around `SpanKind`, please see the current spec and #51, #371.

Oberon00:
Actually, I think I just found a real use-case for having this field.

Consider the case where a trace crosses systems monitored by different backends (not even necessarily different vendors, but backend instances with disconnected data storage). More concretely, a trace that goes A1 -> B -> A2. Thanks to W3C TraceContext, the traceparent and tracestate header get "properly" propagated through. But even though A1 and A2 export spans to the same backend, they will not be able to properly connect the span from A2 as a child of A1, since the parent span ID at A2 will be that of B, of which backend A has no knowledge.

[...]

PR #1852, opened 2021-08-05
Add a section for OTel specific values in TraceState.

carlosalberto:
Related to https://github.com/open-telemetry/oteps/pull/168 as we want to be able to specify OTel-specific values how in `tracestate` (in the case of the aforementioned OTEP, as we want to store probability and a random number, for sampling purposes).

This is a draft for now, in order to get OTEP 168 make progress.

cc @yurishkuro

### DIFF
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -257,6 +257,23 @@ Such changes then make sense only right before
 or [telemetry data exporting](sdk.md#span-exporter).
 In both cases, `Propagator`s and `SpanExporter`s may create a modified `TraceState` copy before serializing it to the wire.
 
+#### OTel specific values
+
+When setting values that are part of the OTel ecosystem, they MUST all be contained in a single entry using
+the `otel` key, and the value MUST be a semicolon separated list of key-value pairs such as:
+
+* `otel=p:8;r:64`
+* `otel=foo:bar;k1=13`
+
+New set values MUST be appended to existing `otel` entry in `TraceState`, unless an explicit override
+is expected to happen, in order to preserve existing values beloging to other concerns.
+For example, if a given concern K wants to set `k1:13`:
+
+* `otel=p:8;r:64` will become `otel=p:8;r:64;k1=13`.
+* `otel=p:8,r:64;tkey=7` will become `otel:p8;r:64;k1=13`.
+
+Observe that the keys and values used within the `otel` value MUST all comply with the `TraceState` rules.
+
 ## Span
 
 A `Span` represents a single operation within a trace. Spans can be nested to


specification/trace/api.md
```diff
@@ -257,6 +257,23 @@ Such changes then make sense only right before
 or [telemetry data exporting](sdk.md#span-exporter).
 In both cases, `Propagator`s and `SpanExporter`s may create a modified `TraceState` copy before serializing it to the wire.
 
+#### OTel specific values
+
+When setting values that are part of the OTel ecosystem, they MUST all be contained in a single entry using
+the `otel` key, and the value MUST be a semicolon separated list of key-value pairs such as:
+
+* `otel=p:8;r:64`
+* `otel=foo:bar;k1=13`
```
yurishkuro, line 266:
```suggestion
* `otel=foo:bar;k1:13`
```

specification/trace/api.md
```diff
@@ -257,6 +257,23 @@ Such changes then make sense only right before
...
+
+New set values MUST be appended to existing `otel` entry in `TraceState`, unless an explicit override
+is expected to happen, in order to preserve existing values beloging to other concerns.
+For example, if a given concern K wants to set `k1:13`:
+
+* `otel=p:8;r:64` will become `otel=p:8;r:64;k1=13`.
```
yurishkuro, line 272:
```suggestion
* `otel=p:8;r:64` will become `otel=p:8;r:64;k1:13`.
```

specification/trace/api.md
```diff
@@ -257,6 +257,23 @@ Such changes then make sense only right before
...
+* `otel=p:8,r:64;tkey=7` will become `otel:p8;r:64;k1=13`.
```
yurishkuro, line 273: why is `tkey` removed in this example?

carlosalberto: Fixed ;)

specification/trace/api.md
```diff
@@ -257,6 +257,23 @@ Such changes then make sense only right before
 or [telemetry data exporting](sdk.md#span-exporter).
 In both cases, `Propagator`s and `SpanExporter`s may create a modified `TraceState` copy before serializing it to the wire.
 
+#### OTel specific values
+
+When setting values that are part of the OTel ecosystem, they MUST all be contained in a single entry using
+the `otel` key, and the value MUST be a semicolon separated list of key-value pairs such as:
```
yurishkuro, line 263: we could also go with `ot` string, most vendors used 2-letter keys

specification/trace/api.md
```diff
@@ -257,6 +257,23 @@ Such changes then make sense only right before
...
+
+* `otel=p:8;r:64`
+* `otel=foo:bar;k1:13`
+
+New set values MUST be appended to existing `otel` entry in `TraceState`, unless an explicit override
```
yurishkuro, line 268: What is the intention here? The API that we should provide to the SDK internals over this key is that of a `map<string, string>`, so I would avoid putting ordering requirements which always complicate things.

Also L242 defines the API for accessing top-level keys of `tracestate`, we should also describe the API for key-values stored within the `otel` key.

specification/trace/api.md
```diff
@@ -257,6 +257,23 @@ Such changes then make sense only right before
 or [telemetry data exporting](sdk.md#span-exporter).
 In both cases, `Propagator`s and `SpanExporter`s may create a modified `TraceState` copy before serializing it to the wire.
 
+#### OTel specific values
```
yurishkuro, line 260: I think this would make more sense in the SDK doc, not the API. The end users of OTEL API are not supposed to have access to the values stored inside the `otel` key, only the SDK components should be able to manipulate it.

specification/trace/api.md
```diff
@@ -257,6 +257,23 @@ Such changes then make sense only right before
...
+
+When setting values that are part of the OTel ecosystem, they MUST all be contained in a single entry using
+the `otel` key, and the value MUST be a semicolon separated list of key-value pairs such as:
+
+* `otel=p:8;r:64`
```
yurishkuro, line 265: we should probably also be more prescriptive about the requirements for keys and values, i.e. keys are alpha-num

specification/trace/api.md
```diff
@@ -257,6 +257,23 @@ Such changes then make sense only right before
...
+* `otel=foo:bar;k1:13`
+
+New set values MUST be appended to existing `otel` entry in `TraceState`, unless an explicit override
+is expected to happen, in order to preserve existing values beloging to other concerns.
+For example, if a given concern K wants to set `k1:13`:
+
+* `otel=p:8;r:64` will become `otel=p:8;r:64;k1:13`.
+* `otel=p:8,r:64;k1:7` will become `otel:p8;r:64;k1:13`.
+
+Observe that the keys and values used within the `otel` value MUST all comply with the `TraceState` rules.
```
yurishkuro, line 275: We actually need more strict rules for values because `:` and `=` are not allowed characters in values. For performance reasons it is better if the API for setting values would not be required to applied something like url encoding (making it the SDK plugin's responsibility to  provide valid strings), but at the same time we may want protection against bugs that could allow an invalid string to break the rest of tracestate.

specification/trace/api.md
jmacd: I'm not sure. In a given SDK version, there will be a set of known fields from this `ot` value that can be parsed into dedicated fields, while any unknown values could be maintained as a pass-through. In my prototype, the data type is:
```
type otelTraceState struct {
	random      int
	probability int
	unknown     []string
}
```
where I've parsed the two fields I care about and left the rest in `unknown`.

specification/trace/api.md
jmacd: I would like to be as strict as possible, so that SDKs can hand-roll a parser for this and not use any kind of parsing framework or heavy-weight string manipulators. My prototype restricts the character set to not use punctuation at all, except the `:` and `;` separators. https://github.com/open-telemetry/opentelemetry-go/compare/main...jmacd:jmacd/propagate?expand=1#diff-3f1cead504691608d45bc4e9a09c8485f2b8da14e6b619391b42c737b52a8d75R363

yurishkuro: > I would like to be as strict as possible

By that do you mean SDK should force-encode non-conformant values or refuse to accept them and throw exception? I am talking about Set(k, v) function, not the parsing function.

jmacd: I was trying to say that because this type we are discussing is specific to OTel's purposes, we don't need support for arbitrary fields via Set(k, v), because for any field that we've specified we should be able to have a `SetField(specificValue)`.  For compatibility, I would force-encode unrecognized values under the assumption that some future-in-time OTel SDK is propagating new state.

specification/trace/sdk.md
```diff
@@ -680,5 +681,23 @@ public interface SpanExporter {
 }
 ```
 
+## OTel TraceState values
+
+When setting [TraceState](api#tracestate) values that are part of the OTel ecosystem,
+they MUST all be contained in a single entry using the `ot` key, with the value being
+a semicolon separated list of key-value pairs such as:
+
+* `ot=p:8;r:64`
```
Oberon00, line 690: I suppose like there will be a need for an `ot@TENANTID` alternative for cases where we cross systems that report to different backends/backend instances. Cf. https://github.com/open-telemetry/opentelemetry-specification/issues/366 (a entry for the tenant-parent-span-id in the `ot@TENANTID` key would also be nice)

specification/trace/sdk.md
```diff
@@ -680,5 +681,23 @@ public interface SpanExporter {
...
+* `ot=foo:bar;k1:13`
+
+Set values MUST be either updated or added to the `ot` entry in `TraceState`,
+in order to preserve existing values belonging to other OTel concerns. For example,
+if a given concern K wants to set `k1:13`:
+
+* `ot=p:8;r:64` will become `ot=p:8;r:64;k1:13`.
+* `ot=p:8,r:64;k1:7` will become `ot:p8;r:64;k1:13`.
+
+Keys and values MUST only contain alphanumeric characters.
```
Oberon00, line 700: Does this allow hyphen (negative numbers) or dots (decimal parts)?

carlosalberto: As per [TraceContext](https://www.w3.org/TR/trace-context/#value), yes, we could include them (and a few others). Have something in mind? I don't see any problem with it (happened to include only alphanumeric values _for now_, in order to simplify this PR).

specification/trace/sdk.md
carlosalberto: Yeah, that's taken for granted, I think - we can double clarify this if needed.

carlosalberto:
@yurishkuro Please review, as I've updated this PR based on your feedback ;)

carlosalberto:
@yurishkuro @Oberon00 Updated the document to use the same notation used by the TraceContext specification. Specific changes:

1. Mentioned the entire list MUST NOT exceed 256 characters, as the related TraceContext's value section specifies.

2. Keys are limited to **two** characters, mostly being alphanumeric (`p`, `rs` or `t1`, for example). I think this will give OTel concerns enough range of sub-keys. Otherwise, happy to change it.

3. Values are limited to a-z / A-Z / 0-9 / "-" / "_" / "." - I think it makes sense to start with a small range of allowed characters, but have otherwise no opinion (will cc a few people who may know more).

4. Did NOT put a limit on the sub values, as choosing anything feels rather random - should it be 10? Should it be 6? Should it be 4? For now, I only mentioned **again** that 256 is the maximum length for the **entire** list.

specification/trace/sdk.md
```diff
@@ -680,5 +681,66 @@ public interface SpanExporter {
...
+
+* `ot=p:8;r:64`
+* `ot=foo:bar;k1:13`
+
+The list can be formally defined as:
+
+```
+list        = list-member *( ";" list-member )
+list-member = key ":" value
+list-member = "" ; empty character
```
yurishkuro, line 702: why do we need this? I would rather be strict and precise in the format, there's no need to cater for readability by allowing whitespace etc.

specification/trace/sdk.md
```diff
@@ -680,5 +681,66 @@ public interface SpanExporter {
...
+```
+
+The complete list length MUST NOT exceed 256 characters, as defined by the
+[TraceState value section](https://www.w3.org/TR/trace-context/#value),
+and the used keys MUST be unique.
+
+### Key
+
+The key is an identifier that describes the OTel concern, consisting
```
yurishkuro, line 711:
```suggestion
The key is an identifier that describes an OTel concern, consisting
```

specification/trace/sdk.md
```diff
@@ -680,5 +681,66 @@ public interface SpanExporter {
...
+of either one or two characters. Simple examples are `p`, `ts`, or `s1`.
```
yurishkuro, line 712: I am not sure I see much value in 2-letter restriction. The keys are not arbitrary, I assume every key will have to be precisely defined in the spec (and as constants in the SDKs), so it's not like someone can just add a key with arbitrary length.

specification/trace/sdk.md
```diff
@@ -680,5 +681,66 @@ public interface SpanExporter {
...
+The value can be formally defined as:
+
+```
+value      = *(chr)
+chr        = ucalpha / lcalpha / DIGIT / "." / "_" / "-"
+ucalpha    = %x41-5A ; A-Z
+lcalpha    = %x61-7A ; a-z
+```
+
+### Setting values
```
yurishkuro, line 736: Q: do we even want to open a Set(k,v) capability to update the `ot` entry? I would consider instead requiring API-only access, like SetInclusionProbabilityLog2(p). This may be a question for a different PR though, as it does not affect the spec.

specification/trace/sdk.md
```diff
@@ -680,5 +681,66 @@ public interface SpanExporter {
...
+
+Set values MUST be either updated or added to the `ot` entry in `TraceState`,
+in order to preserve existing values belonging to other OTel concerns. For example,
+if a given concern K wants to set `k1:13`:
+
+* `ot=p:8;r:64` will become `ot=p:8;r:64;k1:13`.
+* `ot=p:8,r:64;k1:7` will become `ot:p8;r:64;k1:13`.
```
yurishkuro, line 743: Should we also explicitly define the behavior of adding a k:v that makes the total list longer than 256? Namely that such Set operation must be ignored by the SDK (perhaps return `false`). This is again spilling into the API definition for manipulating the `ot` entry, maybe from the format spec perspective it could be unspecified (just delaying the inevitable though).

carlosalberto:
Ping @yurishkuro

yurishkuro (approved):
lgtm, we can improve further later, this looks like a good foundation

specification/trace/sdk.md
jmacd: I agree, I don't think `Set(k,v)` makes sense for the OTel value. Users who want to set arbitrary keys should user their own vendor ID, right?

specification/trace/sdk.md
jmacd: If we take away `Set()` and limit the fields we specify, I think we can avoid this fate possibly forever provided that can prove that the maximum encoding length of all valid settings for the `ot` value is less than 256 bytes.

mwear (approved):
LGTM

carlosalberto:
@Oberon00 Moved it to its own document. Let me know ;)

specification/trace/tracestate-handling.md
```diff
@@ -0,0 +1,74 @@
...
+Simple examples are `p`, `ts`, or `s1`.
+
+The key can be formally defined as:
+
+```
+key        = lcalpha *(lcalpha / DIGIT )
+lcalpha    = %x61-7A ; a-z
+```
+
+Specific keys used by OTel concerns MUST be defined as part as the Specification too.
```
jkwatson, line 39:
```suggestion
Specific keys used by OTel concerns MUST be defined as part as the Specification.
```

jmacd (approved):

specification/trace/tracestate-handling.md
```diff
@@ -0,0 +1,73 @@
+# TraceState Handling
```
arminru, line 1: Please reference this new document from the SDK spec so readers will find it and add mention it in the changelog.

specification/trace/tracestate-handling.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+## Setting values
+
+Set values MUST be either updated or added to the `ot` entry in `TraceState`,
+in order to preserve existing values belonging to other OTel concerns. For example,
+if a given concern K wants to set `k1:13`:
+
+* `ot=p:8;r:64` will become `ot=p:8;r:64;k1:13`.
+* `ot=p:8,r:64;k1:7` will become `ot:p8;r:64;k1:13`.
```
arminru, line 63:
```suggestion
* `ot=p:8,r:64;k1:7` will become `ot=p8;r:64;k1:13`.
```

arminru: Is it necessary to preserve the order or just a coincidence?

arminru: Is it necessary to preserve the order or just a coincidence?

specification/trace/tracestate-handling.md
```diff
@@ -0,0 +1,73 @@
...
+The list can be formally defined as:
+
+```
+list        = list-member *( ";" list-member )
+list-member = key ":" value
+```
+
+The complete list length MUST NOT exceed 256 characters, as defined by the
+[TraceState value section](https://www.w3.org/TR/trace-context/#value),
+and the used keys MUST be unique.
```
dyladan, line 25: Should we mention anywhere that it is forbidden or at least strongly discouraged to use any key not defined in the spec here? 256 is a very small limit if clients/instrumentations start adding things.

specification/trace/tracestate-handling.md
dyladan: hopefully its just a coincidence. Order preservation comes free in some languages but not others and seems an unnecessary burden.

specification/trace/tracestate-handling.md
```diff
@@ -0,0 +1,73 @@
...
+
+If setting a value ends up making the entire `ot` entry exceed the 256 characters limit,
+SDKs are advised to abort the operation and signal the user about the error, e.g.
+
+```go
+traceState, err := SetTraceStateValue(traceState, value)
+if err != nil {
+  // Could not set the specified value, traceState was not updated.
+}
+```
```
dyladan, line 73: In all other places, we are strongly discouraged from throwing. In languages which don't have an error return value like go does, it may not be obvious that a function may throw. Maybe we should specify a success boolean be returned in order to make it very obvious?

jmacd:
Here is a question about whether OTel SDKs should be encourage to preserve their own `otel` key when applying resource limits to the tracestate value:

https://github.com/open-telemetry/oteps/pull/168/files#r704651896

I think that question should be addressed here, not there.

carlosalberto:
> I think that question should be addressed here, not there.

Agreed, let's do that here.

specification/trace/tracestate-handling.md
```diff
@@ -0,0 +1,73 @@
...
+a semicolon separated list of key-value pairs such as:
+
+* `ot=p:8;r:64`
+* `ot=foo:bar;k1:13`
+
+The list can be formally defined as:
+
+```
+list        = list-member *( ";" list-member )
+list-member = key ":" value
```
bogdandrutu, line 20: If it is a list, it means order matter, do we define the order?

specification/trace/tracestate-handling.md
```diff
@@ -0,0 +1,73 @@
...
+The value can be formally defined as:
+
+```
+value      = *(chr)
+chr        = ucalpha / lcalpha / DIGIT / "." / "_" / "-"
+ucalpha    = %x41-5A ; A-Z
+lcalpha    = %x61-7A ; a-z
+```
+
+## Setting values
```
bogdandrutu, line 56: Should we allow custom values? I think we should limit our entry to only what the library needs, users/vendors can define other custom entries in the tracestate.

specification/trace/tracestate-handling.md
```diff
@@ -0,0 +1,73 @@
...
+In alignment to the [TraceContext](https://www.w3.org/TR/trace-context/) specification, this section uses the
+Augmented Backus-Naur Form (ABNF) notation of [RFC5234](https://www.w3.org/TR/trace-context/#bib-rfc5234),
+including the DIGIT rule in that document.
+
+When setting [TraceState](api.md#tracestate) values that are part of the OTel ecosystem,
+they MUST all be contained in a single entry using the `ot` key, with the value being
+a semicolon separated list of key-value pairs such as:
+
+* `ot=p:8;r:64`
+* `ot=foo:bar;k1:13`
```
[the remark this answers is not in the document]
carlosalberto: We had agreed to use `ot` as it's shorter, although I'm fine using `otel` ;) (although I doubt we will ever get to use `ot` for OpenTracing, ever).

specification/trace/tracestate-handling.md
carlosalberto: No, actually the order doesn't matter.

specification/trace/tracestate-handling.md
carlosalberto: Yes, added a mention for this in the Key section.

carlosalberto:
Updated the PR to reflect the latest feedback:

* Mention that ONLY keys defined in the Specification should be used - instrumentation libraries and clients MUST use their own entry.
* Upon updates, no order needs to be preserved.

Question: Should we _really_ go back to `otel` as the entry key?

specification/trace/tracestate-handling.md
carlosalberto: Done ;)

specification/trace/tracestate-handling.md
carlosalberto: Referenced from the API doc, as there's no TraceState mention in the SDK one.

specification/trace/tracestate-handling.md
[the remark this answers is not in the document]
yurishkuro: Many vendors were planning to use 2-letter codes, I would keep `ot`.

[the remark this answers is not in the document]
yurishkuro: https://github.com/w3c/tracestate-ids-registry/blob/main/content/registry.md - we should add `ot` there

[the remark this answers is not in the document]
carlosalberto: @bogdandrutu Ping

carlosalberto:
Ping @bogdandrutu - I think we are ready to go, unless you think there's something we need to tune still.

specification/trace/tracestate-handling.md
```diff
@@ -0,0 +1,80 @@
+# TraceState Handling
+
+**Status**: [Experimental](../document-status.md)
+
+In alignment to the [TraceContext](https://www.w3.org/TR/trace-context/) specification, this section uses the
+Augmented Backus-Naur Form (ABNF) notation of [RFC5234](https://www.w3.org/TR/trace-context/#bib-rfc5234),
+including the DIGIT rule in that document.
+
+When setting [TraceState](api.md#tracestate) values that are part of the OTel ecosystem,
+they MUST all be contained in a single entry using the `ot` key, with the value being
```
Oberon00, line 10: @carlosalberto ping! The conversation at https://github.com/open-telemetry/opentelemetry-specification/pull/1852#discussion_r686916528 (on which GitHub doesn't let me comment as part of a review) about specifying a tenant ID as part of the key needs to be specified here IMHO.

Oberon00 (changes requested):

specification/trace/tracestate-handling.md
carlosalberto: Hey @Oberon00 Sorry for the delayed answer. Actually I'd like, if possible, to have a **single** `ot` entry for the time being at least, and extend it later as needed, unless you have something specific in mind? Let me know.

carlosalberto: Ping @Oberon00

carlosalberto:
Ping @Oberon00 @bogdandrutu - hopefully we can iterate on the remaining bits so this can be merged ;)

specification/trace/tracestate-handling.md
Oberon00: I'm sorry for forgetting about this thread PR!

> Actually I'd like, if possible, to have a single ot entry for the time being at least, and extend it later as needed, unless you have something specific in mind?

Do you have something specific in mind that is usable with a single entry? If you have different OTel instances participating in the trace that report to at least two different backends/tenants?

Oberon00: I mentioned my use case there: https://github.com/open-telemetry/opentelemetry-specification/pull/1852#discussion_r686916528

A more concrete instance of that use case: Suppose you have some cloud provider that helpfully always sets a traceparent header that contains the span ID of a span that is reported to the cloud provider's monitoring service. You cannot or do not want to disable this service. Still, you want to monitor the function (also) in your own backend (say a Jaeger instance). Now the cloud provider will break all your traces at the point where it inserts its own span.
To survive that, you could add a tracestate entry on the calling side that just duplicates your span ID (assuming the cloud provider is W3C-compliant and preserves the trace ID in the traceparent header): `abc123@ot=s:b7ad6b7169203331`.
As long as the cloud provider does not use OTel itself this would even work without a tenant ID, but clearly, that would just be relying on an implementation detail of the cloud provider (it not using OTel).

PR #1863, opened 2021-08-11
Consistently format enums, fix non-identifier IDs.

### LINKED ISSUE #1519: Consistently format semantic convention enums

Oberon00:
Fixes #1519 (again) and also fixes some `id`s containing spaces (`Alibaba Cloud` -> `alibaba_cloud`) or starting with numbers (`1xRTT` -> `csma2000_1xrtt`) which tripped up code generators (open-telemetry/build-tools#59).

## Changes

Changes enum IDs to be lowercase.

Also changes values and briefs of `net.host.connection.subtype` and `rpc.grpc.status_code` to match the IDs. http.flavor is left unchanged since it is already quite old (but http.flavor should probably be refactored anyway separately).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ release.
 
 ### Semantic Conventions
 
+- BREAKING: Change enum member IDs to lowercase without spaces, not starting with numbers.
+  Change values of `net.host.connection.subtype` to match.
+  ([#1863](https://github.com/open-telemetry/opentelemetry-specification/pull/1863))
+
 ### Compatibility
 
 ### OpenTelemetry Protocol
--- a/semantic_conventions/resource/cloud.yaml
+++ b/semantic_conventions/resource/cloud.yaml
@@ -8,16 +8,16 @@ groups:
         type:
           allow_custom_values: true
           members:
-            - id: Alibaba Cloud
+            - id: 'alibaba_cloud'
               value: 'alibaba_cloud'
               brief: 'Alibaba Cloud'
-            - id: AWS
+            - id: 'aws'
               value: 'aws'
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -22,6 +22,10 @@ release.
 
 ### Semantic Conventions
 
+- BREAKING: Change enum member IDs to lowercase without spaces, not starting with numbers.
```
Oberon00, line 25: This is a breaking change, but to an experimental part. The oldest changed enum value is from #1156, Oct 30, 2020.

carlosalberto (approved):

carlosalberto:
@Oberon00 Regarding the breaking portion, is it enough to let Maintainers know next Monday, or should we do something else?

Oberon00:
I think that would be enough, as actual data breakage is only in stuff that is new *and* recent. But probably I'm the wrong person to ask that question to. 😃

carlosalberto:
@tigrannajaryan maybe you have something to say?

carlosalberto:
@Oberon00 Please update the CHANGELOG so we can merge ;)

PR #1837, opened 2021-07-29
Add metrics exporters spec files

reyang:
## Changes

* Updated the library guideline to clarify exporter for logs/metrics/traces.
* Added skeleton files for metrics exporters.

PR #1864, opened 2021-08-12
Add metrics exporter requirements to the metrics SDK spec

reyang:
Follow up #1837.

## Changes

* Added the metrics exporter type
* Added more information to the In-memory exporter

### DIFF
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -296,6 +296,77 @@ example:
 * Exporter D is a pull exporter which reacts to another scraper over a named
   pipe.
 
+### Interface Definition
+
+`MetricExporter` must support the following functions:
+
+#### Export(batch)
+
+Exports a batch of `Metrics`. Protocol exporters that will implement this
+function are typically expected to serialize and transmit the data to the
+destination.
+
+`Export` will never be called concurrently for the same exporter instance.
+`Export` can be called again only after the current call returns.
+
+`Export` MUST NOT block indefinitely, there MUST be a reasonable upper limit
+after which the call must time out with an error result (Failure).
+
+Any retry logic that is required by the exporter is the responsibility of the
+exporter. The default SDK SHOULD NOT implement retry logic, as the required
+logic is likely to depend heavily on the specific protocol and backend the spans
+are being sent to.
+
+**Parameters:**
+
+`batch` - a batch of `Metrics`. The exact data type of the batch is language
+specific, typically it is some kind of list.
+
+Returns: `ExportResult`
+
+`ExportResult` is one of:
+
+* `Success` - The batch has been successfully exported. For protocol exporters
+this typically means that the data is sent over the wire and delivered to the
+destination server.
+* `Failure` - exporting failed. The batch must be dropped. For
+example, this can happen when the batch contains bad data and cannot be
+serialized.
+
+Note: this result may be returned via an async mechanism or a callback, if that
+is idiomatic for the language implementation.
+
+#### ForceFlush()
+
+This is a hint to ensure that the export of any `Metrics` the exporter has
+received prior to the call to `ForceFlush` SHOULD be completed as soon as
+possible, preferably before returning from this method.
+
+`ForceFlush` SHOULD provide a way to let the caller know whether it succeeded,
+failed or timed out.
+
+`ForceFlush` SHOULD only be called in cases where it is absolutely necessary,
+such as when using some FaaS providers that may suspend the process after an
+invocation, but before the exporter exports the completed spans.
+
+`ForceFlush` SHOULD complete or abort within some timeout. `ForceFlush` can be
+implemented as a blocking API or an asynchronous API which notifies the caller
+via a callback or an event. OpenTelemetry client authors can decide if they want
+to make the flush timeout configurable.
+
+#### Shutdown()
+
+Shuts down the exporter. Called when SDK is shut down. This is an opportunity
+for exporter to do any cleanup required.
+
+Shutdown should be called only once for each `MetricExporter` instance. After
+the call to `Shutdown` subsequent calls to `MetricExporter` are not allowed and
+should return a Failure result.
+
+`Shutdown` should not block indefinitely (e.g. if it attempts to flush the data
+and the destination is unavailable). OpenTelemetry client authors can decide if
+they want to make the shutdown timeout configurable.
+
 ### Push Metric Exporter
 
 Push Metric Exporter sends the data on its own schedule. Here are some examples:
@@ -308,3 +379,6 @@ Push Metric Exporter sends the data on its own schedule. Here are some examples:
 Pull Metric Exporter reacts to the metrics scrapers and reports the data
 passively. This pattern has been widely adopted by
 [Prometheus](https://prometheus.io/).
+
+Pull Metric Exporter SHOULD always return immediately with an indication of
+failure when `ForceFlush` is called.
--- a/specification/metrics/sdk_exporters/in-memory.md
+++ b/specification/metrics/sdk_exporters/in-memory.md
@@ -5,3 +5,9 @@
 Note: this specification is subject to major changes. To avoid thrusting
 language client maintainers, we don't recommend OpenTelemetry clients to start
 the implementation unless explicitly communicated.
+
+The in-memory exporter accumulates telemetry data in the local memory and allows
+to inspect it (useful for e.g. unit tests).
+
+The in-memory exporter MUST support both [pull](../sdk.md#pull-metric-exporter)
+mode and [push](../sdk.md#push-metric-exporter) mode.


specification/metrics/sdk.md
```diff
@@ -308,3 +382,6 @@ Push Metric Exporter sends the data on its own schedule. Here are some examples:
 Pull Metric Exporter reacts to the metrics scrapers and reports the data
 passively. This pattern has been widely adopted by
 [Prometheus](https://prometheus.io/).
+
+Pull Metric Exporter SHOULD always return immediately with an indication of
```
jkwatson, line 386: I do wonder if it even makes sense for the "Pull Metric Exporter" to be included in the same interface definition as the push version. In practice, they look completely different. `forceflush()` is just one instance of the difference.

specification/metrics/sdk.md
```diff
@@ -296,6 +296,80 @@ example:
...
+after which the call must time out with an error result (Failure).
+
+Any retry logic that is required by the exporter is the responsibility of the
+exporter. The default SDK SHOULD NOT implement retry logic, as the required
+logic is likely to depend heavily on the specific protocol and backend the spans
+are being sent to.
+
+Individual language clients can decide how to associate
+[Resource](../resource/sdk.md) with `Metrics`. Refer to the [tracing SDK
+specfication](../trace/sdk.md#additional-span-interfaces) for more information.
```
jmacd, line 322: This sounds contradictory with https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#resource-sdk. It already states

```

When used with distributed tracing, a resource can be associated with the TracerProvider when the TracerProvider is created. That association cannot be changed later. When associated with a TracerProvider, all Spans produced by any Tracer from the provider MUST be associated with this Resource.

Analogous to distributed tracing, when used with metrics, a resource can be associated with a MeterProvider. When associated with a MeterProvider, all metrics produced by any Meter from the provider will be associated with this Resource.
```

specification/metrics/sdk.md
jsuereth: I don't see an interface for Pull Metric Exporter specified here.   So, I'd actually like to see it defined if we're going to specify it.

E.g. How do I implement a Pull metric exporter?  What's the interface + methods?

specification/metrics/sdk_exporters/in-memory.md
```diff
@@ -5,3 +5,9 @@
 Note: this specification is subject to major changes. To avoid thrusting
 language client maintainers, we don't recommend OpenTelemetry clients to start
 the implementation unless explicitly communicated.
+
+The in-memory exporter accumulates telemetry data in the local memory and allows
+to inspect it (useful for e.g. unit tests).
+
+The in-memory exporter MUST support both [pull](../sdk.md#pull-metric-exporter)
```
jsuereth, line 12: It's not clear what this entails or what it looks like due to not having a clear interface for PULL expoorter.

specification/metrics/sdk.md
```diff
@@ -296,6 +296,80 @@ example:
 * Exporter D is a pull exporter which reacts to another scraper over a named
   pipe.
 
+### Interface Definition
+
+`MetricExporter` must support the following functions:
+
+#### Export(batch)
```
jsuereth, line 303: This is a push-exporter interface.   Is the intentiono that this interface would be used e.g. by prometheus?

If so, that means effectively we ONLY have push exporters and pull-based exporters need to "save the last point"

The alternative here would be to define a "MetricProducer" interace with a "collect" method that pull exporters are given and can call on demand.

specification/metrics/sdk.md
```diff
@@ -296,6 +296,80 @@ example:
...
+
+`Export` will never be called concurrently for the same exporter instance.
+`Export` can be called again only after the current call returns.
+
+`Export` MUST NOT block indefinitely, there MUST be a reasonable upper limit
+after which the call must time out with an error result (Failure).
+
+Any retry logic that is required by the exporter is the responsibility of the
+exporter. The default SDK SHOULD NOT implement retry logic, as the required
+logic is likely to depend heavily on the specific protocol and backend the spans
```
cijothomas, line 317:
```suggestion
logic is likely to depend heavily on the specific protocol and backend the metrics
```

specification/metrics/sdk.md
reyang: Based on our discussion during the SIG meeting, I've scoped down the PR to focus on Push Metric Exporter. I will cover the Pull Metric Exporter and "MetricReader" in my next PR.

@jkwatson I would wish to have the same interface for push and pull exporters, or at least keep these as close as possible. For example, we can imagine the in-memory exporter to support both push and pull. Having the same interface makes it easier to implement such exporter. I will explain my thinking in the next PR, so we can focus on push exporter in this PR.

specification/metrics/sdk.md
```diff
@@ -296,6 +296,80 @@ example:
...
+`MetricExporter` must support the following functions:
+
+#### Export(batch)
+
+Exports a batch of `Metrics`. Protocol exporters that will implement this
+function are typically expected to serialize and transmit the data to the
+destination.
+
+`Export` will never be called concurrently for the same exporter instance.
+`Export` can be called again only after the current call returns.
```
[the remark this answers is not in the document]
reyang: I agree with your assessment @jmacd.
This is to align with what we already have in the [tracing spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#exportbatch), we just borrowed the exact wording from there.

specification/metrics/sdk.md
```diff
@@ -296,6 +296,80 @@ example:
...
+
+`Export` MUST NOT block indefinitely, there MUST be a reasonable upper limit
+after which the call must time out with an error result (Failure).
+
+Any retry logic that is required by the exporter is the responsibility of the
+exporter. The default SDK SHOULD NOT implement retry logic, as the required
+logic is likely to depend heavily on the specific protocol and backend the spans
+are being sent to.
```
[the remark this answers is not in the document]
reyang: I think each vendor might have different strategy and implementation. For example, I think it might make sense for an eventual consistent system to avoid retry for cumulative temporality, it might also make sense for a fully transactional system to always retry when there is a failure (and the protocol could support some unique ids for de-duplication).

specification/metrics/sdk.md
reyang: @jmacd would you clarify where's the contradiction? I think the resource sdk spec says "a resource CAN be ...", which is what we have here "Individual language clients CAN decide ..." (it sounds like a recommendation to me, rather than a mandatory requirement).

Or maybe we want to say "the resource CAN and CAN ONLY be associated with MeterProvider" and "all metrics produced by any Meter from the provider MUST be associated with ..."?

specification/metrics/sdk_exporters/in-memory.md
reyang: I've scoped down the PR based on our discussion.

specification/metrics/sdk.md
jkwatson: Yep. I'm looking forward to seeing it. :)

specification/metrics/sdk.md
jsuereth: This explicitly allows clients to determine how to associate `Resource` which is way more flexible than what the tracing specification says.  I'd force this to lock `Resource` to match that associated with the MeterProvider.

specification/metrics/sdk.md
```diff
@@ -303,6 +303,80 @@ Push Metric Exporter sends the data on its own schedule. Here are some examples:
...
+  happen when the batch contains bad data and cannot be serialized.
+
+Note: this result may be returned via an async mechanism or a callback, if that
+is idiomatic for the language implementation.
+
+##### ForceFlush()
+
+This is a hint to ensure that the export of any `Metrics` the exporter has
+received prior to the call to `ForceFlush` SHOULD be completed as soon as
+possible, preferably before returning from this method.
```
ocelotl, line 353: [Here](https://github.com/open-telemetry/opentelemetry-specification/pull/1864/files#diff-32d5f289beed96900fa1e6e69a3ee9cc0b6503ce8883672011c44bc772c0fa2aR316) it is stated that `Export` is not to be called concurrently. `ForceFlush` seems to export metrics, have a timeout and can result in success or failure. How is it different from `Export`? Is it necessary if `Export` can be called concurrently?

specification/metrics/sdk.md
```diff
@@ -303,6 +303,80 @@ Push Metric Exporter sends the data on its own schedule. Here are some examples:
...
+via a callback or an event. OpenTelemetry client authors can decide if they want
+to make the flush timeout configurable.
+
+##### Shutdown()
+
+Shuts down the exporter. Called when SDK is shut down. This is an opportunity
+for exporter to do any cleanup required.
+
+Shutdown should be called only once for each `MetricExporter` instance. After
+the call to `Shutdown` subsequent calls to `MetricExporter` are not allowed and
```
ocelotl, line 373: `MetricExporter` is not called, `Export` is.

specification/metrics/sdk.md
```diff
@@ -303,6 +303,80 @@ Push Metric Exporter sends the data on its own schedule. Here are some examples:
...
+should return a Failure result.
+
+`Shutdown` should not block indefinitely (e.g. if it attempts to flush the data
+and the destination is unavailable). OpenTelemetry client authors can decide if
```
ocelotl, line 377:
```suggestion
and the destination is unavailable). OpenTelemetry implementation authors can decide if
```

specification/metrics/sdk_exporters/in-memory.md
```diff
@@ -5,3 +5,9 @@
 Note: this specification is subject to major changes. To avoid thrusting
 language client maintainers, we don't recommend OpenTelemetry clients to start
 the implementation unless explicitly communicated.
+
+The in-memory exporter accumulates telemetry data in the local memory and allows
```
ocelotl, line 9: Should we even have this in our specification? Which is the protocol for this kind of exporter?

I understand that this is a useful idea and having an exporter of this kind can help debug. Probably most of the implementations will have it because of that, but is it necessary to specify a debugging aid in this document?

ocelotl (changes requested):

specification/metrics/sdk.md
```diff
@@ -303,6 +303,80 @@ Push Metric Exporter sends the data on its own schedule. Here are some examples:
...
+This is a hint to ensure that the export of any `Metrics` the exporter has
+received prior to the call to `ForceFlush` SHOULD be completed as soon as
+possible, preferably before returning from this method.
+
+`ForceFlush` SHOULD provide a way to let the caller know whether it succeeded,
+failed or timed out.
+
+`ForceFlush` SHOULD only be called in cases where it is absolutely necessary,
+such as when using some FaaS providers that may suspend the process after an
+invocation, but before the exporter exports the completed spans.
```
cijothomas, line 360:
```suggestion
invocation, but before the exporter exports the completed metrics.
```

specification/metrics/sdk.md
reyang: It is the same as SpanExporter Export vs. ForceFlush https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#forceflush-2.

specification/metrics/sdk.md
reyang: "OpenTelemetry client authors" seems to be a more widely used phrase https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#shutdown-2.

specification/metrics/sdk.md
reyang: Resolved by scoping down https://github.com/open-telemetry/opentelemetry-specification/pull/1864#discussion_r688712055.

jmacd (approved):

specification/metrics/sdk.md
```diff
@@ -33,6 +33,12 @@ Table of Contents
 
 ## MeterProvider
 
+`MeterProvider` MUST provide a way to allow a [Resource](../resource/sdk.md) to
+be specified. If a `Resource` is specfied, it SHOULD be associated with all the
```
cijothomas, line 37:
```suggestion
be specified. If a `Resource` is specified, it SHOULD be associated with all the
```

specification/metrics/sdk.md
reyang: I've adjusted the wording based on the [8/19/2021 SIG discussion](https://docs.google.com/document/d/1pdvPeKjA8v8w_fGKAN68JjWBmVJtPCpqdi9IZrd6eEo/edit#).

@jmacd please take another look.

specification/metrics/sdk.md
```diff
@@ -33,6 +33,12 @@ Table of Contents
...
+metrics produced by any `Meter` from the `MeterProvider`. The [tracing SDK
+specfication](../trace/sdk.md#additional-span-interfaces) has provided some
```
cijothomas, line 39:
```suggestion
specification](../trace/sdk.md#additional-span-interfaces) has provided some
```

specification/metrics/sdk.md
```diff
@@ -303,6 +309,76 @@ Push Metric Exporter sends the data on its own schedule. Here are some examples:
...
+after which the call must time out with an error result (Failure).
+
+Any retry logic that is required by the exporter is the responsibility of the
+exporter. The default SDK SHOULD NOT implement retry logic, as the required
+logic is likely to depend heavily on the specific protocol and backend the metrics
+are being sent to.
+
+**Parameters:**
+
+`batch` - a batch of `Metrics`. The exact data type of the batch is language
```
cijothomas, line 335: would we be adding (i.e is that planned for any future PRs) more details about what information is containing in each of the `Metrics`?
For example:
1. Will the exporter be able to tell if a given `Metric` is of temporality delta or cumulative, so that the exporter can chose to drop/log if it receives a `Metric` of Delta, but it only expects Cumulative.
2. Will the exporter be able to tell which instrument produced the given Metric? Or exporter will have no idea "how" the metric was produced. A Sum could have come from a Counter instrument or even Histogram instrument. (when view is configured to produce SUM from Histogram instruments)

cijothomas: is `Metrics` here referring to the TimeseriesModel https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#timeseries-model or OTLP model https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#opentelemetry-protocol-data-model

jmacd (approved):

specification/metrics/sdk.md
```diff
@@ -33,6 +33,12 @@ Table of Contents
 
 ## MeterProvider
 
+`MeterProvider` MUST provide a way to allow a [Resource](../resource/sdk.md) to
```
jkwatson, line 36:
```suggestion
A `MeterProvider` MUST provide a way to allow a [Resource](../resource/sdk.md) to
```

specification/metrics/sdk.md
```diff
@@ -303,6 +309,76 @@ Push Metric Exporter sends the data on its own schedule. Here are some examples:
 * Sends the data based on a user configured schedule, e.g. every 1 minute.
 * Sends the data when there is a severe error.
 
+#### Interface Definition
+
+Push Metric Exporter MUST support the following functions:
```
jkwatson, line 314:
```suggestion
A Push Metric Exporter MUST support the following functions:
```

jkwatson (approved):

jsuereth (approved):

specification/metrics/sdk.md
jmacd: In the Tuesday SIG call, we discussed a scenario, presumed to be common, where an SDK will push metrics data to two endpoints using two exporters. (1) for upgrading from a legacy protocol to a new protocol, (2) for reading metrics from inside a process.

The question: how much do we need to specify in order to support two push exporters simultaneously. Answer? If they look like a single exporter, that's fine.

specification/metrics/sdk.md
jmacd: :100:

specification/metrics/sdk.md
jmacd: :shrug: Let's follow the convention. :grin:

jmacd:
Discussed in the SIG call, and not mentioned in the comments:

The OTel-Go prototype metrics `Reader` interface, associated with pull exporters and push exporters alike, accepts an `ExportKindSelector` that allows each exporter to choose the export strategy it desires.

The prototype's basic processor, [here](https://github.com/open-telemetry/opentelemetry-go/blob/c912b179fd595cc7c6eec0b22e6c1371e31241d7/sdk/metric/processor/basic/basic.go), keeps sufficient state during the collect-and-process code path to support both delta and cumulative exporters at once, provided it was given the proper configuration up front.

The "Reader" API (presently named `CheckpointSet`) accepts the `ExportKindSelector` [here in its `ForEach()` method](https://github.com/open-telemetry/opentelemetry-go/blob/c912b179fd595cc7c6eec0b22e6c1371e31241d7/sdk/export/metric/metric.go#L250) for iterating over the results of `Collect()` in both Push and Pull code paths.

The "configuration up front" is provided via the basic Processor's constructor. Each Exporter is also an ExportKindSelector, meaning the exporter will be consulted once per `*metric.Descriptor` as to the choice of export strategy. `ExportKind` is a bit set, allowing multiple exporters to select both export strategies when desired.

```
// ExportKind indicates the kind of data exported by an exporter.
// These bits may be OR-d together when multiple exporters are in use.
type ExportKind int

const (
	// CumulativeExportKind indicates that an Exporter expects a
	// Cumulative Aggregation.
	CumulativeExportKind ExportKind = 1

	// DeltaExportKind indicates that an Exporter expects a
	// Delta Aggregation.
	DeltaExportKind ExportKind = 2
)
```

specification/metrics/sdk.md
jmacd: @cijothomas see my comment in the main thread. I think we can resolve this and continue specifying the pull exporter to answer your questions. I've pointed to locations in the Otel-Go prototype where multiple temporality is exporter to push and pull exporters in a consistent manner.

specification/metrics/sdk_exporters/in-memory.md
jmacd: If it were just about debugging use-cases, I agree and would recommend statements about debugging support be included in the [library guidelines](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/library-guidelines.md). In this case specifically, there's a connection to the yet-unspecified Pull Exporter but I think we should remove this file and move it to the pull export PR.

reyang: @jmacd and I discussed offline, we'll not include this change in the current PR.
I'll cover it in my next PR.

PR #1867, opened 2021-08-13
Lambda instrumentations should check if X-Ray parent context is valid

NathanielRN:
## Changes

The behavior of Lambda requires that we check that the context `IsValid` in addition to checking if it sampled. We update the specs to reflect that.

Also changed the wording in the API Gateway section so that we know that the attributes are set in _both_ the "as a pure proxy" and "as a configuration for a REST API" setups of API Gateway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@ release.
 
 ### Semantic Conventions
 
+- Lambda instrumentations should check if X-Ray parent context is valid
+  ([#1867](https://github.com/open-telemetry/opentelemetry-specification/pull/1867))
+
 ### Compatibility
 
 ### OpenTelemetry Protocol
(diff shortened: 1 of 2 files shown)


NathanielRN:
@anuraaga would appreciate your review on this PR!

anuraaga (approved):

carlosalberto (approved):

carlosalberto:
@NathanielRN Maybe this is enough of a change to deserve a CHANGELOG? Or think it's not really needed for this change?

NathanielRN:
@carlosalberto Originally I didn't think it was necessary, but I guess it's safer to add it so that devs instrumenting Lambda know that they have an action item to do 😄 I've added a CHANGELOG entry!

PR #1869, opened 2021-08-16
Added value zipkin_json and zipkin_proto as expected values for OTEL_TRACES_EXPORTER

anirudhbagri:
Fixes #
Added value `zipkin_json` and `zipkin_proto` as expected values for `OTEL_TRACES_EXPORTER`
## Changes

Related issues #

### DIFF
--- a/specification/sdk-environment-variables.md
+++ b/specification/sdk-environment-variables.md
@@ -167,8 +167,9 @@ Known values for OTEL_TRACES_EXPORTER are:
 
 - `"otlp"`: [OTLP](./protocol/otlp.md)
 - `"jaeger"`: [Jaeger gRPC](https://www.jaegertracing.io/docs/1.21/apis/#protobuf-via-grpc-stable)
-- `"zipkin_proto"`: [Zipkin](https://zipkin.io/zipkin-api/) (Defaults to [protobuf](https://github.com/openzipkin/zipkin-api/blob/master/zipkin.proto) format)
+- `"zipkin"`: [Zipkin](https://zipkin.io/zipkin-api/) (Defaults to [protobuf](https://github.com/openzipkin/zipkin-api/blob/master/zipkin.proto) format)
 - `"zipkin_json"`: [Zipkin](https://zipkin.io/zipkin-api/)
+- `"zipkin_proto"`: [Zipkin](https://zipkin.io/zipkin-api/)
 - `"none"`: No automatically configured exporter for traces.
 
 Known values for OTEL_METRICS_EXPORTER are:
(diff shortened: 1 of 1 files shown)


specification/sdk-environment-variables.md
```diff
@@ -167,7 +167,8 @@ Known values for OTEL_TRACES_EXPORTER are:
 
 - `"otlp"`: [OTLP](./protocol/otlp.md)
 - `"jaeger"`: [Jaeger gRPC](https://www.jaegertracing.io/docs/1.21/apis/#protobuf-via-grpc-stable)
-- `"zipkin"`: [Zipkin](https://zipkin.io/zipkin-api/) (Defaults to [protobuf](https://github.com/openzipkin/zipkin-api/blob/master/zipkin.proto) format)
```
jkwatson, line 170: This would be a breaking change. I'd prefer if we didn't get rid of this, since some SDKs only support on zipkin option as it is.

anirudhbagri: Okay, in that case should we add the other two options with this?

jkwatson: I'm not sure. Can you update the title of this PR so that people will know it's related to zipkin formats?

carlosalberto:
cc @anuraaga

specification/sdk-environment-variables.md
lonewolf3739: Hmm, In python SIG we split up the exporters based on protocol + serialisation so that users can avoid taking dependency on unwanted packages. This implementation suggestion was left to each SIG to decide upon and not every SIG does this. I am not sure if this should be addressed at spec level because it is not just `zipkin` but other as well (e.g `otlp_grpc` and `otlp_http`). The author of this PR had an issue where they set the env to `zipkin` and it didn't work. I think python sdk should have used `zipkin_proto` in this case since protobuf is the spec default recommendation.

anuraaga (commented):
There's an environment variable already reserved for use for the Zipkin protocol. Any reason not to use it?

https://github.com/open-telemetry/opentelemetry-specification/blob/c5f1a70340059953244fcc9aabb371347e0989fc/specification/sdk-environment-variables.md#zipkin-exporter

`OTEL_EXPORTER_ZIPKIN_PROTOCOL`

anirudhbagri:
Even with __ setup, you will still need to specify `OTEL_TRACES_EXPORTER`, right? Which doesn't accept plain `zipkin` as a parameter, at least with python SDK.

anuraaga:
I guess Python could be updated to look at both variables? Currently it's not in spec I guess.

PR #1880, opened 2021-08-19
Add OTLP protocol configuration options

### LINKED ISSUE #1879: Add OTEL_EXPORTER_OTLP_PROTOCOL config option

Formalize support for `OTEL_EXPORTER_OTLP_PROTOCOL`, `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL`, and `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL` environment variables that are [currently reserved](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#specify-protocol).

What did you expect to see?

I expected to be able to configure the SDK to specify the OTLP protocol. 

**Additional context.**

Specifically working in the java SDK, which has `otlp/http` support that is not configurable via environment variables.

jack-berg:
Fixes #1879 

## Changes

Add `OTEL_EXPORTER_OTLP_PROTOCOL`, `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL`, and `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL` environment variables for specifying the OTLP protocol.

### DIFF
--- a/specification/sdk-environment-variables.md
+++ b/specification/sdk-environment-variables.md
@@ -117,11 +117,13 @@ See [OpenTelemetry Protocol Exporter Configuration Options](./protocol/exporter.
 | Name                            | Description                                                      | Default                                                                                          |
 |---------------------------------|------------------------------------------------------------------|--------------------------------------------------------------------------------------------------|
 | OTEL_EXPORTER_JAEGER_AGENT_HOST | Hostname for the Jaeger agent                                    | "localhost"                                                                                      |
-| OTEL_EXPORTER_JAEGER_AGENT_PORT | Port for the Jaeger agent                                        | 6832                                                                                             |
+| OTEL_EXPORTER_JAEGER_AGENT_PORT | Port for the Jaeger agent `compact` Thrift protocol              | 6831                                                                                             |
 | OTEL_EXPORTER_JAEGER_ENDPOINT   | HTTP endpoint for Jaeger traces                                  | <!-- markdown-link-check-disable --> "http://localhost:14250"<!-- markdown-link-check-enable --> |
 | OTEL_EXPORTER_JAEGER_TIMEOUT    | Maximum time the Jaeger exporter will wait for each batch export | 10s                                                                                              |
 | OTEL_EXPORTER_JAEGER_USER       | Username to be used for HTTP basic authentication                | -                                                                                                |
 | OTEL_EXPORTER_JAEGER_PASSWORD   | Password to be used for HTTP basic authentication                | -                                                                                                |
+
+See [Jaeger Agent](https://www.jaegertracing.io/docs/latest/deployment/#agent) documentation.
 
 ## Zipkin Exporter
 
--- a/specification/sdk-environment-variables.md
+++ b/specification/sdk-environment-variables.md
@@ -106,21 +106,9 @@ See the SDK [Span Limits](trace/sdk.md#span-limits) section for the definition o
 | OTEL_EVENT_ATTRIBUTE_COUNT_LIMIT       | Maximum allowed attribute per span event count | 128     |       |
 | OTEL_LINK_ATTRIBUTE_COUNT_LIMIT        | Maximum allowed attribute per span link count  | 128     |       |
 
-## OTLP Exporters
+## OTLP Exporter
 
 See [OpenTelemetry Protocol Exporter Configuration Options](./protocol/exporter.md).
-
-[OTLP](./protocol/exporter.md) supports the following transport protocols and corresponding [exporters values](#exporter-selection):
-
-| Protocol          | Exporter Value    | Description |
-| ----------------- | ----------------- | ----------- |
-| `grpc`            | "otlp_grpc"       | Protobuf-encoded data using gRPC wire format over HTTP/2 connection |
-| `http/protobuf`   | "otlp_http_proto" | Protobuf-encoded data over HTTP connection |
-| `http/json`       | "otlp_http_json"  | JSON-encoded data over HTTP connection |
-
-SDKs MUST support either `grpc` or `http/protobuf` and SHOULD support both. They also MAY support `http/json`.
-
-SDKs have an unspecified default protocol, which is used if the exporter selection is "otlp".
 
 ## Jaeger Exporter
 
@@ -177,20 +165,14 @@ The SDK MAY accept a comma-separated list to enable setting multiple exporters.
 
 Known values for OTEL_TRACES_EXPORTER are:
 
-- `"otlp"`: [OTLP](./protocol/otlp.md) with the unspecified default protocol described in [OTLP exporters](#otlp-exporters).
-- `"otlp_grpc"`: [OTLP](./protocol/otlp.md) with the `otlp/grpc` protocol.
-- `"otlp_http_proto"`: [OTLP](./protocol/otlp.md) with the `http/protobuf` protocol.
-- `"otlp_http_json"`: [OTLP](./protocol/otlp.md) with the `http/json` protocol.
+- `"otlp"`: [OTLP](./protocol/otlp.md)
 - `"jaeger"`: [Jaeger gRPC](https://www.jaegertracing.io/docs/1.21/apis/#protobuf-via-grpc-stable)
 - `"zipkin"`: [Zipkin](https://zipkin.io/zipkin-api/) (Defaults to [protobuf](https://github.com/openzipkin/zipkin-api/blob/master/zipkin.proto) format)
 - `"none"`: No automatically configured exporter for traces.
 
 Known values for OTEL_METRICS_EXPORTER are:
 
-- `"otlp"`: [OTLP](./protocol/otlp.md) with the unspecified default protocol as described in [OTLP exporters](#otlp-exporters).
-- `"otlp_grpc"`: [OTLP](./protocol/otlp.md) with the `otlp/grpc` protocol.
-- `"otlp_http_proto"`: [OTLP](./protocol/otlp.md) with the `http/protobuf` protocol.
-- `"otlp_http_json"`: [OTLP](./protocol/otlp.md) with the `http/json` protocol.
+- `"otlp"`: [OTLP](./protocol/otlp.md)
 - `"prometheus"`: [Prometheus](https://github.com/prometheus/docs/blob/master/content/docs/instrumenting/exposition_formats.md)
 - `"none"`: No automatically configured exporter for metrics.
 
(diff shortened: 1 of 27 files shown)


jkwatson (approved):

anuraaga (approved):

carlosalberto:
cc @jsuereth (maybe I'm wrong but I think you were interested in this ;) )

jsuereth (changes requested):
We need to confirm with maintainers that the meaning of these enums i the same across SDKs.

There was concern (previously), that "http/protobuf" and "http/json" are not specific enough in what they mean, and that existing "http/protobuf" or "http/json" exporters in SDKs are actually incompatible with each other.

Opening this as a blocking comments until we have verification from language SDK maintainers that they're comfortable with this as written given the previous attempt to specify this.

specification/sdk-environment-variables.md
```diff
@@ -106,10 +106,22 @@ See the SDK [Span Limits](trace/sdk.md#span-limits) section for the definition o
...
+## OTLP Exporters
 
 See [OpenTelemetry Protocol Exporter Configuration Options](./protocol/exporter.md).
 
+[OTLP](./protocol/exporter.md) supports the following transport protocols and corresponding [exporters values](#exporter-selection):
+
+| Protocol          | Exporter Value    | Description |
+| ----------------- | ----------------- | ----------- |
+| `grpc`            | "otlp_grpc"       | Protobuf-encoded data using gRPC wire format over HTTP/2 connection |
+| `http/protobuf`   | "otlp_http_proto" | Protobuf-encoded data over HTTP connection |
```
jack-berg, line 118: Opted to `otlp_http_proto` instead of `otlp_http_protobuf` to be inline with `zipkin_proto` added [here](https://github.com/open-telemetry/opentelemetry-specification/pull/1869).

jack-berg:
I've added new commits which revert my first attempt in favor of the approach discussed in the sig meeting of adding new exporter options for the different protocols instead of adding a new configuration option.

specification/sdk-environment-variables.md
```diff
@@ -165,14 +177,20 @@ The SDK MAY accept a comma-separated list to enable setting multiple exporters.
 
 Known values for OTEL_TRACES_EXPORTER are:
 
-- `"otlp"`: [OTLP](./protocol/otlp.md)
+- `"otlp"`: [OTLP](./protocol/otlp.md) with the unspecified default protocol described in [OTLP exporters](#otlp-exporters).
+- `"otlp_grpc"`: [OTLP](./protocol/otlp.md) with the `otlp/grpc` protocol.
```
anuraaga, line 181: Using values `_EXPORTER` doesn't seem to fit well with the other variables we have defined. Notably, variables such as `OTEL_EXPORTER_OTLP_ENDPOINT` would need to be extended with `OTEL_EXPORTER_OTLP_GRPC_ENDPOINT`? It seems like an unnecessary explosion.

I understand this was discussed verbally at the sig meeting so am probably missing detail of the motivation for the switcheroo, but I can't find a signficant downside of a `_PROTOCOL` approach - it requires looking at multiple environment variables to configure an exporter anyways (_EXPORTER, _ENDPOINT, _ETC) so adding another one to represent a knob for wire format seems natural. It's true that in most cases, the different formats represent different exporter artifacts, but that seems like an implementation detail that is still compatible with this approach - in fact we've already implemented it in Java with experimental variable names. It doesn't seem to have the downside of allowing multiple exporters like otlp_grpc, otlp_http_proto defined at the same time without being able to configure them.

carlosalberto: Ping @tedsuo ;)

tedsuo: @anuraaga the original proposal does not reflect how exporters are actually implemented in any language. So the result is a bizarre mismatch. And with both proposals, there is no way to configure multiple OTLP exporters, even though it is possible to install them. So, no design is right here. 

I general, I am more and more unhappy about this ad-hoc approach to environment variables as a way to drive the conversation forwards. I'm watching "design theories" be lifted and then dropped as they are convenient. This is a terrible approach to designing the primary surface area that our users are going to interact with. I really want to see a more consistent and coherent approach to configuration design. 

Sorry @jack-berg, I know you're just trying to solve some annoying configuration issues that currently exist. But should we really be adding more env vars when we know they may not fit very well? 

Looking again, I do agree that if what we want is to allow multiple exporters to be installed, but only allow one _OTLP_ exporter to be installed, the original proposal has fewer side effects. But is that what we actually want? Down the line, it will definitely be possible to install multiple exporters. If it's possible to install both `otel_grpc` and `otel_http_json`, how would these env vars be expected to interact with that configuration?

Ironically, the proposal @anuraaga dislikes the most (seperate env vars for each exporter, e.g. OTEL_EXPORTER_OTLP_GRPC_ENDPOINT) is actually the one which correctly maps to our architecture. 

Env vars are sadness. If we had a conf file, I don't think we'd even be asking these questions.

anuraaga: > the original proposal does not reflect how exporters are actually implemented in any language. So the result is a bizarre mismatch. 

I don't see the coupling between the exporter implementation and how the configuration knob is presented. @jack-berg has validated this by implementing some experimental variables in Java and I don't think anyone has found it to be bizarre. What concretely do you think is bizarre about it?

> But is that what we actually want? Down the line, it will definitely be possible to install multiple exporters. 

We currently do allow multiple exporters to be installed with a CSV to `OTEL_*_EXPORTER`, we probably shouldn't have as it has caused a huge explosion in configuration surface in the env vars. Anyways, there seems to be no conceptual basis for allowing `otel,otel_grpc` and not `otel,otel`, which the `_PROTOCOL` approach avoids. I agree that this is where a configuration file approach would allow more flexibility. But it doesn't have to be one or the other - perhaps we are missing text but the env vars still provide a good, simple approach. Perhaps we are missing some text in this MD file that the goal of the env vars is to provide a simple configuration, without maximum flexibility. A way to think about it could be that env vars are likely the primary mechanism when using the collector - you just need a simple set of variables to get the data into the collector which has the flexible pipelines to process them.

At least that's what I figure they're for - it's why I would advocate the `OTEL_EXPORTER_OTLP_PROTOCOL` variable. At a glance it might seem like creep by adding another variable, but it actually does not add as much configurability as adding to the OTEL_EXPORTER types, which opens up the greater configuration surface of handling the case of multiple OTLP exporters being defined.

But freezing variables until there is a more well defined story seems ok too - I'm mostly concerned about adding the new exporter types as currently proposed. My goal I think is the same as @tedsuo's, to reduce the expansion of the configuration surface. Ironically, it is an additional variable that does that I think.

> Env vars are sadness. If we had a conf file, I don't think we'd even be asking these questions.

This isn't fair - personally I've had a much easier time with env vars when deploying apps to k8s for example, and the config file requirement of the OTel collector is actually somewhat annoying. Others probably have different experiences though. I doubt one approach is clearly better - I suspect that a simple set of environment variables along with the ability to define something more flexibly with a file can satisfy both itches.

tedsuo: @anuraaga I agree that env vars are very useful. Sorry, I didn't mean to slight them. In fact, I would prefer that OTel be designed so that env vars and conf files keep a 1:1 basis – anything that can be defined in a conf file should be overridable with an env var. I believe that env vars are critical in that regard, and should be treated as first class citizens. Like you say, there are many situations where it is easier to set an env var than it is to change a conf file, especially in cloud environments where changing a conf file means deploying a new image.

But when I say env vars make me sad, it's more about the nefarious way they seem to make good design difficult. I've been on multiple projects which accidentally painted themselves into a corner by taking our current approach. We would steadily add env vars which made sense at the time but later created headaches, and couldn't be removed for compatibility reasons.

That's actually my concern with these particular proposals (both of them!). It's not the way they expand the surface area now, it's that I can see how they may force us to make a complicated set of rules later, when we expose planned features in the future.

In this case, if we had a proper design and layout, I suspect it might have lead us to a solution which exposed more env vars, but didn't feel "clunky" because it fit a regular pattern we were applying everywhere.

I've learned that it is better to start by designing the conf file and env vars at the same time, always make them match, and always lay out the options in a way that reflects the architecture of the system. For whatever reason, the presence of the conf file seems to help with all of that. Maybe because we are visual critters. We didn't do that with OTel. Oh well. I'm glad we're trying to do it now. We'll fix it in post. :)

At any rate, I propose we add these env vars, as they were originally defined by @jack-berg and @anuraaga. We need to clean our design up, but it's been two weeks now and it's not fair to keep blocking this feature. If we want to freeze future env vars while we bang out a design, fine, but let's not hold up needed additions that have already been waiting. 

Again, many apologies to Jack for having to put up with us while we went around and around. Turns out this was trivial but also important. :)

tedsuo (approved):
Having now discussed this at leeeeeeeeeength, I'm fine with adding the original proposal.

specification/sdk-environment-variables.md
```diff
@@ -106,10 +106,22 @@ See the SDK [Span Limits](trace/sdk.md#span-limits) section for the definition o
...
+
+| Protocol          | Exporter Value    | Description |
+| ----------------- | ----------------- | ----------- |
+| `grpc`            | "otlp_grpc"       | Protobuf-encoded data using gRPC wire format over HTTP/2 connection |
+| `http/protobuf`   | "otlp_http_proto" | Protobuf-encoded data over HTTP connection |
+| `http/json`       | "otlp_http_json"  | JSON-encoded data over HTTP connection |
+
+SDKs MUST support either `grpc` or `http/protobuf` and SHOULD support both. They also MAY support `http/json`.
+
+SDKs have an unspecified default protocol, which is used if the exporter selection is "otlp".
```
lonewolf3739, line 123: Small question, what does unspecified default protocol mean here?

jack-berg: My understanding is that it means that language SDKs have a default, but that default isn't dictated by the specification. 

See [this convo](https://github.com/open-telemetry/opentelemetry-specification/pull/1358#discussion_r564498561) for context.

specification/sdk-environment-variables.md
jack-berg: A great discussion! And definitely agree with the need for a well designed conf file (especially thinking about configuring metric views 😬!), which env vars can be neatly mapped to. I do think that these new environment variables will be straight forward to map to whatever conf file spec otel lands on.

anuraaga (approved):
Thanks everyone!

carlosalberto (approved):

carlosalberto:
Ping @jsuereth

carlosalberto:
@jsuereth Ping ;) Is your original changes request still standing? Else, let's merge this one.

jsuereth (approved):
Think we've left this open with sufficient time for SDK concerns and raised it multiple SiGs.  I left ONE comment in the maintainer channel for last-minute discussions.

I don't think this belongs in the specificatoin but taking note here:

Compliance to protocol should be evaluated via Otel Collector ingestion of telemetry.

carlosalberto:
@jack-berg Oops, looks like we don't have a CHANGELOG entry, please add one ;)

jack-berg:
All set @carlosalberto!

alolita (commented):
+1

Issue #1870, opened 2021-08-17
Propagation Getter spec can't support W3C spec


tsloughter:
The specification for Otel's TextMap propagator `Get` function and the W3C spec conflict in how to handle multiple of the same key in a list of pairs. The Otel spec has an optional `Get` that must return value, https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#get, while in W3C in the case of `traceparent` a new trace should be started: https://github.com/w3c/trace-context/blob/main/test/test.py#L136-L137

For `tracestate` W3C says it must combine the values of the `tracestate` key.

I think `Getter` has to change to return every pair with the key being looked up. But this is not backwards compatible? Maybe it will have to be that the `get` function can return a single value if only one is found but a list if there are multiple? That way existing `Getter` implementations continue to work.

mwear:
It would be possible to combine the multiple headers into one by concatenating the values separated by a `,`. This is briefly mentioned in the [W3C Trace Context spec](https://www.w3.org/TR/trace-context/#tracestate-header-field-values). It says:

 > When receiving multiple `tracestate` header fields, they MUST be combined into a single header according to [RFC7230](https://httpwg.org/specs/rfc7230.html#field.order).
 
 The relevant part of RFC7230 says:
 
 > A recipient MAY combine multiple header fields with the same field name into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field value to the combined field value in order, separated by a comma. The order in which header fields with the same field name are received is therefore significant to the interpretation of the combined field value; a proxy MUST NOT change the order of these field values when forwarding a message.

tsloughter:
Yes, this is what we do in the Erlang W3C propagator implementation. But this became not possible in the propagator itself when adding support for a `Get` function that can be set by the user and only returns the first element.

An option instead of adding a `GetAll` or having `Get` return a list would be to define `Get` to have to combine duplicates using the method of RFC7230 if it is for a data structure that can have duplicates -- meaning if your `Get` is defined for a dictionary it wouldn't have to do anything special, but one that takes a list or array would have to search for all matches and combine them.

tsloughter:
Closing because I found https://github.com/open-telemetry/opentelemetry-specification/issues/433 already exists from a year and a half ago and I've opened https://github.com/open-telemetry/opentelemetry-specification/pull/1884

PR #1884, opened 2021-08-23
update Get function in propagators to combine duplicate keys

### LINKED ISSUE #433: Propagation - Getter handling of multi-value headers/attributes
Some formats (http) allow for a single key to return multiple values.  This is used in [w3c trace-context](https://w3c.github.io/trace-context/#tracestate-header-field-values) which we plan to support:
> Tracestate MAY be sent or received as multiple header fields. ... The tracestate header SHOULD be sent as a single field when possible, but MAY be split into multiple header fields. 

We don't currently specify this requirement in our spec:
https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-propagators.md#getter-argument

As a consequence, we aren't handling that well in java, and likely other languages:
https://github.com/open-telemetry/opentelemetry-java/blob/v0.2.x/api/src/main/java/io/opentelemetry/context/propagation/HttpTextFormat.java#L123

How do we want to handle this?  Should we push responsibility for joining them properly to the `Getter`, or should the API support returning multiple values?

tsloughter:
Fixes #433 and #1870

## Changes

This change makes it so the `Get` function used in `TextMapPropagator` extraction must combine the values of duplicate keys, a requirement for supporting W3C TraceContext.


jkwatson:
Since the Getter implementation is provided by instrumentation, not the API, is this something that we can say MUST to? I'm not sure we can do more than provide strong guidance on this, since we're not in control of all implementations.

carlosalberto:
> ? I'm not sure we can do more than provide strong guidance on this

Agreed, same _feeling_ (**clear** guidance should help a lot, hopefully!)

tsloughter:
Yea, I guess if being a MUST in the api spec implies that the api has to ensure any `Get` function passed to the propagator does indeed do this then it isn't something we can require since, unlike giving an interface or type spec, we can't validate it.

But it was already a MUST to give the first entry, so there are likely other places that will need to be updated.

Would changing it to instead saying that the `Get` function returns the value for a key (removing the existing part about returning the first element) and then noting that the function should do the combining in order to not have potentially unexpected behaviour when used with the builtin propagators?

specification/context/api-propagators.md
```diff
@@ -216,14 +216,19 @@ For example, it can be used to detect all keys following the `uberctx-{user-defi
 
 ##### Get
 
-The Get function MUST return the first value of the given propagation key or return null if the key doesn't exist.
+The Get function returns the value for a given propagation key or returns null if
+the key doesn't exist. If the carrier is a type that allows duplicate keys, like
+a list or array as opposed to a dictionary, then `Get` MUST get all the values
```
anuraaga, line 221: This seems to make sense given that if multiple traceparent come in, they would get concatenated and our parser would fail, satisfying the test case

https://github.com/w3c/trace-context/blob/main/test/test.py#L137

tsloughter:
Sounds like there is agreement so I'll get this updated soon.

tsloughter:
I've made an update but have not removed the use of `MUST`. I still can but didn't because I realized it is used in reference to case sensitivity. I'm not sure if both case sensitivity and duplicates should be changed to use `SHOULD` or if they should remain `MUST` but I feel they should at least be the same.

tsloughter:
Ping? I've updated the PR.

mateuszrzeszutek (approved):

MSNev (approved):

jmacd:
@tsloughter I think we should leave the MUST for case-sensitivity and change the treatment of duplicates to a SHOULD.

tsloughter:
@jmacd I don't see how they are different? Both are required to properly handle HTTP headers.

jmacd:
Sorry @tsloughter, I've neglected this PR. :(
I don't feel I have a qualified opinion on this matter.
I would like to re-assign it to a TC member who I think does.

tsloughter:
Ping @tedsuo

PR #1888, opened 2021-08-26
Add MetricReader interface

reyang:
Follow up #1864.

## Changes

* Added the MetricReader interface (refer to https://github.com/open-telemetry/opentelemetry-specification/pull/1730#discussion_r669024476 for the initial discussion)

### DIFF
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -587,6 +587,11 @@ This is an implementation of the `MetricReader` which collects metrics when
 [Collect](#collect) is called. and passes the metrics to the configured
 [MetricExporter](#metricexporter).
 
+**Note:** The name **"Base" exporting MetricReader** has nothing to do with
+object oriented programming or class inheritance / hierarchy. In case it might
+introduce confusion in certain programming languages, individual language client
+CAN choose a different wording.
+
 Configurable parameters:
 
 * `exporter` - the exporter where the metrics are sent to.
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -621,7 +621,7 @@ If the configured exporter only supports [pull mode](#pull-metric-exporter):
 * [Collect](#collect) SHOULD only be called when the [Pull Metric
   Exporter](#pull-metric-exporter) triggers the scraping, otherwise it SHOULD be
   treated as an error.
-* Individual language client MAY decide if [OnForceFlush](#onforceflush) will
+* Individual language clients MAY decide if [OnForceFlush](#onforceflush) will
   return immediately (with an indication of failure) or wait for the next
   [OnCollect](#oncollect) call to complete.
 
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -537,6 +537,16 @@ measurements using the equivalent of the following naive algorithm:
 * Handling the [ForceFlush](#forceflush) and [Shutdown](#shutdown) signals from
   the SDK.
 
+The SDK MUST support multiple `MetricReader` instances to be registered on the
+same `MeterProvider`, and the [MetricReader.Collect](#collect) invocation on one
+`MetricReader` instance MUST NOT introduce side-effects to other `MetricReader`
+instances. For example, if a `MetricReader` instance is receiving metric data
+points that have [delta temporality](./datamodel.md#temporality) via the
+[MetricReader.OnCollect](#oncollect) callback, it is expected that SDK will
+update the time range - e.g. from (T<sub>n</sub>, T<sub>n+1</sub>] to
+(T<sub>n+1</sub>, T<sub>n+2</sub>] - **ONLY** for this particular `MetricReader`
+instance, so the next `OnCollect` will get a new start time.
+
 ```text
 +-----------------+            +--------------+
 |                 | Metrics... |              |
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -450,7 +450,7 @@ failed or timed out.
 `Collect` does not have any required parameters, however, individual language
 client MAY choose to support optional parameters (e.g. filter, timeout).
 
-#### OnCollect(batch)
+#### OnCollect
 
 Callback function which will be triggered by [Collect](#collect).
 
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -752,10 +752,15 @@ Pull Metric Exporter reacts to the metrics scrapers and reports the data
 passively. This pattern has been widely adopted by
 [Prometheus](https://prometheus.io/).
 
+Unlike [Push Metric Exporter](#push-metric-exporter) which can send data on its
+own schedule, pull exporter can only send the data when it is being asked by the
+scraper, and `ForceFlush` would not make sense.
+
 Implementors MAY choose the best idiomatic design for their language. For
-example, they could apply the [Push Metric Exporter
-interface](#push-metric-exporter) design for consistency, or they could design a
-completely different pull exporter interface.
+example, they could generalize the [Push Metric Exporter
+interface](#push-metric-exporter) design and use that for consistency, they
+could model the pull exporter as [MetricReader](#metricreader), or they could
+design a completely different pull exporter interface.
 
 ## Defaults and Configuration
 
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -679,6 +679,46 @@ in the SDK:
                                +-----------------------------+
 ```
 
+The following sequence diagram shows how `MetricExporter` interacts with other
+components in the SDK. Note that the _SIGNAL_ in the diagram could come from:
+
+* A periodic timer from the `MetricReader`, if the MetricReader is a [Periodic
+  exporting MetricReader](#periodic-exporting-metricreader).
+* A scraping call received by the `MetricExporter`, if the MetricExporter is a
+  [Pull Metric Exporter](#pull-metric-exporter).
+* An explicit call to the `MetricReader.Collect()` from the user.
+* An implicit call to the `MetricReader.Collect()` (e.g.
+  `MeterProvider.ForceFlush()`).
+
+```text
++---------------+                         +--------------+             +----------------+
+|               |                         |              |             |                |
+| MeterProvider |                         | MetricReader |             | MetricExporter |
+|               |                         |              |             |                |
++---------------+                         +--------------+             +----------------+
+        |           +--------+                   |                              |
+        |           |        |                   |                              |
+        |           | SIGNAL |                   |                              |
+        |           |        |                   |                              |
+        |           +---+----+                   |                              |
+        |               |                        |                              |
+        |               | MetricReader.Collect() |                              |
+        |               +------------------------>                              |
+        |                                        |                              |
+        | invoke callbacks from async instuments |                              |
+        <----------------------------------------+                              |
+        |                                        |                              |
+        | MetricReader.OnCollect(batch) callback |                              |
+        +---------------------------------------->                              |
+        |                                        |                              |
+        |                                        | MetricExporter.Export(batch) |
+        |                                        +------------------------------>
+        |                                        |                              |
+        |                                        |                              | send metrics
+        |                                        |                              +-------------->
+        |                                        |                              |
+```
+
 Metric Exporter has access to the [pre-aggregated metrics
 data](./datamodel.md#timeseries-model).
 
@@ -692,13 +732,6 @@ example:
 * Exporter C is a pull exporter which reacts to a scraper over HTTP.
 * Exporter D is a pull exporter which reacts to another scraper over a named
   pipe.
-
-### Push Metric Exporter
-
-Push Metric Exporter sends the data on its own schedule. Here are some examples:
-
-* Sends the data based on a user configured schedule, e.g. every 1 minute.
-* Sends the data when there is a severe error.
 
 #### Interface Definition
 
@@ -770,11 +803,44 @@ return a Failure result.
 and the destination is unavailable). OpenTelemetry client authors can decide if
 they want to make the shutdown timeout configurable.
 
+### Push Metric Exporter
+
+Push Metric Exporter sends the data on its own schedule. Here are some examples:
+
+* Sends the data based on a user configured schedule, e.g. every 1 minute.
+* Sends the data when there is a severe error.
+
 ### Pull Metric Exporter
 
 Pull Metric Exporter reacts to the metrics scrapers and reports the data
 passively. This pattern has been widely adopted by
 [Prometheus](https://prometheus.io/).
+
+```text
++---------------+                         +--------------+          +-----------------------+
+|               |                         |              |          |                       |
+| MeterProvider |                         | MetricReader |          | MetricExporter (pull) |
+|               |                         |              |          |                       |
++---------------+                         +--------------+          +-----------------------+
+        |                                        |                              |          +---------+
+        |                                        |                              | scraping |         |
+        |                                        |                              <----------+ Scraper |
+        |                                        | MetricReader.Collect()       |          |         |
+        |                                        <------------------------------+          +---------+
+        |                                        |                              |
+        | invoke callbacks from async instuments |                              |
+        <----------------------------------------+                              |
+        |                                        |                              |
+        | MetricReader.OnCollect(batch) callback |                              |
+        +---------------------------------------->                              |
+        |                                        |                              |
+        |                                        | MetricExporter.Export(batch) |
+        |                                        +------------------------------>
+        |                                        |                              |
+        |                                        |                              | respond to scraper
+        |                                        |                              +-------------------->
+        |                                        |                              |
+```
 
 ## Defaults and Configuration
 
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -539,7 +539,7 @@ measurements using the equivalent of the following naive algorithm:
 
 The SDK MUST support multiple `MetricReader` instances to be registered on the
 same `MeterProvider`, and the [MetricReader.Collect](#collect) invocation on one
-`MetricReader` instance MUST NOT introduce side-effects to other `MetricReader`
+`MetricReader` instance SHOULD NOT introduce side-effects to other `MetricReader`
 instances. For example, if a `MetricReader` instance is receiving metric data
 points that have [delta temporality](./datamodel.md#temporality), it is expected
 that SDK will update the time range - e.g. from (T<sub>n</sub>, T<sub>n+1</sub>]
@@ -577,9 +577,9 @@ functions will be triggered.
 failed or timed out.
 
 `Collect` does not have any required parameters, however, individual language
-clients MAY choose to support optional parameters (e.g. callback, filter,
-timeout). Individual language clients MAY choose the return value type, or do
-not return anything.
+clients MAY choose to add parameters (e.g. callback, filter, timeout).
+Individual language clients MAY choose the return value type, or do not return
+anything.
 
 ### Base exporting MetricReader
 
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -541,11 +541,10 @@ The SDK MUST support multiple `MetricReader` instances to be registered on the
 same `MeterProvider`, and the [MetricReader.Collect](#collect) invocation on one
 `MetricReader` instance MUST NOT introduce side-effects to other `MetricReader`
 instances. For example, if a `MetricReader` instance is receiving metric data
-points that have [delta temporality](./datamodel.md#temporality) via the
-[MetricReader.OnCollect](#oncollect) callback, it is expected that SDK will
-update the time range - e.g. from (T<sub>n</sub>, T<sub>n+1</sub>] to
-(T<sub>n+1</sub>, T<sub>n+2</sub>] - **ONLY** for this particular `MetricReader`
-instance, so the next `OnCollect` will get a new start time.
+points that have [delta temporality](./datamodel.md#temporality), it is expected
+that SDK will update the time range - e.g. from (T<sub>n</sub>, T<sub>n+1</sub>]
+to (T<sub>n+1</sub>, T<sub>n+2</sub>] - **ONLY** for this particular
+`MetricReader` instance.
 
 ```text
 +-----------------+            +--------------+
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -691,32 +691,32 @@ components in the SDK. Note that the _SIGNAL_ in the diagram could come from:
   `MeterProvider.ForceFlush()`).
 
 ```text
-+---------------+                         +--------------+             +----------------+
-|               |                         |              |             |                |
-| MeterProvider |                         | MetricReader |             | MetricExporter |
-|               |                         |              |             |                |
-+---------------+                         +--------------+             +----------------+
-        |           +--------+                   |                              |
-        |           |        |                   |                              |
-        |           | SIGNAL |                   |                              |
-        |           |        |                   |                              |
-        |           +---+----+                   |                              |
-        |               |                        |                              |
-        |               | MetricReader.Collect() |                              |
-        |               +------------------------>                              |
-        |                                        |                              |
-        | invoke callbacks from async instuments |                              |
-        <----------------------------------------+                              |
-        |                                        |                              |
-        | MetricReader.OnCollect(batch) callback |                              |
-        +---------------------------------------->                              |
-        |                                        |                              |
-        |                                        | MetricExporter.Export(batch) |
-        |                                        +------------------------------>
-        |                                        |                              |
-        |                                        |                              | send metrics
-        |                                        |                              +-------------->
-        |                                        |                              |
++---------------+                          +--------------+             +----------------+
+|               |                          |              |             |                |
+| MeterProvider |                          | MetricReader |             | MetricExporter |
+|               |                          |              |             |                |
++---------------+                          +--------------+             +----------------+
+        |           +--------+                    |                              |
+        |           |        |                    |                              |
+        |           | SIGNAL |                    |                              |
+        |           |        |                    |                              |
+        |           +---+----+                    |                              |
+        |               |                         |                              |
+        |               | MetricReader.Collect()  |                              |
+        |               +------------------------->                              |
+        |                                         |                              |
+        | invoke callbacks from async instruments |                              |
+        <-----------------------------------------+                              |
+        |                                         |                              |
+        | MetricReader.OnCollect(batch) callback  |                              |
+        +----------------------------------------->                              |
+        |                                         |                              |
+        |                                         | MetricExporter.Export(batch) |
+        |                                         +------------------------------>
+        |                                         |                              |
+        |                                         |                              | send metrics
+        |                                         |                              +-------------->
+        |                                         |                              |
 ```
 
 Metric Exporter has access to the [pre-aggregated metrics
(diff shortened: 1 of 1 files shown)


specification/metrics/sdk.md
```diff
@@ -411,6 +413,114 @@ active span](../trace/api.md#context-interaction)).
 +------------------+
 ```
 
+## MetricReader
```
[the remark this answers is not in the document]
reyang: I've updated the MeterProvider part to cover this, PTAL

specification/metrics/sdk.md
```diff
@@ -526,6 +528,117 @@ measurements using the equivalent of the following naive algorithm:
...
+`OnCollect` SHOULD never be called concurrently for the same `MetricReader`
+instance. `OnCollect` can be called again only after the current call returns.
+
+`OnCollect` MUST accept the following parameters:
+
+* A `batch` of metrics. The exact data type of the batch is language specific,
+  typically it is some kind of list, e.g. it could be an enumerator that travels
+  through the in-memory state.
+
+Individual language client MAY choose to add optional parameters (e.g. timeout).
```
jkwatson, line 584:
```suggestion
Individual language clients MAY choose to add optional parameters (e.g. timeout).
```

specification/metrics/sdk.md
```diff
@@ -526,6 +528,117 @@ measurements using the equivalent of the following naive algorithm:
...
+Configurable parameters:
+
+* `exporter` - the exporter where the metrics are sent to.
+
+If the configured exporter only supports [pull mode](#pull-metric-exporter):
+
+* [Collect](#collect) SHOULD only be called when the [Pull Metric
+  Exporter](#pull-metric-exporter) triggers the scraping, otherwise it SHOULD be
+  treated as an error.
+* Individual language client MAY decide if [OnForceFlush](#onforceflush) will
```
jkwatson, line 624:
```suggestion
* Individual language clients MAY decide if [OnForceFlush](#onforceflush) will
```

jkwatson:
I would personally find it very helpful if we included a sequence diagram for both the pull & push cases in this documentation of how this should work. I'm having trouble wrapping my head around it just looking at the words here (I'm a visual learner). If we had sequence diagrams, I would have a much better idea of what is being presented here.

reyang:
> I would personally find it very helpful if we included a sequence diagram for both the pull & push cases in this documentation of how this should work. I'm having trouble wrapping my head around it just looking at the words here (I'm a visual learner). If we had sequence diagrams, I would have a much better idea of what is being presented here.

I've added the sequence diagrams, got a gut feeling that this PR is becoming too big 🧙‍♂️. Let's discuss more during the upcoming SIG meeting.

specification/metrics/sdk.md
```diff
@@ -526,6 +529,117 @@ measurements using the equivalent of the following naive algorithm:
...
+[OnCollect](#oncollect) is called. and passes the metrics to the configured
+[MetricExporter](#metricexporter).
+
+Configurable parameters:
+
+* `exporter` - the exporter where the metrics are sent to.
+
+If the configured exporter only supports [pull mode](#pull-metric-exporter):
+
+* [Collect](#collect) SHOULD only be called when the [Pull Metric
```
jsuereth, line 622: This seems like this is a method on the *returned* object/structure from registering a reader.

Specifically, It seems like you're mixing TWO interface with different purposes here.  

1. An interface that *users will extend* to define a metric exporter (On* methods)
2. AN interface that the *sdk will provide* to interact with metric collection (the non-ON* metthods)

Why not explicitly pull these apart like so:

```
interface MetricReader {
  Status onCollect(MetricBatch metrics);
    Status onForceFlush();
    Status onShutdown();
}
interface RegisteredMetricReader {
  Status collect();
  Status forceFlush();
  Status shutdown();
}

interface MeterProvider {
  RegisteredMetricReader register(MetricReader reader);
}
```

In this case, Periodic exporter can easily be implemented.  I'm not sure why BASE exists except perhaps to provide implementations for collect/forceFlush + shutdown.

specification/metrics/sdk.md
```diff
@@ -411,6 +413,114 @@ active span](../trace/api.md#context-interaction)).
...
++-----------------+            +--------------+
+```
+
+### MetricReader operations
+
+#### Collect
+
+Collects the metrics from the SDK. If there are [asynchronous
+Instruments](./api.md#asynchronous-instrument) involved, their callback
+functions will be triggered. Then the [OnCollect callback](#oncollect) will be
```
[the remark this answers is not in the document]
jsuereth: Where is the description?  I'm still not sure this is clear.

For example, somewhere it should *explicitly* denote where delta time intervals are reset.  If it's per-metric-reader, then the collect method should denote that it resets metric reporting intervals (e.g. setting the next start time for delta metrics).

specification/metrics/sdk.md
```diff
@@ -526,6 +529,117 @@ measurements using the equivalent of the following naive algorithm:
...
+called with the collected metrics from the in-memory state.
+
+`Collect` SHOULD provide a way to let the caller know whether it succeeded,
```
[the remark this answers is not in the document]
reyang: > So this means "collect" blocks until all metrics are collected?

I think in most implementations the answer is YES.

> Is it expected that users override this method in some fashion?

Most cases NO.

> Does it make MORE sense to have the MeterProvider have a collect(MeterReader) method?

I think we already have `MeterProvider.ForceFlush()` which calls the `MeterReader.Collect()` on ALL registered `MeterReader` instances.
Regarding "why it is called MeterProvider.ForceFlush vs. MeterProvider.Collect?":
* Tracer provider also has ForceFlush, I think it is easier for the user to understand (e.g. I expect 99% of the user won't need to know the MetricReader interface), and it is more consistent naming.
* We might want to add functionality to "Collect" (e.g. to indicate whether we only want to trigger the async callbacks in case "the collection cycle is different than the exporting cycle" #1432). This makes more sense for individual MetricReader and makes less sense of MeterProvider.

specification/metrics/sdk.md
reyang: Yes you're right that these are not orthogonal at all.

My thinking is that the spec describes what the requirement, rather than "how a language client can meet that requirement". So in languages which supports multi-inheritance folks can have something like istream/ostream/iostream, but other language clients do not have to so.

Do you see the current approach "preventing" Java from doing what you've described above? I want to make sure that we give such flexibility from the spec (instead of enforcing the interface/class hierarchy).

specification/metrics/sdk.md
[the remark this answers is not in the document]
jkwatson: if we allowed `collect` to return a Future, or language-equivalent, would that satisfy the requirement? 
How about if `collect` took the callback as a parameter, rather than having it separated into two separate things?

specification/metrics/sdk.md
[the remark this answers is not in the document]
reyang: Here goes the new update https://github.com/open-telemetry/opentelemetry-specification/pull/1888/commits/fd9dfc45c33c0895c0d9d7f62d767d9ed4ce6323.

specification/metrics/sdk.md
jsuereth: Yes, I think we'd need to find sufficient reason to deviate from the exact specification of methods you propose.  I'm not happy with how it exposes (through inheritance) core SDK functionality in practice.

Let me prototype some options and see if we can rephrase this in a way that satisfies each SDK better

jmacd:
I have a few general questions. 

For the asynchronous instruments, when we have multiple Readers registered, do we want the asynchronous callbacks to be executed on a per-Reader basis? I would've assumed no, but this PR suggests the opposite. Is there a reason not to share observations across Readers?

fd9dfc4 and #1432 both discuss sequences of observations, where a Reader might want to collect synchronous and asynchronous instruments more frequently than the export interval. I want to avoid requiring the SDK to maintain separate state about each Reader. I think we should somehow state that every instrument, though it may have its own observation/collection frequency, is producing measurements at a single frequency. If there is an arrangement of exporters that wish to downsample from this base frequency, they would be able to do so using temporal aggregation, meaning to store and aggregate metric data between export intervals. This should be do-able by re-using existing Aggregators and other parts of the SDK.

Here's a test to see how you feel about this. If both an OTLP exporter and a Prometheus exporter are configured, and both are configured for cumulative temporality. The OTLP exporter is configured with a 30s export interval. Prometheus collects data on demand, so the interval is unknown. In this case, I would be happy if the Prometheus collect() always received the last result that was pushed by the OTLP exporter. I would not want the Prometheus collect() to separately evaluate the async instruments. I am willing to accept staleness of 30s or less when using either exporter.

Another user might not agree with this, and could say that Prometheus cannot export data less than 5 seconds old. In that case, I would configure a minimum interval of 5s for the instruments and the OTLP exporter would then be required to aggregate six points down to one for a 30s export interval. For cumulative this is as simple as dropping five points out of six. (For delta, this means allocating a new aggregator to sum the six points.) I believe I've described how to satisfy #1432 cc: @Oberon00

reyang:
> For the asynchronous instruments, when we have multiple Readers registered, do we want the asynchronous callbacks to be executed on a per-Reader basis? I would've assumed no, but this PR suggests the opposite. Is there a reason not to share observations across Readers?

I hope that the spec will give certain flexibility to SDKs so they can make their own conscious decision. There are scenarios people would want to pay the cpu/memory cost to get more updated data, and there are also scenarios people are more immune to data that are slightly outdated.
 
> [fd9dfc4](https://github.com/open-telemetry/opentelemetry-specification/commit/fd9dfc45c33c0895c0d9d7f62d767d9ed4ce6323) and #1432 both discuss sequences of observations, where a Reader might want to collect synchronous and asynchronous instruments more frequently than the export interval. I want to avoid requiring the SDK to maintain separate state about each Reader. I think we should somehow state that every instrument, though it may have its own observation/collection frequency, is producing measurements at a single frequency. If there is an arrangement of exporters that wish to downsample from this base frequency, they would be able to do so using temporal aggregation, meaning to store and aggregate metric data between export intervals. This should be do-able by re-using existing Aggregators and other parts of the SDK.

Do you want to require the SDK to avoid maintaining separate state for each Reader? Or you think it is fine to have the spec giving some flexibility so SDKs can decide if they want to maintain separate state or not? From what you've described here I feel that you might be okay with the latter?

> Here's a test to see how you feel about this. If both an OTLP exporter and a Prometheus exporter are configured, and both are configured for cumulative temporality. The OTLP exporter is configured with a 30s export interval. Prometheus collects data on demand, so the interval is unknown. In this case, I would be happy if the Prometheus collect() always received the last result that was pushed by the OTLP exporter. I would not want the Prometheus collect() to separately evaluate the async instruments. I am willing to accept staleness of 30s or less when using either exporter.
> 
> Another user might not agree with this, and could say that Prometheus cannot export data less than 5 seconds old. In that case, I would configure a minimum interval of 5s for the instruments and the OTLP exporter would then be required to aggregate six points down to one for a 30s export interval. For cumulative this is as simple as dropping five points out of six. (For delta, this means allocating a new aggregator to sum the six points.) I believe I've described how to satisfy #1432 cc: @Oberon00

I have trouble with two pull exporters (although I think it might not be a super common use case), how do we possibly know the maximum common divisor? I think it is similar to the scenario you mentioned during the SIG meeting (the throttle mechanism in Golang SDK) - both exporters would fetch the values from async instruments (and the SDK might want to put some logic to avoid reentrancy), and if any of them realized "someone already fetched the snapshot 3 seconds ago" or "someone is already asking for a snapshot", they could decide to wait/reuse that data rather than asking for a brand new version.

specification/metrics/sdk.md
reyang: I don't like the "BASE" name, it feels that the spec is suggesting some type of class inheritance (and it doesn't make sense for languages who don't believe in OOP). Definitely want to rephrase it in a better way (@jkwatson has been helping me on the English and wording part) 🧙

jmacd:
In the case of two pull exporters, I imagine one of them requesting a period of 3s and one of them requesting a period of 5s. In this situation, I would force the 5s interval to accept up to 2s of staleness. We could also call this a configuration error -- in which case all readers should use integer factors of the minimum period, however that's configured.

specification/metrics/sdk.md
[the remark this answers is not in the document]
reyang: We've discussed more during the [9/2/2021 Metrics SIG Meeting](https://docs.google.com/document/d/1pdvPeKjA8v8w_fGKAN68JjWBmVJtPCpqdi9IZrd6eEo/edit#).

One example:
* Java might prefer the multi-inheritance approach as @jsuereth mentioned [here](https://github.com/open-telemetry/opentelemetry-specification/pull/1888#discussion_r699237840), and using different interfaces for push and pull like @jkwatson mentioned [here](https://github.com/open-telemetry/opentelemetry-specification/pull/1888#discussion_r699523510).
* If the spec requires Java to unify the push/pull exporter, it might end up being very awkward/dirty.


The consensus:
* Many languages might prefer their own idiomatic way of implementing the exporter. We should loosen the exporter interface requirement and give certain flexibility.
* @reyang to work with @jsuereth to get some alignment (at least making sure @jsuereth is aware of this and won't waste time prototyping something that is against the direction)
* Once aligned, @reyang to update the PR. The sequence diagrams will be no longer needed so we'll remove them.

specification/metrics/sdk.md
reyang: This is being addressed by https://github.com/open-telemetry/opentelemetry-specification/pull/1888#discussion_r701496630.

jmacd:
As discussed in the SIG call today, I've come around to supporting this PR and the implication that each exporter will independently call the asynchronous instruments. We discussed how the API may be not idiomatic in every language, and that perhaps we do not need to prescribe the API mechanism for the Reader. In some environments, push exporters and pull exporter interfaces to the Reader may look different from each other.

jmacd (approved):

specification/metrics/sdk.md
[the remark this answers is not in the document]
reyang: Chatted with @jsuereth offline and he is fine with the direction. I'll update the PR soon.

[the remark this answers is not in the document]
reyang: PR updated, PTAL.

specification/metrics/sdk.md
reyang: I've updated the PR, PTAL.

specification/metrics/sdk.md
```diff
@@ -185,26 +188,26 @@ meter_provider
         "Bar",
         instrument_name="Y",
         aggregation=HistogramAggregation(buckets=[5.0, 10.0, 25.0, 50.0, 100.0]))
-    .set_exporter(PrometheusExporter())
+    .add_metric_reader(BaseExportingMetricReader(PrometheusExporter()))
```
jsuereth, line 191: As discussed, I'm not sure the notion of `BaseExportingMetricReader` is providing enough value of just the `MetricReader` interface you have spec'd below.

specification/metrics/sdk.md
```diff
@@ -540,17 +630,29 @@ The following diagram shows `MetricExporter`'s relationship to other components
...
+                               |  |                       |  |
+                               |  | MetricExporter (push) +-----> Another process
+                               |  |                       |  |
+                               |  +-----------------------+  |
+                               |                             |
+                               +-----------------------------+
+
++-----------------+            +-----------------------------+
+|                 | Metrics... |                             |
+| In-memory state +------------> Base exporting MetricReader |
```
jsuereth, line 647: We discussed this case offline.   If we want all exporters to share an interface, then basically we should ONLY expose push-exporters and an adapter for pull.  We aren't really providing a pull-based interface here, only an adapter against push.

I think what you have spec'd for MetricReader gives us first-class pull support where this version doesn't really make a lot of sense anymore.

specification/metrics/sdk.md
```diff
@@ -650,6 +752,11 @@ Pull Metric Exporter reacts to the metrics scrapers and reports the data
 passively. This pattern has been widely adopted by
 [Prometheus](https://prometheus.io/).
 
+Implementors MAY choose the best idiomatic design for their language. For
```
jsuereth, line 755: I actually don't think you need to define "Pull Metric Exporter" at all. 

If it's using the same interface as PushMetricExporter, but has a "controlling" reader to force a collect, that's just using the existing APIs for push exporters in a clever way vs. actually supporting pull first class.    That kind of derived usage doesn't necessarily need to be specified now, we should see if it's common across SDKs (especially if we ARE exposing first-class pull support via Reader).

specification/metrics/sdk.md
```diff
@@ -526,6 +529,93 @@ measurements using the equivalent of the following naive algorithm:
...
+
+`Collect` SHOULD provide a way to let the caller know whether it succeeded,
+failed or timed out.
+
+`Collect` does not have any required parameters, however, individual language
+clients MAY choose to add parameters (e.g. callback, filter, timeout).
+Individual language clients MAY choose the return value type, or do not return
+anything.
+
+### Base exporting MetricReader
```
jsuereth, line 584: We discussed offline that this section isn't necessarily providing anything over just the MetricReader interface.

I *do* think you might want to consider if the PrometheusExporter should be specified as a MetricReader *or* MetricExporter, or if you're leaving that up to SDKs.   IIUC from our discussion, you're planning to leave that up to SDKs right now.

jsuereth:
Reiley and I discussed offline.  I added the most important topics (for reference to everyone) and call out my last bits before approving myself:

1. Remove BaseMetricReader as a concept unless it provides behavior beyond the already specified MetricReader interface.
2. Remove Pull MetricExporter from the specification, or clearly denote it's an optional thing you could do, as MetricReader provides the first-class pull exporter support.

Optional;  If we could specify more clearly what "collect" looks like to a Reader so I can leverage that in the OpenCensus <-> OTel Bridge, that'd be ideal, but can be dealt with in follow up.

reyang:
> 2. Remove Pull MetricExporter from the specification, or clearly denote it's an optional thing you could do, as MetricReader provides the first-class pull exporter support.

I've adjusted the wording. PTAL.
I do think it is important to keep a section for "Pull Metric Exporter" so other places can link to it (e.g. when we talk about the push vs. pull model in concrete exporter requirement).

specification/metrics/sdk.md
reyang: I've updated the wording for pull exporter (e.g. PrometheusExporter) in [this commit](https://github.com/open-telemetry/opentelemetry-specification/pull/1888/commits/47f74e98bb00a17efa6ea51302f13a076a426e5a).

specification/metrics/sdk.md
reyang: Trying to address it here https://github.com/open-telemetry/opentelemetry-specification/pull/1888#issuecomment-914911739.

specification/metrics/sdk.md
reyang: `BaseExportingMetricReader` takes an exporter as the input, while MetricReader doesn't take it.
I think there might be a way to reword it so we give some flexibility here, I'll pick your brain offline and report back here.

specification/metrics/sdk.md
```diff
@@ -540,17 +630,29 @@ The following diagram shows `MetricExporter`'s relationship to other components
...
-+-----------------+            +-----------------------+
-
-+-----------------+            +-----------------------+
-|                 | Metrics... |                       |
-| In-memory state +------------> MetricExporter (pull) +--> Another process (scraper)
-|                 |            |                       |
-+-----------------+            +-----------------------+
++-----------------+            +-----------------------------+
+|                 | Metrics... |                             |
+| In-memory state +------------> Base exporting MetricReader |
```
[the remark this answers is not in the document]
reyang: For example, I have a Console Exporter, and I don't want it to export data periodically, what I want is "only prints data to the stdout when I press the ENTER key".
If we force people to use `PeriodicMetricReader`, it could still work and I guess the workaround is to put the exporter interval to 1 million years.
We might also step back and say "it seems to be an over design" and I'm fine to compromise.

specification/metrics/sdk.md
jsuereth: To record what I said previously:

- If you're using `MetricExporter` for pull-based export, do you need the metric reader interface? 
- `MetricReader` as "Pull" API and `MetricExporter` as "Push" API seems like two good things to specify.   The unification of just `MetricExporter` on top of `MetricReader` should be either "optional" for SDKs, or dropped entirely and SDKs can implement such unification if they want it (but does not need to be specified only discussed).   

Right now there's *no* language requiring either `BaseMetricReader` or `PeriodicMetricreader`, but I'd argue the existence in the specification implies they're both required (without language saying otherwise).

My suggestion (at a minimum) would be to declare that `BaseMetricReader` MAY be provided, but `PeriodicMetricReader` MUST be provided.

specification/metrics/sdk.md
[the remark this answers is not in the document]
jsuereth: Rewording here fits the implementation I have.  I fear we've made it harder to understand how to implement the specification BUT we have an implementable specification :).  LGTM for the first prototype.

specification/metrics/sdk.md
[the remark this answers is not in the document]
jsuereth: For your ConsoleExporter example (and any Push -> Pull adaptation), I'd argue what you want (instead of MetricReader) is to specify a helper/adapter for MetricExporter that stores metrics in-memory (performing additional aggregations every export interval) and when you press enter you get the data from the LAST export period.

I believe this is one of the ways the opentelemetry collecter can export to prometheus (via the web front-end).

specification/metrics/sdk.md
jsuereth: Fundamentally there's a few open questions here that I don't think are addressed in the verbage:

- are you specifying that `MetricExporter`, when it's a "Pull" exporter *MUST* be able to call some kind of "go collect things now"?   IIUC BaseMetricReader offers that capability to `MetricExporter` so you could have a Push->Pull adapted `MetricExporter` and manually call `collect`.  I'm against *forcing* this on SDKs via the specification.   If Push->Pull adaptation must go through interval metric reader, I think we have a much simpler system to maintain.
- Will we require PrometheusExporter to extend the `MetricExporter` interface in the specification?  I assume this will be answered in a follow up PR.  I think forcing that will limit the utility of `MetricReader` for pull-export.
- Is `Base exporting MetricReader` required for SDKs to implement?

specification/metrics/sdk.md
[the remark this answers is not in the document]
reyang: I've updated the diagram to use `periodic exporting MetricReader`.
I've moved the pull exporter examples and diagram to the pull exporter section and clarified that these are just examples so folks _can_ do, but it is _not required_.

specification/metrics/sdk.md
```diff
@@ -526,6 +528,72 @@ measurements using the equivalent of the following naive algorithm:
...
+
+This is an implementation of the `MetricReader` which collects metrics based on
+a user-configurable time interval, and passes the metrics to the configured
+[Push Metric Exporter](#push-metric-exporter).
+
+Configurable parameters:
+
+* `exporter` - the push exporter where the metrics are sent to.
+* `exportIntervalMillis` - the time interval in milliseconds between two
+  consecutive exports. The default value is 5000 (milliseconds).
```
jsuereth, line 593: Nit:  This seems pretty aggressive.   We see ~1 minute intervals being common for metric reporting in clients.   (Prometheus [defaults to 1 minute](https://prometheus.io/docs/prometheus/latest/configuration/configuration/)).

GCP itself actually prevents anything [faster than 10 seconds](https://cloud.google.com/monitoring/quotas) to prevent abuse.   5 seconds seems like it would flood many metric systems.

jsuereth (approved):
One blocking comment (the default export interval) but others this looks good to go now.

specification/metrics/sdk.md
reyang: I've updated the PR https://github.com/open-telemetry/opentelemetry-specification/pull/1888/commits/68f4fdc1156ae3a8eed3332006040b8648db32e1.

For folks who need more context here goes the discussion https://github.com/open-telemetry/opentelemetry-java/pull/3578#discussion_r703620370.

specification/metrics/sdk.md
reyang: Addressed in https://github.com/open-telemetry/opentelemetry-specification/pull/1888/commits/7977fbfe721a8e5681d18b15da8fc7e8a145ea80.

specification/metrics/sdk.md
```diff
@@ -411,6 +413,95 @@ active span](../trace/api.md#context-interaction)).
...
++-----------------+            +--------------+
+```
+
+### MetricReader operations
+
+#### Collect
+
+Collects the metrics from the SDK. If there are [asynchronous
+Instruments](./api.md#asynchronous-instrument) involved, their callback
+functions will be triggered. Then the [OnCollect callback](#oncollect) will be
```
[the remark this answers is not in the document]
jmacd: I agree we should cut this scope for an after-GA time. Should we start using release:after-ga for issues like #1432?

specification/metrics/sdk.md
```diff
@@ -526,6 +529,93 @@ measurements using the equivalent of the following naive algorithm:
...
+|                 |            |              |
++-----------------+            +--------------+
+```
+
+The SDK SHOULD provide a way to allow `MetricReader` to respond to
+[MeterProvider.ForceFlush](#forceflush) and [MeterProvider.Shutdown](#shutdown).
+Individual language clients can decide the language idiomatic approach, for
+example, as `OnForceFlush` and `OnShutdown` callback functions.
+
+### MetricReader operations
```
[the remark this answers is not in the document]
jmacd: This makes sense to me, especially having seen this pattern applied to the OTel-Go SDK and its OpenCensus bridge.

specification/metrics/sdk.md
jmacd: the OTel-Go SDK uses 10s, and the Prometheus server default is 60s. I'm not sure how this should be decided --

jmacd (approved):

specification/metrics/sdk.md
```diff
@@ -650,6 +718,46 @@ Pull Metric Exporter reacts to the metrics scrapers and reports the data
...
+  +-----------------+            |                             |
+                                 +-----------------------------+
+  ```
+
+* Use the same MetricExporter design for both push and pull exporters
+
+  ```text
+  +-----------------+            +-----------------------------+
+  |                 | Metrics... |                             |
+  | In-memory state +------------> Base exporting MetricReader |
```
cijothomas, line 750:
```suggestion
  | In-memory state +------------> MetricReader |
```

specification/metrics/sdk.md
```diff
@@ -526,6 +528,72 @@ measurements using the equivalent of the following naive algorithm:
...
+#### Collect
+
+Collects the metrics from the SDK. If there are [asynchronous
+Instruments](./api.md#asynchronous-instrument) involved, their callback
+functions will be triggered.
+
+`Collect` SHOULD provide a way to let the caller know whether it succeeded,
+failed or timed out.
+
+`Collect` does not have any required parameters, however, individual language
```
cijothomas, line 578: do we need some language like 
"If the MetricReader is used for pull-based Exporters, then Collect() should pass the metrics to the configured exporter" ?

reyang: I think the general spirit is to give flexibility, so each language can find their optimal way of achieving the goal. They will need to support Prometheus exporter anyways, so whether we have a `SHOULD` here or not won't change the fact that they will need to do something to make it work, and we probably don't need to force them here.

specification/metrics/sdk.md
reyang: This was intentional, I think having a MetricReader here is wrong since MetricReader itself doesn't know how to take an exporter. This is just an example of how people could model it so I think it's fine that we don't spec out the base exporting MetricReader.

reyang: Refer to the discussion here https://github.com/open-telemetry/opentelemetry-specification/pull/1888#discussion_r704053279.

cijothomas: I see. The term "base exporting" (which is never described before) could be confusing as well. How about simply "Exporting MetricReader"

cijothomas (approved):
Left a comment (about naming suggestion)

jkwatson (approved):
Thanks!

specification/metrics/sdk.md
jmacd: @reyang will apply in a follow-on PR

reyang: Being addressed in #1917.

PR #1890, opened 2021-08-26
Clarify some cases for http.host and http.client_ip.

### LINKED ISSUE #1845: Clarification on when to capture http.client_ip

Clarification whether or not to populate `http.client_ip` with the remote address (e.g. same value as `net.peer.ip`) when there is no `Forwarded` or `X-Forwarded-By` header.

**Additional context.**

The current definition of `http.client_ip` in the [HTTP Server semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-server-semantic-conventions):

> The IP address of the original client behind all proxies, if known (e.g. from X-Forwarded-For). [2]
>
> [2]: This is not necessarily the same as net.peer.ip, which would identify the network-level peer, which may be a proxy.

Oberon00:
Fixes #1845.

## Changes

* Clarify when & why to set http.client_ip (#1845).
* Clarify that an empty string for http.host is valid and in fact recommended if there actually is an empty host header.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@ release.
 - Add `messaging.consumer_id` to differentiate between message consumers.
   ([#1810](https://github.com/open-telemetry/opentelemetry-specification/pull/1810))
 - Clarifications for `http.client_ip` and `http.host`.
-  ([#????](https://github.com/open-telemetry/opentelemetry-specification/pull/????))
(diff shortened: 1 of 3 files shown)


iNikem:
@open-telemetry/technical-committee Anything preventing this PR from merging?

PR #1892, opened 2021-08-27
Exempt resources from attribute limits

owais:
Resources are not susceptible to scenarios where excessive attributes
can be recorded unlike Spans. Resources are also immutable and it can be
hard for some SDKs to apply the limits at source at the time the
attributes are added to a resource. Also Limits and Resources
both are generally defined and passed on to a TracerProvider which
forces a TracerProvider to either mutate the resource or generate a new
one with duplicate attributes in order to apply the limits to it.

I propose we explicitly exempt Resources from attribute limits.

## Changes

Exempt Resources from attribute limits.

### DIFF
--- a/specification/common/common.md
+++ b/specification/common/common.md
@@ -89,6 +89,14 @@ general limit.
 
 #### Exempt Entities
 
+Resource attributes SHOULD be exempt from the limits described above as resources
+are not susceptible to the scenarios (auto-instrumentation) that result in
+excessive attributes count or size. Resources are also sent only once per batch
+instead of per span so it is relatively cheaper to have more/larger attributes
+on them. Resources are also immutable by design and they are generally passed
+down to TracerProvider along with limits. This makes it awkward to implement
+attribute limits for Resources.
+
 Attributes, which belong to Metrics, are exempt from the limits described above
 at this time, as discussed in
 [Metrics Attribute Limits](../metrics/sdk.md#attribute-limits).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ release.
 
 ### Resource
 
+- Exempt Resource from attribute limits.
+  ([#1892](https://github.com/open-telemetry/opentelemetry-specification/pull/1892))
+
 ### Semantic Conventions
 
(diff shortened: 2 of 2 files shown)


carlosalberto (approved):

carlosalberto:
Ping @iNikem @open-telemetry/specs-metrics-approvers

jsuereth:
Related issue for metrics: https://github.com/open-telemetry/opentelemetry-specification/issues/1782

iNikem:
Spec modification as proposed by this PR nicely described why resource attributes CAN be excluded from attribute limits. But it does not explain why they SHOULD or why do we want to have such an exception. PR description has some justification:

>  Also Limits and Resources both are generally defined and passed on to a TracerProvider which forces a TracerProvider to either mutate the resource or generate a new one with duplicate attributes in order to apply the limits to it

Are current attribute limits specific/scoped into `TracerProvider`?

bogdandrutu (approved):

specification/common/common.md
```diff
@@ -89,6 +89,12 @@ general limit.
 
 #### Exempt Entities
 
+Resource attributes are exempt from the limits described above as resources
```
Oberon00, line 92: I think we should use clear language here. I.e. instead of saying "Resource attributes *are* exempt", we should use "Resource attributes *MAY/SHOULD/MUST be* exempt".

owais: I tried to follow the same languages as metrics below but I agree MAY/SHOULD/MUST makes more sense since metrics case is a temporary exemption till we figure them out.

owais:
> Are current attribute limits specific/scoped into TracerProvider?

@iNikem  Yes, I can confirm at least Java, Python, Go and Node/JS do this.

iNikem (approved):
@owais I still think we should add some justification about "Why" from PR description into the text added

owais:
@iNikem added.

carlosalberto:
Ping @jsuereth

tigrannajaryan (approved):

tigrannajaryan:
Reviewers, please respond to unresolved comments. If you want to block merging then request changes, otherwise I am going to merge this since it has the required number of approvals.

tigrannajaryan:
No further comments, merging.

PR #1893, opened 2021-08-27
Prefer global limit over model-specific default

### LINKED ISSUE #1878: Model specific limits should attempt to use global limits before falling back o defaults

We have two types of limits today. Global limits and model specific limits. Here is what the spec has to say about the case where both global and model specific versions of the same limit are defined:

> An SDK MAY implement model-specific limits, for example SpanAttributeCountLimit. If both a general and a model-specific limit are implemented, then the SDK MUST first attempt to use the model-specific limit, if it isn't set and doesn't have a default, then the SDK MUST attempt to use the general limit.

List of [global limits](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#attribute-limits) and [model specific limits](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#span-limits-).

Model specific limits `OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT`, `OTEL_EVENT_ATTRIBUTE_COUNT_LIMIT` and `OTEL_LINK_ATTRIBUTE_COUNT_LIMIT` all have default values of 128. This means if an SDK implements the model specific limits then`OTEL_ATTRIBUTE_COUNT_LIMIT` will always be a noop. I think this can be counter intuitive to users. As a user, if I don't specify any model specific limit and only a single global limit, I'd expect all models to respect the global limit. This would also allow users to specify common limits using the global limit option and then override just one of the model specific limits. For example, if I want to limit link attribute count to 16 but everything else to 32, today I have to specify the following env vars:

```
OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT=32
OTEL_LINK_ATTRIBUTE_COUNT_LIMIT=32
OTEL_EVENT_ATTRIBUTE_COUNT_LIMIT=16
```

Specifying the global OTEL_ATTRIBUTE_COUNT_LIMIT would have no effect here as all of the above have default values (128) which would be used in case these env vars were not specified.

I think it'd be far more intuitive if users could set the following values to achieve the same effect.

```
OTEL_ATTRIBUTE_COUNT_LIMIT=32
OTEL_EVENT_ATTRIBUTE_COUNT_LIMIT=16
```

-------

I propose the SDK should use the following order to load model specific limits:

1. Use the model specific limit if set.
2. else use the corresponding global limit if set. 
3. else use the default value for model specific limit (if any)
4. else use the default value for the corresponding global limit.

owais:
Fixes #1878 

## Changes

Prefer global limit over model-specific limit default when model-specific limit is not set.

### DIFF
--- a/specification/common/common.md
+++ b/specification/common/common.md
@@ -79,7 +79,7 @@ An SDK MAY implement model-specific limits, for example
 `SpanAttributeCountLimit`. If both a general and a model-specific limit are
 implemented, then the SDK MUST first attempt to use the model-specific limit, if
 it isn't set, then the SDK MUST attempt to use the general limit. If neither are
-defined, then the SDK must try to use the model-specific limit default value,
+defined, then the SDK MUST try to use the model-specific limit default value,
 followed by the global limit default value.
 
 <a name="attribute-limits-configuration"></a>
--- a/specification/common/common.md
+++ b/specification/common/common.md
@@ -78,8 +78,9 @@ in the list below.
 An SDK MAY implement model-specific limits, for example
 `SpanAttributeCountLimit`. If both a general and a model-specific limit are
 implemented, then the SDK MUST first attempt to use the model-specific limit, if
-it isn't set and doesn't have a default, then the SDK MUST attempt to use the
-general limit.
+it isn't set, then the SDK MUST attempt to use the general limit. If neither are
+defined, then the SDK must try to use the model-specific limit default value,
+followed by the global limit default value.
 
 <a name="attribute-limits-configuration"></a>
 **Configurable parameters:**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ release.
 
 ### Traces
 
+- Prefer global user defined limits over model-sepcific default values.
+  ([#1893](https://github.com/open-telemetry/opentelemetry-specification/pull/1893))
 - Add InstrumentationLibrary to Sampler.ShouldSample.
   ([#1850](https://github.com/open-telemetry/opentelemetry-specification/pull/1850))
 


jtmal-signalfx (approved):
I suppose the main question is who has already implemented this. It might be easy to miss by SDK implementors.

Otherwise I think that `model-explicit > global-explicit > model-default > global-default` sounds good. Unless more people think that in general global limits are too dangerous, and we'd rather only have the model-specific limits.

owais:
We implemented this in the Python SDK and most people who reviewed it were surprised by the fact that it prefers model-specific default over user specified global.

jtmalinowski:
Right, so when I was writing that rule I considered both cases, and went model-specific defaults, so there's a smaller chance of a user shooting themselves in the foot, by setting something globally and not understanding consequences. But good to hear that it was tested in practice, and now we now which way is preferred.

specification/common/common.md
```diff
@@ -78,8 +78,9 @@ in the list below.
 An SDK MAY implement model-specific limits, for example
 `SpanAttributeCountLimit`. If both a general and a model-specific limit are
 implemented, then the SDK MUST first attempt to use the model-specific limit, if
-it isn't set and doesn't have a default, then the SDK MUST attempt to use the
-general limit.
+it isn't set, then the SDK MUST attempt to use the general limit. If neither are
+defined, then the SDK must try to use the model-specific limit default value,
```
iNikem, line 82:
```suggestion
defined, then the SDK MUST try to use the model-specific limit default value,
```

iNikem (approved):

jmacd (approved):

iNikem (approved):

jsuereth (approved):

carlosalberto (approved):

specification/common/common.md
```diff
@@ -78,8 +78,9 @@ in the list below.
 An SDK MAY implement model-specific limits, for example
 `SpanAttributeCountLimit`. If both a general and a model-specific limit are
 implemented, then the SDK MUST first attempt to use the model-specific limit, if
-it isn't set and doesn't have a default, then the SDK MUST attempt to use the
-general limit.
+it isn't set, then the SDK MUST attempt to use the general limit. If neither are
+defined, then the SDK MUST try to use the model-specific limit default value,
+followed by the global limit default value.
```
arminru, line 83: Please add an entry mentioning this clarification in `CHANGELOG.md` and we're good to go.

owais: Added. Thanks

PR #1895, opened 2021-08-27
OTLP Exporter: Make gzip compression the default

johnnyshields:
- In `specification/protocol/exporter.md`, make `gzip` compression the default.
- To allow disabling compression, I have added `null` as a supported value (currently the only way is to leave the env var blank)

AFAIK there is no practical use case where a user would not want to use gzip compression when exporting traces. It greatly reduces network load and data costs.

After approval, this change should be done in all the client libraries.


johnnyshields:
@arminru can you unmark this as stale? AFAIK this is pending review from OTEL core team.

specification/protocol/exporter.md
```diff
@@ -15,13 +15,14 @@ The following configuration options MUST be available to configure the OTLP expo
...
 | Certificate File     | The trusted certificate to use when verifying a server's TLS credentials. Should only be used for a secure connection. | n/a               | `OTEL_EXPORTER_OTLP_CERTIFICATE` `OTEL_EXPORTER_OTLP_TRACES_CERTIFICATE` `OTEL_EXPORTER_OTLP_METRICS_CERTIFICATE` |
 | Headers              | Key-value pairs to be used as headers associated with gRPC or HTTP requests. See [Specifying headers](./exporter.md#specifying-headers-via-environment-variables) for more details.                   | n/a               | `OTEL_EXPORTER_OTLP_HEADERS` `OTEL_EXPORTER_OTLP_TRACES_HEADERS` `OTEL_EXPORTER_OTLP_METRICS_HEADERS` |
-| Compression          | Compression key for supported compression types. Supported compression: `gzip`| No value              | `OTEL_EXPORTER_OTLP_COMPRESSION` `OTEL_EXPORTER_OTLP_TRACES_COMPRESSION` `OTEL_EXPORTER_OTLP_METRICS_COMPRESSION` |
+| Compression          | Compression key for supported compression types. Supported compression: `gzip`, `null` | `gzip`              | `OTEL_EXPORTER_OTLP_COMPRESSION` `OTEL_EXPORTER_OTLP_TRACES_COMPRESSION` `OTEL_EXPORTER_OTLP_METRICS_COMPRESSION` |
```
arminru, line 18: This sounds like a good idea to me and would've been a reasonable default but I'm afraid we likely can't introduce a breaking change like this at this point given that the spec here has been marked as stable already.

Currently I only see this being implemented in Go, Python, PHP and Ruby which have not been GA'ed yet, however.
@open-telemetry/go-maintainers @open-telemetry/php-maintainers @open-telemetry/python-maintainers @open-telemetry/ruby-maintainers: Would you still be able and willing to pick up this change if we would decide to accept it?

cc @open-telemetry/technical-committee

tigrannajaryan:
> AFAIK there is no practical use case where a user would not want to use gzip compression when exporting traces. It greatly reduces network load and data costs.

gzip compression is possibly pointless in localhost communication (e.g. between Otel SDK and Otel Collector running on localhost) and likely is just a CPU overhead. I did not test for this but this may be a practical use case which disproves the above statement.

yurishkuro:
> gzip compression is possibly pointless in localhost communication (e.g. between Otel SDK and Otel Collector running on localhost)

probably, but is the localhost-only setup a prevalent configuration or an edge case that can get by with proactively disabling gzip?

specification/protocol/exporter.md
Aneurysm9: In the maintainers' meeting this week we discussed this and there was not clarity regarding the impact of this change or what specific actions would be required of SDKs or the collector.  As Go is preparing to release `v1.0.0` this week I would like to avoid disruption of that release with work of unknown scope.

johnnyshields: Support for this default could be stated as "the default SHOULD be gzip (contingent on the SDK supporting gzip) otherwise the default in a given SDK MAY be null until such time the SDK implements gzip."

Re: specific actions, I would suggest:

1. The various collectors should confirm they support gzip. The vast majority of web frameworks handle gzip out-of-the-box; typically Nginx and Apache will pass-thru the gzipped body by default and the downstream web framework (Rails, Laravel, etc.) will decompress it. Perhaps you could mass-contact all collector owners give a 3 week window for anyone to state that they do **not** support gzip (the assumption should be that they **do**.)

2. Assuming step 1 is a non-issue, each SDK should implement the switch from null to gzip according to their own schedule. I would suggest it be done as a minor version bump as it is not expected to break any actual interface behavior.

johnnyshields:
> the localhost-only setup a prevalent configuration

It is most certainly an edge case; the standard usage is to send OTEL data directly to a third-party service such as DataDog or Lightstep.

Even in the case of local collectors, presumably these collectors forward requests to an external host, in which case it would probably still make sense to gzip your requests before they are forwarded. Nginx/Apache handle gzipped requests in this manner; they receive them as gzipped and forward as gzipped. (Note that gzip applies only to the request body and not to the HTTP headers.)

specification/protocol/exporter.md
carlosalberto: cc @open-telemetry/javascript-maintainers

Oberon00:
> It is most certainly an edge case; the standard usage is to send OTEL data directly to a third-party service such as DataDog or Lightstep.

AFAIK AWS Lambda is one setup where you typically send to a local collector.

fbogsany:
> Even in the case of local collectors, presumably these collectors forward requests to an external host, in which case it would probably still make sense to gzip your requests before they are forwarded.

This is not true for the OpenTelemetry Collector when used as an "agent" (e.g. as a sidecar or daemonset). The collector will decompress, deserialize, process (perhaps including batching), re-serialize, and re-compress before forwarding. In this scenario, compression in the app and decompression in the Collector are almost certainly needless overhead.

That said, at Shopify we ended up forwarding "as-is" before abandoning the Collector daemonset altogether.

johnnyshields:
@fbogsany there are two questions we should think about here:

1. Which is more common: using a local "collector" daemon, or directly forwarding OTEL to an external host? I would assert in 90%+ of cases it's the later, and that seems to be the pattern favored by the majority of Otel analytics vendors (DataDog, Lightstep, etc.)

2. What is the penalty for "getting it wrong"? If we compress when going to a local collector, it's a miniscule bit of CPU wasted. However, if we *don't* compress when going to an external sink, the user's network and data ingestion bills will be 10x higher.

anuraaga:
> and that seems to be the pattern favored by the majority of Otel analytics vendors (DataDog, Lightstep, etc.)

FWIW AWS encourages the sidecar collector to allow rich resource detection and attributes processing. My impression had been this is the common recommendation among vendors but could be wrong. But up until now when thinking of tradeoffs, my mental model has always considered sidecar to be a large majority.

Just an aside that it's always pretty annoying working with OTLP - the tree structure is harder to serialize by needing more state, and parsing with several loops every single time just to get some spans is tedious. It is some builtin in compression in the structs but if gziping anyways, which indeed will probably shrink it even more, then the tedium of the tree structure becomes quite disappointing.

tigrannajaryan:
I think this PR is superseded by https://github.com/open-telemetry/opentelemetry-specification/pull/1923 which allowed languages to make their own choice of default for compression.

Unless there is something remaining I think we can close this PR.

johnnyshields:
@tigrannajaryan, yes, there is. `gzip` should become the recommended default for the major language OTEL libraries like Ruby, Python, etc. It is fine to say that if a specific lib has a strong reason for not using `gzip` as default, that is fine, but the assumption should be that `gzip` is default.

The language is not currently clear in the spec, and will cause debate within each lib about what the default should be.

johnnyshields:
@yurishkuro no, #1923 did not explicitly state that gzip should be the default where possible. I am proposing to make the wording even more clear.

yurishkuro:
@johnnyshields I don't mind reopening, but I do not see a consensus in the discussion that gzip is the best choice in the majority of use cases.

jack-berg:
Not an approver, but on New Relic would prefer the default to be gzip.

yurishkuro:
> I see no need for this change. If language SDKs are already required to support gzip and can use it by default if it makes sense for their expected primary use cases then what do we gain by recommending a particular default?

@Aneurysm9 if we had data showing clear performance gains (not just in traffic but looking more comprehensively at other measures like CPU cost), then the benefit of centralized recommendation for a default is that it can be made once, instead of individual SIGs deciding the same thing over and over (and what are they basing their decision on?) However, this thread showed that there are many factors that go into this, including the deployment mode of collectors (i.e. not just language concerns that are in SIG's purview), so I am skeptical that a single central benchmark can justify a default one way or another.

anuraaga:
The issue is mainly that we don't have a standard or expected deployment configuration as @yurishkuro points out. Anecdotally I've heard sidecar collector is common - I don't pretend to know for sure what's really the common scenario though. But the way I look at it is that the endpoint variables default to `localhost` - so they do seem to default to a sidecar scenario. For sidecars, we'd generally expect to not use gzip since it's not going over a network - a default of enabling gzip in one variable but localhost in the other doesn't seem to provide a consistent experience. We do already have language to allow SDKs to use a different default though if they'd like which seems sufficient - it's unlikely any language will shift their default if the spec was tweaked again.

> I'm citing the wide prevalence of "compression by default" in various frameworks/dbs as further evidence of this industry-wide preference.

FWIW, MySQL defaults to `uncompressed`

https://dev.mysql.com/doc/refman/8.0/en/connection-options.html#option_general_compression-algorithms

johnnyshields:
Sidecars ideally should accept gzip-compressed traces and pass them thru (i.e. not re-compress them.) Nginx behaves in a similar way for gzip-compressed HTTP payloads.

johnnyshields:
@anuraaga if the spec is defaulting to `https` (which implies non-sidecar), doesn't it make sense to also default to `gzip`? At least we should be consistent.

I do think that non-sidecar is the more common way, looking at NewRelic, Datadog, Lightstep, and others. The whole selling point of these platforms is that you just need to add a single plugin lib to your app with no daemons, etc.

anuraaga:
> The choice of localhost does not imply that Sidecar is considered to be the "standard setup".

This is an unexpected interpretation for me and also seems irrelevant - it can still be part of the matrix which is just listing out the user experiences, right? The alternative could be specifying no default.

johnnyshields:
Keep alive

tigrannajaryan:
The TC has discussed this. I am posting on behalf of the TC:
- We believe that having a locally running Collector is common.
- The SDK defaults are optimized for this scenario. For example the default endpoints that SDKs send to are "localhost".
- In line with this we also believe that the default setting for compression options should be "none". Enabling compression when sending to localhost increases CPU consumption and is unnecessary.
- We think that it is not desirable to make one option's default value depend on another option's value (e.g. make "gzip" default compression if the destination is not "localhost").
- Some SDKs will know for certain or with high probability that the typical destination they send to is not local. For example this may be the case for in-browser JS SDK or for mobile Android SDK. For such SDKs we want to allow to choose a different default setting for compression.

Therefore the TC believes that:
- The default recommended value for `OTEL_EXPORTER_OTLP_*COMPRESSION` option in the spec should be "none".
- The spec will allow SDKs to deviate from this recommendation as explained above.

johnnyshields:
OK. Scratching my head because none of the major commercial platforms (New Relic, Datadog, Lightstep) use a sidecar. I guess Splunk does? Who were the representatives on the TC?

bogdandrutu:
@johnnyshields just for my curiosity by what measurement did use to determine "major commercial platforms"?

https://github.com/open-telemetry/community/blob/main/community-members.md#technical-committee

johnnyshields:
My comment was not meant to be an exhaustive list--the 3 I listed are indeed major platforms, among others. Should have written "e.g."

johnnyshields:
I'm going to close this for now because the current language is adequate. I still don't understand the TC's logic given the reality of most commercial platforms 🤷‍♂️

PR #1897, opened 2021-08-30
Define minimum stale time for issues/PRs in SIGs

carlosalberto:
Fixes https://github.com/open-telemetry/community/issues/806

Initial proposal to define 1 year as a **minimum** period of time before a issue without activity can be closed; use the 7 days stale, 7 days closed policy for PRs, as currently done in the Specification.

### DIFF
--- a/issue-management.md
+++ b/issue-management.md
@@ -71,6 +71,13 @@ action, please provide the rationale for closing, and indicate that OP can
 re-open for discussion if there are additional info, justification and
 questions.
 
+Issues with no activity during 1 year can be considered stale
+and can be closed. However, SIGs are free to use a longer period of time.
+
+Pull requests with no activity during 7 days can be marked stale,
+and they should stay at least 7 days as stale before they can be closed.
+Similarly to issues, SIGs are free to use longer periods of time.
+
 ## When Issues Get Stuck
 
 Some issues are not directly related to a particular code change. If an


issue-management.md
```diff
@@ -71,6 +71,13 @@ action, please provide the rationale for closing, and indicate that OP can
 re-open for discussion if there are additional info, justification and
 questions.
 
+Issues with no activity during 1 year can be considered stale
+and can be closed. However, SIGs are free to use a longer period of time.
```
yurishkuro, line 75: closed immediately or marked stale first?
```suggestion
Issues with no activity during 1 year can be marked stale and can be closed
7 days later. However, SIGs are free to use a longer period of time.
```

carlosalberto: I'd vow for closing them immediately. 1 year is rather enough time to be able to forget about them ;)

Oberon00:
My unqualified opinion as an "armchair (non-)maintainer": I think old issues should not automatically be closed. This creates an incentive to let issues rot to let them resolve itself. If a SIG is overwhelmed with the amount of issues, they can always manually sort by "updated by" and close out old issues on demand. I don't think a fixed age is the best policy for that, but rather something like "if we have more than 200 open issues we will close all issues that were created before the release of 1.3 except marked as important".

issue-management.md
arminru: > However, SIGs are free to use a longer period of time.

I think this should rather be defined in the repo guidelines in https://github.com/open-telemetry/community.

issue-management.md
```diff
@@ -71,6 +71,13 @@ action, please provide the rationale for closing, and indicate that OP can
...
+
+Pull requests with no activity during 7 days can be marked stale,
+and they should stay at least 7 days as stale before they can be closed.
+Similarly to issues, SIGs are free to use longer periods of time.
```
arminru, line 79: > However, SIGs are free to use a longer period of time.

>Similarly to issues, SIGs are free to use longer periods of time.

I think this should rather be defined in the repo guidelines in https://github.com/open-telemetry/community.

carlosalberto: We already have a lot regarding issues in this section - we can try to move it out in a follow up, however.

issue-management.md
```diff
@@ -71,6 +71,13 @@ action, please provide the rationale for closing, and indicate that OP can
 re-open for discussion if there are additional info, justification and
 questions.
 
+Issues with no activity during 1 year can be considered stale
```
iNikem, line 74: Does this mean that maintainers are not allowed to close issues faster? Regardless of the activity.

iNikem (commented):
I agree on the part about PRs, assuming that our goal is to have enough time for all interested parties to participate in PRs.

I don't understand the part about issues. What is the goal of this 1 year period? IMO maintainers should be free to close issues whenever they want. And I don't see any benefit in auto-closing.

issue-management.md
arminru: @carlosalberto Makes sense.

bogdandrutu (approved):
LGTM overall, regarding where this should be can be in the repo setup process steps

issue-management.md
arminru: This is only about closing due to it being stale. If the issue has been sufficiently addressed it can be closed anyway regardless of this.

issue-management.md
yurishkuro: Marking issue stale is a form of notification to people, who can react to it, come back and say "I'm still interested", which could signal the bot to remove stale marker. They technically can reply to a closed issue, but that won't reopen / extend the issue without manual intervention from repo owners (who can always close manually if necessary).

I personally don't find it that useful to auto-close issues in the first place. Closing PRs makes sense because they rot much faster, but an issue might record just an idea on the roadmap, or even some registry like  https://github.com/jaegertracing/jaeger/issues/638. So if we introduce auto-close for issues I think it must be accompanied by a tagging mechanism that exempts certain issues from auto-close, like `roadmap` or `keep_open` labels.

carlosalberto:
> I don't understand the part about issues. What is the goal of this 1 year period? IMO maintainers should be free to close issues whenever they want. And I don't see any benefit in auto-closing.

This is mostly about auto-closing issues (that have had no activity in long time, hence considered stale), _if_ that is supported by a specific SIG. Because of this, we have the "minimum" timeline. If that makes sense, I can clarify the wording.

carlosalberto:
Based on @yurishkuro we could also add a `roadmap` or `keep_open` tag that would prevent auto-closing issues, _if_ this is supported by a given SIG.

iNikem:
> > I don't understand the part about issues. What is the goal of this 1 year period? IMO maintainers should be free to close issues whenever they want. And I don't see any benefit in auto-closing.
> 
> This is mostly about auto-closing issues (that have had no activity in long time, hence considered stale), _if_ that is supported by a specific SIG. Because of this, we have the "minimum" timeline. If that makes sense, I can clarify the wording.

Then I agree with https://github.com/open-telemetry/opentelemetry-specification/pull/1897#issuecomment-908472396

Issue #1089, opened 2020-10-13
Add BeforeEnd to have a callback where the span is still writeable


gugahoa:
**What are you trying to achieve?**
Set an attribute to all spans in a trace

**What did you expect to see?**
Writeable spans on SpanProcessor's OnEnd(Span)

**Additional context.**
What I tried to implement is a feature called trace field in Honeycomb (reference: [`AddFieldToTrace`](https://godoc.org/github.com/honeycombio/beeline-go#AddFieldToTrace))

For that I first tried to use [`SpanProcessor`](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#span-processor), but soon  found out that [`OnEnd(Span)`](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#onendspan) is not writeable.

The solution I arrived was to write a custom exporter which was able to read from a thread-local storage (at the time of the implementation, Baggage had yet to be spec'ed out) to list all attributes related to a Trace and insert into the span before exporting.

From what I understood from the [`SpanProcessor`](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#span-processor) specification, that's the place where I should have placed this logic, as it's a transformation that should happen before the span is sent to any exporter

garthk:
Required for dynamic resource-like attributes desired in every span but computed at run-time because they're not constant, _eg._ instance uptime and recent CPU utilisation. Injecting them in the exporter gives results that become less accurate during the spikes we're most interested in.

arminru:
IIRC it was discussed to make spans writeable in OnEnd but this was left out because users will not always have full control over the order in which span processors are registered and in turn invoked, which could result in unreliable behavior.
From your use case it sounds like you would not need any attributes that are added/overwritten in OnEnd visible to any other span processor, right? If you only need to have those changes reflected once passed to exporters, this shouldn't be an issue then.

@carlosalberto you seemed to remember some of that discussion as well in the meeting yesterday. Do you know the rationale for that decision?

iNikem:
Why these attributes cannot be set in `onStart`, which already has writable span?

arminru:
Reading the spec again, I noticed that it explicitly says that spans are already marked as ended at the time they are passed to SpanProcessor.OnEnd:

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#onendspan

[...]

PR #1898, opened 2021-08-30
Capture HTTP request/response headers as span attributes

### LINKED ISSUE #1061: Capture all request and response headers

I want to capture selected/all HTTP request and response headers.

The instrumentation libraries should be configurable to capture selected or all HTTP request/response headers.

To move this forward we probably need to define:
* data semantics e.g. `http.request.header.<header-name>`/`http.response.header.<header-name>` 
* configuration for instrumentations e.g. `OTEL_INSTRUMENTATION_CAPTURE_REQUEST_HEADERS=<name1,name2>` `OTEL_INSTRUMENTATION_CAPTURE_REQUEST_HEADERS_ALL=bool`

**Additional context.**

Related issues:
* java agent capture headers https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1316
* https://github.com/opentracing-contrib/java-web-servlet-filter/issues/50

mateuszrzeszutek:
Fixes #1061

## Changes

- Add semantic conventions for capturing HTTP request and response headers.

### DIFF
--- a/specification/trace/semantic_conventions/http.md
+++ b/specification/trace/semantic_conventions/http.md
@@ -94,7 +94,7 @@ It is recommended to also use the general [network attributes][], especially `ne
 | `http.request.header.<key>` | string[] | HTTP request headers, `<key>` being the HTTP Header name (case preserving), the value being the header values. [1] [2] | `http.request.header.Content-Type=["application/json"]`; `http.request.header.X-Forwarded-for=["1.2.3.4", "1.2.3.5"]` | No |
 | `http.response.header.<key>` | string[] | HTTP response headers, `<key>` being the HTTP Header name (case preserving), the value being the header values. [1] [2] | `http.response.header.Content-Type=["application/json"]`; `http.response.header.My-custom-header=["abc", "def"]` | No |
 
-**[1]:** Instrumentations SHOULD require an explicit configuration of which headers are to be captured (e.g. an allowlist).
+**[1]:** Instrumentations SHOULD require an explicit configuration of which headers are to be captured.
 Including all request/response headers can be a security risk - explicit configuration helps avoid leaking sensitive information.
 
 Some HTTP headers - `Host` and `User-Agent` - are already captured in the `http.host` and `http.user_agent` attributes.
--- a/specification/trace/semantic_conventions/http.md
+++ b/specification/trace/semantic_conventions/http.md
@@ -87,6 +87,17 @@ Don't set the span status description if the reason can be inferred from `http.s
 
 It is recommended to also use the general [network attributes][], especially `net.peer.ip`. If `net.transport` is not specified, it can be assumed to be `IP.TCP` except if `http.flavor` is `QUIC`, in which case `IP.UDP` is assumed.
 
+### HTTP request and response headers
+
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `http.request.header.<key>` | string[] | HTTP request headers, `<key>` being the HTTP Header name (case preserving), the value being the header values. [1] | `http.request.header.Content-Type=["application/json"]`; `http.request.header.X-Forwarded-for=["1.2.3.4", "1.2.3.5"]` | No |
+| `http.response.header.<key>` | string[] | HTTP response headers, `<key>` being the HTTP Header name (case preserving), the value being the header values. [1] | `http.response.header.Content-Type=["application/json"]`; `http.response.header.My-custom-header=["abc", "def"]` | No |
+
+**[1]:** Certain headers MAY be sanitized or omitted to avoid leaking sensitive information.
+
+**[2]:** The attribute value MAY contain separate header value as separate items, or concatenated header values as one item, depending on the way the HTTP library provides access to headers.
+
 [network attributes]: span-general.md#general-network-connection-attributes
 
 ## HTTP client
--- a/specification/trace/semantic_conventions/http.md
+++ b/specification/trace/semantic_conventions/http.md
@@ -94,7 +94,11 @@ It is recommended to also use the general [network attributes][], especially `ne
 | `http.request.header.<key>` | string[] | HTTP request headers, `<key>` being the HTTP Header name (case preserving), the value being the header values. [1] [2] | `http.request.header.Content-Type=["application/json"]`; `http.request.header.X-Forwarded-for=["1.2.3.4", "1.2.3.5"]` | No |
 | `http.response.header.<key>` | string[] | HTTP response headers, `<key>` being the HTTP Header name (case preserving), the value being the header values. [1] [2] | `http.response.header.Content-Type=["application/json"]`; `http.response.header.My-custom-header=["abc", "def"]` | No |
 
-**[1]:** Certain headers MAY be sanitized or omitted to avoid leaking sensitive information.
+**[1]:** Instrumentations SHOULD require an explicit configuration of which headers are to be captured (e.g. an allowlist).
+Including all request/response headers can be a security risk - explicit configuration helps avoid leaking sensitive information.
+
+Some HTTP headers - `Host` and `User-Agent` - are already captured in the `http.host` and `http.user_agent` attributes.
+These headers SHOULD NOT be captured as a part of this convention.
 
 **[2]:** The attribute value MAY contain separate header value as separate items, or concatenated header values as one item, depending on the way the HTTP library provides access to headers.
 
--- a/specification/trace/semantic_conventions/http.md
+++ b/specification/trace/semantic_conventions/http.md
@@ -91,8 +91,8 @@ It is recommended to also use the general [network attributes][], especially `ne
 
 | Attribute  | Type | Description  | Examples  | Required |
 |---|---|---|---|---|
-| `http.request.header.<key>` | string[] | HTTP request headers, `<key>` being the HTTP Header name (case preserving), the value being the header values. [1] | `http.request.header.Content-Type=["application/json"]`; `http.request.header.X-Forwarded-for=["1.2.3.4", "1.2.3.5"]` | No |
-| `http.response.header.<key>` | string[] | HTTP response headers, `<key>` being the HTTP Header name (case preserving), the value being the header values. [1] | `http.response.header.Content-Type=["application/json"]`; `http.response.header.My-custom-header=["abc", "def"]` | No |
+| `http.request.header.<key>` | string[] | HTTP request headers, `<key>` being the HTTP Header name (case preserving), the value being the header values. [1] [2] | `http.request.header.Content-Type=["application/json"]`; `http.request.header.X-Forwarded-for=["1.2.3.4", "1.2.3.5"]` | No |
+| `http.response.header.<key>` | string[] | HTTP response headers, `<key>` being the HTTP Header name (case preserving), the value being the header values. [1] [2] | `http.response.header.Content-Type=["application/json"]`; `http.response.header.My-custom-header=["abc", "def"]` | No |
 
 **[1]:** Certain headers MAY be sanitized or omitted to avoid leaking sensitive information.
 
--- a/specification/trace/semantic_conventions/http.md
+++ b/specification/trace/semantic_conventions/http.md
@@ -98,7 +98,7 @@ It is recommended to also use the general [network attributes][], especially `ne
 Including all request/response headers can be a security risk - explicit configuration helps avoid leaking sensitive information.
 
 Some HTTP headers - `Host` and `User-Agent` - are already captured in the `http.host` and `http.user_agent` attributes.
-These headers SHOULD NOT be captured as a part of this convention.
+Users MAY explicitly configure instrumentations to capture them even though it is not recommended.
 
 **[2]:** The attribute value MAY contain separate header value as separate items, or concatenated header values as one item, depending on the way the HTTP library provides access to headers.
 
--- a/specification/trace/semantic_conventions/http.md
+++ b/specification/trace/semantic_conventions/http.md
@@ -101,7 +101,7 @@ Including all request/response headers can be a security risk - explicit configu
 Some HTTP headers - `Host` and `User-Agent` - are already captured in the `http.host` and `http.user_agent` attributes.
 Users MAY explicitly configure instrumentations to capture them even though it is not recommended.
 
-**[2]:** The attribute value MAY contain multiple header values as an array of strings, or as a single-item array containing a possibly comma-concatenated string, depending on the way the HTTP library provides access to headers.
+**[2]:** The attribute value MUST consist of either multiple header values as an array of strings or a single-item array containing a possibly comma-concatenated string, depending on the way the HTTP library provides access to headers.
 
 [network attributes]: span-general.md#general-network-connection-attributes
 
--- a/specification/trace/semantic_conventions/http.md
+++ b/specification/trace/semantic_conventions/http.md
@@ -101,7 +101,7 @@ Including all request/response headers can be a security risk - explicit configu
 Some HTTP headers - `Host` and `User-Agent` - are already captured in the `http.host` and `http.user_agent` attributes.
 Users MAY explicitly configure instrumentations to capture them even though it is not recommended.
 
-**[2]:** The attribute value MAY contain multiple header values as an array of strings, or as a single, possibly comma-concatenated, string, depending on the way the HTTP library provides access to headers.
+**[2]:** The attribute value MAY contain multiple header values as an array of strings, or as a single-item array containing a possibly comma-concatenated string, depending on the way the HTTP library provides access to headers.
 
 [network attributes]: span-general.md#general-network-connection-attributes
 
--- a/specification/trace/semantic_conventions/http.md
+++ b/specification/trace/semantic_conventions/http.md
@@ -14,6 +14,7 @@ and various HTTP versions like 1.1, 2 and SPDY.
   - [Name](#name)
   - [Status](#status)
   - [Common Attributes](#common-attributes)
+    - [HTTP request and response headers](#http-request-and-response-headers)
   - [HTTP client](#http-client)
   - [HTTP server](#http-server)
     - [HTTP server definitions](#http-server-definitions)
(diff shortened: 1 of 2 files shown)


specification/trace/semantic_conventions/http.md
```diff
@@ -85,6 +85,17 @@ Don't set the span status description if the reason can be inferred from `http.s
...
+### HTTP request and response headers
+
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `http.request.header.<key>` | string[] | HTTP request headers, `<key>` being the HTTP Header name (case preserving), the value being the header values. [1] | `http.request.header.Content-Type=["application/json"]`; `http.request.header.X-Forwarded-for=["1.2.3.4", "1.2.3.5"]` | No |
+| `http.response.header.<key>` | string[] | HTTP response headers, `<key>` being the HTTP Header name (case preserving), the value being the header values. [1] | `http.response.header.Content-Type=["application/json"]`; `http.response.header.My-custom-header=["abc", "def"]` | No |
+
+**[1]:** Certain headers MAY be sanitized or omitted to avoid leaking sensitive information.
+
+**[2]:** The attribute value MAY contain separate header value as separate items, or concatenated header values as one item, depending on the way the HTTP library provides access to headers.
```
jkwatson, line 97: is this footnote actually referenced anywhere?

jkwatson (approved):

specification/trace/semantic_conventions/http.md
mateuszrzeszutek: Oops🤭 
Thanks for catching this!

specification/trace/semantic_conventions/http.md
```diff
@@ -85,6 +85,17 @@ Don't set the span status description if the reason can be inferred from `http.s
 
 It is recommended to also use the general [network attributes][], especially `net.peer.ip`. If `net.transport` is not specified, it can be assumed to be `IP.TCP` except if `http.flavor` is `QUIC`, in which case `IP.UDP` is assumed.
 
+### HTTP request and response headers
+
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `http.request.header.<key>` | string[] | HTTP request headers, `<key>` being the HTTP Header name (case preserving), the value being the header values. [1] [2] | `http.request.header.Content-Type=["application/json"]`; `http.request.header.X-Forwarded-for=["1.2.3.4", "1.2.3.5"]` | No |
```
Oberon00, line 92: We have quite a few HTTP semantic conventions that will usually (but not necessarily) be just copies of respective headers, namely `http.host`, `http.user_agent`. What is the guidance on these? I would propose to prefer the specific semantic convention for now and not duplicate them with this by default.

Oberon00, line 92: I think it is important to provide guidance which headers should be captured by default. Capturing all headers will certainly be too much. I propose that instrumentations need to have captured headers explicitly configured. We should even have an standardized configuration (envvar) for this. The details are outside of scope of this PR probably, but in this PR I think it is important to at least state that capturing everything by default is wrong.

Oberon00 (commented):
This convention is breaking new ground, because it only defines a prefix, not a full semantic convention. We don't really have support for that in the semantic convention generator (yet?), so it's fine to only have it as a markdown table.

There are a few questions that arise from this convention, that I'll post as separate comments.

specification/trace/semantic_conventions/http.md
mateuszrzeszutek: I imagined that a user would have to configure an allowlist of headers that are to be included as span attributes, capturing all headers is definitely not a good idea from the security perspective.

specification/trace/semantic_conventions/http.md
```diff
@@ -85,6 +85,21 @@ Don't set the span status description if the reason can be inferred from `http.s
...
+| `http.response.header.<key>` | string[] | HTTP response headers, `<key>` being the HTTP Header name (case preserving), the value being the header values. [1] [2] | `http.response.header.Content-Type=["application/json"]`; `http.response.header.My-custom-header=["abc", "def"]` | No |
+
+**[1]:** Instrumentations SHOULD require an explicit configuration of which headers are to be captured (e.g. an allowlist).
```
Oberon00, line 95:
```suggestion
**[1]:** Instrumentations SHOULD require an explicit configuration of which headers are to be captured (e.g. a capturelist).
```

Or remove the sample. I think "allowlist" does not fit here (allowlist cannot replace all usages of the old term).

Oberon00 (approved):

specification/trace/semantic_conventions/http.md
carlosalberto: Also, any reason to not use MUST here? I think this is sensitive enough that should be explicit all the time.

mateuszrzeszutek: I used the `db.statement` attribute as an analogy: the DB query can contain very sensitive information too but the spec says that `The value may be sanitized to exclude sensitive information` and doesn't place any hard requirements on instrumentations to make sanitization required (or even configurable).

jsuereth:
Should HTTP Request/Response be an "Event" with these as attributes on that event?

yurishkuro:
> Should HTTP Request/Response be an "Event" with these as attributes on that event?

I'd say no, because there are no meaningful timestamps one could attach to such events that are different from the span start/end.

specification/trace/semantic_conventions/http.md
```diff
@@ -85,6 +85,21 @@ Don't set the span status description if the reason can be inferred from `http.s
...
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `http.request.header.<key>` | string[] | HTTP request headers, `<key>` being the HTTP Header name (case preserving), the value being the header values. [1] [2] | `http.request.header.Content-Type=["application/json"]`; `http.request.header.X-Forwarded-for=["1.2.3.4", "1.2.3.5"]` | No |
+| `http.response.header.<key>` | string[] | HTTP response headers, `<key>` being the HTTP Header name (case preserving), the value being the header values. [1] [2] | `http.response.header.Content-Type=["application/json"]`; `http.response.header.My-custom-header=["abc", "def"]` | No |
+
+**[1]:** Instrumentations SHOULD require an explicit configuration of which headers are to be captured.
+Including all request/response headers can be a security risk - explicit configuration helps avoid leaking sensitive information.
+
+Some HTTP headers - `Host` and `User-Agent` - are already captured in the `http.host` and `http.user_agent` attributes.
+These headers SHOULD NOT be captured as a part of this convention.
```
owais, line 99: Can we remove this to simplify implementation a bit? Users will most likely not configure this feature to capture headers that are already part of the spec and supported by their SDK.

specification/trace/semantic_conventions/http.md
owais: Since users will have to explicitly mention all headers they want to capture, I think it is unlikely they'll specify headers that their SDK already captures. Can we leave it out to simplify a bit?

specification/trace/semantic_conventions/http.md
```diff
@@ -85,6 +85,21 @@ Don't set the span status description if the reason can be inferred from `http.s
...
 It is recommended to also use the general [network attributes][], especially `net.peer.ip`. If `net.transport` is not specified, it can be assumed to be `IP.TCP` except if `http.flavor` is `QUIC`, in which case `IP.UDP` is assumed.
 
+### HTTP request and response headers
+
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `http.request.header.<key>` | string[] | HTTP request headers, `<key>` being the HTTP Header name (case preserving), the value being the header values. [1] [2] | `http.request.header.Content-Type=["application/json"]`; `http.request.header.X-Forwarded-for=["1.2.3.4", "1.2.3.5"]` | No |
+| `http.response.header.<key>` | string[] | HTTP response headers, `<key>` being the HTTP Header name (case preserving), the value being the header values. [1] [2] | `http.response.header.Content-Type=["application/json"]`; `http.response.header.My-custom-header=["abc", "def"]` | No |
+
+**[1]:** Instrumentations SHOULD require an explicit configuration of which headers are to be captured.
```
owais, line 95: Should we specify the env vars that should be used to configure this?

specification/trace/semantic_conventions/http.md
Oberon00: I'd see this as a suggestion for the user how to use the configuration. An explicit request by the user is a valid reason not to honor the SHOULD.

specification/trace/semantic_conventions/http.md
mateuszrzeszutek: I don't think that we have any instrumentation config prop/env var spec in this repo - I agree that this needs to be done, but as a separate effort and not a part of this PR.

specification/trace/semantic_conventions/http.md
owais: I'm concerned SDK authors might not read it like that. Your comment is very clear. Perhaps we can re-word it to sound more like that?

mateuszrzeszutek: Done 👍

Oberon00:
After seeing #1907, I wonder if my proposal to preserve casing was so good after all, or if we need something additional. IMHO a good goal would be to be able to capture header information in such a way that #1907 is resolved as well. Some ideas:
1. Normalize header names to lowercase, `-` to `_` (like lowercased CGI envvars), do not preserve casing.
2. Optionally capture a list of configured headers normalized (in addition to case-preserving or only). The prefix could be the same or `http.(request|response).normalized_header` or similar. Or we could rename the current capturing to `raw_header`. The normalization could be done by the instrumentation, a span processor (if we have #1089), the exporter, the collector or even the backend.
3. Just state that normalization / accessing keys in a case-insensitive way is the responsibility of whoever processes the data (maybe in addition to defining the normalization but define nothing more. I don't think that would be good enough to satisfy proposals like #1907 in the future though.

CC @Blacksmoke16

Blacksmoke16:
@Oberon00 HTTP/2 requires headers be lowercase while in HTTP/1 it doesn't matter. Normalizing them to all however would be backwards compatible, while also being future proof.

mateuszrzeszutek:
> 1. Normalize header names to lowercase, `-` to `_` (like lowercased CGI envvars), do not preserve casing.

I like this idea because it's the simplest out of the three you suggested - and the most intuitive one ("why don't we lowercase headers" was the first question I asked in the linked issue 😄 ). We do lose a little bit of information that way, but since HTTP headers are supposed to be case insensitive by design it's probably not worth that much.
I'll rework my PR with idea 1. in mind.

specification/trace/semantic_conventions/http.md
```diff
@@ -85,6 +86,21 @@ Don't set the span status description if the reason can be inferred from `http.s
 
 It is recommended to also use the general [network attributes][], especially `net.peer.ip`. If `net.transport` is not specified, it can be assumed to be `IP.TCP` except if `http.flavor` is `QUIC`, in which case `IP.UDP` is assumed.
 
+### HTTP request and response headers
+
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `http.request.header.<key>` | string[] | HTTP request headers, `<key>` being the normalized HTTP Header name (lowercase, with `-` characters replaced by `_`), the value being the header values. [1] [2] | `http.request.header.content_type=["application/json"]`; `http.request.header.x_forwarded_for=["1.2.3.4", "1.2.3.5"]` | No |
+| `http.response.header.<key>` | string[] | HTTP response headers, `<key>` being the normalized HTTP Header name (lowercase, with `-` characters replaced by `_`), the value being the header values. [1] [2] | `http.response.header.content_type=["application/json"]`; `http.response.header.my_custom_header=["abc", "def"]` | No |
```
bogdandrutu, line 94: Why not have this as attributes with "type" `map[string][]string` (too much golang)? Then the key would be fixed.

yurishkuro: I had the same suggestion on the ticket itself https://github.com/open-telemetry/opentelemetry-specification/issues/1061#issuecomment-908447921. I think nested map is a much cleaner approach than inventing bespoke translation rules and polluting the overall attributes namespace.

mateuszrzeszutek: So what should we do with this PR then? Should we wait with it until we reach a consensus on #376 and implement headers as nested structures?

iNikem: If we accept this PR, will it possible later to define `http.request.headers` attribute to contain a map of headers? I assume "yes", so we can proceed and merge this PR for now. Although @bogdandrutu may object to having two separate semantic conventions to represent the same entity :)

mateuszrzeszutek: > If we accept this PR, will it possible later to define `http.request.headers` attribute to contain a map of headers?

It will - although having 2 conventions that basically do the same thing will be far from elegant.
I also had some question in #376, maybe answering them will make the whole picture a bit clearer.

carlosalberto:
> Why not have this as attributes with "type" map[string][]string

Since we don't support this now (and probably it will take a little while before we do so), should we proceed with this approach for now? @bogdandrutu @yurishkuro

carlosalberto (approved):

Oberon00:
> and probably it will take a little while before we do so

You are assuming that we will support it 😉 That's not decided at all.

carlosalberto:
> That's not decided at all.

Yeah, definitely. More the reason to go with the current approach.

Oberon00:
I'm somewhat strongly against adding support for more complex attributes, like maps. I think this PR should be merged as-is, and then we should close #376 because the only use case that came up so far has been solved differently. What would the purpose of adding maps *after* this PR be? The added-value of maps is already very small (negative IMHO), and if we add maps after this PR, there is one more negative to it, namely the disruption it causes for this semantic convention.

specification/trace/semantic_conventions/http.md
```diff
@@ -85,6 +86,21 @@ Don't set the span status description if the reason can be inferred from `http.s
...
+
+**[1]:** Instrumentations SHOULD require an explicit configuration of which headers are to be captured.
+Including all request/response headers can be a security risk - explicit configuration helps avoid leaking sensitive information.
+
+Some HTTP headers - `Host` and `User-Agent` - are already captured in the `http.host` and `http.user_agent` attributes.
+Users MAY explicitly configure instrumentations to capture them even though it is not recommended.
+
+**[2]:** The attribute value MAY contain separate header value as separate items, or concatenated header values as one item, depending on the way the HTTP library provides access to headers.
```
yurishkuro, line 102:
```suggestion
**[2]:** The attribute value MAY contain multiple header values as an array of strings, or as a single, possibly comma-concatenated, string, depending on the way the HTTP library provides access to headers.
```

yurishkuro (approved):

specification/trace/semantic_conventions/http.md
Oberon00: @yurishkuro, no the current wording is right: We want to always have an array, even if it has just a single element.

Oberon00: See #831

yurishkuro: Ok, but it doesn't mention array and the wording is confusing. Not a blocker.

Oberon00 (changes requested):
I'm against having the value be set as a single string instead of single-element array, as it makes typing more complicated. See https://github.com/open-telemetry/opentelemetry-specification/pull/1898#discussion_r711013196

specification/trace/semantic_conventions/http.md
Oberon00: The array is in the type column (note the `[]`).

mateuszrzeszutek: I've changed the wording so that it's somewhat in between those two versions, but strongly saying that it should be an array.

specification/trace/semantic_conventions/http.md
```diff
@@ -85,6 +86,21 @@ Don't set the span status description if the reason can be inferred from `http.s
 
 It is recommended to also use the general [network attributes][], especially `net.peer.ip`. If `net.transport` is not specified, it can be assumed to be `IP.TCP` except if `http.flavor` is `QUIC`, in which case `IP.UDP` is assumed.
 
+### HTTP request and response headers
```
arminru, line 89: Changelog please 🙂

arminru (approved):

Oberon00 (approved):

specification/trace/semantic_conventions/http.md
mateuszrzeszutek: Done 👍

specification/trace/semantic_conventions/http.md
```diff
@@ -87,6 +88,21 @@ Don't set the span status description if the reason can be inferred from `http.s
...
+| `http.response.header.<key>` | string[] | HTTP response headers, `<key>` being the normalized HTTP Header name (lowercase, with `-` characters replaced by `_`), the value being the header values. [1] [2] | `http.response.header.content_type=["application/json"]`; `http.response.header.my_custom_header=["abc", "def"]` | No |
+
+**[1]:** Instrumentations SHOULD require an explicit configuration of which headers are to be captured.
+Including all request/response headers can be a security risk - explicit configuration helps avoid leaking sensitive information.
+
+Some HTTP headers - `Host` and `User-Agent` - are already captured in the `http.host` and `http.user_agent` attributes.
+Users MAY explicitly configure instrumentations to capture them even though it is not recommended.
+
+**[2]:** The attribute value MAY contain multiple header values as an array of strings, or as a single-item array containing a possibly comma-concatenated string, depending on the way the HTTP library provides access to headers.
```
Oberon00, line 104: Actually, the wording is a bit odd here with the "MAY". If we opt to not use the "MAY", what else can we set? Nothing, right? So instead of "MAY ... or ...", I suggest "MUST either ... or ...".

carlosalberto:
Will merge by EOD unless somebody opposes.

iNikem (approved):

Oberon00 (approved):

Issue #1819, opened 2021-07-16
Sampling SIG progress report


jmacd:
**What are you trying to achieve?**

Report to the general tracing community the problems and solutions that have been discussed.

The Sampling SIG has had two meetings and themes are emerging:

1. The `TraceIDRatio` sampler is our preferred default mechanism because it supports independent span sampling without requiring propagation of the effective sampling probability.
2. To complete the `TraceIDRatio` specification for its intended use, we need to agree on and specify a hashing mechanism, as we do not expect TraceID values to be uniformly distributed.
3. Any Sampler or export/processing stage that knows the output inclusion probability for a span SHOULD output that value in the form of an adjusted count (i.e., the inverse of inclusion probability) using an attribute named `sampling.adjusted_count` (proposed)
4. The name of the effective sampling policy should be attached as a resource attribute to indicate further information about what sampling is taking place, e.g., `sampling.policy=xyz`. If the sampling policy is dynamic on a per-span basis, the attribute MAY be a span attribute instead of a resource attribute.

**The problem with incomplete traces**

When collecting spans that may or may not have been sampled, there is a well-known problem with identifying when a Trace is complete. This problem exists with or without sampling, but it becomes substantially worse with sampling. Without sampling, a trace could be incomplete because spans are dropping, so you could suspect incomplete traces whenever spans are dropping. With sampling, you have to suspect traces may be incomplete even under normal operation. 

We are searching for mechanisms to detect incomplete traces, and the ones we know of are:

1. Ensure that collection is perfect (i.e., no dropped spans), use a Parent Sampler for sampling decisions
2. Ensure that collection is perfect (i.e., no dropped spans), use the `TraceIDRatio` Sampler and record the per-process minimum TraceIDRatio threshold that is in effect. Assume traces are complete when they fall below the global threshold.
3. The "missing parent" heuristic: when a parent is missing you can tell because there is a child without a parent (this doesn't work for missing leaves).
4. The approach OpenCensus took, which is to count the expected number of child spans and check that they are all present.

[...]

PR #1899, opened 2021-08-30
Probability sampling specification

### LINKED ISSUE #1413: Review approach & specify algorithm for TraceIdRatioBasedSampler (ProbabilitySampler)
See also discussion on https://github.com/open-telemetry/opentelemetry-specification/pull/1412#issuecomment-776101910

The sampling alorithm for TraceIdRatioBasedSampler is unspecified. As a result, trace IDs that are sampled by some implementations might get non-sampled or re-sampled by SDKs in other languages, even though they have the same or a a higher probability than the parent.

TODO list for this issue:

1. Is this a problem at all? Or is the ParentBased approach enough (in combination with any, not necessarily trace-id-based probability-based sampling of root spans)?
2. ~~If it is a problem, since trace IDs can come from untrusted, non-random sources, do we open up a DDoS/Security/performance issue when using trace IDs as sole, deterministic input for our sampling algorithm? Do we need to put a warning there?~~ Let's assume it is no problem for this issue, this should be handled in #1188.
3. If applicable after checking the above we determine that we need a consistent algorithm, actually do specify it (maybe based on #331).

jmacd:
This proposes changes in the specification implied by OTEPs 168 and 170:

https://github.com/open-telemetry/oteps/pull/168
https://github.com/open-telemetry/oteps/pull/170

The justification for these changes is given in those two OTEPs. Our goal is to ensure that Span-to-Metrics pipelines can be built to statistically count spans after they have been sampled and before they have been assembled into traces.

The two OTEPs are not receiving enough attention. This PR is meant to consolidate the implied changes of those two OTEPs combined in a way that makes it easier to see what is being specified (WHAT) without all the related justification (WHY). If you approve these changes, please also review and approve those OTEPs first.

Fixes #1413.
Part of #1819.

This cannot be merged before OTEPs 168 and 170 are merged, but this is also not a draft, it is a complete proposal. This has been prototyped twice, once in Java and once in Go. See OTEP 170 for more detail about these prototypes.

### DIFF
--- a/specification/trace/sdk.md
+++ b/specification/trace/sdk.md
@@ -119,15 +119,18 @@ Thus, the SDK specification defines sets of possible requirements for
 
 ## Sampling
 
-Sampling is a mechanism to control the noise and overhead introduced by
-OpenTelemetry by reducing the number of samples of traces collected and sent to
-the backend.
+Sampling is a mechanism to control the cost of OpenTelemetry tracing by
+reducing the number of spans collected and sent to the backend.
+OpenTelemetry gives participants in a distributed trace the option to sample
+based on whether their context was sampled or to make the decision
+independently.
 
-Sampling may be implemented on different stages of a trace collection. The
-earliest sampling could happen before the trace is actually created, and the
-latest sampling could happen on the Collector which is out of process.
+Sampling may be implemented in different stages of trace collection. The
+earliest sampling happens inside the trace SDK, but later sampling decisions
+can happen outside the process, for example in an OpenTelemetry collector.
 
-The OpenTelemetry API has two properties responsible for the data collection:
+The OpenTelemetry API has two properties associated with Spans that allow the
+user to become aware of the sampling decision:
 
 * `IsRecording` field of a `Span`. If `false` the current `Span` discards all
   tracing data (attributes, events, status, etc.). Users can use this property
@@ -165,6 +168,59 @@ The following table summarizes the expected behavior for each combination of
 The SDK defines the interface [`Sampler`](#sampler) as well as a set of
 [built-in samplers](#built-in-samplers) and associates a `Sampler` with each [`TracerProvider`].
 
+### Probability sampling
+
+OpenTelemetry specifies a mechanism for conveying sampling probability, when
+known, to enable accurate statistical counting of the population of spans
+using only the portion that were sampled and collected.
+
+Samping probability is conveyed in a form known as "adjusted count", which
+is the number of spans in the population accurately represented by the
+individual.  In common terms, a "1-in-N" sampling scheme produces spans with
+adjusted count N, where every sample span represents N in the general
+population.  Adjusted count is the inverse of sampling probability except
+when the probability is zero, which is an important special case.
+
+Although sampling can be carried out in multiple stages, OpenTelemetry
+specifies a dedicated field in the Span data model for representing
+probability at the "head" of the distributed trace, where it describes the
+probability the `Sampled` flag was set in the Span's initial sampling
+decision.
+
+To reduce the cost of conveying Sampling information through propagators,
+OpenTelemetry limits head sampling probabilities to powers of two or zero.
+Adjusted counts are likewise limited to powers of two or zero.  As not all
+Sampler implementations will be probabilistic in nature, a special value for
+the "unknown" adjusted count is included in the data model.  Aside from the
+zero and unknown cases, adjusted count values can be encoded using their
+base-2 logarithm in a small number of bits.
+
+The OpenTelemetry Span field for encoding adjusted count is named
+`log_head_adjusted_count`, with the default value zero representing the case
+of unknown adjusted count.  Values 1 through 62 encode one plus the base-2
+logarithm of adjusted count (i.e., adjusted count equals
+`2**(encoded_value-1)`), and zero adjusted count is encoded by value 63.
+Thus, it takes 6 bits to encode the `log_head_adjusted_count` field, with
+interpretation specified in the table below.
+
+<a name="log-head-adjusted-count-table"></a>
+
+| Encoded Value | Head adjusted count | Head sampling probability |
+| ------------- | ------------------- | ------------------------- |
+| 0             | Unknown             | Unknown                   |
+| 1             | 1                   | 1                         |
+| 2             | 2                   | 1/2                       |
+| 3             | 4                   | 1/4                       |
+| 4             | 8                   | 1/8                       |
+| ...           | ...                 | ...                       |
+| N             | 2**(N-1)            | 2**(-N+1)                 |
+| ...           | ...                 | ...                       |
+| 62            | 2**61               | 2**-61                    |
+| 63            | 0                   | 0                         |
+
+The built-in Samplers are defined so that when this specification is
+followed, all spans will have known head sampling probability.
+
 ### SDK Span creation
 
 When asked to create a Span, the SDK MUST act as if doing the following in order:
@@ -175,7 +231,7 @@ When asked to create a Span, the SDK MUST act as if doing the following in order
 2. Query the `Sampler`'s [`ShouldSample`](#shouldsample) method
    (Note that the [built-in `ParentBasedSampler`](#parentbased) can be used to
    use the sampling decision of the parent,
-   translating a set SampledFlag to RECORD and an unset one to DROP).
+   translating a set SampledFlag to `RECORD_AND_SAMPLE` and an unset one to `DROP`).
 3. Generate a new span ID for the `Span`, independently of the sampling decision.
    This is done so other components (such as logs or exception handling) can rely on
    a unique span ID, even if the `Span` is a non-recording instance.
@@ -227,12 +283,22 @@ object must be immutable (multiple calls may return different immutable objects)
   If the sampler returns an empty `Tracestate` here, the `Tracestate` will be cleared,
   so samplers SHOULD normally return the passed-in `Tracestate` if they do not intend
   to change it.
+* The value of `log_head_adjusted_count` for the Span data model, reflecting
+  known and unknown values for adjusted count defined above.  The following guidance 
+  is given for user-defined Samplers:
+  * If the Sampler can be implemented as choice of built-in Sampler behavior
+    made at runtime, the resulting _composite_ sampler gives the correct
+    behavior.
+  * `RECORD_ONLY` and `DROP` sampling decisions indicate zero adjusted count.  It is
+    an error to return a `RECORD_ONLY` or `DROP` decision with non-zero adjusted count.
+  * `RECORD_AND_SAMPLE` sampling decisions that cannot be defined in terms of
+    a built-in Sampler SHOULD use unknown adjusted count.
 
 #### GetDescription
 
 Returns the sampler name or short description with the configuration. This may
 be displayed on debug pages or in the logs. Example:
-`"TraceIdRatioBased{0.000100}"`.
+`"TraceIdRatioBased{.000122}"`.
 
 Description MUST NOT change over time and caller can cache the returned value.
 
@@ -261,11 +327,8 @@ the `ParentBased` sampler specified below.
   represented as a decimal number. The precision of the number SHOULD follow
   implementation language standards and SHOULD be high enough to identify when
   Samplers have different ratios. For example, if a TraceIdRatioBased Sampler
-  had a sampling ratio of 1 to every 10,000 spans it COULD return
-  `"TraceIdRatioBased{0.000100}"` as its description.
-
-TODO: Add details about how the `TraceIdRatioBased` is implemented as a function
-of the `TraceID`. [#1413](https://github.com/open-telemetry/opentelemetry-specification/issues/1413)
+  had a sampling ratio of 1-in-(2**13) spans it COULD return
+  `"TraceIdRatioBased{0.000122}"` as its description.
 
 ##### Requirements for `TraceIdRatioBased` sampler algorithm
 
--- a/specification/trace/sdk.md
+++ b/specification/trace/sdk.md
@@ -311,45 +311,138 @@ The default sampler is `ParentBased(root=AlwaysOn)`.
 
 * Returns `RECORD_AND_SAMPLE` always.
 * Description MUST be `AlwaysOnSampler`.
+* Behavior of `ShouldSample` is identical to `TraceIdRatioBased{1}`
 
 #### AlwaysOff
 
 * Returns `DROP` always.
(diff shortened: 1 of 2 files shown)


specification/trace/sdk.md
```diff
@@ -119,15 +119,18 @@ Thus, the SDK specification defines sets of possible requirements for
 
 ## Sampling
 
-Sampling is a mechanism to control the noise and overhead introduced by
-OpenTelemetry by reducing the number of samples of traces collected and sent to
-the backend.
+Sampling is a mechanism to control the cost of OpenTelemetry tracing by
+reducing the number of spans collected and sent to the backend.
```
yurishkuro, line 123:
```suggestion
reducing the number and/or verbosity of spans collected and sent to the backend.
```

specification/trace/sdk.md
```diff
@@ -165,6 +168,59 @@ The following table summarizes the expected behavior for each combination of
 The SDK defines the interface [`Sampler`](#sampler) as well as a set of
 [built-in samplers](#built-in-samplers) and associates a `Sampler` with each [`TracerProvider`].
 
+### Probability sampling
+
+OpenTelemetry specifies a mechanism for conveying sampling probability, when
```
yurishkuro, line 173:
```suggestion
OpenTelemetry specifies a mechanism for recording sampling probability, when
```

specification/trace/sdk.md
```diff
@@ -165,6 +168,59 @@ The following table summarizes the expected behavior for each combination of
...
+known, to enable accurate statistical counting of the population of spans
+using only the portion that were sampled and collected.
+
+Samping probability is conveyed in a form known as "adjusted count", which
+is the number of spans in the population accurately represented by the
+individual.  In common terms, a "1-in-N" sampling scheme produces spans with
+adjusted count N, where every sample span represents N in the general
```
yurishkuro, line 180:
```suggestion
adjusted count N, where every sample span represents N spans in the general
```

specification/trace/sdk.md
```diff
@@ -165,6 +168,59 @@ The following table summarizes the expected behavior for each combination of
...
+population.  Adjusted count is the inverse of sampling probability except
+when the probability is zero, which is an important special case.
+
+Although sampling can be carried out in multiple stages, OpenTelemetry
+specifies a dedicated field in the Span data model for representing
+probability at the "head" of the distributed trace, where it describes the
+probability the `Sampled` flag was set in the Span's initial sampling
```
yurishkuro, line 187: > probability the `Sampled`

probability _when_ the `Sampled`?

specification/trace/sdk.md
```diff
@@ -245,48 +311,138 @@ The default sampler is `ParentBased(root=AlwaysOn)`.
...
+    R := 0
+    for {
+      if nextRandomBit() {
+        break
+      }
+      R++
+    }
+	if R < 63 {
+	  return R
+    }
```
yurishkuro, line 424:
```suggestion
    if R < 63 {
        return R
    }
```

specification/trace/sdk.md
```diff
@@ -245,48 +311,138 @@ The default sampler is `ParentBased(root=AlwaysOn)`.
...
+random bits limited to maximum value 62.  Here is example code to compute a
+value for `r`:
+
+```golang
+func nextRandomness() int {
+  // Repeat until a valid result is produced.
+  for {
+    R := 0
+    for {
+      if nextRandomBit() {
```
yurishkuro, line 417:
```suggestion
      if nextRandomBit() == 1 {
```

yurishkuro (commented):
+1

tedsuo (approved):
Once Yuri's suggestions are responded to, this LGTM

specification/trace/sdk.md
```diff
@@ -245,48 +311,138 @@ The default sampler is `ParentBased(root=AlwaysOn)`.
...
+	if R < 63 {
+	  return R
+    }
+	// Reject, try again.
+  }
+}
+```
+
+Given a randomness value `r` and a power-of-two `TraceIdRatioBased` sampling
+ratio expressed as`2**(-s)`, the `TraceIdRatioBased` Sampler MUST make the
```
oertl, line 431: should be `2**(1-s)` to be consistent with the way the head sampling probability is encoded (compare table above)

specification/trace/sdk.md
```diff
@@ -245,48 +311,138 @@ The default sampler is `ParentBased(root=AlwaysOn)`.
...
+###### TraceIdRatioBased: Non-root case
+
+The `r` value should have been generated by the root `Sampler`.  If no `r`
+value was provided in the context, the Sampler SHOULD return an unknown
+adjusted count as the `log_head_adjusted_count`.  No `p` value should be
+propagated to child contexts, in this case.
+
+When `r` was properly generated by a root Sampler, the decision is the same
+for the root case.  Given a randomness value `r` and a power-of-two
+`TraceIdRatioBased` sampling ratio expressed as`2**(-s)`, the
```
oertl, line 443: `2**(1-s)`

yurishkuro (approved):

specification/trace/sdk.md
```diff
@@ -245,48 +311,138 @@ The default sampler is `ParentBased(root=AlwaysOn)`.
...
+This Sampler supports making independent sampling decisions at each Span in
+a trace, making it possible for each SDK to independently control its rate
+of sample spans.  The decision is defined to be consistent, such that for
+any two spans `T` and `U` of the same trace configured with
+`TraceIDRatioBased` ratios `ratio_T` and `ratio_U`, respectively, if
+`ratio_T <= ratio_U` and span `T` was sampled, span `U` MUST also be
+sampled.
+
+The constructor for `TraceIDRatioBased` MUST accept ratios in the range 0
+through 1, inclusive.  When the input ratio is not a power of two exactly,
```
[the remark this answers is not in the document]
jmacd: Fixed!

specification/trace/sdk.md
```diff
@@ -245,48 +311,138 @@ The default sampler is `ParentBased(root=AlwaysOn)`.
...
+the Sampler MUST round the ratio down to the nearest power of two.
```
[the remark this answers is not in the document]
jmacd: Fixed!

specification/trace/sdk.md
jmacd: Fixed!

specification/trace/sdk.md
jmacd: Fixed!

specification/trace/sdk.md
```diff
@@ -245,48 +311,138 @@ The default sampler is `ParentBased(root=AlwaysOn)`.
...
+| 0         | 1/2       |
+| 1         | 1/4       |
+| 2         | 1/8       |
+| 3         | 1/16      |
+| ...       | ...       |
+| N         | 2**(-N-1) |
+| ...       | ...       |
+| 62        | 2**(-63)  |
+| 63        | Reject    |
+
```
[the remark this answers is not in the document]
jmacd: You are correct, and thank you for the long-form proof! I will update the doc.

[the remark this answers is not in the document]
jmacd: To convince myself of this, I worked out an example for a 3-bit encoding scheme, which I will incorporate into OTEP 168 as I incorporate the latest feedback. By the way, I think I prefer to reason about the sampling probability in the form of `2**-S`, not `2**-S+1`. We should stop using the variable `P` for probability, since it's being used as the offset-by-one encoded value for head probability.

With `2**-S` defining `S` we have 6 defined values for `R` in this scheme and 8 values for `P`.  The `P` values are:

| `P` value | Adjusted count |
| -----     | -----          |
| 0         | Unknown        |
| 1         | 1              |
| 2         | 2              |
| 3         | 4              |
| 4         | 8              |
| 5         | 16             |
| 6         | 32             |
| 7         | 0              |

The `R` values, with their selection probability, and each possible `S` value with the corresponding adjusted count:

| `R` value | `R` selection probability | `S=0` | `S=1` | `S=2` | `S=4` | `S=5` | `S=6` |
| --        | --                        | --    | --    | --    | --    | --    | --    |
| 0         | 1/2                       | 1    | 0     | 0     | 0     | 0     | 0     |
| 1         | 1/4                       | 1     | 2     | 0     | 0     | 0     | 0     |
| 2         | 1/8                       | 1     | 2     | 4     | 0     | 0     | 0     |
| 3         | 1/16                      | 1     | 2     | 4     | 8     | 0     | 0     |
| 4         | 1/32                      | 1     | 2     | 4     | 8     | 16    | 0     |
| 5         | 1/32                      | 1     | 2     | 4     | 8     | 16    | 32    |

The test is that any value for `R` and any value for `S` produce one expected span when sampling. For example, in the `S=5`
column we have `0*1/2 + 0*1/4 + 0*1/8 + 0*1/16 + 16*1/32 + 16*1/32 =
16/32 + 16/32 = 1`. In the `S=2` column we have `0*1/2 + 0*1/4 +
4*1/8 + 4*1/16 + 4*1/32 + 4*1/32 = 4/8 + 4/16 + 4/32 + 4/32 = 1/2 +
1/4 + 1/8 + 1/8 = 1`.

For the 6-bit encoding, therefore I would prefer we allow values of R in [0,61], leaving 62 and 63 invalid values.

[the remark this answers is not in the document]
oertl: @jmacd, column `S=3` is missing in your second table, but I assume that `S` is meant to be from `[0,5]`.

Here is the explanation why I have chosen `R in [1,62]`:

`P` is used to encode the adjusted count. I would prefer that the adjusted count is given by `2**P`. However, I guess this is not possible because `UNKNOWN` has to be equal to the default value which is 0, right? So we have to encode the adjusted count as `2**(P-1)` with `P in [1,62]`.

`S` is used for the sampling probability. I think encoding the sampling probability as `2**(1-S)` with `S in [1,62]` and having `S = 63` for zero probability is more consistent with the encoding of the adjusted count `2**(P-1)` as we will always have `P = S`.

`R` is used to encode the randomness. By having `R in [1,62]` the sampling condition writes as `S <= R`. For `R in [0,61]` the sampling condition would be `S <= R + 1` which is slightly more complex. This was my main reason to have `R in [1,62]`.

If we could encode `UNKNOWN` with `P = 63`, for example, we could encode the adjusted count by `2**P` the sampling probability as `2**-S` and take `R` from `[0,61]` with the sampling condition `S <= R`. This would be all much more natural and intuitive.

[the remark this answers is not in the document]
jmacd: I have updated this PR to reflect the current state of OTEP 168.

jmacd:
Reviewers, please consider another look at https://github.com/open-telemetry/oteps/pull/168 which now covers a number of corner cases and is generally finished. Once that merges we can continue working on this PR. The questions that I propose we defer for "future debate" are summarized here in this comment https://github.com/open-telemetry/oteps/pull/168#issuecomment-916459431 and I propose that we debate them here afterhttps://github.com/open-telemetry/oteps/pull/168 merges.

jmacd:
FYI https://github.com/open-telemetry/oteps/pull/175 adds a glossary of sampling terms used that would be used to flesh out a tracing `datamodel.md` document. Unlike Metrics and Logs, Tracing does not have one of these.

specification/trace/sdk.md
```diff
@@ -165,6 +168,59 @@ The following table summarizes the expected behavior for each combination of
...
+To reduce the cost of conveying Sampling information through propagators,
+OpenTelemetry limits head sampling probabilities to powers of two or zero.
+Adjusted counts are likewise limited to powers of two or zero.  As not all
+Sampler implementations will be probabilistic in nature, a special value for
+the "unknown" adjusted count is included in the data model.  Aside from the
+zero and unknown cases, adjusted count values can be encoded using their
+base-2 logarithm in a small number of bits.
+
+The OpenTelemetry Span field for encoding adjusted count is named
+`log_head_adjusted_count`, with the default value zero representing the case
```
bogdandrutu, line 199: Where is this defined in the data-model? Why is this needed if the tracestate is already present?

bogdandrutu (commented):
@jsuereth I thought that we don't immediately define specs for OTEPs, especially when we are in a new unknown area. Also this changes a "stable" document without having a prototype which we should not correct?

specification/trace/sdk.md
jmacd: (Remember this PR is a draft until all the OTEPs merge.)

The field `log_head_adjusted_count` is described in https://github.com/open-telemetry/oteps/blob/main/text/trace/0170-sampling-probability.md.

You make a valid point about `tracestate`. On the other hand, it's less clearly part of the data model if the user is required to parse the `tracestate` (twice, as there are two syntaxes) to recover the information. Besides, `tracestate` is not an ideal solution in the long term. Our best long term solution is to modify the W3C `traceparent` to include a few extra bytes. Then, the rationale of storing `log_head_adjusted_count` as an independent field is that we may have other means of determining `log_head_adjusted_count` in the future.

For now, I would suggest that OTel erases its own `tracestate` record from the span data and specify fields for anything that's part of the data model. As it stands, the `r` value from OTEP 168 is not being recorded and as far as I know, no one has asked for that.

specification/trace/sdk.md
```diff
@@ -119,15 +119,18 @@ Thus, the SDK specification defines sets of possible requirements for
 
 ## Sampling
 
-Sampling is a mechanism to control the noise and overhead introduced by
-OpenTelemetry by reducing the number of samples of traces collected and sent to
-the backend.
+Sampling is a mechanism to control the cost of OpenTelemetry tracing by
+reducing the number and/or verbosity of spans collected and sent to the backend.
+OpenTelemetry gives participants in a distributed trace the option to sample
```
reyang, line 124: Not a native English speaker, I wonder if this would improve the readability?

```suggestion
OpenTelemetry gives distributed trace participants the option to sample
```

specification/trace/sdk.md
```diff
@@ -119,15 +119,18 @@ Thus, the SDK specification defines sets of possible requirements for
...
+based on whether their context was sampled or to make the decision
+independently.
 
-Sampling may be implemented on different stages of a trace collection. The
-earliest sampling could happen before the trace is actually created, and the
-latest sampling could happen on the Collector which is out of process.
+Sampling may be implemented in different stages of trace collection. The
+earliest sampling happens inside the trace SDK, but later sampling decisions
+can happen outside the process, for example in an OpenTelemetry collector.
```
reyang, line 130:
```suggestion
can happen outside the process, for example in an OpenTelemetry Collector.
```

reyang (approved):

PR #1905, opened 2021-09-06
Fix propagator's Extract required argument

pellared:
Fixes propagator's Extract required argument.
1. Capitalize `HTTP`
2. It should be a "request", not a "response"


anuraaga (approved):

arminru (approved):
Thanks!

Issue #1902, opened 2021-08-31
Support summary aggregation type for metrics


jack-berg:
**What are you trying to achieve?**

Summary aggregations are mentioned in the [metrics data model](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#summary-legacy) anad the [opentelemetry protos](https://github.com/open-telemetry/opentelemetry-proto/blob/v0.9.0/opentelemetry/proto/metrics/v1/metrics.proto#L274-L276) but [are not listed](https://github.com/open-telemetry/opentelemetry-specification/blame/main/specification/metrics/sdk.md#L265-L274) as an aggregation that must be supported by the SDK.

Tried searching through the issues / PRs for context, but couldn't find a reason for the omission. 

From my perspective, summaries are a good lightweight alternative to histograms that don't significantly bloat the scope of the SDK spec, and they should be supported in the initial version. 

**Additional context.**

N/A.

jack-berg:
Clarifying this request, I'm looking for some way to represent a min, max, sum, count aggregation. 

We discussed this in the metrics 9/2 metrics SIG and proposed using a 0 bucket histogram, extended with optional min and max fields, to represent this situation. 

An old PR on [opentelemetry-proto](https://github.com/open-telemetry/opentelemetry-proto/pull/279) has been re-opened to propose the necessary additions to the histogram protos which enable this.

reyang:
Discussed more during the [9/9/2021 Metrics SIG meeting](https://docs.google.com/document/d/1pdvPeKjA8v8w_fGKAN68JjWBmVJtPCpqdi9IZrd6eEo/edit#):

1. The OpenTelemetry clients won't implement Summary. Summary only exists in the Data Model and OTLP for Prometheus compatibility (so OTLP can represent Prometheus data without having to drop them).
2. Min/Max is a good fit for Histogram.
3. Suggested that we use Proto3+ field presence and options.

@jsuereth will follow up in the Spec SIG regarding the proto3 field presence and options.

PR #1912, opened 2021-09-08
Prepare for the metrics SDK experimental release

reyang:
This PR is pending on #1888 and #1913 to be merged.

Changes:
* Removed the warning & banner.
* Updated changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,8 @@ release.
   [#1864](https://github.com/open-telemetry/opentelemetry-specification/pull/1864),
   [#1828](https://github.com/open-telemetry/opentelemetry-specification/pull/1828),
   [#1888](https://github.com/open-telemetry/opentelemetry-specification/pull/1888),
-  [#1911](https://github.com/open-telemetry/opentelemetry-specification/pull/1911))
+  [#1912](https://github.com/open-telemetry/opentelemetry-specification/pull/1912),
+  [#1913](https://github.com/open-telemetry/opentelemetry-specification/pull/1913))
 
 ### Logs
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,16 @@ release.
 
 ### Metrics
 
+- Added Experimental Metrics SDK specification.
+  ([#1673](https://github.com/open-telemetry/opentelemetry-specification/pull/1673),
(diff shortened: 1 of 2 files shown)


jack-berg:
Hi @reyang - I'm curious to get your (and others) thoughts on when best to resolve Issue #1902 regarding some sort of support for a min, max, sum, count style aggregation.

The current thought is that we can achieve this by [extending the histogram proto](https://github.com/open-telemetry/opentelemetry-proto/pull/279) with optional min / max fields. If we went down this route, it would be nice to have this type of histogram specified in the SDK spec. Specifically, I think we could add a `Zero Bucket Summary Histogram Aggregation` specification akin to the [Explicit Bucket Histogram Aggregation](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#explicit-bucket-histogram-aggregation). 

Personally, I think it would be valuable to make this aggregation available in the experimental release.

jsuereth:
@jack-berg Adding min/max fields to histogram proto would be part of "metrics data model" specification.   If we add min/max there, then implicitly histogram aggregators will need to be updated to support it.  I think that, combined with providing no buckets to histogram explicit bucket will get you exactly what you want out of `Summary` from the `Histogram` instrument type.


Happy to help shepherd you through the process and drive Prs on appropriate repositories.

reyang:
@jack-berg thanks for calling this out. I think these are not blockers for SDK spec Experimental release, and this is definitely a nice-to-have and we should get it before Feature-freeze. Here goes the [Metrics SIG meeting notes](https://docs.google.com/document/d/1pdvPeKjA8v8w_fGKAN68JjWBmVJtPCpqdi9IZrd6eEo/edit#).

There are several parts I could imagine:
1. The data model spec (which is Stable) has to be updated.
2. The proto has to be updated, using the protobuf v3 OPTIONAL.
3. The SDK/View might need to be updated so folks can specify if they want to spend the extra CPU/memory to get min/max, and whether it should be opt-in (default off) or opt-out (default on).

jmacd (approved):

jsuereth (approved):
Approved, pending PRs listed as blocking.

cijothomas (approved):

bogdandrutu (approved):

arminru (approved):

PR #1913, opened 2021-09-09
Spec out the MeterProvider ForceFlush and Shutdown requirement

reyang:
This was previously left as TODO since it has dependency on the exporter (#1864) and reader (#1888) interface.

### DIFF
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -85,22 +85,28 @@ to make the shutdown timeout configurable.
 
 ### ForceFlush
 
-This method provides a way for provider to immediately export all metrics data
-that have not yet been exported. Note: unlike [Push Metric
-Exporter](#push-metric-exporter) which can send data on its own schedule, [Pull
-Metric Exporter](#pull-metric-exporter) can only send the data when it is being
-asked by the scraper, so `ForceFlush` would not make sense.
+This method provides a way for provider to notify the registered
+[MetricReader](#metricreader) and [MetricExporter](#metricexporter) instances,
+so they can do as much as they could to consume or send the metrics. Note:
+unlike [Push Metric Exporter](#push-metric-exporter) which can send data on its
+own schedule, [Pull Metric Exporter](#pull-metric-exporter) can only send the
+data when it is being asked by the scraper, so `ForceFlush` would not make much
+sense.
 
 `ForceFlush` SHOULD provide a way to let the caller know whether it succeeded,
-failed or timed out.
+failed or timed out. `ForceFlush` SHOULD return some **ERROR** status if there
+is an error condition; and if there is no error condition, it should return some
+**NO ERROR** status, language implementations MAY decide how to model **ERROR**
+and **NO ERROR**.
 
 `ForceFlush` SHOULD complete or abort within some timeout. `ForceFlush` can be
 implemented as a blocking API or an asynchronous API which notifies the caller
 via a callback or an event. OpenTelemetry client authors can decide if they want
 to make the flush timeout configurable.
 
-`ForceFlush` MUST invoke `ForceFlush` on all registered [push
-exporters](#push-metric-exporter).
+`ForceFlush` MUST invoke `ForceFlush` on all registered
+[MetricReader](#metricreader) [Push Metric Exporter](#push-metric-exporter)
+instances.
 
 ### View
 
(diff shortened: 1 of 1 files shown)


specification/metrics/sdk.md
```diff
@@ -64,11 +64,41 @@ configuration only via this reference.
 
 ### Shutdown
 
-TODO
+This method provides a way for provider to do any cleanup required.
+
+`Shutdown` MUST be called only once for each `MeterProvider` instance. After the
+call to `Shutdown`, subsequent attempts to get a `Meter` are not allowed. SDKs
```
jsuereth, line 70: I'd suggest that subsequent calls just return failure status.

specification/metrics/sdk.md
```diff
@@ -64,11 +64,41 @@ configuration only via this reference.
...
+`Shutdown` SHOULD provide a way to let the caller know whether it succeeded,
+failed or timed out.
+
+`Shutdown` SHOULD complete or abort within some timeout. `Shutdown` can be
+implemented as a blocking API or an asynchronous API which notifies the caller
+via a callback or an event. OpenTelemetry client authors can decide if they want
+to make the shutdown timeout configurable.
+
+`Shutdown` MUST be implemented at least by invoking `Shutdown` on all registered
+[exporters](#metricexporter).
```
jsuereth, line 82: Should this also call out `MetricReader` shutdowns?

jsuereth (approved):

specification/metrics/sdk.md
reyang: I took the wording from https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#shutdown.

I think the tricky part is "return failure status", we try to give language clients flexibility by saying "provide some way" so they can decide if exception or return value or something else (e.g. a global error handler) makes more sense.

specification/metrics/sdk.md
reyang: Technically YES.
Practically - we need to wait for the MetricReader PR to be merged and clean that part (I cannot link to something that isn't yet merged), I think this is not a blocker though.

cijothomas (approved):

specification/metrics/sdk.md
bogdandrutu: >`Shutdown` MUST be implemented at least by invoking `Shutdown` on all registered
[exporters](#metricexporter).

Concerned about this, and now I kind of understand your "on_collect" idea, and I am worried that because of this requirement a specific kind of complicated designed is required. I do understand the compatibility story between Traces and Metrics, but kind of this complicates the interface between MeterProvider and MeterReader.

Curios if something simpler (and accepted by others) can be achieved if "ForceFlush" is removed from the MeterProvider at this moment and maybe added later, and Shutdown will be present only on the reader.

This will simplify the "pull" problem, and may simplify overall design.

reyang: Discussed during the [9/9/2021 Metrics SIG meeting](https://docs.google.com/document/d/1pdvPeKjA8v8w_fGKAN68JjWBmVJtPCpqdi9IZrd6eEo/edit#), we'll keep `ForceFlush` and improve the wording here.

reyang: I've updated the wording based on the discussing during the SIG meeting. PTAL.

jmacd (approved):

specification/metrics/sdk.md
```diff
@@ -66,11 +66,47 @@ configuration only via this reference.
...
+to make the shutdown timeout configurable.
+
+`Shutdown` MUST be implemented at least by invoking `Shutdown` on all registered
+[MetricReader](#metricreader) and [MetricExporter](#metricexporter) instances.
 
 ### ForceFlush
 
-TODO
+This method provides a way for provider to notify the registered
+[MetricReader](#metricreader) and [MetricExporter](#metricexporter) instances,
```
bogdandrutu, line 89: If I understand correctly MetricExporter is not directly connected to the Provider, so just notify the Reader is enough correct? It is Reader's responsibility to notify Exporter

bogdandrutu (approved):

specification/metrics/sdk.md
reyang: For Pull Exporters we've agreed to give implementers flexibility so they could (although we haven't seen a strong reason to do so) implement the Pull Exporter bypassing the Reader. https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#pull-metric-exporter

_Implementors MAY choose the best idiomatic design for their language. For example, they could generalize the Push Metric Exporter interface design and use that for consistency, they **could model the pull exporter as MetricReader**, or they **could design a completely different pull exporter interface**._

The wording here ensures that all the Readers and Exporters will be covered. An implementation could achieve this by having the Provider notifying the Readers, and Readers notifying Exporters.

PR #1914, opened 2021-09-09
Avoid using gRPC when generic RPC systems have same properties

bogdandrutu:
The concept of "streaming" is supported by multiple protocols including gRPC, but also others.

### DIFF
--- a/semantic_conventions/trace/rpc.yaml
+++ b/semantic_conventions/trace/rpc.yaml
@@ -2,7 +2,7 @@ groups:
   - id: rpc
     prefix: rpc
     brief: 'This document defines semantic conventions for remote procedure calls.'
-    events: [rpc.grpc.message]
+    events: [rpc.message]
     attributes:
       - id: system
         type: string
@@ -142,10 +142,10 @@ groups:
         note: >
           This is always required for jsonrpc. See the note in the general
           RPC conventions for more information.
-  - id: rpc.grpc.message
-    prefix: "message"  # TODO: Change the prefix to rpc.grpc.message?
+  - id: rpc.message
+    prefix: "message"  # TODO: Change the prefix to rpc.message?
     type: event
-    brief: "gRPC received/sent message."
+    brief: "RPC received/sent message."
     attributes:
       - id: type
         type:
--- a/specification/trace/semantic_conventions/rpc.md
+++ b/specification/trace/semantic_conventions/rpc.md
@@ -14,11 +14,11 @@ This document defines how to describe remote procedure calls
     - [Span name](#span-name)
     - [Attributes](#attributes)
       - [Service name](#service-name)
+    - [Events](#events)
     - [Distinction from HTTP spans](#distinction-from-http-spans)
   - [gRPC](#grpc)
     - [gRPC Attributes](#grpc-attributes)
     - [gRPC Status](#grpc-status)
-    - [Events](#events)
   - [JSON RPC](#json-rpc)
     - [JSON RPC Attributes](#json-rpc-attributes)
 
@@ -95,6 +95,32 @@ Generally, a user SHOULD NOT set `peer.service` to a fully qualified RPC service
 [`service.name`]: ../../resource/semantic_conventions/README.md#service
 [`peer.service`]: span-general.md#general-remote-service-attributes
 
+### Events
+
+In the lifetime of a RPC stream, an event for each message sent/received on
+client and server spans SHOULD be created. In case of unary calls only one sent
+and one received message will be recorded for both client and server spans.
+
+The event name MUST be `"message"`.
+
+<!-- semconv rpc.message -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `message.type` | string | Whether this is a received or sent message. | `SENT` | No |
+| `message.id` | int | MUST be calculated as two different counters starting from `1` one for sent messages and one for received message. [1] |  | No |
+| `message.compressed_size` | int | Compressed size of the message in bytes. |  | No |
+| `message.uncompressed_size` | int | Uncompressed size of the message in bytes. |  | No |
+
+**[1]:** This way we guarantee that the values will be consistent between different implementations.
+
+`message.type` MUST be one of the following:
+
+| Value  | Description |
+|---|---|
+| `SENT` | sent |
+| `RECEIVED` | received |
+<!-- endsemconv -->
+
 ### Distinction from HTTP spans
 
 HTTP calls can generally be represented using just [HTTP spans](./http.md).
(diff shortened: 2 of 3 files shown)


specification/trace/semantic_conventions/rpc.md
```diff
@@ -95,6 +95,32 @@ Generally, a user SHOULD NOT set `peer.service` to a fully qualified RPC service
 [`service.name`]: ../../resource/semantic_conventions/README.md#service
 [`peer.service`]: span-general.md#general-remote-service-attributes
 
+### Events
+
+In the lifetime of a RPC stream, an event for each message sent/received on
```
arminru, line 100:
```suggestion
In the lifetime of an RPC stream, an event for each message sent/received on
```

semantic_conventions/trace/rpc.yaml
```diff
@@ -142,10 +142,10 @@ groups:
         note: >
           This is always required for jsonrpc. See the note in the general
           RPC conventions for more information.
-  - id: rpc.grpc.message
-    prefix: "message"  # TODO: Change the prefix to rpc.grpc.message?
+  - id: rpc.message
```
arminru, line 145: Changelog please 🙂

arminru (approved):

semantic_conventions/trace/rpc.yaml
bogdandrutu: Done :)

jsuereth (approved):

jsuereth:
Think this may need a telemetry schema update cc @tigrannajaryan

bogdandrutu:
@jsuereth no idea since the JSON was added after the last release, only change is that the event applies to all RPC systems :D which I have no idea how to define in schema url change :D

specification/trace/semantic_conventions/rpc.md
bogdandrutu: @jsuereth can you confirm? I think is "a" since the next word is not a vowel....

jkwatson: When you say this out loud, though, you say "arr" for the letter 'R', so it gets the `an` here.

tigrannajaryan:
> Think this may need a telemetry schema update cc @tigrannajaryan

I do no think this needs a schema update. If I understand correctly this change says that a source that previously was not emitting the event should now do so (i.e. all RPCs, not just gRPC).

There are no changes in the attribute names or values so the schema does not change in any way.

specification/trace/semantic_conventions/rpc.md
```diff
@@ -95,6 +95,32 @@ Generally, a user SHOULD NOT set `peer.service` to a fully qualified RPC service
...
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `message.type` | string | Whether this is a received or sent message. | `SENT` | No |
+| `message.id` | int | MUST be calculated as two different counters starting from `1` one for sent messages and one for received message. [1] |  | No |
+| `message.compressed_size` | int | Compressed size of the message in bytes. |  | No |
+| `message.uncompressed_size` | int | Uncompressed size of the message in bytes. |  | No |
+
+**[1]:** This way we guarantee that the values will be consistent between different implementations.
+
+`message.type` MUST be one of the following:
```
yurishkuro, line 116: I would call it message.kind, similar to span.kind. "type" sounds more like talking about the content of the message.

Do we use type in other similar scenarios?

Given that this is a new namespace, we could make this change now.

semantic_conventions/trace/rpc.yaml
```diff
@@ -2,7 +2,7 @@ groups:
   - id: rpc
     prefix: rpc
     brief: 'This document defines semantic conventions for remote procedure calls.'
-    events: [rpc.grpc.message]
```
yurishkuro, line 5: This is breaking change, no? Should there be a corresponding change to the scheme transformation instructions?

tigrannajaryan: I do not fully understand how our convention build tool works but it seems like the `rpc.grpc.` prefix has no effect on the final attribute names. The attribute names start with `message.` so there are no actual changes to the schema (at least that's what I see by looking at the generated .md file).

Someone please correct me if I am wrong.

specification/trace/semantic_conventions/rpc.md
jsuereth: @tigrannajaryan @bogdandrutu This is the kind of change that would require a schema change.

Also, this field (`message.type`) was specific to gRPC (and adapted from OpenCensus), as the type of message is a concept in gRPC.

I'd prefer not to loose this if we promote gRPC conventions up to all RPC conventions.

semantic_conventions/trace/rpc.yaml
bogdandrutu: It is not since this ID was anyway added after the last release (never published). The ID as defined today is just a reference, anyway, so not used in the semantics.

specification/trace/semantic_conventions/rpc.md
bogdandrutu: The namespace is a no-op change as defined today, so the entire PR is a not a breaking change. I would prefer to not do this change now.

tigrannajaryan: >This is the kind of change that would require a schema change.

Why would it be a schema change? What has changed? The attribute name is the same, the values are the same. 

If I understand correctly this  PR merely says that an event can be recorded by all RPC sources, not just gRPC sources as it was before. I fail to see a schema change here.

carlosalberto (approved):

semantic_conventions/trace/rpc.yaml
```diff
@@ -142,10 +142,10 @@ groups:
         note: >
           This is always required for jsonrpc. See the note in the general
           RPC conventions for more information.
-  - id: rpc.grpc.message
-    prefix: "message"  # TODO: Change the prefix to rpc.grpc.message?
+  - id: rpc.message
+    prefix: "message"  # TODO: Change the prefix to rpc.message?
```
Oberon00, line 146: While at it: Should we change it? Claiming the "message" namespace for rpc messages seems dubious to me.

bogdandrutu: That is a breaking change :) I keep trying to keep the PR as non breaking change :D

bogdandrutu: Joking: Why did you not do that when added the entry ;)

Oberon00: I didn't want to do any semantic changes in my PR back then 😅 Looks like this renaming is something of a hot potato being passed around. Fine to do it in a (quick) follow-up.

bogdandrutu:
@open-telemetry/specs-approvers @open-telemetry/specs-trace-approvers @open-telemetry/technical-committee all comments are addressed, if no more comments in the next 24h we will merge this.

semantic_conventions/trace/rpc.yaml
Oberon00: Yes, the ID is used internally within the YAMLs for cross-referencing, and typically by code generators to generate constant names.

Oberon00 (approved):

Oberon00 (commented):
Actually, I have one more question about this: The message.id attribute has a _very_ specific definition. Does that make sense for RPC generally? Did it even make sense for gRPC? Should we (temporarily) mark it as deprecated until we decide if it makes sense?

jsuereth:
@Oberon00 The message.id thing is for streaming RPCs and is the current status quo for gRPC instrumentation See: https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/gRPC.md#message-events

Oberon00:
Then maybe that property should be kept as rpc.grpc.id?

jsuereth:
I'd be ok with keeping things grpc specific when we're not sure if it applies to non gRPC systems.

carlosalberto:
> I'd be ok with keeping things grpc specific when we're not sure if it applies to non gRPC systems.

@bogdandrutu ^ ?

bogdandrutu:
@Oberon00 @jsuereth without the ID you cannot map sent with received messages in a streaming RPC, which is one of the critical property to be able to determine network delays, reorderings, etc. I think it should be a very generic property for any streaming implementation not just gRPC.

Oberon00:
> without the ID you cannot map sent with received messages

Well we need some ID, but maybe not one generated in exactly that way. Some RPC systems may already have an ID.

PR #1915, opened 2021-09-09
Add optional min and max fields to histogram data model.

jack-berg:
Related to Issue #1902. See this [convo](https://github.com/open-telemetry/opentelemetry-specification/pull/1912#issuecomment-915645835) as well.

## Changes

Expand histogram description in metric datamodel to include optional min and max fields.

The [Metric Points](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#metric-points) section of the data model is marked as stable, but I believe that the optional designation on these fields makes this a non breaking change.

Related issues #

N/A.


jack-berg:
If folks want to proceed with this change, I'd also like to update the [model-data-histogram.png](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/img/model-delta-histogram.png) image, but don't have the original file.

I can also update the changelog if necessary.

specification/metrics/datamodel.md
```diff
@@ -398,6 +400,11 @@ denotes Delta temporality where accumulated event counts are reset to zero after
 and a new aggregation occurs. Cumulative, on the other hand, continues to
 aggregate events, resetting with the use of a new start time.
 
+The min and max fields behave differently based on aggregation temporality. When
+Delta, they represent the minimum and maximum values recorded over the time window. When
+Cumulative, they represent the minimum and maximum values recently recorded. The max age
```
jsuereth, line 405: This cumulative behavior deviates from cumulative in all other cases.  Is there a reason for this divergence here?

jack-berg: This is an attempt to make min and max at least somewhat useful in the cumulative case. While having a sliding min and max with an unspecified window isn't perfect, its more useful than a true cumulative min or max would be. 

An additional piece of context is that this mimics the behavior of quantiles in prometheus summaries.

jmacd: +1 to mimicking Prometheus.

bogdandrutu: -2 :) I think we should not diverge from the temporality in case of histogram. Summary (quantiles) is different because we don't have temporality and we know we just have that to support prometheus.

bogdandrutu: Also this is not useful unless you know the time window for these values.

jack-berg: Good point bout mergeability. But are cumulative histograms with "recent" min and max values actually unmergeable, or is the result of merging just a bit strange? You could take two of these histograms and merge them by taking the min of mins, and the max of the maxes. But the interpretation of the output would be strange: the merged max would be the max recorded value over the duration of the merged windows PLUS some unspecified (max age) duration before the earlier of the two windows. 

However, since the usefulness of min and max on cumulative histograms is questionable (regardless of whether the values are truly cumulative or represent recent values), we might consider leaving them unspecified for cumulative temporality.

bogdandrutu: The big problem is if they have a different window than the temporality. 

Also we need to define how to do cumulative to delta, probably no min/max

jack-berg: > The big problem is if they have a different window than the temporality.

I'm not sure I follow. 

> Also we need to define how to do cumulative to delta, probably no min/max

Yes. The options appear to be either to drop the min / max fields, or to take them as is. 

If the max age of the min / max fields is equal to the size of the windows for the delta metrics, then taking the min / max as is for the delta metrics is accurate. In this case, the cumulative min / max are being calculated just like a delta min / max. However, if the max age is a different size than the windows, then taking the min / max as is will result in inaccuracy in the resulting delta metrics.

bogdandrutu (changes requested):
I think this PR needs to clarify in more details the correlation between these new fields and the temporality and how to calculate "delta -> cumulative -> delta"

jack-berg:
Here's a proposal for how to handle "delta to cumulative" and "cumulative to delta" conversion.

### Cumulative to Delta Conversion
**Summary**: The min / max on cumulative histograms include values recorded more recently than an unspecified max age. Assume the max age is equal to the cumulative collection interval. 
**Pros**: Simple to compute min / max: just take the cumulative min / max as is which is effectively delta already.
**Cons**: Inaccurate if the cumulative max age is not equal to the delta window size. For example, if the cumulative max age is larger than the delta window, then the reported delta max / min may include values from previous windows. If the cumulative max age is smaller than the delta window, then the reported delta min / max may be missing values included in the count / sum. However, the min / max recorded in cumulative histogram would already be subject to this inaccuracy. In general, the cumulative max age needs to be greater or equal to the cumulative collection interval. 
```
histogram {"temporality":  "cumulative", "start": "00:00:00", "end": "00:05:00", "count":  5, "sum": 50, "min":  5, "max": 20}
histogram {"temporality":  "cumulative", "start": "00:00:00", "end": "00:10:00", "count":  8, "sum": 70, "min":  4, "max": 14}
histogram {"temporality":  "cumulative", "start": "00:00:00", "end": "00:15:00", "count":  13, "sum": 100, "min":  3, "max": 16}

converts to:

histogram {"temporality":  "delta", "start": "00:00:00", "end": "00:05:00", "count":  5, "sum": 50, "min":  5, "max": 20}
histogram {"temporality":  "delta", "start": "00:05:00", "end": "00:10:00", "count":  3, "sum": 20, "min":  4, "max": 14}
histogram {"temporality":  "delta", "start": "00:10:00", "end": "00:15:00", "count":  5, "sum": 30, "min":  3, "max": 16}
```

### Delta to Cumulative Conversion
**Summary**: The min / max on cumulative histograms include values recorded more recently than an unspecified max age. Use the delta min / max values as is in the cumulative histogram, assuming the max age is equal to the delta window size. 
**Pros**: Simple to compute min / max: just take the delta min / max as is.
**Cons**: The min / max in the resulting cumulative histograms won't have any standard max age. 
```
histogram {"temporality":  "delta", "start": "00:00:00", "end": "00:05:00", "count":  5, "sum": 50, "min":  5, "max": 20}
histogram {"temporality":  "delta", "start": "00:05:00", "end": "00:10:00", "count":  3, "sum": 20, "min":  4, "max": 14}
histogram {"temporality":  "delta", "start": "00:10:00", "end": "00:15:00", "count":  5, "sum": 30, "min":  3, "max": 16}

converts to:

histogram {"temporality":  "cumulative", "start": "00:00:00", "end": "00:05:00", "count":  5, "sum": 50, "min":  5, "max": 20}
histogram {"temporality":  "cumulative", "start": "00:00:00", "end": "00:10:00", "count":  8, "sum": 70, "min":  4, "max": 14}
histogram {"temporality":  "cumulative", "start": "00:00:00", "end": "00:15:00", "count":  13, "sum": 100, "min":  3, "max": 16}
```

**Thoughts**
The delta to cumulative conversion I propose has little downside. However the cumulative to delta conversion definitely has the opportunity to introduce error into the resulting delta histograms. IMO, this error is probably unavoidable, and a consequence of min / max being a poor match for cumulative metrics to begin with. If this error is unacceptable, options include: 
- Ensure delta histograms are generated at point of data generation so that cumulative to delta conversion is not needed.
- Ensure the max age used when collecting cumulative min / max is equal to the cumulative collection interval. 
- Drop the min and max fields. Although I assume in many scenarios data with some inaccuracy is better than no data at all.

specification/metrics/datamodel.md
jsuereth: I think I agree w/ bogdan here.  Reporting min/max is really problematic.

A few main points:

## Aggregation + backend implementations

One of the benefits of cumulative aggregation is you can ignore missing data points and "just take the last one" (you don't merge).    If you're doing time-alignment or data compression (e.g. moving from 1 minute resolution, to 5 minute) you can just grab closest points to the time-alignment you want to keep the values.

## Prometheus doesn't recommend it

There's a nice [openmetrics warning about summary](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#summary):

> They MAY also be used for backwards compatibility, because some existing instrumentation libraries expose precomputed quantiles and do not support Histograms. Precomputed quantiles SHOULD NOT be used, because quantiles are not aggregatable and the user often can not deduce what timeframe they cover.

## We have no place to store the "delta" window
Right now the start/end timestamps we record represent the FULL cumualtive window.   If you want Histograms to report some kind of delta-time-window or "recency" min/max you'd also need to add a "recent start timestamp" to cumulative metrics.  This would negate all the benefits of cumulative metrics (e.g. the time-alignment property) and force us to pay very close attention to dropped points (one of the benefits of cumulative is you can be resilient to dropped points).

jsuereth (changes requested):
I think min/max need to abide by the aggregation temporality, just like all the other fields in histogram.

jack-berg:
> I think min/max need to abide by the aggregation temporality, just like all the other fields in histogram.

Just want to call out that if min/max abide by aggregation temporality and represent the cumulative min / max since recording began, then it's most reasonable to drop them when converting from cumulative to delta. The conversion from delta to cumulative is straight forward.

specification/metrics/datamodel.md
```diff
@@ -398,6 +400,11 @@ denotes Delta temporality where accumulated event counts are reset to zero after
...
+of recently recorded is commonly set to 10 minutes, and is not specified in the histogram.
```
reyang, line 406: I'm worried that we're opening a can of worm here.
Given histogram has clear semantic regarding temporality, I would wish we follow it strictly:

* For delta, if we have `(Tn, Tn+1]`, the MAX value means "the maximum value we've observed during (Tn, Tn+1]".
* For cumulative, if we have `(T0, Tn+1]`, the MAX value means "the maximum value we've observed during (T0, Tn+1]".
* For cumulative, if somehow we need to reset the start time due to cardinality explosion / memory pressure, and we have `(Tm, Tm+n]`, the MAX value means "the maximum value we've observed during (Tm, Tm+n]".

If we need to merge histogram, or perform delta/cumulative conversion, we should also make sure the result follows the exact same semantic, and if it's impossible to achieve that goal (e.g. the information is insufficient for us to deduce the correct value), we should drop the data.

If we have other useful/valid scenario that cannot be covered by the above proposal, we should consider using something else instead of Histogram Min/Max (e.g. some crazy idea - maybe a special type of exemplar).

reyang (changes requested):
Please find my concerns and suggestions here https://github.com/open-telemetry/opentelemetry-specification/pull/1915/files#r710727328.

specification/metrics/datamodel.md
jack-berg: These are valid points. 

The two routes I see are: 
1. Min / max on cumulative histogram are recently recorded values. The semantics for merging and converting to / from delta histograms are problematic in the ways described by @jsuereth, @bogdandrutu, and @reyang. But, **the information is still somewhat useful**. 
2. Min / max on cumulative histograms are true cumulative values. Semantics for merging are intuitive. Delta to cumulative conversion is intuitive. There is no way to convert from cumulative to delta so min / max should be dropped in this operation. So good logical semantics, but the **information is essentially useless**. 

I'm fine with the second option. Frankly, I would rather see a data model in which the min / max on cumulative are useful, even at the expense of questionable merging and conversion. But my primary objective is to have some lightweight mechanism to capture min, max, sum, and count. If min and max are added to histograms, but effectively only useful for delta temporality, I'm still happy.

jmacd:
I'm skeptical about the resistance to this proposal.

On the surface, remember that this discussion is meant to enable a MinMaxSumCount style of aggregation (i.e., a zero-bucket histogram). We want to encode Min/Max/Sum/Count without any further resolution, so the buckets of a histogram are not a substitute. We're interested in Min and Max quantiles **specifically because they are mergeable**. The comments above saying they are not mergeable address temporal aggregation, not spatial aggregation.

Let's say there wasn't a cumulative temporality, and we were only discussing delta-temporality histograms. There's still a problem with temporal alignment, but it's not enough of a problem for me to discard the data. A great deal of metrics data will pass through temporal alignment before it is finally shown to the user. When splitting and merging deltas to perform realignment, the same problem shows up. Why is this such an obstacle?

I think of Min and Max values as like Gauges. When it comes to merging multiple of them, you select the last value of each series. We can write rules for merging histograms of either temporality, always output the last value of the min/max that you have.

specification/metrics/datamodel.md
jsuereth: I'm curious why the information is essentially useless in the #2 case.   For explicit bucket histograms, this can still inform you whether your bucket boundaries have been chosen well.

My suggestion for min/max in histograms is akin to your (2):

1. We add min,max to histogram.  They are somewhat of a gauge, but with a slightly more robust meaning (i.e. they leverage the aggregation temporality). 
2. We drop min/max when doing CUMULATIVE->DELTA translations.  I think it's a meaningless value at the point, as we can't really know if the min/max were seen during any particular interval.

However, I'd like to propose a few other avenues for you achieving your goal:

1. You could create a view configuration that exports 2 gauges + 1 sum for every histogram instrument.  Specifically you'd create 3 metric streams:   `{metric}_sum` (Sum) `{metric}_max` (Gauge) and `{metric}_min` (Gauge).  This is effectively how Prometheus actually reports histograms today, and requires no adjustments to the data-model.  We could expand the SDK to support this.

2. You could create an Exemplar sampler that guarantees min/max are recorded during every reporting interval.   This would have the advantage of having the timestamp of the observed min/max and you can encode a special label on the "filtered_attributes" component.   We *do* have Exemplar sampling exposed as overridable API in the SDK specification.

3. We explicit mark min/max as "Gauge" values in the DataModel and do NOT prescribe merging via mathematical min/max but "last-value".  Effectively this is like option 1 but forever encoded in the datamodel.  I believe this is what jmacd@ is suggesting (and this is the wording I'd suggest you change to if we go this route).

specification/metrics/datamodel.md
jsuereth: Not to duplicate, but my thoughts are here: https://github.com/open-telemetry/opentelemetry-specification/pull/1915#discussion_r711424211

jsuereth:
@jmacd I commented [here](https://github.com/open-telemetry/opentelemetry-specification/pull/1915#discussion_r711424211) with  my thoughts on the way forward.

I think the major issue for us to figure out is basically:  Should we leverage the natural mathematical function "min" as the merge algorithm for the value "min" on histogram (similarly for max)?

I think it would be totally viable to say 'no' to this and make min/max have Gauge semantics within histogram (just like "sum" and "count" have Sum semantics).

I *do* think the natural min/max functions have meaning, even spatially but perhaps there's a scenario I'm not thinking of.  THis is why my bias is towards giving "min" and "max" their own aggregate function.   Can you give me an example of spatial min-max merge where mathematical "min" and "max" wouldn't make sense?

specification/metrics/datamodel.md
jack-berg: > I'm curious why the information is essentially useless in the #2 case. For explicit bucket histograms, this can still inform you whether your bucket boundaries have been chosen well.

To me, what you describe falls into the gap between "essentially useless" and "completely useless" 😁. It can only tell you whether the top / bottom ends of your bucket bounds have been chosen well. And after the boundaries have been tuned, what then? 

> My suggestion for min/max in histograms is akin to your (2):

👌 

Thoughts on the other avenues you propose:
1. Would work but has weak ergonomics. I think you want this information collated on the same metric. Also, min and max don't want gauge last value merge semantics. E.g. if I have a `[{start: t0, end:t1, max: 10}, {start: t1: end: t2, max 3}]` I want the merge to look like `[{start: t0, end: t2, max: 10}]`. 
2. Also think this has weak ergonomics. You don't want to go digging through examplars to find min / max. Plus these are fields that benefit from merging. (Is merging exemplars a thing? 🤔 If it is, would be strange to have some merge function that combined special exemplars.)
3. I think, at least for the delta case, it's crucial that min / max are merged using min / max functions rather than last value. Treating them as gauges with last value merging for both delta and cumulative is overly accommodating for cumulative (for which min / max aren't a great fit to begin with) at the expense of value for delta.

jmacd:
> Can you give me an example of spatial min-max merge where mathematical "min" and "max" wouldn't make sense?

No, I believe these always make sense for spatial merge. I was proposing "Gauge semantics" as a trick to let min and max work for temporal aggregation, it's just careful language to allow delta temporality to do one thing and cumulative temporality to do another without breaking any definitions. 😀 

If we defined the min/max to be interpreted over the window `(X, Time]` where `X=Max(StartTime, Time-10m)`, then delta temporality with intervals <= 10 minutes are able to encode the exact min and max and cumulative exports at least follow the same semantics. You could convert delta to cumulative and correctly meet the specification by keeping up to 10 minutes of state, and when you convert from cumulative back to delta now the min/max fields are no longer exact. This doesn't bother me, but we could add a flag to indicate when min/max are exact if that will help.

specification/metrics/datamodel.md
jsuereth: Given your proposal I think it means you have to pick whether or not you want the natural aggregation for min/max to be min + max.   That decision needs to happen first before we can agree on a path forward.

If we want that, then CUMULATIVE should use the same natural aggregation function as DELTA, and we give up CUMULATIVE->DELTA (or at some reduced capacity), and likely don't bother filling out min/max for cumulative histograms (if those values are useless).

jack-berg: I think that's a valid path forward @jsuereth. 👍

reyang:
Discussed during the [9/21/21 Metrics SIG Meeting](https://docs.google.com/document/d/1pdvPeKjA8v8w_fGKAN68JjWBmVJtPCpqdi9IZrd6eEo/edit#) and concluded that @jack-berg will update the PR:

Conclusion:
min/max would be added to Histogram, and they respect the temporality for both Delta and Cumulative
Add a section saying that Delta->Cumulative will work (and is mathematically correct), and Cumulative->Delta won’t work mathematically (due to the information loss), however folks can take the min/max and convert it to something else (e.g. Gauge) or simply drop min/max on the floor
[reyang] +1
[jsuereth] +1
[jack-berg] +1 this would work for us
[bogdan] +1 happy with the proposal, need to follow up regarding how hard it would be for the Collector to implement that (need to double check before making the decision, it might be a huge change)

jack-berg:
I've updated the description of the min and max fields to reflect the conclusion of the 9/21 SIG meeting.

jsuereth (approved):
This LGTM now.  If you'd like to update the pictures, ping me on slack and I'll get you access to the google drawing to reproduce them.

CHANGELOG.md
```diff
@@ -54,7 +56,9 @@ release.
 
 ### OpenTelemetry Protocol
 
-- Add environment variables for configuring the OTLP exporter protocol (`grpc`, `http/protobuf`, `http/json`) ([#1880](https://github.com/open-telemetry/opentelemetry-specification/pull/1880))
+- Add environment variables for configuring the OTLP exporter protocol
```
bogdandrutu, line 59: Why this change? Consider a separate PR

bogdandrutu (commented):
PR looks better, would prefer couple of days to think more on this but seems to be the right track

CHANGELOG.md
jack-berg: Oh sorry. I wrote that line in an earlier PR and realized it wasn't conforming to the line length recommendations of the rest of the file. Figured I could fix it up while I was making additional changes to the changelog. 

However I understand the desire for PRs that are focused and will revert it.

jmacd (approved):

Issue #1747, opened 2021-06-08
Define span modelling conventions for HTTP client spans


jkwatson:
For consistency across languages and HTTP client libraries, we should provide suggested conventions for modeling details of HTTP client calls. Here is a short list of items to include in the modeling effort:

* How do we represent HTTP retries?
* How do we represent HTTP redirects?
* How do we represent DNS queries that take place during the HTTP client call?
* How do we represent TLS handshakes?
* If more than one instrumentation library considers themselves the logical HTTP client, how should that be handled?

tedsuo:
@jkwatson the simplest (and cheapest) model would be a single span for a logical HTTP request with all of the data listed above added as events. Would that model work, or do you have examples of how would be painful?

jkwatson:
> @jkwatson the simplest (and cheapest) model would be a single span for a logical HTTP request with all of the data listed above added as events. Would that model work, or do you have examples of how would be painful?

I have no opinions on the matter...I just want to make sure we capture the recommendations of the experts.

The last question is a stickier one, though, that wouldn't be captured via events, I don't think.

tedsuo:
> If more than one instrumentation library considers themselves the logical HTTP client, how should that be handled?

Ideally, the higher level client should presume that instrumentation for the lower level client is present, and simply decorate the span that the lower level client produces with additional information. If no lower level instrumentation is available, the author of the higher-level instrumentation should create a separate instrumentation package for the lower lever client, rather than duplicate the work.

For situations where there is a categorical difference between the two client levels (DB client span vs HTTP client span) I'm less certain. There will definitely be a nested client spans when there is not a 1:1 relationship between DB and HTTP semantics. For example, some DB clients may make more than one logical HTTP request per DB transaction. In that case there would be a client span for the DB transaction, containing multiple HTTP client spans.

For databases/queues/etc where there is always 1:1 relationship between a DB transaction and an HTTP request; I'm tempted to say that the DB transaction should decorate the HTTP span rather than wrap it. Operation name would be the only potential conflict of interest in that case.

jkwatson:
> > If more than one instrumentation library considers themselves the logical HTTP client, how should that be handled?
> 
> Ideally, the higher level client should presume that instrumentation for the lower level client is present, and simply decorate the span that the lower level client produces with additional information. If no lower level instrumentation is available, the author of the higher-level instrumentation should create a separate instrumentation package for the lower lever client, rather than duplicate the work.

[...]

PR #1916, opened 2021-09-09
Define HTTP attributes that should be provided at span creation time

lmolkova:
## Changes

This change clarifies which HTTP attributes should be provided before making sampling decision: all attributes that are required and available before span starts.

Optional attributes are not provided at start time to avoid the overhead of calculating and adding them

**Note** it actually goes against current spec recommendation (see #620 for discussion):

> Whenever possible, users SHOULD set any already known attributes at span creation instead of calling SetAttribute later.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span-creation

Related issues #1747


Oberon00:
> Note it actually goes against current spec recommendation:
>
>> Whenever possible, users SHOULD set any already known attributes at span creation instead of calling SetAttribute later.
>
> https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span-creation
>
> And I would like to change that as well because it seems to be not very efficient

Please don't open new, mostly unrelated issues in the PR description 😃
I agree that there is a fundamental performance problem with the current sampling API, but this should be discussed in #620.

Oberon00:
Your change currently fails the semantic convention check. I see two possibilities:

1. Use the existing but so far unused `sampling_relevant: true` field in YAML to declare which attributes should be set at span creation time (that's what it was meant for). Then the markdown generation part of the semantic convention generator would need to be adapted to include that information, see https://github.com/open-telemetry/build-tools/blob/main/semantic-conventions/src/opentelemetry/semconv/templating/markdown/__init__.py
2. Manually write the "at creation time" restrictions outside the `<!--semconv-->` tags for now.

semantic_conventions/trace/http.yaml
```diff
@@ -19,10 +20,12 @@ groups:
...
           `http.url` MUST NOT contain credentials passed via URL in form of `https://username:password@www.example.com/`.
           In such case the attribute's value should be `https://www.example.com/`.
+        sampling_relevant: true
         examples: ['https://www.foo.bar/search?q=OpenTelemetry#SemConv']
       - id: target
         type: string
         brief: 'The full request target as passed in a HTTP request line or equivalent.'
+        sampling_relevant: true
         examples: ['/path/12314/?q=ddds#123']
       - id: host
```
Oberon00, line 30: Why is the host not sampling relevant?

Same goes for http.server_name / net.peer|host.name / net.peer|host.ip et al. which are alternatives to host.

Oberon00:
I agree to the semantic content of this PR, but it needs to be changed to pass the semantic convention check.

semantic_conventions/trace/http.yaml
jmacd: It looks like this PR is scoped to http.* variables, and http.server_name is now listed as "Yes, if provided".

Oberon00: http.server_name is potentially required, see the any_of constraint at the end.

lmolkova: it is sampling_relevant, see line 41 
https://github.com/open-telemetry/opentelemetry-specification/blob/d4430d39cbe4349bbd491ee1113a635555b20a13/semantic_conventions/trace/http.yaml#L30-L41

lmolkova: regarding `net.peer.*` attributes: I agree they should be provided, and I've just mentioned them here spec. I believe their population time requirements should be in the general spec too.  But until we generalize start time attributes beyond HTTP and categorize all general attributes, I would keep them in the HTTP spec (and maybe even after for extra clarity for anyone working on HTTP instrumentation).

Oberon00: I think you could `ref`  them and then override `sampling_relevant` just for the HTTP use. But that probably wouldn't make much sense.

lmolkova: thanks and I'll try to fix the build sometime this week.

bogdandrutu (approved):
Please fix the checks, everything else LGTM

specification/trace/semantic_conventions/http.md
```diff
@@ -84,6 +87,17 @@ Don't set the span status description if the reason can be inferred from `http.s
 | `2.0` | HTTP 2 |
 | `SPDY` | SPDY protocol. |
 | `QUIC` | QUIC protocol. |
+
+Following attributes MUST be provided **at span creation time** (when provided at all):
```
yurishkuro, line 91: MUST here seems an overkill to me. There are modes where tracing is collecting only skeleton information about the request execution, these attributes are not required in such modes.

arminru, line 91: I think we should give a hint on the reasoning for this requirement since it might not be obvious to everyone.
We could either adapt the wording produced here to say something like "_, so it can be considered for sampling decisions_" or refine the wording in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span-creation.
@lmolkova WDYT?
Either way, this can be done in a separate follow-up PR.

yurishkuro (dismissed):
The change sneaks MUST in the text

jmacd:
This doesn't look like "sneaking" to me, it's just a grammar question.

@yurishkuro are you suggesting to change

"following attributes MUST be provided **at span creation time** (when provided at all)"

to something like:

"If they will be provided during the lifetime of a span, the following attributes MUST be provided **at span creation time**."

?

jmacd (approved):

yurishkuro:
@jmacd the PR title says "should be provided", but the text says MUST, that's why I called it "sneaking in". 

I would not object to SHOULD, I do object to MUST, I think it is an unreasonable requirement.

jmacd:
"If condition is met, something MUST be." sounds equivalent to SHOULD, it's just more specific. Right?

lmolkova:
I'm sorry for any confusion, there was no intention to sneak anything, 

> There are modes where tracing is collecting only skeleton information about the request execution, these attributes are not required in such modes.

this seems to go against the current HTTP spec which requires you to populate attributes. If it doesn't populate even required HTTP attributes, it shouldn't be considered HTTP spec.

specification/trace/semantic_conventions/http.md
lmolkova: I agree and happy to do so once we reach an agreement on should/must. It has to be done as a separate PR in the build-tools repo (which generates everything within semconv block).

yurishkuro:
> this seems to go against the current HTTP spec which requires you to populate attributes.

Fair enough, but the spec does not *require* attributes to be provided on span creation. I am not even convinced that it is possible in all existing HTTP frameworks, there could be hooks for starting a span as early as the incoming packet, before anything from the request was parsed.

The use case in this ticket is to provide attributes to enable sampling decisions - that is already a special case that is not applicable to every situation,  e.g. a lot of deployments today are sampling randomly without paying any attention to span attributes.

So while I agree it is a good idea to provide these attributes as early as possible (SHOULD), I don't agree that it must be a requirement for instrumentation (not a MUST).

jmacd:
> there could be hooks for starting a span as early as the incoming packet, before anything from the request was parsed.

I believe in these circumstances, the Span start should be delayed until the required attributes are known, and for this reason we support setting the start time of the span explicitly. My understanding is that Luidmila's goal is essentially to ensure that Samplers have access to required HTTP attributes when sampling HTTP spans.

lmolkova:
Thanks @yurishkuro  and @jmacd for the explanation.

Yes, my goal is to make sure the sampling experience will be consistent across instrumentations. Assuming some instrumentations populate attributes later, make them incompatible with certain samples, which makes a bad user story.

If it's not feasible, then I agree with `SHOULD`.

@yurishkuro can you help me understand the case when a span starts with the first packet:
- you at least have to read headers (context) to start a span
- at this point, you've read the first request line as well and can likely populate other attributes (e.g. method, scheme, host, target)
what am I missing?

yurishkuro:
I think we're getting hung up on concrete use cases. I have a simpler question - if my sampling does not depend on http attributes in any way, why should instrumentation be required to supply them on span-start instead of when-convenient? There is performance overhead in providing attributes on-start since in most cases it requires allocation of a dictionary, instead of doing this:
```
sp := tracer.start("name")
if sp.IsRecording() {
    sp.SetAttribute("http.method", req.Method)
    ...
}
```
By using MUST we are adding a very sweeping requirement.

lmolkova:
> I think we're getting hung up on concrete use cases. I have a simpler question - if my sampling does not depend on http attributes in any way

This is where our disagreement is: if you ship distro and control instrumentations and samplers - everything is fine. Otherwise, users combine instrumentations and samplers and things get broken. 

I think we have to add another pre-sampling/early check to minimize extra allocations: https://github.com/open-telemetry/opentelemetry-specification/issues/620, which is an addition to the spec.

The problem I have with `SHOULD` is that instrumentations just do not follow it. Example:

> Whenever possible, users SHOULD set any already known attributes at span creation instead of calling SetAttribute later.
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span-creation

There is no consistency around it - everyone does its own thing.

lmolkova:
I.e. how I think optimization should look like:

```
if tracer.shouldStart("name", ...) {
    sp:= start tracer.start("name", ..., sampling_relevant_attributes)
    sp.IsRecording() {
        sp.SetAttribute(optional attributes)
         ...
   }
```

jmacd:
If we want to optimize the sampling API, we should take it up in #620 and let this PR proceed. A lazy-attribute mechanism is how I'd do this, e.g.,

```
    sp:= tracer.start("name", ..., tracing.WithLazyAttribute("http.path", func() attribute.Value { return ... }))
    // ...
    sp.SetAttribute(attribute.LazyString("http.thing", func() attribute.Value { return ... }))
    if sp.IsRecording() {
        expensiveVar := expensiveMethod()
        calculated := doLogic(expensiveVar)
        sp.SetAttribute(attribute.String("http.expensive", calculated))
    }
```

#620 is saying we should be able to refactor the Sampler API to allow the Sampler to include this "shouldSample(name)" logic and selectively evaluate any of the lazy span attributes in order to make its decision. Most call-sites would be able to use the lazy attribute and avoid a call to `sp.IsRecording()`, if we had such a thing, like the example above.

yurishkuro:
> The problem I have with SHOULD is that instrumentations just do not follow it. 

So either they don't care about following the spec (MUST wouldn't change that), or they have reasons (MUST would break them without a path forward).

lmolkova:
> So either they don't care about following the spec (MUST wouldn't change that), or they have reasons (MUST would break them without a path forward).

my assumptions about reasons (and I own Azure SDK http instrumentation, so know a bit about it):
- historic implementations - follow old versions and owners not  aware of sampling and creation time attribute requirements
- not-optimal (optional attributes)
- not-optimal (no early check)

Let me give you another reason for MUST  (a bit beyond sampling): calculating metrics without a new instrumentation code. Is it worth a `MUST` assuming #620 solves perf concerns (however it does it)?

yurishkuro:
> Is it worth a MUST assuming #620 solves perf concerns (however it does it)?

Are we prepared to park this PR until #620 (which is ~18mo old) is resolved?

lmolkova:
> > Is it worth a MUST assuming #620 solves perf concerns (however it does it)?
> 
> Are we prepared to park this PR until #620 (which is ~18mo old) is resolved?

I'm happy to change it to `SHOULD (please, it will be a MUST soon)` until #620 is resolved. Do you agree that it's the only blocker for `MUST`?

yurishkuro:
Sorry, but I still haven't seen a convincing argument why this needs to be a requirement. I cannot overcome the fact that the *only* reason you want to require these fields is to do sampling, and not _every conceivable form of sampling_, but a very specific form that most people do not actually use today. And even if we start talking about attribute-based sampling, there can be plenty of other attributes could be way more valuable in sampling decisions than these http attributes, so why require these specific ones? This simply does not pass the bar for me for a narrow use case to be elevated to a requirement. I think you'd get as much mileage from SHOULD as from MUST.

> calculating metrics without a new instrumentation code

Of the three RED metrics, error and duration cannot be calculated until span is finished, so it doesn't matter when span attributes are set.

lmolkova:
I don't care much about this change, I want to have clear rules for implementations to follow.

> And even if we start talking about attribute-based sampling, there can be plenty of other attributes could be way more valuable in sampling decisions than these http attributes, so why require these specific ones?

I'm trying to move HTTP spec forward, not sampling story.  I want to have a good, consistent  e2e story for users. Lack of clarity in this spec causes inconsistencies in instrumentations - MUST is a way to provide clarity.

> Of the three RED metrics, error and duration cannot be calculated until span is finished, so it doesn't matter when span attributes are set.

For metrics, we don't even have to start a real span if it's sampled out - we can have a lightweight one that only supports a few required attributes, provided at start time and measures duration. But you're right that all required attributes indeed can be provided after start for it to work.

iNikem:
> a very specific form that most people do not actually use today.

Please be careful with such sweeping statements. In my pond the situation is the opposite: nobody cares about probabilistic sampling but a lot of people want attribute-based sampling (or Views, as you called them in another issue).

anuraaga:
For better or worse, the sample method accepts `Attributes`. I agree with @iNikem that this is an important use case - but without a consistent base then that use case basically gets blocked. If some library, or some language even, does not populate an attribute while others do, setting up any sort of attribute-based sampling becomes a nightmare for users, if it's even possible.

It seems to be the same as the `required` field for conventions - presumably they are defined so that backends have a consistent base to work with, even though not all backends will actually use them all. But that base needs to be defined. I don't think I see a definition for what `required` actually means in RFC-speak - if `required` is a MUST, I'd expect this definition to also be MUST, or otherwise both could be SHOULD. 

> There is performance overhead in providing attributes on-start since in most cases it requires allocation of a dictionary, instead of doing this:

This comment specifically seems to point out the dangers - we can't provide a consistent experience to users if one language SIG happens to deem the performance implication too high and doesn't implement these while other languages do.

carlosalberto (approved):

specification/trace/semantic_conventions/http.md
lmolkova: @yurishkuro based on the new input in this PR, I'm going to resolve this conversation (but will keep the PR stay open for a bit more). Feel free to unresolve.

lmolkova: addressed in https://github.com/open-telemetry/build-tools/pull/70

bogdandrutu, line 91: I agree with @yurishkuro that SHOULD here makes more sense, see https://datatracker.ietf.org/doc/html/rfc2119#section-3 which covers that in normal circumstances this will be the case unless special case requires to not do this.

lmolkova: then we probably should use SHOULD everywhere? The required attributes SHOULD be populated or MUST? Help me understand why we use one or another.

Let's say if we write an instrumentation certification kind of a test, should we fail if SHOULD requirement is not satisfied? Perhaps not, but if it's MUST, we should fail.

Why do we want instrumentations to have some wiggle room here? 
Do we believe it's not feasible?
Or this scenario of attribute-based sampling is more important than perf gains (that probably would dissapear for metrics collection anyway)?

lmolkova: From some conversations I had about implementing this spec, SHOULD is not strong enough to force people even if something is totally possible. 

This is an experimental spec and we're working on stabilizing it, I assume implementing the new version in current contrib repos in multiple languages would be a prereq for stabilization. 

Can we try with MUST and see if it's doable during the stabilization and implementation? We can change it to SHOULD after. Switching from SHOULD to MUST is a bigger deal.

Oberon00: You are thinking about the instrumentation (producer) side. For the sampler or backend (consumer) side, switching from MUST to SHOULD would be the bigger deal.

Oberon00 (approved):

specification/trace/semantic_conventions/http.md
lmolkova: agreed. I believe an attempt to go with MUST works best for backends. But If it's not possible and we have to switch to SHOULD, it's still possible while spec is experimental.

pyohannes (approved):

carlosalberto:
Merging as further details will be decided as part of #2011, as discussed in the previous Spec SIG call.

PR #1917, opened 2021-09-10
Metrics SDK fixes

reyang:
Addressing the non-blocking comments / minor issues (these are nice to have, and won't block the SDK spec Experimental release):
* Addressed https://github.com/open-telemetry/opentelemetry-specification/pull/1888#discussion_r704923731 from @cijothomas
* Updated the ToC to reflect changes from #1828.

### DIFF
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -747,7 +747,7 @@ modeled to interact with other components in the SDK:
   ```text
   +-----------------+            +-----------------------------+
   |                 | Metrics... |                             |
-  | In-memory state +------------> Base exporting MetricReader |
+  | In-memory state +------------> Exporting MetricReader      |
   |                 |            |                             |
   +-----------------+            |  +-----------------------+  |
                                  |  |                       |  |
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -25,11 +25,16 @@ Table of Contents
 * [MeterProvider](#meterprovider)
 * [Attribute Limits](#attribute-limits)
 * [MeasurementProcessor](#measurementprocessor)
+* [Exemplar](#exemplar)
+  * [Exemplar Filter](#exemplar-filter)
+  * [Exemplar Reservoir](#exemplar-reservoir)
+  * [Exemplar Defaults](#exemplar-defaults)
 * [MetricReader](#metricreader)
   * [Periodic exporting MetricReader](#periodic-exporting-metricreader)
 * [MetricExporter](#metricexporter)
   * [Push Metric Exporter](#push-metric-exporter)
   * [Pull Metric Exporter](#pull-metric-exporter)
+* [Defaults and Configuration](#defaults-and-configuration)
 
 </details>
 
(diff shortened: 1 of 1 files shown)


cijothomas (approved):

jmacd (approved):

jsuereth (approved):

specification/metrics/sdk.md
```diff
@@ -440,7 +445,7 @@ A Metric SDK SHOULD provide extensible hooks for Exemplar sampling, specifically
 - `ExemplarFilter`: filter which measurements can become exemplars
 - `ExemplarReservoir`: determine how to store exemplars.
 
-### Exemplar Filter
+### ExemplarFilter
```
bogdandrutu, line 448: Not related to this, should we consider `ExemplarSampler` since we use the "sampled" word?

reyang: I actually like this proposal 👍

reyang: @jsuereth what's your take?

jsuereth: I'm calling an `ExemplarSampler` a combination of both the reservoir + filter configuration.  I can update this to denote that concept if you'd like.

The reason we've torn that concept in two is to make it easy to "sample w/ sampled traces" vs. "on/off" behavior.  I.e. unlike a regular sampler implementation we ALSO want filtering as a separate configuration parameter (hence the specification around it).


The reservoir itself will make further sampling decisions beyond the filter, hence calling JUST the filter a "sampler" being a bit disingenuous to what's really going on.

bogdandrutu: Do we need both concepts exposed from the SDK API? Asking because this may be an implementation detail

jmacd: I think it's OK to use the term "sampler".  In a an OTEP-170 revision here https://github.com/open-telemetry/oteps/pull/175 I've categorized samplers as probability and non-probability samplers. A filter can be called a non-probability sampler, it's fine.

jmacd (approved):

specification/metrics/sdk.md
jsuereth: At a minimum, we expose quick-access to the filter via configuration (NEVER, ALL or SAMPLE_WITH_TRACE).   This was the primary reason to expose this, and I think the ability to quickly configure exemplars is important enough to require.

At a maximum, the ExemplarReservoir is a per-metric-stream reservoir that gets pre-filtered measurements.

An ExemplarSampler (and I can update the specification for this), is the thing which decides (upon understanding there is a metric stream being collected) which Reservoir to instantiate *and* what filter to attach on the reservoir.

I specifically think we want these three concepts, but the names can be changed.  However, if you think of "ExemplarSampler" as "the thing which determines which measurements are sampled into Exemplars" it's the whole kit + kaboodle of "The factory that decides which ExemplarReservoir belongs to which metric stream" and "The measurement prefilter".

bogdandrutu: Indeed I think as a whole, but let's move this discussion to an issue, no point in polluting this PR.

jsuereth: sounds good.

PR #1921, opened 2021-09-15
Discourage faas.trigger for client/outgoing spans.

Oberon00:
## Changes

faas.trigger does not really make sense for outgoing calls. Change wording to discourage its use for client spans..

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,8 @@ release.
   ([#1810](https://github.com/open-telemetry/opentelemetry-specification/pull/1810))
 - Clarifications for `http.client_ip` and `http.host`.
   ([#1890](https://github.com/open-telemetry/opentelemetry-specification/pull/1890))
+- Change meaning and discourage use of `faas.trigger` for FaaS clients (outgoing).
+  ([#????](https://github.com/open-telemetry/opentelemetry-specification/pull/????))
 
 ### Compatibility
 
--- a/semantic_conventions/trace/faas.yaml
+++ b/semantic_conventions/trace/faas.yaml
@@ -10,10 +10,12 @@ groups:
         required:
           conditional: >
             On FaaS instances, faas.trigger MUST be set on incoming invocations.
-            Clients invoking FaaS instances MUST set `faas.trigger` on outgoing
-            invocations, if it is known to the client. This is, for example, not
-            the case, when the transport layer is abstracted in a FaaS client
-            framework without access to its configuration.
+            Clients invoking FaaS instances usually cannot set `faas.trigger`,
+            since they would typically need to look in the payload to determine
+            the event type. If clients set it, it should be the same as the
+            trigger that corresponding incoming would have (i.e., this has
+            nothing to do with the underlying transport used to make the API
+            call to invoke the lambda, which is often HTTP).
         brief: 'Type of the trigger on which the function is executed.'
         type:
           allow_custom_values: false
--- a/specification/trace/semantic_conventions/faas.md
+++ b/specification/trace/semantic_conventions/faas.md
@@ -37,7 +37,7 @@ If Spans following this convention are produced, a Resource of type `faas` MUST 
 | `faas.trigger` | string | Type of the trigger on which the function is executed. | `datasource` | Conditional [1] |
 | `faas.execution` | string | The execution ID of the current function execution. | `af9d5aa4-a685-4c5f-a22b-444f80b3cc28` | No |
 
-**[1]:** On FaaS instances, faas.trigger MUST be set on incoming invocations. Clients invoking FaaS instances MUST set `faas.trigger` on outgoing invocations, if it is known to the client. This is, for example, not the case, when the transport layer is abstracted in a FaaS client framework without access to its configuration.
+**[1]:** On FaaS instances, faas.trigger MUST be set on incoming invocations. Clients invoking FaaS instances usually cannot set `faas.trigger`, since they would typically need to look in the payload to determine the event type. If clients set it, it should be the same as the trigger that corresponding incoming would have (i.e., this has nothing to do with the underlying transport used to make the API call to invoke the lambda, which is often HTTP).
 
 `faas.trigger` MUST be one of the following:
 


semantic_conventions/trace/faas.yaml
```diff
@@ -10,10 +10,12 @@ groups:
         required:
           conditional: >
             On FaaS instances, faas.trigger MUST be set on incoming invocations.
-            Clients invoking FaaS instances MUST set `faas.trigger` on outgoing
-            invocations, if it is known to the client. This is, for example, not
-            the case, when the transport layer is abstracted in a FaaS client
-            framework without access to its configuration.
+            Clients invoking FaaS instances usually cannot set `faas.trigger`,
```
Oberon00, line 13: The previous definition did not make any sense to me. On incoming calls, the trigger would probably depend on the event type. But on outgoing calls it would suddenly be the protocol used to make the API call that invokes the lambda?

Even if I make a HTTP REST call to AWS to invoke a lambda, I can (typically will) pass something different from an API Gateway (HTTP) event, e.g. a custom event type, that would result in `other` being set as trigger on the incoming side.

CHANGELOG.md
```diff
@@ -48,6 +48,8 @@ release.
   ([#1810](https://github.com/open-telemetry/opentelemetry-specification/pull/1810))
 - Clarifications for `http.client_ip` and `http.host`.
   ([#1890](https://github.com/open-telemetry/opentelemetry-specification/pull/1890))
+- Change meaning and discourage use of `faas.trigger` for FaaS clients (outgoing).
+  ([#????](https://github.com/open-telemetry/opentelemetry-specification/pull/????))
```
arminru, line 52:
```suggestion
  ([#1921](https://github.com/open-telemetry/opentelemetry-specification/pull/1921))
```
!!!!

semantic_conventions/trace/faas.yaml
```diff
@@ -10,10 +10,12 @@ groups:
...
+            since they would typically need to look in the payload to determine
+            the event type. If clients set it, it should be the same as the
+            trigger that corresponding incoming would have (i.e., this has
+            nothing to do with the underlying transport used to make the API
+            call to invoke the lambda, which is often HTTP).
```
arminru, line 18: I think it would make sense to have that clarification in the attribute's `brief` or `note` rather than it being "hidden" in the `required` attribute that might not show up in generated API docs, for example.

semantic_conventions/trace/faas.yaml
```diff
@@ -10,10 +10,12 @@ groups:
...
         brief: 'Type of the trigger on which the function is executed.'
         type:
           allow_custom_values: false
```
arminru, line 21: (reference to line 26 below)
I noticed that the `brief` for `http` does not really make sense from a client's perspective ("To provide an answer to an inbound HTTP request"). Should we update that as well (even though it's mostly obsolete with this change here)?

Oberon00: I don't think changing that helps much .The name of the convention already does not make sense because from a client perspective it is not a "trigger". It would make more sense as "event_type".

semantic_conventions/trace/faas.yaml
Oberon00: I added 032c08e1379c25c3922b74d61b27ba36f20a594f in which I also `ref` the trigger in `faas_span.in` with `required: true`. I'm not sure if I'm happy with the resulting markdown, but I think from a model perspective it makes sense.

specification/trace/semantic_conventions/faas.md
```diff
@@ -87,6 +95,17 @@ For incoming FaaS spans, the span kind MUST be `Server`.
 | Attribute  | Type | Description  | Examples  | Required |
 |---|---|---|---|---|
 | `faas.coldstart` | boolean | A boolean that is true if the serverless function is executed for the first time (aka cold-start). |  | No |
+| `faas.trigger` | string | Type of the trigger which caused this function execution. [1] | `datasource` | Yes |
```
Oberon00, line 98: Note the "Yes" in the Required column.

Oberon00:
@anuraaga Maybe you can review this?

Oberon00:
If nobody has an opinion on this, I would actually say that's a sign in favor of this PR. We can always put back recommendations to use it in some way, recommending to not use it for client-side spans seems like a safe thing to do for now if nobody knows what it is about.

semantic_conventions/trace/faas.yaml
```diff
@@ -7,14 +7,17 @@ groups:
...
-            invocations, if it is known to the client. This is, for example, not
-            the case, when the transport layer is abstracted in a FaaS client
-            framework without access to its configuration.
-        brief: 'Type of the trigger on which the function is executed.'
+        brief: 'Type of the trigger which caused this function execution.'
+        note: |
+          For the server/consumer span on the incoming side,
+          `faas.trigger` MUST be set.
+
+          Clients invoking FaaS instances usually cannot set `faas.trigger`,
```
anuraaga, line 15: This makes sense to me. There doesn't seem to be a generic way of populating these

To provide an example in terms of AWS, `faas.trigger` would be populated  by the AWS SDK instrumentation of the `Lambda.Invoke` API. In this case, the correct setting of `faas.trigger` would be

`datasource` if the AWS SDK invocation happens to be made from the S3 service itself (AWS managed code)
`pubsub` if the AWS SDK invocation happens to be made from the SQS service itself (AWS managed code)
`http` if the AWS SDK invocation happens to be made from the API Gateway service itself (AWS managed code)

An end user never actually creates a span for a faas span in these cases,  all they do is configure the trigger in the backend, and the app itself is independent from it.


If there was a trigger `manual`, then that could actually be populated by AWS SDK instrumentation by default though - almost all AWS SDK invocations of `Lambda.Invoke` outside of these servers is a manual invocation without a trigger. But such a trigger doesn't exist - presumably that lack of populating something indicates "manual", in which case indeed this PR is definitely reinforcing that.

That being said, it could be OK to add text encouraging cloud service providers to set trigger on client spans if this seems like it can add value, but generic instrumentation shouldn't be doing so.

anuraaga (approved):

PR #1923, opened 2021-09-15
Allow implementations to use their own default for OTLP compression.

carlosalberto:
As discussed in yesterday's Spec SIG call, allow SIGs to define their own default. Specially important as OTel JS will be going GA very soon, and we want to allow it to use gzip as their default (at least for the browser case) - also a motivating reason to include this change in this week's Specification release.

(Also removed the mention of SIGs supporting other compression methods, as they should be vetted first by the Collector, etc)

Alternative to #1895 

cc @dyladan @bogdandrutu

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@ release.
 ### OpenTelemetry Protocol
 
 - Add environment variables for configuring the OTLP exporter protocol (`grpc`, `http/protobuf`, `http/json`) ([#1880](https://github.com/open-telemetry/opentelemetry-specification/pull/1880))
+- Allow SIGs to use their own default for OTLP compression ([#1923](https://github.com/open-telemetry/opentelemetry-specification/pull/1923))
 
 ### SDK Configuration
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,7 +58,8 @@ release.
 ### OpenTelemetry Protocol
 
 - Add environment variables for configuring the OTLP exporter protocol (`grpc`, `http/protobuf`, `http/json`) ([#1880](https://github.com/open-telemetry/opentelemetry-specification/pull/1880))
-- Allow SIGs to use their own default for OTLP compression ([#1923](https://github.com/open-telemetry/opentelemetry-specification/pull/1923))
+- Allow implementations to use their own default for OTLP compression, with `none` denotating no compression
+  ([#1923](https://github.com/open-telemetry/opentelemetry-specification/pull/1923))
 
 ### SDK Configuration
 
--- a/specification/protocol/exporter.md
+++ b/specification/protocol/exporter.md
@@ -20,7 +20,7 @@ The following configuration options MUST be available to configure the OTLP expo
 | Protocol             | The transport protocol. Options MAY include `grpc`, `http/protobuf`, and `http/json`. See [Specify Protocol](./exporter.md#specify-protocol) for more details. | n/a               | `OTEL_EXPORTER_OTLP_PROTOCOL` `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL` |
 
 **[1]**: If no compression value is explicitly specified, SIGs can default to the value they deem
-most useful among the supported options. This is specially important in presence of technical constraints,
+most useful among the supported options. This is especially important in the presence of technical constraints,
 e.g. directly sending telemetry data from mobile devices to backend servers.
 
 Supported values for `OTEL_EXPORTER_OTLP_*COMPRESSION` options:
--- a/specification/protocol/exporter.md
+++ b/specification/protocol/exporter.md
@@ -25,7 +25,7 @@ e.g. directly sending telemetry data from mobile devices to backend servers.
 
 Supported values for `OTEL_EXPORTER_OTLP_*COMPRESSION` options:
 
-- `` if compression is disabled.
+- `none` if compression is disabled.
 - `gzip` is the only specified compression method for now.
 
 Example 1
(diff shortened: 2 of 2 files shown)


codeboten (approved):
👍

dyladan (approved):
This looks good from JS perspective. Sorry for late review i had vacation

bogdandrutu (approved):

tigrannajaryan:
As discussed in the SIG, I think the best that we do this:
- The OTLP [protocol spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md) to require all OTLP receivers to support both "none" and "gzip" compression.
- The SDK spec to allow SDKs exporters to use "none" or "gzip" as the default compression method when OTEL_EXPORTER_OTLP_TRACES_COMPRESSION env variable is unspecified. The choice of the default is language-specific.

carlosalberto:
@tigrannajaryan Updated it to match the feedback, please review again.

@dyladan FYI, now the new explicit `none` option exists (besides `gzip`).

specification/protocol/exporter.md
```diff
@@ -15,14 +15,18 @@ The following configuration options MUST be available to configure the OTLP expo
...
 
+**[1]**: If no explicit compression value is specified, SIGs can default to the value they deem
+most useful among the supported options. This is specially important in presence of technical constraints,
+e.g. directly sending telemetry data from mobile devices to backend servers.
+
 Supported values for `OTEL_EXPORTER_OTLP_*COMPRESSION` options:
 
-- If the value is missing, then compression is disabled.
-- `gzip` is the only specified compression method for now. Other options MAY be supported by language SDKs and should be documented for each particular language.
+- `none` if compression is disabled.
```
tigrannajaryan, line 28: I think it is best to keep it the old way, i.e. missing value indicates no compression. This is I think how current implementations do it, and it is also how Collector does it.

tigrannajaryan (commented):
LGTM, except the "none" part.

specification/protocol/exporter.md
carlosalberto: You mean, the explicit "" literal, right?

tigrannajaryan: Yes, not sure how we normally specify empty strings.

carlosalberto: I adjusted this to use the "" literal. Looks about right?

specification/protocol/exporter.md
```diff
@@ -15,14 +15,18 @@ The following configuration options MUST be available to configure the OTLP expo
...
+| Compression          | Compression key for supported compression types. Supported compression: `gzip`| No value [1]          | `OTEL_EXPORTER_OTLP_COMPRESSION` `OTEL_EXPORTER_OTLP_TRACES_COMPRESSION` `OTEL_EXPORTER_OTLP_METRICS_COMPRESSION` |
 | Timeout              | Maximum time the OTLP exporter will wait for each batch export. | 10s               | `OTEL_EXPORTER_OTLP_TIMEOUT` `OTEL_EXPORTER_OTLP_TRACES_TIMEOUT` `OTEL_EXPORTER_OTLP_METRICS_TIMEOUT` |
 | Protocol             | The transport protocol. Options MAY include `grpc`, `http/protobuf`, and `http/json`. See [Specify Protocol](./exporter.md#specify-protocol) for more details. | n/a               | `OTEL_EXPORTER_OTLP_PROTOCOL` `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL` |
 
+**[1]**: If no compression value is explicitly specified, SIGs can default to the value they deem
+most useful among the supported options. This is specially important in presence of technical constraints,
```
yurishkuro, line 23:
```suggestion
most useful among the supported options. This is especially important in presence of technical constraints,
```

specification/protocol/exporter.md
```diff
@@ -15,14 +15,18 @@ The following configuration options MUST be available to configure the OTLP expo
...
+e.g. directly sending telemetry data from mobile devices to backend servers.
+
 Supported values for `OTEL_EXPORTER_OTLP_*COMPRESSION` options:
 
-- If the value is missing, then compression is disabled.
-- `gzip` is the only specified compression method for now. Other options MAY be supported by language SDKs and should be documented for each particular language.
+- `` if compression is disabled.
```
yurishkuro, line 28: I am not sure what this means. `set COMPRESSION=''`? Why should this be different from `unset COMPRESSION`?

Since we need explicit statement to disable compression, I would go with `set COMPRESSION=none` over `set COMPRESSION=''`

tigrannajaryan: > I am not sure what this means. `set COMPRESSION=''`? Why should this be different from `unset COMPRESSION`?

Unset (non-existing) env variable is distinguishable from an env variable that is set to empty string. So, I believe there is a difference even if it is not immediately clear.

> Since we need explicit statement to disable compression, I would go with `set COMPRESSION=none` over `set COMPRESSION=''`

The only problem with this is that I think existing implementations don't look for "none", they look for empty string (AFAIK, that's the case with the corresponding Collector config setting). If we change it to "none" it is a breaking change, isn't it? (Although I imagine most user simply did not set any value, likely noone tried to set it to empty string, so realistically perhaps we are not breaking anyone).

carlosalberto: Supporting "" but deprecating it, while making "none" the official value would a decent enough solution?

tigrannajaryan: > Supporting "" but deprecating it, while making "none" the official value would a decent enough solution?

I think it is doable but there may be multiple cases on how the deprecation should be done depending on whether:
- the implementation was previously differentiating between empty string and unset env var.
- what was the default value when the env var was unset.

carlosalberto:
@tigrannajaryan @yurishkuro  Used `none` for no compression now. Please confirm so we can go ahead and merge.

jsuereth (approved):

MSNev (approved):

tigrannajaryan (approved):

CHANGELOG.md
```diff
@@ -58,6 +58,7 @@ release.
 ### OpenTelemetry Protocol
 
 - Add environment variables for configuring the OTLP exporter protocol (`grpc`, `http/protobuf`, `http/json`) ([#1880](https://github.com/open-telemetry/opentelemetry-specification/pull/1880))
+- Allow SIGs to use their own default for OTLP compression ([#1923](https://github.com/open-telemetry/opentelemetry-specification/pull/1923))
```
tigrannajaryan, line 61: Perhaps also mention that "none" now stands for "no compression".

bogdandrutu, line 61: Allow "implementations" not SIGs :) There may be other valid implementations of the protocol.

carlosalberto: Done.

carlosalberto: Done.

Issue #1932, opened 2021-09-16
Specify data model details for ExponentialHistogram


jmacd:
**What are you trying to achieve?**

The metrics protocol is about to add a new data point type, `ExponentialHistogram`. The protocol PR specified in https://github.com/open-telemetry/opentelemetry-proto/pull/322 is able to express certain combinations of data that are unreasonable or have no practical use at this time and implementations are not expected to support all inputs. Details of this sort belong in the data model, not comments in the protocol definition.

See https://github.com/open-telemetry/opentelemetry-proto/pull/322#issuecomment-917283619

tigrannajaryan:
@jmacd reassigning this to you since you are the expert.

PR #1933, opened 2021-09-16
Add detail about explicit histogram buckets and boundaries

jmacd:
While working on a fix for #1932, I noticed some details of the current explicit-boundary histogram are not spelled out in the data model on Histogram.

## Changes

States what it means when a histogram has no buckets. 
Discusses exclusive and inclusive boundaries.

Related issues #1932 
Related proto PR https://github.com/open-telemetry/opentelemetry-proto/pull/322

### DIFF
--- a/specification/metrics/datamodel.md
+++ b/specification/metrics/datamodel.md
@@ -398,6 +398,21 @@ denotes Delta temporality where accumulated event counts are reset to zero after
 and a new aggregation occurs. Cumulative, on the other hand, continues to
 aggregate events, resetting with the use of a new start time.
 
+Buckets and counts are optional.  A Histogram without buckets conveys
+a population in terms of only the sum and count, and may be
+interpreted as a histogram with single bucket covering `(-Inf, +Inf)`.
+
+Bucket upper-bounds are inclusive while bucket lower-bounds are
+exclusive. That is, buckets express the number of values that are
+greater than their lower bound and less than or equal to their upper
+bound.  Importers and exporters working with OpenTelemetry Metrics
+data are meant to disregard this specification when translating to and
+from histogram formats that use inclusive lower bounds and exclusive
+upper bounds.  Changing the inclusivity and exclusivity of bounds is
+an example of worst-case Histogram error; users should choose
+Histogram boundaries so that worst-case error is within their error
+tolerance.
+
 ### Summary (Legacy)
 
 [Summary](https://github.com/open-telemetry/opentelemetry-proto/blob/v0.9.0/opentelemetry/proto/metrics/v1/metrics.proto#L268)


specification/metrics/datamodel.md
```diff
@@ -398,6 +398,21 @@ denotes Delta temporality where accumulated event counts are reset to zero after
 and a new aggregation occurs. Cumulative, on the other hand, continues to
 aggregate events, resetting with the use of a new start time.
 
+Buckets and counts are optional.  A Histogram without buckets conveys
+a population in terms of only the sum and count, and may be
+interpreted as a histogram with single bucket covering `(-Inf, +Inf)`.
+
+Bucket upper-bounds are inclusive while bucket lower-bounds are
```
reyang, line 405: Probably nit picking:

```suggestion
Bucket upper-bounds are inclusive (except for the case where the upper-bound is +Inf) while bucket lower-bounds are
```

specification/metrics/datamodel.md
```diff
@@ -398,6 +398,21 @@ denotes Delta temporality where accumulated event counts are reset to zero after
...
+exclusive. That is, buckets express the number of values that are
+greater than their lower bound and less than or equal to their upper
+bound.  Importers and exporters working with OpenTelemetry Metrics
+data are meant to disregard this specification when translating to and
```
reyang, line 409: Thanks for calling this out! 👍 💯

reyang (approved):
LGTM.

specification/metrics/datamodel.md
jsuereth: We could (for consistency, just include +Inf] there.  I don't see any reason not to...

specification/metrics/datamodel.md
jsuereth: +100

jsuereth (approved):

specification/metrics/datamodel.md
```diff
@@ -398,6 +398,21 @@ denotes Delta temporality where accumulated event counts are reset to zero after
 and a new aggregation occurs. Cumulative, on the other hand, continues to
 aggregate events, resetting with the use of a new start time.
 
+Buckets and counts are optional.  A Histogram without buckets conveys
```
bogdandrutu, line 401: Is `Count` optional? What do we keep inside then? Probably you want to say bucket counts not the count.

jmacd: Yes, fixed.

specification/metrics/datamodel.md
jmacd: I think we should say whether Inf values are permitted at the API level. 
NaN values too. I think we should disallow these special cases, then it doesn't matter whether infinite bounds are inclusive.

jmacd: (Accepted this suggestion.) 
https://github.com/open-telemetry/opentelemetry-specification/issues/1936

bogdandrutu (approved):

jsuereth (approved):

PR #1935, opened 2021-09-16
Add ExponentialHistogram to Metrics data model

### LINKED ISSUE #1932: Specify data model details for ExponentialHistogram

jmacd:
Fixes #1932.

## Changes

Adds statements describing the ExponentialHistogram data point introduced by https://github.com/open-telemetry/opentelemetry-proto/pull/322. 

This is meant to capture details about acceptable precision limits, how to handle out-of-range values for producers and consumers. See https://github.com/open-telemetry/opentelemetry-proto/pull/322#issuecomment-917283619

Related issues #

### DIFF
--- a/specification/metrics/datamodel.md
+++ b/specification/metrics/datamodel.md
@@ -398,6 +398,82 @@ denotes Delta temporality where accumulated event counts are reset to zero after
 and a new aggregation occurs. Cumulative, on the other hand, continues to
 aggregate events, resetting with the use of a new start time.
 
+### ExponentialHistogram
+
+[ExponentialHistogram](TBD_after_PR322_merges) data points are an
+alternate representation, like the [Histogram](#histogram) that convey
+a population of recorded measurements in a compressed format.
+ExponentialHistogram compresses bucket boundaries using an exponential
+formula, making it suitable for conveying high-resolution histogram
+data with a large number of buckets.
+
+Statements about `Histogram` that refer to aggregation temporality,
+attributes, timestamps, as well as the `sum`, `count`, `exemplars`
+fields are identical for `ExponentialHistogram`.
+
+The resolution of the ExponentialHistogram is characterized by a
+parameter known as `scale`, with larger values of `scale` offering
+greater precision.  Bucket boundaries of the ExponentialHistogram are
+located at integer powers of the `base`, where:
+
+```
+base = 2**(2**(-scale))
+```
+
+The symbol `**` in these formulas represents exponentiation, thus
+`2**x` is read "Two to the power of X", typically computed by an
+expression like `math.Pow(2.0, x)`.  Calculated `base` values for
+selected scales are shown below:
+
+| Scale | Base              | Expression  |
+| --    | --                | --          |
+| 10    | 1.000677130693066 | 2**(1/1024) |
+| 9     | 1.001354719892108 | 2**(1/512)  |
+| 8     | 1.002711275050203 | 2**(1/256)  |
+| 7     | 1.005429901112803 | 2**(1/128)  |
+| 6     | 1.010889286051700 | 2**(1/64)   |
+| 5     | 1.021897148654117 | 2**(1/32)   |
+| 4     | 1.044273782427414 | 2**(1/16)   |
+| 3     | 1.090507732665258 | 2**(1/8)    |
+| 2     | 1.189207115002721 | 2**(1/4)    |
+| 1     | 1.414213562373095 | 2**(1/2)    |
+| 0     | 2                 | 2**1        |
+| -1    | 4                 | 2**2        |
+| -2    | 16                | 2**4        |
+| -3    | 256               | 2**8        |
+| -4    | 65536             | 2**16       |
+
+The ExponentialHistogram bucket identified by `index`, a signed
+integer, represents values in the population that are greater than or
+equal to `base**index` and less than `base**(index+1)`.
+
+The positive and negative ranges of the histogram are expressed
+separately.  Negative values are mapped by their absolute value
+into the negative range using the same scale as the positive range.
+
+Each range of the ExponentialHistogram data point uses a dense
+representation of the buckets, where a range of buckets is expressed
+as a single `offset` value, a signed integer, and an array of count
+values, where array element `i` represents the bucket count for bucket
+index `offset+i`.
+
+For a given range, positive or negative:
+
+- The absolute value 1.0 has bucket index `0`
+- Bucket index `0` counts measurements greater than or equal to 1.0 and less than `base`
+- Negative indexes correspond with absolute values less than 1.0.
+
+The ExponentialHistogram contains a special `zero_count` field
+containing is the count of values that are either exactly zero or
+within the region considered zero by the instrumentation at the
+tolerated degree of precision.  This bucket stores values that cannot
+be expressed using the standard exponential formula as well as values
+that have been rounded to zero.
+
+#### Producers and consumer requirements
+
+TODO: Work-In-Progress.
+
 ### Summary (Legacy)
 
 [Summary](https://github.com/open-telemetry/opentelemetry-proto/blob/v0.9.0/opentelemetry/proto/metrics/v1/metrics.proto#L268)


specification/metrics/datamodel.md
```diff
@@ -398,6 +398,82 @@ denotes Delta temporality where accumulated event counts are reset to zero after
 and a new aggregation occurs. Cumulative, on the other hand, continues to
 aggregate events, resetting with the use of a new start time.
 
+### ExponentialHistogram
```
jsuereth, line 401: This should have a stability label.

jmacd: Fixed. I marked this "experimental" is that right?

jmacd:
@open-telemetry/specs-metrics-approvers this is ready for review.

jmacd:
FYI I have posted reference implementations that calculate bucket indexes described by the formulas here. These were captured from the review thread of https://github.com/open-telemetry/opentelemetry-proto/pull/322 here, now available:
https://github.com/open-telemetry/oteps/pull/179

specification/metrics/datamodel.md
```diff
@@ -398,14 +404,153 @@ denotes Delta temporality where accumulated event counts are reset to zero after
...
+representation of the buckets, where a range of buckets is expressed
+as a single `offset` value, a signed integer, and an array of count
+values, where array element `i` represents the bucket count for bucket
+index `offset+i`.
+
+For a given range, positive or negative:
+
+- Bucket index `0` counts measurements in the range `[1, base)`
+- Positive indexes correspond with absolute values greater or equal to `base`
+- Negative indexes correspond with absolute values less than 1
```
jamesmoessis, line 486: Question: Assuming bucket counts are represented by a list, how would you expect implementations to represent bucket counts with a negative index? Would this be achieved by adjusting the `offset` according to the lowest measured value?

specification/metrics/datamodel.md
```diff
@@ -398,14 +404,153 @@ denotes Delta temporality where accumulated event counts are reset to zero after
...
+- There are `2**scale` buckets between successive powers of 2.
+
+For example, with `scale=3` there are `2**3` buckets between 1 and 2.
+Note that the lower boundary for bucket index 4 in a `scale=3`
+histogram maps into the lower boundary for bucket index 2 in a
+`scale=2` histogram and maps into the lower boundary for bucket index
+1 (i.e., the `base`) in a `scale=1` histogram.
```
yzhuge, line 493: maybe add "This is an example of perfect subsetting"

specification/metrics/datamodel.md
```diff
@@ -398,14 +404,153 @@ denotes Delta temporality where accumulated event counts are reset to zero after
...
+containing the count of values that are either exactly zero or within
+the region considered zero by the instrumentation at the tolerated
+level of precision.  This bucket stores values that cannot be
+expressed using the standard exponential formula as well as values
+that have been rounded to zero.
+
+#### Producer expectations
+
+The ExponentialHistogram design makes it possible to express values
+that are too large or small to be represented in computer hardware.
```
yzhuge, line 518: It's more like "computer software". Suggest changing "computer hardware" to "the 64 bit "double" floating point format"

specification/metrics/datamodel.md
```diff
@@ -398,14 +404,153 @@ denotes Delta temporality where accumulated event counts are reset to zero after
...
+Certain values for `scale`, while meaningful, are not necessarily
+useful.
+
+The range of data represented by an ExponentialHistogram determines
+which scales can be usefully applied.  Therefore, producers SHOULD
+ensure that bucket indices are within the range of a signed 64-bit
```
yzhuge, line 524: you mean "32 bit (because we currently use 32 bit for index offset in protocol)"? 
The issue here is that even with the 32bit offset, a producer can still produce an index beyond 32 bit. For example, when offset is close to 32 bit max, then offset + arrayIndex will still go above 32 bit max. 

Shall we say something about such index overflow in consumer section?

specification/metrics/datamodel.md
```diff
@@ -398,14 +404,153 @@ denotes Delta temporality where accumulated event counts are reset to zero after
...
+integer by changing scale.
+
+Producers MAY use a built-in logarithm function to calculate the
+bucket index of a value.  The use of a built-in logarithm function
+could lead to results that differ from the bucket index that would be
+computed using arbitrary precision or a lookup table, however
+producers are not required to perform an exact computation.  As a
+result, ExponentialHistogram exemplars could map into buckets with
+zero count.  Instead, we expect to find such values counted in the
```
yzhuge, line 533: "instead" seems extraneous.

specification/metrics/datamodel.md
```diff
@@ -398,14 +404,153 @@ denotes Delta temporality where accumulated event counts are reset to zero after
...
+adjacent bucket.
+
+#### Consumer expectations
+
+ExponentialHistogram buckets are expected to map into numbers that can be
+represented using normalized IEEE 754 double-width floating point
+values (i.e., subnormal values are excluded).  Consumers SHOULD reject
```
yzhuge, line 540: Subnormal exclusion here seems odd. We didn't talk about it anywhere else in this doc, most importantly, not in the producer section. I'd like to remove this exclusion. If we have to say something about subnormals, we can say:
Consumer may choose to round subnormals into the zero_count bucket, or reject such numbers, if it is incapable of handling subnormal numbers.

specification/metrics/datamodel.md
yzhuge: "adjusting the offset according to the lowest measured value" is certainly one way. But it involves copying the array when offset changes. Another way is [WindowedCounterArray](https://github.com/newrelic-experimental/newrelic-sketch-java/blob/main/src/main/java/com/newrelic/nrsketch/WindowedCounterArray.java), used in [NrSketch](https://github.com/newrelic-experimental/newrelic-sketch-java#windowedcounterarray). It automatically adjusts to the incoming data's index range without shifting or copying the array.

specification/metrics/datamodel.md
```diff
@@ -398,14 +404,153 @@ denotes Delta temporality where accumulated event counts are reset to zero after
...
+
+The ExponentialHistogram design makes it possible to express values
+that are too large or small to be represented in computer hardware.
+Certain values for `scale`, while meaningful, are not necessarily
+useful.
+
+The range of data represented by an ExponentialHistogram determines
+which scales can be usefully applied.  Therefore, producers SHOULD
+ensure that bucket indices are within the range of a signed 64-bit
+integer by changing scale.
```
jmacd, line 525: @yzhuge reviewing https://github.com/newrelic-experimental/newrelic-sketch-java/blob/1ce245713603d61ba3a4510f6df930a5479cd3f6/src/main/java/com/newrelic/nrsketch/indexer/ScaledExpIndexer.java#L26 -- would you say 52bits is a better limit for the index?

yzhuge: starting at https://github.com/newrelic-experimental/newrelic-sketch-java/blob/main/src/test/java/com/newrelic/nrsketch/indexer/BucketIndexerTest.java#L242, I have scale accuracy tests 
 testLogIndexerScales, testSubBucketLogIndexerScales, testSubBucketLookupIndexerScales, and testExponentIndexerScales. Summary here:

64 bit index is needed to cover double min (including subnormals) to max at scale -11 to 52. At scale -11, there are just two buckets, 0 for >= 1 values, and -1 for < 1 values. At scale 52, index "mantissa" bits == value mantissa bits. Higher scale is pointless (they will just create many empty buckets). 

Log indexer can go up to scale 42, with index off by at most 1, which is expected due to inherent floating point inaccuracy. Beyond that, it will be off by 2 or more.

Some indexers, like SubBucketLogIndexer can go to scale 52 with off by 1. SubBucket Lookup indexer of course is limited by memory size. The subbucket indexers get better accuracy because they only use log on the mantissa, which has limited range of 1 to 2.

specification/metrics/datamodel.md
jmacd: I wrote 
```
Regardless of scale, producers
SHOULD ensure that the index of any encoded bucket falls within the
range of a signed 32-bit integer.  This recommendation is applied to
limit the width of integers used in standard processing pipelines such
as the OpenTelemetry collector.  The wire-level protocol could be
extended for 64-bit bucket indices in a future release.
```

specification/metrics/datamodel.md
jmacd: I removed it. I believe the mapping functions are well-defined for subnormals, I just don't want to think about them 😁

jmacd (commented):
Thank you @yzhuge!

specification/metrics/datamodel.md
```diff
@@ -398,14 +404,153 @@ denotes Delta temporality where accumulated event counts are reset to zero after
...
+### ExponentialHistogram
+
+**Status**: [Experimental](../document-status.md)
+
+[ExponentialHistogram](https://github.com/open-telemetry/opentelemetry-proto/blob/cfbf9357c03bf4ac150a3ab3bcbe4cc4ed087362/opentelemetry/proto/metrics/v1/metrics.proto#L222)
+data points are an alternate representation to the
+[Histogram](#histogram) data point, used to convey a population of
+recorded measurements in a compressed format.  ExponentialHistogram
+compresses bucket boundaries using an exponential formula, making it
+suitable for conveying high-resolution data using a relatively large
```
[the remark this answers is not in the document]
yzhuge: sounds good

yzhuge:
@jmacd all looks good now.

yzhuge (approved):

specification/metrics/datamodel.md
```diff
@@ -413,14 +419,158 @@ Changing the inclusivity and exclusivity of bounds is an example of
...
+
+Producers MAY use a built-in logarithm function to calculate the
+bucket index of a value.  The use of a built-in logarithm function
+could lead to results that differ from the bucket index that would be
+computed using arbitrary precision or a lookup table, however
+producers are not required to perform an exact computation.  As a
+result, ExponentialHistogram exemplars could map into buckets with
+zero count.  We expect to find such values counted in the adjacent
+bucket.
+
```
oertl, line 556: I would prefer "in adjacent buckets" over "in the adjacent bucket" as values may also be mapped to non-direct neighbors.

specification/metrics/datamodel.md
```diff
@@ -413,14 +419,158 @@ Changing the inclusivity and exclusivity of bounds is an example of
...
+which allows consumers to lower the resolution of a histogram (i.e.,
+downscale) without introducing error.
+
+#### Exponential buckets
+
+The ExponentialHistogram bucket identified by `index`, a signed
+integer, represents values in the population that are greater than or
+equal to `base**index` and less than `base**(index+1)`.  Note that the
+ExponentialHistogram specifies a lower-inclusive bound while the
+explicit-boundary Histogram specifies an upper-inclusive bound.
```
jsuereth, line 487: Glad you call this one out!

jsuereth (approved):

specification/metrics/datamodel.md
```diff
@@ -413,14 +419,239 @@ Changing the inclusivity and exclusivity of bounds is an example of
...
+}
+```
+
+2. For negative scales, the index of a value equals the normalized
+   base-2 exponent (as by `GetExponent()` above) shifted to the right
+   by `-scale`.  Note that because of sign extension, this shift performs
+   correct rounding for the negative indices.  This may be written as:
+   
+```golang
+  return GetExponent(value) << -scale
```
yzhuge, line 609: you mean ">>"?

specification/metrics/datamodel.md
```diff
@@ -413,14 +419,239 @@ Changing the inclusivity and exclusivity of bounds is an example of
...
+```
+   
+3. For positive scales, use of the built-in natural logarithm
+   function.  A multiplicative factor equal to `2**scale / ln(2)`
+   proves useful (where `ln()` is the natural logarithm), for example:
+   
+```golang
+    scaleFactor := math.Log2E * math.Exp2(1<<scale)
```
yzhuge, line 617: should this be "math.Exp2(scale)"?
And log method is generally applicable on 0 and negative scales too, though less efficient. For simplicity, an implementation may just do a one size fits all log method, especially when it expects that 0 and negative scales are used rarely.

specification/metrics/datamodel.md
jmacd: Done.

specification/metrics/datamodel.md
jmacd: Yes, thanks.

specification/metrics/datamodel.md
jmacd: Fixed. Thanks.

jmacd:
@yzhuge while testing the mapping function for `scale=-4` I noticed that index -68 is possibly problematic and wanted to see what you think about it.

Using the algorithm as given in the current text, with value 0x1p-1073 or 0x1p-1074, the algorithm computes index -68. However, the lower-boundary of index -68 cannot be represented using a double. Should the producer bother to check that an index can be reverse-mapped before using it?

specification/metrics/datamodel.md
jmacd: Thank you!

yzhuge:
@jmacd "the lower-boundary of index -68 cannot be represented using a double". 
In negative scales, the lowest bucket may be a "partial bucket" (partially representable by "double" range). 
An implementation's lower bound function will need to guard against floating point underflow. In NrSketch, I have https://github.com/newrelic-experimental/newrelic-sketch-java/blob/main/src/main/java/com/newrelic/nrsketch/indexer/ExponentIndexer.java#L43

Similarly, to be bullet proof, we need guard against overflow at the high end of "double":
https://github.com/newrelic-experimental/newrelic-sketch-java/blob/main/src/main/java/com/newrelic/nrsketch/indexer/ScaledExpIndexer.java#L114

Not sure it is worth mentioning such extreme cases in the spec though. Maybe it is, spec has to be bullet proof.

"Should the producer bother to check that an index can be reverse-mapped before using it?"
I think producer should still map it to the lowest (partial) bucket. Its other choices are: round off to zero; throw an exception. None is great.

And I assume you are talking about the exponent extraction method. The log method generally does not work well in the subnormal range. I had to skip it in subnormal tests (https://github.com/newrelic-experimental/newrelic-sketch-java/blob/main/src/test/java/com/newrelic/nrsketch/indexer/BucketIndexerTest.java#L400)

specification/metrics/datamodel.md
```diff
@@ -651,6 +651,12 @@ lookup table with `O(2**scale)` entries.
 
 ##### Producer recommendations
 
+Regardless of scale or mapping technique, it can be difficult to
```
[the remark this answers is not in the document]
jmacd: Done -- approximately. I put your final sentence into the Consumer section and reworded slightly. IIUC, this can be stated as "Consumers MAY round partially representable bucket indices up or down to the adjacent fully-representable bucket index."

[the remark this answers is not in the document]
yzhuge: I meant consumer keeping the partial bucket, but capping the bound to a representable number, not folding the bucket to another bucket. For example, if last bucket is logically 90 to 100, but the format can only represent up to 95, then the consumer shrinks the bucket to 90 to 95. The returned bound is typically used for charting or percentile calculation. With the capping, the result is still logically correct. No "rounding" error is introduced. Rounding to a full bucket is an alternative, but it introduces rounding error.

specification/metrics/datamodel.md
```diff
@@ -422,14 +435,262 @@ Changing the inclusivity and exclusivity of bounds is an example of
...
+The following choices of mapping function have been validated through
+reference implementations.
+
+##### Scale zero: extract the exponent
+
+For scale zero, the index of a value equals its normalized base-2
+exponent, meaning the value of _exponent_ in the base-2 fractional
+representation `1._significand_ * 2**_exponent_`.  Normal IEEE 754
+double-width floating point values have indices in the range
+`[-1022,+1023]` and subnormal values have indices in the range
```
reyang, line 601:
```suggestion
`[-1022, +1023]` and subnormal values have indices in the range
```

specification/metrics/datamodel.md
```diff
@@ -422,14 +435,262 @@ Changing the inclusivity and exclusivity of bounds is an example of
...
+`[-1074,-1023]`.  This may be written as:
```
reyang, line 602:
```suggestion
`[-1074, -1023]`.  This may be written as:
```

specification/metrics/datamodel.md
```diff
@@ -19,6 +19,17 @@
   * [Sums](#sums)
   * [Gauge](#gauge)
   * [Histogram](#histogram)
+  * [ExponentialHistogram](#exponentialhistogram)
+    + [Exponential scale](#exponential-scale)
```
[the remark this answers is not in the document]
jmacd: Done!

reyang (approved):

specification/metrics/datamodel.md
```diff
@@ -422,14 +435,262 @@ Changing the inclusivity and exclusivity of bounds is an example of
...
+differ from the bucket index that would be computed using arbitrary
+precision or a lookup table, however producers are not required to
+perform an exact computation.  As a result, ExponentialHistogram
+exemplars could map into buckets with zero count.  We expect to find
+such values counted in the adjacent buckets.
+
+#### Consumer Expectations
+
+ExponentialHistogram bucket indices are expected to map into buckets
+where both the uppwer and lower boundaries that can be represented
```
aabmass, line 676:
```suggestion
where both the upper and lower boundaries can be represented
```

aabmass (approved):

PR #1926, opened 2021-09-15
Release v1.7.0

carlosalberto:
September release, following our monthly cadence.

@reyang Are we ready to go, metrics-wise?

@bogdandrutu @dyladan Should we include #1923 in this release?

cc @tigrannajaryan

PR #186, opened 2019-07-18
Add exporter API and more details about SDK implementation

tigrannajaryan:
This clarifies what functionality should be the core part of
the SDK and what should be available for optional usage and
also defines the API that exporters need to define.

Co-authored with @bogdandrutu

PR #1938, opened 2021-09-17
Remove MeasurementProcessor from the Metrics SDK spec

reyang:
This was discussed during the [08/12/2021 #186 Metrics SIG at 11am PT](https://docs.google.com/document/d/1pdvPeKjA8v8w_fGKAN68JjWBmVJtPCpqdi9IZrd6eEo/edit#heading=h.rf686wxmv2ru) and we concluded that `MeasurementProcessor` should be removed.

I didn't make the change at that time because we were focusing on getting the initial Experimental version done (which is achieved now).

**Note:** this CAN be part of the [spec v1.7.0 release](https://github.com/open-telemetry/opentelemetry-specification/pull/1926#issuecomment-921932216), although it does NOT HAVE TO be.

### DIFF
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -402,13 +402,18 @@ series and it requires further analysis.
 
 ## Exemplar
 
-An [Exemplar](./datamodel.md#exemplars) is a recorded measurement that exposes
-the following pieces of information:
+An [Exemplar](./datamodel.md#exemplars) is a recorded
+[Measurement](./api.md#measurement) that exposes the following pieces of
+information:
 
 - The `value` that was recorded.
-- The `time` the measurement was seen.
-- The set of [Attributes](../common/common.md#attributes) associated with the measurement not already included in a metric data point.
-- The associated [trace id and span id](../trace/api.md#retrieving-the-traceid-and-spanid) of the active [Span within Context](../trace/api.md#determining-the-parent-span-from-a-context) of the measurement.
+- The `time` the `Measurement` was seen.
+- The set of [Attributes](../common/common.md#attributes) associated with the
+  `Measurement` not already included in a metric data point.
+- The associated [trace id and span
+  id](../trace/api.md#retrieving-the-traceid-and-spanid) of the active [Span
+  within Context](../trace/api.md#determining-the-parent-span-from-a-context) of
+  the `Measurement`.
 
 A Metric SDK MUST provide a mechanism to sample `Exemplar`s from measurements.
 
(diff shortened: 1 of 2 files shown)


cijothomas (approved):

specification/metrics/sdk.md
```diff
@@ -401,47 +400,6 @@ Attributes which belong to Metrics are exempt from the
...
-* The `Measurement`
-* The `Instrument`, which is used to report the `Measurement`
-* The `Resource`, which is associated with the `MeterProvider`
-
-In addition to things listed above, if the `Measurement` is reported by a
-synchronous `Instrument` (e.g. [Counter](./api.md#counter)),
-`MeasurementProcessor` MUST have access to:
-
-* [Baggage](../baggage/api.md)
-* [Context](../context/context.md)
```
jsuereth, line 421: Do you think the Exemplar specification says enough about access to Baggage/Context or should we improve the verbage in that section now that this is disappearing?

reyang: I think it's okay. I've tweaked the Exemplar wording a bit (although we probably can do it in another PR), please take a look.

reyang: @jsuereth do you think ExemplarFilter and ExemplarReservoir should have access to Resource? I think no but I can also imagine some folks might want to have a general ExemplarFilter that handles different Resources differently.

specification/metrics/sdk.md
```diff
@@ -469,14 +432,16 @@ A Metric SDK SHOULD provide extensible hooks for Exemplar sampling, specifically
...
-- The value of the measurement.
-- The complete set of `Attributes` of the measurment.
-- the `Context` of the measuremnt.
-- The timestamp of the measurement.
+- The `value` of the measurement.
+- The complete set of `Attributes` of the measurement.
+- The [Context](../context/context.md) of the measurement, which covers the
+  [Baggage](../baggage/api.md) and the current actve
+  [Span](../trace/api.md#span).
+- The `timestamp` of the measurement.
```
reyang, line 444: @jsuereth I guess `the timestamp of the measurement` might generate confusion since [Measurement](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#measurement) by definition does not contain a timestamp (for performance reason).

@jmacd what do you feel?

specification/metrics/sdk.md
jsuereth: I don't think they need it.  Resource is per-SDK (or per-MeterProvider), as is exemplar configuration.  Implicitly they have access to it because they're per-SDK and you configure those at the same time.

InsturmentationLibrary is a different story.   We may want to discuss that in the next metrics SiG

specification/metrics/sdk.md
jsuereth: Right, for the Java prototype we actually pull timestamp for Exemplars *after* the sampling decision was made, but in-line with measurement collection.

We can reword this to be "A timestamp that best represents when the measurement was taken"

jsuereth (approved):

bogdandrutu (approved):

specification/metrics/sdk.md
reyang: Thanks @jsuereth, I've updated the PR applying your suggestion.

carlosalberto:
How are we looking? Ready to merge? I see three reviews, so confirm this can be merged (prior to 1.7) @open-telemetry/specs-metrics-approvers

jsuereth:
@carlosalberto This can be merged, was giving it the business-day-wait-for-review required.

reyang:
Another attempt here #4318.

Issue #955, opened 2020-09-15
Propagating Exceptions in Span Events


HaloFour:
I'm working with the Java implementation of the reference OpenTelemetry Tracer and using a span exporter that renders spans to logs via log4j2.  Currently the Span interface exposes a helper method `recordException` which can record the details of an exception as an event with attributes to the span.  However, as the specification supports a small subset of possible data types for attributes those details and added as String attributes with a raw stack trace and the original exception instance is lost.  This prevents me from being able to use the log4j2 configuration to specify how I want the exception and stack trace to be rendered into the log.  What I am interested in is some way to propagate an exception into an event on the span so that a span exporter can obtain that original exception instance to interpret/render.

I don't wish a prescribe a solution but I do believe that exceptions are a fairly common concern across most programming languages so it makes sense having a standard way to describe them in the specification.

Here are some possible solutions if this is considered:

1. Allow for an Event to have an optional Exception/Error property
1. Allow for an Attribute to be of type Exception
1. Allow for custom Event implementations to be supported
    1. In the Java SDK `Event` is an interface and I can provide my own implementation however the reference SDK Tracer implementation wraps my `Event` and prevents access to any custom members I expost

This conversation might fall into the larger conversation as to what constitutes a successful Span, although being able to track exceptions as discrete events within the lifetime of a span doesn't impact how those exceptions might be interpreted into a single status/success.

jamescrosswell:
We've bumped into [similar issues in .NET](https://github.com/open-telemetry/opentelemetry-dotnet/issues/2439#issuecomment-1577314568). 

Without access to the original Exception (not just the `Type` or `StackTrace` as a `string`), it's impossible to match the exception up with debug info to provide accurate symbolication and error diagnostics. So we can tell people there was a problem, but we can't help them fix it properly 😢.

I don't know that the exception needs to be propagated across process boundaries (and that might not be desirable for privacy reasons). However when receiving telemetry in the same process where an exception occurs, it would be great if there there was a `Activity.GetExceptions()` method or possible an `Event.GetException()` method that could be used to get the original exception.

PR #697, opened 2020-07-09
Exception reporting specification

mwear:
This PR is another take on exception / error reporting. It's based on #427 and conversations from the errors working group. Below is a summary of differences between this proposal and #427:

* Use exception terminology rather than error. Whether or not an exception is an error will depend on context.
* Change stacktrace from array of strings, to a single string
  * The format of the string is to be determined on a language by language basis
  * Structure can be derived from the string format and platform information, if necessary
* Add a `Span.RecordException` API method

The goal of this PR is to solve the 90% use case, without painting ourselves into a corner for future improvements.

PR #427, opened 2020-01-27
Add semantic conventions for errors

mwear:
Having a way to record errors and annotate spans with that information is essential functionality that is currently missing from OpenTelemetry. 

There is currently [OTEP-69](https://github.com/open-telemetry/oteps/pull/69) that proposes typed events and includes structures for recording errors.

Since it was authored, [array-value attributes](https://github.com/open-telemetry/opentelemetry-specification/pull/368) have been added to the specification. This addition makes it possible to record errors using the existing event type by representing a stack trace as an array of strings, where each entry corresponds to a stackframe.

This PR doesn't intend to compete with OTEP-69, but offers another approach to be considered. Hopefully, between this proposal and OTEP-69 we can begin to drive a discussion on how to record errors forward.

PR #1939, opened 2021-09-19
Recommend Exception.StackTrace for C#

bgrainger:
## Changes

The StackTrace property "gets a string representation of the immediate frames on the call stack." It's a better way to get the stack trace than calling `.ToString()`, which will include the exception type and message (which should already be in the `exception.type` and `exception.message` attributes. It's supported on all .NET Frameworks and .NET Standard versions, so no reason not to use it.

### DIFF
--- a/specification/trace/semantic_conventions/exceptions.md
+++ b/specification/trace/semantic_conventions/exceptions.md
@@ -81,7 +81,7 @@ to adopt them if they see fit.
 
 | Language   | Format                                                              |
 | ---------- | ------------------------------------------------------------------- |
-| C#         | the return value of [Exception.ToString()][csharp-stacktrace]       |
+| C#         | the return value of [Exception.StackTrace][csharp-stacktrace]       |
 | Go         | the return value of [runtime.Stack][go-stacktrace]                  |
 | Java       | the contents of [Throwable.printStackTrace()][java-stacktrace]      |
 | Javascript | the return value of [error.stack][js-stacktrace] as returned by V8  |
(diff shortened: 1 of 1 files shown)


Oberon00 (changes requested):
I think you are running into a common misunderstanding here: exception.stacktrace should not only contain the frames on the stack, but everything that is usually included when an "exception stacktrace" is printed in the language, which is exactly what ToString does. I fought to have the name changed to something like full_details instead of stacktrace (https://github.com/open-telemetry/opentelemetry-specification/pull/697#discussion_r453662519), but stacktrace is also a common, though ambigous name for the concept of "human-readable exception details". 

With this change, we would lose important details, like the cause of the exception (inner exception). Also, the stacktraces for AggregateExceptions are usually quite useless, while the full ToString is more helpful.

Oberon00:
> Sending raw strings vs. structured data can make this hard as you need to re-parse the data per-language.
>
> I'm wondering if maybe we should have a best of both worlds? e.g. reporting a structured stack trace and allowing "caused by" chains in the stack trace (or some kind of better error string)

I agree with that and I think it was also consensus back then that we would want a structured form of exception reporting in the future. In particular, see https://github.com/open-telemetry/opentelemetry-specification/pull/427#discussion_r371698926. But this would be an entirely different issue, and a separate (set of) attribute(s). Related #955.

bgrainger:
@Oberon00 Thanks for the link to that comment thread. I think the attribute name and description (of `exception.stacktrace`) in this document are incredibly misleading (but I don't need to tell you that). Comparing to the recommendations for other languages, it seems that `.ToString()` is the desired data so I'll close this.

PR #1941, opened 2021-09-20
Add missing fields to logs mapping from universal model to Splunk HEC

dmitryax:
Since we already have logs mapping from the unified model to Splunk HEC, we should define it for all fields in the unified model. This PR adds missing request context fields to the mapping.


tigrannajaryan (approved):

dmitryax:
After a discussion with @tigrannajaryan, we decided to go with already established field names for request context related fields like `trace_id`, `span_id` and `trace_flags`

bogdandrutu (approved):
Not in this PR, but probably worth changing this document to use md tables instead of html tables.

kumoroku (approved):

PR #1944, opened 2021-09-21
Refer and link to exception semantic conventions

### LINKED ISSUE #1868: Error semantics in Log Data Model don't match general semantics for Exceptions

Figure out what the correct/proposed OTel attribute names in a structured log should be for information about an exception.

**Additional context.**

The [Log Data Model spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md) doesn't mention exceptions at all. However, [the example mapping for Elastic Common Schema](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#elastic-common-schema) recommends a number of attributes beginning with `error.`

There is already [another spec documenting semantic conventions for Exceptions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/exceptions.md), and they are different. 
It seems like these should be used in the Elastic mapping, as well as mentioned elsewhere in the Log Data Model spec.

GrahamLea:
Fixes #1868

## Changes

Adds a section to the logs spec referring to the semantic conventions for exception-related attributes.

### DIFF
--- a/specification/logs/data-model.md
+++ b/specification/logs/data-model.md
@@ -27,6 +27,7 @@
     - [Field: `Body`](#field-body)
     - [Field: `Resource`](#field-resource)
     - [Field: `Attributes`](#field-attributes)
+      - [Errors and Exceptions](#errors-and-exceptions)
   - [Example Log Records](#example-log-records)
   - [Appendix A. Example Mappings](#appendix-a-example-mappings)
     - [RFC5424 Syslog](#rfc5424-syslog)
@@ -426,6 +427,14 @@ information about the request context (other than TraceId/SpanId). SHOULD follow
 OpenTelemetry
 [semantic conventions for Attributes](../trace/semantic_conventions/README.md).
 This field is optional.
+
+#### Errors and Exceptions
+
+Additional information about errors and/or exceptions that are associated with
+a log record MAY be included in the structured data in the `Attributes` section
+of the record.
+If included, they MUST follow the OpenTelemetry
+[semantic conventions for exception-related attributes](../trace/semantic_conventions/exceptions.md#attributes).
 
 ## Example Log Records
 


specification/logs/data-model.md
```diff
@@ -427,6 +428,14 @@ OpenTelemetry
 [semantic conventions for Attributes](../trace/semantic_conventions/README.md).
 This field is optional.
 
+#### Errors and Exceptions
+
+Additional information about errors and/or exceptions that are associated with
+a log reccord MAY be included in the structured data in the `Attributes` section 
+of the record.
+If inclucded, they SHOULD follow the OpenTelemetry
+[semantic conventions for exceptions](../trace/semantic_conventions/exceptions.md).
```
tigrannajaryan, line 437: Let's link directly to the `Attributes` section for more clarity: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/exceptions.md#attributes

specification/logs/data-model.md
```diff
@@ -427,6 +428,14 @@ OpenTelemetry
 [semantic conventions for Attributes](../trace/semantic_conventions/README.md).
 This field is optional.
 
+#### Errors and Exceptions
+
+Additional information about errors and/or exceptions that are associated with
+a log reccord MAY be included in the structured data in the `Attributes` section 
+of the record.
+If inclucded, they SHOULD follow the OpenTelemetry
```
tigrannajaryan, line 436:
```suggestion
If included, they SHOULD follow the OpenTelemetry
```

tigrannajaryan (commented):
@GrahamLea thanks for the contribution. Please sign the CLA.

specification/logs/data-model.md
GrahamLea: Good one

GrahamLea:
>Please sign the CLA.

It wasn't working yesterday, but is done now.

tigrannajaryan (approved):

specification/logs/data-model.md
```diff
@@ -427,6 +428,14 @@ OpenTelemetry
 [semantic conventions for Attributes](../trace/semantic_conventions/README.md).
 This field is optional.
 
+#### Errors and Exceptions
+
+Additional information about errors and/or exceptions that are associated with
+a log reccord MAY be included in the structured data in the `Attributes` section
```
codeboten, line 434:
```suggestion
a log record MAY be included in the structured data in the `Attributes` section
```

codeboten (approved):

carlosalberto (approved):

specification/logs/data-model.md
```diff
@@ -427,6 +428,14 @@ OpenTelemetry
 [semantic conventions for Attributes](../trace/semantic_conventions/README.md).
 This field is optional.
 
+#### Errors and Exceptions
+
+Additional information about errors and/or exceptions that are associated with
+a log record MAY be included in the structured data in the `Attributes` section
+of the record.
+If included, they SHOULD follow the OpenTelemetry
```
arminru, line 436:
```suggestion
If included, they MUST follow the OpenTelemetry
```
Is there any reason to make these a SHOULD given it's a MUST for traces?
If it's a SHOULD, consumers cannot rely on it being in the format they'd expect from the semantic convention. If a producer cannot adhere to the specified semantics, they could use different attribute keys just like a trace producer would have to.

GrahamLea: I used SHOULD here only because the containing section about `Attributes` more generally uses SHOULD.
So, I have no argument why it shouldn't be MUST.
I'll leave it to other people with more experience in the spec work to debate whether it should or shouldn't change.

arminru: @open-telemetry/specs-logs-approvers @open-telemetry/specs-approvers Any opinions?

tigrannajaryan: I do not see a reason why this conventions should not be followed, so I think `MUST` works. We may change our mind if we discover cases when the conventions cannot be followed. If that happens this spec can be updated, perhaps with additional explanation and the we can change MUST to SHOULD. The logs spec is still experimental so we are free to make changes as needed.

carlosalberto: Ping @GrahamLea

tigrannajaryan:
@arminru PTAL.

arminru (approved):

GrahamLea:
Thanks for giving this more attention than it probably deserved, folks.

tigrannajaryan:
@GrahamLea thanks for the contribution.

tigrannajaryan:
@jkwatson FYI, the exceptions convention added to logs.

PR #1955, opened 2021-09-22
Clarify OTLP server components MUST support none/gzip compression.

carlosalberto:
As mentioned in #1923 we need to be explicit on the compression support for the transports on the client AND server side (i.e.g collector receivers at this time).

cc @tigrannajaryan @bogdandrutu

### DIFF
--- a/specification/protocol/otlp.md
+++ b/specification/protocol/otlp.md
@@ -70,6 +70,12 @@ OTLP is a request/response style protocols: the clients send requests, the
 server replies with corresponding responses. This document defines one requests
 and response type: `Export`.
 
+All transports MUST support in the client and server components the following
+compression options:
+
+* No compression, denoted by `none`.
+* Gzip compression, denoted by `gzip`.
+
 ### OTLP/gRPC
 
 **Status**: [Stable](../document-status.md)
--- a/specification/protocol/otlp.md
+++ b/specification/protocol/otlp.md
@@ -70,10 +70,9 @@ OTLP is a request/response style protocols: the clients send requests, the
 server replies with corresponding responses. This document defines one requests
 and response type: `Export`.
 
-All transports MUST support in the client and server components the following
-compression options:
+All transports MUST support in the server components the following compression options:
 
-* No compression, denoted by `none`.
+* No compression.
 * Gzip compression, denoted by `gzip`.
 
 ### OTLP/gRPC
--- a/specification/protocol/otlp.md
+++ b/specification/protocol/otlp.md
@@ -70,7 +70,7 @@ OTLP is a request/response style protocols: the clients send requests, the
 server replies with corresponding responses. This document defines one requests
 and response type: `Export`.
 
-All transports MUST support in the server components the following compression options:
+All server components MUST support the following transport compression options:
 
 * No compression, denotated by `none`.
 * Gzip compression, denoted by `gzip`.
--- a/specification/protocol/otlp.md
+++ b/specification/protocol/otlp.md
@@ -72,7 +72,7 @@ and response type: `Export`.
 
 All transports MUST support in the server components the following compression options:
 
-* No compression.
+* No compression, denotated by `none`.
 * Gzip compression, denoted by `gzip`.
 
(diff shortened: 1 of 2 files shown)


specification/protocol/otlp.md
```diff
@@ -70,6 +70,12 @@ OTLP is a request/response style protocols: the clients send requests, the
 server replies with corresponding responses. This document defines one requests
 and response type: `Export`.
 
+All transports MUST support in the client and server components the following
+compression options:
+
+* No compression, denoted by `none`.
```
tigrannajaryan, line 76: In the Collector config, the "No compression" is chosen when the "Compression" setting is left empty. The `OTEL_EXPORTER_OTLP_*COMPRESSION` env variable for SDK also has the same behavior.

To avoid confusion perhaps delete the "denoted by `none`" part here?

specification/protocol/otlp.md
```diff
@@ -70,6 +70,12 @@ OTLP is a request/response style protocols: the clients send requests, the
 server replies with corresponding responses. This document defines one requests
 and response type: `Export`.
 
+All transports MUST support in the client and server components the following
```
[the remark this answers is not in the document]
carlosalberto: Good call!

specification/protocol/otlp.md
carlosalberto: Updated.

johnnyshields: I actually think that we should have an explicit value for "no compression".

carlosalberto: @johnnyshields I think the intent _would be_ to have the empty literal for that, i.e. `*_COMPRESSION=""` - is that right @tigrannajaryan ?

tigrannajaryan: Yes, if the implementations are done correctly it should be possible to explicitly force the "no compression" by setting OTEL_EXPORTER_OTLP_*COMPRESSION="". This would indicate to the implementation that "no compression" is requested. If OTEL_EXPORTER_OTLP_*COMPRESSION is undefined it would indicate to the implementation that the default is requested (whatever is the default for that implementation).

I would not want to change this behavior now. AFAIK all existing implementations treat empty string as "no compression".

johnnyshields: For my tastes, using blank/empty string is bad practice for an enumeration, but it's your rodeo.

yurishkuro: > using blank/empty string is bad practice for an enumeration

+1

yurishkuro: it is especially bad when it is drastically different in meaning from unset value

tigrannajaryan: @johnnyshields @yurishkuro  I don't disagree. 
My concern is that it may be a breaking change, but perhaps my concern is pointless if the SDK don't really implement more than one compression method at the moment (either gzip or none) which I suspect may be the case.
I would like maintainers to chime in and tell what their SDKs do _currently_ and whether it is a breaking change for them if we make "none" to mean "no compression".

johnnyshields: *not specified* can default to the value "none". This will not be a breaking change in terms of existing implementations, since "none" is a newly added value. Clients should be advised to check that they handle "none" appropriately (i.e. rather than returning an error.)

carlosalberto:
@open-telemetry/specs-approvers Please review/approve. In yesterday's Maintainers meeting this item was discussed and it seems we won't be breaking anything out there.

cc @open-telemetry/proto-go-approvers @open-telemetry/javascript-approvers

jsuereth (approved):

tigrannajaryan (approved):

specification/protocol/otlp.md
```diff
@@ -70,6 +70,11 @@ OTLP is a request/response style protocols: the clients send requests, the
 server replies with corresponding responses. This document defines one requests
 and response type: `Export`.
 
+All transports MUST support in the server components the following compression options:
```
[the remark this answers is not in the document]
carlosalberto: Applied, thanks!

PR #1965, opened 2021-09-24
Update env variable values to match other env variables.

jsuereth:
Exemplar environment variable configuration doesn't match lower-case convention in other SDK environment variable parameters.

Hat tip to @anuraaga for calling this out.

_Note: This change is still ok because Metrics SDK specification is not stable, and this is not a new feature, just a change to an existing one.  I believe there is no (released) implementation  of this feature yet._


tigrannajaryan (approved):

tigrannajaryan:
Should this env var be also moved to https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md where all the other env vars are?

jsuereth:
@tigrannajaryan I think once metrics is marked stable (not feature freeze) then yes this should move.

jkwatson (approved):

reyang (approved):

arminru (approved):

carlosalberto (approved):

jmacd (approved):

Issue #1891, opened 2021-08-26
Metrics SDK: define standard circuit breakers for uncollected, infinite cardinality recordings


jkwatson:
It is possible (perhaps even likely) that instrumentation will generate metrics with cardinality explosions. There may be backends that can handle this, so it's not necessarily an error in the instrumentation. However, it will often be a mistake in the instrumentation, and the SDK needs to include some sort of mechanism so that the the application will not die with a memory leak in this case. This could especially happen with a pull-based exporter if nothing is pulling the data from it, for example.

The SDK specification should define the standard strategy for this circuit breaker. There are at least several possible options to solving this, and it would be good if all languages implemented the same default strategy.

Some options:
a) Count the cardinality and start dropping recordings for instruments that exceed it.
b) Watch for specific dimensions that are causing the explosion and start automatically dropping those dimensions.
c) Something more clever.

reyang:
I'm trying to address these topics in #1966, leave it as guidance/suggestion rather than spec requirements.

jack-berg:
An initial implementation of cardinality defenses was [recently implemented](https://github.com/open-telemetry/opentelemetry-java/pull/3831) in java.

Im summary, each view has a max number of metric streams. Currently, this max is hardcoded to be 2000, which was chosen rather arbitrarily. If exceeded, we log a warning and drop subsequent recordings. At the same time we aggressively forget metric streams - any metric stream that doesn't have recordings during a collection is forgotten. This applies to cumulative / delta and sync / async cases.

jmacd:
@jack-berg -- 

@MrAlias asked for guidance on this topic in https://github.com/open-telemetry/opentelemetry-go/issues/3006#issuecomment-1189376884

I am aware of incompleteness in our specification around the appearance or disappearance of metrics from one collection to the next, depending on async/sync and depending on temporality. I came up with my own answers to this set of questions for my (i.e., Lightstep's) experimental SDK, and [one of the deciding factors came from trying to pin down the behavior of synchronous gauge instruments](https://github.com/lightstep/otel-launcher-go/tree/main/lightstep/sdk/metric#synchronous-gauge-instrument) -- while prototyping for #2318.

At the highest level, my approach to these questions is "when cumulative temporality, do what a Prometheus user expects" and "when delta temporality, do what a Statsd user expects". 

[...]

Issue #1873, opened 2021-08-17
Define Metric SDK behavior when presented with invalid metric recording values


jkwatson:
1) We know that the metrics API cannot throw exceptions/panic on bad inputs.
2) The Counter (for example) cannot accept negative values to its `add` operation. Histograms should only accept positive values for recordings.
3) How should SDK implementations behave when presented with syntactically valid but semantically invalid inputs such as this for metric recordings?

PR #1966, opened 2021-09-24
Add metrics spec supplementary guidelines

reyang:
Related to #1873 and #1891.

If you find it hard to look at the raw markdown file, here is a rendered version https://github.com/reyang/opentelemetry-specification/blob/reyang/guideline/specification/metrics/supplementary-guidelines.md.

### DIFF
--- a/specification/metrics/supplementary-guidelines.md
+++ b/specification/metrics/supplementary-guidelines.md
@@ -227,7 +227,7 @@ keeping things in memory unless there is a must need. One good example is the
 
 **Design a better memory layout**, so the storage is efficient and accessing the
 storage can be fast. This is normally specific to the targeting programming
-language and platform. For example, aliging the memory to the CPU cache line,
+language and platform. For example, aligning the memory to the CPU cache line,
 keeping the hot memories close to each other, keeping the memory close to the
 hardware (e.g. non-paged pool,
 [NUMA](https://en.wikipedia.org/wiki/Non-uniform_memory_access)).
@@ -235,7 +235,7 @@ hardware (e.g. non-paged pool,
 **Pre-allocate and pool the memory**, so the SDK doesn't have to allocate memory
 on-the-fly. This is especially useful to language runtimes that have garbage
 collectors, as it ensures the hot path in the code won't trigger garbage
-colletion.
+collection.
 
 **Limit the memory usage, and handle critical memory condition.** The general
 expectation is that a telemetry SDK should not fail the application. This can be
--- a/specification/metrics/supplementary-guidelines.md
+++ b/specification/metrics/supplementary-guidelines.md
@@ -234,7 +234,7 @@ important things for the OpenTelemetry SDK:
   on-the-fly. This is especially useful to language runtimes that have garbage
   collectors, as it ensures the hot path in the code won't trigger garbage
   colletion.
-* Limit the memmory usage, and handle critical memory condition. In general the
+* Limit the memory usage, and handle critical memory condition. The general
   expectation is that a telemetry SDK should not fail the application. This can
   be done via some dimension-capping algorithm - e.g. start to combine/drop some
   data points when the SDK hits the memory limit, and provide a mechanism to
--- a/specification/metrics/supplementary-guidelines.md
+++ b/specification/metrics/supplementary-guidelines.md
@@ -23,23 +23,23 @@ The OpenTelemetry Metrics [Data Model](./datamodel.md) and [SDK](./sdk.md) are
 designed to support both Cumulative and Delta
 [Temporality](./datamodel.md#temporality). It is important to understand that
 temporality will impact how the SDK could manage memory usage. Let's take the
-following example:
+following HTTP requests example:
 
 * During the time range (T<sub>0</sub>, T<sub>1</sub>]:
-  * HTTP verb = `GET`, status = `200`, duration = `50 (ms)`
-  * HTTP verb = `GET`, status = `200`, duration = `100 (ms)`
-  * HTTP verb = `GET`, status = `500`, duration = `1 (ms)`
+  * verb = `GET`, status = `200`, duration = `50 (ms)`
+  * verb = `GET`, status = `200`, duration = `100 (ms)`
+  * verb = `GET`, status = `500`, duration = `1 (ms)`
 * During the time range (T<sub>1</sub>, T<sub>2</sub>]:
   * no HTTP request has been received
 * During the time range (T<sub>2</sub>, T<sub>3</sub>]
-  * HTTP verb = `GET`, status = `500`, duration = `5 (ms)`
-  * HTTP verb = `GET`, status = `500`, duration = `2 (ms)`
+  * verb = `GET`, status = `500`, duration = `5 (ms)`
+  * verb = `GET`, status = `500`, duration = `2 (ms)`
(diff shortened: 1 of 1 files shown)


specification/metrics/supplementary-guidelines.md
```diff
@@ -0,0 +1,182 @@
...
+  * ProcessId = `1001`, ThreadId = `2`, PageFaults = `42`
+* During the time range (T<sub>3</sub>, T<sub>4</sub>]:
+  * ProcessId = `1001`, ThreadId = `1`, PageFaults = `60`
+  * ProcessId = `1001`, ThreadId = `2`, PageFaults = `47`
+* During the time range (T<sub>4</sub>, T<sub>5</sub>]:
+  * thread 1 died, thread 3 started
+  * ProcessId = `1001`, ThreadId = `2`, PageFaults = `53`
+  * ProcessId = `1001`, ThreadId = `3`, PageFaults = `5`
+
+If we export the metrics using **Cumulative Temporality**, it is quite
```
utpilla, line 152: Should this be Delta Temporality and the next paragraph be Cumulative?

specification/metrics/supplementary-guidelines.md
```diff
@@ -0,0 +1,182 @@
...
+impossible** to reliably deduce the Delta temporality from Cumulative
+temporality. For example:
+
+* If the maximum value is 10 during (T<sub>0</sub>, T<sub>2</sub>] and the
+  maximum value is 20 during (T<sub>0</sub>, T<sub>3</sub>], we know that the
+  maximum value during (T<sub>2</sub>, T<sub>3</sub>] must be 20.
+* If the maximum value is 20 during (T<sub>0</sub>, T<sub>2</sub>] and the
+  maximum value is also 20 during (T<sub>0</sub>, T<sub>3</sub>], we wouldn't
+  know what the maximum value is during (T<sub>2</sub>, T<sub>3</sub>], unless
+  we know that there is no value (count = 0).
```
utpilla, line 174: Maybe reword this to be more specific?

"unless we know that there is no value during (T<sub>2</sub>, T<sub>3</sub>] (count = 0) in which case the maximum value during (T<sub>2</sub>, T<sub>3</sub>] will be 0."

specification/metrics/supplementary-guidelines.md
```diff
@@ -0,0 +1,182 @@
...
+    `5 (ms)`
+* (T<sub>0</sub>, T<sub>5</sub>]
+  * dimensions: {verb = `GET`, status = `200`}, count: `6`, min: `30 (ms)`, max:
+    `100 (ms)`
+  * dimensions: {verb = `GET`, status = `500`}, count: `3`, min: `1 (ms)`, max:
+    `5 (ms)`
+
+If we choose **Cumulative Temporality**, the SDK **has to track what has
+happened prior to the latest collection/export cycle**, in this worst case, the
+SDK **will remember what has happened since the ever beginning of the process**.
```
utpilla, line 101: nit: SDK **will have to remember what has happened ever since the beginning of the process**.

specification/metrics/supplementary-guidelines.md
reyang: Nope, this is Cumulative (we're doing Cumulative->Cumulative here).

The next paragraph is about Cumulative->Delta conversion.

specification/metrics/supplementary-guidelines.md
reyang: Probably no? I think if count=0, the maximum value technically should not be exist, or it can be NaN.

specification/metrics/supplementary-guidelines.md
reyang: Fixed, thanks!

specification/metrics/supplementary-guidelines.md
utpilla: If the maximum value shouldn't technically exist, then maybe we should get rid of the "unless we know ..." part.

jsuereth (commented):
This generally looks very good.  I need to take a second pass on grammar/phrasing (That's my own worst enemy), but the content looks like a great addition/treatment of the issue for implementers.

One thing I'd like to call out though is a suggestion around Memory Management and Allocations.  While it's impossible to remove all allocations on the hot path of synchronous metric collection, particularly if using Baggage, SDK authors should be encouraged to reduce memory contention *in the hot path*, and pay most overhead at startup.   This aligns with OTel's vision of "least disruption to runtime", which shows up in recommendations around error-reporting during operation and likely should also apply to things like memory usage.

I like the idea of limiting memory as suggested here, just want to also reinforce this "allocate ahead of time and reuse" notion, which we actually expose a bit more directly in Java (and I need to annotate).


Specifically, one scenario to call out (more important for Logs/Traces than metrics) is how a system behaves when memory can no longer be allocated.   Do we still get observability or is it gone?

reyang:
> I like the idea of limiting memory as suggested here, just want to also reinforce this "allocate ahead of time and reuse" notion, which we actually expose a bit more directly in Java (and I need to annotate).

I've updated the PR to cover the pre-allocation for garbage collected languages. PTAL.
 
> Specifically, one scenario to call out (more important for Logs/Traces than metrics) is how a system behaves when memory can no longer be allocated. Do we still get observability or is it gone?

I've updated the PR trying to scratch the surface. If we need to dig deeper, a separate PR might be better?

jsuereth (approved):

specification/metrics/supplementary-guidelines.md
```diff
@@ -0,0 +1,251 @@
...
+* (T<sub>0</sub>, T<sub>5</sub>]
+  * dimensions: {verb = `GET`, status = `200`}, count: `6`, min: `30 (ms)`, max:
+    `100 (ms)`
+  * dimensions: {verb = `GET`, status = `500`}, count: `3`, min: `1 (ms)`, max:
+    `5 (ms)`
+
+You can see that we are performing Delta->Cumulative conversion, and the SDK
+**has to track what has happened prior to the latest collection/export cycle**,
+in the worst case, the SDK **will have to remember what has happened since the
+ever beginning of the process**.
```
alanwest, line 103:
```suggestion
beginning of the process**.
```

specification/metrics/supplementary-guidelines.md
```diff
@@ -0,0 +1,251 @@
...
+  * verb = `GET`, status = `200`, duration = `30 (ms)`
+  * verb = `GET`, status = `200`, duration = `50 (ms)`
+
+Let's imagine we export the metrics as [Histogram](./datamodel.md#histogram),
+and to simplify the story we will only have one histogram bucket `(-Inf, +Inf)`:
+
+If we export the metrics using **Delta Temporality**:
+
+* (T<sub>0</sub>, T<sub>1</sub>]
+  * dimensions: {verb = `GET`, status = `200`}, count: `2`, min: `50 (ms)`, max:
```
cijothomas, line 51: nit: Histogram also has Sum

jmacd (approved):
❤️ ❤️ ❤️

specification/metrics/supplementary-guidelines.md
reyang: Yeah, I think I omitted many things here intentionally so the reader could focus on the key points.

Issue #1885, opened 2021-08-23
Default protocol for OTLP exporter selection SDK env


lonewolf3739:
**What are you trying to achieve?**

Exporter selection https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#exporter-selection
 part of spec has default for jaeger and zipkin but there is no default for otlp. It is unclear whether to use `otlp/grpc` or `otlp/http` for the client libraries that have support for both.

tigrannajaryan:
OTLP/HTTP is slightly more CPU efficient than OTLP/gRCP so it may be preferable. However, it may be that most SDKs implement OTLP/gRCP already, but not OTLP/HTTP so making OTLP/HTTP may not work.

This likely needs a discussion in Maintainers meeting. Adding to the next week's agenda.

mtwo:
From the maintainers call: general agreement that gRPC should generally be the default on each artifact, unless the HTTP implementation is substantially better on that artifact. Most maintainers in the call were already following this pattern.

tigrannajaryan:
Here is some informal performance data for Go implementation ([benchmarks](https://github.com/tigrannajaryan/exp-otelproto/blob/master/runbenchmarks.sh) an OTLP client+server):
```
32000 nano batches, 1 spans per batch, 10 attrs per span
OTLP/HTTP1.1/1                 32000 spans|  6.2 cpusec|  4.7 wallsec| 5144.7 batches/cpusec|  6832.8 batches/wallsec|194.4 cpuμs/span
OTLP/GRPC-Unary/Sequential     32000 spans| 13.2 cpusec|  6.0 wallsec| 2418.7 batches/cpusec|  5370.4 batches/wallsec|413.4 cpuμs/span

8000 tiny batches, 10 spans per batch, 10 attrs per span
OTLP/HTTP1.1/1                 80000 spans|  4.4 cpusec|  3.8 wallsec| 1830.7 batches/cpusec|  2097.9 batches/wallsec| 54.6 cpuμs/span
OTLP/GRPC-Unary/Sequential     80000 spans|  6.0 cpusec|  4.1 wallsec| 1322.3 batches/cpusec|  1932.3 batches/wallsec| 75.6 cpuμs/span

4000 small batches, 100 spans per batch, 10 attrs per span
[...]

PR #1969, opened 2021-09-27
Make OTLP/HTTP the recommended default transport

tigrannajaryan:
Resolves https://github.com/open-telemetry/opentelemetry-specification/issues/1885

### DIFF
--- a/specification/protocol/exporter.md
+++ b/specification/protocol/exporter.md
@@ -17,7 +17,7 @@ The following configuration options MUST be available to configure the OTLP expo
 | Headers              | Key-value pairs to be used as headers associated with gRPC or HTTP requests. See [Specifying headers](./exporter.md#specifying-headers-via-environment-variables) for more details.                   | n/a               | `OTEL_EXPORTER_OTLP_HEADERS` `OTEL_EXPORTER_OTLP_TRACES_HEADERS` `OTEL_EXPORTER_OTLP_METRICS_HEADERS` |
 | Compression          | Compression key for supported compression types. Supported compression: `gzip`| No value [1]          | `OTEL_EXPORTER_OTLP_COMPRESSION` `OTEL_EXPORTER_OTLP_TRACES_COMPRESSION` `OTEL_EXPORTER_OTLP_METRICS_COMPRESSION` |
 | Timeout              | Maximum time the OTLP exporter will wait for each batch export. | 10s               | `OTEL_EXPORTER_OTLP_TIMEOUT` `OTEL_EXPORTER_OTLP_TRACES_TIMEOUT` `OTEL_EXPORTER_OTLP_METRICS_TIMEOUT` |
-| Protocol             | The transport protocol. Options MAY include `grpc`, `http/protobuf`, and `http/json`. See [Specify Protocol](./exporter.md#specify-protocol) for more details. | n/a               | `OTEL_EXPORTER_OTLP_PROTOCOL` `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL` |
+| Protocol             | The transport protocol. Options MAY include `grpc`, `http/protobuf`, and `http/json`. See [Specify Protocol](./exporter.md#specify-protocol) for more details. | `http/protobuf` (see below)        | `OTEL_EXPORTER_OTLP_PROTOCOL` `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL` |
 
 **[1]**: If no compression value is explicitly specified, SIGs can default to the value they deem
 most useful among the supported options. This is especially important in the presence of technical constraints,
@@ -95,9 +95,14 @@ The `OTEL_EXPORTER_OTLP_PROTOCOL`, `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL`, and `OT
 - `http/protobuf` for protobuf-encoded data over HTTP connection
 - `http/json` for JSON-encoded data over HTTP connection
 
-SDKs MUST support either `grpc` or `http/protobuf` and SHOULD support both. They also MAY support `http/json`.
+SDKs SHOULD support both `grpc` and `http/protobuf` transports and MUST support
+at least one of them. If they support only one, it SHOULD be `http/protobuf`.
+They also MAY support `http/json`.
 
-SDKs have an unspecified default, if no configuration is provided.
+If no configuration is provided the default transport SHOULD be `http/protobuf`
+unless SDKs have good reasons to choose `grpc` as the default (e.g. for backward
+compatibility reasons when `grpc` was already the default in a stable SDK
+release).
 
 ### Specifying headers via environment variables
 
--- a/specification/protocol/exporter.md
+++ b/specification/protocol/exporter.md
@@ -17,7 +17,7 @@ The following configuration options MUST be available to configure the OTLP expo
 | Headers              | Key-value pairs to be used as headers associated with gRPC or HTTP requests. See [Specifying headers](./exporter.md#specifying-headers-via-environment-variables) for more details.                   | n/a               | `OTEL_EXPORTER_OTLP_HEADERS` `OTEL_EXPORTER_OTLP_TRACES_HEADERS` `OTEL_EXPORTER_OTLP_METRICS_HEADERS` |
 | Compression          | Compression key for supported compression types. Supported compression: `gzip`| No value [1]          | `OTEL_EXPORTER_OTLP_COMPRESSION` `OTEL_EXPORTER_OTLP_TRACES_COMPRESSION` `OTEL_EXPORTER_OTLP_METRICS_COMPRESSION` |
 | Timeout              | Maximum time the OTLP exporter will wait for each batch export. | 10s               | `OTEL_EXPORTER_OTLP_TIMEOUT` `OTEL_EXPORTER_OTLP_TRACES_TIMEOUT` `OTEL_EXPORTER_OTLP_METRICS_TIMEOUT` |
-| Protocol             | The transport protocol. Options MAY include `grpc`, `http/protobuf`, and `http/json`. See [Specify Protocol](./exporter.md#specify-protocol) for more details. | `http/protobuf` (see below)        | `OTEL_EXPORTER_OTLP_PROTOCOL` `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL` |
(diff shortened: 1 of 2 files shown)


tigrannajaryan:
@open-telemetry/ruby-approvers @open-telemetry/php-approvers please review. This makes gRPC the default and required transport and the compliance matrix currently says that Ruby and PHP don't support it OTLP/gRPC so this impacts you the most.

SergeyKanzhelev (approved):

fbogsany:
gRPC is less commonly used in the Ruby ecosystem than in other languages and Ruby appears to be a lower priority than other languages for gRPC maintainers. This means that we have much less experience with gRPC in the Ruby SIG and little demand for it so far. Battle scars from spectacular Ruby + gRPC production failures also make some of us quite wary of the support burden.

We have had considerable success with OTLP/HTTP in OpenTelemetry Ruby. Is there a strong reason to force gRPC as a default transport rather than leaving the default up to individual SIGs? Similarly, is there a strong reason to **require** gRPC as a transport? Both of these are challenging for the Ruby SIG.

tigrannajaryan:
@fbogsany can you please post your comment here https://github.com/open-telemetry/opentelemetry-specification/pull/1969 and let's discuss there until we decide what's the way forward. I will keep this as draft for now.

tigrannajaryan:
@fbogsany can you please post your comment here https://github.com/open-telemetry/opentelemetry-specification/issues/1885 and let's discuss there until we decide what's the way forward. I will keep this as draft for now.

Oberon00:
> I will keep this as draft for now.

Do you want to convert it to a draft?

image

tigrannajaryan:
Discussed again in Maintainers meeting. Raw meeting notes:
```
Feedback from Bob: +1, gRPC has great implementations in C++, Go, Python, and Java, but the other language clients have usability and performance challenges. HTTP is ubiquitous.
Ted: we should still ensure that languages support both, even if they have a preference for HTTP or gRPC!
Proposal: maintainers / the TC no longer have a preference between gRPC or HTTP being the default, but OpenTelmetry components should support both.
No objections in the group
```

To summarize this is what we want to do:
1. Strongly recommend SDKs to implement both OTLP/gRPC and OTLP/HTTP+Protobuf exporters (SHOULD clause).
2. Strongly recommend OTLP/HTTP+Protobuf to be the default protocosl (SHOULD clause) but allow language SDKs to choose a different default if they have good reasons (MAY clause).

I am going to rewrite this PR according to the plan above. If there are any objections to the plan please speak up.

specification/protocol/exporter.md
```diff
@@ -95,9 +95,13 @@ The `OTEL_EXPORTER_OTLP_PROTOCOL`, `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL`, and `OT
 - `http/protobuf` for protobuf-encoded data over HTTP connection
 - `http/json` for JSON-encoded data over HTTP connection
 
-SDKs MUST support either `grpc` or `http/protobuf` and SHOULD support both. They also MAY support `http/json`.
+SDKs SHOULD support both `grpc` and `http/protobuf` transports. They also MAY
+support `http/json`.
```
arminru, line 99:
```suggestion
SDKs SHOULD support both `grpc` and `http/protobuf` transports and MUST support at least one of them.
They also MAY support `http/json`.
```

tigrannajaryan: Done.

specification/protocol/exporter.md
```diff
@@ -95,9 +95,13 @@ The `OTEL_EXPORTER_OTLP_PROTOCOL`, `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL`, and `OT
 - `http/protobuf` for protobuf-encoded data over HTTP connection
 - `http/json` for JSON-encoded data over HTTP connection
 
-SDKs MUST support either `grpc` or `http/protobuf` and SHOULD support both. They also MAY support `http/json`.
+SDKs SHOULD support both `grpc` and `http/protobuf` transports and MUST support
+at least one of them. They also MAY support `http/json`.
```
Oberon00, line 99: If only 
```suggestion
at least one of them. If they support only one, they SHOULD prefer `http/protobuf`. They also MAY support `http/json`.
```

Oberon00 (approved):

specification/protocol/exporter.md
tigrannajaryan: Done.

specification/protocol/exporter.md
```diff
@@ -17,7 +17,7 @@ The following configuration options MUST be available to configure the OTLP expo
...
 | Compression          | Compression key for supported compression types. Supported compression: `gzip`| No value [1]          | `OTEL_EXPORTER_OTLP_COMPRESSION` `OTEL_EXPORTER_OTLP_TRACES_COMPRESSION` `OTEL_EXPORTER_OTLP_METRICS_COMPRESSION` |
 | Timeout              | Maximum time the OTLP exporter will wait for each batch export. | 10s               | `OTEL_EXPORTER_OTLP_TIMEOUT` `OTEL_EXPORTER_OTLP_TRACES_TIMEOUT` `OTEL_EXPORTER_OTLP_METRICS_TIMEOUT` |
-| Protocol             | The transport protocol. Options MAY include `grpc`, `http/protobuf`, and `http/json`. See [Specify Protocol](./exporter.md#specify-protocol) for more details. | n/a               | `OTEL_EXPORTER_OTLP_PROTOCOL` `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL` |
+| Protocol             | The transport protocol. Options MAY include `grpc`, `http/protobuf`, and `http/json`. See [Specify Protocol](./exporter.md#specify-protocol) for more details. | `http/protobuf`               | `OTEL_EXPORTER_OTLP_PROTOCOL` `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL` |
```
anuraaga, line 20: I don't think this table defining a default makes sense if it's then clarified to not be a required default. I've seen some other threads about the table format being too limiting, here it seems also perhaps the table should be changed to a bullet list?

anuraaga (commented):
> gRPC has great implementations in C++, Go, Python, and Java, but the other language clients have usability and performance challenges. HTTP is ubiquitous.

I hope this is referring specifically to support for HTTP/2 in these languages. Unary gRPC is a trivial wire format and I think is basically the same to implement as HTTP given a library that supports HTTP/2. I don't know if some languages lack HTTP/2 client libraries, even in 2021, but maybe they do. I just hope people aren't getting confused into thinking gRPC is something very different, which it isn't, it's just modern HTTP like here where we don't use the gRPC library at all, it's more or less equivalent to the code implementing HTTP/protobuf.

https://github.com/open-telemetry/opentelemetry-java/blob/3640eb0129d5611ddbaea86212e27ba240ac8ef6/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/grpc/GrpcRequestBody.java

The point is the HTTP/2 requirement - while it has become far more usable there are still plenty of situations such as internal load balancing strategies that can make it difficult. HTTP/1 is a better baseline. I just don't want people to be misunderstanding that this is about gRPC, which for unary is just a very small convention on top of HTTP/2, not something that itself would cause performance / usability problems. The Google API proto embedded in the OTLP HTTP/protobuf response probably provides more usability issues ;)

specification/protocol/exporter.md
Oberon00: As a quickfix, maybe just insert "(see below)" instead of a concrete default.

Oberon00:
gRPC clients might be easy to implement (TIL!) but that is not obvious, and I don't think there is readily available documentation for that? I had the impression that gRPC is meant to be used with Google's toolchain for it (stub generator, ...) and everything else is rather unsupported.

bogdandrutu (approved):

specification/protocol/exporter.md
carlosalberto: A "see below" is definitely required IMHO. We can iterate on the table later on to fix things on a more general level.

carlosalberto (approved):

carlosalberto:
Left a small comment regarding adding a "see below" note for the default value, but otherwise LGTM.

specification/protocol/exporter.md
tigrannajaryan: Added "see below".

specification/protocol/exporter.md
```diff
@@ -17,7 +17,7 @@ The following configuration options MUST be available to configure the OTLP expo
...
 | Compression          | Compression key for supported compression types. Supported compression: `gzip`| No value [1]          | `OTEL_EXPORTER_OTLP_COMPRESSION` `OTEL_EXPORTER_OTLP_TRACES_COMPRESSION` `OTEL_EXPORTER_OTLP_METRICS_COMPRESSION` |
 | Timeout              | Maximum time the OTLP exporter will wait for each batch export. | 10s               | `OTEL_EXPORTER_OTLP_TIMEOUT` `OTEL_EXPORTER_OTLP_TRACES_TIMEOUT` `OTEL_EXPORTER_OTLP_METRICS_TIMEOUT` |
-| Protocol             | The transport protocol. Options MAY include `grpc`, `http/protobuf`, and `http/json`. See [Specify Protocol](./exporter.md#specify-protocol) for more details. | n/a               | `OTEL_EXPORTER_OTLP_PROTOCOL` `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL` |
+| Protocol             | The transport protocol. Options MAY include `grpc`, `http/protobuf`, and `http/json`. See [Specify Protocol](./exporter.md#specify-protocol) for more details. | `http/protobuf` (see below)        | `OTEL_EXPORTER_OTLP_PROTOCOL` `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL` |
```
carlosalberto, line 20:
```suggestion
| Protocol             | The transport protocol. Options MAY include `grpc`, `http/protobuf`, and `http/json`. See [Specify Protocol](./exporter.md#specify-protocol) for more details. | `http/protobuf` [2]        | `OTEL_EXPORTER_OTLP_PROTOCOL` `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL` |
```

specification/protocol/exporter.md
```diff
@@ -95,9 +95,14 @@ The `OTEL_EXPORTER_OTLP_PROTOCOL`, `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL`, and `OT
 - `http/protobuf` for protobuf-encoded data over HTTP connection
 - `http/json` for JSON-encoded data over HTTP connection
 
-SDKs MUST support either `grpc` or `http/protobuf` and SHOULD support both. They also MAY support `http/json`.
+SDKs SHOULD support both `grpc` and `http/protobuf` transports and MUST support
```
carlosalberto, line 98:
```suggestion
**[2]**: SDKs SHOULD support both `grpc` and `http/protobuf` transports and MUST support
```

specification/protocol/exporter.md
tigrannajaryan: Done.

specification/protocol/exporter.md
tigrannajaryan: Done.

Issue #1920, opened 2021-09-14
otlp default endpoint port for http is different from collector receiver default


blumamir:
[According to the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#configuration-options), the default value for `Endpoint (OTLP/HTTP)` is `https://localhost:4317`, However, [the collector default listening port for otlp over HTTP is 4318](https://github.com/open-telemetry/opentelemetry-collector/tree/main/receiver/otlpreceiver#getting-started):

> endpoint (default = 0.0.0.0:4317 for grpc protocol, 0.0.0.0:4318 http protocol)

Should the spec change the default endpoint port for OTLP/HTTP from 4317 to 4318 to be synced with the collector's port?

Oberon00:
It seems like this is a recurring problem, we had a similar issue in #1816.
CC @tigrannajaryan

tigrannajaryan:
This is an unfortunate result of moving back and forth on the decision to have a single port or 2 separate ports. Collector currently has 2 ports but there is a draft PR which tries to combine them according to the spec.

We can change change the spec to match the reality for now and then if we are able to merge the ports in the Collector amend the spec again. The SDKs are/will likely use different ports and changing a default port is a breaking change so I am not sure if this is a good idea. I don't know how to nicely handle this so that it is not a breaking change. On the Collector side we can make the OTLP receiver accept data on both ports even after the port merging is implemented, but that doesn't help other receivers (e.g. backends).

Any thoughts on how to make this transition without breaking stuff?

carlosalberto:
As we have mentioned, at Lightstep we had a few problems using a single port for such functionality, and I'd rather have a unified port as a _feature_ (if the prototype proves it's all good). That would be a rather conservative position, but probably it's a good position at this time.

carlosalberto:
@tigrannajaryan any though after it was mentioned that https://github.com/open-telemetry/opentelemetry-collector/pull/3765 may not receive enough iterations in the near future?

tigrannajaryan:
I think given that we heard using one port caused pains for others in the past and given that we do not have an existing Collector implementation that uses one port and do not have anyone with experience who can contribute an implementation that uses one port we should stop pursuing one port and go with 2 ports.
Collector already uses 2 ports, it works and is known to not have any problems. I do not think it is very important to have one port and risk having problems down the road.

[...]

PR #1847, opened 2021-08-04
Revert "Change OTLP/HTTP port from 4317 to 4318 (#1839)"

tigrannajaryan:
Apparently there is new evidence that one port may work after all:
https://github.com/open-telemetry/opentelemetry-collector/pull/3765/files

This reverts commit cce1d5996873de38a68e05eafa4d5e224df9b8f1 until we have the final
decision about the ability to have a single port, see:

https://github.com/open-telemetry/opentelemetry-specification/issues/1846#issuecomment-892523208

PR #1970, opened 2021-09-27
Change OTLP/HTTP port from 4317 to 4318 (#1839)

tigrannajaryan:
Related to https://github.com/open-telemetry/opentelemetry-specification/issues/1816
Fixes https://github.com/open-telemetry/opentelemetry-specification/issues/1835
Fixes https://github.com/open-telemetry/opentelemetry-specification/issues/1920

Some historical context: we wanted to make grpc and http use the same port and we
had an open issue in the Collector to do so:
https://github.com/open-telemetry/opentelemetry-collector/issues/1256

The conclusion is that there are technical hurdles that make it unfeasible:
https://github.com/open-telemetry/opentelemetry-collector/issues/1256#issuecomment-880048833

Because of that we need to keep grpc and http ports separate. This means we need to
change the spec to say that otlp/http uses port 4318. Once this PR is merged we
will also need to submit for port 4318 registration with IANA like we did previously
with port 4317:
https://github.com/open-telemetry/opentelemetry-specification/issues/1148#issuecomment-725067404

There was also a draft PR to merge the ports in the Collector but it was not completed:
https://github.com/open-telemetry/opentelemetry-collector/pull/3765

Note that this change was initially submitted in PR https://github.com/open-telemetry/opentelemetry-specification/pull/1839
and then reverted in PR https://github.com/open-telemetry/opentelemetry-specification/pull/1847
because we hoped that the merging could be successfully done. We believe that we should
no longer pursue this and should consider the ports separate from now on.

Note 2: we consider this a spec bug fix (spec was impossible to implement), rather than a 
functionality change, that's why we believe this is an allowed change.


tigrannajaryan:
@open-telemetry/specs-approvers please review this PR with a long history of back-and-forth changes. Hopefully this is the last time we touch the port.

carlosalberto (approved):

iNikem (approved):

jsuereth (approved):

carlosalberto:
@open-telemetry/specs-approvers I think we should include this as part of 1.7 to avoid further confusion. Sounds about right?

bogdandrutu (approved):
Is this a breaking change?

codeboten (approved):

tigrannajaryan:
> Is this a breaking change?

Not for Collector, because Collector already uses separate ports 4317 and 4318 for a while now.

It may be a breaking change for a receiver that expects to receive OTLP/HTTP on port 4317. However I am not aware of any existing implementation that does and given that we met difficulties when trying to have one port on the receiver side it is very likely that no such receivers exist at all.

So, even if it is a breaking change the impact is likely very small and fixing it should be easy (just set an env var on the SDK side to use a different port).

From formal perspective, yes, it is a breaking change because the spec says OTLP/HTTP port is 4318 and it is a Stable section: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp
Formally we are not allowed to change any Stable section, but I feel that the real world impact is close to zero. 

However, since we are _unable_ to implement in the Collector what the spec says I think we can treat this as a spec bug, something that is simply wrong and needs to be fixed in the spec. This makes it an allowed change, we are supposed to fix spec bugs even if they mean someone who relied on the bug may be broken as a result of the change.

tigrannajaryan:
@open-telemetry/specs-approvers I will keep this open for one more day at least. If there are any doubts please speak up otherwise I will merge and after that will be strongly opposed to reversing/reverting this decision again.

anuraaga:
FWIW, `443` will be a highly common port anyways for non-collector use cases. So indeed the risk of the breaking change seems extremely low.

jmacd (approved):

carlosalberto:
@tigrannajaryan Are we really to roll? Or you want to wait a little longer?

tigrannajaryan:
> @tigrannajaryan Are we really to roll? Or you want to wait a little longer?

2 more hours to complete 24 hours since I promised at least one more day :-)

tigrannajaryan:
6 approvals, 2 days since last changed, I believe this can be merged.

tigrannajaryan:
@mtwo please file port 4318 with IANA (I believe you know how it's done since you did it last time).

mtwo:
I’ve forgotten, but I’ll read the docs and will file again :)

mtwo:
Filed

mtwo:
Update: my request to extend our existing reservation to the adjacent port was declined, as new port reservations require a new submission. I've since filed submission 1213852 with the IANA and I will reply here when I hear back.

prashant-shahi:
@tigrannajaryan any update on the IANA submission would be very appreciated.

tigrannajaryan:
IANA rejected the request since they don't believe the justification we provided warrants another port registration. The new port will remain unregistered for now until we find stronger arguments to ask for the new port.

anuraaga:
FWIW I completely agree with IANA's decision. HTTP is HTTP regardless of whether the application decides to send gRPC frames or not.

tigrannajaryan:
> FWIW I completely agree with IANA's decision. HTTP is HTTP regardless of whether the application decides to send gRPC frames or not.

@anuraaga Can you help make it happen in the Collector's OTLP receiver? We tried and were unable.

anuraaga:
Maybe, but I am wondering why as the creators of both gRPC and OpenTelemetry (via OC), Google isn't stepping up to take responsibility and fix the mess.

PR #1972, opened 2021-09-27
Allow unit to be specified in View

### LINKED ISSUE #1822: Support `unit` customization in Metrics View
https://github.com/open-telemetry/opentelemetry-specification/pull/1730/files#r668191496

reyang:
Fixes #1822.

### DIFF
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -151,6 +151,7 @@ are the inputs:
   stream](./datamodel.md#events--data-stream--timeseries).
 * The configuration for the resulting [metrics
   stream](./datamodel.md#events--data-stream--timeseries):
+  * The `unit`. If not provided, the Instrument `unit` would be used by default.
   * The `description`. If not provided, the Instrument `description` would be
     used by default.
   * A list of `attribute keys` (optional). If not provided, all the attribute


jsuereth (approved):

cijothomas (approved):

specification/metrics/sdk.md
```diff
@@ -151,6 +151,7 @@ are the inputs:
   stream](./datamodel.md#events--data-stream--timeseries).
 * The configuration for the resulting [metrics
   stream](./datamodel.md#events--data-stream--timeseries):
+  * The `unit`. If not provided, the Instrument `unit` would be used by default.
```
jkwatson, line 154: this doesn't imply any sort of auto-conversion of units, correct?

SergeyKanzhelev: +1, would be nice to clarify in text.

bogdandrutu: It is not correct to say that "the Instrument `unit` would be used by default". As discussed previously please provide examples/use cases when this is needed. I cannot find any use-case besides the auto-conversion in case of "Count" to drop the unit since no longer apply, but besides that I cannot find any other example. I am reluctant on adding more configs unless we prove they are needed.

bogdandrutu (changes requested):

reyang:
Discussed during the [09/28/2021 Metrics SIG Meeting](https://docs.google.com/document/d/1pdvPeKjA8v8w_fGKAN68JjWBmVJtPCpqdi9IZrd6eEo/edit#), moving this to post-Stable release, abandon the PR for now.

PR #1973, opened 2021-09-27
Remove db.cassandra.keyspace and db.hbase.namespace, and clarify db.name

### LINKED ISSUE #1760: Propose to remove db.cassandra.keyspace and use db.name instead

Consistency across different database client spans.

Cassandra keyspace captured as `db.name`, similar to other database client spans.

**Additional context:**

The specification [currently says](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#cassandra) that `db.cassandra.keyspace` is

> to be used instead of the generic `db.name` attribute.

But I'm not sure why there is this special case for Cassandra. It seems to make sense to use `db.name` instead.

This attribute appears to come from #575, but I couldn't find any discussion in that PR (or in the related issue) about why `db.cassandra.keyspace` should be used instead of `db.name`.

This issue stems from discussion in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3199

trask:
Fixes #1760

## Changes

Removes `db.cassandra.keyspace` and `db.hbase.namespace` (in preference for using `db.name`).

Clarifies `db.name` in the case where a database product provides multiple concepts that could be mapped to `db.name`.

### DIFF
--- a/semantic_conventions/trace/database.yaml
+++ b/semantic_conventions/trace/database.yaml
@@ -176,15 +176,16 @@ groups:
         tag: call-level
         type: string
         required:
-          conditional: >
-            Required, if applicable and no more-specific attribute is defined.
+          conditional: Required, if applicable.
         brief: >
-          If no [tech-specific attribute](#call-level-attributes-for-specific-technologies)
-          is defined, this attribute is used to report the name of the database being accessed.
+          This attribute is used to report the name of the database being accessed.
           For commands that switch the database, this should be set to the target database
           (even if the command fails).
         note: >
           In some SQL databases, the database name to be used is called "schema name".
+          In case there are multiple layers that could be considered for database name
+          (e.g. Oracle instance name and schema name),
+          the database name to be used is the more specific layer (e.g. Oracle schema name).
         examples: [ 'customers', 'main' ]
       - id: statement
         tag: call-level
@@ -256,13 +257,6 @@ groups:
     brief: >
       Call-level attributes for Cassandra
     attributes:
-      - id: keyspace
-        type: string
-        required: always
-        tag: call-level-tech-specific-cassandra
-        brief: >
-          The name of the keyspace being accessed. To be used instead of the generic `db.name` attribute.
-        examples: 'mykeyspace'
       - id: page_size
         type: int
         tag: call-level-tech-specific-cassandra
@@ -302,7 +296,7 @@ groups:
         tag: call-level-tech-specific-cassandra
         required:
           conditional: Recommended if available.
-        brief: The name of the primary table that the operation is acting upon, including the schema name (if applicable).
+        brief: The name of the primary table that the operation is acting upon, including the database name (if applicable).
         note: >
           This mirrors the db.sql.table attribute but references cassandra rather than sql.
           It is not recommended to attempt any client-side parsing of
@@ -334,21 +328,6 @@ groups:
         brief: >
           The data center of the coordinating node for a query.
         examples: 'us-west-2'
-
-  - id: db.hbase
-    prefix: db.hbase
-    extends: db
-    brief: >
-      Call-level attributes for Apache HBase
-    attributes:
-      - id: namespace
-        type: string
-        required: always
-        tag: call-level-tech-specific
-        brief: >
-          The [HBase namespace](https://hbase.apache.org/book.html#_namespace) being accessed.
-          To be used instead of the generic `db.name` attribute.
-        examples: 'default'
 
   - id: db.redis
     prefix: db.redis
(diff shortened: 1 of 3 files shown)


tigrannajaryan:
> Fixes #1760
> 
> ## Changes
> Removes `db.cassandra.keyspace` and `db.hbase.namespace` (in preference for using `db.name`).
> 
> Clarifies `db.name` in the case where a database product provides multiple concepts that could be mapped to `db.name`.

Do we expect that the exact same value that we previously recommended to record in `db.cassandra.keyspace` and `db.hbase.namespace` to be now recorded in `db.name`? If that is the case then this is semantically a renaming of an attribute from schema evolution perspective and should be captured in the schema file for the next release (in the [trace attributes section](https://github.com/open-telemetry/oteps/blob/main/text/0152-telemetry-schemas.md#rename_attributes-transformation) and any other applicable section).

trask:
> Do we expect that the exact same value that we previously recommended to record in `db.cassandra.keyspace` and `db.hbase.namespace` to be now recorded in `db.name`? If that is the case then this is semantically a renaming of an attribute from schema evolution perspective and should be captured in the schema file for the next release

👍 schema translation added

iNikem (approved):

semantic_conventions/trace/database.yaml
```diff
@@ -176,15 +176,16 @@ groups:
...
-          If no [tech-specific attribute](#call-level-attributes-for-specific-technologies)
-          is defined, this attribute is used to report the name of the database being accessed.
+          This attribute is used to report the name of the database being accessed.
           For commands that switch the database, this should be set to the target database
           (even if the command fails).
         note: >
           In some SQL databases, the database name to be used is called "schema name".
+          In case there are multiple layers that could be considered for database name
+          (e.g. Oracle instance name and schema name),
+          the database name to be used is the more specific layer (e.g. Oracle schema name).
```
Oberon00, line 188: What is the criteria for considering a layer as database name?

Blocking comment: We lost the wording that for hbase we must use the namespace and for Cassandra the keyspace name as db.name.

Oberon00 (changes requested):

schemas/1.7.0
```diff
@@ -0,0 +1,12 @@
...
+schema_url: https://opentelemetry.io/schemas/1.7.0
+versions:
+  1.7.0:
+    spans:
+      changes:
+        - rename_attributes:
+            attribute_map:
+              db.cassandra.keyspace: db.name
+              db.hbase.namespace: db.name
+  1.5.0:
```
mateuszrzeszutek, line 11: Missing 1.6.1?

semantic_conventions/trace/database.yaml
trask: > What is the criteria for considering a layer as database name?

I don't have an answer for this, but it doesn't seem any more vague than the existing `db.name` definition.

> Blocking comment: We lost the wording that for hbase we must use the namespace and for Cassandra the keyspace name as db.name.

fixed 👍

specification/trace/semantic_conventions/database.md
```diff
@@ -194,10 +195,9 @@ Separated for clarity.
 <!-- semconv db.tech(tag=call-level-tech-specific-cassandra) -->
 | Attribute  | Type | Description  | Examples  | Required |
 |---|---|---|---|---|
-| `db.cassandra.keyspace` | string | The name of the keyspace being accessed. To be used instead of the generic `db.name` attribute. | `mykeyspace` | Yes |
 | `db.cassandra.page_size` | int | The fetch size used for paging, i.e. how many rows will be returned at once. | `5000` | No |
 | `db.cassandra.consistency_level` | string | The consistency level of the query. Based on consistency values from [CQL](https://docs.datastax.com/en/cassandra-oss/3.0/cassandra/dml/dmlConfigConsistency.html). | `all` | No |
-| `db.cassandra.table` | string | The name of the primary table that the operation is acting upon, including the schema name (if applicable). [1] | `mytable` | Recommended if available. |
+| `db.cassandra.table` | string | The name of the primary table that the operation is acting upon, including the database name (if applicable). [1] | `mytable` | Recommended if available. |
```
Oberon00, line 200: Are you sure this change makes sense? It might change the meaning, I don't know enough about cassandra to know.

specification/trace/semantic_conventions/database.md
```diff
@@ -170,17 +168,20 @@ If, for example, the [`HMSET` command][] is invoked, `"HMSET myhash field1 'Hell
 In **CouchDB**, `db.operation` should be set to the HTTP method + the target REST route according to the API reference documentation.
 For example, when retrieving a document, `db.operation` would be set to (literally, i.e., without replacing the placeholders with concrete values): [`GET /{db}/{docid}`][CouchDB get doc].
 
+In **Cassandra**, `db.name` should be set to the keyspace name.
+
+In **HBase**, `db.name` should be set to the HBase namespace.
```
Oberon00, line 173:
```suggestion
In **Cassandra**, `db.name` SHOULD be set to the keyspace name.

In **HBase**, `db.name` SHOULD be set to the HBase namespace.
```

specification/trace/semantic_conventions/database.md
trask: good point, I updated it to `keyspace`

carlosalberto (approved):

trask:
@open-telemetry/specs-approvers @open-telemetry/specs-trace-approvers

any additional feedback on this?

anuraaga (approved):

Issue #328, opened 2019-10-20
Exporter configuration


tsloughter:
I still need to do some renaming in the Erlang implementation to match the spec, but I completed implementation of the exporter process yesterday (https://github.com/open-telemetry/opentelemetry-erlang) and had chose a different strategy which I planned to bring up as an otep because I didn't realize there was already a spec that mostly covered it :)

I wanted to still mention it before switching to the current spec in case others think it could be useful as an update/option to the current spec.

Basically the difference is that `maxQueueSize` is not checked when finishing a span and it does not drop a span if it has reached the max. Instead, the exporter continually checks the queue size and if it reaches the configured maximum it kills the current export process, if one is ongoing, and immediately calls the exporter on the current queue. So there will potentially be dropped span, just not as a blocking operation when finishing a span.

---

Sort of unrelated, I think `maxExportBatchSize` should be optional to support. There is already a max queue size and in some implementations it may be more efficient to not do any additional batching.

For example, the Erlang impl stores finished spans in a shared in-memory table, after `scheduledDelayMillis` it points future finished spans to a new table and the former table is passed to the user configured `exporter` (with precautions so that it can't be written to again). The exporter can read as many or as a few of the spans at a time from the table.

Only a reference to the table is passed as an argument, not the actual spans. To support a batch size would entail extra processing and memory usage to read the batch size spans from the table into the process's heap.

Actually, went ahead and opened a PR to make the max export batch size a "MAY" part of the spec: https://github.com/open-telemetry/opentelemetry-specification/pull/327

andrewhsu:
from the spec sig today, curious to know if this issue needs another PR to resolve since #327 was closed?

andrewhsu:
may need more guidance on prioritization, setting to P2 for now

tsloughter:
@andrewhsu I will close this.

I think the SDK spec should be looser on its requirements since there are multiple ways to achieve the same goal that have different implications on performance for different languages/runtimes, but I also don't think this issue is clear on that and would be better done with a new PR, or a new issue if I find it needs more discussion before a PR.

PR #1974, opened 2021-09-27
Add specification for otlp retry configuration parameters and defaults

### LINKED ISSUE #1742: Additional specification around OTLP exporter retry requirements

The OTLP exporter spec states the following concerning [retry logic](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#retry): 

> Transient errors MUST be handled with a retry strategy. This retry strategy MUST implement an exponential back-off with jitter to avoid overwhelming the destination until the network is restored or the destination has recovered.

There should be more specificity around how exactly the retry strategy should be configured with gRPC. [Here's the proto](https://github.com/grpc/grpc-proto/blob/master/grpc/service_config/service_config.proto#L129-L151) definition for the gRPC Retry Policies. Without more specificity, language implementations may end up choosing arbitrary values for these parameters. 

Perhaps there should also be configuration parameters for adjusting the retry parameters. Maybe retry can be enabled / disabled altogether, or maybe the various parameters can be tuned. 

**Additional context.**

Discussing this in an opentelemetry-java [issue](https://github.com/open-telemetry/opentelemetry-java/issues/3271).

jack-berg:
Fixes #1742.

## Changes

Expand the OTLP exporter retry language to include language for using the built-in gRPC retry mechanism. This allows SDKs to leverage the mechanism as a standard without precluding the use of other retry mechanisms (such as the one [used in the collector](https://github.com/open-telemetry/opentelemetry-collector/blob/0e3b991f9f2a5310ad60b26dfba8897d7bb94745/exporter/exporterhelper/queued_retry.go#L255)).

It also establishes declares the configurable parameters if the built-in gRPC retry mechanism is used, with default values. 

Related issues #

- opentelemetry-java [PR #3636](https://github.com/open-telemetry/opentelemetry-java/pull/3636) prototypes adding a retry mechanism but is blocked on spec language
- opentelemetry-php [Issue #328](https://github.com/open-telemetry/opentelemetry-php/issues/328) requests an OTLP retry implementation
- opentelemetry-dotnet [Issue #1779](https://github.com/open-telemetry/opentelemetry-dotnet/issues/1779) requests an OTLP retry implementation

### DIFF
--- a/specification/protocol/exporter.md
+++ b/specification/protocol/exporter.md
@@ -62,6 +62,20 @@ The `OTEL_EXPORTER_OTLP_HEADERS`, `OTEL_EXPORTER_OTLP_TRACES_HEADERS`, `OTEL_EXP
 
 Transient errors MUST be handled with a retry strategy. This retry strategy MUST implement an exponential back-off with jitter to avoid overwhelming the destination until the network is restored or the destination has recovered.
 
-For OTLP/HTTP, the errors `408 (Request Timeout)` and `5xx (Server Errors)` are defined as transient, detailed information about erros can be found in the [HTTP failures section](otlp.md#failures). For the OTLP/gRPC, the full list of the gRPC retryable status codes can be found in the [gRPC response section](otlp.md#otlpgrpc-response).
+For OTLP/HTTP, the errors `408 (Request Timeout)` and `5xx (Server Errors)` are defined as transient, detailed information about errors can be found in the [HTTP failures section](otlp.md#failures). For the OTLP/gRPC, the full list of the gRPC retryable status codes can be found in the [gRPC response section](otlp.md#otlpgrpc-response).
+
+SDKs MAY use the built-in [gRPC Retry](https://github.com/grpc/proposal/blob/master/A6-client-retries.md) mechanism to facilitate exponential back-off. If the built-in gRPC mechanism is used, the following values SHOULD be available for configuration:
+
+- `maxAttempts`: The maximum number of attempts, including the original request. Must be an integer greater than 1 and less than 6. (The built-in gRPC mechanism treats values greater than 5 as 5.) Defaults to `5`.
+- `initialBackoff`: Must be a duration greater than 0. Defaults to `1s`
+- `maxBackoff`: Must be a duration greater than 0. Defaults to `5s`.
+- `backoffMultiplier` Must be a number greater than 0. Defaults to `1.5`.
+
+These properties are used to compute the backoff as follows:
+
+- The initial retry attempt will occur after `random(0, initialBackoff)`
+- The `n`-th retry attempt will occur after `random(0, min(initialBackoff*backoffMultiplier**(n-1), maxBackoff))`
+
+Language SDKs SHOULD have retry configuration and mechanics that are consistent across protocols. For example, if the built-in gRPC Retry mechanism is used for the `grpc` protocol, the `http/protobuf` and `http/json` protocols should expose the same configuration options and compute the backoff duration in the same manner.
 
 [otlphttp-req]: otlp.md#otlphttp-request
--- a/specification/protocol/exporter.md
+++ b/specification/protocol/exporter.md
@@ -66,16 +66,16 @@ For OTLP/HTTP, the errors `408 (Request Timeout)` and `5xx (Server Errors)` are 
 
 SDKs MAY use the built-in [gRPC Retry](https://github.com/grpc/proposal/blob/master/A6-client-retries.md) mechanism to facilitate exponential back-off. If the built-in gRPC mechanism is used, the following values SHOULD be available for configuration:
 
-- `maxAttempts`: The maximum number of attempts, including the original request. Must be an integer greater than 1 and less than 6. (The built-in gRPC mechanism treats values greater than 5 as 5.) Defaults to `5`.
-- `initialBackoff`: Must be a duration greater than 0. Defaults to `1s`
-- `maxBackoff`: Must be a duration greater than 0. Defaults to `5s`.
-- `backoffMultiplier` Must be a number greater than 0. Defaults to `1.5`.
+- `maxAttempts`: The maximum number of attempts, including the original request. Must be an integer greater than 1 and less than 6.
+- `initialBackoff`: Must be a duration greater than 0.
+- `maxBackoff`: Must be a duration greater than 0.
+- `backoffMultiplier` Must be a number greater than 0.
 
-These properties are used to compute the backoff as follows:
+SDKs have unspecified default values for these properties. They are used to compute the backoff as follows:
 
 - The initial retry attempt will occur after `random(0, initialBackoff)`
 - The `n`-th retry attempt will occur after `random(0, min(initialBackoff*backoffMultiplier**(n-1), maxBackoff))`
 
-Language SDKs SHOULD have retry configuration and mechanics that are consistent across protocols. For example, if the built-in gRPC Retry mechanism is used for the `grpc` protocol, the `http/protobuf` and `http/json` protocols should expose the same configuration options and compute the backoff duration in the same manner.
+Language SDKs SHOULD have retry configuration and mechanics that are consistent across OTLP protocols. For example, if the built-in gRPC Retry mechanism is used for the `grpc` protocol, the `http/protobuf` and `http/json` protocols should expose the same configuration options and compute the backoff duration in the same manner.
 
 [otlphttp-req]: otlp.md#otlphttp-request
--- a/specification/protocol/exporter.md
+++ b/specification/protocol/exporter.md
@@ -64,18 +64,8 @@ Transient errors MUST be handled with a retry strategy. This retry strategy MUST
 
 For OTLP/HTTP, the errors `408 (Request Timeout)` and `5xx (Server Errors)` are defined as transient, detailed information about errors can be found in the [HTTP failures section](otlp.md#failures). For the OTLP/gRPC, the full list of the gRPC retryable status codes can be found in the [gRPC response section](otlp.md#otlpgrpc-response).
 
-SDKs MAY use the built-in [gRPC Retry](https://github.com/grpc/proposal/blob/master/A6-client-retries.md) mechanism to facilitate exponential back-off. If the built-in gRPC mechanism is used, the following values SHOULD be available for configuration:
+SDKs MAY use the built-in [gRPC Retry](https://github.com/grpc/proposal/blob/master/A6-client-retries.md) mechanism to facilitate exponential back-off.
 
-- `maxAttempts`: The maximum number of attempts, including the original request. Must be an integer greater than 1 and less than 6.
-- `initialBackoff`: Must be a duration greater than 0.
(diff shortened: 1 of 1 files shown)


specification/protocol/exporter.md
```diff
@@ -62,6 +62,20 @@ The `OTEL_EXPORTER_OTLP_HEADERS`, `OTEL_EXPORTER_OTLP_TRACES_HEADERS`, `OTEL_EXP
...
+For OTLP/HTTP, the errors `408 (Request Timeout)` and `5xx (Server Errors)` are defined as transient, detailed information about errors can be found in the [HTTP failures section](otlp.md#failures). For the OTLP/gRPC, the full list of the gRPC retryable status codes can be found in the [gRPC response section](otlp.md#otlpgrpc-response).
+
+SDKs MAY use the built-in [gRPC Retry](https://github.com/grpc/proposal/blob/master/A6-client-retries.md) mechanism to facilitate exponential back-off. If the built-in gRPC mechanism is used, the following values SHOULD be available for configuration:
+
+- `maxAttempts`: The maximum number of attempts, including the original request. Must be an integer greater than 1 and less than 6. (The built-in gRPC mechanism treats values greater than 5 as 5.) Defaults to `5`.
+- `initialBackoff`: Must be a duration greater than 0. Defaults to `1s`
+- `maxBackoff`: Must be a duration greater than 0. Defaults to `5s`.
+- `backoffMultiplier` Must be a number greater than 0. Defaults to `1.5`.
```
jack-berg, line 72: The defaults for these were informed by the defaults for the [Batch Span Processor](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#batch-span-processor), which specifies a default max of 30s for maximum time to export.

If 5 total attempts are to be done (4 retries),  the default parameters would produce the following target backoff durations (before jitter randomness is applied):
- Attempt 1: 0 delay
- Attempt 2 (retry 1): 1s
- Attempt 3 (retry 2): 2.25s
- Attempt 4 (retry 3): 3.375s
- Attempt 5 (retry 4): 5s

That means that if round trips for each of the attempts is instantaneous, 11.625s of the 30s would be spend waiting for retries. This affords each of the 5 attempts 3.675s (`=(30-11.625)/5`) to complete, which seemed reasonable to me.

I looked at the default values for the `queued_retry` helper in the [collector](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/queued_retry.go#L54-L75), but thought that its 5 minute total time made the other settings incompatible with the 30s default max export time of the batch span processor.

tigrannajaryan:
>opentelemetry-java PR #3636 prototypes adding a retry mechanism but is blocked on spec language

Can you please point to the part of the spec that is blocking this?

specification/protocol/exporter.md
tigrannajaryan: 1. Why do we need the retries to complete within the batch processor's default export time?
2. I see exact numbers for attempt intervals. Does this mean no jitter is used by grpc's built-in retrying mechanism?

specification/protocol/exporter.md
```diff
@@ -62,6 +62,20 @@ The `OTEL_EXPORTER_OTLP_HEADERS`, `OTEL_EXPORTER_OTLP_TRACES_HEADERS`, `OTEL_EXP
 
 Transient errors MUST be handled with a retry strategy. This retry strategy MUST implement an exponential back-off with jitter to avoid overwhelming the destination until the network is restored or the destination has recovered.
 
-For OTLP/HTTP, the errors `408 (Request Timeout)` and `5xx (Server Errors)` are defined as transient, detailed information about erros can be found in the [HTTP failures section](otlp.md#failures). For the OTLP/gRPC, the full list of the gRPC retryable status codes can be found in the [gRPC response section](otlp.md#otlpgrpc-response).
+For OTLP/HTTP, the errors `408 (Request Timeout)` and `5xx (Server Errors)` are defined as transient, detailed information about errors can be found in the [HTTP failures section](otlp.md#failures). For the OTLP/gRPC, the full list of the gRPC retryable status codes can be found in the [gRPC response section](otlp.md#otlpgrpc-response).
+
+SDKs MAY use the built-in [gRPC Retry](https://github.com/grpc/proposal/blob/master/A6-client-retries.md) mechanism to facilitate exponential back-off. If the built-in gRPC mechanism is used, the following values SHOULD be available for configuration:
```
reyang, line 67: For retry, I guess we need to consider the following situation:
1. When transport error happened - the exporter didn't hear back from the receiver. (in this case the retry policy might be different from the situation where the exporter got the server response - where the server might give hint on Retry-After or Server Too Busy / Not Available).
2. When partial success occurred - e.g. HTTP 206. The response payload needs to include what succeeded and what failed, and the exporter will need to understand and respect that.
3. How to de-dupe - if there is no partial success, a simple unique sequence id (scoped to the connection or session) might be used. If there is partial success, a more sophisticated mechanism might be needed.

specification/protocol/exporter.md
jack-berg: > 1. Why do we need the retries to complete within the batch processor's default export time?
Its a matter of how to interpret the batch processors export time. Should it be the time allotted for a single attempt of potentially many? Or should it be the total allotted time for all attempts? I think total allotted time is more intuitive. 
> 2. I see exact numbers for attempt intervals. Does this mean no jitter is used by grpc's built-in retrying mechanism?
I suppose my comment was a bit misleading. gRPC does support jitter. The way it works is that a target upper bound for delay is computed from the config, and then a `random(0, upperBound)` is selected.

specification/protocol/exporter.md
```diff
@@ -62,6 +62,20 @@ The `OTEL_EXPORTER_OTLP_HEADERS`, `OTEL_EXPORTER_OTLP_TRACES_HEADERS`, `OTEL_EXP
...
+- `initialBackoff`: Must be a duration greater than 0. Defaults to `1s`
+- `maxBackoff`: Must be a duration greater than 0. Defaults to `5s`.
+- `backoffMultiplier` Must be a number greater than 0. Defaults to `1.5`.
+
+These properties are used to compute the backoff as follows:
+
+- The initial retry attempt will occur after `random(0, initialBackoff)`
+- The `n`-th retry attempt will occur after `random(0, min(initialBackoff*backoffMultiplier**(n-1), maxBackoff))`
+
+Language SDKs SHOULD have retry configuration and mechanics that are consistent across protocols. For example, if the built-in gRPC Retry mechanism is used for the `grpc` protocol, the `http/protobuf` and `http/json` protocols should expose the same configuration options and compute the backoff duration in the same manner.
```
reyang, line 79: Maybe call out explicitly "across OTLP transports".

I guess it is impossible to have consistent retry configuration/mechanism across all protocols. Maybe it is possible to have consistency across OTLP protocols, but definitely not the same for Prometheus.

specification/protocol/exporter.md
jack-berg: In the 10/5/2021 Spec SIG, we discussed that retry behavior in partial success scenarios could be a separate area of future work.

Upon closer inspection, of the otlp spec, I believe each of these points are already documented as known limitations. 

> 1. When transport error happened - the exporter didn't hear back from the receiver. (in this case the retry policy might be different from the situation where the exporter got the server response - where the server might give hint on Retry-After or Server Too Busy / Not Available).

This falls under the [Duplicate Data](https://github.com/open-telemetry/opentelemetry-specification/blame/main/specification/protocol/otlp.md#L506-L511) limitation.

> 2. When partial success occurred - e.g. HTTP 206. The response payload needs to include what succeeded and what failed, and the exporter will need to understand and respect that.

This falls under the [Partial Success](https://github.com/open-telemetry/opentelemetry-specification/blame/main/specification/protocol/otlp.md#L516) limitation. 

> 3. How to de-dupe - if there is no partial success, a simple unique sequence id (scoped to the connection or session) might be used. If there is partial success, a more sophisticated mechanism might be needed.

This falls under the [Request Acknowledgements](https://github.com/open-telemetry/opentelemetry-specification/blame/main/specification/protocol/otlp.md#L506) / [Partial Success](https://github.com/open-telemetry/opentelemetry-specification/blame/main/specification/protocol/otlp.md#L516) limitations. 

I could add language clarifying that these limitations also apply to retry attempts, but that's already implied. I think the OTLP spec's language is already sufficiently clear on what happens in retry edge cases: data duplication can occur if things like network interruptions happen, and there's no accommodation for partial success (for retry attempts or in general).

specification/protocol/exporter.md
jack-berg: Per the conversation in the 10/05/2021 Spec SIG, I've removed default values from these properties, clarifying that SDKs have their own unspecified defaults. 

Resolving this convo since the subject was the default values I had proposed.

specification/protocol/exporter.md
jack-berg: I've done this in the latest commit.

specification/protocol/exporter.md
```diff
@@ -62,6 +62,20 @@ The `OTEL_EXPORTER_OTLP_HEADERS`, `OTEL_EXPORTER_OTLP_TRACES_HEADERS`, `OTEL_EXP
...
+For OTLP/HTTP, the errors `408 (Request Timeout)` and `5xx (Server Errors)` are defined as transient, detailed information about errors can be found in the [HTTP failures section](otlp.md#failures). For the OTLP/gRPC, the full list of the gRPC retryable status codes can be found in the [gRPC response section](otlp.md#otlpgrpc-response).
+
+SDKs MAY use the built-in [gRPC Retry](https://github.com/grpc/proposal/blob/master/A6-client-retries.md) mechanism to facilitate exponential back-off. If the built-in gRPC mechanism is used, the following values SHOULD be available for configuration:
+
+- `maxAttempts`: The maximum number of attempts, including the original request. Must be an integer greater than 1 and less than 6.
+- `initialBackoff`: Must be a duration greater than 0.
+- `maxBackoff`: Must be a duration greater than 0.
+- `backoffMultiplier` Must be a number greater than 0.
+
+SDKs have unspecified default values for these properties. They are used to compute the backoff as follows:
```
anuraaga, line 74: Wasn't in the SIG where I think it was discussed, but why do we leave the defaults unspecified? Actually having agreed upon defaults to use in the implementation was one of the main points I was hoping for from the spec language.

anuraaga: If gRPC had defaults to copy from, using those could probably work well but indeed I can't find them.

I think having defaults in the spec is important because I expect all languages to have the same retry behavior as a user - there doesn't seem to be anyone language specific about it. We've seen some env variables diverge, probably due to ordering of SDK implementation and spec, but I think the intent is for it to define cross-language concerns where possible to avoid divergence. Making it easier to implement an SDK is a great side-benefit.

jack-berg: I came to the same conclusion as @anuraaga [here](https://github.com/open-telemetry/opentelemetry-specification/pull/1974#discussion_r723373951). We need this language to ensure a consistent user experience.

Issue #1925, opened 2021-09-15
Reading semantic convention tables is cumbersome


Oberon00:
(This has to be implemented in https://github.com/open-telemetry/build-tools/tree/main/semantic-conventions but affects mostly the spec and needs discussion, so opening issue here)

The current presentation of the semantic convention attribute tables has a few problems:

1. Some tables are too wide so you need horizontal scrolling. If the table is also very long, this becomes very awkward since you have to scroll down to scroll right and then scroll up again. E.g., https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#messaging-attributes and https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#common-attributes are such a tables on my screen.
2. The definitions of enum members are easy to miss since they come after the table and there is no hint that further information follows within the table (AFAIR previous versions of the semantic convention generator at least added "string enum" instead of just "string" as type, but this was removed). E.g. https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#messaging-attributes  the `destination_kind` attribute.
3. The notes and required-conditions are annotated with `[1]`, `[2]` etc. If you read the note, it is often not obvious to which attribute it belongs. This makes it impossible to read a semantic convention from top to bottom without scrolling up and down constantly. E.g. https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#messaging-attributes

Worst case currently is that you have a enum with a note and a condition and enum members can have a note as well. Sample UX:

* Read definition of enum in table. Scroll down to the note for the definition. Scroll up again because you forgot the number of the note for the condition. Scroll up again to continue reading in the table.
* Under the table, realize that this is an enum. Scroll up again to the definition and read it because you already forgot it.
* Scroll down again.
* Read the enum members. If its a long enum with many members, don't forget to scroll up and down repeatedly to read any notes on enum members.

I would like to have a presentation that you can read in a single pass without cross-referencing stuff below the table and inside the table. For example, this could be a presentation as a list, so there is enough space to inline the notes and enum definitions. This could be complemented by a (very) short list of attributes with links at the top.

yurishkuro:
How important is it to even have the table? Looking at the examples you provided, not only do you need to scroll the table, but the formatting is also 80% whitespace. A bullet-list format would've been more compact:
[...]

PR #1975, opened 2021-09-28
Clarify OTEL_EXPORTER_OTLP_ENDPOINT envvar for OTLP/HTTP.

Oberon00:
The current wording of the OTLP endpoint config was confusing, especially the "if not present already" wording. Instead, clarify that we must always append when using the envvar for all signals (it was already clearly specified that the per-signal vars do not get the path appended).

This came up in https://github.com/open-telemetry/opentelemetry-java/issues/3650 and again at https://github.com/open-telemetry/opentelemetry-java/pull/3666#discussion_r717347648.

Also make it a MUST (not SHOULD) since this kind of thing would be extremely annoying to have differences per-language in. Also, without appending, the variable cannot be used to configure more than one signal which would defeat its sole purpose.


specification/protocol/exporter.md
```diff
@@ -10,7 +10,7 @@ The following configuration options MUST be available to configure the OTLP expo
...
 | Endpoint (OTLP/gRPC) | Target to which the exporter is going to send spans or metrics. The endpoint SHOULD accept any form allowed by the underlying gRPC client implementation. Additionally, the endpoint MUST accept a URL with a scheme of either `http` or `https`. A scheme of `https` indicates a secure connection and takes precedence over the `insecure` configuration setting. If the gRPC client implementation does not support an endpoint with a scheme of `http` or `https` then the endpoint SHOULD be transformed to the most sensible format for that implementation. | `https://localhost:4317` | `OTEL_EXPORTER_OTLP_ENDPOINT` `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` |
 | Insecure             | Whether to enable client transport security for the exporter's gRPC connection. This option only applies to OTLP/gRPC - OTLP/HTTP always uses the scheme provided for the `endpoint`. Implementations MAY choose to not implement the `insecure` option if it is not required or supported by the underlying gRPC client implementation. | `false` | `OTEL_EXPORTER_OTLP_INSECURE` `OTEL_EXPORTER_OTLP_SPAN_INSECURE` `OTEL_EXPORTER_OTLP_METRIC_INSECURE` |
```
Oberon00, line 15: (We should also re-think the table-format, it is gruesome to look at in diffs with the long lines. Maybe a list of bullet points would be better. Cf. also #1925)

carlosalberto (approved):

specification/protocol/exporter.md
```diff
@@ -10,7 +10,7 @@ The following configuration options MUST be available to configure the OTLP expo
...
+| Endpoint (OTLP/HTTP) | Target to which the exporter is going to send spans or metrics. The endpoint MUST be a valid URL with scheme (http or https) and host, and MAY contain a port and path. A scheme of https indicates a secure connection. When using `OTEL_EXPORTER_OTLP_ENDPOINT`, exporters MUST append the version and signal to the path (`/v1/traces` for traces and `/v1/metrics` for metrics, correcting for double-slashes). The per-signal endpoint configuration options take precedence and can be used to override this behavior. See the [OTLP Specification][otlphttp-req] for more details. | `https://localhost:4317` | `OTEL_EXPORTER_OTLP_ENDPOINT` `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` |
```
tigrannajaryan, line 13: What does `correcting for double-slashes` mean?

Oberon00: If you specify `OTEL_EXPORTER_OTLP_ENDPOINT=http://example.com/foo/` and you literally append `/v1/traces`, you would get `http://example.com/foo//v1/traces` with a double-slash inside the path of the URL. That wording wants to say that you should "correct" that to `http://example.com/foo/v1/traces`.

Maybe there is a better way to word this?
MUST send traces to the path `v1/traces` relative to  `OTEL_EXPORTER_OTLP_ENDPOINT` maybe? That also sounds a bit weird.

iNikem: So if there is already a path in `OTEL_EXPORTER_OTLP_ENDPOINT`, then we should preserve that path and append `v1/traces` to it?

Oberon00: Yes. E.g. if you have `http://example.com/collectorapi/`, then you should get `http://example.com/collectorapi/v1/traces` for the traces.
This does not change with this PR.

Oberon00: Since we only have a single line in the markdown source, adding examples seems hard, but I'll try.

tigrannajaryan: Yeah, the table format is not very good for lots of content. Maybe add as a footnote?

Oberon00: I saw that there already are some examples below. I extended them, added a more elaborate description above them and linked from the table in 6ca48b9.

specification/protocol/exporter.md
```diff
@@ -28,24 +28,65 @@ Supported values for `OTEL_EXPORTER_OTLP_*COMPRESSION` options:
...
+#### Example 3
+
+The following configuration sends all signals except for metrics to the same collector:
+
+```bash
+export OTEL_EXPORTER_OTLP_ENDPOINT=http://collector:4317/mycollector/
+export OTEL_EXPORTER_OTLP_METRICS_ENDPOINT=https://collector.example.com/v1/metrics/
+```
+
+Traces are sent to `http://collector:4317/mycollector/mycollector/v1/traces`
```
fbogsany, line 85:
```suggestion
Traces are sent to `http://collector:4317/mycollector/v1/traces`
```

tigrannajaryan (approved):
LGTM, once duplicate `mycollector/` that @fbogsany noticed is fixed.

arminru (approved):

specification/protocol/exporter.md
```diff
@@ -28,24 +28,65 @@ Supported values for `OTEL_EXPORTER_OTLP_*COMPRESSION` options:
...
+   Non-normatively, this could be implemented by ensuring that the base URL ends with
+   a slash and then appending the relative URLs as strings.
+
+#### Example 1
 
 The following configuration sends all signals to the same collector:
 
 ```bash
-export OTEL_EXPORTER_OTLP_ENDPOINT=http://collector:4317
+export OTEL_EXPORTER_OTLP_ENDPOINT=http://collector:4318
```
Oberon00, line 56: I also changed the ports here to follow #1970.

specification/protocol/exporter.md
```diff
@@ -28,24 +28,65 @@ Supported values for `OTEL_EXPORTER_OTLP_*COMPRESSION` options:
...
 
+```bash
+export OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=http://collector:4318
 export OTEL_EXPORTER_OTLP_METRICS_ENDPOINT=https://collector.example.com/v1/metrics
 ```
 
+This will send traces directly to the root path `http://collector:4318/`
+(`/v1/traces` is only automatically added when using the non-signal-specific
+environment variable) and metrics
+to `https://collector.example.com/v1/metrics`.
```
Oberon00, line 74: So these are sent to `https://collector.example.com:443/v1/metrics` then, since that is the default HTTPs port? Or should we add wording to use port 4318 if the URL contains no port? Having the https scheme would suggest 443.

tigrannajaryan: I was thinking about this yesterday night.
When the port is missing do we use the default port as defined by "https" schema (443) or do we use the default port defined by OTLP spec? I don't know what's the right answer.
I think we should choose the least surprising behavior (unless there is some HTTPS RFC which tells us what we have to do). That said, I have no idea what's the least suprising behavior :-)

Oberon00: First, I think this is out of scope for the PR, even though I brought it up myself. Second, I think both behaviors could be surprising. Third, I just checked and at Dynatrace we tell you to configure an URL without port and then listen only on 443 (https://www.dynatrace.com/support/help/how-to-use-dynatrace/transactions-and-services/purepath-distributed-traces/opentelemetry-ingest/#activate). Since nobody has complained so far, most SDKs seem to default to the scheme's default port, and in doubt I would say that is fine.

fbogsany (approved):

carlosalberto:
@iNikem Any further comment? Or are you ok with the current PR?

specification/protocol/exporter.md
anuraaga: Looking forward to a follow up issue to clarify the port to add a vote :-) Many systems have defined a port for HTTP APIs (Elasticsearch, Zipkin, many more) and I have never seen a client library for them default to anything other than 80/443 for portless URLs, anything else would be quite unexpected given how people use URLs.

anuraaga (approved):

iNikem (approved):

specification/protocol/exporter.md
tigrannajaryan: @Oberon00 can you please add a follow up for that. I believe this PR can be merged.

Oberon00: Created https://github.com/open-telemetry/opentelemetry-specification/pull/1985

PR #1978, opened 2021-09-29
Introduce new semantic conventions for CloudEvents

### LINKED ISSUE #1949: Add semantic conventions for CloudEvents

joaopgrassi:
Closes #1951, Resolves #1949

## Changes

This PR is a follow-up of PR [#1951](https://github.com/open-telemetry/opentelemetry-specification/pull/1951). 

In that PR, the CloudEvents semantic conventions were added together with the messaging ones. After the messaging WG meeting on 23.09.2021 ([meeting notes](https://docs.google.com/document/d/1dWHhyXnfVife-cQ2DW5-d5Ldp1Lq8Rre2UsHpyo8cEE/edit#heading=h.tshzguzdor6k)), it was agreed that the attributes introduced there make sense, but they should be outside of the messaging semantic conventions, to avoid confusion of CloudEvents being a messaging system.

This spec adds guidance on how/when to create Spans for CloudEvents, how context should be propagated and lists attributes that can be added to such Spans.


lmolkova:
Since CloudEvents spec defines [distributed tracing extension](https://github.com/cloudevents/spec/blob/v1.0.1/extensions/distributed-tracing.md), OTel conventions should cover how it's populated.

And then it would still collide with WIP messaging spec.

I suggest we either figure it out first or explicitly mention in this spec that instrumentation SHOULD NOT populate cloud event distributed tracing context until we know how it should work.

joaopgrassi:
@lmolkova I'm adding instrumentation to a project that uses the CloudEvents Go SDK and in order to get "proper" traces, I inject the tracecontext into the event and I used the distributing tracing extension for that: https://github.com/dynatrace-oss-contrib/sdk-go/blob/add-otel-obsservice/observability/opentelemetry/v2/client/cloudevents_carrier.go. 

I understand your point, but until a proper solution is in place, what should users do? Is my solution there completely discouraged?

lmolkova:
> @lmolkova I'm adding instrumentation to a project that uses the CloudEvents Go SDK and in order to get "proper" traces, I inject the tracecontext into the event and I used the distributing tracing extension for that: https://github.com/dynatrace-oss-contrib/sdk-go/blob/add-otel-obsservice/observability/opentelemetry/v2/client/cloudevents_carrier.go.
> 
> I understand your point, but until a proper solution is in place, what should users do? Is my solution there completely discouraged?

I work on the SDKs which do the same, so I feel your pain. I think our SDKs should keep doing what they are doing if it helps users, but it's not ready to be an official spec - we should be ready for how this context is propagated to change (or maybe a priority of this context to change when cloud event is transferred over something that supports context propagation too, e.g. AMQP). 

Maybe there is some middle ground where we say that you MAY propagate context and here's how, but the details are still in the works and may change?

joaopgrassi:
>Maybe there is some middle ground where we say that you MAY propagate context and here's how, but the details are still in the works and may change?

Yes that seems reasonable. The "POC" I did with CloudEvents in Go injecting/extracting using the distributed tracing extension worked well I think, but definitely having something "built-in" is the best in the long run. 

From my point maybe it's not so bad that things could change later. The PR I sent there for CloudEvents offers the `inject/extract` but you don't need to use it, if you are sending the events via HTTP and all is auto-instrumented then things will work. But for my case, it was necessary and I need it "today". So, until we have a clear path going forward, having a middle ground solution like you said makes sense. Me as a user can benefit from OTel today. If later I just update the SDK and things work as well, even better.

pyohannes:
> Since CloudEvents spec defines distributed tracing extension, OTel conventions should cover how it's populated.

I submitted a separate issue for this: https://github.com/open-telemetry/opentelemetry-specification/issues/1979

We will discuss this when we discuss context propagation in relation to the messaging semantic conventions. At this point we can't define how we populate it, as we don't know how to use it (and whether it will be of use at all).

In this way the context propagation question shouldn't block this PR, but it can be amended later if we come to a conclusion on this.

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -0,0 +1,47 @@
+# CloudEvents
+
+This document defines how to describe sending and receiving [CloudEvents](https://cloudevents.io/) with spans.
+
+The transport involved in sending/receiving CloudEvents is represented by a separate span, depending on the underlying protocol used.
```
pyohannes, line 5: What does "depending on the underlying protocol used" mean? Does it mean some underlying protocols require a separate span to be created, others do not?

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -0,0 +1,47 @@
+# CloudEvents
+
+This document defines how to describe sending and receiving [CloudEvents](https://cloudevents.io/) with spans.
```
pyohannes, line 3: I wonder if we could restrict these conventions to just defining attributes, and leaving it open how they are added to spans?

I'm not sure if having separate spans for CloudEvents makes sense in all cases.

Oberon00: I also think that in some cases you'd want to add the span attributes to the span of the operation that is described by the event (e.g. for `com.github.pull_request.opened` you would add the attributes to the span that traces opening the PR).

Oberon00:
@lmolkova 

> And then it would still collide with WIP messaging spec.
>
> I suggest we either figure it out first or explicitly mention in this spec that instrumentation SHOULD NOT populate cloud event distributed tracing context until we know how it should work.

I think @joaopgrassi took part in a messaging SIG meeting and the outcome was that CloudEvents are not really messaging. I would agree to that.

CloudEvents IMHO might be sent over a messaging system, but they might be sent over HTTP or stored in a database or be only used in-process as well. So IMHO, CloudEvents are a concept in their own right, not tied to any other semantic convention.

specification/trace/semantic_conventions/cloudevents.md
joaopgrassi: I wanted to basically say what you mentioned below: 

> I wonder if we could restrict these conventions to just defining attributes, and leaving it open how they are added to spans?

What I wanted to say is that you can use the attributes here to create spans if you want but all the other details of propagation and probably separate spans that need to be created when actually sending/receiving the event is a different topic. I'm having a hard time to produce words to explain this 😅.

An actual example:

```go
event := cloudevents.NewEvent()

// create a span for this operation, and another one maybe be created, depending if the transport is instrumented
ctx, span := o.tracer.Start(ctx, 
		"CloudEvents " + e.Context.GetType(),
		trace.WithSpanKind(trace.SpanKindProducer),
		trace.WithAttributes(...)) // with the attributes from OTel semconv
		
client, _ := client.New(t, copt...)
// Another span might be created here, depending if the transport being used is instrumented
client.Send(ctx, event) 
span.End()
```

specification/trace/semantic_conventions/cloudevents.md
joaopgrassi: Yeah I agree and it was what I tried to explain in the opening statement, but it's not good. Would you have an idea how we can communicate this more clearly?

lmolkova: Since we are separating creation and publishing in the new messaging SDKs, CloudEvent span in this case is creation, and HTTP POST  is publishing. Cloud events spans seem to be essential at least for this SDK and attributes should not be stamped on any other spans - there is no technical solution that would even allow that except hacky manual instrumentation colliding with an automatic one. 

They would be duplicates with messaging SDKs that use cloud events (have 'create' span too and propagate context differently) - but that's what we have to solve in messaging spec.
I think if users turn cloud events instrumentation on (or don't turn it off), they should have cloud events spans regardless of messaging layer.

Oberon00: > CloudEvent span in this case is creation

I disagree. I was only there in the first SIG meeting, but there "creation" was defined as the logical send operation, so it would include retries, etc. but not the creation of the data from which the message is formed.

Oberon00: With such a broad definition of "creation", all business logic that happens to also notify something via message about the result would be traced with a "message creation" span.

lmolkova: > "creation" was defined as the logical send operation

I don't think so, Creation is a separate step according to the [draft](https://github.com/open-telemetry/oteps/blob/922295985f2f1c0e440d4fe7c91f34f43e5e87e3/text/trace/0157-messaging-semantic-conventions.md)

Since you can send multiple cloud events over whatever protocol in a batch and CloudEvent has an individual context, you have to create a span for it.

lmolkova: and here's the thread that clarifies it https://github.com/open-telemetry/oteps/pull/173#discussion_r701762856

lmolkova:
To create a spec that fits other SDKs that do CloudEvents, let me share a few examples here

https://docs.microsoft.com/en-us/java/api/overview/azure/messaging-eventgrid-readme?view=azure-java-stable#sending-cloudevent-to-a-topic-that-accepts-cloudevent-schema

```java
List<CloudEvent> events = new ArrayList<>();
User user = new User("John", "James");
events.add(new CloudEvent("https://source.example.com", "Com.Example.ExampleEventType",
    BinaryData.fromObject(user), CloudEventDataFormat.JSON, "application/json"));
cloudEventClient.sendEvents(events);
```

I can't really spec on behalf of AWS, but AWS EventBridge does something similar (AFAIK using CloudEvents at least something similar that has a bridge to them) https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/examples-cloudwatch-send-events.html

```java
PutEventsRequestEntry request_entry = new PutEventsRequestEntry()
    .withDetail(EVENT_DETAILS)
    .withDetailType("sampleSubmitted")
    .withResources(resource_arn)
    .withSource("aws-sdk-java-cloudwatch-example");

PutEventsRequest request = new PutEventsRequest()
    .withEntries(request_entry);

PutEventsResult response = cwe.putEvents(request);
```

So
- creation is separate from publishing,
- batching is common
- send span can be related to message but does not have to (I can buffer up events and send later).

I'll leave a few comments on the spec to make it work for Azure EventGrid with CloudEvents

specification/trace/semantic_conventions/cloudevents.md
lmolkova: > sending and receiving

I suggest changing it to creation and consumption (sending and receiving are independent at least for other SDKs)

specification/trace/semantic_conventions/cloudevents.md
lmolkova: > The transport involved in sending/receiving CloudEvents

I suggest this spec does not cover this subject at all and it's up to messaging spec to define

lmolkova (commented):
I would still add a remark on context propagation, saying that it's being defined and you MAY use cloud events extension, but if you send cloud events over some messaging system covered by OTel messaging conventions, you SHOULD follow messaging conventions instead.

specification/trace/semantic_conventions/cloudevents.md
joaopgrassi: Sounds good, but I'm not sure how express this here in this spec. Do you have a suggestion?

lmolkova: I'd suggest something along the lines of "This specification covers creation and consumption of cloud events only and does not cover transport aspects of publishing and receiving cloud events.  Please refer to the messaging semantic conventions that will cover creation, consumption, and transport. 
The integration between CloudEvents and messaging libraries tracing instrumentation is not defined yet. Instrumentations should follow either CloudEvents convention or Messaging convention, but not both"

specification/trace/semantic_conventions/cloudevents.md
joaopgrassi: I read the thread you linked, I think it makes sense. But there it's said:

> "Create" (as well as "Process") are agnostic of the underlying protocol (AMQP, MQTT, HTTP), whereas all the other stages are not.

So probably we should use here also create and process to be consistent.

specification/trace/semantic_conventions/cloudevents.md
joaopgrassi: Sorry I read after your comment below about using "create", but I think consumption should be process? As it's in the thread in the OTEP you linked. Thanks for the suggestion I'll try to work something up!

specification/trace/semantic_conventions/cloudevents.md
lmolkova: sure, I agree to use terminology that's used in the OTEP and 'process' seems to be more precise.

pyohannes: > Since we are separating creation and publishing in the new messaging spec, CloudEvent span in this case is creation, and HTTP POST is publishing.

We didn't agree on any span structure yet, but we decided to treat creation and publishing as different conceptual stages.

I agree with @lmolkova we should have a clear guidance of when and whether to create dedicated CloudEvents span, but I think currently we don't have currently alignment on that.

I propose we just specify the attributes for now, maybe with a similar disclaimer than [for general span attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#general-attributes):

> The attributes described in this section are not specific to a particular operation but rather generic. They may be used in any Span they apply to.

We might want to add a hint that this will be specified in more detail in the future  (when we work out the conventions for messaging).

lmolkova: > I propose we just specify the attributes for now, maybe with a similar disclaimer than for general span attributes:

With spec that intentionally avoids specifying which spans should be created, I can't take it as a library owner and implement it. Backends can't take it and build experiences on it. General attributes beauty that they can go an almost any span, these can only go on something related to cloud events.

This spec (among others) is experimental and if messaging spec would contradict it, we can come back and change it, but leaving it intentionally vague does not seem to help.

Oberon00: >> @Oberon00:  I was only there in the first SIG meeting, but there "creation" was defined as the logical send operation, so it would include retries, etc. but not the creation of the data from which the message is formed.
>
> @lmolkova: I don't think so, Creation is a separate step according to the draft
>
> and here's the thread that clarifies it https://github.com/open-telemetry/oteps/pull/173#discussion_r701762856

But in this thread I said exactly that at https://github.com/open-telemetry/oteps/pull/173#discussion_r702089063:

>> @Oberon00: From the WG meeting it sounded like Create is actually supposed to mean "logical publish operation" (potentially with retries) while Publish is the "physical publish operation".
>
> @pyohannes: That's my understanding, seems we have to make this clearer in the document.

joaopgrassi: Hey all. I read the thread in the OTEP and although it seems we reached an agreement that it's helpful to have the create/publish stages separate, we still don't have a clear way on how we want to do it, as we can see here in this thread.

With that in mind, I took a few suggestions from @lmolkova and @pyohannes and I adapted the document to explicit say this are "generic" attributes and that further guidance will be added once we figure out the details on how CloudEvents will relate with messaging. Let me know your thoughts on this update.

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -1,38 +1,26 @@
...
-- [Attributes](#attributes)
-
-<!-- tocstop -->
+- How and when Spans are created for CloudEvents
 
-## Span name
+- Transport aspects of sending and receiving CloudEvents
 
-The span name should be a combination of the `CloudEvents` prefix, followed by the event type as in the following format:
+The above points are being worked on in the messaging semantic conventions.
```
Oberon00, line 14: I would not even mention messaging. I'm still not sure if there is much overlap.

semantic_conventions/trace/cloudevents.yaml
```diff
@@ -0,0 +1,33 @@
+groups:
+  - id: cloudevents
```
Oberon00, line 2:
```suggestion
  - id: cloudevent
```

etc

specification/trace/semantic_conventions/cloudevents.md
joaopgrassi: From the discussion above, I got that most of the details will come from the messaging WG/conventions, no? For example, if a Span should be created when creating CloudEvents. So I thought of mentioned it as @pyohannes suggested:

> We might want to add a hint that this will be specified in more detail in the future (when we work out the conventions for messaging).

semantic_conventions/trace/cloudevents.yaml
joaopgrassi: I think we should stick with the trademark, no? https://www.linuxfoundation.org/trademarks

Oberon00: Why? We are not describing the abstract "CloudEvents specification" or "CloudEvents technology" here but a particular CloudEvent. The spec uses "CloudEvent" singular form several times, e.g. see https://github.com/cloudevents/spec/blob/master/spec.md#overview

> Event Formats specify how to serialize **a CloudEvent** with certain encoding formats.

Oberon00: Although we also use "messaging.message_id" in the messaging conventions...

Oberon00: Looking through the attributes you already use `event_` as a prefix. I think that is fine then. I would just suggest to rename `subject` to `event_subject` for consistency.

specification/trace/semantic_conventions/cloudevents.md
Oberon00: > From the discussion above, I got that most of the details will come from the messaging WG/conventions, no? For example, if a Span should be created when creating CloudEvents.

Why do we see CloudEvents as a messaging system? Does messaging SIG also consider general event notification systems in scope? I guess it makes sense then, fine with me then.

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -0,0 +1,35 @@
...
+The attributes described in this document are not specific to a particular operation but rather generic. They may be used in any Span they apply to.
+
+At this time, this specification does not cover:
+
+- How and when Spans are created for CloudEvents
+
+- Transport aspects of sending and receiving CloudEvents
+
+The above points are being worked on in the messaging semantic conventions.
+This document will be updated with further guidance once the messaging semantic conventions reaches a stable state.
```
Oberon00, line 15: Do we intend to wait until after the messaging semantic conventions reach "stable"? No semantic convention has ever reached that state yet (except service.name).

specification/trace/semantic_conventions/cloudevents.md
lmolkova: messaging systems carry a payload that historically is a blob or arbitrary data, also they all have different metadata. CloudEvents are intended to replace the arbitrary blob and some metadata, they are designed to be used with messaging systems https://github.com/cloudevents/spec#cloudevents-documents

The intersection between this spec and messaging spec is the trace context. They'd both define when the context is created, how it's injected and extracted: producer, consumer spans and links

specification/trace/semantic_conventions/cloudevents.md
joaopgrassi: Yeah probably not, but I couldn't find a better way to explain it. =/

I wanted to express that some concepts are "overlapping"  and that more details with be added when we figure out things in the messaging semantic conventions. (e.g.** how to propagate context? Should users use CloudEvents extensions, or should it be taken care of by the transport instrumentation, see Liudmila's concerns) 

I had this version in a draft, but I'm not sure it's better:
```
This document will be updated with further guidance once the messaging working group defines a clear guidance on how CloudEvents related with messaging systems.
```

specification/trace/semantic_conventions/cloudevents.md
pyohannes: > Why do we see CloudEvents as a messaging system?

We don't. That's why we asked to not define it in the messaging semantic conventions (which was the initial PR). As @lmolkova pointed out though, it's heavily used in the context of messaging systems.

> I would not even mention messaging. I'm still not sure if there is much overlap.

I'm fine with that too. It might be better to not have disguised TODO statements in the spec. Work that is left to be done can be captured in issues. In this context, one could just add a note to https://github.com/open-telemetry/opentelemetry-specification/issues/1979.

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -0,0 +1,35 @@
+# CloudEvents
+
+**Status**: [Experimental](../../document-status.md)
+
+This specification defines attributes for [CloudEvents](https://cloudevents.io/).
+The attributes described in this document are not specific to a particular operation but rather generic. They may be used in any Span they apply to.
+
+At this time, this specification does not cover:
+
+- How and when Spans are created for CloudEvents
```
lmolkova, line 10: and still, I don't understand the benefit of defining a spec that's not possible to implement. Who does it help?

specification/trace/semantic_conventions/cloudevents.md
pyohannes: I'm fine leaving this statement out. So let's not elaborate on what's not currently covered and when it will (prospectively) be covered. Again, maybe this is better captured in an issue.

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -0,0 +1,35 @@
...
+- [Attributes](#attributes)
+
+<!-- tocstop -->
+
+## Attributes
+
+<!-- semconv cloudevents -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `cloudevents.event_id` | string | The [event_id](https://github.com/cloudevents/spec/blob/master/spec.md#id) identifies the event within the scope of a producer. | `123e4567-e89b-12d3-a456-426614174000`; `producer-1` | No |
```
pyohannes, line 30: It's not clear to me what "identifies the event within the scope of a producer" means. Aren't the id and the source supposed to help uniquely identify an event also on consumer side?

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -0,0 +1,35 @@
...
+| `cloudevents.source` | string | The [source](https://github.com/cloudevents/spec/blob/master/spec.md#source-1) identifies the context in which an event happened. | `https://github.com/cloudevents`; `/cloudevents/spec/pull/123`; `my-service` | No |
+| `cloudevents.spec_version` | string | The [version of the CloudEvents specification](https://github.com/cloudevents/spec/blob/master/spec.md#specversion) which the event uses. | `1.0` | No |
+| `cloudevents.event_type` | string | The [event_type](https://github.com/cloudevents/spec/blob/master/spec.md#type) contains a value describing the type of event related to the originating occurrence. | `com.github.pull_request.opened`; `com.example.object.deleted.v2` | No |
+| `cloudevents.subject` | string | The [subject](https://github.com/cloudevents/spec/blob/master/spec.md#subject) of the event in the context of the event producer (identified by source). | `mynewfile.jpg` | No |
```
pyohannes, line 34: Just a question about consistent naming: as far as I understand, all those attributes pertain to a single event. Only two of them (`event_id` and `event_type`) have an `event_` prefix, others (`source` or `subject`) don't.

Do you think it makes sense to make this consistent?

specification/trace/semantic_conventions/cloudevents.md
joaopgrassi: From the CloudEvents spec, the `id` field lists some constraints, one of them being:

> MUST be unique within the scope of the producer

But you are right, it is a bit confusing. I've changed to only say it uniquely identifies the event, and added a `note` about the `source` + `id` combination as well.

specification/trace/semantic_conventions/cloudevents.md
joaopgrassi: I think it makes sense to add the `event_` prefix to all of them. I initially copied them from the CloudEvents Go SDK and only added the prefix to `id` and `type` because they were "too general" and might be confused with other conventions. Changed it now.

specification/trace/semantic_conventions/cloudevents.md
joaopgrassi: Alright - I removed both statements. Does it also makes sense then to remove this portion:

> At this time, this specification does not cover:

specification/trace/semantic_conventions/cloudevents.md
joaopgrassi: As per discussion above, I removed it now. Now it simply has the attributes. I understand and agree with your points, but it seems we are a bit far to reaching an conclusion. It seems we will only be "unblocked" once we figure out the propagation portion. I think what @pyohannes said of capturing these missing things in an issue makes more sense. What do you think?

semantic_conventions/trace/cloudevents.yaml
joaopgrassi: I added the `event_` prefix to all of them now for consistency.

specification/trace/semantic_conventions/cloudevents.md
joaopgrassi: I removed those statements and what's currently not covered.

semantic_conventions/trace/cloudevents.yaml
```diff
@@ -0,0 +1,37 @@
...
+          Producers MUST ensure that event_source + event_id is unique for each distinct event. 
+          If a duplicate event is re-sent (e.g. due to a network error) it MAY have the same id. 
+          Consumers MAY assume that Events with identical event_source and event_id are duplicates.
+        examples: ['123e4567-e89b-12d3-a456-426614174000', '0001']
+      - id: event_source
+        type: string
+        brief: >
+          The [source](https://github.com/cloudevents/spec/blob/master/spec.md#source-1) identifies the context in which an event happened.
+        examples: ['https://github.com/cloudevents', '/cloudevents/spec/pull/123', 'my-service' ]
+      - id: event_spec_version
```
Oberon00, line 23: This is the one key where I would rather not add `event_` in front, although maybe it makes sense even here. But if we use `event` everywhere, maybe using `cloudevent` (singular) as prefix *instead* would be nicer. Not sure...

Oberon00 (approved):

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -0,0 +1,28 @@
+# CloudEvents
+
+**Status**: [Experimental](../../document-status.md)
+
+This specification defines attributes for [CloudEvents](https://cloudevents.io/).
+The attributes described in this document are not specific to a particular operation but rather generic. They may be used in any Span they apply to.
```
Oberon00, line 6: What about span events? Rather not, right?

joaopgrassi: Yeah I'd say not at this point. Maybe it will make sense later on though.

semantic_conventions/trace/cloudevents.yaml
joaopgrassi: Yes I was also unsure, but I added anyway for the sake of consistency.

> But if we use event everywhere, maybe using cloudevent (singular) as prefix instead would be nicer. Not sure...

I personally don't have a preference, but I tend to want to stick with the trademark for the prefix. When actually using the constants in code this is how it will look like in Go at least

```go
semconv.CloudEventsEventIdKey
semconv.CloudEventsEventSourceKey
```

I think it isn't so bad. There's `Event*` twice but well.

joaopgrassi: Thinking now, when using this to generate the constants. isn't `cloudevents` prefix going to be `Cloudevents`? As in: `semconv.CloudeventsEventIdKey`? That is not so great actually... :(

joaopgrassi: OK I tested with the https://github.com/open-telemetry/opentelemetry-go-build-tools/tree/main/semconvgen and this generates as expected `CloudeventsEventIDKey`. I assume in languages like Java this will be `CLOUDEVENTS_EVENT_ID`. I guess we'll have to live with `Cloudevents` then 🤔

Oberon00: This topic seems to be problematic enough in Go that there is a complete post-processor for the semantic  conventions at https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/2ceabab23375c1e33f6c6e82a1baaefa8cdbc7f3/semconvgen/generator.go#L268 (the Go maintainers opted to not use the capabilities of the existing semantic conventions generator because they have no Jinja2/Python expertise in their team and instead write a post-processor in Go)

joaopgrassi: Yeah I see. It changes highly between languages but I guess this is out of scope for this anyway. Will mark this as solved.

specification/trace/semantic_conventions/cloudevents.md
pyohannes: I'll reserve a slot in this Thursday's messaging workgroup so we can align on this (whether to add instrumentation guidance besides just attributes).

I had a quick chat with @lmolkova, she has a strong interest in having this instrumentation guidance worked out, and she also offered to contribute to this work.

tigrannajaryan:
> Does that help clarify things a bit?

Yes, thank you.

tigrannajaryan:
@open-telemetry/specs-approvers PTAL.

pyohannes:
There is ongoing discussion about this PR in the messaging workgroup.

@lmolkova showed a prototype for using CloudEvents in a messaging system context, in this case using links is preferable to using parent/child relationships. @joaopgrassi will work on a prototype for the HTTP scenario using links.

We will that in the messaging workgroup next week, then this PR will be updated.

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -37,14 +37,13 @@ In case a CloudEvent is passed to the instrumented library with the Distributed
...
-
-If multiple events are processed together, for each event being processed, if the event has the Distributed Tracing Extension populated, instrumentation MUST add a link to the trace context from the extension on the processing span.
+When instrumented library supports processing of a single CloudEvent, instrumentation SHOULD create a new span to trace it. Instrumentation SHOULD set the remote trace context from the Distributed Tracing Extension as a link on the processing span.
 
 **Span name:** `CloudEvents Process <event_type>`
 **Span kind:** CONSUMER
 
+Note: CloudEvents processing does not follow a common pattern. Some "CloudEvents-enabled" libraries expose handlers for processing (e.g.  [CloudEvents Go SDK](https://github.com/cloudevents/sdk-go#receive-your-first-cloudevent)) and can auto-instrument the processing calls. Alternatively, a library might offer a handler, but the user chooses not to use it, or a library does not offer one at all. In such cases, the instrumentation responsibility falls under the application owner.
```
pyohannes, line 45:
```suggestion
Note: CloudEvents processing does not follow a common pattern. Some "CloudEvents-enabled" libraries expose handlers for processing (e.g.  [CloudEvents Go SDK](https://github.com/cloudevents/sdk-go#receive-your-first-cloudevent)) and can auto-instrument the processing calls. Alternatively, a library might offer a handler, but the user chooses not to use it, or a library does not offer one at all. In such cases, the instrumentation responsibility lies with the application owner.
```

I guess "falls on" would work too.

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -0,0 +1,61 @@
...
+  * [Processing](#processing)
+- [Attributes](#attributes)
+
+<!-- tocstop -->
+
+## Spans
+
+### Creation
+
+Instrumentation SHOULD create a new span and populate the [Distributed Tracing Extension](https://github.com/cloudevents/spec/blob/v1.0.1/extensions/distributed-tracing.md) on the event. This applies when:
```
pyohannes, line 28: From discussions in the messaging workgroup: we should add a link to this document in the CloudEvents spec, to make clear where the DT extension is used.

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -0,0 +1,61 @@
...
+**Span kind:** PRODUCER
+
+### Processing
+
+When instrumented library supports processing of a single CloudEvent, instrumentation SHOULD create a new span to trace it. Instrumentation SHOULD set the remote trace context from the Distributed Tracing Extension as a link on the processing span.
+
+**Span name:** `CloudEvents Process <event_type>`
+**Span kind:** CONSUMER
+
+Note: CloudEvents processing does not follow a common pattern. Some "CloudEvents-enabled" libraries expose handlers for processing (e.g.  [CloudEvents Go SDK](https://github.com/cloudevents/sdk-go#receive-your-first-cloudevent)) and can auto-instrument the processing calls. Alternatively, a library might offer a handler, but the user chooses not to use it, or a library does not offer one at all. In such cases, the instrumentation responsibility lies with the application owner.
```
pyohannes, line 45: From discussions in the messaging workgroup: this section could be more authoritative and tell SDKs what to do. E. g.: SDKs SHOULD preferably auto-instrument or alternatively provide a handler.

pyohannes (commented):
I added some comments from discussions in today's messaging workgroup.

I think this is good otherwise, however, let's try to achieve some consensus in next weeks workgroup meeting.

specification/trace/semantic_conventions/cloudevents.md
joaopgrassi: I changed the paragraph a bit, we can go over it during this week's messaging wg meeting.

specification/trace/semantic_conventions/cloudevents.md
joaopgrassi: I have a PR in the pipeline. I'll open it later today or, at the latest tomorrow (against the CloudEvents spec).

joaopgrassi: PR is there: https://github.com/cloudevents/spec/pull/912

joaopgrassi:
@clemensv @pyohannes and others: Pinging since this week the meeting is cancelled. I addressed the two points from our discussion from the other week - There's a PR open against the CloudEvents spec https://github.com/cloudevents/spec/pull/912 and I changed the paragraph in this spec and tried to be more clear about this spec "expects" from CloudEvents libraries. Let me know your thoughts :)

pyohannes (approved):
Thanks @joaopgrassi for working on this.

It's not yet clear how those semantic conventions interoperate with the semantic conventions for messaging, which are currently being designed. Once messaging semantic conventions are stable, we will need to discuss and revisit the CloudEvents semantic conventions, in order to clarify how CloudEvents and messaging semantic conventions relate to each other.

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -0,0 +1,68 @@
+# CloudEvents
+
+**Status**: [Experimental](../../document-status.md)
+
+This specification defines semantic conventions for [CloudEvents](https://cloudevents.io/). It covers creation, processing and context propagation between producer and consumer. It does not cover transport aspects of publishing and receiving cloud events.
+
+## Overview
+
+To be individually traceable, each CloudEvent has to have its own trace context. The trace context is populated in the event using the [Distributed Tracing Extension](https://github.com/cloudevents/spec/blob/v1.0.1/extensions/distributed-tracing.md).
```
arminru, line 9:
```suggestion
To be individually traceable, each CloudEvent has to have its own trace context.
The trace context is populated in the event using the [CloudEvents Distributed Tracing Extension](https://github.com/cloudevents/spec/blob/v1.0.1/extensions/distributed-tracing.md).
```
To make clear this is provided by CloudEvents, not OTel.

specification/trace/semantic_conventions/cloudevents.md
arminru:
```suggestion
Instrumentation SHOULD create a new span and populate the [CloudEvents Distributed Tracing Extension](https://github.com/cloudevents/spec/blob/v1.0.1/extensions/distributed-tracing.md) on the event. This applies when:
```

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -0,0 +1,68 @@
...
+
+<!-- tocstop -->
+
+## Spans
+
+### Creation
+
+Instrumentation SHOULD create a new span and populate the [Distributed Tracing Extension](https://github.com/cloudevents/spec/blob/v1.0.1/extensions/distributed-tracing.md) on the event. This applies when:
+
+- A CloudEvent is created by the instrumented library. It may be impossible or impractical to create the Span during event creation (e.g. inside the constructor or in a factory method) - instrumentation MAY create the Span later, when passing the event to the transport layer.
```
arminru, line 31:
```suggestion
- A CloudEvent is created by the instrumented library. It may be impossible or impractical to create the Span during event creation (e.g. inside the constructor or in a factory method), so instrumentation MAY create the Span later, when passing the event to the transport layer.
```

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -0,0 +1,68 @@
...
+- A CloudEvent is created outside of the instrumented library (e.g. directly constructed by the application owner, without calling a constructor or factory method), and passed without the Distributed Tracing Extension populated.
+
+In case a CloudEvent is passed to the instrumented library with the Distributed Tracing Extension populated, instrumentation MUST NOT create a span and MUST NOT modify the Distributed Tracing Extension on the event.
```
arminru, line 34:
```suggestion
In case a CloudEvent is passed to the instrumented library with the Distributed Tracing Extension already populated, instrumentation MUST NOT create a span and MUST NOT modify the Distributed Tracing Extension on the event.
```

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -0,0 +1,68 @@
...
+
+**Span name:** `CloudEvents Create <event_type>`
+**Span kind:** PRODUCER
+
+### Processing
+
+When instrumented library supports processing of a single CloudEvent, instrumentation SHOULD create a new span to trace it. Instrumentation SHOULD set the remote trace context from the Distributed Tracing Extension as a link on the processing span.
```
arminru, line 41:
```suggestion
When an instrumented library supports processing of a single CloudEvent, instrumentation SHOULD create a new span to trace it.
Instrumentation SHOULD set the remote trace context from the Distributed Tracing Extension as a link on the processing span.
```

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -0,0 +1,68 @@
...
+**Span name:** `CloudEvents Process <event_type>`
+**Span kind:** CONSUMER
+
+## Instrumentation
+
+CloudEvents libraries SHOULD have built-in mechanisms to make the aforementioned
+Span operations (creation, extraction, and injection of context into events)
+transparent for application owners. When auto-instrumentation
+is not feasible, CloudEvents libraries SHOULD offer such mechanisms in a form that
+application owners can "plug in" to apply the Span operations above by themselves.
```
arminru, line 52: The part "CloudEvents libraries SHOULD" should rather go into https://github.com/cloudevents/spec than the OTel spec I think.

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -0,0 +1,68 @@
...
+|---|---|---|---|---|
+| `cloudevents.event_id` | string | The [event_id](https://github.com/cloudevents/spec/blob/master/spec.md#id) uniquely identifies the event. [1] | `123e4567-e89b-12d3-a456-426614174000`; `0001` | No |
+| `cloudevents.event_source` | string | The [source](https://github.com/cloudevents/spec/blob/master/spec.md#source-1) identifies the context in which an event happened. | `https://github.com/cloudevents`; `/cloudevents/spec/pull/123`; `my-service` | No |
+| `cloudevents.event_spec_version` | string | The [version of the CloudEvents specification](https://github.com/cloudevents/spec/blob/master/spec.md#specversion) which the event uses. | `1.0` | No |
+| `cloudevents.event_type` | string | The [event_type](https://github.com/cloudevents/spec/blob/master/spec.md#type) contains a value describing the type of event related to the originating occurrence. | `com.github.pull_request.opened`; `com.example.object.deleted.v2` | No |
+| `cloudevents.event_subject` | string | The [subject](https://github.com/cloudevents/spec/blob/master/spec.md#subject) of the event in the context of the event producer (identified by source). | `mynewfile.jpg` | No |
```
arminru, line 65: https://github.com/cloudevents/spec renamed master to main and restructured directories so the links here are broken.

joaopgrassi: Thanks! I fixed it and used the `v1.0.1` default branch now.

specification/trace/semantic_conventions/cloudevents.md
joaopgrassi: Makes sense. I also broke the long lines to make it easier to review.

specification/trace/semantic_conventions/cloudevents.md
joaopgrassi: I tend to agree as well. I think this should be a section in the  https://github.com/cloudevents/spec/blob/v1.0.1/extensions/distributed-tracing.md. Looping in @pyohannes, @lmolkova. What do you think? I vote to remote this from here as well. I spoke with the maitainer of the JS CloudEvents SDK and he also feel this is not the place to be. As a SDK authors people will look at the CloudEvents spec on what the SDK should do, and not here.

arminru (approved):
Looks good to me provided that the CloudEvents vs. OTel spec responsibility for the library requirements is sorted out.
As mentioned above otherwise this state should be good to merge and as it's experimental it will be revisited as the messaging semantic conventions mature further.

joaopgrassi:
@open-telemetry/specs-approvers sorry for the spam, but can someone re-open this? I pushed some commits 8 days ago, I thought this would count for removing the stale label but it didn't 🤔.

specification/trace/semantic_conventions/cloudevents.md
joaopgrassi: Done.

joaopgrassi:
Hi @tigrannajaryan. We discussed this PR during today's messaging working group meeting (meeting notes https://docs.google.com/document/d/1dWHhyXnfVife-cQ2DW5-d5Ldp1Lq8Rre2UsHpyo8cEE) and it's ready to be merged.

tigrannajaryan:
@open-telemetry/specs-approvers please review, especially if you are familiar with CloudEvents.

Question for eveyone: do we want `cloudevents` to be a top-level namespace it needs to be a subnamespace under some top-level namespace that groups other similar sets of conventions?

joaopgrassi:
Hi @tigrannajaryan, 

If I understood you right, you mean this being part of some other existing convention? 

I submitted an initial PR (#1951) where I had this under the messaging conventions (`semantic_conventions/trace/messaging.yaml`) but it was decided that this should be on its own file/area: https://github.com/open-telemetry/opentelemetry-specification/pull/1951#issuecomment-925974455.

tigrannajaryan:
> If I understood you right, you mean this being part of some other existing convention?

I mean part of a group of conventions for similar concepts. Not sure if that's the right approach, but typically putting something at the top-level requires that concept to be really unique and widely supported. Perhaps cloudevents qualify for that, I don't know if that is the case or no.

If we wanted to put it understand a different top-level namespace then attributes would become `foo.cloudevents.*` instead of `cloudevents.*`, where `foo` the concept to which we believe cloudevents belong.

tedsuo (approved):

joaopgrassi:
> If we wanted to put it understand a different top-level namespace then attributes would become foo.cloudevents.* instead of cloudevents.*, where foo the concept to which we believe cloudevents belong.

Right. In the initial PR, I had it under `messaging.cloudevents.*` because that was my naive assumption at the time. During the review and our meetings it was concluded that it didn't belong under the messaging namespace. Since CloudEvents are an abstraction around a concept and can be sent in different ways (via HTTP, messaging systems, etc) then we thought it deserved its own namespace. Maybe @pyohannes can add more.

pyohannes:
> If we wanted to put it understand a different top-level namespace then attributes would become `foo.cloudevents.*` instead of `cloudevents.*`, where foo the concept to which we believe cloudevents belong.

One problem here is, that cloud events are somewhat orthogonal to current top-level namespaces. They can be used via HTTP, in messaging scenarios, or in RPC scenarios. As cloud events aim to describe _event_ data in a common way, `event.cloudevents` might be a fit, but then _event_ is a very overloaded term.

If it's a requirement to put it under a namespace, `messaging` would probably be the best fit, although it suggests cloud events are only used in messaging scenarios (which is not the case).

I don't have a strong preference, but as those semantic conventions are currently experimental, I'd opt to keep it as a top-level namespace. If necessary, it can be moved to a suitable namespace later.

joaopgrassi:
Hi @tigrannajaryan, just pinging to see if you had a chance to see the replies.

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -0,0 +1,76 @@
...
+**Span name:** `CloudEvents Process <event_type>`
+
+**Span kind:** CONSUMER
+
+## Attributes
+
+The following attributes are applicable to creation and processing Spans.
+
+<!-- semconv cloudevents -->
+| Attribute  | Type | Description  | Examples  | Required |
```
tigrannajaryan, line 69: Are all of the attributes optional? Typically we have at least one required attribute in a set of attributes that describe a concept. The presence of such attribute makes it possible to detect that the span is about that particular concept.
If there is an attribute that you know will always be available for a cloudevent it would be useful to mark that attribute as Required.

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -0,0 +1,76 @@
+# CloudEvents
```
tigrannajaryan, line 1: Is it possible to add some context/background to this doc? A short paragraph describing what is a cloudevent and why and when (in what scenarios) one would want to record it in span attributes?
I don't know anything about CloudEvents and tried reading https://cloudevents.io/ and still can't figure out why it needs to be recorded on a span.

specification/trace/semantic_conventions/cloudevents.md
joaopgrassi: Thanks for pointing this out. There are several attributes in the CloudEvent specification that are required in the event. (https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#required-attributes). I added these here as required as well.

specification/trace/semantic_conventions/cloudevents.md
joaopgrassi: Hi @tigrannajaryan I pushed some changes where I tried to answer your questions. Let me know if this helps clarifies things.

joaopgrassi: I should also mention that this is still experimental and many things we came up in this PR (we, as the Messaging working group) are being used in stabilizing semantic conventions for messaging systems (https://github.com/open-telemetry/oteps/pull/192). 

As we progress with messaging systems semantic conventions, this document will change to reflect that. Specially in things like context propagation.

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -0,0 +1,103 @@
...
+  * [Processing](#processing)
+- [Attributes](#attributes)
+
+<!-- tocstop -->
+
+## Spans
+
+### Creation
+
+Instrumentation SHOULD create a new span and populate the
```
tigrannajaryan, line 59: Is a new span created for each cloudevent? What's the duration of the span?

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -0,0 +1,103 @@
...
+[CloudEvents Distributed Tracing Extension](https://github.com/cloudevents/spec/blob/v1.0.1/extensions/distributed-tracing.md)
```
tigrannajaryan, line 60: Do I understand correctly that the purpose of this is to capture the context of the execution so that when the cloudevent is propagated elsewhere this context is preserved and can be continued on? This is essentially the context propagation mechanism and the cloudevent is the medium through which the context propagates. Is this true?

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -0,0 +1,103 @@
...
+that happened as the result of an event. One can, for example, answer the
+following questions:
+
+- What components in a system reacted to an event
+- What further events were sent due to an incoming event
+- Which event caused the exception
+
+## Overview
+
+To be individually traceable, each CloudEvent has to have its own trace context.
```
tigrannajaryan, line 38: What does "individually traceable" mean?

specification/trace/semantic_conventions/cloudevents.md
joaopgrassi: Yes, you understood it correctly.

A CloudEvent can be sent directly from producer to consumer. In this manner the customary parent-child trace model works well and we don't need extra context propagation. But, CloudEvents also often go through many hops until it reaches a consumer, specially in distributed systems. This makes it hard to create a trace that "makes sense". This challenge is similar (if not the same) as the ones we face with messaging systems. 

The CloudEvents spec already offers as mentioned above an "extension" where a span context can be added to the event as a way of preserving the initial create context. This specification then takes uses of it and offers guidelines on how/when to populate it and how to consume it in order to continue the trace in a consistent way. 

To help exemplify, consider a scenario like this:

image

Today, we will get a trace like this (given there's some auto-instrumentation configured, e.g. http):
image

Essentially you have 2 traces that you can't correlate to each other. Now, here's a trace that implements the proposed conventions here:
image

The two "red" spans are the `Create` and `Process` this conventions introduces. This gives a chance of connecting producer and consumer(s), basically enabling users to have a "map" of everything that happened after an event was sent.
This model also creates a trace that is independent of any protocols.

To also help clarify, this way of using the event as a medium to propagate the initial create context is also similar to what we are discussing for messaging. In messaging we will also attach the initial create context to the message in some way (message metadata, etc).

specification/trace/semantic_conventions/cloudevents.md
joaopgrassi: The answer above goes into more details, but an event can go over multiple hops in a distributed system and one would want to trace every action/reaction that happened due to an event being sent. In order to do that, each event must have its own context, that "travels" together with it and other traces can link to it.

specification/trace/semantic_conventions/cloudevents.md
joaopgrassi: Yes, a new span should be created for each new event. There will be essentially no duration as this span is more a way of creating a "snapshot" that this happen so this context can travel along with the event and can be linked later by any involved component.

specification/trace/semantic_conventions/cloudevents.md
tigrannajaryan: This explanation and the diagrams are very useful. I think it is worth adding as a section to the conventions document. It makes the intent clear.

joaopgrassi: I'll try to move some of it to the document. I'm not sure how to handle images in here though.. I guess I can try reproducing it with some fancy ASCII art.

tigrannajaryan: Either ASCII art or add pngs. We do that elsewhere in specs already, e.g.: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/library-guidelines.md

joaopgrassi: @tigrannajaryan I added the diagrams + modified the sections a bit to include the explanation from this thread.

joaopgrassi:
Hi @pyohannes, @lmolkova, @arminru, I updated the PR a bit, according to this conversation I had with @tigrannajaryan (https://github.com/open-telemetry/opentelemetry-specification/pull/1978#discussion_r808532035). Could you please take another look? 

The changes are around explaining the context propagation portion and adding some example diagram/traces to clarify why this is necessary. Let me know your thoughts.

specification/trace/semantic_conventions/cloudevents.md
joaopgrassi: Resolving this as most things got answered in this thread https://github.com/open-telemetry/opentelemetry-specification/pull/1978#discussion_r808532035

tigrannajaryan (approved):

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -2,33 +2,152 @@
...
 
-Once the trace context is set on the event, it MUST not be modified.
+<!-- tocstop -->
 
-<!-- Re-generate TOC with `markdown-toc --no-first-h1 -i` -->
+## Definitions
 
-<!-- toc -->
+ From the
+ [CloudEvents specification:](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#overview)
```
arminru, line 22:
```suggestion
 [CloudEvents specification](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#overview):
```

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -2,33 +2,152 @@
...
+                                           |                                                                                   
+                                           v                                                                                   
+      +----------+                    +----------+                                                                             
+      | Consumer | <----------------- |  Queue   |                                                                             
+      +----------+                    +----------+ 
+```
+
+With the traditional parent-child trace model, the above scenario would produce
+two traces, completely independent from each other. It is not possible to
+correlate a producer with a consumer(s).
```
arminru, line 59:
```suggestion
two traces, completely independent from each other because the consumer
starts receiving (and thus has to specify a parent span) before it receives the event.
It is not possible to correlate a producer with a consumer(s) solely via a parent-child relationship.
```

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -2,33 +2,152 @@
...
+## Conventions
+
+To achieve the trace above, it is necessary to capture the context of
+the event creation so that when the CloudEvent reaches its destination(s), this
+context can be continued. Each CloudEvent acts then as the medium of this
+context propagation.
+
+This document relies on the CloudEvents specification, which defines this
+context propagation mechanism via the
+[CloudEvents Distributed Tracing Extension](https://github.com/cloudevents/spec/blob/v1.0.1/extensions/distributed-tracing.md).
```
arminru, line 141:
```suggestion
[CloudEvents Distributed Tracing Extension](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/extensions/distributed-tracing.md).
```
Since you reference 1.0.2 above as well (the files were reorganized in the meantime).

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -2,33 +2,152 @@
...
+The remainder of this section describes the semantic conventions for Spans
+required to achieve the proposed trace.
 
-### Creation
+### Spans
+
+#### Creation
 
 Instrumentation SHOULD create a new span and populate the
 [CloudEvents Distributed Tracing Extension](https://github.com/cloudevents/spec/blob/v1.0.1/extensions/distributed-tracing.md)
```
arminru, line 153:
```suggestion
[CloudEvents Distributed Tracing Extension](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/extensions/distributed-tracing.md)
```

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -47,9 +166,10 @@ Distributed Tracing Extension already populated, instrumentation MUST NOT create
 a span and MUST NOT modify the Distributed Tracing Extension on the event.
 
 **Span name:** `CloudEvents Create <event_type>`
+
 **Span kind:** PRODUCER
```
arminru, line 170: This renders just like the h4 for Creation and Processing. You could make the hierarchy clearer by making it italic instead of bold or by indenting it via bullet points, for example.

semantic_conventions/trace/cloudevents.yaml
```diff
@@ -8,25 +8,25 @@ groups:
     attributes:
       - id: event_id
         type: string
+        required: always
         brief: >
           The [event_id](https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#id) uniquely identifies the event.
```
arminru, line 13:
```suggestion
          The [event_id](https://github.com/cloudevents/spec/blob/v1.0.2/spec.md#id) uniquely identifies the event.
```
Would be the up-to-date version (+ a few more occurrences below).

specification/trace/semantic_conventions/cloudevents.md
```diff
@@ -58,28 +178,19 @@ Instrumentation SHOULD set the remote trace context from the
 Distributed Tracing Extension as a link on the processing span.
```
arminru, line 178: One might wonder which attributes to use. If the spec would say "with no attributes" it would prohibit adding them but you could say "with optional attributes" or "MAY add attributes" to make clear they are not specified.

arminru (approved):
Thanks for adding additional context and explanations. LGTM!

semantic_conventions/trace/cloudevents.yaml
joaopgrassi: Good find. Fixed it everywhere.

specification/trace/semantic_conventions/cloudevents.md
joaopgrassi: I will go with the bullet points approach, that renders better I think.

specification/trace/semantic_conventions/cloudevents.md
joaopgrassi: OK - I misunderstood and you were talking about attributes when adding the Link. I also addressed this on 714543a8963eea5f7cf7ebbe8efa29f52a3dae78

joaopgrassi:
Thanks all for the help with the reviews! 🎉

PR #1980, opened 2021-09-30
Clarify difference between `container.name` and `k8s.container.name`

dmitryax:
Add more details to container name attributes to clarify the difference.

Following from the thread in https://github.com/open-telemetry/opentelemetry-specification/pull/1945


semantic_conventions/resource/k8s.yaml
```diff
@@ -61,7 +61,9 @@ groups:
       - id: name
         type: string
         brief: >
-          The name of the Container in a Pod template.
+          The name of the Container from Pod specification, must be unique
+          within a Pod. Container runtime usually uses different name unique
+          within a Cluster (`container.name`).
```
Oberon00, line 66: The container runtime does not usually call that "cluster".

semantic_conventions/resource/container.yaml
```diff
@@ -7,7 +7,7 @@ groups:
       - id: name
         type: string
         brief: >
-          Container name.
+          Container name used by container runtime.
         examples: ['opentelemetry-autoconf']
```
Oberon00, line 11: Maybe add a typical docker-auto-generated name as example.

dmitryax: Do you mean names that docker generates if user doesn't provide anything? Those names are like gracious_fermi, quirky_kalam, nervous_bohr etc. Not sure if it will be better than existing example :)

k8s gives docker engine meaningful names that include k8s objects "owning" the container, but I don't think we should use k8s names in this section.

semantic_conventions/resource/k8s.yaml
Oberon00, line 66: This is a bit confusing, because (a) I think you mean the k8s Cluster here but (b) you are explaining something about the container.name in the description of k8s.container.name, mixing the two different levels. I think we can't really avoid that, but maybe we can be more clear:

1. Is the container-runtime-level container name actually unique in the whole k8s-cluster?
2. Is the name actually only unique on the k8s node (i.e. within the local container runtime's scope)?

I think only node-local uniqueness is guaranteed although if UUIDs are used, then the ID is not only unique within that single k8s cluster but globally.

semantic_conventions/resource/container.yaml
Oberon00: Yes, I meant the furious_heisenberg et.al. names 😃

semantic_conventions/resource/k8s.yaml
dmitryax: The container name provided to docker engine is actually globally unique, it uses pod's UUID, format is `k8s_POD_<k8s.container.name>_<k8s.namespace>_<k8s.pod.uid>_<k8s.restart_count>`. I'll update the description

tigrannajaryan:
@open-telemetry/specs-approvers we need more reviews of this small change.

bogdandrutu (approved):

bogdandrutu:
Just clarifications, ok to merge.

PR #1842, opened 2021-08-02
Add details for Aggregator in Metrics Spec.

victlu:
Add more details for an Aggregator in the Metrics Spec.

This PR is a rework of #1804 after discussions and feedback from metric SIG meetings.

PR #320, opened 2019-10-18
Update CONTRIBUTING.md

songy23:
Mention markdown style requirement and rearrange sections.

Follow-up of https://github.com/open-telemetry/opentelemetry-specification/pull/268#issuecomment-543306010.

PR #303, opened 2019-10-11
Add semantic conventions for Resource labels

tigrannajaryan:
This document previously contained semantic conventions for Spans
but was missing conventions for Resources. I copied Resource conventions
from OpenCensus project [1] and also added Service conventions which
were a TODO in OpenCensus conventions.

[1] - https://github.com/census-instrumentation/opencensus-specs/blob/master/resource/StandardResources.md

PR #1983, opened 2021-10-01
Incorporate histogram min and max fields into histogram language, add…

### LINKED ISSUE #1902: Support summary aggregation type for metrics

jack-berg:
… zero bucket histogram aggregation

Resolves #1902.

## Changes

* Update explicit bucket histogram language to describe capturing min and max
* Add new Zero Bucket Histogram aggregation, a convenience method for specifying an explicit bucket histogram with zero buckets which produces histogram data points that capture summary data (min, max, sum, count)

### DIFF
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -301,6 +301,7 @@ The SDK MUST provide the following `Aggregation` to support the
 - [Last Value](./sdk.md#last-value-aggregation)
 - [Histogram](./sdk.md#histogram-aggregation)
 - [Explicit Bucket Histogram](./sdk.md#explicit-bucket-histogram-aggregation)
+- [Zero Bucket Histogram](./sdk.md#zero-bucket-histogram-aggregation)
 
 #### None Aggregation
 
@@ -388,12 +389,33 @@ This Aggregation honors the following configuration parameters:
 | Temporality | Delta, Cumulative | Cumulative | See [Temporality](./datamodel.md#temporality). |
 | Boundaries | double\[\] | [ 0, 5, 10, 25, 50, 75, 100, 250, 500, 1000 ] | Array of increasing values representing explicit bucket boundary values.<br><br>The Default Value represents the following buckets:<br>(-&infin;, 0], (0, 5.0], (5.0, 10.0], (10.0, 25.0], (25.0, 50.0], (50.0, 75.0], (75.0, 100.0], (100.0, 250.0], (250.0, 500.0], (500.0, 1000.0], (1000.0, +&infin;) |
 
-\[1\]: Language implementations may choose the best strategy for handling errors. (i.e. Log, Discard, etc...)
-
 This Aggregation informs the SDK to collect:
 
 - Count of `Measurement` values falling within explicit bucket boundaries.
 - Arithmetic sum of `Measurement` values in population.
+- Min `Measurement` value in population.
+- Max `Measurement` value in population.
+
+#### Zero Bucket Histogram Aggregation
+
+The Zero Bucket Aggregation informs the SDK to collect data for
+the [Histogram Metric Point](./datamodel.md#histogram) using zero buckets. This
+produces lightweight histogram data points which capture the min, max, sum, and
+count of a population of measurements.
+
+| Key | Value | Default Value | Description |
+| --- | --- | --- | --- |
+| Monotonic | boolean | true | if true, non-positive values are treated as errors<sup>1</sup>. |
+| Temporality | Delta, Cumulative | Cumulative | See [Temporality](./datamodel.md#temporality). |
+
+This Aggregation informs the SDK to collect:
+
+- Count of `Measurement` values in population
+- Arithmetic sum of `Measurement` values in population.
+- Min `Measurement` value in population.
+- Max `Measurement` value in population.
+
+\[1\]: Language implementations may choose the best strategy for handling errors. (i.e. Log, Discard, etc...)
 
 ## Attribute limits
 


specification/metrics/sdk.md
```diff
@@ -301,6 +301,7 @@ The SDK MUST provide the following `Aggregation` to support the
 - [Last Value](./sdk.md#last-value-aggregation)
 - [Histogram](./sdk.md#histogram-aggregation)
 - [Explicit Bucket Histogram](./sdk.md#explicit-bucket-histogram-aggregation)
+- [Zero Bucket Histogram](./sdk.md#zero-bucket-histogram-aggregation)
```
jack-berg, line 304: As mentioned in the PR description, this is a convenience method. The same thing can be achieved by configuring an explicit bucket histogram with an empty array of bucket boundaries. 

I believe it improves the ergonomics of the sdk enough to justify its existence.

specification/metrics/sdk.md
```diff
@@ -388,12 +389,33 @@ This Aggregation honors the following configuration parameters:
...
-\[1\]: Language implementations may choose the best strategy for handling errors. (i.e. Log, Discard, etc...)
-
 This Aggregation informs the SDK to collect:
 
 - Count of `Measurement` values falling within explicit bucket boundaries.
 - Arithmetic sum of `Measurement` values in population.
+- Min `Measurement` value in population.
+- Max `Measurement` value in population.
+
+#### Zero Bucket Histogram Aggregation
```
jack-berg, line 399: I wanted to call this "Summary Histogram Aggregation", but figured the word summary has a lot of baggage associated with it so may be better to avoid it.

Also considered "MinMaxSumCount Histogram Aggregation".

Also considered omitting the word "Histogram" from these aggregation names. Although including the word "histogram" is confusing because a histogram with zero buckets isn't really a histogram at all, I figured including it is still good to do because it conveys the underlying data point representation.

specification/metrics/sdk.md
```diff
@@ -388,12 +389,33 @@ This Aggregation honors the following configuration parameters:
...
+| --- | --- | --- | --- |
+| Monotonic | boolean | true | if true, non-positive values are treated as errors<sup>1</sup>. |
+| Temporality | Delta, Cumulative | Cumulative | See [Temporality](./datamodel.md#temporality). |
+
+This Aggregation informs the SDK to collect:
+
+- Count of `Measurement` values in population
+- Arithmetic sum of `Measurement` values in population.
+- Min `Measurement` value in population.
+- Max `Measurement` value in population.
```
jack-berg, line 416: Unlike Explicit Bucket Histogram, count and sum capture values across the whole population here instead of just those within the bucket boundaries. This distinction is an additional reason to keep an aggregation around for this specific use case.

specification/metrics/sdk.md
reyang: 1. Naming is probably the hardest thing for developers.
2. "**Zero** bucket" sounds like a bug, because technically we have one bucket which is either `(-Inf, +Inf)` or `[0, +Inf)`.
3. I did some search and I found this https://www.cise.ufl.edu/~adobra/approxqp/histograms2, which suggests **Trivial Histogram**, which I would give 👍.

reyang (approved):
LGTM besides the [naming](https://github.com/open-telemetry/opentelemetry-specification/pull/1983/files#r720501828), which is subjective though 🤷

specification/metrics/sdk.md
jack-berg: Yes, naming and cache invalidation 😁.

I don't mind Trivial.

What do you think about "Degenerate Histogram". The [definition of degeneracy](https://en.wikipedia.org/wiki/Degeneracy_(mathematics)) as it applies to math is quite fitting:

> a degenerate case is a limiting case of a class of objects which appears to be qualitatively different from (and usually simpler than) the rest of the class

reyang: No strong opinion, either way sounds fine to me.

specification/metrics/sdk.md
```diff
@@ -388,12 +389,33 @@ This Aggregation honors the following configuration parameters:
...
+#### Degenerate Histogram Aggregation
+
+The Degenerate Histogram Aggregation informs the SDK to collect data for
+the [Histogram Metric Point](./datamodel.md#histogram) using zero buckets. This
+produces lightweight histogram data points which capture the min, max, sum, and
+count of a population of measurements.
+
+| Key | Value | Default Value | Description |
+| --- | --- | --- | --- |
+| Monotonic | boolean | true | if true, non-positive values are treated as errors<sup>1</sup>. |
```
bogdandrutu, line 408: There is no "monotonic" property in the histogram.

specification/metrics/sdk.md
bogdandrutu: Not sure I understand, this was from the beginning the case.

bogdandrutu (changes requested):

specification/metrics/sdk.md
jack-berg: Cherry picked this language from the [explicit bucket histogram](https://github.com/open-telemetry/opentelemetry-specification/blame/main/specification/metrics/sdk.md#L387). Relevant convo [here](https://github.com/open-telemetry/opentelemetry-specification/pull/1842/files#r685602689).

[PR #320](https://github.com/open-telemetry/opentelemetry-proto/pull/320) in the proto proposed a way to solve this but was ultimately closed, but [Issue #303](https://github.com/open-telemetry/opentelemetry-proto/issues/303) it aimed to resolve is still open. 

I could come back in a PR if we think its best to remove `monotonic` until that issue is resolved, but I do think that language should be consistent for now.

specification/metrics/sdk.md
jack-berg: Mispoke on `sum`, but for `count` the language of explicit bucket histogram [states](https://github.com/open-telemetry/opentelemetry-specification/blame/main/specification/metrics/sdk.md#L395-L396) that it only counts values falling within the bucket bounds:
```
- Count of `Measurement` values falling within explicit bucket boundaries.
- Arithmetic sum of `Measurement` values in population.
```
We should include all values in count when there are zero buckets.

PR #1985, opened 2021-10-04
Clarify OTLP/HTTP port handling.

Oberon00:
Clarifies that default ports for the URL schemes, not default OTLP/HTTP port should be used if no port is given in URL.

See https://github.com/open-telemetry/opentelemetry-specification/pull/1975#discussion_r719238784

### DIFF
--- a/specification/protocol/exporter.md
+++ b/specification/protocol/exporter.md
@@ -48,6 +48,12 @@ for each signal as follow:
    Non-normatively, this could be implemented by ensuring that the base URL ends with
    a slash and then appending the relative URLs as strings.
 
+If no port is given, the URL MUST be used without any extra port modification,
+letting the HTTP library use whatever default port it normally uses for the scheme
+(i.e., 443 for `https://` and 80 for `http://`, though some libraries/configurations
+might allow overriding that by some means like
+[`/etc/services`](https://man7.org/linux/man-pages/man5/services.5.html))
+
 #### Example 1
 
 The following configuration sends all signals to the same collector:
@@ -71,7 +77,7 @@ export OTEL_EXPORTER_OTLP_METRICS_ENDPOINT=https://collector.example.com/v1/metr
 This will send traces directly to the root path `http://collector:4318/`
 (`/v1/traces` is only automatically added when using the non-signal-specific
 environment variable) and metrics
-to `https://collector.example.com/v1/metrics`.
+to `https://collector.example.com/v1/metrics`, using the default https port (443).
 
 #### Example 3
 
@@ -83,9 +89,10 @@ export OTEL_EXPORTER_OTLP_METRICS_ENDPOINT=https://collector.example.com/v1/metr
 ```
 
 Traces are sent to `http://collector:4318/mycollector/v1/traces`
-and metrics to `https://collector.example.com/v1/metrics/`
-(other signals, would they be defined, would be sent to their specific paths
-relative to `http://collector:4318/mycollector/`).
+and metrics to `https://collector.example.com/v1/metrics/`, using the default
+https port (443).
+Other signals, (if there were any) would be sent to their specific paths
+relative to `http://collector:4318/mycollector/`.
 
 ### Specify Protocol
 


CHANGELOG.md
```diff
@@ -93,8 +97,6 @@ release.
 ### OpenTelemetry Protocol
 
 - Add environment variables for configuring the OTLP exporter protocol (`grpc`, `http/protobuf`, `http/json`) ([#1880](https://github.com/open-telemetry/opentelemetry-specification/pull/1880))
-- Specify the behavior of the OTLP endpoint variables for OTLP/HTTP more strictly
```
Oberon00, line 96: This was not in 1.7. Moved up.

anuraaga (approved):

iNikem (approved):

specification/protocol/exporter.md
```diff
@@ -71,7 +77,7 @@ export OTEL_EXPORTER_OTLP_METRICS_ENDPOINT=https://collector.example.com/v1/metr
 This will send traces directly to the root path `http://collector:4318/`
 (`/v1/traces` is only automatically added when using the non-signal-specific
 environment variable) and metrics
-to `https://collector.example.com/v1/metrics`.
+to `https://collector.example.com/v1/metrics`, using the default https port (443).
```
jsuereth, line 80: What's the plan here over time?   If we get IANA port approval are we going to use our own URL prefix for the default OTLP/HTTP port?

specification/protocol/exporter.md
```diff
@@ -48,6 +48,12 @@ for each signal as follow:
    Non-normatively, this could be implemented by ensuring that the base URL ends with
    a slash and then appending the relative URLs as strings.
 
+If no port is given, the URL MUST be used without any extra port modification,
+letting the HTTP library use whatever default port it normally uses for the scheme
```
tigrannajaryan, line 52: I think this gives up too much control to the HTTP library. Essentially it says it is implementation specific. I think we need to be more precise here and dictate an exact behavior (just enforce 443 and 80).

specification/protocol/exporter.md
tigrannajaryan: > use our own URL prefix

Do you mean to register our own URI scheme, e.g. `otlphttp://collector.example.com/v1/metrics` and `otlpgrpc://collector.example.com`?

specification/protocol/exporter.md
anuraaga: This thought came to mind, but I thought of many instances of implementing API client libraries on top of an HTTP library - never once has this been an issue. I think in practice developers want to just pass an HTTP URL as-is to an HTTP library, and users don't actually pay a price for it - since everyone's happy I liked @Oberon00's wording the best.

Oberon00: Then you need to do extra work e.g. when using https://man7.org/linux/man-pages/man3/getaddrinfo.3.html
We can add extra wording that 443 and 80 must always be used, I guess it won't be relevant in practice but will be (slightly) harder to implement.
Do we care what is sent in the `Host`  header? That one would usually contain `:80` / `:443` if you add something to the URL I think.

Oberon00: Maybe it would be best to just remove the whole detailed wording, i.e. remove this:

> (i.e., 443 for `https://` and 80 for `http://`, though some libraries/configurations
might allow overriding that by some means like
[`/etc/services`](https://man7.org/linux/man-pages/man5/services.5.html))

It's maybe rather useless detail. Though it might happen that the question comes up while implementing in some SIG, we could come back to that then.

Oberon00: How about copying https://datatracker.ietf.org/doc/html/rfc7230#section-2.7.1 then?

> If the port subcomponent is empty or not given, TCP port 80 (the reserved port for WWW services) is the default.

We could use that as-is.

Oberon00: This "is the default" wording is still not completely ruling out using another port but I'd say if that RFC doesn't do it and there are no interoperability problems with HTTP URLs then we should also be fine.

tigrannajaryan: Sounds good to copy and quote. Or link to the RFC.

anuraaga: I think the key point is we should not force SDKs to "do something" related to this. I think what we want is port 80 for http and 443 for https, but this may just be confirming that is the behavior of underlying HTTP library (I'd be surprised to find one that doesn't!). I don't think any SDKs should be doing things like "if no colon and starts with http:// add :80 else add :443") type of logic - these are aspects of HTTP not OTLP. We shouldn't use an HTTP library that doesn't have this behavior within the library, but in practice they all do since otherwise they aren't HTTP libraries.

tigrannajaryan: I agree with you @anuraaga 
My only objection is for the case when someone uses some weird http library which for some reason defaults to a different port (I can't imagine why) and then claims that that is what spec told them to do. I don't want to be responsible for mess it can cause :-)

Oberon00: I reworded that paragraph in 75c184c. What do you think?

tigrannajaryan: Sorry, when I read I feel like we are trying to avoid saying directly what we want to say. Any reason we can't simply say that it just needs to be 80 and 443? (we can also link RFC if we want to, but it seems pretty widely known fact, so maybe that's also unnecessary).

I may be missing something, but to me it looks like we should just say it directly.

anuraaga: > Any reason we can't simply say that it just needs to be 80 and 443? 

Perhaps guidance can both specify this while still being clear.

"If port is left out, it MUST be 80 for http scheme and 443 for https scheme. Most underlying HTTP libraries will follow this behavior in which case the SDK SHOULD not set any explicit port."

Oberon00: > Any reason we can't simply say that it just needs to be 80 and 443?

The RFC is also not saying that, it just says that this should be the "default".

> Most underlying HTTP libraries will follow this behavior **in which case** the SDK SHOULD not set any explicit port.

Modifying the URL to include the port should never be done as that would change the `Host` header. I think I will just copy the RFC as literally as it makes sense.

Oberon00: Added 345e87e. I think we are now exactly as explicit about 443 and 80 as the RFC.

tigrannajaryan (approved):
LGTM.

carlosalberto (approved):

Issue #1593, opened 2021-04-02
Editorial change - call out that Counter is monotonic


reyang:
This is a small editorial change pointed out by @jonatan-ivanov in a [PR review comment](https://github.com/open-telemetry/opentelemetry-specification/pull/1590#discussion_r606382472).

PR #1995, opened 2021-10-06
Add supplementary doc about monotonicity

reyang:
Related to #1593.

I think before calling out something is monotonic, we need to clarify what monotonicity means.


specification/metrics/supplementary-guidelines.md
```diff
@@ -62,6 +64,70 @@ Here is one way of choosing the correct instrument:
...
+  * the process didn't ask the operating system to report the page faults
+* At T<sub>3</sub>:
+  * the operating system reported with `1050` page faults for the process
+* At T<sub>4</sub>:
+  * the operating system reported with `1200` page faults for the process
+
+You can see that the number being reported is the absolute value rather than
+increments, the value is non-negative, and the value is monotonically
+increasing. Note that the mathematical definition of "monotonically increasing"
+does not require the value to be always non-negative, this additional semantic
```
reyang, line 124: I want to see if folks would agree that Asynchronous Counter should always report non-negative value.

reyang: > in which case it's better to qualify that explicitly as "monotonic, non-negative" rather than to change the math definition of monotonic.

+1, this is why I'm calling it out in the PR and see if we want to update the spec wording.

PR #1997, opened 2021-10-06
Clarify default for OTLP endpoint should, not must, be https

### LINKED ISSUE #1984: Document discrepancy with spec or loosen spec for default value of exporter endpoint
I randomly noticed the default value for endpoint for exporter is `https://localhost`

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#configuration-options

Java uses `http://localhost` - I think fixing this now would be too disruptive to existing users without enough benefit. Should this be documented here somehow or the spec wording loosened to allow it? FWIW, I believe `https://localhost` has no real user case and suspect where secure is implemented as the default in the SDK, 100% of users are setting the variable themselves - in case that matters here.

anuraaga:
Fixes #1984

## Changes

Adds a subnote clarifying that while the OTLP endpoint variables should default to https, it is not a must

I mostly copied the wording from #1969


tigrannajaryan (approved):
LGTM, to avoid breaking a stable SDK release.

carlosalberto (approved):

tigrannajaryan:

jack-berg:
After more digging, I found [this](https://github.com/open-telemetry/opentelemetry-specification/pull/699) PR that seems to have been the seed for the original `https` default. It added the original `OTEL_EXPORTER_OTLP_*_INSECURE` flags, which defaulted to `false`, and seem to have inspired the `https` default. 

I can't find any context in that PR related to why to default to secure.

jsuereth:
This has enough approvals.  Going to leave open for 1 more business day for feedback before enabling auto-merge.

PR #1998, opened 2021-10-06
Don't set Span.Status for 4xx http status codes for SERVER spans

### LINKED ISSUE #1818: Span.Status and http semantic convention
[HTTP semantic convention](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#status) says:

> For HTTP status codes in the 4xx and 5xx ranges, as well as any other code the client failed to interpret, status MUST be set to Error.

This was introduced in the first version of this semantic convention and never discussed or challenged later. But although this requirement probably makes sense for CLIENT http spans, it is hardly sensible for SERVER http spans. SERVER spans are created based on the "external event": somebody made a request to my server for the URL that I cannot serve. I have no control over such requests and thus I have no way to fix this "error".

I propose:

-  to relax semantic convention and say that "For HTTP status codes in 4xx range, status MAY be set to Error"
- to allow auto-instrumentation to expose some kind of configuration how to interpret http codes in 4xx range

## Alternative approach

We can say (after https://github.com/open-telemetry/opentelemetry-collector/issues/3571 is fixed) that application developer or Operator are free to override `Span.Status` from `Error` to `Ok`. But I think this use-case is common enough not to force all affected users to perform the same manual work.

iNikem:
Fixes #1818

In addition to all the reasons given in #1818 we have numerous complaints from end-users who are frustrated by seeing all those "error" traces that they can do nothing about.

## Changes

Don't set Span.Status for 4xx http status codes for SERVER spans

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ release.
 
 - Add optional min / max fields to histogram data model.
   ([#1915](https://github.com/open-telemetry/opentelemetry-specification/pull/1915))
+- Add exponential histogram to the metrics data model.
+  ([#1935](https://github.com/open-telemetry/opentelemetry-specification/pull/1935))
 
 ### Logs
 
@@ -22,12 +24,17 @@ release.
 
 ### Semantic Conventions
 
+- BREAKING: Remove db.cassandra.keyspace and db.hbase.namespace, and clarify db.name
+  ([#1973](https://github.com/open-telemetry/opentelemetry-specification/pull/1973))
+
 - Don't set Span.Status for 4xx http status codes for SERVER spans.
   ([#1998](https://github.com/open-telemetry/opentelemetry-specification/pull/1998))
 
 ### Compatibility
 
 ### OpenTelemetry Protocol
+
+- Make OTLP/HTTP the recommended default transport ([#1969](https://github.com/open-telemetry/opentelemetry-specification/pull/1969))
 
 ### SDK Configuration
 
@@ -88,6 +95,8 @@ release.
   ([#1890](https://github.com/open-telemetry/opentelemetry-specification/pull/1890))
 - Add HTTP request and response headers semantic conventions.
   ([#1898](https://github.com/open-telemetry/opentelemetry-specification/pull/1898))
+- Add `k8s.container.restart_count` Resource attribute.
+  ([#1945](https://github.com/open-telemetry/opentelemetry-specification/pull/1945))
 
 ### Compatibility
 
--- a/schemas/1.8.0
+++ b/schemas/1.8.0
@@ -1,0 +1,14 @@
+file_format: 1.0.0
+schema_url: https://opentelemetry.io/schemas/1.8.0
+versions:
+  1.8.0:
+    spans:
+      changes:
+        - rename_attributes:
+            attribute_map:
+              db.cassandra.keyspace: db.name
+              db.hbase.namespace: db.name
+  1.7.0:
+  1.6.1:
+  1.5.0:
+  1.4.0:
--- a/semantic_conventions/resource/k8s.yaml
+++ b/semantic_conventions/resource/k8s.yaml
@@ -63,6 +63,13 @@ groups:
         brief: >
           The name of the Container in a Pod template.
         examples: ['redis']
+      - id: restart_count
+        type: int
+        brief: >
+          Number of times the container was restarted. This attribute can be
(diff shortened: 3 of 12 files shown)


specification/trace/semantic_conventions/http.md
```diff
@@ -46,8 +46,11 @@ default span name.
 the response body; or 3xx codes with max redirects exceeded), in which case status
 MUST be set to `Error`.
 
-For HTTP status codes in the 4xx and 5xx ranges, as well as any other code the client
-failed to interpret, status MUST be set to `Error`.
+For HTTP status codes in the 4xx range span status MUST be left unset in case of `SpanKind.SERVER`
+and MUST be set to `Error` otherwise.
```
Oberon00, line 50:
```suggestion
and MUST be set to `Error` in case of `SpanKind.CLIENT`.
```

"otherwise" could also mean "if not in the 4xx range"

specification/trace/semantic_conventions/http.md
```diff
@@ -46,8 +46,11 @@ default span name.
 the response body; or 3xx codes with max redirects exceeded), in which case status
 MUST be set to `Error`.
 
-For HTTP status codes in the 4xx and 5xx ranges, as well as any other code the client
-failed to interpret, status MUST be set to `Error`.
+For HTTP status codes in the 4xx range span status MUST be left unset in case of `SpanKind.SERVER`
```
Oberon00, line 49: I agree with these being MUSTs instead of SHOULDs because I think that instrumentations should not be burdened with implementing configuration for this.
Maybe we should make this more clear here and also hint that backends might provide more elaborate configuration for error detection (per status-code + route/url-pattern, etc.)

Oberon00:
I'm weakly against this change, but since I don't have user feedback, I yield. 😀

specification/trace/semantic_conventions/http.md
iNikem: should we mention `SpanKand.INTERNAL` here? Just to avoid ambiguity, although INTERNAL spans should not have HTTP status codes.

specification/trace/semantic_conventions/http.md
iNikem: We can do that, but I am not sure if spec should talk about telemetry consumers/backends much. Unless this is actually makes sense, such as we did for interpretation of `SpanStatus.OK` by backends.

Oberon00: Theoretically the hint I mentioned does not belong in a spec, I agree. It would belong in a user manual, but we don't have one. Maybe some md file outside spec, but then not in this PR.

specification/trace/semantic_conventions/http.md
iNikem: > How about wording it as "client-side spans" and "server-side spans", independent of the concrete span king chosen?

Too vague for the spec, IMO. We don't have "client-side spans"/"server-side spans" defined anywhere in the spec.

Oberon00: In the HTTP semconv, we say:

> HTTP client
>
> This span type represents an outbound HTTP request.
>
> For an HTTP client span, SpanKind MUST be Client.

Similarly for server. So the spec more or less forbids using INTERNAL (or CONSUMER/PRODUCER) for now. Should we rely on that here?

tigrannajaryan (approved):
I am weakly in favour of the proposed change (4xx not a server error), however I understand that for some people the opposite is desirable. Unfortunately I cannot think of definitive arguments in favour or against either approach. If we cannot reach consensus on this we can make this behavior configurable.

yurishkuro (approved):

Oberon00:
> If we cannot reach consensus on this we can make this behavior configurable.

That is the one thing I definitely don't want. That configurability should be on the backend.

tedsuo:
@Oberon00 I disagree about configuration. We need to define a configuration language for OpenTelemetry instrumentation, to handle these issues.  But this should be done thoughtfully, as part of the spec. Configuration for instrumentation should not be implemented in an ad-hoc manner; all instrumentation we provide should use the same configuration language.

The instrumentation SIG is working on this configuration issue, but we would greatly appreciate more participation from core Spec members.

As far as this PR goes, I'm fine with the change. My request is that there is some follow up with contrib maintainers and the instrumentation SIG to make sure this change is actually implemented.

PR #2003, opened 2021-10-08
Framework to support W3C test-driven specification

### LINKED ISSUE #1210: Make it possible to automatically list all specification requirements

ocelotl:
Fixes #1210 

There is a framework for specifications from the W3C, the [W3C QA Framework](https://www.w3.org/TR/qaframe-spec/), this is its abstract:

> The goal of this document is to help W3C editors write better specifications, by making a specification easier to interpret without ambiguity and clearer as to what is required in order to conform. It focuses on how to define and specify conformance. It also addresses how a specification might allow variation among conforming implementations. The document presents guidelines or requirements, supplemented with good practices, examples and techniques.

#### What does this PR do?

1. It changes one of the spec documents (`specification/baggage/api.md`) to make it follow the W3C QA framework. This change is intended to show how a document would look after being changed to follow the W3C QA Framework.
2. It adds a script that can parse the spec documents that follow the W3C QA Framework and produce JSON files that can list the included requirements, how mandatory they are and their identifier.
3. Adds a [conformance clause](https://www.w3.org/QA/WG/2004/08/SpecGL-template-text.html) for OpenTelemetry that details how the documents that form this specification are to be written.

#### Why did I create this PR?

While I was working on implementing Metrics, I found myself first having to read all the related documents to find the requirements so that I could know what to implement. This is something that should be as easy as looking for all instances of `MUST`, `SHOULD`, `MAY`, etc but it is not so because there is no guarantee that as the specification is written today every requirement is written in this way (please see [this](https://github.com/open-telemetry/opentelemetry-specification/pull/2003#issuecomment-940984533) comment for examples). I found it necessary then to make this PR more strict in the way it is being written so that its requirements could be extracted from it.

#### What is the general strategy here?

With the approach suggested here we should be able to generate JSON files that contain all requirements so that they can be then transformed into test cases in each implementation. Once an implementation implements and passes all of the mandatory test cases (the ones whose corresponding RFC 2119 keyword is `MUST [NOT]` or equivalent) and it can consider itself compliant.

This idea will require of course rewriting all the documents, but this can be done little by little with one PR per document until we can have all the specification in this format. It is not necessary for this project to stop being developed to format all documents, it continue with some documents in this format and some not.

#### What about the compliance matrix?

The [compliance matrix](https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md) gathers all the features that each SIG has implemented and marks them as optional or not. I find it convenient but I don't think it is something that gives the level of detail that this approach gives. The compliance matrix is feature-oriented, this approach is requirement-oriented. I think both serve their specific purpose and can continue to exist together.


tigrannajaryan:
It would be useful to to start with the motivation for the change, what are the cons and pros and why we should adopt it.

This appears related to https://github.com/open-telemetry/opentelemetry-specification/pull/1211/files which did not gain traction. What has changed that we should consider this now?

jmacd:
I am a really big fan of this work, @ocelotl!

The OpenTelemetry specification is expansive and written by many authors. As we see from your work on #1956, it is difficult if not impossible to construct a compliance matrix from the text of the specification without having the authors and the community members there to help with interpretation.

It looks like this can be done iteratively: one step at a time, a part of the specification will be rewritten to match the style given here, at which point the compliance matrix row headings can be automatically checked or generated.

Because this is a large PR, I want to encourage you to first (sorry, procedural nits!) file an issue describing the problem "Make the OpenTelemetry specification automatically testable" or something similar. You've already written the motivation above, but we use issues to prioritize discussions about large changes ahead of actual pull requests, usually.

Please file an issue that repeats the motivation and also gives a brief explanation of the steps, e.g., (1) merge the spec-writing guidelines, (2) update document status states to manage the transition, (3) merge the tools for generating x/y/z (4) finally for each component of the specification, (a) refactor, (b) regenerate compliance matrix, (c) ...

hdost:
Is this something that's still being worked on, but maybe elsewhere?

ocelotl:
> Is this something that's still being worked on, but maybe elsewhere?

Not that I am aware of, curious why you ask :sweat_smile:

PR #2008, opened 2021-10-12
Add telemetry schemas to the specification

tigrannajaryan:
Add telemetry schemas to the specification

This merges the remaining bits of OTEP0152 to the specification
and links to already existing schema-related sections in the API.

The documents are marked as "Experimental", however I would like
to promote file_format_v1.0.0.md to Stable as soon as possible since we
already have published such files.

Related to https://github.com/open-telemetry/oteps/pull/152

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,8 @@ release.
 ### Metrics
 
 - Add optional min / max fields to histogram data model.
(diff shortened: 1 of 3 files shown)


tigrannajaryan:
@open-telemetry/specs-approvers this needs a review. It is merging of an already approved OTEP.

PR #2014, opened 2021-10-13
Change default OTLP endpoint to use http:// scheme

anuraaga:
## Changes

Changes the default scheme for OTLP exporters to http:// because the default host is `localhost`.

With the additional language that SDKs may choose a different default, actually it stops being a breaking change to update this default. I argue that the current default has no use case - it is extremely rare to use TLS with a localhost connection. This prevents the defaults to provide an out-of-the-box experience for any practical situation. It is a better experience if they can provide an OOB experience for something at least. Or otherwise I don't think we would want to have any defaults at all.

While this still seemed relatively harmless, I'm finding that it can cause difficulty in reasoning out other changes, such as in #1895. All the variables should be defined with some target experience in mind - currently there isn't, but the use of `localhost` seems to strongly indicate a target of sidecar.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,11 +26,17 @@ release.
 
 - BREAKING: Remove db.cassandra.keyspace and db.hbase.namespace, and clarify db.name
   ([#1973](https://github.com/open-telemetry/opentelemetry-specification/pull/1973))
+- Add AWS App Runner as a cloud platform
+  ([#2004](https://github.com/open-telemetry/opentelemetry-specification/pull/2004))
 - Add Tencent Cloud as a cloud provider.
   ([#2006](https://github.com/open-telemetry/opentelemetry-specification/pull/2006))
-
 - Don't set Span.Status for 4xx http status codes for SERVER spans.
   ([#1998](https://github.com/open-telemetry/opentelemetry-specification/pull/1998))
+- Add attributes for Apache RocketMQ.
+  ([#1904](https://github.com/open-telemetry/opentelemetry-specification/pull/1904))
+
+- Define http tracing attributes provided at span creation time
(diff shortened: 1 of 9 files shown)


johnnyshields:
See this comment: https://github.com/open-telemetry/opentelemetry-specification/pull/1895#issuecomment-941985063

tigrannajaryan:
> Note, I considered adding some explicit language like "Default values are chosen to provide an out-of-the-box experience for sidecar usage" but left it out. Let me know if something like that makes sense / could be helpful.

I think it is not necessary because the sidecar is not the only deployment configuration when this is suitable. For example http to localhost is also the correct default for apps deployed on a VM/physical host with a local Collector installed as an agent.

I think it is a reasonable approach to make sure the default values are consistent around the assumption that there is a Collector running on the localhost. However I believe it is important to ensure it is not easy to accidentally send insecurely to a non-local host.

If we accept this PR the defaults will be:

| Setting | Default Value |Notes|
|--------|----------|---|
| Endpoint | http://localhost |
| Protocol | http/protobuf | grpc allowed as a default in SDKs for backward compatibility reasons|
|Insecure | false| Ignored, because http protocol is used|

When the destination is not the localhost (i.e. sending directly to a backend or to an intermediary Collector on a different host) then at least the `Endpoint` must be specified, which forces the user to choose `http` or `https` as the scheme in the URL (assuming they keep the default http/protobuf as the `Protocol`). This is good, they need to consciously choose a non-secure transport.

If instead they decide to use grpc to send to a non-local host, then they will have to specify at least `Endpoint` and `Protocol`. Once the `Protocol` is switched to grpc, the `Insecure=false` option kicks in. Again, this is good, can't accidentally send insecurely.

So, it looks good to me.

Issue #1908, opened 2021-09-07
User feedback for the semantic convention


jamesmoessis:
During the 2021-08-31 instrumentation SIG, @tedsuo asked me to provide feedback on the semantic convention from an Atlassian perspective. We have been using and implementing the semantic convention for over 6 months now. Here are a few pieces of feedback we have for it. 

## Problem 1

The Semantic Convention seems flooded with specific details about technologies that aren’t relevant for the vast majority of users. As an arbitrary example, Cassandra takes up a sizeable chunk of the database trace convention. Is this unnecessary complexity for what could be a more general document? 

### Thoughts on Problem 1

Is there a good reason that the specific technologies for each category are listed as part of the core document? Would it be productive to separate out the technologies to "extend" the "core" sections of the conventions? I see that AWS SDK has its own instrumentation section, and wonder if something similar could be done for the many specific technologies that form part of the convention. 

## Problem 2 

Some Semantic Convention guidelines are not practical. 

Examples:
1. HTTP span attribute `http.url`: the guidelines says to send the full URL, however this comes with it's own problems of user privacy, with possibly difficult-to-detect PII/UGC appearing in query strings. 
2. HTTP span name: guideline says use the "HTTP Route", which increases "key operations" in Lightstep. This causes large cost increases. 

## Problem 3

Breaking changes. In its experimental stages, the semantic convention has understandably been brittle to breaking changes. 

### Thoughts on Problem 3 

I don't have any specific solution to this, but I believe mechanisms achieving translations between semantic convention versions are already being discussed. We would be highly in favour of this.

Oberon00:
> HTTP span name: guideline says use the "HTTP Route", which increases "key operations" in Lightstep. This causes large cost increases.

The set of routes should be bounded by the application code, so while there may be quite a few routes, at least it should not depend on what input data you get, etc.
Regarding the key operations/cost, this sounds more like a Lightstep problem. Or would it help if we prefixed the routes with the usual "HTTP GET"?

Oberon00:
> HTTP span attribute http.url: the guidelines says to send the full URL, however this comes with it's own problems of user privacy, with possibly difficult-to-detect PII/UGC appearing in query strings.

[...]

PR #2019, opened 2021-10-14
Refactor trace semantic convention yaml structure

jamesmoessis:
## Changes

I've refactored the semantic convention yaml so it lines up with @tedsuo's draft PR #1977. This is an initial piece of work which precedes refactoring the markdown into a similar structure. 

The structure is as follows: 

```
├── trace
│   ├── database
│   │   ├── database.yaml
│   │   ├── implementations
│   │   │   ├── cassandra.yaml
│   │   │   ├── mongodb.yaml
│   │   │   ├── mssql.yaml
│   │   │   └── redis.yaml
│   │   └── sql.yaml
│   ├── exception
│   │   └── exception.yaml
│   ├── faas
│   │   ├── faas.yaml
│   │   └── implementations
│   │       └── lambda.yaml
│   ├── general.yaml
│   ├── http
│   │   └── http.yaml
│   ├── instrumentation
│   │   └── aws-sdk.yml
│   ├── messaging
│   │   ├── implementations
│   │   │   ├── kafka.yaml
│   │   │   ├── rabbitmq.yaml
│   │   │   └── rocketmq.yaml
│   │   └── messaging.yaml
│   └── rpc
│       ├── implementations
│       │   ├── grpc.yaml
│       │   └── jsonrpc.yaml
│       └── rpc.yaml
```

I deleted the `aws` folder as it had one file `lambda.yaml` and moved it under `faas/implementations`. 

Additionally, `make table-generation` produces no changes to the markdown, so it looks like it isn't breaking anything. 

### Why restructure? 

It has been raised on several occasions that the semantic convention is becoming bloated with implementation-specific conventions. This is the first step to separating them out.

This will help the semantic convention towards stability because we will be able to declare the "core" conventions as stable, while the "implementations" can remain experimental. 

Related issues #

- #1908
- #1977


jamesmoessis:
@tigrannajaryan Only traces and resources have YAML definitions, and I can't think of a way to combine them that would make sense. Restructuring to unify metrics/traces would need to happen in the markdown, since the metrics semconv is defined directly in the markdown.

> many (most?) of the conventions that are applicable to traces are also applicable to logs. 

Do logs have a semantic convention? I can't find it in the spec, so I haven't taken logs into consideration at all.

Ted's PR says: 

> Centralize the location of all semantic conventions, rather than split them across many subsections of the spec. 

> Create one folder per topic (HTTP, faas, messaging, etc). All relevant tracing, metric, and resource conventions are combined into these topics. 

I'm trying to think of how to tackle these in terms of the YAML. In the name of centralization, we could:
- Rename the `trace` folder to something more general, and keep the YAML organised by "topic" like is mentioned. But currently there are only YAML definitions for traces, so I kept the `trace` folder name.
- Move the YAML and markdown nearer to each other in the spec. 

If anyone has ideas about how to structure the yaml, would be interesting to hear.

PR #2022, opened 2021-10-14
Add semantic conventions for log media attributes

### LINKED ISSUE #1772: Semantic Conventions for File Attributes
When consuming logs from a file, information about the file is commonly included as metadata on the log record.

1. Should this information be standardized? If so:
2. What is the schema for this metadata?
3. Does the metadata belong in Resource or Attributes?

As a starting point for discussion, I propose:
1. Yes
2. Establish a `file.*` namespace:
 - `file.name` - The basename of the file (i.e. `mylog.log`).
 - `file.path` - The absolute path of the file (i.e. `/var/log/mylog.log`).
 - `file.name.resolved` - Same as `file.name`, but with symlinks resolved.
 - `file.path.resolved` - Same as `file.name`, but with symlinks resolved.
 - `file.stream` - When relevant, `stdout` or `stderr`.
3. In the case of log ingestion, Attributes is appropriate. The file from which logs are being consumed is not the component that emitted the logs. Rather, it is a medium for conveying logs.

djaglowski:
Resolves #1772 

## Changes

Adds semantic conventions for file log source.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,8 @@ release.
 
 - Fix Syslog severity number mapping in the example.
   ([#2091](https://github.com/open-telemetry/opentelemetry-specification/pull/2091))
+- Add log.* attributes.
+  ([#2022](https://github.com/open-telemetry/opentelemetry-specification/pull/2022))
 
 ### Resource
 
--- a/specification/logs/data-model.md
+++ b/specification/logs/data-model.md
@@ -426,8 +426,8 @@ Description: Additional information about the specific event occurrence. Unlike
 the `Resource` field, which is fixed for a particular source, `Attributes` can
 vary for each occurrence of the event coming from the same source. Can contain
 information about the request context (other than TraceId/SpanId). SHOULD follow
-OpenTelemetry
-[semantic conventions for Attributes](../trace/semantic_conventions/README.md).
+OpenTelemetry [semantic conventions for Log Attributes](./semantic_conventions/README.md) or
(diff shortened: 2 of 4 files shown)


djaglowski:
> I see file related attributes as Resource attributes

I disagree on this point because the log file is not really the _source of the log_, which would be necessary to categorize file attributes as resource attributes. The file is really just a communication mechanism between the source of the log and the consumer of the log.

> Should we define here which type of log data model attributes these should be applied to?

Probably we should try to describe any new resource types as separate documents in [this](https://github.com/open-telemetry/opentelemetry-specification/tree/main/semantic_conventions/resource) directory, so that there are no a log-specific resource types.

dmitryax:
> I disagree on this point because the log file is not really the _source of the log_, which would be necessary to categorize file attributes as resource attributes. The file is really just a communication mechanism between the source of the log and the consumer of the log.

This makes sense, but the attribute namespace name `source_log` contradicts with this statement. Should we consider other name if log file is not source of the log? Maybe `log.file.path` and ``log.file.name`?

zenmoto (approved):

kumoroku (approved):

pmm-sumo (approved):

dmitryax (approved):

tigrannajaryan:
Please rebase.

djaglowski:
This should be ready to merge, unless there is any more feedback.

kumoroku (approved):

PR #2034, opened 2021-10-19
Experimentation attributes

HarryEMartland:
## Changes

Adds new semantic conventions for experiment attributes on spans.

### New Attributes
`experiment.seed`
`experiment.experiment_name.variant`
`experiment.experiment_name.id`


iNikem:
I think there is not enough description of the use-case and it is certainly not self-explanatory. Who and when should apply them?

HarryEMartland:
Thanks for the feedback @iNikem, Is there enough now? Should I add more info on individual attribute descriptions?

Issue #1953, opened 2021-09-21
Create metrics feature spec compliance matrix


jmacd:
**What are you trying to achieve?**

https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md has a "TBD" for metrics feature compliance.

We need to fill in this table. @ocelotl has prepared a very compelling draft of this table, here:
https://github.com/open-telemetry/opentelemetry-python/discussions/2122

PR #2035, opened 2021-10-19
Define how to enable Meter attributes for Prometheus Exporter

### LINKED ISSUE #1906: How to map the Meter name in non-OTLP exporters?
In the [Tracing SDK spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/non-otlp.md#instrumentationlibrary) we have specified the following mapping for the [Tracer name and version](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#get-a-tracer) (= InstrumentationLibrary name/version) for non-OTLP exporters:

> OpenTelemetry `InstrumentationLibrary`'s fields MUST be reported as key-value
pairs associated with the Span using the following mapping:
> | OpenTelemetry InstrumentationLibrary Field | non-OTLP Key |
> | ------------------- | --- |
> | `InstrumentationLibrary.name`|`otel.library.name`|
> | `InstrumentationLibrary.version`|`otel.library.version`|

The Metrics API specifies the [Meter name and version](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#get-a-meter) accordingly. In [OTLP](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto) this is reported as `InstrumentationLibrary` as well.

Should we define the same mapping for non-OTLP metrics exporters as well, i.e., should the meter name and version be added as attributes/labels/tags/dimensions called `otel.library.name` and `otel.library.version`, or would it make more sense to call this `otel.meter.*` here? This would be inconsistent with the Tracing spec, however, since that one could have also just called it `otel.tracer.*` but didn't.
I'd be in favor of staying consistent here, even though seeing `otel.library.*` on the backend might not be as obvious to the user as the more "tangible" notation of a "Meter name".

hdost:
Fixes #1906

## Changes

* Add `OTEL_METRICS_LIBRARY_ATTRIBUTES` variable
* Update Metric SDK to clarify when Meter attributes should be exported.

Related issues #1953

### DIFF
--- a/specification/sdk-environment-variables.md
+++ b/specification/sdk-environment-variables.md
@@ -149,8 +149,14 @@ thrift or protobuf.  As of 1.0 of the specification, there
 
 | Name                          | Description                     | Default                      |
 | ----------------------------- | --------------------------------| ---------------------------- |
-| OTEL_EXPORTER_PROMETHEUS_HOST | Host used by the Prometheus exporter | All addresses: "0.0.0.0"|
-| OTEL_EXPORTER_PROMETHEUS_PORT | Port used by the Prometheus exporter | 9464                    |
+| `OTEL_EXPORTER_PROMETHEUS_HOST` | Host used by the Prometheus exporter | All addresses: "0.0.0.0"|
+| `OTEL_EXPORTER_PROMETHEUS_PORT` | Port used by the Prometheus exporter | 9464                    |
+| `OTEL_EXPORTER_PROMETHEUS_ADD_METER_ATTRIBUTES` | Filter for if `name` and `version` from [Meter](./metrics/api.md#get-a-meter) are exported. | `false` | |
+
+Known values for `OTEL_EXPORTER_PROMETHEUS_ADD_METER_ATTRIBUTES` are:
+
+- `true` : Add the Meter `name` and `version` as Attributes on the metrics exported.
+- `false` : Don't add Meter `name` and `version` as Attributes on the metrics exported.
 
 ## Exporter Selection
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ release.
 - Add clarifications on how to determine aggregation temporality.
   ([#2013](https://github.com/open-telemetry/opentelemetry-specification/pull/2013),
   [#2032](https://github.com/open-telemetry/opentelemetry-specification/pull/2032))
+- Define how Prometheus should handle Meter attributes.
+  ([#2035](https://github.com/open-telemetry/opentelemetry-specification/pull/2035))
 
 ### Logs
 
--- a/specification/metrics/sdk_exporters/prometheus.md
+++ b/specification/metrics/sdk_exporters/prometheus.md
@@ -5,3 +5,10 @@
 Prometheus Exporter is a [Pull Metric Exporter](../sdk.md#pull-metric-exporter)
 which reacts to the Prometheus scraper and report the metrics passively to
 [Prometheus](https://prometheus.io/).
+
+Prometheus Exporter MUST export Meter `name` and `version` information when
+enabled by the [Metrics SDK Configuration](../../sdk-environment-variables.md#metrics-sdk-configuration).
+The Prometheus Exporter MUST respect the configuration preferences of the application
+provide a method to configure whether these data are exported as
+[Attributes](../common/common.md#attributes). The `name` and `version` information
+will follow the same naming as [trace exporters.](../../trace/sdk_exporters/non-otlp.md#instrumentationlibrary)


hdost:
> > > 1. @hdost Would you be willing to put together a prototype or more example use case here around what attributes across all of OTLP should be in prometheus?  I'm happy to help, just overloaded in time at the moment trying to build the otel roadmap.
> 
> > I think I can arrange time to do that. What is the timeframe we are looking so I can prioritize?
> 
> I'd like to see a prototype/usage example in prometheus prior to merging the specification change, if that's feasible.

@jsuereth  Yup, I was more wondering in the context of timeline, is the goal to have this behavior defined before Feature freeze of he SDK or just stabilization.

jsuereth:
@hdost IIUC - Prometheus Exporter specification can happen post SDK-freeze unless you need a feature from the SDK for this.

It's my understanding that we do not need SDK features for this, so we're free to take some time to get this right.   I'd like to make sure whatever you do here lines up with the formal specification for prometheus.

I think the only thing I'm objecting too on this PR right now is that I think adding `otel.library` (or whatever the label is we'll use for insturmentation library) should be *default on* for prometheus export.

jmacd:
We discussed this at length in the Prometheus working group SIG and consulted with @brian-brazil.

There is a fundamental question which is practically orthogonal to the issue at hand, which is when can we ever say that two metrics are the "same". This question borders on philosophical--we discussed a hypothetical set of two libraries both with a "request_count" metric. Brian's experience says that rarely are two implementations exactly the same, so they should be considered different almost always. While `cpu.seconds` may be given a single meaning, very often "request_count" differs slightly based on where in a pipeline you are counting requests, so a more-refined metric name should be used. E.g., if you were counting Kafka requests and Redis requests, you should probably use different metric names. 

Here are the recommendations we arrived at when we landed this discussion:
- Inside the same binary, 
  - two instruments with the same name SHOULD use the same instrument kind, SDK MUST warn the user and reject one
  - two instruments with the same name SHOULD use distinct attribute sets, because the collector can't validate
-  Inside a collector, when exporting to Prometheus:
  - do nothing by default; if the application follows the rules, nothing bad happens except possibly combining metrics w/ slightly different implementations
  - optionally, if the user wants to consistently apply an instrumentation-library attribute, this should be an option, but it's not a good default behavior because it has to be done consistently across a system
- General recommendations
  - Prefix the library name to the metric name if two libraries might use the same attribute set inside a single binary
  - Follow distinct resource practices to avoid breaking the single-writer rule, so ultimately this issue boils down to another case of #1782

hdost:
> Prefix the library name to the metric name if two libraries might use the same attribute set inside a single binary

Oh so now that is quite interesting 🤔 basically this would namespace things, so if you're instrumenting
"io.otel.metrics" and "io.otel.traces" then the metrics could optionally be appended with those names sanitized for prometheus as an example:`io_otel_traces_<metric_name>`

I could see this exploding the metric name length depending on what comes through.

specification/sdk-environment-variables.md
```diff
@@ -149,8 +149,14 @@ thrift or protobuf.  As of 1.0 of the specification, there
...
 | ----------------------------- | --------------------------------| ---------------------------- |
-| OTEL_EXPORTER_PROMETHEUS_HOST | Host used by the Prometheus exporter | All addresses: "0.0.0.0"|
-| OTEL_EXPORTER_PROMETHEUS_PORT | Port used by the Prometheus exporter | 9464                    |
+| `OTEL_EXPORTER_PROMETHEUS_HOST` | Host used by the Prometheus exporter | All addresses: "0.0.0.0"|
+| `OTEL_EXPORTER_PROMETHEUS_PORT` | Port used by the Prometheus exporter | 9464                    |
+| `OTEL_EXPORTER_PROMETHEUS_ADD_METER_ATTRIBUTES` | Filter for if `name` and `version` from [Meter](./metrics/api.md#get-a-meter) are exported. | `false` | |
+
+Known values for `OTEL_EXPORTER_PROMETHEUS_ADD_METER_ATTRIBUTES` are:
+
+- `true` : Add the Meter `name` and `version` as Attributes on the metrics exported.
```
reyang, line 158: I did some quick prototype in .NET https://github.com/open-telemetry/opentelemetry-dotnet/pull/2593.

Couple findings:
1. the label name `name` and `version` are concerning, I think it is very likely to conflict with existing labels (it is totally possible to invent some rule although I didn't cover it in the prototype).
2. the [Metrics API Spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#get-a-meter) guarantees a non-empty meter name, however there is no guarantee on the meter version. I think we should call out here what's the behavior if version is not available or is an empty string.
3. I think there are many cases people might care about the meter name, but not the version, I think the current proposal (having both or nothing) might be less useful.

specification/sdk-environment-variables.md
```diff
@@ -149,8 +149,14 @@ thrift or protobuf.  As of 1.0 of the specification, there
...
+- `false` : Don't add Meter `name` and `version` as Attributes on the metrics exported.
```
reyang, line 159: I guess it meant to be "Prometheus labels" instead of "Attributes"?

reyang (changes requested):
Please find addition feedback based on the prototype https://github.com/open-telemetry/opentelemetry-specification/pull/2035/files#r747236300.

jonatan-ivanov (commented):
Since some backends has reserved words for tag name, I would try to choose something more specific than `name` and `version`, like `meter_name` and `meter_version`. Also, if I see name and version on the tags, I'm not sure it will be clear to me what these exactly mean.

(Not really Prometheus-related but [datadog reserves `version`](https://docs.datadoghq.com/getting_started/tagging/))

jmacd:
As discussed in the SIG call today, we should be using the same attribute names already specified for tracing systems for consistency. This means "otel.library.name" and "otel.library.version", which in a Prometheus setting should have dots converted to underscores.

PR #2045, opened 2021-10-20
Unset and empty environment variables are equivalent

pellared:
## Changes

Describe how an empty value of an environment variable should be interpreted.

Related PR: https://github.com/open-telemetry/opentelemetry-specification/pull/2052

## Why

The problem is that not all languages/OSes/shells are handling an empty value in the same way.

In .NET (C# etc) setting an empty value unsets the environmental variable (sic!). See: https://docs.microsoft.com/en-us/dotnet/api/system.environment.setenvironmentvariable?view=net-5.0

On Windows setting an empty env var is VERY tricky. I suspect 95% of Windows IT users may have real problems doing it. You cannot do it in the command prompt. not via PowerShell, not via GUI. It is possible via Windows Registry or WinAPI, but it is VERY inconvenient and hacky.

Also from a pure usability point of view, `echo $SOMEVAR` does not allow to distinguish between empty and unset.

Having a different interpretation of an empty and unset environment variable can be seen as a kind of "billion-dollar mistake" 😉

Reference: https://unix.stackexchange.com/questions/27708/is-there-a-difference-between-setting-an-environment-variable-to-the-empty-strin

## TODOs after this PR

- Unify the default value formatting: https://github.com/open-telemetry/opentelemetry-specification/pull/2045#discussion_r734251696  and https://github.com/open-telemetry/opentelemetry-specification/pull/2045#discussion_r735350485
- Enhance the "no limit" description: https://github.com/open-telemetry/opentelemetry-specification/pull/2045#discussion_r734092202
- Propose `-1` to set "no limit" https://github.com/open-telemetry/opentelemetry-specification/pull/2045#issuecomment-949713265


pellared:
@tigrannajaryan Thanks for your feedback

> Before we accept this change the entire list of existing env variables must be scanned to verify if any of these variables have an expectation that the empty variable can have a particular meaning compared to the unset value.

That is one of the main reasons why I propose the `SHOULD` to allow exceptions and rather use the spec as a recommendation.

As an example `OTEL_SERVICE_NAME=""` I am not aware of any SDK that would make set this as an empty name. At least Go and .NET SDK will treat it in the same way as unset.

❗  Take also into consideration that for example on Windows it is (almost?) NOT possible to set an empty env var. Not in the command prompt. not via PowerShell, not via GUI. Maybe it is possible via Windows Registry or WinAPI, but I am not sure and it would be VERY inconvenient.

tigrannajaryan:
>As an example OTEL_SERVICE_NAME="" I am not aware of any SDK that would make set this as an empty name. At least Go and .NET SDK will treat it in the same way as unset.

Fair enough. I didn't check the actual behavior but it did appear like it could matter. I want to make sure there are no other env variables where it does actually matter. I suppose it is unlikely that any SDK implementations make a difference between unset and empty value but it is certainly technically possible, so better to double check.

> Take also into consideration that for example on Windows it is (almost?) NOT possible to set an empty env var. Not in the command prompt. not via PowerShell, not via GUI. Maybe it is possible via Windows Registry or WinAPI, but I am not sure and it would be VERY inconvenient.

I think this confirms one more time that we need to revise all env variables to make sure there are none where being able to set an empty value is an important use case and then we should accept this PR.

owais:
Luckily neither of the the two fields specify a default so if we make this change, we can also update attribute spec to set the default for these to "infinity". That way it shouldn't be a breaking change?

pellared:
@owais 
I think that the spec is not clear here.

This is how I understand it: the default value of `*LENGTH_LIMIT` is empty which is interpreted as infinite. The reason why there is nothing inside the "default" value is the the "numeric type" does not allow any value other than a nonnegative decimal number. Maybe we should put `-` there instead of nothing and add some legend that `-` means empty or unset value?

mateuszrzeszutek:
> As an example `OTEL_SERVICE_NAME=""` I am not aware of any SDK that would set this with an empty name. At least Go and .NET SDK will treat it in the same way as unset and use `unknown` value.

The Java SDK allows setting an empty name: https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/EnvironmentResource.java#L43
We only check for the `null` value (which means there was no configuration at all), empty string is treated as a valid service name.
I think that probably most (all?) of the string-typed properties used by the SDK/agent treat null and empty values differently.

pellared:
@mateuszrzeszutek Do you think that changing this behavior in Java SDK/AutoInstr could be dangerous or even seen as something unwelcome?

pellared:
The problem is that not all languages/OSes/shells are handling an empty value in the same way.

In .NET (C# etc) setting an empty value unsets the environmental variable (sic!). See: https://docs.microsoft.com/en-us/dotnet/api/system.environment.setenvironmentvariable?view=net-5.0

On Windows setting an empty env var is VERY tricky. I suspect 95% of Windows IT users may have real problems doing it.

For me having a different interpretation of an empty and unset environment variable is a kind of "billion-dollar mistake" 😉 

I would be happy if the specs would at least encourage us to interpret empty and unset environment variables in the same way. However, maybe it makes even sense to require it?

I am mostly worried about a scenario where setting an empty value may be an important scenario. E.g. `OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT= ` in the case of Python SDK.

pellared:
For reference: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#service

specification/sdk-environment-variables.md
```diff
@@ -16,6 +22,7 @@ If an SDK chooses to support an integer-valued environment variable, it SHOULD s
 
 For variables which accept a known value out of a set, i.e., an enum value, SDK implementations MAY support additional values not listed here.
 For variables accepting an enum value, if the user provides a value the SDK does not recognize, the SDK MUST generate a warning and gracefully ignore the setting.
+`"none"` SHOULD be used as the value that repesents an "empty" selection.
```
bogdandrutu, line 25: Why is this change necessary? I understand the others

carlosalberto: This goes along #2052 - that is, to clarify empty/unset AND a no-op behavior.

pellared: @bogdandrutu So far this PR has 8 approves, but if you have some concerns I can move it to a separate PR. However, I feel that creating another PR would just produce more "noise" 😉

tigrannajaryan: I think the wording is a bit unclear. I understand and support the intent (also evidenced in https://github.com/open-telemetry/opentelemetry-specification/pull/2052/) but when I read just this sentence I don't understand it. What is `"empty" selection`?

I think we need to explain that `none` is necessary for the cases when the default value of the option is not an empty string, but empty string is valid value that the user may need to specify for that option. In such case `none` is a special literal that signals to the SDK that the user wishes to set the value of the option to empty string. I am not sure how to best explain this in the most concise way but I don't think the current description explains it well enough.

At least show an example here, that may help.

carlosalberto: @pellared I'd advise to create a separate PR for that (so @tigrannajaryan's feedback can be addressed), which would allow us to merge this PR today 👍

pellared: Line removed https://github.com/open-telemetry/opentelemetry-specification/pull/2045/commits/66b4b52d9eccc577a6e05b0bd2ccfeed51ea72e4

I will create a new PR tomorrow

tigrannajaryan (approved):

specification/sdk-environment-variables.md
tigrannajaryan: Can merge after @bogdandrutu confirms he has no other objections.

tigrannajaryan:
Please also rebase from main to make ready for merging.

bogdandrutu (approved):

carlosalberto (approved):

PR #2046, opened 2021-10-20
Add OpenTracing section to the compliance matrix.

carlosalberto:
As we are looking to stabilize the OpenTracing compatibility section, we want to keep track of the actual features in different SIGs.


spec-compliance-matrix.md
```diff
@@ -204,6 +204,36 @@ Note: Support for environment variables is optional.
...
+| [Tracer](specification/compatibility/opentracing.md#tracer-shim)             |   |    |   |      |    |   |    |   |    |     |
+| Inject SpanContext                                                           |   |    |   |      |    |   |    |   |    |     |
+| Extract SpanContext                                                          |   |    |   |      |    |   |    |   |    |     |
+| [Span](specification/compatibility/opentracing.md#span-shim)                 |   |    |   |      |    |   |    |   |    |     |
+| Get Context                                                                  |   |    |   |      |    |   |    |   |    |     |
+| Get Baggage Item                                                             |   |    |   |      |    |   |    |   |    |     |
+| Set Baggage Item                                                             |   |    |   |      |    |   |    |   |    |     |
+| SetTag                                                                       |   |    |   |      |    |   |    |   |    |     |
+| SetTag with error mapping                                                    |   |    |   |      |    |   |    |   |    |     |
```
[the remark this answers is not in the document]
carlosalberto: I have 'compacted' the matrix with the top level areas, and included a migration guide point as well.

carlosalberto:
@yurishkuro Updated, please review - I'm keeping the `Baggage` support among execution units for now, while we decide whether we discard support for this or not.

spec-compliance-matrix.md
```diff
@@ -296,6 +296,23 @@ Note: Support for environment variables is optional.
...
+| Create OpenTracing Shim                                                      |   |    |   |      |    |   |    |   |    |     |
+| Create OpenTracing Shim with custom propagators                              |   |    |   |      |    |   |    |   |    |     |
+| [Tracer](specification/compatibility/opentracing.md#tracer-shim)             |   |    |   |      |    |   |    |   |    |     |
+| [Span](specification/compatibility/opentracing.md#span-shim)                 |   |    |   |      |    |   |    |   |    |     |
+| [SpanContext](specification/compatibility/opentracing.md#spancontext-shim)   |   |    |   |      |    |   |    |   |    |     |
+| SpanContext changes visible among different execution units                  |   |    |   |      |    |   |    |   |    |     |
+| [ScopeManager](specification/compatibility/opentracing.md#scopemanager-shim) |   |    |   |      |    |   |    |   |    |     |
+| ScopeManager caching Shim objects in Context                                 |   |    |   |      |    |   |    |   |    |     |
+| Semantic convention mapping for attributes/events                            |   |    |   |      |    |   |    |   |    |     |
```
yurishkuro, line 313: This should not be language-specific imo.

Are we saying that shims are expected to perform runtime conversion from OpenTracing naming convention to OpenTelemetry (when they are different)? Personally, I would prefer not making it a requirement, let the backends deal with it.

spec-compliance-matrix.md
```diff
@@ -296,6 +296,23 @@ Note: Support for environment variables is optional.
...
+Languages not covered by the OpenTracing project do not need to be listed here, e.g. Erlang.
+
+| Feature                                                                      |Go |Java|JS |Python|Ruby|PHP|Rust|C++|.NET|Swift|
+|------------------------------------------------------------------------------|---|----|---|------|----|---|----|---|----|-----|
+| Create OpenTracing Shim                                                      |   |    |   |      |    |   |    |   |    |     |
+| Create OpenTracing Shim with custom propagators                              |   |    |   |      |    |   |    |   |    |     |
+| [Tracer](specification/compatibility/opentracing.md#tracer-shim)             |   |    |   |      |    |   |    |   |    |     |
+| [Span](specification/compatibility/opentracing.md#span-shim)                 |   |    |   |      |    |   |    |   |    |     |
+| [SpanContext](specification/compatibility/opentracing.md#spancontext-shim)   |   |    |   |      |    |   |    |   |    |     |
+| SpanContext changes visible among different execution units                  |   |    |   |      |    |   |    |   |    |     |
```
yurishkuro, line 310: a) the wording is quite vague, what are "execution units"?
b) do we actually want this as a requirement? Is this exclusively related to Baggage?

spec-compliance-matrix.md
```diff
@@ -296,6 +296,23 @@ Note: Support for environment variables is optional.
...
 | TBD                                                                            |          |    |   |    |             |      |        |     |      |     |      |       |
 
+## OpenTracing Compatibility
+
+Languages not covered by the OpenTracing project do not need to be listed here, e.g. Erlang.
+
+| Feature                                                                      |Go |Java|JS |Python|Ruby|PHP|Rust|C++|.NET|Swift|
+|------------------------------------------------------------------------------|---|----|---|------|----|---|----|---|----|-----|
+| Create OpenTracing Shim                                                      |   |    |   |      |    |   |    |   |    |     |
+| Create OpenTracing Shim with custom propagators                              |   |    |   |      |    |   |    |   |    |     |
```
yurishkuro, line 306: is this needed? 
a) OTEL SDKs already support custom propagators
b) I view OpenTracing compatibility as an intersection of OpenTracing API and OTEL API, not OTEL SDK

bhs:
ping @carlosalberto on this (and @yurishkuro's comments)... this PR is on the critical path to formally archiving the OpenTracing project within CNCF so would be good to keep it moving, though obviously not P0 or anything. :)

spec-compliance-matrix.md
carlosalberto: For a), this is because we want to include threads but also light thread (i.e. Python coroutines). Happy to use a different word (or even explicitly say "threads or coroutines").

For b), it's the requirement to keep the `SpanContext` + `Baggage` invariant. Will change the wording (and we still need to discuss whether we keep support for baggage or not ;) )

spec-compliance-matrix.md
carlosalberto: a) This is for _not_ relying on the global propagators at all, but simply use separate `TextMap` propagators for OpenTracing's `TextMap` and `HttpHeader` formats (if the user specifies them). I think that needs a re-wording as well.

b) The Propagators AP in OpenTelemetry is not dependent on the SDK :)

spec-compliance-matrix.md
carlosalberto: That's an interesting one. I've been dealing with it back and forth. Observe that at this moment `Span` error mapping is **all** the mapping being done.

One curious case would be a codebase that is using both the OT Shim and OTel directly, performing gradual adoption - errors, for the same code base, would be reported in some parts as `error` attributes, and in other ones as `Status`.

Bottom line, this mapping will be to be done also in the backend (maybe also through a Collector processor?). So I'm fine either way. Any opinion @tedsuo ?

jmacd: I would not try to translate semantic conventions, personally, but the error state feels worth handling correctly.

carlosalberto: @jmacd agreed, we could go that path, only handling the error case, and not handling semantic conventions at all. WDYT @yurishkuro ?

spec-compliance-matrix.md
yurishkuro: > threads or coroutines

+ 1

spec-compliance-matrix.md
yurishkuro: +1 to only spec error handling

carlosalberto:
@yurishkuro Applied feedback. The only remaining bit is the one regarding custom propagators - which is useful in case users want to use a different propagator for OpenTracing's `HTTP_HEADERS` and `TEXT_MAP` (by default we use the global OTel propagators for **both** types).

spec-compliance-matrix.md
yurishkuro: Got it. I agree it needs rewording. Actually, I think all these entries should be linked to sections in opentracing.md, similar to how you already link to Tracer/Span/etc. Otherwise nobody knows what we mean by these entries, like "custom propagators" or "Error mapping".

carlosalberto: Changed as actually this is also included in the `Create Tracer` section (which is very small anyway).

yurishkuro (approved):

carlosalberto:
@yurishkuro Please review. Removed some parts (such as the `Create Tracer with custom propagators`) as they are already included in features that MUST be supported in their respective sections.

Left the error mapping one without a link as it is mentioned in both sections. It can probably be also removed though, although I'd personally keep that one in order to signal implementers of its importance ;)

tedsuo (approved):
Thanks for picking this up Carlos.

Issue #1524, opened 2021-03-09
Clarify TraceIdRatioBased Sampler description


MrAlias:
[From the specification](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.0.1/specification/trace/sdk.md#traceidratiobased):

> Description MUST be `TraceIdRatioBased{0.000100}`

Does this mean this sampler should always return the static string `"TraceIdRatioBased{0.000100}"` even if the ratio is not `0.000100`?

Or should it return `"TraceIdRatioBased{$RATIO}"` with `$RATIO` replaced with a decimal representation of the sampling ratio? If so, what precision should this ratio be?

MrAlias:
Based on what SIGs are doing (especially the ones that have a released a 1.0.0 version) it would follow the specification needs to be updated to reflect the description should be dynamic. E.g.

```
Description MUST return a string of the form `"TraceIdRatioBased{RATIO}"` with `RATIO` replaced with the Sampler instances' trace sampling ratio represented as a decimal number.
The precision of the number SHOULD follow language standards and SHOULD be high enough to identify when Samplers have different ratios.
For example, if a TraceIdRatioBased Sampler had a sampling ratio of 1 to every 10,000 spans it could return `"TraceIdRatioBased{0.000100}"` as its description.
```

arminru:
I'm glad no one interpreted it as "it has to be this constant string regardless of the actual ratio" 😄 

Thanks for digging up the status-quo and presenting it here, Tyler!
Your proposal sounds good to me. Could you please create a PR for that?

MrAlias:
> Your proposal sounds good to me. Could you please create a PR for that?

Sure thing :+1:

PR #2047, opened 2021-10-20
Probability sampling in `tracestate` specification

### LINKED ISSUE #2224: Support for probability sampling at-a-glance

Probability sampling PR #2047 is waiting for reviews. I would like to help the community support this proposal.

#2047 adds support for building Span-to-Metrics pipelines; it is narrowly focused on how we count spans that have been probabilistically sampled. It does not attempt to solve all challenges for OpenTelemetry trace users that want sampling, so it may be a source of confusion that so much effort has been spent to specify a partial solution.

**Why has this been proposed?**

[The current OpenTelemetry specification has a "TODO" written in it, along with a "WARNING". The TODO and the WARNING were written because the `TraceIdRatioBased` sampler specification is incomplete. The `TraceIdRatioBased` sampler was envisioned as a solution for consistent sampling, but it never delivered that.](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#traceidratiobased)

We have a narrowly focused PR that stops short of fixing the TODO. Instead, it introduces new Sampler definitions that could in the future be swapped in to complete the TODO. The next step for the Sampling SIG is to work on configurable sampling and remote-configurable sampling. To that end, #2047 is only a building block that delivers a fix for the TODO/WARNING. #2047 is a technical replacement for `TraceIdRatioBased`, `ParentBased`, `AlwaysOn`, and `AlwaysOff` samplers. These four samplers meet the same requirements and support Span-to-Metrics.

**Where is the user guide for Sampling?**

#2047 is a specification to enable interoperability between Samplers and Span-to-metrics systems. It is no more a user-guide than the existing SDK specification of the built-in Samplers is. Although the lack of a user guide can be seen as a failing, the OTel specification has already failed to explain how it wants users to configure sampling. See also #1597 and #2113. 

**In a nutshell, how does this proposal work?**

Today's W3C trace context does not specify that any of the 128 bits in a TraceId are true, uniform-distributed random bits. Lacking this certainty, the options for correct probability sampling were:

- consider hashing the TraceId; however we found hashing is expensive, and a portable hashing algorithm of sufficient quality is not widely available
- add additional randomness to the context.

#2047 takes the second approach, adds a new field known as "r-value" that carries a small amount of randomness in the OpenTelemetry tracestate. The amount of randomness is tailored to the number of supported sampling probabilities, which are limited to powers-of-two under this proposal. Thus r-value is contained in 6-bits of information.

#2047 uses the term "Adjusted count", which is how much we should add to the span count for each span when received. A field known as "p-value" in the tracestate conveys the context's adjusted count. **Adjusted count equals `2**p` under this proposal.** Sampling probability is the inverse of adjusted count (i.e., `probability == 2**-p`).

By including r-value and p-value in the tracestate, these two values automatically propagate through the context and are recorded on every span. Because tracestate is automatically recorded, typically only a root Sampler needs to be changed to add Span-to-metrics support to a system. Trace consumers only need to recognize p-value to correctly count spans.
 
**What alternatives have been considered?**

As it stands, the cost to users who enable Span-to-metrics is not nearly as low as it could be. The options should be evaluated in terms of how many bytes are added to each context when sampled vs. not sampled. The current proposal requires r-value for both sampled and unsampled contexts.  While p-value is only required for sampled contexts, this hardly matters given that r-value forces the introduction of `tracestate` for all contexts.

Restricting the choice of sampling probability to powers of two seems unnecessarily restrictive. An alternative, while more expensive, would eliminate the use of a base-2 logarithm so that adjusted count could be any integer in [1, 2**62], for example. 

**How can this be improved, working with W3C?**

The W3C `traceparent` has unused trace flag bits that could be used without increasing the size of a context or requiring alternative header syntax. [Suppose that `traceparent` had an option to specify at least 62 of the 128 TraceId bits be true random bits](https://github.com/w3c/trace-context/issues/467); "r-value" can be calculated from these TraceIDs, in which case `tracestate` is only needed for p-value--meaning, r-value would go away and tracestate would only be needed when sampled.

**How can this be extended to support non-power of two probabilities?**

[If a proposal to change W3C `traceparent` were adopted](https://github.com/w3c/trace-context/issues/467), so that the r-value of #2047 were calculated from TraceID instead of encoded in the tracestate, then other schemes can be identified with consistent sampling behavior that could be mixed in the future.

Consider a hypothetical "c-value", an alternative to p-value that directly states the adjusted count. Probability equals `1/c`. This can be consistently computed from 62 random bits so that the two schemes are equivalent at power-of-two probabilities. In #2047, r-value equals the number of leading bits in the 62-bit unsigned integer `D` and the sampling equation is `sampled <=> p <= LeadingZeros(D)`. The alternative c-value defines a 62-bit threshold equal to `2**62 / c` with the sampling equation `sampled <=> D < 2**62 / c`.  (These schemes are equivalent when `c = 2**p`, I'm leaving out details about zero adjusted count.)

While this "c-value" proposal leads to a more human-readable tracestate, it is also more difficult to specify and currently substantially more expensive to transmit in the unsampled case. Considering the effort and reward, it makes sense to leave this option until after W3C supports built-in randomness. At that point, it will make sense to use either p-value with power-of-two probabilities or c-value for non-power-of-two, integer-reciprocal probabilities.

**Conclusions?**

With the above in mind (i.e., potential `traceparent` randomness, potential non-power-of-two support),#2047 should be seen as a compromise to let us demonstrate demand for Span-to-Metrics pipelines to use as evidence for adding `traceparent` randomness. Power-of-two probability sampling is relatively easy to specify and implement, at the same time it leaves a low commitment behind as trace consumers will only need to recognize p-value going forward (i.e., not r-value).  #2047 asks us to agree that p-value will always be a compact way to represent power-of-two adjusted counts, whether or not we admit the hypothetical c-value or other schemes in the future.

jmacd:
## Changes

Specifies how to interpret and use p-value and r-value for consistent probability sampling decisions, including two optional samplers named `ConsistentProbabilityBased` and `ConsistentParentProbabilityBased`.

Does not modify existing built-in Sampler definitions.

This is ready for general review.

Implemented in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/1379

OTEPs 
https://github.com/open-telemetry/oteps/blob/main/text/trace/0168-sampling-propagation.md
https://github.com/open-telemetry/oteps/blob/main/text/trace/0170-sampling-probability.md

Part of #1414 
Part of #1412 
Part of #1524 
Part of #570 
Another take on #1899 
Fixes #2224
Part of #2179

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,8 @@ release.
   ([#2191](https://github.com/open-telemetry/opentelemetry-specification/pull/2191))
 - Add `device.manufacturer` to describe mobile device manufacturers.
   ([2100](https://github.com/open-telemetry/opentelemetry-specification/pull/2100))
+- Add JVM memory runtime semantic
+  conventions. ([#2272](https://github.com/open-telemetry/opentelemetry-specification/pull/2272))
 
 ### Compatibility
 
--- a/specification/metrics/semantic_conventions/runtime-environment-metrics.md
+++ b/specification/metrics/semantic_conventions/runtime-environment-metrics.md
@@ -55,5 +55,20 @@ consider, for example pthreads vs green thread implementations.
 
 **Description:** Java Virtual Machine (JVM) metrics captured under `process.runtime.jvm.`
 
-| Name | Description | Units | Instrument Type | Value Type | Attribute Key | Attribute Values |
-|------|-------------|-------|-----------------|------------|---------------|------------------|
+All JVM metric attributes are required unless otherwise indicated.
+
+| Name                                 | Description                         | Unit  | Unit ([UCUM](README.md#instrument-units)) | Instrument Type            | Value Type | Attribute Key | Attribute Values      |
+|--------------------------------------|-------------------------------------|-------|-------------------------------------------|----------------------------|------------|---------------|-----------------------|
+| process.runtime.jvm.memory.usage     | Measure of memory used              | Bytes | `By`                                      | Asynchronous UpDownCounter | Int64      | type          | `"heap"`, `"nonheap"` |
+|                                      |                                     |       |                                           |                            |            | pool          | Name of pool [1]      |
+| process.runtime.jvm.memory.init      | Measure of initial memory requested | Bytes | `By`                                      | Asynchronous UpDownCounter | Int64      | type          | `"heap"`, `"nonheap"` |
+|                                      |                                     |       |                                           |                            |            | pool          | Name of pool [1]      |
+| process.runtime.jvm.memory.committed | Measure of memory committed         | Bytes | `By`                                      | Asynchronous UpDownCounter | Int64      | type          | `"heap"`, `"nonheap"` |
+|                                      |                                     |       |                                           |                            |            | pool          | Name of pool [1]      |
+| process.runtime.jvm.memory.max       | Measure of max obtainable memory    | Bytes | `By`                                      | Asynchronous UpDownCounter | Int64      | type          | `"heap"`, `"nonheap"` |
+|                                      |                                     |       |                                           |                            |            | pool          | Name of pool [1]      |
+
+**[1]**: Pool names are generally obtained
+via [MemoryPoolMXBean#getName()](https://docs.oracle.com/en/java/javase/11/docs/api/java.management/java/lang/management/MemoryPoolMXBean.html#getName())
+. Examples include `G1 Old Gen`, `G1 Eden space`, `G1 Survivor Space`
+, `Metaspace`, etc.


specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,352 @@
...
+| 1       | 1/2                | 2              |
+| 2       | 1/4                | 4              |
+| ...     | ...                | ...            |
+| N       | 2**-N              | 2**N           |
+| ...     | ...                | ...            |
+| 61      | 2**-61             | 2**61          |
+| 62      | 2**-62             | 2**62          |
+| 63      | 0                  | 0              |
+
+##### Requirement 1
```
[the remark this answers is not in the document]
jmacd: Got it, will fix.

jmacd:
@open-telemetry/specs-trace-approvers the two optional samplers specified here are implemented by this PR in opentelemetry-go-contrib: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/1379

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,619 @@
...
+decision or interpret adjusted count.
+
+#### R-value
+
+R-value is set in the `tracestate` by the Sampler at the root of the
+trace, in order to support consistent probability sampling.  When the
+value is omitted or not present, child spans in the trace are not able
+to participate in consistent probability sampling.
+
+R-value determines which sampling probabilities and will not sample
```
yurishkuro, line 287: "determines which sampling probabilities" - incomplete sentence?

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,619 @@
...
+as the input.  This scheme is limited to power-of-two sampling
+probabilities, as follows.
+
+1. Express the sampling probability as `2**-s`. For example, 25%
+   equals `2**-2` with `s=2`
+2. Count `r`, the number of consecutive zero bits in the input string
+3. If `s <= r`, select the item with adjusted count `2**s`.
+
+This algorithm is the basis of the consistent probability sampling
+approach used in OpenTelemetry, defined in greater detail below.
```
yurishkuro, line 129: > defined in greater detail below

do you mean some specific section?

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,619 @@
...
+
+For non-probability sampling, adjusted count is unknown.
+
+Zero adjusted count is defined in a way that supports composition of
+probability and non-probability sampling.  Zero is assigned as the
+adjusted count when a probability sampler does not select a span.
+
+#### Power-of-two random sampling
+
+A simple sampling scheme can be implemented using a random bit string
```
yurishkuro, line 119: Suggestion: give this string a variable name and refer to it later, e.g. `using a random bit string B as the input".

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,619 @@
...
+as the input.  This scheme is limited to power-of-two sampling
+probabilities, as follows.
+
+1. Express the sampling probability as `2**-s`. For example, 25%
+   equals `2**-2` with `s=2`
+2. Count `r`, the number of consecutive zero bits in the input string
```
yurishkuro, line 125: Q: is the bit string really relevant here? It seems like a (possible) implementation detail, couldn't we explain the actual methodology without it? I.e. I would prefer to separate math definition from algorithmic implementation.

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,619 @@
...
+effectively excluding it from the population.  This is not defined as
+a probability sampler with zero probability, because these spans are
+effectively unrepresented.
+
+#### Non-probability sampler
+
+A non-probability sampler is a Sampler that makes its decisions not
+based on chance, but instead uses arbitrary logic and internal state.
+
+## Probability sampling
```
yurishkuro, line 188: should this be titled Consistent Probability Sampling?

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,619 @@
...
+
+Both fields are propagated via the OpenTelemetry `tracestate` under
+the `ot` vendor tag using the rules for [tracestate
+handling](tracestate-handling.md).  Both fields are represented as
+unsigned integers requiring at most 6 bits of information.  An
+invariant will be stated that connects the `sampled` trace flag found
+in `traceparent` context to the r-value and p-value found in
+`tracestate` context.
+
+### Conformance
```
yurishkuro, line 204: I would expect the text to first explain how p/r are actually used in the algorithm & sampling, before going into conformance etc. We must assume this is the first time a person reads this material.

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,619 @@
...
+
+R-value is set in the `tracestate` by the Sampler at the root of the
+trace, in order to support consistent probability sampling.  When the
+value is omitted or not present, child spans in the trace are not able
+to participate in consistent probability sampling.
+
+R-value determines which sampling probabilities and will not sample
+for spans of a given trace, as follows:
+
+| r-value          | Probability of r-value   | Implied sampling probabilities |
```
yurishkuro, line 290: I recommend explaining "Probability of r-value" column - why is it here?

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,619 @@
...
+produce a valid `tracestate`.  In the case where it is used in a
+non-root context, it is required to validate the incoming `tracestate`
+and to produce a valid `tracestate` for the outgoing context.
+
+The `ConsistentProbabilityBased` sampler is required to support
+probabilities that are not exact powers of two.  To do so,
+implementations are required to select between the nearest powers of
+two probabilistically.  For example, 5% sampling can be achieved by
+selecting 1/16 sampling 60% of the time and 1/32 sampling 40% of the
+time.
```
PeterF778, line 350: There's a potential trouble here - a consistent probability sampler is defined as one "that supports independent sampling decisions at each span in a trace while maintaining that traces will be complete with probability equal to the minimum sampling probability across the trace". Now, let's assume that we have multiple probability samplers configured with 5% sampling rate, handling different branches of the trace, and that 5% is actually the minimum sampling probability across the trace. If those samplers select their nearest powers of two probabilistically and independently, completeness of the trace will be less likely than 5%. This can be "fixed" by a number of ways, none of them seems ideal though.

oertl: Yes, this needs to be reformulated. Completeness is only guaranteed with a probability that is the largest power of 1/2 smaller than or equal to the mimum sampling rate across the trace.

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,619 @@
...
+and r-value are independent settings, each can be meaningfully set
+without the other present.
+
+#### Sampled flag
+
+Probability sampling uses additional information to enable consistent
+decision making and to record the adjusted count of sampled spans.
+When both values are defined and in the specified range, the invariant
+between r-value and p-value and the `sampled` trace flag states that
+`sampled` is equivalent to the expression `p <= r || p == 63`.
```
oertl, line 243: If the parent uses an always-off sampler, which behaves like a consistent probability sampler with probability 0, I would expect that `sampled == FALSE` and `p == 63`. That means an always-off sampler would always break this invariant.

The invariant `(p <= r) == sampled` must always hold except for the case that `sampled == TRUE` while `p == 63` which has a special meaning and leads to zero adjusted counts. Therefore, I think the invariant should rather be
`((p <= r) == sampled) OR (sampled AND (p == 63)) == TRUE`.

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,619 @@
...
+for spans of a given trace, as follows:
+
+| r-value          | Probability of r-value   | Implied sampling probabilities |
+| ---------------- | ------------------------ | ----------------------         |
+| 0                | 1/2                      | 1                              |
+| 1                | 1/4                      | 1/2 and above                  |
+| 2                | 1/8                      | 1/4 and above                  |
+| 3                | 1/16                     | 1/8 and above                  |
+| ...              | ...                      | ...                            |
+| 0 <= r <= 61     | 1/(2**(-r-1))            | 2**(-r) and above              |
```
robbkidd, line 297: Is the inclusion of `1/` a typo here? Combined with a negative exponent, `1/...` produces a dramatically invalid probability. `1/` is not present in the rows for r-value 59-62.

```ruby
> (0..61).map{|r| 2**(-r-1)}.reduce(&:+)
=> (4611686018427387903/4611686018427387904)
> (0..61).map{|r| 2**(-r-1)}.reduce(&:+).to_f
=> 1.0

> (0..61).map{|r| 1/2**(-r-1)}.reduce(&:+)
=> (9223372036854775806/1)
```

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,619 @@
...
+| ...              | ...                      | ...                            |
+| 59               | 2**-60                   | 2**-59 and above               |
+| 60               | 2**-61                   | 2**-60 and above               |
+| 61               | 2**-62                   | 2**-61 and above               |
+| 62               | 2**-62                   | 2**-62 and above               |
```
robbkidd, line 302: Possibly silly question: what's special about r = 62 that the "probably of r-value formula" is computed differently than for 0 <= r <= 61? I assume it _is_ a special case because the range given above in the row for "r" stops at 61, but I'm missing the importance of special handling.

```suggestion
| 62               | 2**-63                   | 2**-62 and above               |
```

jmacd: This is a subtle detail, but stands correct. The last row in the table of r-values does not follow the pattern, so that the probabilities sum to 1. This detail caused confusion during review of the OTEPs (myself included) and I wrote this [section to explain why this detail is correct using a small example](https://github.com/open-telemetry/oteps/blob/main/text/trace/0168-sampling-propagation.md#worked-3-bit-example).

specification/trace/tracestate-probability-sampling.md
jmacd: Fixed.

specification/trace/tracestate-probability-sampling.md
jmacd: Fixed. Except for the final row, the `r=x` has probability 2**-(x+1).

specification/trace/tracestate-probability-sampling.md
jmacd: Since I've use "power of two" to describe negative powers of two throughout, I prefer to phrase this as:

"only ensured at the smallest power of two greater than or equal to the minimum sampling rate across the trace." 

Hope that sounds good.

specification/trace/tracestate-probability-sampling.md
robbkidd: … Ooooh. This is "[snub Zeno, let Atalanta get to the end of the path](https://en.m.wikipedia.org/wiki/Zeno's_paradoxes#Dichotomy_paradox)?" Our series is finite. We'll declare the last item equal to the next-to-last item, making the sum of all the possible probabilities exactly equal 1 instead of approaching 1, and everyone gets to move on with life.

a square showing the series 1/2, 1/4, … 1/64

jmacd: Yes! Another way I've found to think about this is that in a 62-bit random signed integer, there is exactly one "1" and exactly one "0", corresponding with r=61 and r=62, which therefore have the same probabilities.

jmacd:
@yurishkuro Thank you! Your feedback pointed to several rearrangements, which I've now applied.

specification/trace/tracestate-probability-sampling.md
jmacd: I removed this example, found a simpler way to explain how p- and r- relate to simple sampling.

specification/trace/tracestate-probability-sampling.md
jmacd: Fixed. The sections have been reordered.

specification/trace/tracestate-probability-sampling.md
jmacd: It was supposed to read "determines which sampling probabilities WILL and will not ...". I've edited this section.

jmacd:
@PeterF778 I've added the recommendation that we discussed in today's SIG stating that users are recommended to use power-of-two sampling probability for non-root spans, explaining that only at the root will non-power-of-two sampling work w/o incomplete traces.

oertl:
> @PeterF778 I've added the recommendation that we discussed in today's SIG stating that users are recommended to use power-of-two sampling probability for non-root spans, explaining that only at the root will non-power-of-two sampling work w/o incomplete traces.

This is not root-specific.

specification/trace/tracestate-probability-sampling.md
robbkidd: > […] in a 62-bit random signed integer, there is exactly one "1" and exactly one "0", corresponding with r=61 and r=62, which therefore have the same probabilities.

I look forward to understanding this.

jmacd:
@oertl  You are correct. I'll adjust the wording here to say something different about root probability. It's true that a trace can be incomplete, but it can only be missing one span (the root) and this condition is testable, so we have an easy test for completeness.

As discussed in yesterday's SIG there are a bunch of other ways traces can be incomplete and OTel users can't verify completeness, so I consider this acceptable and part of a larger issue.

jmacd:
@oertl and @PeterF778 please see a38e321.

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,707 @@
...
+sample when the non-probability sample does sample, p-value 63 takes
+effect signifying zero adjusted count.
+
+#### List of requirements
+
+##### Requirement: Combining multiple sampling decisions using logical `or`
+
+When multiple samplers are combined using composition, the sampling
+decision is to sample if at least of the combined samplers decies to
+sample.
```
ocelotl, line 480: Missing RFC 2119 keyword

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,707 @@
...
+
+If `r` is not set on the input `tracecontext` and the Span is not a
+root span, `ConsistentProbabilityBased` SHOULD set `r` as if it were a
+root span and MAY warn the user that a potentially inconsistent trace
+is being produced.
+
+##### Requirement: ConsistentProbabilityBased sampler acts like AlwaysOff sampler for probabilities less than 2**-62
+
+If the configured sampling probability is less than `2**-62`, the
+Sampler should round down to zero probability and make the same
```
ocelotl, line 452:
```suggestion
Sampler SHOULD round down to zero probability and make the same
```

ocelotl (changes requested):

ocelotl:
Just for fun, this would be the generated JSON file from the requirements in this PR:

``` json
[
    {
        "id": "Inconsistent p-values are unset",
        "clean id": "inconsistent_p_values_are_unset",
        "content": "Samplers SHOULD unset `p` when the invariant between the `sampled`,\n`p`, and `r` values is violated before using the `tracestate` to make\na sampling decision or interpret adjusted count.",
        "RFC 2119 keyword": "SHOULD"
    },
    {
        "id": "Out-of-range p-values are unset",
        "clean id": "out_of_range_p_values_are_unset",
        "content": "Consumers SHOULD unset `p` from the `tracestate` if the unsigned value\nis greater than 63 before using the `tracestate` to make a sampling\ndecision or interpret adjusted count.",
        "RFC 2119 keyword": "SHOULD"
    },
    {
        "id": "Out-of-range r-values unset both p and r",
        "clean id": "out_of_range_r_values_unset_both_p_and_r",
        "content": "Samplers SHOULD unset both `r` and `p` from the `tracestate` if the\nunsigned value is of `r` is greater than 62 before using the\n`tracestate` to make a sampling decision.",
        "RFC 2119 keyword": "SHOULD"
    },
    {
        "id": "ParentBased API compatibility",
        "clean id": "parentbased_api_compatibility",
        "content": "The `ParentConsistentProbabilityBased` Sampler MUST have the same\nconstructor signature as the built-in `ParentBased` sampler in each\nOpenTelemetry SDK.",
        "RFC 2119 keyword": "MUST"
    },
    {
        "id": "ParentConsistentProbabilityBased does not modify valid tracestate",
        "clean id": "parentconsistentprobabilitybased_does_not_modify_valid_tracestate",
        "content": "The `ParentConsistentProbabilityBased` Sampler MUST NOT modify a\nvalid `tracestate`.",
        "RFC 2119 keyword": "MUST"
    },
    {
        "id": "ParentConsistentProbabilityBased and ParentBased make identical decisions",
        "clean id": "parentconsistentprobabilitybased_and_parentbased_make_identical_decisions",
        "content": "The `ParentConsistentProbabilityBased` Sampler MUST make the same\ndecision specified for the `ParentBased` sampler.",
        "RFC 2119 keyword": "MUST"
    },
    {
        "id": "TraceIdRatioBased API compatibility",
        "clean id": "traceidratiobased_api_compatibility",
        "content": "The `ConsistentProbabilityBased` Sampler MUST have the same\nconstructor signature as the built-in `TraceIdRatioBased` sampler in\neach OpenTelemetry SDK.",
        "RFC 2119 keyword": "MUST"
    },
    {
        "id": "ConsistentProbabilityBased sampler sets r for root span",
        "clean id": "consistentprobabilitybased_sampler_sets_r_for_root_span",
        "content": "The `ConsistentProbabilityBased` Sampler MUST set `r` when it makes a\nroot sampling decision.",
        "RFC 2119 keyword": "MUST"
    },
    {
        "id": "ConsistentProbabilityBased sampler unsets p when not sampling",
        "clean id": "consistentprobabilitybased_sampler_unsets_p_when_not_sampling",
        "content": "The `ConsistentProbabilityBased` Sampler MUST unset `p` from the\n`tracestate` when it decides not to sample.",
        "RFC 2119 keyword": "MUST"
    },
    {
        "id": "ConsistentProbabilityBased sampler sets p when sampling",
        "clean id": "consistentprobabilitybased_sampler_sets_p_when_sampling",
        "content": "The `ConsistentProbabilityBased` Sampler MUST set `p` when it decides\nto sample.",
        "RFC 2119 keyword": "MUST"
    },
    {
        "id": "ConsistentProbabilityBased sampler sets p using an unbiased algorithm",
        "clean id": "consistentprobabilitybased_sampler_sets_p_using_an_unbiased_algorithm",
        "content": "The `ConsistentProbabilityBased` Sampler MUST set `p` so that the\nadjusted count interpreted from the `tracestate` is an unbiased\nestimate of the number of representative spans in the population.",
        "RFC 2119 keyword": "MUST"
    },
    {
        "id": "ConsistentProbabilityBased sampler sets r for non-root span",
        "clean id": "consistentprobabilitybased_sampler_sets_r_for_non_root_span",
        "content": "If `r` is not set on the input `tracecontext` and the Span is not a\nroot span, `ConsistentProbabilityBased` SHOULD set `r` as if it were a\nroot span and MAY warn the user that a potentially inconsistent trace\nis being produced.",
        "RFC 2119 keyword": "SHOULD"
    },
    {
        "id": "ConsistentProbabilityBased sampler acts like AlwaysOff sampler for probabilities less than 2**-62",
        "clean id": "consistentprobabilitybased_sampler_acts_like_alwaysoff_sampler_for_probabilities_less_than_2___62",
        "content": "If the configured sampling probability is less than `2**-62`, the\nSampler should round down to zero probability and make the same\nsampling decision as the builtin `AlwaysOff` sampler would.",
        "RFC 2119 keyword": null
    },
    {
        "id": "Combining multiple sampling decisions using logical `or`",
        "clean id": "combining_multiple_sampling_decisions_using_logical__or_",
        "content": "When multiple samplers are combined using composition, the sampling\ndecision is to sample if at least of the combined samplers decies to\nsample.",
        "RFC 2119 keyword": null
    },
    {
        "id": "Combine multiple consistent probability samplers using the minimum p-value",
        "clean id": "combine_multiple_consistent_probability_samplers_using_the_minimum_p_value",
        "content": "When combining Sampler decisions for multiple consistent probability\nSamplers and at least one decides to sample, the minimum of the \"yes\"\ndecision `p` values MUST be set in the `tracestate`.",
        "RFC 2119 keyword": "MUST"
    },
    {
        "id": "Unset p when multiple consistent probability samplers decide not to sample",
        "clean id": "unset_p_when_multiple_consistent_probability_samplers_decide_not_to_sample",
        "content": "When combining Sampler decisions for multiple consistent probability\nSamplers and none decides to sample, p-value MUST be unset in the\n`tracestate`.",
        "RFC 2119 keyword": "MUST"
    },
    {
        "id": "Use probability sampler p-value when its decision to sample is combined with non-probability samplers",
        "clean id": "use_probability_sampler_p_value_when_its_decision_to_sample_is_combined_with_non_probability_samplers",
        "content": "When combining Sampler decisions for a consistent probability Sampler\nand a non-probability Sampler, and the probability Sampler decides to\nsample, its p-value MUST be set in the `tracestate` regardless of the\nnon-probability Sampler decision.",
        "RFC 2119 keyword": "MUST"
    },
    {
        "id": "Use p-value 63 when a probability sampler decision not to sample when combined with non-probability sampler decision to sample",
        "clean id": "use_p_value_63_when_a_probability_sampler_decision_not_to_sample_when_combined_with_non_probability_sampler_decision_to_sample",
        "content": "When combining Sampler decisions for a consistent probability Sampler\nand a non-probability Sampler, and the probability Sampler decides not\nto sample but the non-probability does sample, p-value 63 MUST be set\nin the `tracestate`.",
        "RFC 2119 keyword": "MUST"
    },
    {
        "id": "Pass 15 non-power-of-two statistical tests",
        "clean id": "pass_15_non_power_of_two_statistical_tests",
        "content": "For the test with 20 trials and 1 million spans each, the test MUST\ndemonstrate a random number generator seed such that the ChiSquared\ntest statistic is below 0.102587 exactly 1 out of 20 times.",
        "RFC 2119 keyword": "MUST"
    },
    {
        "id": "Pass 5 power-of-two statistical tests",
        "clean id": "pass_5_power_of_two_statistical_tests",
        "content": "For the teset with 20 trials and 1 million spans each, the test MUST\ndemonstrate a random number generator seed such that the ChiSquared\ntest statistic is below 0.003932 exactly 1 out of 20 times.",
        "RFC 2119 keyword": "MUST"
    }
]
```

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,707 @@
...
+consistent probability samplers in the same trace.  One way to avoid
+producing incomplete traces is to use parent-based samplers except for
+root spans.
+
+There is a simple test for trace incompleteness, but it is a one-way
+test and does not detect when child spans are uncollected.  One way to
+avoid producing incomplete traces is to avoid configuring
+non-power-of-two sampling probabilities for non-root spans, because
+completeness is not guaranteed for non-power-of-two sampling
+probabilities.
```
kalyanaj, line 523: I am looking to understand one thing better: In that case, are there situations where it will still be useful to support non-power-of-two sampling probabilities? Would it be an option to consider a requirement/constraint that samplers should support only power-of-two sampling probabilities (to avoid this situation)?

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,707 @@
...
+
+Incomplete traces that result from sampling configuration (i.e., on
+purpose) are known as partial traces.  An important subset of the
+partial traces are those which are also complete subtraces.  A
+complete subtrace is defined at span S when:
+
+- S is not a root span
+- S's parent was not collected
+- S either has no children or all the children of S are also complete
+subtraces.
```
kalyanaj, line 180: I think I understand the intuition behind this, but I have a couple of questions to understand this formal definition better. Let's take the example of the following spans and their sampling rates: 

A (6.25%) -> B (12.5%) -> C (25%) -> D (25%) -> E (25%) -> F (6.25%).  

When we have a partial trace that only has spans from C, D, and E (due to their higher sampling rates ):
1) We would regard the subset of C, D, E to be a complete subtrace at span C, correct?
2) When we apply the above definition to recursively evaluate it for the children of "C", then the formal definition has  "S's parent was not collected" which will be false for D, correct? Since the parent-not-being collected seems like a valid condition only for the "root of that sub-trace", wondering if we should change the formal definition to reflect it.

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,707 @@
...
+### Completeness guarantee
+
+This specification defines consistent sampling for power-of-two
+sampling probabilities.  When a sampler is configured with a
+non-power-of-two sampling probability, the sampler will
+probabilistically choose between the nearest powers of two
+
+When a single consistent probability sampler is used at the root of a
+trace, the resulting traces are always complete (ignoring collection
+errors).  This property holds even for non-power-of-two sampling
```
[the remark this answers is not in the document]
jmacd: Yes, thank you. I've added "and all other spans use a parent-based sampler".

specification/trace/tracestate-probability-sampling.md
jmacd: Good question. We could surely specify that consistent probability samplers are restricted to powers of two. One answer is that not every application for trace data cares about completeness. The research by @oertl (https://arxiv.org/pdf/2107.07703.pdf) shows how we can use this data without completeness.

The situation where it may be useful to support a non-power-of-two sampling decisions is the root, where completeness is not broken if all other samplers are parent-based. This sounds contrived, but it is the normal case in historical systems, and as a result the OTel 1.0 tracing specification includes the `TraceIDRatioBased` sampler which is specified to support non-power-of-two probabilities. For the ConsistentProbabilityBased sampler to replace the TraceIDRatioBased sampler, I proposed that we relax the constructor to allow non-powers-of-two. The OTel 1.0 tracing speciication also includes a TODO and a warning against using the TraceIDRatioBased sampler at non-root spans, because it wasn't specified consistently. The consistent probability sampler can replace the TraceIDRatioBased sampler, and this is a replacement warning to be careful configuring this at non-root spans. It was a compromise.

specification/trace/tracestate-probability-sampling.md
jmacd: I made an error in this section, had it completely wrong as you noticed. I do not consider the set of spans C, D, and E to be complete because F is missing.

I've shortened the definition to "A complete subtrace is defined at span S when every descendent span is collected."

specification/trace/tracestate-probability-sampling.md
```diff
@@ -512,8 +508,8 @@ in the `tracestate`.
 As stated in the [completeness guarantee](#completeness-guarantee),
 traces will be possibly incomplete when configuring multiple
 consistent probability samplers in the same trace.  One way to avoid
-producing incomplete traces is to use parent-based samplers except for
-root spans.
+producing incomplete traces is to use parent-based samplers except forr
```
kalyanaj, line 511:
```suggestion
producing incomplete traces is to use parent-based samplers except for 
```

specification/trace/tracestate-probability-sampling.md
```diff
@@ -512,8 +508,8 @@ in the `tracestate`.
...
+oot spans.
```
kalyanaj, line 512:
```suggestion
root spans.
```

specification/trace/tracestate-probability-sampling.md
kalyanaj: Got it, thanks Joshua for the additional context here.

jmacd:
@open-telemetry/specs-trace-approvers This PR has received a round of reviews from subject matter experts and I recommend that we try to get this merged. 

The OTel-Go implementation sits here: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/1379

@dyladan and @oertl will you please post pointers to the current/best implementation of this work in Javascript and Java?

ocelotl (approved):

oertl:
@jmacd 
> @open-telemetry/specs-trace-approvers This PR has received a round of reviews from subject matter experts and I recommend that we try to get this merged.
> 
> The OTel-Go implementation sits here: [open-telemetry/opentelemetry-go-contrib#1379](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/1379)
> 
> @dyladan and @oertl will you please post pointers to the current/best implementation of this work in Javascript and Java?

This is my proof of concept written in Java: https://github.com/dynatrace-research/opentelemetry-sampling-poc. But that needs to be updated according to the latest specification proposal.

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,709 @@
...
+
+##### Requirement: R-value is generated with the correct probabilities
+
+Samplers MUST generate r-values using a randomized scheme that
+produces each value with the specified probability.
+
+### Samplers
+
+#### ParentConsistentProbabilityBased sampler
+
```
oertl, line 370: I think keeping `ParentConsistentProbabilityBased` fully compatible with the current `ParentBased` Sampler is a problem. The delegation to other samplers based on the parent's sampling decision may lead to inconsistencies. For example, when delegating to an Always-on consistent probability sampler, the p-value will be set to 0 (corresponding to 100% sampling probability). This overwrites the p-value of the parent which represents the correct sampling probability. Therefore, I think `ParentConsistentProbabilityBased` must not delegate to other samplers. It should should just reuse the parent's p- and r-values which also leads to the same sampling decision.

oertl:
@jmacd 

Preliminary Java implementations of consistent samplers can be found in this branch:
https://github.com/dynatrace-oss-contrib/opentelemetry-java/tree/consistent-sampling
It also includes a prototype of a new span processor realizing consistent reservoir sampling. This span processor allows to sample a maximum number of spans per period.

specification/trace/tracestate-probability-sampling.md
jmacd: Good catch! You're right that this configuration mechanism can lead to incorrect interpretation. This has to be fixed.

My understanding of the existing ParentBased interface is that a user might use it to configure alternate samplers, which might for example add new span attributes. This is worthy of maintaining, and I hope we can do this without too much complicated new wording. This may require a bit more study, but the following rules seem to do it:

1. The delegated-to Sampler's _input_ tracestate will be stripped of `r` and `p`
2. The delegated-to Sampler's _output_ tracestate will be stripped of `r` and `p` and the original `r` always passes through
3. The delegated-to Sampler can change the sampling decision; if no change, `p` passes through unmodified
4. If the parent's decision was to sample and the delegated-to's decision is not to sample, `p` must be dropped
5. If the parent's decision was not to sample and the delegated-to's decision is to sample, `p` must be set to 63

oertl: I think, this will not fix it. `p` represents the parent sampling probability and allows to determine the adjusted count. According to 3) `p` is only passed through with a certain probability, which requires increasing `p` to get adjusted counts right. 

Consistent sampling requires that the sampling probability of any sampler does not depend on `r`. The sampled flag is given by the parent sampling decision `r >= p` and therefore the sampled flag depends on `r`. Since the parent-based sampler delegates dependent on this flag, its sampling probability will be dependent on `r` which leads to inconsistent sampling. To follow the parent sampling decision in a consistent way, a sampler could just set its `p` equal to the propagated parent `p` as demonstrated by https://github.com/dynatrace-oss-contrib/opentelemetry-java/blob/consistent-sampling/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/ConsistentParentBasedSampler.java.

I am interested in particular use cases of the current ParentBased sampler. I am pretty sure that those can be covered with consistent sampling, but not by delegating to different samplers based on the sampled flag.

jmacd:
@oertl @kalyanaj I have added more recommendations for trace producers to avoid inconsistent configurations. 

In my comment above (https://github.com/open-telemetry/opentelemetry-specification/pull/2047#discussion_r758753139), I've described how I think we can extend the ConsistentParentProbabilityBased sampler to check and constrain its delegating sampler specifically to avoid the inconsistencies that @oertl referred to. I support doing that in a future PR. Since this spec is already marked experimental, I think we can leave that out of this PR, sound good?

specification/trace/tracestate-probability-sampling.md
jmacd: You're right!  A correction is required, and I'm not sure what we should do with this. The correction needed was described in OTEP 170 while describing the [Dapper inflationary sampling approach](https://github.com/open-telemetry/oteps/blob/main/text/trace/0170-sampling-probability.md#dappers-inflationary-sampler). I feel that we could specify how to do this correctly but that we should not.

I have included warnings against improper configuration, so the question is whether the risk of misconfiguration outweighs the cost of not having those interfaces. AFAICT the reason for `ParentBased` to support conditional delegation is not to provide a means of changing the sampling decision, but to provide a way to change the attribute set on the resulting span. Choosing a different sampler delegate for unsampled contexts appears useful to limit the cost of sampling -- there's no reason to synthesize attributes for unsampled contexts.

To summarize, the `ParentBased` sampler supports conditional delegation, and I see reasons for this to exist, but this facility also makes it easy to accidentally subvert our adjusted count logic. I would like to ask @yurishkuro or any @open-telemetry/specs-trace-approvers -- can anyone remember why `ParentBased` has this facility to choose the delegating sampler based on the parent's sampled flag? It's risky.

A "safer" version of the `ConsistentParentProbabilityBased` sampler could break from the `ParentBased` interface, allow configuring the delegating sampler for remote and non-remote contexts, but not based on the `sampled` flag. Is the added safety worth the loss of control over performance? If I could choose my own solution to this problem, I would return us to #620 -- the problem of expensive attributes that are conditionally evaluated should be separated from the `Sampler` API with the use of lazy attributes. If we had lazy attributes, there would be no "good" reason for `ParentBased` samplers to support delegation based on the sampled flag.

yurishkuro: > can anyone remember why ParentBased has this facility to choose the delegating sampler based on the parent's sampled flag? It's risky

This PR/discussion notwithstanding, the original motivation for ParentBased was to support consistent head-based sampling, where if you have parent context, you just respect it without any additional sampling decisions, and if there isn't, you delegate to whichever sampler. This was the default behavior supported by most existing SDKs (Zipkin, Jaeger, OC -- in Jaeger it wasn't even configurable, it was built into the Tracer logic).

oertl: > the original motivation for ParentBased was to support consistent head-based sampling, where if you have parent context, you just respect it without any additional sampling decisions, and if there isn't, you delegate to whichever sampler. 

This could also be achieved with a consistent sampler that sets the sampling probability according to the parent's `p` value. Since `p` and `r` will be the same, the consistent sampling condition `r >= p` will lead to the same sampling decision as the parent. In case there is no parent, it could delegate to any other consistent sampler (e.g ConsistentProbabilityBased or ConsistentAlwaysOn).

jmacd: Thank you @yurishkuro. Given what we know, it's starting to look like the OTel-1.0 built-in `ParentBased` can and should be replaced by a more-primitive set of samplers. I want to find a way to avoid blocking this PR, and @oertl's suggestion is the way to go. `ConsistentParentProbabilityBased` only needs one parameter in the current framing, which is the sampler to use when the context is a root.

I'll find another place to discuss the larger issue. 

----

It looks like we ought to have two sampler APIs, one for root spans and one for non-root spans. A built-in sampler named `DelegatingSampler(root, localParent, remoteParent)` takes one root sampler and two non-root samplers and calls the appropriate one. A new non-root `FlagSampler` would simply honor the `sampled` flag and requires zero parameters. The default-parameter `ParentBased` Sampler is equivalent to `DelegatingSampler(root, localParent=FlagSampler(), remoteParent=FlagSampler())`.

Under this formulation, we'd have a zero-parameter non-root `ConsistentParentProbabilitySampler` which first validates the tracestate and then acts like `FlagSampler`. 

Returning to consistency problem that prompted this discussion (https://github.com/open-telemetry/opentelemetry-specification/pull/2047#discussion_r757691632), the problem stems from the `DelegatingSampler` aspect of `ParentBased`, and to enforce consistent adjusted counts will require consistency checking whenever non-root sampler delegation happens. We can say something like that for non-root sampler decisions, the delegating sampler MUST prevent bias (i.e., ensure consistency) and spell out the rules for doing so. I think they amount to preventing `r` and `p` values from changing except to permit explicit change of the sampled flag. If sampling changes from on (parent) to off (child), the effective behavior should be equivalent to `AlwaysOff`. If sampling changes from off (parent) to on (child), the effective behavior is spelled out in the composition rules: the context takes on zero adjusted count.

jmacd: I've changed the requirements as @oertl suggested.

jmacd: Leaving this comment thread open, now referred to by https://github.com/open-telemetry/opentelemetry-specification/pull/2047#discussion_r759697500.

jmacd:
@open-telemetry/specs-trace-approvers There are no outstanding comments on this PR and this design has already been through the OTEPs process. If there are reasons not to merge this work, please speak up, otherwise please approve.

oertl (approved):

carlosalberto (approved):

carlosalberto:
Sorry for the late review. LGTM!

As a follow up I'd love to have a few clarifications for probabilistic sampling newcomers (like me) - I will create an issue after this PR is merged.

Namely:

* Even though there are value tables, I feel actual examples of P/R values in a trace/spans would be a great thing to have.
* I feel there are some places where the special case of the R-value being `63` could be more concise/clear, e.g. `The adjusted count of spans sampled by a non-probability sampler is unknown.` and then `A special p-value of 63 is defined to mean zero adjusted count...`.
* In the R-value section there is a value table with “probability of r” and “implied sampling probabilities” columns, and the related values may not be super clear during a first read.

jmacd:
Thanks @carlosalberto 
Please see: 0c2b1cd 318bdd2 5879cfa

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,972 @@
...
+distinct probability values. The set of supported probabilities
+includes the integer powers of two between 1 and 2**-62.
+
+R-value encodes which among the 63 possibilities will consistently
+decide to sample for a given trace.  Specifically, r-value specifies
+the smallest probability that will decide to sample a given trace in
+terms of the corresponding p-value.  For example, a trace with r-value
+0 will sample spans configured for 100% sampling, while r-value 1 will
+sample spans configured for 50% or 100% sampling, and so on through
+r-value 62 which will sample every span.
```
PeterF778, line 247: There's still a corner case of ConsistentAlwaysOffSampler which will not sample even for r-value of 62.

jmacd: Thank you @PeterF778, please see if 7c374c2 helps.

jmacd:
@open-telemetry/specs-trace-approvers I've added a few more examples.

jmacd:
@yurishkuro Because you reviewed the OTEPs on this topic, will you please review the finished specification? The Sampling SIG would like to focus on configurable samplers (a.k.a., span "Views"), and as far as I can tell there's nothing left to do here. After this merges we can lobby the W3C trace context group to add support for embedding r-value in the TraceId using a new trace flag, when there is sufficient randomness, after which point the cost of probability sampling and transmitting adjusted counts will be 0 for untraced contexts.

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,1057 @@
...
+and "p-value" meant to support interoperable Sampler implementations.
+Rules are given for creating, validating, interpreting, and mutating
+these fields in an OpenTelemetry [context](../context/context.md).
+
+Two Samplers are specified for optional inclusion in OpenTelemetry
+tracing SDKs, named `ParentConsistentProbabilitityBased` and
+`ConsistentProbabilityBased`.  These are meant as optional
+replacements for the built-in `ParentBased` and `TraceIdRatioBased`
+Samplers.
+
```
bdarfler, line 108:
```suggestion

### Overview

This document specifies an approach to sampling based on an "r-value" and a 
"p-value". At a very high level the "r-value" is a source of randomness and the 
"p-value" encodes the sampling probability. A span is then sampled if `p <= r`.

For efficient representation, the number of supported sampling probabilities is limited 
to powers-of-two. Rather than encoding the sampling probably directly, the "p-value" 
represents the base-2 exponent of the sampling probability. The true sampling 
probability is then `2**p`.

Since the W3C trace context does not specify that any of the 128 bits in a TraceID are
true uniform-distributed random bits, the "r-value" is introduced as an additional source
of randomness is. The amount of randomness in the "r-value" is then tailored to the 
number of supported sampling probabilities. This allows the r-value to be represented 
in just 6-bits.

Finally, by including the "r-value" and "p-value" in the `tracestate`, these two values 
automatically propagate through the context and are recorded on every span. This 
allows Trace consumers to correctly count spans by recognizing the "p-value" on
a given span.

```

bdarfler (commented):
@jmacd here is my offering of a short overview section. Feel free to adapt as you see fit. I tried to borrow heavily from #2224.

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,1057 @@
...
+The `ConsistentProbabilityBased` Sampler MUST have the same
+constructor signature as the built-in `TraceIdRatioBased` sampler in
+each OpenTelemetry SDK.
+
+##### Requirement: ConsistentProbabilityBased sampler sets r for root span
+
+The `ConsistentProbabilityBased` Sampler MUST set `r` when it makes a
+root sampling decision.
+
+##### Requirement: ConsistentProbabilityBased sampler unsets p when not sampling
```
bogdandrutu, line 531: "Sampling" sounds like the process, should this be "sampled"?

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,1057 @@
...
+
+A consistent probability sampler is a Sampler that supports
+independent sampling decisions at each span in a trace while
+maintaining that traces will be complete with a certain minimum
+probability across the trace.
+
+Consistent probability sampling requires that for any span in a given
+trace, if a Sampler with lesser sampling probability selects the span
+for sampling, then the span would also be selected by a Sampler
+configured with greater sampling probability.
```
bogdandrutu, line 204: is the current `TraceIdRatioBased` not a consistent probability sampler? It feel it is based on your requirements, then the problem is that the `ConsistentProbabilityBased` suggests that the current `TraceIdRatioBased` is not.

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,1057 @@
...
+Incomplete traces that result from sampling configuration (i.e., on
+purpose) are known as partial traces.  An important subset of the
+partial traces are those which are also complete subtraces.  A
+complete subtrace is defined at span S when every descendent span is
+collected.
+
+Since the test for an incompleteness is one-way, it is important to
+know which sampling configurations may lead to incomplete traces.
+Sampling configurations that lead naturally to complete traces and
+complete subtraces are [discussed below](#trace-producer-completeness).
```
bogdandrutu, line 227: OpenCensus had a "number" of child spans, to make this a two way check. Please think if that would help.

specification/trace/tracestate-probability-sampling.md
jmacd: I integrated your text and the other few leading paragraphs, see how this looks?

specification/trace/tracestate-probability-sampling.md
jmacd: Fixed (twice).

specification/trace/tracestate-probability-sampling.md
bogdandrutu: I also have hard time reading this.

specification/trace/tracestate-probability-sampling.md
jmacd: I think we could say it was _intended to be_ the definition given here, but the TODO and WARNING in the current specification are not so easily fixed without making changes at the W3C level. This specification is a deliberate compromise.

specification/trace/tracestate-probability-sampling.md
jmacd: I documented this in OTEP 170:

> One approach,
[lost in the transition from OpenCensus to OpenTelemetry is to count
the number of children of each
span](https://github.com/open-telemetry/opentelemetry-specification/issues/355).

Let's use https://github.com/open-telemetry/opentelemetry-specification/issues/355 for this topic.

specification/trace/tracestate-probability-sampling.md
jmacd: I rewrote this sentence as:

R-value determines which sampling probabilities will decide to sample
or not decide to sample for spans of a given trace, as follows:

bogdandrutu (commented):
Overall these looks great as a starting point. I wish we could merge more the current samplers with the newly proposed once, but maybe we can do that as a followup before marking this as stable and once we have couple of concrete prototypes.

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,1069 @@
...
+The `ParentConsistentProbabilityBased` sampler is meant as an optional
+replacement for the [`ParentBased` Sampler](sdk.md#parentbased). It is
+required to first validate the `tracestate` and then respect the
+`sampled` flag in the W3C traceparent.
+
+##### Requirement: ParentConsistentProbabilityBased API
+
+The `ParentConsistentProbabilityBased` Sampler constructor SHOULD take
+a single Sampler argument, which is the Sampler to use in case the
+`ParentConsistentProbabilityBased` Sampler is called for a root span.
```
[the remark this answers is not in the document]
jmacd: Discussed here: https://github.com/open-telemetry/opentelemetry-specification/pull/2047#discussion_r759697500

I documented this in issue #2179.

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,1069 @@
...
+Probability sampling uses additional information to enable consistent
+decision making and to record the adjusted count of sampled spans.
+When both values are defined and in the specified range, the invariant
+between r-value and p-value and the `sampled` trace flag states that
+`((p <= r) == sampled) OR (sampled AND (p == 63)) == TRUE`.
+
+The invariant between `sampled`, `p`, and `r` only applies when both
+`p` and `r` are present.  When the invariant is violated, the
+`sampled` flag takes precedence and `p` is unset from `tracestate` in
+order to signal unknown adjusted count.
```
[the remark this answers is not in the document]
jmacd: That is correct. There is no contradiction here.

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,1069 @@
...
+tracestate: ot=r:3
+```
+
+This supports consistent probability sampling in child contexts by
+virtue of having an r-value.  P-value is not set, consistent with an
+unsampled context.
+
+### Samplers
+
+#### ParentConsistentProbabilityBased sampler
```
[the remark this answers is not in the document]
jmacd: I described what I hoped we can do in #2179.

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,1069 @@
...
+
+The `ParentConsistentProbabilityBased` sampler is meant as an optional
+replacement for the [`ParentBased` Sampler](sdk.md#parentbased). It is
+required to first validate the `tracestate` and then respect the
+`sampled` flag in the W3C traceparent.
```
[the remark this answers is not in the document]
jmacd: I prefer the proposed `DelegatingSampler` concept in #2179.  I don't think the `TextMapPropagator` interface as specified has access to the `sampled` flag.

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,1069 @@
...
+
+##### Example: Setting R-value for a root span
+
+A new root span is sampled by a consistent probability sampler at 25%.
+A new r-value should be generated (see the appendix for suitable
+methods), in this example r-value 5 is used which happens 1.5625% of
+the time and indicates to sample:
+
+```
+tracestate: ot=r:5;p:2
```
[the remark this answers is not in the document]
jmacd: This was an arbitrary decision, not easily made. In other versions of this proposal I've specified base16 because that's how `traceparent` is specified, but in the specification of `./tracestate-handling.md` it just permits alpha-numeric characters and the choice of number base is not fixed by the format.

specification/trace/tracestate-probability-sampling.md
jmacd: I take it your goal is to keep two primary samplers, which are today TraceIDRatioBased and ParentBased. I agree these are the two important concepts. There have been disagreements over whether the new functionality should be on-by-default or an option; since this is still experimental, my goal was to provide always-on probability samplers, not to entertain the possibility of _turning it off_ as an option.

Simply, I'd rather have two samplers than one sampler with two behaviors as an option. However, if we could agree to have the functionality on by default, which I think means waiting for a W3C trace flag to encode r-value in the TraceID, then it would make sense to roll probability sampling into the built-in samplers.

Also #2179 mentions how `ParentBased` has an API specified that unintentionally breaks probability sampling and proposes a fix. If we accept the proposed fixes, and standardize a W3C randomness flag, the outcome you are looking for will likely happen.

specification/trace/tracestate-probability-sampling.md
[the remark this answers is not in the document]
jmacd: Note this is the basis of saying that you can modify only the root to use probability sampling, all the existing ParentBased samplers will do the right thing by default. It's the ParentBased options that create trouble, see #2179.

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,1069 @@
...
+##### Requirement: ParentConsistentProbabilityBased API
+
+The `ParentConsistentProbabilityBased` Sampler constructor SHOULD take
+a single Sampler argument, which is the Sampler to use in case the
+`ParentConsistentProbabilityBased` Sampler is called for a root span.
+
+##### Requirement: ParentConsistentProbabilityBased does not modify valid tracestate
+
+The `ParentConsistentProbabilityBased` Sampler MUST NOT modify a
+valid `tracestate`.
```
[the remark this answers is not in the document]
jmacd: I believe you would configure a ConsistentProbabilitySampler (i.e., ignore the parent decision) if you want to change p to capture a sub-trace.

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,1069 @@
...
+The probability sampler decision is composed with a non-probability
+sampler that decides to sample the context.  Setting `sampled` when
+the probability sampler has not sampled requires setting `p:63`,
+indicating zero adjusted count.
+
+The resulting context:
+
+```
+traceparent: 00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01
+tracestate: ot=r:3;p:63
```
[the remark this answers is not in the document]
jmacd:
```suggestion
tracestate: ot=r:4;p:63
```

specification/trace/tracestate-probability-sampling.md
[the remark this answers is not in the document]
bogdandrutu: Probably we should do it since otherwise this cannot be put on the wire correct?

jmacd:
@bogdandrutu I'm leaving open the discussions you opened above. I wonder if issues #2224 and #2179 capture enough of the state required to understand these compromises that you would consider approving this in its current form?

specification/trace/tracestate-probability-sampling.md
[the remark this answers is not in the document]
jmacd: Are you saying we should use base64 for `r` and `p`? I am not sure that the characters for code 62 and 63 are currently allowed by our `tracestate-handling.md` spec, and this makes it less readable to humans. I can be convinced this is OK. OTOH, if we ever admit a "directly-stated" adjusted count option such as `c:15` (e.g., discussed in #2224), I would argue it should be decimal because it's meant to be somewhat human-readable.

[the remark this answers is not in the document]
PeterF778: I think both r and p should be encoded as decimals. There's hardly anything to gain with alternate encodings, and readability would suffer.

bogdandrutu (approved):
Happy to approve as long as we make sure we resolve the 2 issues you pointed before stable.

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,1069 @@
...
+Samplers SHOULD unset both `r` and `p` from the `tracestate` if the
+unsigned decimal value of `r` is greater than 62 before using the
+`tracestate` to make a sampling decision.
+
+##### Requirement: R-value is generated with the correct probabilities
+
+Samplers MUST generate r-values using a randomized scheme that
+produces each value with the probabilities equivalent to those
+produced by counting the number of leading 0s in a string of 62 random
+bits.  See the appendix for details.
```
ocelotl, line 420:
```suggestion
bits.
```

That suggestion should be added outside this normative seciton.

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,1069 @@
...
+to sample according to its configured sampling probability.
+
+##### Requirement: ConsistentProbabilityBased sampler records unbiased adjusted counts
+
+The `ConsistentProbabilityBased` Sampler with non-zero probability
+MUST set `p` so that the adjusted count interpreted from the
+`tracestate` is an unbiased estimate of the number of representative
+spans in the population.
+
+A test specification for this requirement is given in the appendix.
```
ocelotl, line 548:
```suggestion
```

specification/trace/tracestate-probability-sampling.md
```diff
@@ -0,0 +1,1069 @@
...
+
+##### Requirement: ConsistentProbabilityBased sampler sets r for non-root span
+
+If `r` is not set on the input `tracecontext` and the Span is not a
+root span, `ConsistentProbabilityBased` SHOULD set `r` as if it were a
+root span and MAY warn the user that a potentially inconsistent trace
```
ocelotl, line 554: Two RFC keywords are present in this normative section, this should be separated in two.

ocelotl (changes requested):

ocelotl (commented):
Changing to comment instead of requesting changes since these are not critical changes needed.

ocelotl (approved):

specification/trace/tracestate-probability-sampling.md
jmacd: I removed this sentence. The reader will have to keep reading. :-)

specification/trace/tracestate-probability-sampling.md
jmacd: Done.

specification/trace/tracestate-probability-sampling.md
jmacd: I changed the MAY into a SHOULD by making this requirement compound, as in "SHOULD do A and B" instead of "SHOULD do A and MAY do B". 

Error handling doesn't really need to be specified, there are guidelines in place. I think this is OK.

ocelotl (approved):

oertl (approved):

jmacd:
@open-telemetry/specs-trace-approvers This meets our minimum criteria for merging. I have been trying to get more stamps of approval, but it's not happening (or not happening quickly). I propose to merge this PR.

codeboten (approved):

reyang (approved):

mtwo:
Looks good, a bit long but that's likely necessary given the topic. I couldn't find an example for what I suspect might be a common scenario: when an edge service uses a probability-based sampler but a downstream service decides to sample when it sees a trace with a sampled flag OR some other condition is satisfied (an error, for example).

I think (?) that this is similar to the "interoperability with TraceIDRatioBased sampler" example, though the second sampler in that case uses different criteria.

mtwo (approved):
Looks good, a bit long but that's likely necessary given the topic. I couldn't find an example for what I suspect might be a common scenario: when an edge service uses a probability-based sampler but a downstream service decides to sample when it sees a trace with a sampled flag OR some other condition is satisfied (an error, for example).

I think (?) that this is similar to the "interoperability with TraceIDRatioBased sampler" example, though the second sampler in that case uses different criteria.

jmacd:
@mtwo There isn't an exact example corresponding to your question. The section "Example: Probability and non-probability sampler in a root context" comes close, and there is a related discussion in #2179. As discussed in the issue, a better Delegating sampler API would be nice, but is not critical to the answer here. Sampler API details aside (it's a bit messy, we could use a v2 API proposal to clear things up--see #2179), the child span is expected to have adjusted count equal to the parent when the span is sampled, and adjusted count 0 when the parent was not sampled but there were errors (`p=63`).

carlosalberto:
We have more than enough reviews and no standing questions. Let's merge this now and address any improvements in future PRs.

Issue #1351, opened 2021-01-19
Add duration unit parsing for environment variables


carlosalberto:
Follow up to #1325 and its related issue #1312 

Add duration unit parsing for environment variables, i.e. formatted strings with an integer followed immediately by a unit string:

- `"ms"` - milliseconds (default if not unit was specified, to not break the current usage)
- `"s"` - seconds
- `"m"` - minutes
- `"h"` - hours
- `"d"` - days

Although this is a really good thing to have, we need to make sure that (at least) the main SIGs implement this before we make this official.

tigrannajaryan:
@carlosalberto can you also specify that the unit is optional and if missing the duration is assumed to be a number in milliseconds for backward compatibility reasons.

carlosalberto:
@tigrannajaryan Done :)

svrnm:
The label removal we applied during triage stopped the `stale` bot from closing this issue, so we close it manually, if this is incorrect please re-open or raise a new issue, cc @mx-psi @jpkrohling

PR #2051, opened 2021-10-21
Remove unit from duration environment variables

pellared:
## Why
From https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#duration

> Any value that represents a duration, for example a timeout, MUST be an integer representing a number of
> milliseconds. The value is non-negative - if a negative value is provided, the SDK MUST generate a warning,
> gracefully ignore the setting and use the default value if it is defined.
> 
> For example, the value `12000` indicates 12000 milliseconds, i.e., 12 seconds.

I think that currently:

1. someone implementing the Go SDK will parse the value using https://pkg.go.dev/time#ParseDuration
2. somebody using an instrumented application may try to add the duration units when setting the env vars 

## What

Remove unit from duration environment variables to make the specification more straightforward


tigrannajaryan:
Yes, found here https://github.com/open-telemetry/opentelemetry-specification/issues/1351

The proposal was to make milliseconds the default and allow for units. We didn't end up adding it to the spec, but I think it is still doable and desirable.

PR #2052, opened 2021-10-22
Add "none" as a possible value to OTEL_PROPAGATORS

pellared:
Per https://github.com/open-telemetry/opentelemetry-specification/pull/2045#discussion_r734090912

## Changes

Add `none` as a possible value for `OTEL_PROPAGATORS` to disable context propagation. Because the default is `tracecontext,baggage`, there is no way to disable signal export via env vars.

The same approach has been specified for `OTEL_TRACES_EXPORTER` and `OTEL_METRICS_EXPORTER` here: #1439

It is also important given the fact that an env var with empty env var should be interpreted in the same way as an usnset env var. More: #2045


carlosalberto:
@open-telemetry/proto-go-maintainers A last look perhaps?

MrAlias (approved):

PR #2061, opened 2021-10-26
Add semantic conventions for process metrics

### LINKED ISSUE #1817: Define metric semantic conventions for OS processes
The list of process metrics TBD, but can possibly include things like process uptime, cpu and memory usage, etc.

We have an empty [Semantic Conventions for OS Process Metrics](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/process-metrics.md) page where we intended to list the process metrics.

djaglowski:
## Changes

Adds semantic conventions for process metrics

Resolves #1817 

Prerequisite for [(collector) #4232](https://github.com/open-telemetry/opentelemetry-collector/issues/4232)

## Notes

I would appreciate perspectives on `process.cpu.time`, in regards to nuances described [here](https://github.com/open-telemetry/opentelemetry-collector/issues/4232#issuecomment-952130568).

### DIFF
--- a/specification/metrics/semantic_conventions/process-metrics.md
+++ b/specification/metrics/semantic_conventions/process-metrics.md
@@ -16,9 +16,24 @@ metrics](runtime-environment-metrics.md).
 <!-- toc -->
 
 - [Metric Instruments](#metric-instruments)
+  * [Process](#process)
+- [Attributes](#attributes)
 
 <!-- tocstop -->
 
 ## Metric Instruments
 
-TODO
+### Process
+
+Below is a table of Process metric instruments.
+
+| Name | Instrument | Units | Description | Labels |
+|------|------------|-------|-------------|--------|
+| `process.cpu.time` | Asynchronous Counter | s | Total CPU seconds broken down by different states. | `state`, if specified, SHOULD be one of: `system`, `user`, `wait`. A process SHOULD be characterized _either_ by data points with no `state` labels, _or only_ data points with `state` labels. |
+| `process.memory.usage` | Asynchronous UpDownCounter | By | The amount of physical memory in use. |  |
+| `process.memory.virtual` | Asynchronous UpDownCounter | By | The amount of committed virtual memory. |  |
+| `process.disk.io` | Asynchronous Counter | By | Disk bytes transferred. | `direction` SHOULD be one of: `read`, `write` |
+
+## Attributes
+
+Process metrics SHOULD be associated with a [`process`](../../resource/semantic_conventions/process.md#process) resource whose attributes provide additional context about the process.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@ release.
   ([#2210](https://github.com/open-telemetry/opentelemetry-specification/pull/2210))
 - Use UCUM units in Metrics Semantic Conventions.
   ([#2199](https://github.com/open-telemetry/opentelemetry-specification/pull/2199))
+- Add semantic conventions for process metrics.
+  [#2032](https://github.com/open-telemetry/opentelemetry-specification/pull/2061)
 
 ### Logs
 
--- a/specification/metrics/semantic_conventions/README.md
+++ b/specification/metrics/semantic_conventions/README.md
@@ -121,8 +121,10 @@ for the total amount of memory on a system.
 amount of memory in a each state. Where appropriate, the sum of **usage**
 over all attribute values SHOULD be equal to the **limit**.
 
-  A measure of the amount of an unlimited resource consumed is differentiated
-  from **usage**.
+  A measure of the amount consumed of an unlimited resource, or of a resource
+  whose limit is unknowable, is differentiated from **usage**. For example, the
+  maximum possible amount of virtual memory that a process may consume may
+  fluctuate over time and is not typically known.
 
 - **utilization** - an instrument that measures the *fraction* of **usage**
 out of its **limit** should be called `entity.utilization`. For example,


djaglowski:
> Do we also need to make the "process id" an attribute for all process metrics?

It looks like it is [already defined](https://github.com/open-telemetry/opentelemetry-specification/blob/3e380e249f60c3a5f68746f5e84d10195ba41a79/specification/resource/semantic_conventions/process.md#process) as a resource attribute.

tigrannajaryan:
> > Do we also need to make the "process id" an attribute for all process metrics?
> 
> It looks like it is [already defined](https://github.com/open-telemetry/opentelemetry-specification/blob/3e380e249f60c3a5f68746f5e84d10195ba41a79/specification/resource/semantic_conventions/process.md#process) as a resource attribute.

Right!

djaglowski:
@tigrannajaryan I misunderstood your point. I think you suggesting we should explicitly state the requirement in this document. I'll add this!

specification/metrics/semantic_conventions/process-metrics.md
```diff
@@ -16,9 +16,28 @@ metrics](runtime-environment-metrics.md).
...
+| `process.memory.virtual_usage` | Asynchronous UpDownCounter | By | Virtual memory size. |  |
+| `process.disk.io` | Asynchronous Counter | By | Disk bytes transferred. | `direction` SHOULD be one of: `read`, `write` |
+| `process.uptime` | Asynchronous Counter | s | Uptime of the process. | |
+| `process.runtime.memory.heap.allocated` | Asynchronous UpDownCounter | By | Bytes of allocated heap objects (see 'go doc runtime.MemStats.HeapAlloc.'). |  |
+| `process.runtime.memory.heap.allocated_cumulative` | Asynchronous Counter | By | Cumulative bytes allocated for heap objects (see 'go doc runtime.MemStats.TotalAlloc'). |  |
+| `process.runtime.memory.system` | Asynchronous UpDownCounter | By | Total bytes of memory obtained from the OS (see 'go doc runtime.MemStats.Sys'). |  |
+
+## Attributes
+
+[`process`](../../resource/semantic_conventions/process.md#process) and [`process.runtime`](../../resource/semantic_conventions/process.md#process) resource attributes SHOULD be included on Process metric events as appropriate.
```
djaglowski, line 43: @tigrannajaryan Do you think it's appropriate to reference resource attributes like this? My thinking is that it would be an antipattern to re-define them here.

specification/metrics/semantic_conventions/process-metrics.md
```diff
@@ -16,9 +16,23 @@ metrics](runtime-environment-metrics.md).
...
+Below is a table of Process metric instruments.
+
+| Name | Instrument | Units | Description | Labels |
+|------|------------|-------|-------------|--------|
+| `process.cpu.time` | Asynchronous Counter | s | Total CPU seconds broken down by different states. | `state`, if specified, SHOULD be one of: `system`, `user`, `wait`. A process SHOULD be characterized _either_ by data points with no `state` labels, _or only_ data points with `state` labels. |
+| `process.memory.physical_usage` | Asynchronous UpDownCounter | By | The amount of physical memory in use. |  |
+| `process.memory.virtual_usage` | Asynchronous UpDownCounter | By | Virtual memory size. |  |
+| `process.disk.io` | Asynchronous Counter | By | Disk bytes transferred. | `direction` SHOULD be one of: `read`, `write` |
+| `process.uptime` | Asynchronous Counter | s | Uptime of the process. | |
+| `process.runtime.memory.heap.allocated` | Asynchronous UpDownCounter | By | Bytes of allocated heap objects (see 'go doc runtime.MemStats.HeapAlloc.'). |  |
```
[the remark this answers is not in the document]
djaglowski: Looking into this further, it seems the spec has defined the `process.runtime` namespace for runtime resource attributes, but recommends `runtime.*` for runtime metric names.

specification/metrics/semantic_conventions/process-metrics.md
```diff
@@ -16,9 +16,28 @@ metrics](runtime-environment-metrics.md).
...
 
-TODO
+### Process
+
+Below is a table of Process metric instruments.
+
+| Name | Instrument | Units | Description | Labels |
+|------|------------|-------|-------------|--------|
+| `process.cpu.time` | Asynchronous Counter | s | Total CPU seconds broken down by different states. | `state`, if specified, SHOULD be one of: `system`, `user`, `wait`. A process SHOULD be characterized _either_ by data points with no `state` labels, _or only_ data points with `state` labels. |
+| `process.memory.physical_usage` | Asynchronous UpDownCounter | By | The amount of physical memory in use. |  |
```
jsuereth, line 33: Why is memory split into two counters instead of just `process.memory.usage` with attributes to denote `physical` vs. `virtual`?

The metric semconv recommends one with labels.

specification/metrics/semantic_conventions/process-metrics.md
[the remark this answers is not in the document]
jsuereth: Yeah, this would be `runtime.os` vs. `runtime.jvm`

jsuereth (commented):
It seems to me that the process collector + the general naming conventions for metrics don't really line up.  If we're going to create semconv, we should figure out how to line up with those naming conventions or purposefully deviate.

specification/metrics/semantic_conventions/process-metrics.md
djaglowski: I actually had it that way, but [Tigran pointed out](https://github.com/open-telemetry/opentelemetry-collector/issues/4232#issuecomment-952088591) that the sum of these values is not a meaningful aggregation, so we reverted to this.

specification/metrics/semantic_conventions/process-metrics.md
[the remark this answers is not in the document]
djaglowski: > Yeah, this would be `runtime.os` vs. `runtime.jvm`

Does it matter that the values are pulled from go's `runtime` package?

Also, @jsuereth do you mind sharing your thoughts on whether or not we need to reconcile the [`process.runtime.*` resource namespace](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/process.md#process-runtimes) w/ the [`runtime.*` metric namespace](https://github.com/open-telemetry/opentelemetry-specification/blob/401bb9c944323d5e3fc773b88ad3319c01abcff0/specification/metrics/semantic_conventions/runtime-environment-metrics.md#runtime-environment-specific-metrics---runtimeenvironment)?

djaglowski:
> It seems to me that the process collector + the general naming conventions for metrics don't really line up. If we're going to create semconv, we should figure out how to line up with those naming conventions or purposefully deviate.

I'm not sure what you're saying here. Are you saying that the _proposed_ metric names don't line up with general naming conventions? (Certainly there are some metrics within the collector that do not, but those will be changed based on the outcome of this PR.)

specification/metrics/semantic_conventions/process-metrics.md
tigrannajaryan: I think the reference is fine. However I am unsure about which of the Process Resource attributes should be required vs optional. The linked page says that `pid` is optional while at least one of name, executable.path, command, command_line or command_args is required. That may be a reasonable requirement for the Resource but in the context of reporting process metrics I am not so sure. Naively I would imagine `pid` is the most important attribute to report because it uniquely identifies the Process on the host while the rest don't.

specification/resource/semantic_conventions/process.md
```diff
@@ -27,7 +27,7 @@
 <!-- semconv process -->
 | Attribute  | Type | Description  | Examples  | Required |
 |---|---|---|---|---|
-| `process.pid` | int | Process identifier (PID). | `1234` | No |
+| `process.pid` | int | Process identifier (PID). | `1234` | Yes |
```
tigrannajaryan, line 30: Oh, I see you changed it here. This answers my previous comment. I would like to confirm that we want to indeed change this for the Process Resource in general. Anyone know why this was not Required?

jsuereth:
> I'm not sure what you're saying here. Are you saying that the proposed metric names don't line up with general naming conventions? (Certainly there are some metrics within the collector that do not, but those will be changed based on the outcome of this PR.)

Yes, the proposed names don't line up.  See: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/README.md#general-metric-semantic-conventions

djaglowski:
> Yes, the proposed names don't line up.

I tried to follow those guidelines as closely as possible, but I don't think they cover every scenario or generalize well enough to be taken as clear rules in all cases.

Here's how I'm looking at each of the proposed names. I'm happy to reconsider but you'll have to be more specific.

Clearly following the guidelines:
- `process.cpu.time` 
- `process.disk.io`

Naming guidelines [appear to contradict aggregation rule of thumb](https://github.com/open-telemetry/opentelemetry-specification/pull/2061#discussion_r740504935):
- `process.memory.physical_usage` 
- `process.memory.virtual_usage`

I took it for granted that uptime is a special case, but the naming guidelines suggest this should be `process.time`. It seems to me that uptime communicates the meaning much more intuitively though. 
- `process.uptime`

There's an open question [here](https://github.com/open-telemetry/opentelemetry-specification/pull/2061#discussion_r740509308) about whether there is an issue with the root namespace. Otherwise, it's not clear to me how these go against the naming guidelines.
- `process.runtime.memory.heap.allocated`
- `process.runtime.memory.heap.allocated_cumulative`
- `process.runtime.memory.system`

jsuereth:
> Naming guidelines appear to contradict aggregation rule of thumb:
>
> process.memory.physical_usage
> process.memory.virtual_usage


Yeah, I found this last week but didn't get a chance to open a bug on it. 

> Otherwise, it's not clear to me how these go against the naming guidelines.

I'll try to be less clear in my objections going forward.  I'm only objecting to the physical vs. virtual usage.  I think you could mark an argument that "limit" doesn't make sense for virtual usage but does for physical, so they need to be called out and I could by that (and would want that decision recorded somewhere for all time).

If `physical_usage` and `virutal_usage` are better documented, then I have no objections besides `runtime.process` vs. `process.runtime`.   Specifically, I *suspect* that distinction should be covered by instrumentation-library going forward, but we just don't have enough prototypes to justify such a stringent change to semconv.

djaglowski:
> I think you could mark an argument that "limit" doesn't make sense for virtual usage but does for physical, so they need to be called out and I could by that (and would want that decision recorded somewhere for all time).

@jsuereth I've renamed the memory metrics, so that physical memory is clearly described as a `usage` metric, while virtual is not. I've also proposed some additional language on the general guidelines for `usage` metrics. Let me know what you think.

specification/metrics/semantic_conventions/process-metrics.md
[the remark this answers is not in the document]
djaglowski: Regarding these runtime metrics - I think it would be appropriate to move these to `runtime-environment-metrics.md`, and define them only as golang-specific metrics, corresponding to the golang runtime namespace that is tentatively established [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/process.md#go-runtimes). (I will push for that namespace to be reviewed as well.)

Does that seem like a reasonable approach?

specification/metrics/semantic_conventions/process-metrics.md
djaglowski: I've opened an [issue](https://github.com/open-telemetry/opentelemetry-specification/issues/2116) specifically to address the question of requiring the `process.pid` attribute.

specification/metrics/semantic_conventions/process-metrics.md
```diff
@@ -16,9 +16,28 @@ metrics](runtime-environment-metrics.md).
...
+
+Below is a table of Process metric instruments.
+
+| Name | Instrument | Units | Description | Labels |
+|------|------------|-------|-------------|--------|
+| `process.cpu.time` | Asynchronous Counter | s | Total CPU seconds broken down by different states. | `state`, if specified, SHOULD be one of: `system`, `user`, `wait`. A process SHOULD be characterized _either_ by data points with no `state` labels, _or only_ data points with `state` labels. |
+| `process.memory.usage` | Asynchronous UpDownCounter | By | The amount of physical memory in use. |  |
+| `process.memory.virtual` | Asynchronous UpDownCounter | By | The amount of committed virtual memory. |  |
+| `process.disk.io` | Asynchronous Counter | By | Disk bytes transferred. | `direction` SHOULD be one of: `read`, `write` |
+| `process.uptime` | Asynchronous Counter | s | Uptime of the process. | |
```
jsuereth, line 36: I still have the justification in draft form, but please see this: https://github.com/jsuereth/oteps/blob/wip-uptime-monitoring/text/metrics/%3F%3F%3F%3F-uptime-monitoring-semantic-conventions.md

TL;DR: it's likely we shouldn't treat process.uptime as a Counter.

jsuereth (commented):
Sorry for delay in response.  I made a bunch, bet apparently left them as pending!

specification/metrics/semantic_conventions/process-metrics.md
tigrannajaryan: @jsuereth I read your OTEP and I agree with you on Gauge vs Counter (not that I am an expert, but your arguments seem sound).

specification/resource/semantic_conventions/process.md
djaglowski: I am going to back out this change from this PR and allow the decision to be made independently on [this issue](https://github.com/open-telemetry/opentelemetry-specification/issues/2116).

specification/metrics/semantic_conventions/process-metrics.md
djaglowski: I will leave the link, but suggest that I should leave the pid requirement decision to #2116.

specification/metrics/semantic_conventions/process-metrics.md
djaglowski: I will remove this metric from this PR. We can codify the appropriate uptime metric(s) later based on the resolution of the otep.

specification/metrics/semantic_conventions/process-metrics.md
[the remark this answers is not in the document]
djaglowski: Ok, circling back to this. #2112 was merged, which aligns the metric-name and resource-attribute namespaces (both now `process.runtime`). 

In an effort to streamline things, I've reduced the scope of this PR by removing the proposed `process.runtime` metrics. These can be added in a future PR.

specification/metrics/semantic_conventions/process-metrics.md
```diff
@@ -16,9 +16,23 @@ metrics](runtime-environment-metrics.md).
...
-TODO
+### Process
+
+Below is a table of Process metric instruments.
+
+| Name | Instrument | Units | Description | Labels |
+|------|------------|-------|-------------|--------|
+| `process.cpu.time` | Asynchronous Counter | s | Total CPU seconds broken down by different states. | `state`, if specified, SHOULD be one of: `system`, `user`, `wait`. A process SHOULD be characterized _either_ by data points with no `state` labels, _or only_ data points with `state` labels. |
+| `process.memory.physical_usage` | Asynchronous UpDownCounter | By | The amount of physical memory in use. |  |
+| `process.memory.virtual_usage` | Asynchronous UpDownCounter | By | Virtual memory size. |  |
```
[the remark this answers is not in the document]
djaglowski: Resolving this comment, as the proposed metrics have been removed from this PR.

tigrannajaryan (approved):

tigrannajaryan:

djaglowski:

MrAlias (approved):

tigrannajaryan:

specification/metrics/semantic_conventions/process-metrics.md
```diff
@@ -16,9 +16,24 @@ metrics](runtime-environment-metrics.md).
...
+| Name | Instrument | Units | Description | Labels |
+|------|------------|-------|-------------|--------|
+| `process.cpu.time` | Asynchronous Counter | s | Total CPU seconds broken down by different states. | `state`, if specified, SHOULD be one of: `system`, `user`, `wait`. A process SHOULD be characterized _either_ by data points with no `state` labels, _or only_ data points with `state` labels. |
+| `process.memory.usage` | Asynchronous UpDownCounter | By | The amount of physical memory in use. |  |
+| `process.memory.virtual` | Asynchronous UpDownCounter | By | The amount of committed virtual memory. |  |
+| `process.disk.io` | Asynchronous Counter | By | Disk bytes transferred. | `direction` SHOULD be one of: `read`, `write` |
+
+## Attributes
+
+[`process`](../../resource/semantic_conventions/process.md#process) and [`process.runtime`](../../resource/semantic_conventions/process.md#process) resource attributes SHOULD be included on Process metric events as appropriate.
```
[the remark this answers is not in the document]
djaglowski: I see your point about this wording being ambiguous. I'll clean it up.

The statement here is just meant to call attention to the fact that the notion of a process resource is established elsewhere, and to suggest that process metrics should be associated with a process resource.

To be a overly verbose about it. Roughly: 

```yaml
resources:
  # resource 1
  - attributes: # as defined in resource semantic conventions
      process.command_args: cmd/otelcol --config=config.yaml
    metrics:
    - name: process.cpu.time
      data_points: ... # with attribute "state = system | user | wait" as defined in this doc
    - name: process.disk.io
      data_points: ... # with attribute "direction = read | write" as defined in this doc    
  # resource 2
  - attributes: # as defined in resource semantic conventions
      process.command_args: bin/foo --bar
    metrics: ... # whose names and attributes are defined in this doc
```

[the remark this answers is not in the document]
djaglowski: I've pushed an update. Please let me know if it looks right to you. 

I also removed the reference to `process.runtime` attributes, since `process.runtime` metrics are no longer proposed in this PR.

MrAlias (approved):

bogdandrutu (approved):

tigrannajaryan:
@djaglowski please resolve any comments that are addressed. If none are remaining we can start the countdown to merge the PR.

tigrannajaryan:
This has enough approvals and the last change was more than 2 days ago, this is good to merge. @djaglowski please rebase from main so that we can merge.

tigrannajaryan:
@djaglowski please resolve the conflicts and ping me to merge.

djaglowski:
@tigrannajaryan, should be good to go.

Issue #174, opened 2019-07-06
Replace tracer format methods with `inject` and `extract`


rochdev:
Right now there are `getHttpTextFormat` and `getBinaryFormat` defined on `Tracer`. These are inflexible as they don't allow extensions without vendor lock-in by requiring the users to use  an external class directly as it cannot be dynamically added to the list of supported formats.

Having single `inject` and `extract` methods for all formats on `Tracer` similar to OpenTracing would allow adding formats without changing the API. This would apply to formats added by OpenTelemetry, vendors, and users.

I proposed this in JS but it was [rejected](https://github.com/open-telemetry/opentelemetry-js/pull/85/files#r300810012) as it's incompatible with the spec, even though the current API could very easily be replaced by the above which would be a lot cleaner.

bogdandrutu:
Not sure I understand how can you add new capabilities without changing the API, at least even in the OpenTracing API which I suppose you refer to there is an enum which needs to change when adding a new format. If you do use a String for the format that is very unsafe because the type may not be supported.

So the argument that the API does not change is weak, and having the type defined by a string has a lot of problems as well.

Imagine if someone writes a code where they call inject with `format = foo`, if the format is not well define this is a definition of vendor locking. By having well defined formats we actually avoid vendor locking.

rochdev:
> If you do use a String for the format that is very unsafe because the type may not be supported.

In JavaScript an enum is a string or a number, so this argument about type safety is completely invalid. My understanding is that this is a cross-language specification so it shouldn't have Java-centric requirements. In JavaScript this type of pattern is very common to allow extensibility. We currently use this functionality of OpenTracing in our own tracer to stay 100% compliant.

> if the format is not well define this is a definition of vendor locking

The argument was about vendor lock in based on the API. If I switch to a vendor that doesn't support a format that is defined as a string, it will only break propagation. If I need to add a method to add a format and then the user is required to use that method, then when they change vendor the code will actually break. This means that in the first scenario there is no lock in as you can freely change vendor without causing your code to completely break. For the purpose of OpenTelemetry, this extensibility could be used by specific vendors to implement their own formats that could be registered with the tracer and would be ignored by the others. While there could be the argument that these additional formats should be added to OpenTelemetry, this is not always possible as it may be very specific to a single vendor.

bogdandrutu:
[...]

PR #2078, opened 2021-10-28
Define span structure for HTTP retries and redirects

denisivan0v:
## Changes

This PR clarifies semantic conventions for HTTP retries and redirects and defines a span structure and linking as well as span attributes for retries. Changes were discussed recently at Instrumentation SIG meetings.

This change addresses a scenario which is in the scope for bringing the existing HTTP semantic conventions for tracing to an initial stable state, see related [otep #174](https://github.com/open-telemetry/oteps/pull/174).


specification/trace/semantic_conventions/http.md
```diff
@@ -121,6 +122,12 @@ Users MAY explicitly configure instrumentations to capture them even though it i
 
 [network attributes]: span-general.md#general-network-connection-attributes
 
+### HTTP request retries and redirects
+
+A span for each subsequent HTTP request re-sending attempt MUST be created and it SHOULD be linked with a previous one by creating a single `Link`. `http.retry_count` attribute SHOULD be added to each retry span with the value that reflects the ordinal number of request retry attempt.
+
+A span for each subsequent HTTP request redirect MUST be created and it SHOULD be linked with a span for initial request or previous redirect by creating a single `Link`.
```
kenfinnigan, line 129: Was a redirect span being a child considered?

specification/trace/semantic_conventions/http.md
```diff
@@ -121,6 +122,12 @@ Users MAY explicitly configure instrumentations to capture them even though it i
 
 [network attributes]: span-general.md#general-network-connection-attributes
 
+### HTTP request retries and redirects
+
+A span for each subsequent HTTP request re-sending attempt MUST be created and it SHOULD be linked with a previous one by creating a single `Link`. `http.retry_count` attribute SHOULD be added to each retry span with the value that reflects the ordinal number of request retry attempt.
```
kenfinnigan, line 127: I'm having some trouble parsing what is expected to be set on `http.retry_count`. Could you add an example?

Given a situation where a single request failed and we're retrying. Is `http.retry_count` set to `1` because it's the first "retry", or `2` because it's the second attempt to make the request?

specification/trace/semantic_conventions/http.md
denisivan0v: Yes, initially there were several options discussed at Instrumentation SIGs:
- retry/redirect span being a child
- retry/redirect span created for a "physical" request only
- retry/redirect span created for a "logical" request only
- retry/redirect span created for both "physical" and "logical" (+ adding the "logical" span in case there was a retry/redirect)

It was a consensus that by having spans for "physical" requests only + adding links to previous try/redirect we can cover most of customer needs.

specification/trace/semantic_conventions/http.md
denisivan0v: Sure. 
In case a single request was successful there is no `http.retry_count` attribute added.
In case a single request failed and we're retrying, `http.retry_count` set to `1` will be added to the first _retry_ which is the second attempt.

semantic_conventions/trace/http.yaml
```diff
@@ -104,6 +104,13 @@ groups:
         brief: >
           The size of the uncompressed response payload body after transport decoding. Not set if transport encoding not used.
         examples: 5493
+      - id: retry_count
```
jamesmoessis, line 107: What is the reasoning behind this being an attribute? Can't this just be implied by the parent span and the overall structure of the trace?

denisivan0v:
Do folks know how to reopen this PR? Don't see the reopen button.

denisivan0v:
The implementation of changes proposed in this PR for .NET is done within the PR here [opentelemetry-dotnet #2756](https://github.com/open-telemetry/opentelemetry-dotnet/pull/2756).

semantic_conventions/trace/http.yaml
denisivan0v: As per behavior proposed in this PR, a new span must be created for each try/redirect, however, these spans can correspond to different traces (in case there is no trace started manually by a user) and have no parent-child relationship, but a next try is _linked_ to the previous one.

This attribute stores a number of retries and can be useful in querying scenarios.

tonivj5:

Up?

specification/trace/semantic_conventions/http.md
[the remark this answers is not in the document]
anuraaga: Unfortunately for the AWS SDK, it's not currently possible to create a span per retry because the request is only signed once, so the headers, e.g. trace context, need to be set up once before all retries. MUST is too strong here I think.

specification/trace/semantic_conventions/http.md
```diff
@@ -121,6 +123,12 @@ Users MAY explicitly configure instrumentations to capture them even though it i
...
```
cijothomas, line 128: it'd be good to put an example here.

original http call - span1
retry - span2 (linked to span1), retry.count=1
retry - span3 (linked to span1), retry.count=2

tedsuo (approved):
Thanks for putting this together. Based on our conversations at the instrumentation SIG, I approve this.

specification/trace/semantic_conventions/http.md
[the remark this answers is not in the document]
MSNev: In the case of running in a browser and using XMLHttpRequest the browser will automatically follow redirects, but retries (based on responses) needs to be under JS control. From memory *if* you use the fetch API you can tell it not to automatically follow redirects.

MSNev (approved):

specification/trace/semantic_conventions/http.md
jmacd: I found myself asking Ken's question, and I find your answer satisfying -- that this was discussed in group and a consensus was reached. I think there's a danger of losing this justification after the PR merges, and future confusion as this question continues to arise. 

Two ideas: (1) would an example with a span relationship diagram help? (2) would a paragraph of introduction using the words "logical" and "physical" help?  Could be:

```
These semantic conventions call for Spans to be created corresponding to physical HTTP requests. When retries and redirects cause more than one physical request to be sent, multiple sibling spans will be created, each linked to the previous physical request. No span is created corresponding to the logical request, users are expected to interpret span links to understand the behavior of a logical request in the presence of retries and redirects. 
```

jmacd (approved):

specification/trace/semantic_conventions/http.md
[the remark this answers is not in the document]
MSNev: I agree with that from a specification standpoint that decorating a span with a link and count makes sense and therefore agree with this approach of "how" the Span should be decorated to indicate this.

But, after some offline discussion, I'm concerned with the following comment below.

> Essentially, to capture retries and redirects we need to store a state within a request object

Actual language implementation assumptions should __not__ be included as part of the general specification as it WILL NOT be possible for all languages / runtimes.

As such I would suggest that this be reworded to something like (including some ideas from @jmacd in the process 😄 )

------------------------

If creating a Span to identify a retry or redirect attempt, the span SHOULD include a `http.retry_count` attribute with the value that reflects the ordinal number of retry attempt and it SHOULD include a link to the previous one by creating a single `Link`.

Consumers are expected to interpret span links and/or the `http.retry_count` attribute to understand the behavior of HTTP request being represented by the span and to apply any application or service level logic for what the presence may represent, examples may include de-duping, alerting (eg. excessive retries) etc. As these will most likely depend on other attributes of the Span (`http.url`, `http.method`)

------------------------

As such
- SDK's MAY or may not create new Span's to identify retries (AWS SDK)
- Instrumentations may use internal techniques and assumptions to automatically "link" spans that represent the physical HTTP requests. (.Net). Note: Any assumptions SHOULD be documented by the Instrumentation.
- SDK's and/or Instrumentation MAY provide helpers to support the providing / injecting the information required to be added to any new Span when automatic techniques / solutions are not possible. (JavaScript/NodeJS; Python; etc)
- Or specific SDK's and Instrumentations may not provide any additional "support" and the application developers will need to create and provide the details themselves (not ideal, but may be necessary for some Languages / libraries).

specification/trace/semantic_conventions/http.md
lzchen: Isn't span3 supposed to be linked to span2?

lzchen, line 128: Are implementations freely allowed to choose whether or not retry/redirect spans are children of the original request span? In Python, for particular libraries like urllib3, the redirect scenario creates a new tree where as the redirect scenario is a recursive call (part of the original request). For the links, it makes sense to link to the previous call, but can implementations have freedom over the parent->child relationship?

cijothomas: yes I typed wrong.
I was giving an example of an example 😆

denisivan0v: Thank you @cijothomas for the suggestion. Examples added in bb26521.

specification/trace/semantic_conventions/http.md
denisivan0v: It turned out the [same question](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/883#discussion_r794329721) has arisen in the PR for the prototype of this in Python. I've re-written this section in order to make it as clear as possible + added few examples.

@jmacd many thanks for the feedback and suggestions!

specification/trace/semantic_conventions/http.md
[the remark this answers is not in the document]
denisivan0v: @anuraaga and @MSNev thanks a lot for your feedback on this! Based in that and discussions at Instrumentation SIG meeting I've changed wording fro `MUST` to `SHOULD` reflecting that an instrumentation should follow the spec unless it's physically impossible to do so for some languages/runtimes.

Also, I'm completely agree with this point:

> SDK's and/or Instrumentation MAY provide helpers to support the providing / injecting the information required to be added to any new Span when automatic techniques / solutions are not possible.

specification/trace/semantic_conventions/http.md
denisivan0v: I've made some rewording in order to make it clear. Here is the current thinking:

> In order to represent relationship between multiple spans created for retries
or redirects, `Link` SHOULD be used with each subsequent span linked to the
span for the immediately previous physical request. These spans SHOULD be
created as siblings.

Will that work for Python/urllib3?

denisivan0v:
@tedsuo, @jmacd I've made some significant changes to the text in order to address the feedback and make it more straightforward, however, in essence the proposal stays pretty much the same. Can you please have another look?

specification/trace/semantic_conventions/http.md
```diff
@@ -121,6 +124,25 @@ Users MAY explicitly configure instrumentations to capture them even though it i
...
+### HTTP request retries and redirects
+
+Retries and redirects cause more than one physical HTTP request to be sent.
+A CLIENT span SHOULD be created for each one of these physical requests.
+No span is created corresponding to the "logical" (encompassing) request.
+
+In order to represent relationship between multiple spans created for retries
+or redirects, `Link` SHOULD be used with each subsequent span linked to the
+span for the immediately previous physical request. These spans SHOULD be
+created as siblings. In case there is no trace started before retries or
```
anuraaga, line 136: Is there a way to indicate sibling - an attribute on the Link?

anuraaga (approved):

specification/trace/semantic_conventions/http.md
```diff
@@ -280,3 +302,65 @@ If set, it would be
...
+Example of retries in the presence of a trace started by an inbound request:
+
+```
+request (SERVER, trace=t1, span=s1)
+  |
+  -- GET / - 500 (CLIENT, trace=t1, span=s2)
+  |   |
+  |   --- server (SERVER, trace=t1, span=s3)
+  |
+  -- GET / - 500 (CLIENT, trace=t1, span=s4, links=[ {trace=t1, span=s2} ], http.retry_count=1)
```
bogdandrutu, line 317: typo: l/L

specification/trace/semantic_conventions/http.md
bogdandrutu, line 136: Personally I would postpone adding the "links" yet. I would not change anything except removing the "links" suggestion for the moment: The links add the necessity to do an "extra" propagation between siblings. This may be a too much, and would like to reconsider this.

The most common case where there is no parent "span" are:
1. Web apps/mobile apps - these may be solved by creating a notion of "Session" and all of them will be part of the same session.
2. Cron/Background jobs - Maybe it is reasonable to ask users to start a trace there instead of complicating from our side doing siblings propagations.

bogdandrutu: It may be a good proposal to start having "relationship" on links :)

specification/trace/semantic_conventions/http.md
denisivan0v: Fixed in bb2c4e0

specification/trace/semantic_conventions/http.md
denisivan0v: Yes, it might be a good idea to start using attributes on links, so it's possible to indicate linked spans belong to the same "session". Seems this also would be useful for messaging/async scenarios. /cc @pyohannes

specification/trace/semantic_conventions/http.md
lzchen: Yes this is clearer thanks!

lzchen (approved):

specification/trace/semantic_conventions/http.md
denisivan0v: @bogdandrutu Although links require some additional efforts, it's still very powerful tool that helps a lot modeling scenarios like that. To be honest, I don't see some other backward-compatible and simple enough way to represent retries and redirects. Let's take this example with 2 parallel threads sending requests to different hosts:

```
request (SERVER, trace=t1, span=s1)
  |
  -- Host: example1.com
  |  GET / - 200 (CLIENT, trace=t1, span=s2)
  |  |
  |   --- server (SERVER, trace=t1, span=s3)
  |
  -- Host: example2.com
  |  GET / - 500 (CLIENT, trace=t1, span=s4)
  |   |
  |   --- server (SERVER, trace=t1, span=s5)
  |
  -- Host: example2.com
  |  GET / - 500 (CLIENT, trace=t1, span=s6, links=[ {trace=t1, span=s4} ], http.retry_count=1)
  |   |
  |   --- server (SERVER, trace=t1, span=s7)
  |
  -- Host: exampl3.com
  |  GET / - 200 (CLIENT, trace=t1, span=s8)
  |   |
  |   --- server (SERVER, trace=t1, span=s9)
  |
  -- Host: example2.com
  |  GET / - 200 (CLIENT, trace=t1, span=s10, links=[ {trace=t1, span=s6} ], http.retry_count=2)
  |   |
  |   --- server (SERVER, trace=t1, span=s11)
  |
  -- Host: example2.com
     GET / - 200 (CLIENT, trace=t1, span=s12)
      |
      --- server (SERVER, trace=t1, span=s13)
```
Despite we have top-level parent span (a trace is started upfront) here, it's very hard to say which spans represent request retries without links. To re-iterate, even in case we have encompassing span, it doesn't really solve the "session" problem.

@tedsuo, @tigrannajaryan it would be great to get your inputs on this.

denisivan0v: Ping @bogdandrutu.

tedsuo: I think links are a logical choice here - we are creating a linked list of spans that represents a casual relationship - each span is a retry or a redirect triggered by the previous span.

We currently have three examples use cases for links: batching/fan-in, async messaging, and now retries. I don't think links are sufficient to model all of these situations without additional attributes - do we need a "link type" attribute to differentiate between them? A "TraceGroupID" attribute which makes these linked traces easier to index? As part of Microsoft and others implementing links, we will get some clarification about what may be needed.

But I think it is fine to start with this, as links are the correct data structure, and this allows us to model retries in our instrumentation. But we should be cautious about slapping on any additional attributes until we understand what would actually be useful.

denisivan0v: @bogdandrutu can I resolve this conversation to be able to move forward?

jamesmoessis (approved):
LGTM 👍

specification/trace/semantic_conventions/http.md
bogdandrutu: @denisivan0v my opinion, in order to make progress I would remove the links part for the moment, and split that in a separate PR.

My motivation is to better document how/where to use links, including the need/or not to have a relationship between links, etc. This is also related to https://github.com/open-telemetry/opentelemetry-specification/pull/2278 where we have the same problem.

tigrannajaryan:
This needs to wait for the 2 days cool-down period since the last change.

specification/trace/semantic_conventions/http.md
```diff
@@ -280,3 +294,65 @@ If set, it would be
...
+Example of retries in the presence of a trace started by an inbound request:
+
+```
+request (SERVER, trace=t1, span=s1)
+  |
+  -- GET / - 500 (CLIENT, trace=t1, span=s2)
+  |   |
+  |   --- server (SERVER, trace=t1, span=s3)
+  |
+  -- GET / - 500 (CLIENT, trace=t1, span=s4, http.retry_count=1)
```
bogdandrutu, line 309: nit: should this be 2?

bogdandrutu (approved):

specification/trace/semantic_conventions/http.md
denisivan0v: That's the first _retry_:

> `http.retry_count` attribute SHOULD be added to each retry span with the value that reflects the ordinal number of request retry attempt.

Retry semantics allows us to add additional attributes to retry spans only, so there is no changes to span structure/attributes in case there were no retries.

bogdandrutu:
Merging this as the initial step for this work.

specification/trace/semantic_conventions/http.md
```diff
@@ -280,3 +294,65 @@ If set, it would be
 but due to `http.scheme`, `http.host` and `http.target` being set, it would be redundant.
 As explained above, these separate values are preferred but if for some reason the URL is available but the other values are not,
 URL can replace `http.scheme`, `http.host` and `http.target`.
+
+## HTTP retries and redirects examples
```
cijothomas, line 298: minor: I think it'd be better to split retries and redirects into own headings

specification/trace/semantic_conventions/http.md
```diff
@@ -121,6 +124,17 @@ Users MAY explicitly configure instrumentations to capture them even though it i
 
 [network attributes]: span-general.md#general-network-connection-attributes
 
+### HTTP request retries and redirects
+
+Retries and redirects cause more than one physical HTTP request to be sent.
+A CLIENT span SHOULD be created for each one of these physical requests.
+No span is created corresponding to the "logical" (encompassing) request.
```
cijothomas, line 131: This statement is not clear? what exactly does it mean? (in the context of the given example)

bogdandrutu:
@denisivan0v I swear I tried, but @cijothomas commented before I was able to press the button, and we configured to have all conversations resolved. Please resolve them and I will merge after. you can ping me on slack when everything is resolved.

specification/trace/semantic_conventions/http.md
denisivan0v: Done in 7f7d801

specification/trace/semantic_conventions/http.md
cijothomas: got it. But I don't see a need to mention that statement here.

PR #2090, opened 2021-11-01
Add aerospike to db.system semantic conventions

verdie-g:
## Changes

Adds [Aerospike](https://aerospike.com) to the list of well-known identifiers for `db.system`.

~Also, I could specify what `db.name` should be for that technology. There is a concept of namespace (see [Aerospike data model](https://docs.aerospike.com/docs/architecture/data-model.html)) but it's~
> ~Similar to a tablespace in an RDBMS, a namespace is a collection of records that share one specific storage engine, with common policies such as repliation factor, encryption and more. A database can contain multiple namespaces.~

~So I think `db.name` should be empty and `db.aerospike.namespace` and `db.aerospike.set` could be optionally set.~


bogdandrutu (changes requested):
Confused about PR description which seems to be more of a question/discussion.

verdie-g:
I've striked the part irrelevant to the PR. I'll open an issue if `db.name` should be documented for aerospike.

Issue #2042, opened 2021-10-20
Should Syslog Emergency map to FATAL instead of ERROR3?


markdascher:
The Log Data Model's [Appendix B](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#appendix-b-severitynumber-example-mappings) shows an example mapping from syslog severity levels to SeverityNumber values. All of the syslog severities make sense except for Emergency and Alert, which are apparently swapped.

In syslog, Emergency is the highest severity possible, and Alert is just below Emergency, at least numerically. Quoting from RFC5424:

```
              0       Emergency: system is unusable
              1       Alert: action must be taken immediately
              2       Critical: critical conditions
              3       Error: error conditions
              4       Warning: warning conditions
              5       Notice: normal but significant condition
              6       Informational: informational messages
              7       Debug: debug-level messages
```

I'd therefore expect Emergency=FATAL and Alert=ERROR3, but Appendix B does the opposite. Looking through the commit history, https://github.com/tigrannajaryan/rfcs/commit/784e409408ad50b35d99772e6192004b8a3385bd seems relevant, possibly in response to [this conversation](https://github.com/open-telemetry/oteps/pull/97#discussion_r423030569). But it doesn't quite match the mappings suggested in that comment. (The Alert=FATAL mapping was already there anyway.)

Was this just a typo, or is there more to it?

markdascher:
Submitted #2087 with the described changes.

Here's the discussion I was thinking of: https://docs.google.com/document/d/1WQDz1jF0yKBXe3OibXWfy3g6lor9SvjZ4xT-8uuDCiA/edit?disco=AAAAGb77xaY

PR #2091, opened 2021-11-02
Fix Syslog severity number mapping in the example

tigrannajaryan:
Emergency and Alert were incorrectly ordered (Emergency is was
less severe but it is actually more severe in Syslog).

This fixes the mapping.

Fixes https://github.com/open-telemetry/opentelemetry-specification/issues/2042


kumoroku (approved):

djaglowski (approved):

zenmoto (approved):

tigrannajaryan:
docfx action is not feeling well.

kumoroku (approved):

tigrannajaryan:
We have 2 approvals which supposedly should be enough but I still can't merge. Let's get one more approval if possible.

bogdandrutu (approved):

Issue #1808, opened 2021-07-08
Google Cloud Logging specification:  log_name should not map to Name


jsuereth:
**What are you trying to achieve?**

According to the logging data model: `Name: Short event identifier.`, and additionally in the [Google Cloud mapping](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#google-cloud-logging), `Name` is set to `log_name`, which is documented [here](https://cloud.google.com/logging/docs/reference/v2/rpc/google.logging.v2#logentry).

TL;DR; `log_name` in cloud represents more of a "source file" of the log and not an event identifier.  It has values like: 

```
"projects/[PROJECT_ID]/logs/[LOG_ID]"
"organizations/[ORGANIZATION_ID]/logs/[LOG_ID]"
"billingAccounts/[BILLING_ACCOUNT_ID]/logs/[LOG_ID]"
"folders/[FOLDER_ID]/logs/[LOG_ID]"
```

Which does not seem to be in the spirit of the logging data model.

### Proposal

We should move `log_name` to either an Attribute or Resource.Attribute in the logging data model.

djaglowski:
This was discussed today in the Log SIG as part of our effort to resolve any issues that may affect the log data model. PR #2092 potentially resolves the issue, pending review from @jsuereth.

PR #2092, opened 2021-11-02
Change mapping for log_name of Google Cloud Logging

tigrannajaryan:
Fixes https://github.com/open-telemetry/opentelemetry-specification/issues/1808

The log_name should not map to Name, they are not the same thing.

log_name description: The resource name of the log to which this log entry belongs.
(See https://cloud.google.com/logging/docs/reference/v2/rpc/google.logging.v2#logentry)

Name description: Short event identifier that does not contain varying parts.

log_name is now mapped to a Google-specific attribute. If better mapping is known
I can change to that but I don't see anything else in log data model log_name maps
nicely to.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@ release.
   ([#2196](https://github.com/open-telemetry/opentelemetry-specification/pull/2196))
 - Added ObservedTimestamp to the Log Data Model.
   ([#2184](https://github.com/open-telemetry/opentelemetry-specification/pull/2184))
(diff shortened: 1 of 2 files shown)


tigrannajaryan:
@jsuereth PTAL

jmacd (approved):

djaglowski:
@jsuereth Would you please give this a review? The Log SIG has interest in seeing the underlying issue closed before the data model is declared stable.

tigrannajaryan:
@jsuereth PTAL.

jsuereth (approved):

PR #2100, opened 2021-11-04
Add `device.manufacturer` to semantic conventions for resources

ladd:
## Changes

- Proposed manufacturer field

### DIFF
--- a/semantic_conventions/resource/device.yaml
+++ b/semantic_conventions/resource/device.yaml
@@ -36,5 +36,5 @@ groups:
         type: string
         brief: 'The name of the device manufacturer'
         note: >
-          Android devices provide this field via [Build](https://developer.android.com/reference/android/os/Build#MANUFACTURER)
+          The Android OS provides this field via [Build](https://developer.android.com/reference/android/os/Build#MANUFACTURER). iOS apps SHOULD hardcode the value `Apple`.
         examples: ['Apple', 'Samsung']
--- a/semantic_conventions/resource/device.yaml
+++ b/semantic_conventions/resource/device.yaml
@@ -32,3 +32,9 @@ groups:
           It's recommended this value represents a human readable version of the
           device model rather than a machine readable alternative.
         examples: ['iPhone 6s Plus', 'Samsung Galaxy S6']
+      - id: manufacturer
+        type: string
+        brief: 'The name of the device manufacturer'
+        note: >
+          Android devices provide this field via [Build](https://developer.android.com/reference/android/os/Build#MANUFACTURER)
+        examples: ['Apple', 'Samsung']
--- a/specification/resource/semantic_conventions/device.md
+++ b/specification/resource/semantic_conventions/device.md
@@ -12,10 +12,14 @@
 | `device.id` | string | A unique identifier representing the device [1] | `2ab2916d-a51f-4ac8-80ee-45ac31a28092` | No |
 | `device.model.identifier` | string | The model identifier for the device [2] | `iPhone3,4`; `SM-G920F` | No |
 | `device.model.name` | string | The marketing name for the device model [3] | `iPhone 6s Plus`; `Samsung Galaxy S6` | No |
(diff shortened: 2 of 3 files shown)


specification/resource/semantic_conventions/device.md
```diff
@@ -12,10 +12,14 @@
 | `device.id` | string | A unique identifier representing the device [1] | `2ab2916d-a51f-4ac8-80ee-45ac31a28092` | No |
 | `device.model.identifier` | string | The model identifier for the device [2] | `iPhone3,4`; `SM-G920F` | No |
 | `device.model.name` | string | The marketing name for the device model [3] | `iPhone 6s Plus`; `Samsung Galaxy S6` | No |
+| `device.manufacturer.name` | string | The name of the device manufacturer [4] | `Apple`; `Samsung` | No |
```
iNikem, line 15: Why not just `device.manufacturer`?

ladd: Yep, fine with me. I was thinking there could be other subfields, but I can't think of any off the top of my head.

iNikem: > Note: the fact that an entity is located within another entity is typically not a sufficient reason for forming nested namespaces. The purpose of a namespace is to avoid name clashes, not to indicate entity hierarchies. This purpose should primarily drive the decision about forming nested namespaces

From https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-naming.md#attribute-naming

iNikem (approved):

ladd:
Could I get a review, please?

arminru:
@ladd Please note that you will need to update the YAML definitions in this folder following the readme there:
https://github.com/open-telemetry/opentelemetry-specification/tree/main/semantic_conventions
By running the command documented there, the markdown table you edited manually will be generated automatically an make the build pass.

ladd:
> @ladd Please note that you will need to update the YAML definitions in this folder following the readme there: https://github.com/open-telemetry/opentelemetry-specification/tree/main/semantic_conventions By running the command documented there, the markdown table you edited manually will be generated automatically an make the build pass.

Thanks -- done!

Oberon00:
Is this something we'd actually want to send? Since we already send device IDs, couldn't this be handled with a device database at the backend? Or do you think the practical advantages will be enough? Does one usually have an API that the instrumentation could call to get the manufacturer? Or would it have to have a DB?

ladd:
Device ids are typically not mappable to manufacturer strings, unless you have a separate instrumentation mechanism to populate a DB. Device model strings can sometimes be mapped, but it normally requires licensing a proprietary database (like GSM Arena) I do think the practical advantages are there -- specific Android manufacturers often tweak the OS in compatibility breaking ways. The API exists for Android -- see my footnote in the PR. For iOS, you would normally just hardcode `Apple`.

ladd:
Can I get a review / merge if the comments have been addressed?

semantic_conventions/resource/device.yaml
```diff
@@ -32,3 +32,9 @@ groups:
           It's recommended this value represents a human readable version of the
           device model rather than a machine readable alternative.
         examples: ['iPhone 6s Plus', 'Samsung Galaxy S6']
+      - id: manufacturer
+        type: string
+        brief: 'The name of the device manufacturer'
+        note: >
+          Android devices provide this field via [Build](https://developer.android.com/reference/android/os/Build#MANUFACTURER)
```
Oberon00, line 39: If we expect this to be hard-coded for iOS, specifying that hardcoded string would be good (e.g. as `Apple` like in the example). Otherwise we may end up seeing `Apple Inc.` or other variants.

Oberon00 (approved):
Please add a CHANGELOG entry.

ladd:
Can I get a review / merge if the comments have been addressed?

jsuereth (approved):

trask (approved):

PR #2106, opened 2021-11-05
Clarify Metric data type

### LINKED ISSUE #2073: Clarification on model metric of exporter.export(batch)

> ##### Export(batch)
> Exports a batch of `Metrics`. Protocol exporters that will implement this function are typically expected to serialize and transmit the data to the destination.
> **Parameters:**
> batch - a batch of `Metrics`. The exact data type of the batch is language specific, typically it is some kind of list.
> https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#exportbatch

The type name `Metrics` is not mentioned anywhere else (as far as I can tell, I didn't find anything at https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md) and it is unclear what it contains.

reyang:
Fixes #2073.

This change adds extra clarification, it doesn't change the spec requirement.


specification/metrics/sdk.md
```diff
@@ -728,8 +728,32 @@ are being sent to.
...
+                                    |
+                                    +--> timestamps, dimensions, value (or buckets), exemplars, ...
+```
+
+Refer to the [Metric points](./datamodel.md#metric-points) section from the
+Metrics Data Model specification for more details.
+
+Note: it is highly recommended that implementors design the `Metric` data type
+_based on_ the [Data Model](./datamodel.md), rather than directly use the data
+types generated from the proto files (because the types generated from proto
```
cijothomas, line 755:
```suggestion
types generated from the [proto files](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto) (because the types generated from proto
```

cijothomas (approved):
lgtm

jsuereth (approved):

specification/metrics/sdk.md
```diff
@@ -728,8 +728,34 @@ are being sent to.
...
+
+Refer to the [Metric points](./datamodel.md#metric-points) section from the
+Metrics Data Model specification for more details.
+
+Note: it is highly recommended that implementors design the `Metric` data type
+_based on_ the [Data Model](./datamodel.md), rather than directly use the data
+types generated from the [proto
+files](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto)
+(because the types generated from proto files are not guaranteed to be backward
+compatible).
```
dyladan, line 758: Does this refer to breaking proto changes or just changes to the proto which are backward compatible but the generated types are not backwards compatible

reyang: Both and more (e.g. the type generated by one tool is incompatible with another tool).

specification/metrics/sdk.md
```diff
@@ -729,8 +729,34 @@ are being sent to.
...
+
+```text
+       +--------+ +--------+     +--------+
+Batch: | Metric | | Metric | ... | Metric |
+       +---+----+ +--------+     +--------+
+           |
+           +--> name, unit, description, meter information, ...
+           |
+           |                  +-------------+ +-------------+     +-------------+
+           +--> MetricPoints: | MetricPoint | | MetricPoint | ... | MetricPoint |
```
findmyway, line 745: Just a quick question, the `MetricPoint`s here is mainly distinguished by `dimensions`, right?

reyang: Short answer is Yes.

Longer version - you can find more details from the [data model spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#metric-points).

MrAlias (commented):
This tries to clarify the types that implementations can use, but I think the original issue that this was opened to resolve is that the term `Metric` itself is not well defined. Here in the specification we are using the term to describe a group of `MetricPoint`s, but above it also describes a ["metric time series"](https://github.com/open-telemetry/opentelemetry-specification/blob/a405c96ae4300b69d2994278e5b8a7be8d1d6092/specification/metrics/sdk.md#attribute-limits). This seems like a overloading of the term.

Additionally, above we refer to [`Aggregated Metrics`](https://github.com/open-telemetry/opentelemetry-specification/blob/a405c96ae4300b69d2994278e5b8a7be8d1d6092/specification/metrics/sdk.md#aggregation) above, but it seems to be referring to the same conceptual thing we a describing here.

I think adding these changes can help clarify implementation details for implementers, but we should also be consistent in our use of terms. I would recommend using `Aggregated Metrics` here if they are the same as above, or maybe instead just say "a batch of `MetricPoint` groups".

specification/metrics/sdk.md
```diff
@@ -729,8 +729,34 @@ are being sent to.
 
 **Parameters:**
 
-`batch` - a batch of `Metrics`. The exact data type of the batch is language
-specific, typically it is some kind of list.
+`batch` - a batch of `Metric`s. The exact data type of the batch is language
+specific, typically it is some kind of list. The exact type of `Metric` is
```
jonatan-ivanov, line 733: List sometimes mean an ordered data type that also allows duplicates. As far as I understand here, we don't have that constraint here so I would use a more general term like collection:

```suggestion
specific, typically it is some kind of collection (list, array, set, etc). The exact type of `Metric` is
```

jonatan-ivanov (approved):

specification/metrics/sdk.md
findmyway: Thanks!

specification/metrics/sdk.md
reyang: The intent is to be consistent with https://github.com/open-telemetry/opentelemetry-specification/blob/3e380e249f60c3a5f68746f5e84d10195ba41a79/specification/trace/sdk.md#exportbatch.

I think `collection` is more frequently used in some languages (e.g. C#, Java). No strong opinion though.

reyang:
> I think adding these changes can help clarify implementation details for implementers, but we should also be consistent in our use of terms. I would recommend using `Aggregated Metrics` here if they are the same as above, or maybe instead just say "a batch of `MetricPoint` groups".

I've updated the wording, PTAL.

jmacd (approved):

specification/metrics/sdk.md
jmacd: Given this is already a "typically", I think it's OK to leave "some kind of list".

PR #2112, opened 2021-11-08
Align runtime metric and resource namespaces

djaglowski:
A 'process.runtime' namespace is currently defined in the resource
semantic conventions. The metric semantic conventions suggest that
runtime metrics use a 'runtime' namespace. Since the runtime metrics
are directly related to the runtime resources, they ought to share a
common namespace.

Related to issue #2061, discussed in [this comment](https://github.com/open-telemetry/opentelemetry-specification/pull/2061#discussion_r740509308).

### DIFF
--- a/specification/metrics/semantic_conventions/README.md
+++ b/specification/metrics/semantic_conventions/README.md
@@ -62,8 +62,8 @@ where similar metrics have significantly different implementations across the
 breadth of all existing metrics. For example, every garbage collected runtime
 has slightly different strategies and measures. Using a single set of metric
 names for GC, not divided by the runtime, could create dissimilar comparisons
-and confusion for end users. (For example, prefer `runtime.java.gc*` over
-`runtime.gc.*`.) Measures of many operating system metrics are similarly
+and confusion for end users. (For example, prefer `process.runtime.java.gc*` over
+`process.runtime.gc.*`.) Measures of many operating system metrics are similarly
 ambiguous.
 
 ### Units
--- a/specification/metrics/semantic_conventions/runtime-environment-metrics.md
+++ b/specification/metrics/semantic_conventions/runtime-environment-metrics.md
@@ -13,7 +13,8 @@ semantic conventions when instrumenting runtime environments.
 <!-- toc -->
 
 - [Metric Instruments](#metric-instruments)
-  * [Runtime Environment Specific Metrics - `runtime.{environment}.`](#runtime-environment-specific-metrics---runtimeenvironment)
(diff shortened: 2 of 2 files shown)


specification/metrics/semantic_conventions/README.md
```diff
@@ -62,8 +62,8 @@ where similar metrics have significantly different implementations across the
 breadth of all existing metrics. For example, every garbage collected runtime
 has slightly different strategies and measures. Using a single set of metric
 names for GC, not divided by the runtime, could create dissimilar comparisons
-and confusion for end users. (For example, prefer `runtime.java.gc*` over
-`runtime.gc.*`.) Measures of many operating system metrics are similarly
+and confusion for end users. (For example, prefer `process.runtime.java.gc*` over
```
jsuereth, line 65: I believe (specifically because we're starting to have Java instrumentation using runtime.java.gc) show up this is a case where SCHEMA should provide a migration rule.   @tigrannajaryan is that true?

jsuereth (commented):
I appreciate the goal of trying to align the semantic conventions!  

Would be good to have a x-link to the [process runtime specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/process.md#process-runtimes)

specification/metrics/semantic_conventions/README.md
tigrannajaryan: A Schema file can record a fact of a renamed metric, but it needs the actual metrics names. I am not sure but I do not see actual metrics defined in runtime-environment-metrics.md at the moment, it appears to be just a recommendation for metric names at the moment and a couple examples, right? If we know the exact names of metrics used then yes we can/should define a `rename_metric` section in the next Schema file.

djaglowski: Ok, so it looks like we don't need a schema file then. I'm resolving this comment, but please challenge if necessary.

djaglowski:
@jsuereth I've added the link.

Any other feedback @jsuereth, @tigrannajaryan?

jsuereth (approved):

SergeyKanzhelev (approved):

tigrannajaryan:
@aabmass you are the author of the OTEP that introduced the `runtime.*` metrics. Please review this PR.

Also, @james-bebbington @andrewhsu @justinfoote @jmacd @MrAlias @lzchen you approved the PR that introduced `runtime.*`. Please review this PR, it moves the `runtime.*` to `process.runtime.*`.

jmacd (approved):
:100:

aabmass (approved):
Makes sense to me 👍

carlosalberto (approved):

tigrannajaryan (approved):

PR #2115, opened 2021-11-10
Add Resource browser and app attributes, clarify device attributes as client-side only

martinkuba:
In order to identify client-side telemetry, this PR specifies that `device` attributes should be used only for client-side devices. It also adds a section for `browser` attributes. The presence of either a `device` or `browser` attribute would be used to distinguish that the signal comes from a client-side instrumentation.

This was discussed in the Client-side SIG as well as [Spec SIG on 11/2/21](https://docs.google.com/document/d/1pdvPeKjA8v8w_fGKAN68JjWBmVJtPCpqdi9IZrd6eEo/edit#heading=h.erc0axt1ayqq)

### DIFF
--- a/specification/resource/semantic_conventions/README.md
+++ b/specification/resource/semantic_conventions/README.md
@@ -90,6 +90,23 @@ namespace = Company
 service.name = Shop.shoppingcart
 ```
 
+## App
+
+**type:** `app`
+
+**Description:** A client application instance.
+
+<!-- semconv service -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `app.name` | string | Logical name of the client application. [1] | `web storefront` | Yes |
+| `app.namespace` | string | A namespace for `app.name`. [2] | `Shop` | No |
+| `app.version` | string | The version string of the app build or implementation. | `2.0.0` | No |
+
+**[1]:** MUST be the same for all instances of the application running on individual client devices. If the value was not specified, SDKs MUST fallback to `unknown_application`.
+
+**[2]:** A string value having a meaning that helps to distinguish a group of client applications, for example the team name that owns a group of applications or web site name that consists of multiple components. `app.name` is expected to be unique within the same namespace. If `app.namespace` is not specified in the Resource then `app.name` is expected to be unique for all applications that have no explicit namespace defined (so the empty/unspecified namespace is simply one more valid namespace). Zero-length namespace string is assumed equal to unspecified namespace.
+
 ## Telemetry SDK
 
 **type:** `telemetry.sdk`
--- a/specification/resource/semantic_conventions/README.md
+++ b/specification/resource/semantic_conventions/README.md
@@ -27,7 +27,6 @@ This document defines standard attributes for resources. These attributes are ty
 ## TODOs
 
 * Add more compute units: AppEngine unit, etc.
-* Add Web Browser.
 * Decide if lower case strings only.
 * Consider to add optional/required for each attribute and combination of attributes
   (e.g when supplying a k8s resource all k8s may be required).
@@ -151,11 +150,12 @@ Attributes defining a computing instance (e.g. host):
 Attributes defining a running environment (e.g. Operating System, Cloud, Data Center, Deployment Service):
 
 - [Operating System](./os.md)
-- [Device](./device.md)
+- [Device](./device.md) (client-side)
 - [Cloud](./cloud.md)
 - Deployment:
   - [Deployment Environment](./deployment_environment.md)
   - [Kubernetes](./k8s.md)
+- [Browser](./browser.md)
 
 ## Version attributes
 
--- a/specification/resource/semantic_conventions/README.md
+++ b/specification/resource/semantic_conventions/README.md
@@ -101,11 +101,19 @@ service.name = Shop.shoppingcart
 |---|---|---|---|---|
 | `app.name` | string | Logical name of the client application. [1] | `web storefront` | Yes |
 | `app.namespace` | string | A namespace for `app.name`. [2] | `Shop` | No |
-| `app.version` | string | The version string of the app build or implementation. | `2.0.0` | No |
+| `app.bundle` | string | Name of the bundle or package. [3] | `ShopApplication`, `com.example.ShopApplication` | No |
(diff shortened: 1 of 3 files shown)


specification/resource/semantic_conventions/device.md
```diff
@@ -4,7 +4,7 @@
 
 **type:** `device`
 
-**Description**: The device on which the process represented by this resource is running.
+**Description**: The device on which the process represented by this resource is running. The `device.*` attributes MUST be used only for resources that represent client-side devices - the presence of these attributes can be used to identify client-side telemetry.
```
jkwatson, line 7: I know we had some discussion at the SIG about whether IOT devices would also fall into this category. If they didn't fall into this resource category, where would they end up?

specification/resource/semantic_conventions/README.md
```diff
@@ -91,6 +90,23 @@ namespace = Company
 service.name = Shop.shoppingcart
 ```
 
+## App
```
iNikem, line 93: How is this different from existing `service.*` attributes?

jkwatson: An app is not a service. It has different qualitative properties and behaviors. Separating the namespaces will allow backends to account for these differences and treat RUM signals differently.

mtwo (approved):
We discussed one or two pending changes during today's call, but LGTM otherwise

specification/resource/semantic_conventions/browser.md
```diff
@@ -0,0 +1,18 @@
...
+<!-- semconv device -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `browser.name` | string | The web browser name | `Chrome` | No |
+| `browser.version` | string | The web browser version | `90` | No |
+| `browser.platform` | string | The operating system | `Windows` | No |
+| `browser.mobile` | boolean | Flag indicating a mobile device | | No |
+| `browser.userAgent` | string | Full user-agent string provided by the browser [1] | `'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/95.0.4638.54 Safari/537.36'` | No |
+
+**[1]:** The user-agent value should be provided only from browsers that do not have a mechanism to retrieve name, version, and platform individually (for example from the User-Agent Client Hints API).
```
scheler, line 18: Snake case is recommended for words in attribute names, so browser.userAgent should be  browser.user_agent.
https://opentelemetry.io/docs/reference/specification/common/attribute-naming/

specification/resource/semantic_conventions/browser.md
```diff
@@ -0,0 +1,18 @@
+# Browser
+
+**Status**: [Experimental](../../document-status.md)
+
+**type:** `browser`
+
+**Description**: The web browser in which the application represented by the resource is running. The `browser.*` attributes MUST be used only for resources that represent applications running in a web browser. The presence of these attributes is intended to identify client-side telemetry.
+
```
mhennoch, line 8: Should spec have semconv for things that are currently mostly added during ingest? I don't think anyone parses useragent client side and there is currently no other way of getting the browser version etc.

specification/resource/semantic_conventions/README.md
```diff
@@ -91,6 +90,31 @@ namespace = Company
...
+| `app.bundle` | string | Name of the bundle or package. [3] | `ShopApplication`, `com.example.ShopApplication` | No |
+| `app.version` | string | The version string of the app build or implementation. This can include more verbose information, such as the build number. [4] | `2.0.0` | No |
+| `app.short_version` | string | Human-readable version or version used to identify a release to end users. [5] | `10.14.1` | No |
+
+**[1]:** MUST be the same for all instances of the application running on individual client devices. If the value was not specified, SDKs MUST fallback to `unknown_application`.
+
+**[2]:** A string value having a meaning that helps to distinguish a group of client applications, for example the team name that owns a group of applications or web site name that consists of multiple components. `app.name` is expected to be unique within the same namespace. If `app.namespace` is not specified in the Resource then `app.name` is expected to be unique for all applications that have no explicit namespace defined (so the empty/unspecified namespace is simply one more valid namespace). Zero-length namespace string is assumed equal to unspecified namespace.
```
ssidhu-apm, line 110: Is it correct to say that one app can belong to only one app.namespace? In order words, a single app.namespace can have multiple apps, or app.name(s) but an app.name can only belong to one namespace?

specification/resource/semantic_conventions/browser.md
ivomagi: Maybe indeed worth calling out the specific case for the browser telemetry, where the information in `browser.*` might be represented in `User-Agent` header in the exported telemetry and not in the payload itself?

specification/resource/semantic_conventions/README.md
t2t2: Adding onto above, another case where updating resource would be useful when adding [user identifiying attributes like `enduser.*`](https://github.com/open-telemetry/opentelemetry-specification/blob/1dc24efb2fed4ad5b90277c8d79a1a371df681e6/specification/trace/semantic_conventions/span-general.md#general-identity-attributes) (while on server request model having them added per span is more useful as generally every request may be served to different users, in a stateful application it'd be simpler to put it into resource (as you'd generally have one user using it at a time) and update when application state changes (user logs in / out) as it happens less frequently)

If not mutating the resource then at least swapping out the currently attached Resource on TracerProvider / MeterProvider / ... so any new spans afterwards would use the new state / easily differentiated based on different Resource instance

specification/resource/semantic_conventions/browser.md
martinkuba: I can make the change. However, I wonder if that will make it unclear whether an SDK should implement collecting these values. I think if we assume that all browsers will send these headers, then the SDKs will not be sending any of these attributes, and we might as well not define them?

specification/resource/semantic_conventions/README.md
martinkuba: @ivomagi @t2t2 Good point, I added this as a topic to the Client-side SIG agenda on 12/1.

MSNev:
@open-telemetry/specs-approvers @open-telemetry/technical-committee Can we please have this PR reopened for discussion or would you prefer a new PR?

PR #2119, opened 2021-11-12
draft for metric http yaml

jamesmoessis:
@ocelotl here's the yaml we discussed. 

A few notes: 

* I've tried to make the same assumption that the tracing convention makes: "extending" a group means inheriting its attributes. To achieve this I had to duplicate status code being defined but it's no biggy. 
* We won't be able to recreate the markdown page exactly. It's too unstructured and different to the tracing convention. We might not be able to have the "recommended" column, and in general have attribute tables align with the trace convention. 
* There are some `ref`'s to attributes from the tracing convention. Eventually we will define those in metrics yaml too. 
* My general thinking was to split it up into two top level sections:
  *  "groups" which is what all the trace convention exclusively uses 
  * "instruments" which are metric specific. The build tool won't understand these yet.
* Groups are more abstract like `http.server` and then the instrument is more concrete like `http.server.active_requests` would extend that group and inherit its attributes, as well as defining the instrument type and units. 
* Feel free to play with the yaml file, you may end up finding it easier if the instruments aren't separated from the groups (they only have two extra fields, instrument and units). Separating them was just my preliminary thinking.

### DIFF
--- a/semantic_conventions/metric/http.yaml
+++ b/semantic_conventions/metric/http.yaml
@@ -85,7 +85,7 @@ groups:
         type: string
         brief: >
             The primary server name of the matched virtual host. This should be obtained via configuration.
-            If no such configuration can be obtained, this attribute MUST NOT be set ( `net.host.name` should be used instead).
+            If no such configuration can be obtained, this attribute MUST NOT be set (`net.host.name` should be used instead).
         note: >
             `http.url` is usually not readily available on the server side but would have to be assembled in a cumbersome and
             sometimes lossy process from other information (see e.g. open-telemetry/opentelemetry-python/pull/148).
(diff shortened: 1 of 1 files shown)


semantic_conventions/metric/http.yaml
```diff
@@ -0,0 +1,123 @@
...
+          - [http.scheme, http.host, http.target]
+          - [http.scheme, http.server_name, net.host.port, http.target]
+          - [http.scheme, net.host.name, net.host.port, http.target]
+          - [http.url]
+    attributes:
+      - id: server_name
+        type: string
+        brief: >
+            The primary server name of the matched virtual host. This should be obtained via configuration.
+            If no such configuration can be obtained, this attribute MUST NOT be set ( `net.host.name` should be used instead).
```
reyang, line 88:
```suggestion
            If no such configuration can be obtained, this attribute MUST NOT be set (`net.host.name` should be used instead).
```

PR #2121, opened 2021-11-13
no-op StartSpan: always return the parent

tsloughter:
## Changes

This change makes it so no span id generation is required in the API's no-op tracer. My understanding from the discussion back when this part was added to the spec the implementation is supposed to create a new span id in the no-op tracer if the parent is a recording span -- while saying this is almost never the case it still would require it to be possible.

This is what we had implemented in the Erlang API:

```
    %% Spec: Behavior of the API in the absence of an installed SDK
    case otel_tracer:current_span_ctx(Ctx) of
        Parent=#span_ctx{trace_id=TraceId,
                         is_recording=false} when TraceId =/= 0 ->
            %% if the parent is a non-recording Span it MAY return the parent Span
            Parent;
        Parent=#span_ctx{trace_id=TraceId} when TraceId =/= 0 ->
            %% API MUST create a non-recording Span with the SpanContext in the parent Context
            SpanCtx = Parent#span_ctx{span_id=opentelemetry:generate_span_id(),
                                      is_recording=false},
            SpanCtx;
        _ ->
            %% If the parent Context contains no valid Span,
            %% an empty non-recording Span MUST be returned
            ?NOOP_SPAN_CTX
    end.
```

After implementing configurable IdGenerators which required moving the id generators to the SDK, https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#id-generators, I realized the above code could no longer be possible and would not make sense even if it were.

Go and Java both always return the parent or the empty non-recording span, and maybe others. If this is the case for those two and others then this change is an attempt to bring the spec inline with existing implementations. So I think that makes it a non-breaking change :)


specification/trace/api.md
```diff
@@ -790,14 +790,13 @@ for how propagators are to be distributed.
...
-it MAY return the parent Span back from the creation method.
-If the parent `Context` contains no `Span`, an empty non-recording Span MUST be returned instead
-(i.e., having a `SpanContext` with all-zero Span and Trace IDs, empty Tracestate, and unsampled TraceFlags).
-This means that a `SpanContext` that has been provided by a configured `Propagator`
-will be propagated through to any child span and ultimately also `Inject`,
-but that no new `SpanContext`s will be created.
+This means that operations on a Tracer, or on Spans, should have no side effects
+and do nothing. However, there is one important exception to this general rule,
+and that is related to propagation of a `SpanContext`: The API MUST return the
+`Span` in the parent `Context` (whether explicitly given or implicit current).
```
bogdandrutu, line 796: What do we do in the "extractor" part I think is missing from here.

tsloughter: The extractor part doesn't change. It still extracts the propagated trace context to a non-recording span as before.

bogdandrutu (approved):

iNikem (approved):

specification/trace/api.md
```diff
@@ -795,14 +795,13 @@ for how propagators are to be distributed.
...
-if the parent is a non-recording Span (which it usually always is if no SDK is present),
-it MAY return the parent Span back from the creation method.
-If the parent `Context` contains no `Span`, an empty non-recording Span MUST be returned instead
-(i.e., having a `SpanContext` with all-zero Span and Trace IDs, empty Tracestate, and unsampled TraceFlags).
-This means that a `SpanContext` that has been provided by a configured `Propagator`
-will be propagated through to any child span and ultimately also `Inject`,
-but that no new `SpanContext`s will be created.
+This means that operations on a Tracer, or on Spans, should have no side effects
+and do nothing. However, there is one important exception to this general rule,
+and that is related to propagation of a `SpanContext`: The API MUST return the
```
jmacd, line 800:
```suggestion
and that is related to propagation of a `SpanContext`: The Tracer API to get the `Span` from the `Context` MUST return the
```

jmacd (approved):

specification/trace/api.md
tsloughter: This isn't the API for getting the Span from the Context but is the API for starting a Span.

specification/trace/api.md
```diff
@@ -795,14 +795,13 @@ for how propagators are to be distributed.
 ## Behavior of the API in the absence of an installed SDK
 
 In general, in the absence of an installed SDK, the Trace API is a "no-op" API.
-This means that operations on a Tracer, or on Spans, should have no side effects and do nothing. However, there
```
Oberon00, line 798: Can you try to do the change without additional line-break changes? This way it will be easier to spot the differences.

specification/trace/api.md
```diff
@@ -795,14 +795,13 @@ for how propagators are to be distributed.
...
-it MAY return the parent Span back from the creation method.
-If the parent `Context` contains no `Span`, an empty non-recording Span MUST be returned instead
-(i.e., having a `SpanContext` with all-zero Span and Trace IDs, empty Tracestate, and unsampled TraceFlags).
-This means that a `SpanContext` that has been provided by a configured `Propagator`
-will be propagated through to any child span and ultimately also `Inject`,
-but that no new `SpanContext`s will be created.
+This means that operations on a Tracer, or on Spans, should have no side effects
+and do nothing. However, there is one important exception to this general rule,
+and that is related to propagation of a `SpanContext`: The API MUST return the
+`Span` in the parent `Context` (whether explicitly given or implicit current).
```
Oberon00, line 801: I would like to preserve the post-condition that a Span returned by a no-op tracer is always non-recording:

```suggestion
and that is related to propagation of a `SpanContext`: The API MUST return a
non-recording `Span` with the `SpanContext` in the parent `Context` (whether explicitly given or implicit current).
If the `Span` in the parent `Context` is already non-recording, it SHOULD be returned directly
without instantiating a new `Span`.
```

This could be achieved by using the mechanisms for "wrapping a SpanContext in a Span".

Oberon00 (changes requested):

specification/trace/api.md
tsloughter: I was wondering about that. So should the parent be set to non-recording even if it was recording?

jmacd: > This could be achieved by using the mechanisms for "wrapping a SpanContext in a Span".

I also noticed the relationship, I guess you could write "if the parent-span is already non-recording the SDK MUST behave as if wrapping the parent's span context in a span". I think that's less clear, and wish to accept this suggestion and then merge this PR.

jmacd (approved):

Oberon00 (approved):

carlosalberto (approved):

carlosalberto:
Hey, we forgot to add a CHANGELOG entry! @tsloughter think you could cook one?

tsloughter:
@carlosalberto woops, yup, will open a PR in a minute.

PR #2017, opened 2021-10-13
Initial draft of Prometheus <-> OTLP datamodel specification.

jsuereth:
DO NOT SUBMIT (draft only)

For shared contribution/discussion and improvements.

Updates data model specification to denote prometheus compatibilty.
- Should align with all the work done in the collector / prometheus-wg
- Should allow Prometheus-exporter requirements for metrics SDK

Pending items:

- [ ] Full review from prometheus-wg folks, with additional items
  - [ ] Prometheus -> OTLP -> Prometheus is fully detailed/working in Collector
  - [ ] (non-prometheus OTLP) -> Prometheus needs fleshing out
- [ ] Attach bugs/tickets around missing prometheus data types
- [ ] Exemplar mappings

Related issues:

- https://github.com/open-telemetry/wg-prometheus/issues/66
- https://github.com/open-telemetry/wg-prometheus/issues/65
- https://github.com/open-telemetry/wg-prometheus/issues/64

cc @alolita for help shaping this up.

Issue #1078, opened 2020-10-08
Metrics semantic convention: "up" metric


jmacd:
**What are you trying to achieve?**

`up` is a standard metric in Prometheus systems to indicate that a particular combination of `job` and `instance` was observed to be healthy. This comes from an active role taken by Prometheus servers in collecting metrics, but OpenTelemetry OTLP exporters can synthesize the same information on export to indicate that they are, in fact, up.

This issue proposes we introduce and specify this metric. Prometheus specifies this as 0- or 1-valued metric labeled by the `job`, `instance` labels. In OpenTelemetry the natural expression of this would be a label-free metric named "up", again 0- or 1-valued, reported along with the monitored Resource.

**Additional context.**

This variable would be synthesized in receivers for other metrics protocols. For example, the OTel collector's Prometheus or OpenMetrics receiver would be changed to generate this metric when it scrapes a target.

jgals:
I'd like to work on this one with @cwildman. You can assign it to me @andrewhsu.

jmacd:
Wondering if we should mark this release:allowed-for-ga? 

This is an important step for a push-based metrics pipeline to produce equivalent data as a pull-based metrics pipeline would produce, which is a prerequisite for existing Prometheus users (or a barrier to migration to OTel SDKs), but it is not a requirement (maybe?).

jmacd:
## Objective

Define how to transform a OTel-Collector Prometheus receiver (pull)
stream into OTLP such that it transforms into exactly the PRW output
that Prometheus or Grafana agent would write.  State how the
OTel-Collector receiver should generate the `up` metric and staleness
markers to be semantically equivalent to Prometheus or Grafana agent
in this regard.

AND

Identify a way to transform an OTel SDK's exporter (push) stream into
OTLP such that it transforms into semantically correct PRW output,
using the same translation applied for OTLP to PRW described above.
Define a push-based method for reporting liveness that can be
semantically translated into `up` by a collector when received through
the OTLP receiver and exported through a PRW exporter.

## Background: `up` and staleness markers in the pull data model

The `up` metric is synthesized in the Prometheus server after
completing a scrape, 1 for success, and 0 for failure. The
interpretation of a 0 is that the scrape failed, and when that
happens, staleness markers are entered in all the timeseries that
target has produced.  The PRW consumer will see NaN values in the
timeseries, and the interpretation is that the process is either (a)
[...]

PR #2124, opened 2021-11-15
Add details to Prometheus Exporter requirement

reyang:
## Changes

* Added requirements to the Prometheus Exporter spec (based on the learnings from the prototype).

Related issues #2017, #2035


jmacd (approved):

specification/metrics/sdk_exporters/prometheus.md
```diff
@@ -2,6 +2,32 @@
...
-which reacts to the Prometheus scraper and report the metrics passively to
-[Prometheus](https://prometheus.io/).
+Prometheus Exporter is a [MetricExporter](../sdk.md#metricexporter) which reacts
+to the Prometheus scraper and reports the metrics passively to
+[Prometheus](https://prometheus.io/), or [pushes
+metrics](https://prometheus.io/docs/instrumenting/pushing/) to a [Prometheus
+Pushgateway](https://github.com/prometheus/pushgateway).
+
+Prometheus Exporter SHOULD support [Pull mode](../sdk.md#pull-metric-exporter),
+if possible.
```
MrAlias, line 12: It seems like this should be strengthened to a requirement. It is the default and expected operation of any prometheus system and would be a compatibility issue if one implementation does not support this.

reyang: I think this is already the strongest wording we can put here, considering the fact that some languages/runtimes don't allow HTTP server (e.g. JavaScript running inside a browser, embedded C module running on IoT devices).

specification/metrics/sdk_exporters/prometheus.md
```diff
@@ -2,6 +2,32 @@
...
+
+Prometheus Exporter MAY support [Push mode](../sdk.md#push-metric-exporter), if
+there is a need to support short-lived applications (e.g. short-lived
+service-level batch jobs, JavaScript applications running inside web browsers,
+etc.).
```
jack-berg, line 17: Here's the prometheus remote write [specification](https://docs.google.com/document/d/1LPhVRSFkGNSuU1fBd81ulhsCPR4hkSZyyBj1SZ8fWOM/edit#). While remote write is certainly useful for all the reasons that push metrics are generally useful, the doc has an interesting blurb:

> The remote write protocol is not intended for use by applications to push metrics to Prometheus Remote Write compatible receivers.  It is intended that a Prometheus remote write compatible sender scrapes instrumented applications or exporters and sends remote write messages to a server.

Maybe prometheus has changed its tune and it more inclined to encourage applications to use remote write; however as originally conceived, it seems that was not the intent. 

I think we should leave push out of the initial feature set, and potentially come back to it post release if there is demand.

jmacd:
The Prom pushgateway comes with a lot of caveats, for example:
https://github.com/prometheus/pushgateway#about-the-job-and-instance-labels

I would prefer to see us put our resources behind finishing push support for OTLP. Prometheus provides a way to join service-discovered attributes with application-provided attributes, and we would like a way to do this with OTLP. When we can do that with OTLP, then clients can send OTLP to a collector, have their service discovery labels applied, and then exported via Prometheus Remote Write. 

We have not discussed adding Prom pushgateway support in the collector, and I don't think we should.

jmacd:
(Issue about service discovery for pushed metrics: https://github.com/open-telemetry/opentelemetry-specification/issues/1078)

specification/metrics/sdk_exporters/prometheus.md
jmacd: I think we should remove ths paragraph and not mention push-prometheus because it comes with a lot of caveats and we would be better off designing a complete push-based metrics solution. See https://github.com/open-telemetry/opentelemetry-specification/issues/1078 and https://github.com/open-telemetry/oteps/pull/185

specification/metrics/sdk_exporters/prometheus.md
reyang: I've updated the PR. Regarding this comment I think we have several options:
1. Make the pull mode a `MUST`, which means JavaScript running inside a browser probably won't be able to support it (and will leave `N/A` in the feature matrix).
2. Use `SHOULD` + "if possible".

specification/metrics/sdk_exporters/prometheus.md
reyang: Thanks for the feedback, I've updated the PR and mentioned "Support for Push mode is out of scope for the initial release."

specification/metrics/sdk_exporters/prometheus.md
MrAlias: Since there isn't a requirement to implement these exporter, and this is written with the expected audience of implementations that can implement this I would prefer to go with option (1). That way it would be clearly defined that if an implementation supports a Prometheus exporter they MUST support pull based metrics. Having implementation that cannot support this report an `N/A` seems appropriate.

MrAlias: Saying "SHOULD if possible" is a poor way to specify a normative statement. If we want to go this route it would be better to describe the cases where it is not possible explicitly. Otherwise, this opens up the specification to ambiguity of subjective definition of what "if possible" is.

specification/metrics/sdk_exporters/prometheus.md
```diff
@@ -2,6 +2,25 @@
...
 **Status**: [Experimental](../../document-status.md)
 
-Prometheus Exporter is a [Pull Metric Exporter](../sdk.md#pull-metric-exporter)
-which reacts to the Prometheus scraper and report the metrics passively to
-[Prometheus](https://prometheus.io/).
+Prometheus Exporter is a [MetricExporter](../sdk.md#metricexporter) which reacts
+to the Prometheus scraper and reports the metrics passively to
+[Prometheus](https://prometheus.io/), or [pushes
+metrics](https://prometheus.io/docs/instrumenting/pushing/) to a [Prometheus
+Pushgateway](https://github.com/prometheus/pushgateway).
```
MrAlias, line 9: - The "Prometheus Exporter" noun phrase needs an indefinite article.
- Remove mention of the push functionality give we are not including it and it should not be a part of the definition.
- Clarify the request/response relationship with word choice.
- Clarify the data format being communicate (metric aggregate values, right?).

```suggestion
A Prometheus Exporter is a [MetricExporter](../sdk.md#metricexporter) which responds
to [Prometheus](https://prometheus.io/) scraper requests with collected metric aggregate
values.
```

reyang: If we put a strong wording which indicates Prometheus Exporter is a "Pull Exporter" (e.g. responds to scraper...), later (after initial Stable release) when we add push support it'll be a breaking change (so we have to either bump the major version of the spec, or we need to introduce another exporter, e.g. `PrometheusPushExporter`).

MrAlias: > ... or we need to introduce another exporter, e.g. `PrometheusPushExporter`

I would support this.

tigrannajaryan: FWIW, Pull and Push are also separate Prometheus exporters in Otel Collector: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter
The Push one is called prometheusremotewrite exporter.

specification/metrics/sdk_exporters/prometheus.md
reyang: I've updated the PR based on the discussions and comments. PTAL.

specification/metrics/sdk_exporters/prometheus.md
reyang: I've updated the PR based on the discussions and comments. PTAL.

specification/metrics/sdk_exporters/prometheus.md
```diff
@@ -2,6 +2,32 @@
...
+service-level batch jobs, JavaScript applications running inside web browsers,
+etc.).
+
+Prometheus Exporter MUST support version `0.0.4` of the [Text-based
+format](https://github.com/prometheus/docs/blob/main/content/docs/instrumenting/exposition_formats.md#text-based-format).
+
+Prometheus Exporter SHOULD only support [Cumulative
+Temporality](../datamodel.md#temporality). Attempt to change the [Aggregation
+Temporality](../datamodel.md#temporality) to
+[Delta](../datamodel.md#temporality) SHOULD be treated as an error. It is
```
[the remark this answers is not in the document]
reyang: I've updated the wording, PTAL.

specification/metrics/sdk_exporters/prometheus.md
```diff
@@ -3,5 +3,21 @@
 **Status**: [Experimental](../../document-status.md)
 
 Prometheus Exporter is a [Pull Metric Exporter](../sdk.md#pull-metric-exporter)
```
MrAlias, line 5:
```suggestion
A Prometheus Exporter is a [Pull Metric Exporter](../sdk.md#pull-metric-exporter)
```

specification/metrics/sdk_exporters/prometheus.md
```diff
@@ -3,5 +3,21 @@
...
-which reacts to the Prometheus scraper and report the metrics passively to
-[Prometheus](https://prometheus.io/).
+which reports metrics by responding to the [Prometheus](https://prometheus.io/)
+scraper requests.
+
+Prometheus Exporter MUST support [Pull mode](../sdk.md#pull-metric-exporter).
```
MrAlias, line 9:
```suggestion
A Prometheus Exporter MUST support [Pull mode](../sdk.md#pull-metric-exporter).
```

specification/metrics/sdk_exporters/prometheus.md
```diff
@@ -3,5 +3,21 @@
...
+
+Prometheus Exporter MUST NOT support [Push
```
MrAlias, line 11:
```suggestion
A Prometheus Exporter MUST NOT support [Push
```

specification/metrics/sdk_exporters/prometheus.md
```diff
@@ -3,5 +3,21 @@
...
+mode](../sdk.md#push-metric-exporter).
+
+Prometheus Exporter MUST only support [Cumulative
```
MrAlias, line 14:
```suggestion
A Prometheus Exporter MUST only support [Cumulative
```

specification/metrics/sdk_exporters/prometheus.md
```diff
@@ -3,5 +3,21 @@
...
+Temporality](../datamodel.md#temporality).
+
+Prometheus Exporter MUST support version `0.0.4` of the [Text-based
```
MrAlias, line 17:
```suggestion
A Prometheus Exporter MUST support version `0.0.4` of the [Text-based
```

specification/metrics/sdk_exporters/prometheus.md
```diff
@@ -3,5 +3,21 @@
...
+format](https://github.com/prometheus/docs/blob/main/content/docs/instrumenting/exposition_formats.md#text-based-format).
+
+Prometheus Exporter MAY support [OpenMetrics Text
```
MrAlias, line 20:
```suggestion
A Prometheus Exporter MAY support [OpenMetrics Text
```

cijothomas (approved):

jsuereth (approved):

jack-berg (approved):

jmacd (approved):

Issue #2130, opened 2021-11-16
Clarify the need and the difference between supported and preferred temporality?


bogdandrutu:
From https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#temporality-override-rules it seems that if there is a "preferred" use that -> it can translate into I support only the "preferred". No need for the both concepts if setting the "preferred" means will produce that anyway.

reyang:
@bogdandrutu had a question if we need both - if we need both and later we remove one, it'll be a breaking change.

PR #2132, opened 2021-11-16
Clarify aggregation temporality selection

### LINKED ISSUE #2129: Clarify SDK/MetricReader/MetricExporter boundaries

Clear design and terminology used. What is the "SDK", what are the components that an SDK communicates with, external boundaries, configurations, etc.

**Additional context.**

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#temporality-override-rules

It is very unclear if an SDK "knows" about the MetricExporter interface or just about the "MetricReader" interface. From all the examples and text in the https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md it seems that the SDK only exposes one "external point" which is the MetricReader interface, and some implementations on the MetricReader may expose a "MetricExporter" interface. If that is the case the "SDK" never sees directly an Exporter and does not know about temporality from there. So the SDK cannot and should not resolve any temporality conflict.

reyang:
Fixes #2129

## Changes

Clarified that the SDK has access to the MetricReader temporality, and the MetricReader could/should proxy the temporality of its underlying MetricExporter (if there is one).


jmacd (approved):

specification/metrics/sdk.md
```diff
@@ -564,7 +562,33 @@ functions.
...
+    refer to [Error handling in OpenTelemetry](../error-handling.md) for the
+    general guidance.
+* If the temporality is not explicitly specified:
+  * If the `MetricReader` only supports one temporality (either Cumulative or
+    Delta), use the supported temporality.
+  * If the `MetricReader` supports both Cumulative and Delta:
+    * If the `MetricReader` has a preferred temporality, use the preferred
+      temporality.
+    * If the `MetricReader` does not have a preferred temporality, use
+      Cumulative.
```
bogdandrutu, line 583: See this https://github.com/open-telemetry/opentelemetry-specification/issues/2130

reyang: I can handle #2130 either in this PR (by making the PR bigger) or as a separate PR after we land this one. Do you have a preference?

specification/metrics/sdk.md
```diff
@@ -564,7 +562,33 @@ functions.
...
+
+If a `MetricReader` instance has an underlying `MetricExporter` instance (e.g. a
```
jsuereth, line 585: Might want to call out what to do if one reader supports multiple exporters

jsuereth (approved):

specification/metrics/sdk.md
```diff
@@ -564,7 +562,33 @@ functions.
 
 The SDK SHOULD provide a way to allow [Aggregation
 Temporality](./datamodel.md#temporality) to be specified for a `MetricReader`
-instance during the creation time. [OpenTelemetry SDK](../overview.md#sdk)
+instance at the creation time. The following logic MUST be followed to determine
```
bogdandrutu, line 565: When "installed/applied" to the SDK correct?

specification/metrics/sdk.md
reyang: Good call, let me make an update later today.

specification/metrics/sdk.md
reyang: Right. In other places I've seen "SDK initialization" (which again is not super accurate). Let me see how to improve the wording.

reyang: I've tweaked the wording, PTAL.

specification/metrics/sdk.md
```diff
@@ -166,9 +165,8 @@ are the inputs:
     applies to [synchronous Instruments](./api.md#synchronous-instrument).
   * The `aggregation` (optional) to be used. If not provided, the SDK SHOULD
     apply a [default aggregation](#default-aggregation). If the aggregation has
-    temporality, the SDK SHOULD use the [temporality override
-    rules](#temporality-override-rules) to determine the aggregation
-    temporality.
+    temporality, the SDK SHOULD handle the aggregation temporality based on the
+    temporality of each [MetricReader](#metricreader) instance.
```
aabmass, line 169: I remember previously that some aggregations had a temporality option, but it seems like that option was removed from all of the aggregators. There is no mention of "temporality" in the Aggregation section anymore so I don't follow what this means.

specification/metrics/sdk.md
```diff
@@ -564,7 +562,34 @@ functions.
 
 The SDK SHOULD provide a way to allow [Aggregation
 Temporality](./datamodel.md#temporality) to be specified for a `MetricReader`
-instance during the creation time. [OpenTelemetry SDK](../overview.md#sdk)
+instance during the setup (e.g. initialization, registration, etc.) time. The
```
aabmass, line 565: Should this be called out in the capabilities section above?

specification/metrics/sdk.md
```diff
@@ -564,7 +562,34 @@ functions.
...
+    * If the `MetricReader` has a preferred temporality, use the preferred
+      temporality.
+    * If the `MetricReader` does not have a preferred temporality, use
+      Cumulative.
+
+If a `MetricReader` instance has an underlying `MetricExporter` instance (e.g. a
+[Periodic exporting MetricReader](#periodic-exporting-metricreader) configured
+with an [OTLP Exporter](./sdk_exporters/otlp.md)), the `MetricReader` SHOULD
+report its supported temporality and preferred temporality based on the
+temporality of the underlying `MetricExporter`.
```
aabmass, line 590: What if the temporality is explicitly specified for the `MetricReader` and it conflicts with the `MetricExporter`? It sounds like Periodic exporting MetricReader should not provide a way for temporality to be specified besides by proxying to the exporter

specification/metrics/sdk.md
reyang: Previously the temporality can be specified at fine granularity, for example - a `MeterProvider` could have 2 instruments configured (via view/aggregation) to pipe metrics to an OTLP exporter, the 1st instrument can result in a stream of Delta sum, the 2nd instrument can result in a stream of Cumulative sum (so the OTLP exporter will export two streams with different temporality). Also this would cause conflict - e.g. if the view/aggregation asked for Delta, but the reader/exporter only supports Cumulative. Later we collected the feedback and decided to simplify the model as there is no good use case (and if later there is, we can add it back without breaking existing stuff).

Now the model is - if a reader is Cumulative, the SDK should do necessary job to feed Cumulative value to the reader (e.g. convert the counter.Add increment to Cumulative). Here are more details https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/supplementary-guidelines.md#aggregation-temporality.

specification/metrics/sdk.md
reyang: Good point. I think either it doesn't provide such way to for temporality to be specified explicitly, or it generates error when temporality is provided explicitly other than implied by the exporter.

I think the current wording `SHOULD` is good enough? Do you feel that `MUST` is needed (I guess some languages might model this as a sealed derived class with temporality setting marked as `NotSupportedOperationException`).

specification/metrics/sdk.md
reyang: The way I think about this - it's an option for the capability "Collecting metrics from the SDK", instead of an independent capacity. Do you have a suggestion / strong opinion?

specification/metrics/sdk.md
jsuereth: @reyang You could rephrase this to say 

```
If the aggregation outputs metric points that use "AggregationTemporality" (e.g. Histogram, Sum) ...
```
We did remove aggregation temporality from view configuration, so from a specification perspective, we should make sure this section is not confusing.

specification/metrics/sdk.md
jsuereth: The way I'd phrase this:

```
If a MetricReader is backed by a MetricExporter it MUST use the underlying MetricExporter's preferred + supported temporality.
```

This is a MUST requirement with an 'if' caveat on when it applies.

cijothomas (approved):
Good clarifictions!

bogdandrutu (approved):

specification/metrics/sdk.md
aabmass: I agree capabilities is probably not the correct section. I don't have a strong opinion please resolve

specification/metrics/sdk.md
aabmass: > Good point. I think either it doesn't provide such way to for temporality to be specified explicitly [...]

I would also prefer that it doesn't provide a way to specify it for both, and I suppose the wording in this PR allows you to implement it in this way. Thanks for clarifying.

Both your wording and Josh's make sense to me, it was just a small nit.

specification/metrics/sdk.md
aabmass: Or can we remove this sentence altogether? It seems irrelevant since aggregations no longer have a temporality option.

reyang: I've made the wording adjustment based on @jsuereth's [comment](https://github.com/open-telemetry/opentelemetry-specification/pull/2132#discussion_r753222272).

specification/metrics/sdk.md
reyang: I took @jsuereth's wording from this [comment](https://github.com/open-telemetry/opentelemetry-specification/pull/2132#discussion_r753223681). (thanks @jsuereth!)

specification/metrics/sdk.md
reyang: Based on the discussion in the SIG meetings regarding #2130 and the offline discussion, we will follow up outside this PR.

1. The SDK spec can be simplified by using one setting instead of two (preferred + supported).
2. A reader might be able to output to multiple exporters (e.g. PeriodMetricReader -> Send to OTLP, also send to local logs)

reyang: Related to #2156.

PR #2070, opened 2021-10-27
Add database semantic convention section and redis server metrics

codeboten:
## Changes

Adding a section for semantic conventions around database metrics. Additionally capturing the redis metrics produced by the otel collector [redis receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/redisreceiver/metric_functions.go) today as a first step.

Related issues # https://github.com/open-telemetry/opentelemetry-specification/issues/1012

Issue #1367, opened 2021-01-22
Move semantic conventions that are applicable to the Resource from "trace" spec to a common area


tigrannajaryan:
Certain conventions, such as `db.*`, `messaging.*` are in the "trace" section of the spec. These are applicable to the Resource as well, not just traces. 

For example metrics emitted by / collected from Redis should have "db.system=redis" Resource attribute.

We need to move them to a common area of semantic conventions.

kenfinnigan:
Is this the case for messaging?

`messaging.*` in trace semantic conventions is referring to where a message was received from, or being sent to, it does not refer to the existing process creating the span.

As I understood `Resource` to be referencing the entity producing the span, it would not reference `messaging.system`, for example.

Oberon00:
The same applies to "db.". What would it mean to set "db.name" on a process? Is the process a database? Does it only ever communicate with that database?

Oberon00:
Just saw this is a quite old issue. Is it still needed?

tigrannajaryan:
@open-telemetry/specs-approvers how do we define semantic conventions for attributes that may be recorded both at the Resource level and at the Span or Metric or Log level?

For example we have `db.redis.*` and `db.*` conventions defined but for some reason they are defined as trace attributes. I don't think they are applicable to traces only. What do we do if we want to record database metrics, for example? I think `db.*` attributes should be possible to use in the Resource too, certainly also in Log Records, maybe also in Metric attributes.

This discussion is related to [this PR](https://github.com/open-telemetry/opentelemetry-specification/pull/2145) wants to add `redis.instance` as a Resource attribute, but I think it should be `db.redis.instance`. However, this in turn requires us to make a decision about moving (or copying?) `db.*` namespace to the Resource semantic conventions.

pyohannes:
I think @tedsuo started some work to refactor semantic conventions, one goal being to represent metric, trace, and resource conventions in a single model: https://github.com/open-telemetry/opentelemetry-specification/pull/1977#issuecomment-930514401

Unfortunately, this PR went stale. I think the approach in the PR was a good start for solving the problem this issue is about.

iNikem:
> The same applies to "db.". What would it mean to set "db.name" on a process? Is the process a database? Does it only ever communicate with that database?

I think comment from @Oberon00 is very relevant. What if my services talks to several databases. How exactly db attributes should look like on the Resource?

tigrannajaryan:
[...]

PR #2145, opened 2021-11-22
Add semantic conventions for redis receiver

manang-splunk:
## Changes

Added new resource attribute `redis.instance` for redis receiver.


specification/resource/semantic_conventions/redis.md
```diff
@@ -0,0 +1,13 @@
...
+**Status**: [Experimental](../../document-status.md)
+
+**type:** `redis`
+
+**Description:** A redis instance
+
+<!-- semconv redis -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `redis.instance` | string | Reported name of the redis instance | `localhost:6379` | No |
```
arminru, line 12: Hi @manang-splunk! Could you please elaborate on how this attribute would be used?
As this is a resource attribute, it would have to be populated by the Redis instance itself, i.e., an OpenTelemetry instrumentation built into Redis server directly. If this is your intention, please clarify it accordingly.

manang-splunk: `redis.instance` attribute would be used as a resource attribute whose value is configurable by user. This attribute was missing from redis metrics so we added it as resource attribute.

manang-splunk: @arminru Please share your thoughts on this.

arminru: @manang-splunk  With Redis metrics you are referring to @codeboten's open PR #2070 and this collector plugin, right?
https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/redisreceiver

manang-splunk: @arminru I am referring to PR [#5922](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/5922) where we have added a resource attribute `redis.instance` in code and we are simply adding its semantic convention for the same in this PR. Please go through PR 5922 to understand the usage of `redis.instance` attribute.

arminru: Thanks for the reference @manang-splunk, now it makes more sense to me!
I wonder if the `service.name` and `service.instance.id` resource attributes defined [here](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions#service) and suggested by @bogdandrutu on your PR would be sufficient or not, have you looked into these? They would be used to describe a "logical" name for the service and are expected to be provided by users.
If the value is supposed to describe the host on which the Redis instance is running like you indicate in your example, the `host.name` attribute defined [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/host.md#host) could be suitable but having the port in there might be unexpected, even if the definition says "_or another name specified by the user_".
@bogdandrutu WDYT?

manang-splunk: Hello @arminru ,
CC: @bogdandrutu 

The current task is to add a resource attribute in the [Redis Receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/redisreceiver) similar to what is existing in the SignalFx agent [Collectd/Redis monitor](https://github.com/signalfx/signalfx-agent/blob/ad8dd278074424b70804e3f9b427fcd99da74193/pkg/monitors/collectd/redis/redis.go#L42-L44).

The SignalFx agent [Collectd/Redis monitor](https://github.com/signalfx/signalfx-agent/blob/ad8dd278074424b70804e3f9b427fcd99da74193/pkg/monitors/collectd/redis/redis.go#L42-L44) has a [plugin_instance](https://github.com/signalfx/redis-collectd-plugin/blob/master/redis_info.py#L181-L185) attribute which expects a value in the format of `{host}:{port}` or a normal string. This helps to label the metrics with user-configured values. Similar functionality does not exist today in the [Redis receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/redisreceiver)  and needs to be implemented as a part of the current task.

To implement this functionality, we have the below two possibilities :

1. Option 1 - We can leverage existing resource attributes in the [Redis receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/redisreceiver) such that it can accommodate the value of both the formats `{host}:{port}` or a normal string based on the above requirements. We have three attributes, [service.name, service.instance.id](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions#service) and [host.name](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/host.md#host) but if `{host}:{port}` format of value gets populated within these existing attributes, then it would seem unexpected as the data format `{host}:{port}` is more like an endpoint type of data rather than service-name-related, service-instance-id-related or only host-name-related data.  Hence, this option is not suggested.

2. Option 2 - We can introduce a new resource attribute called `redis.instance` which can accommodate both these formats `{host}:{port}` or a normal string. Hence, this option is suggested and this logical workflow is implemented as a part of this task.

Please let us know your inputs or queries.

arminru: Thanks for the explanation.
I looked at `service.instance.id` again and agree that it is not really suitable as it is intended to distinguish multiple instances of the same service, so this would only be applicable for Redis cluster nodes.
Having just the form `host:port` for `service.name` is not ideal indeed as this does not correspond well with the definition of a logical service name. If we'd add `redis.instance`, which `service.name` would the Redis receiver set?

specification/resource/semantic_conventions/redis.md
```diff
@@ -0,0 +1,13 @@
+# Redis
```
arminru, line 1: Please reference the new file in the list of contents in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md and also add an entry for your changes in https://github.com/open-telemetry/opentelemetry-specification/blob/main/CHANGELOG.md (Unreleased section).

semantic_conventions/resource/redis.yaml
```diff
@@ -0,0 +1,11 @@
+groups:
+  - id: redis
+    prefix: redis
+    brief: >
+        A redis instance
+    attributes:
+      - id: instance
+        type: string
+        brief: >
+          Reported name of the redis instance
```
arminru, line 10:
```suggestion
          Reported name of the Redis instance. This can be in the form `{host}:{port}` or any other name provided manually when setting up the instrumentation.
```
Not yet happy with my suggestion but this lacks clarification. I'm open for better suggestions :)

semantic_conventions/resource/redis.yaml
```diff
@@ -0,0 +1,11 @@
...
+        examples: ['localhost:6379']
```
arminru, line 11:
```suggestion
        examples: ['localhost:6379', 'product_info_redis']
```
To make clear it cannot only be in that form but any name.

specification/resource/semantic_conventions/redis.md
arminru: I'm open for adding `redis.instance` if we don't see a good fit with the existing attributes but would like to get more opinions from the other @open-telemetry/specs-approvers. This would be the first technology/framework-specific resource attribute to be added and I wonder if we were already able to cover similar cases with the existing, generic attributes.

tigrannajaryan:
We have conventions for databases: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md which uses prefix `db.redis.*` for Redis attributes. Should this attribute be under `db.redis.*`?

Side note: I am not sure it is a good idea to have `db.<dbname>.*` namespace when we also have `db.*` namespace. This has a lot of potential for name clashes. These conventions may need to be revamped.

specification/resource/semantic_conventions/redis.md
manang-splunk: The usage of `service.name` attribute is user specific. Based on the observed data of the Redis server, there is no service-name-related data and as per our usecase `host:port` doesn't fit with `service.name`. Hence, we are not setting the `service.name` attribute. I have attached the sample data we observed from the Redis server. [data.txt](https://github.com/open-telemetry/opentelemetry-specification/files/7649848/data.txt)

specification/resource/semantic_conventions/redis.md
manang-splunk: Done, I have added reference of new file in [README.md](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md) and added the changes in [CHANGELOG.md](https://github.com/open-telemetry/opentelemetry-specification/blob/main/CHANGELOG.md)

semantic_conventions/resource/redis.yaml
manang-splunk: Done

semantic_conventions/resource/redis.yaml
manang-splunk: Done, added an appropriate description.

hprajapati-splunk:
> We have conventions for databases: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md which uses prefix `db.redis.*` for Redis attributes. Should this attribute be under `db.redis.*`?
> 
> Side note: I am not sure it is a good idea to have `db.<dbname>.*` namespace when we also have `db.*` namespace. This has a lot of potential for name clashes. These conventions may need to be revamped.

Hello @tigrannajaryan,
We have added `redis.instance` as a resource attribute. The [database semantics](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md) you referred are for trace attributes, which are used for database client calls. If we add `db.redis.*` in the resource attribute then it will have conflicts with `db.*` namespace as you mentioned earlier, so we have added `redis.instance` (it can be an endpoint or any string value provided by the user). You can refer [this conversation](https://github.com/open-telemetry/opentelemetry-specification/pull/2145#discussion_r757516366) for more details about the field we have added(`redis.instance`).

Please let us know your thoughts.

tigrannajaryan:
> The [database semantics](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md) you referred are for trace attributes, which are used for database client calls. If we add `db.redis.*` in the resource attribute then it will have conflicts with `db.*` namespace as you mentioned earlier

I do not see any reason to limit `db.*` attributes to traces only. They can be also valid for metrics describing databases and I would expect for example `db.system` or `db.name` to be recorded in the Resource for metrics. I think we can have `db.*` namespace in the Resource, which can also contain `db.redis.instance` attribute.

This raises a good question though: how do we define semantic conventions for attributes that may be recorded both at the Resource level and at the Span or Metric level? I don't think Otel spec has a good answer for this and we have an open issue about that https://github.com/open-telemetry/opentelemetry-specification/issues/1367

To make progress on this I suggest the following. Change this PR to use `db.redis.instance`. I do not think top-level `redis.*` is acceptable. We can accept this as experimental convention. Be prepared that the convention may be changed when https://github.com/open-telemetry/opentelemetry-specification/issues/1367 is resolved.

manang-splunk:
> > The [database semantics](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md) you referred are for trace attributes, which are used for database client calls. If we add `db.redis.*` in the resource attribute then it will have conflicts with `db.*` namespace as you mentioned earlier
> 
> I do not see any reason to limit `db.*` attributes to traces only. They can be also valid for metrics describing databases and I would expect for example `db.system` or `db.name` to be recorded in the Resource for metrics. I think we can have `db.*` namespace in the Resource, which can also contain `db.redis.instance` attribute.
> 
> This raises a good question though: how do we define semantic conventions for attributes that may be recorded both at the Resource level and at the Span or Metric level? I don't think Otel spec has a good answer for this and we have an open issue about that #1367
> 
> To make progress on this I suggest the following. Change this PR to use `db.redis.instance`. I do not think top-level `redis.*` is acceptable. We can accept this as experimental convention. Be prepared that the convention may be changed when #1367 is resolved.

We tried to add `db.redis.instance` with namespace as `db.*` as well as with `db.redis.*` at the resource level, we got the following error while running `make table-generation` command as the namespace already exists at the trace level:
```
Error parsing /source/trace/database.yaml

Semantic convention 'db' is already defined.
```

```
Error parsing /source/trace/database.yaml

Semantic convention 'db.redis' is already defined.
```

Is there something that we are missing from our end?

tigrannajaryan:
@manang-splunk I am not sure what the problem is, perhaps the semantic convention generator has limitations that prevent having the same namespace for resources and traces. This may require fixing the generator first. @open-telemetry/build-tools-approvers any thoughts?

tigrannajaryan:
@manang-splunk can you create an issue describing the limitation of the semantic convention generator tool, along with a proposal of how to fix it (if you have one)?

manang-splunk:
> @manang-splunk can you create an issue describing the limitation of the semantic convention generator tool, along with a proposal of how to fix it (if you have one)?

@tigrannajaryan As we were setting group id as “db.redis” and it was already used at trace level we were getting error that “db.redis” is already there. So we used different group id as “redis” for this. Now we are able to add the “db.redis.instance” as a resource semantic convention. We have pushed the changes accordingly can you please review it?

As it is a type of database, we have created a folder “db” (semantic_conventions/resource/db) and kept conventions of Redis under that. Any other database conventions can be added under that in future.

tigrannajaryan (approved):
The namespace and the name of the new attribute look good to me.
I do not know much of Redis to tell if the attribute value is what we want to record. I would like others to chime in too.

tigrannajaryan:
@open-telemetry/specs-approvers @open-telemetry/instr-wg please review.

semantic_conventions/resource/db/redis.yaml
```diff
@@ -0,0 +1,14 @@
+groups:
+  - id: redis
+    prefix: db.redis
```
joaopgrassi, line 3: I quickly searched and I didn't find another convention that where the `id` is at a "different level" than the `prefix`. I'm not sure if this is a problem though, just something that stand up for me.

I guess you will also get a conflict if you have a structure like this: `id: db > id: db.redis > attributes > id: instance`? Similar to this https://github.com/open-telemetry/opentelemetry-specification/blob/d7a6b9f012aa3e3388ed8af12e5cb1b3fac9b3e5/semantic_conventions/trace/database.yaml#L237
?

semantic_conventions/resource/db/redis.yaml
```diff
@@ -0,0 +1,14 @@
...
+    brief: >
+        A redis instance
+    attributes:
+      - id: instance
+        type: string
+        brief: >
+          Reported name of the Redis instance. This can be in the form
```
joaopgrassi, line 10:
```suggestion
          The reported name of the Redis instance. This can be in the form of
```

semantic_conventions/resource/db/redis.yaml
```diff
@@ -0,0 +1,14 @@
...
+          `{host}:{port}` or any other name provided manually while configuring
+          the instrumentation and defaults to the `endpoint` value provided in the
```
joaopgrassi, line 12: I think the sentence is a bit too long and hard to follow. A suggestion would be maybe ending the sentence in `the instrumentation.` and starting a new one with. `The default value is the endpoint... ` or `When not provided, the default value is...`. What do you think?

joaopgrassi (commented):
I'm not familiar with the Redis receiver but I went through the PRs and I think this makes sense. Just some editorial suggestions and a question about the structure of the convention.

specification/resource/semantic_conventions/db/redis.md
```diff
@@ -0,0 +1,13 @@
+# Redis
+
+**Status**: [Experimental](../../document-status.md)
```
joaopgrassi, line 3: The link is missing one more level I think (that's why the docfx check is failing)

manang-splunk: Pushed the changes accordingly.

semantic_conventions/resource/db/redis.yaml
manang-splunk: Pushed the changes accordingly.

semantic_conventions/resource/db/redis.yaml
manang-splunk: Pushed the changes accordingly.

semantic_conventions/resource/db/redis.yaml
manang-splunk: Yes, I was getting conflicts while using id as `db` and `db.redis`. 
As we already have `db` and `db.redis` as id value in [database](https://github.com/manang-splunk/opentelemetry-specification/blob/9faf40fa2ba5a6ce73be7f4f0c002133ea33872c/semantic_conventions/trace/database.yaml#L2) attributes, we cannot use the same here. Therefore we have kept id as `redis`.

joaopgrassi:
Sorry for the back and forth, but I'm not sure the current solution (due to the conflict in the namespace) is the best. As @tigrannajaryan mentioned, we might need to change the build-tool or how we invoke it.

I tried something quickly and I managed to call the `table-generation` without the conflict, but it's also a bit "hackish". If you change your `redis.yaml` to have

```yaml
id: db.redis
    prefix: db.redis
    ...
 ```
 
 Then running the `make table-generation` will fail with the conflict. As far as I could see, that's because the command invokes the generation tool passing the whole semconv folder which contains `traces` and `resources`. If you change it, passing the `resource` folder alone it works. Like this:
 
 ```yaml
 # Generate markdown tables from YAML definitions
.PHONY: table-generation
table-generation:
	docker run --rm -v $(PWD)/semantic_conventions/resource:/source -v $(PWD)/specification/resource:/spec \
		otel/semconvgen:$(SEMCONVGEN_VERSION) -f /source markdown -md /spec
```

This will generate the table on `db/redis.md` correctly. 

I'm not sure if this is a feasible solution though. Maybe @Oberon00 can chime in and share what he thinks about it. Maybe we will really need to change the generator.

Issue #2147, opened 2021-11-22
Mark the OpenTracing compatibility section as stable


carlosalberto:
This issue will serve to track the remaining work to mark the OpenTracing compatibility as stable:

 - [x] Add OpenTracing to the compliance Matrix (#2046)
 - [x] **Only** error mapping should happen in the Shim (#2148)
 - [x] Finalize defining Baggage support (#2137)
 - [x] Decide whether a common `tracer name` should be used when [creating the Shim](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/opentracing.md#create-an-opentracing-tracer-shim)
 - [x] Document what happens when a `null` value is passed to ScopeManager.activate() (see https://github.com/open-telemetry/opentelemetry-java/issues/3952)
 - [x] Add a missing **Start Span** section (addressed by #2228)
 - [x] Document version compatibility among different runtime versions (#2218)
 - [x] Document the supported capabilities for OT languages that never implemented `ScopeManager` (#2219)
 - [ ] Support Baggage-only Propagation of `SpanContext` (#2326)
  
The plan has been updated to mark the Spec as stable in _very early_ Q1, and have the first compliant implementations ready by Q1 a well (Python and Java for a start).

@bhs @yurishkuro @tedsuo Please review in case anything else needs to be added.

carlosalberto:
@yurishkuro Please review, as I've update this with a few issues that appeared lately.

carlosalberto:
We have completed all the important tasks, and I'd only wait for #2298 and #2297 as final improvements before we mark this section stable (we could work on the porting guides/documentation after the release).

Opinions @bhs @yurishkuro @tedsuo ?

bhs:
Mega-apologies for the delay on this. I'm fine marking this as stable, particularly given that OT is now officially archived.

yurishkuro:
sgtm

jack-berg:
I've been thinking a bit about this PR to implement [Tracer.close()](https://github.com/open-telemetry/opentelemetry-java/pull/5151) in the java OpenTracing shim. 

The [OpenTracing shim](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/opentracing.md#abstract) spec says:

> This functionality will be provided as a bridge layer implementing the [OpenTracing API](https://github.com/opentracing/specification) using the OpenTelemetry API. This layer MUST NOT rely on implementation specific details of any SDK.

[...]

PR #2148, opened 2021-11-23
State that ONLY error mapping can happen in the OT Shim layer.

carlosalberto:
Addresses part of #2147 by stating that ONLY error mapping must happen in the OT Shim layer, as opposed to do any other semantic convention mapping.

cc @jmacd @yurishkuro

### DIFF
--- a/specification/compatibility/opentracing.md
+++ b/specification/compatibility/opentracing.md
@@ -48,7 +48,7 @@ The OpenTracing Shim and the OpenTelemetry API/SDK are expected to be consumed
 simultaneously in a running service, in order to ease migration from the former
 to the latter.
 
-Observe that semantic convention mapping MUST NOT be performed, with the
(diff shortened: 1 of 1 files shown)


specification/compatibility/opentracing.md
```diff
@@ -48,6 +48,10 @@ The OpenTracing Shim and the OpenTelemetry API/SDK are expected to be consumed
 simultaneously in a running service, in order to ease migration from the former
 to the latter.
 
+Observe that semantic convention mapping MUST NOT be performed, with the
```
yurishkuro, line 51:
```suggestion
Semantic convention mapping SHOULD NOT be performed, with the
```

yurishkuro (approved):

yurishkuro (approved):

tedsuo (approved):
LGTM

PR #2154, opened 2021-11-23
Remove the concept of supported temporality, keep preferred

bogdandrutu:

Fixes https://github.com/open-telemetry/opentelemetry-specification/issues/2130

### DIFF
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -561,33 +561,11 @@ The SDK SHOULD provide a way to allow `MetricReader` to respond to
 idiomatic approach, for example, as `OnForceFlush` and `OnShutdown` callback
 functions.
 
-The SDK SHOULD provide a way to allow [Aggregation
+The SDK SHOULD provide a way to allow the preferred [Aggregation
 Temporality](./datamodel.md#temporality) to be specified for a `MetricReader`
-instance during the setup (e.g. initialization, registration, etc.) time. The
-following logic MUST be followed to determine which temporality to be used for a
-`MetricReader`:
-
-* If the temporality is explicitly specified during `MetricReader` creation:
-  * If the specified temporality is supported by the `MetricReader`, use the
-    specified temporality.
-  * If the specified temporality is not supported by the `MetricReader`, treat
-    the conflicts as an error. It is unspecified _how_ these error should be
-    handled (e.g. it could fail fast during the SDK configuration time). Please
-    refer to [Error handling in OpenTelemetry](../error-handling.md) for the
-    general guidance.
-* If the temporality is not explicitly specified:
-  * If the `MetricReader` only supports one temporality (either Cumulative or
-    Delta), use the supported temporality.
-  * If the `MetricReader` supports both Cumulative and Delta:
-    * If the `MetricReader` has a preferred temporality, use the preferred
-      temporality.
-    * If the `MetricReader` does not have a preferred temporality, use
-      Cumulative.
-
-If a `MetricReader` is backed by a `MetricExporter` (e.g. a [Periodic exporting
-MetricReader](#periodic-exporting-metricreader) configured with an [OTLP
-Exporter](./sdk_exporters/otlp.md)) it MUST use the underlying
-`MetricExporter`'s preferred + supported temporality.
+instance during the setup (e.g. initialization, registration, etc.) time. If the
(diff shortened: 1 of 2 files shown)


specification/metrics/sdk.md
```diff
@@ -561,33 +561,11 @@ The SDK SHOULD provide a way to allow `MetricReader` to respond to
 idiomatic approach, for example, as `OnForceFlush` and `OnShutdown` callback
 functions.
 
-The SDK SHOULD provide a way to allow [Aggregation
+The SDK SHOULD provide a way to allow preferred [Aggregation
```
reyang, line 564: nit:

```suggestion
The SDK SHOULD provide a way to allow the preferred [Aggregation
```
```suggestion
The SDK SHOULD provide a way to allow preferred [Aggregation
```

reyang (approved):
LGTM. Need to update the changelog.

jsuereth (approved):

cijothomas (approved):

specification/metrics/sdk.md
```diff
@@ -561,33 +561,11 @@ The SDK SHOULD provide a way to allow `MetricReader` to respond to
...
-    * If the `MetricReader` does not have a preferred temporality, use
-      Cumulative.
-
-If a `MetricReader` is backed by a `MetricExporter` (e.g. a [Periodic exporting
-MetricReader](#periodic-exporting-metricreader) configured with an [OTLP
-Exporter](./sdk_exporters/otlp.md)) it MUST use the underlying
-`MetricExporter`'s preferred + supported temporality.
+instance during the setup (e.g. initialization, registration, etc.) time. If the
+preferred temporality is explicitly specified then the SDK SHOULD respect that,
+otherwise use Cumulative.
```
MrAlias, line 568: We should strengthen the normative requirements here to ensure compativility and keep error behavior explicitly to avoid ambiguity.

```suggestion
preferred temporality is explicitly specified then the SDK MUST provide data using that temporality if it can.
If the SDK is not capable of providing data with the explicitly specified temporality it MUST be treated as an error.
Otherwise, if no preferred temporality is specified, the SDK SHOULD provide data with Cumulative temporality.
```

bogdandrutu: > MUST provide data using that temporality if it can.

MUST and "if it can" are kind of opposite, is it a MUST or not.

The reason not to use must is because 20 lines below we have:

> [OpenTelemetry SDK](../overview.md#sdk)
authors MAY choose the best idiomatic design for their language:

* Whether to treat the temporality settings as recommendation or enforcement.
  For example, if the temporality is set to Delta, would the SDK want to perform
  Cumulative->Delta conversion for an [Asynchronous
  Counter](./api.md#asynchronous-counter), or downgrade it to a
  [Gauge](./datamodel.md#gauge), or keep consuming it as Cumulative due to the
  consideration of [memory
  efficiency](./supplementary-guidelines.md#memory-management)?

MrAlias: > > MUST provide data using that temporality if it can.
> 
> MUST and "if it can" are kind of opposite, is it a MUST or not.
> 
> The reason not to use must is because 20 lines below we have:
> 
> > [OpenTelemetry SDK](../overview.md#sdk)
> > authors MAY choose the best idiomatic design for their language:
> 
> * Whether to treat the temporality settings as recommendation or enforcement.
>   For example, if the temporality is set to Delta, would the SDK want to perform
>   Cumulative->Delta conversion for an [Asynchronous
>   Counter](./api.md#asynchronous-counter), or downgrade it to a
>   [Gauge](./datamodel.md#gauge), or keep consuming it as Cumulative due to the
>   consideration of [memory
>   efficiency](./supplementary-guidelines.md#memory-management)?

Gotcha :+1: . Thoughts on being explicit with the errors response?

pirgeo: I think most (all?) implementations currently do the cumulative -> delta conversion, right? This would not be an error state. So since everything is allowed, do you mean that there should be a standard way of letting users know what is happening?

bogdandrutu: golang does not, and probably other languages may not do as well (not sure).

MrAlias: > Thoughts on being explicit with the errors response?

Wait, how was this resolved?

PR #2161, opened 2021-11-24
Change some would be -> MUST in metrics spec.

anuraaga:
`would be` seems to be wording that would be intended for an end user, but the spec is for implementation authors so MUST seems like what was intended (though let me know if it's SHOULD)

### DIFF
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -150,11 +150,11 @@ are the inputs:
     refer to [Error handling in OpenTelemetry](../error-handling.md) for the
     general guidance.
 * The `name` of the View (optional). If not provided, the Instrument `name`
-  would be used by default. This will be used as the name of the [metrics
+  MUST be used by default. This will be used as the name of the [metrics
   stream](./datamodel.md#events--data-stream--timeseries).
 * The configuration for the resulting [metrics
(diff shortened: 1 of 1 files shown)


cijothomas (approved):

reyang:
Please find my comment on another PR https://github.com/open-telemetry/opentelemetry-specification/pull/2163#discussion_r755725126.

anuraaga:
@reyang My impression from reading this was it's the behavior for a View that matched - so I think RFC wording makes more clear what to do when a view matches

reyang:
> @reyang My impression from reading this was it's the behavior for a View that matched - so I think RFC wording makes more clear what to do when a view matches

Sounds good to me (and I've expressed the same [here](https://github.com/open-telemetry/opentelemetry-specification/pull/2163#discussion_r755725126)).

reyang (approved):

carlosalberto (approved):

MrAlias (approved):

bogdandrutu (approved):

jsuereth (approved):

pirgeo (approved):

tigrannajaryan (approved):

jmacd (approved):

PR #2162, opened 2021-11-24
Rename extra dimensions to extra attributes

anuraaga:
I don't think there is a very clear definition of the word "dimension" in the spec. I did find one line that seems to indicate it's an alias of the word "attribute"

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#timeseries-model

It seems like the spec would be more consistent using the word attribute here.


specification/metrics/sdk.md
```diff
@@ -160,8 +160,8 @@ are the inputs:
     not in the list will be ignored. If not provided, all the attribute keys
     will be used by default (TODO: once the Hint API is available, the default
     behavior should respect the Hint if it is available).
-  * The `extra dimensions` which come from Baggage/Context (optional). If not
-    provided, no extra dimension will be used. Please note that this only
+  * The `extra attributes` which come from Baggage/Context (optional). If not
```
reyang, line 163: This might introduce confusion.

The name `dimension` was used here since Baggage is a string->string map, not the normal [attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attribute). And they could have different implications such as limits.

anuraaga: They still get populated as Attributes in data model, notably OTLP, though don't they? I think baggage being a restricted data model is a feature of baggage and it's natural to expect it to be populated as string attributes when desired. I personally wouldn't expect any confusion due to baggage being a string->string map.

I'm not sure about what implications there could be as far as limits go - it seems for now metrics SDK defines no limits anyways though.

reyang: I understand.

So far this is what we have:

The API will report [measurements](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#measurement).
The SDK will receive measurements (that have attributes), and extra **dimensions** from the context/baggage (if the user asked for them).

I don't have strong opinion though, these seems to be editorial / naming thing.

bogdandrutu: @anuraaga I want to make sure that we fix all places where this is used. Thank you

pirgeo: There are actually quite a few places where `dimension` is used, e.g. [further up in the view section](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md?plain=1#L137), where it seems to be used with mixed meanings. I would lean towards also using attribute there. It's also used quite heavily in [the supplementary guidelines](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/supplementary-guidelines.md), however I think there it feels more natural to use dimension over attribute.

anuraaga: Thanks - I went ahead and changed more instances. Going through the supplementary guidelines, I figured attribute makes sense, especially things like `attributes: { verb: GET }` - those are literally the attributes the instrumentation will add. A few places I changed dimension to "set of attributes" which seemed to be the actual meaning.

By the way I found one usage that I kept since it seems to refer to the mathematical concept of dimension, the unit of a number
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/README.md#instrument-units

reyang (approved):
LGTM.

pirgeo (approved):

bogdandrutu (approved):

cijothomas (approved):

cijothomas (approved):

jmacd (approved):

PR #2163, opened 2021-11-24
Clarify SDK must apply a default aggregation if not specified.

anuraaga:
If the SDK doesn't apply a default aggregation, what can it do? This seemed like a MUST to me but if not, I think this could be changed to also mention a potential alternative.

### DIFF
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -163,7 +163,7 @@ are the inputs:
   * The `extra dimensions` which come from Baggage/Context (optional). If not
     provided, no extra dimension will be used. Please note that this only
     applies to [synchronous Instruments](./api.md#synchronous-instrument).
-  * The `aggregation` (optional) to be used. If not provided, the SDK SHOULD
+  * The `aggregation` (optional) to be used. If not provided, the SDK MUST
     apply a [default aggregation](#default-aggregation). If the aggregation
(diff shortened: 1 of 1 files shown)


cijothomas (approved):

reyang (approved):

specification/metrics/sdk.md
```diff
@@ -163,7 +163,7 @@ are the inputs:
   * The `extra dimensions` which come from Baggage/Context (optional). If not
     provided, no extra dimension will be used. Please note that this only
     applies to [synchronous Instruments](./api.md#synchronous-instrument).
-  * The `aggregation` (optional) to be used. If not provided, the SDK SHOULD
+  * The `aggregation` (optional) to be used. If not provided, the SDK MUST
```
reyang, line 166: IIRC the original wording was taking the consideration that View might fail (e.g. a View with a specific name might never match with an actual instrument).

There is similar approach, for example:

_If not provided, the Instrument `name` would be used by default._

instead of:

_If not provided, the Instrument `name` MUST be used by default._

carlosalberto (approved):

MrAlias (approved):

jsuereth (approved):

pirgeo (approved):

jmacd (approved):

PR #2164, opened 2021-11-24
Use RFC SHOULD NOT wording in histogram aggregation

anuraaga:
The main motivation is fixing the case of the RPC wording, but also thought keeping the note about sum next to the information about sum to be clearer.

Also added a spelling / grammar fix that IDEA was showing


cijothomas (approved):

reyang (approved):

specification/metrics/sdk.md
```diff
@@ -386,22 +386,20 @@ This Aggregation honors the following configuration parameters:
...
 | RecordMinMax | true, false | true | Whether to record min and max. |
 
-Note: This aggregator should not fill out `sum` when used with instruments
-that record negative measurements, e.g. `UpDownCounter` or `ObservableGauge`.
-
 This Aggregation informs the SDK to collect:
 
 - Count of `Measurement` values falling within explicit bucket boundaries.
-- Arithmetic sum of `Measurement` values in population.
+- Arithmetic sum of `Measurement` values in population. This SHOULD NOT be collected when used with
```
reyang, line 392: There are discussion about whether we could support negative values in histogram.
Would this change prevent negative value being supported in the future (as additive changes)?

anuraaga: As far as I know SHOULD NOT can be changed to SHOULD, but SHOULD NOT cannot be changed to MUST. I don't know of a wording that can better capture the potential for change post-release than this unfortunately.

reyang: @jmacd @jsuereth @pirgeo do you think preventing negative value in Histogram is a no-go?

specification/metrics/sdk.md
```diff
@@ -386,22 +386,20 @@ This Aggregation honors the following configuration parameters:
...
 - Max (optional) `Measurement` value in population.
 
 ## Attribute limits
 
 Attributes which belong to Metrics are exempt from the
 [common rules of attribute limits](../common/common.md#attribute-limits) at this
-time. Attribute truncation or deletion could affect identitity of metric time
-series and it requires further analysis.
+time. Attribute truncation or deletion could affect identity of metric time
+series and the topic requires further analysis.
```
bogdandrutu, line 402: Good change, usually typos like this can be submitted as different PRs since simpler to merge.

specification/metrics/sdk.md
bogdandrutu: For the moment that breaks Prometheus.

jsuereth: On the fence here myself.   Once we have the support in Histograms for the prometheus case, this is actually fine, and I think we can encode Histograms of UpDownCounter/ObservableGauage into prometheus and things will work out (effectively you just remove `sum` metric sample), and if we need we can bundle negative measurements into one large `<= 0` bucket.

if SHOULD NOT can become SHOULD, this is fine.

pirgeo: I would prefer not to prevent negative bucket boundaries generally. Sums for negative buckets are only a problem with Prometheus, if i understand correctly? Otherwise, I think its fine, and if we can change SHOULD NOT to SHOULD I don't see a problem here.

specification/metrics/sdk.md
reyang: Related to #2174.

specification/metrics/sdk.md
jmacd: I agree with this thread. We can change SHOULD NOT to SHOULD later. I also support the idea that we can extend the data model with information to distinguish histogram instruments used with and without negative measurements. I had a draft OTEP (https://github.com/open-telemetry/oteps/pull/156) which I decided not to pursue further -- the conclusion was that we could in a backward compatible way extend the current histogram data point "temporality" field with more possible configurations to reflect more semantics.

The reason I hesitate to invest in deeper conversation is that there are a number of crossing concerns. We don't have a synchronous Gauge instrument, and histogram is a synchronous instrument, so a histogram of temporally-aggregated synchronous Gauge measurements can't actually happen even though I know what I mean when I write about one. Aggregation temporality for a histogram of Gauge values is not defined presently, but it could mean two things depending on whether we're talking about synchronous measurements or asynchronous measurements. In order to a develop an understanding of temporality for a "histogram of asynchronous gauge measurements" (which could happen for a view of an async gauge, see #2208), we are beginning to distinguish spatial vs. temporal aggregation. We're going to need gauge histogram, which I want us to model as a histogram where the counts are not monotonic: see https://github.com/open-telemetry/opentelemetry-proto/issues/308.

specification/metrics/sdk.md
```diff
@@ -386,22 +386,20 @@ This Aggregation honors the following configuration parameters:
...
 
-Note: This aggregator should not fill out `sum` when used with instruments
-that record negative measurements, e.g. `UpDownCounter` or `ObservableGauge`.
-
 This Aggregation informs the SDK to collect:
 
 - Count of `Measurement` values falling within explicit bucket boundaries.
-- Arithmetic sum of `Measurement` values in population.
+- Arithmetic sum of `Measurement` values in population. This SHOULD NOT be collected when used with
+instruments that record negative measurements, e.g. `UpDownCounter` or `ObservableGauge`.
```
jmacd, line 393:
```suggestion
instruments that record negative measurements, e.g. `UpDownCounter`.
```

jmacd (approved):

reyang (approved):
LGTM.

bogdandrutu (approved):

PR #2166, opened 2021-11-24
Add notes to MetricReader.Collect about overriding

### LINKED ISSUE #2139: Clarify if MetricReader would ever be implemented by the user

Implement the SDK spec in Python.

**Additional context.**

It is not clear to me from the spec if `MetricReader` is supposed to be implementable by the user or if they would only ever use the two provided implementations (for push and pull).

1. Would a user ever subclass `MetricReader`? What would they gain from this, or is it intended that they can only use MetricReader for pull and PeriodMetricReader for push?
2. Should a user be able to override `MetricReader.Collect()`, or should it be considered a final method? I was conceptualizing this method as the hook into the internals of the SDK, so it doesn't make sense for a user to override this to me.

My take is that there should only be two impls of `MetricReader`, one for pull based exporter to call `Collect()` on and one that passes metrics to a push exporter periodically. We only call out the latter in the spec.

reyang:
Fixes #2139.

### DIFF
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -618,9 +618,8 @@ filter, timeout). [OpenTelemetry SDK](../overview.md#sdk) authors MAY choose the
 return value type, or do not return anything.
 
 Note: it is expected that the `MetricReader.Collect` implementations will be
-provided by the SDK, so it is RECOMMENDED to prevent the user from
-accidentally overriding it, if possible (e.g. `final` in C++ and Java, `sealed`
-in C#).
+provided by the SDK, so it is RECOMMENDED to prevent the user from accidentally
+overriding it, if possible (e.g. `final` in C++ and Java, `sealed` in C#).
 
 ### Shutdown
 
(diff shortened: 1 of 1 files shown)


carlosalberto (approved):

specification/metrics/sdk.md
```diff
@@ -623,6 +623,11 @@ SDK](../overview.md#sdk) authors MAY choose to add parameters (e.g. callback,
 filter, timeout). [OpenTelemetry SDK](../overview.md#sdk) authors MAY choose the
 return value type, or do not return anything.
 
+Note: it is expected that the `MetricReader.Collect` implementations will be
+provided by the SDK, so the recommendation is to prevent the user from
```
MrAlias, line 627: Should this be a normative statement?

```suggestion
provided by the SDK, so it is RECOMMENDED to prevent the user from
```

cijothomas (approved):

jmacd (approved):

Issue #2137, opened 2021-11-18
OpenTracing layer support for Baggage


carlosalberto:
In OpenTracing, `SpanContext` and `Baggage` are part of the same object, whereas in OpenTelemetry they are different objects. This would be straightforward and trivial to support **if** the **OpenTracing Shim** couldn't be used along the OpenTelemetry API in the same code base -- which, however, has been deemed important to support gradual adoption in large codebases. 

This currently is supported in the **OpenTracing Shim** by associating any OpenTelemetry `Span` with a given `SpanContext` + `Baggage` combo at all times, in order to see changes reflected in different threads at all times. A typical use case is

```java
// Main thread, OTel API
io.opentelemetry.trace.Span span1 = ...; // Create OpenTelemetry Span alone
propagateAsCurrentSpanToThreads(span1);

// A thread consuming the the propagated span1 through the OT Shim
openTracingTracer.activeSpan().setBaggageItem("hello", "foo");

// Another thread consuming the propagated span1 through the OT Shim too
openTracingTracer.activeSpan().getBaggageItem("hello"); // returns foo
```

This is supported in [Python easily](https://github.com/open-telemetry/opentelemetry-python/compare/main...carlosalberto:ot_shim_baggage_multithread?expand=1) by decorating a pair of values in the OpenTelemetry `Span` itself, whereas in Java (and similar languages) it requires a global map with a global read-write lock, which _may_ still impact the performance greatly. 

### Alternative 1

Completely discard support for Baggage, providing the related functionality as no-op, given:

* Its complicated handling (for some languages) when the OT Shim is used along the OTel API.
* Its potential lack of common usage in OpenTracing
* Additionally: its heavy memory footprint (as `SpanContext` baggage values are **copied** from the the parent `Span`).

### Alternative 2

Provide a _best-hope_ design, in a relatively simplistic fashion, in order to handle common cases (such as an OpenTracing `Span` being created, have its baggage set, and have its baggage propagated).

* Have `SpanContext` + `Baggage` be part of the `Span` **Shim** object (hence, no need for global maps/locks), and propagate this object itself (along its current `Baggage`). This would allow:
  * OTel API usage: simply consume the propagated `Baggage`. 
  * OT Shim usage: have different threads **update** the `Baggage` value as needed.
* If the current `Context` has no `Span` **Shim** object, all the OT Shim Baggage operations are no-op.

This WOULD NOT satisfy the expected OT Handling, but it would offer limited support though.

[...]

PR #2167, opened 2021-11-24
Add execution-unit to glossary.md

### LINKED ISSUE #1511: Add execution-unit to the glossary.
In the context section we use "execution unit" to talk about threads/coroutines/fibers/etc. We should add it to the glossary (or alternatively use a better term).

svrnm:
Fixes #1511 ; since I was looking into adding the term service already, I thought I could also give this one a try. Hope it helps:)

## Changes

- Update glossary to contain a definition of the term execution unit
- Link back from opentracing.md and context.md where this term is used

### DIFF
--- a/specification/compatibility/opentracing.md
+++ b/specification/compatibility/opentracing.md
@@ -126,7 +126,7 @@ MUST be done by the Shim layer in order to retain these invariants.
 
 Because of the previous requirement, a given OpenTelemetry `Span`
 MUST be associated with ONE AND ONLY ONE `SpanContext` Shim object at all times
-for ALL execution units, in order to keep any linked `Baggage` consistent
+for ALL [execution units](../glossary.md#execution-unit), in order to keep any linked `Baggage` consistent
 at all times. It MUST BE safe to get and set the associated `SpanContext` Shim
 object for a specified OpenTelemetry `Span` from different execution units.
 
--- a/specification/context/context.md
+++ b/specification/context/context.md
@@ -21,7 +21,7 @@ Table of Contents
 ## Overview
 
 A `Context` is a propagation mechanism which carries execution-scoped values
-across API boundaries and between logically associated execution units.
+across API boundaries and between logically associated [execution units](../glossary.md#execution-unit).
 Cross-cutting concerns access their data in-process using the same shared
 `Context` object.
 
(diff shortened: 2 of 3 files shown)


specification/compatibility/opentracing.md
```diff
@@ -126,7 +126,7 @@ MUST be done by the Shim layer in order to retain these invariants.
 
 Because of the previous requirement, a given OpenTelemetry `Span`
 MUST be associated with ONE AND ONLY ONE `SpanContext` Shim object at all times
-for ALL execution units, in order to keep any linked `Baggage` consistent
+for ALL [execution units](../glossary.md#execution-unit), in order to keep any linked `Baggage` consistent
```
yurishkuro, line 129: @carlosalberto I think this section is overly restrictive, which may cause difficulties in implementation. For example:

> OpenTracing `SpanContext`, just as OpenTelemetry `SpanContext`, MUST be immutable

I don't think this was an explicitly stated requirement in OT, because SpanContext was an empty interface, thus there was no way to distinguish mutable from immutable. However, It was allowed that `span.context().baggageItems()` could return different results at different times. Given that, the next requirement:

> a given OpenTelemetry `Span` MUST be associated with ONE AND ONLY ONE `SpanContext` Shim object at all times

also seems somewhat problematic (it's actually hard to follow why this is written this way). In contrast, in Jaeger SDKs `span.context()` would return an immutable object, but could return a different object (with different baggage) on another invocation of `span.context()`. The context was immutable only from the POV of public interface of jaeger.SpanContext - internally the object could still be mutable e.g. for tracking deferred sampling.

So in summary, this seems to be dictating a lot of restrictions which were not present in OT. It's fine if you think this restrictions actually help with the shim implementation.

yurishkuro (approved):
I'm fine with the PR itself, but have concerns with the existing section that uses the term.

specification/compatibility/opentracing.md
carlosalberto: @yurishkuro We will actually overhaul completely this entire section based on the discussion in #2137 -- that is, we will NOT have this requirement, and we will do as Jaeger does (which is what OT says as well):

> span.context() would return an immutable object, but could return a different object (with different baggage) on another invocation of span.context()

The paragraph above was, yes, an intent to help with the Shim layer, but as we will be taking a simpler, better, correct route, we don't need the text above.

carlosalberto (approved):

specification/compatibility/opentracing.md
svrnm: Looks like this is not part of this PR and is covered by another issue, so I can resolve this conversation?

PR #2168, opened 2021-11-24
Limit view-rename capabilities in the presence of multi-select views

### LINKED ISSUE #2141: Clarify what happens if a View matches with multiple instruments

Implement the SDK spec in Python.

**Additional context.**

The spec isn't clear what happens if a view matches multiple instruments, for example if it selects all counters.

@jsuereth in chat:
> the view selector can apply to multiple instruments
> but you get one "instance" per instrument
> If you select multiple instruments and set a name, you've effectively got a bug
> but it’s not explicitly called out in the spec

In other words, because the view’s output stream name [defaults to that of the instrument](https://github.com/open-telemetry/opentelemetry-specification/blob/5418462162aeabc7b13b10545554226d4da1d1e8/specification/metrics/sdk.md#:~:text=If%20not%20provided%2C%20the%20Instrument%20name%20would%20be%20used%20by%20default.), there is no single-writer violation if you create one view “instance” per instrument. However if you specify the name, there is a conflict.

jsuereth:
Fixes #2141

- Remove the ability to force multiple-instrument metric streams to the same name.


specification/metrics/sdk.md
```diff
@@ -172,6 +172,15 @@ are the inputs:
     This should be a factory or callback similar to aggregation which allows
     different reservoirs to be chosen by the aggregation.
 
+The SDK MUST not allow the following configurations to be used together:
+
+- An instrument selection criteria lacking any parameters (matches everything)
```
cijothomas, line 177: "If _none_ the optional criteria is provided, the SDK SHOULD treat it as an
    error"

^ The above comment (https://github.com/open-telemetry/opentelemetry-specification/pull/2168/files#diff-32d5f289beed96900fa1e6e69a3ee9cc0b6503ce8883672011c44bc772c0fa2aR148-R151) already says if none of the criteria is provided, its SHOULD be an error.

specification/metrics/sdk.md
```diff
@@ -172,6 +172,15 @@ are the inputs:
...
+  combined with a specified `name` for the View. This leads to gauranteed
+  view conflicts between instruments.
+- An instrument selection criteria with wildcard name support and no specified
+  instrument type and a specified `name` for the View.  This leads to gauranteed
+  view conflicts between instruments.
```
cijothomas, line 182: The spec already make it clear (few lines below) how to deal with conflict. So this section feels like not necessary.

jsuereth: This actually prevents the entire view from being registered.  The section below allows *one* view to succeed for one instrument and the rest fail.

specification/metrics/sdk.md
jsuereth: Ah, good call out.  This one is intended to be more about matching against multiple-instrument types, let me clarify

specification/metrics/sdk.md
cijothomas: I see. Thanks for clarifying.

This is sometimes not feasible (in .NET SDK implm), as we evaluate Views dynamically (i.e when an instrument is created)  in certain cases.

This is definitely feasible in below case
```
.AddView(instrumentName: "counter*", name: "newname")
```

But not feasible in our advanced View API, where user registers a Function, which is evaluated whenever new instruments are created.

```
// Advanced selection criteria and config via Func<Instrument, MetricStreamConfiguration>
.AddView((instrument) =>
 {
     if (instrument.Name.StartsWith("counter*")
     {
         return new MetricStreamConfiguration() { name = "newname" };
     }
     return null;
 })
```

cijothomas: Basically I am asking to relax the MUST to SHOULD, as detecting if a View would result in a conflict is not feasible in all the scenarios.

specification/metrics/sdk.md
```diff
@@ -172,6 +172,15 @@ are the inputs:
     This should be a factory or callback similar to aggregation which allows
     different reservoirs to be chosen by the aggregation.
 
+The SDK MUST not allow the following configurations to be used together:
```
reyang, line 175:
```suggestion
The SDK MUST NOT allow the following configurations to be used together:
```

specification/metrics/sdk.md
```diff
@@ -172,6 +172,15 @@ are the inputs:
...
+
+- An instrument selection criteria lacking instrument type and instrument name
+  criteria combined with a specified `name` for the View. This leads to
+  gauranteed view conflicts between instruments.
+- An instrument selection criteria with wildcard name support and no specified
+  instrument type and a specified `name` for the View.  This leads to gauranteed
```
reyang, line 181:
```suggestion
  instrument type and a specified `name` for the View.  This leads to guaranteed
```

specification/metrics/sdk.md
```diff
@@ -196,6 +205,8 @@ made with an Instrument:
     default behavior off.
 * END.
 
+
```
reyang, line 208: Are these blank lines intentional?

jsuereth: no, accidental

specification/metrics/sdk.md
```diff
@@ -172,6 +172,15 @@ are the inputs:
...
 
+The SDK MUST not allow the following configurations to be used together:
+
+- An instrument selection criteria lacking instrument type and instrument name
+  criteria combined with a specified `name` for the View. This leads to
+  gauranteed view conflicts between instruments.
+- An instrument selection criteria with wildcard name support and no specified
+  instrument type and a specified `name` for the View.  This leads to gauranteed
+  view conflicts between instruments.
+
```
jack-berg, line 183: I'm not sure this criteria is right. 

Take the first statement: "An instrument selection criteria lacking instrument type and instrument name criteria combined with a specified `name` for the View. This leads to gauranteed view conflicts between instruments."

Reading that literally, I would think that unless BOTH instrument type and instrument name are set, I can't specify the view name. But really the instrument type is immaterial as long as long as the instrument name is specified. 

And the second statement: "An instrument selection criteria with wildcard name support and no specified instrument type and a specified `name` for the View.  This leads to gauranteed view conflicts between instruments."

The conflict is not guaranteed, since the regular expression may actually match exactly one instrument. 

We really want to ensure that if your view has a name, your instrument selector selects at most 1 instrument. I think a better phrasing would be something like:

"In order to avoid conflicts, views which specify a name should have an instrument selector which selects at most one instrument. Views with a name should not be used with instrument selectors which select by instrument type or wildcard."

specification/metrics/sdk.md
jsuereth: I think this needs to be MUST, we can loosen it later.

specification/metrics/sdk.md
jsuereth: Sorry for taking so long to get back no this, had to think some things through.


IIUC - `AddView(callback)` is *unspecified* in the specification, AFAICT.   The restriction written here I think is ok for these reasons:

1. The restriction applies to the *specified* API where users provide selection criteria + metric-names.  The API you're talking about is *NOT* that APi, so this restrictions wouldn't apply.
2. If/when we had a callback-based version of this API specified, then we could add a caveat to the language that this MUST only applies to the other API.

specification/metrics/sdk.md
jsuereth: Good point.  I'll fix the language.

jsuereth:
@cijothomas  / @jack-berg  -  I retweaked the wording here.


Cijo - I reworded this to make sure it's understood it applies to view registration where criteria is provided apriori (not the callback approach, which I think falls outside this specification).

Jack - Retweaked the wording you provided with the limitation we feel like we need in the spec right now.

jack-berg (approved):

specification/metrics/sdk.md
```diff
@@ -192,6 +192,12 @@ are the inputs:
     This should be a factory or callback similar to aggregation which allows
     different reservoirs to be chosen by the aggregation.
 
+In order to avoid conflicts, views which specify a name should have an
+instrument selector which selects at most one instrument. For the registration
```
aabmass, line 196:
```suggestion
In order to avoid conflicts, views which specify a name SHOULD have an
instrument selector which selects at most one instrument. For the registration
```

aabmass (approved):

specification/metrics/sdk.md
```diff
@@ -192,6 +192,12 @@ are the inputs:
     This should be a factory or callback similar to aggregation which allows
     different reservoirs to be chosen by the aggregation.
 
+In order to avoid conflicts, views which specify a name should have an
```
jmacd, line 195:
```suggestion
In order to avoid conflicts, views that specify a name should have an
```

specification/metrics/sdk.md
```diff
@@ -192,6 +192,12 @@ are the inputs:
...
+instrument selector which selects at most one instrument. For the registration
+mechanism described above, where selection is provided via configuration, the
+SDK MUST NOT allow Views with a specified name to be declared with instrument
+selectors which select by instrument type or wildcard.
```
jmacd, line 199:
```suggestion
selectors that select by instrument type or wildcard.
```

jmacd (approved):

specification/metrics/sdk.md
```diff
@@ -192,6 +192,12 @@ are the inputs:
     This should be a factory or callback similar to aggregation which allows
     different reservoirs to be chosen by the aggregation.
 
+In order to avoid conflicts, views which specify a name SHOULD have an
+instrument selector which selects at most one instrument. For the registration
```
jmacd, line 196:
```suggestion
instrument selector that selects at most one instrument. For the registration
```

jmacd (approved):

MrAlias (approved):

specification/metrics/sdk.md
jmacd: Thanks for clarifying @jsuereth. I agree this "advanced View API" is not the thing being specified.

cijothomas (approved):

PR #2150, opened 2021-11-23
Mark Metrics SDK as Mixed/Stable

reyang:
This moves the Metrics SDK specification to Mixed/Stable:
1. Once the spec is marked as "Stable", **breaking changes are no longer allowed**.
2. Additive changes are allowed.

These are the issues that should be resolved before we could merge this PR, unless the issue is specific about things that are not marked as Stable in this PR (e.g. Exemplar):
https://github.com/open-telemetry/opentelemetry-specification/issues?q=is%3Aissue+is%3Aopen+label%3Aspec%3Ametrics+label%3Aarea%3Asdk+label%3Arelease%3Arequired-for-ga+-label%3Aarea%3Asemantic-conventions+

Project board https://github.com/orgs/open-telemetry/projects/9.

image
<sub>([original link](https://www.monkeyuser.com/2018/release-day/))</sub>

PR #2175, opened 2021-11-29
Mark In-Memory/OTLP/Stdout exporters as Stable

reyang:
This has a dependency on the MetricExporter spec, so it should wait for #2150 to be merged.

### DIFF
--- a/specification/metrics/sdk_exporters/in-memory.md
+++ b/specification/metrics/sdk_exporters/in-memory.md
@@ -1,6 +1,6 @@
 # OpenTelemetry Metrics Exporter - In-memory
 
-**Status**: [Feature-freeze](../../document-status.md)
+**Status**: [Stable](../../document-status.md)
 
 In-memory Metrics Exporter is a [Push Metric
 Exporter](../sdk.md#push-metric-exporter) which accumulates metrics data in the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ release.
   ([#2131](https://github.com/open-telemetry/opentelemetry-specification/pull/2131))
 - Remove the concept of supported temporality, keep preferred.
   ([#2154](https://github.com/open-telemetry/opentelemetry-specification/pull/2154))
+- Mark In-memory, OTLP and Stdout exporter specs as Stable.
+  ([#TBD](https://github.com/open-telemetry/opentelemetry-specification/pull/TBD))
 
 ### Logs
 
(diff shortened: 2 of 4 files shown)


jsuereth (approved):

cijothomas (approved):

jmacd (approved):

mtwo (approved):

specification/metrics/sdk_exporters/in-memory.md
```diff
@@ -1,6 +1,6 @@
 # OpenTelemetry Metrics Exporter - In-memory
 
-**Status**: [Feature-freeze](../../document-status.md)
+**Status**: [Stable](../../document-status.md)
```
jack-berg, line 3: What should the behavior of flush be for in-memory metric exporter? In java, [as noted in this ticket](https://github.com/open-telemetry/opentelemetry-java/issues/3943#issue-1067754880), it triggers a collection, which seems incorrect according to the [force flush description](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#forceflush-1).

reyang: I can share what .NET did:
1. InMemoryExporter.ForceFlush simply returns true.
2. If the user is reporting measurements and want to make sure the resulting metrics are collected, they should use either of the following way:

    2.1 Call `MeterProvider.ForceFlush`, which in turn will call `MetricReader.Collect` on all MetricReader instances, which will call `InMemoryExporter.ForceFlush`, the callstack looks like:
    ```text
    InMemoryExporter.ForceFlush
    BaseExportingMetricReader.Collect (or PeriodicExportingMetricReader, depending on the configuration)
    MeterProvider.ForceFlush
    ````
    2.2. Call `MetricReader.Collect` on the MetricReader instance that is associated with the InMemoryExporter.

The difference between 2.1 and 2.2: if there are multiple readers/exporters, 2.1 will trigger all of them (except for exporters that only support pull - e.g. Prometheus), 2.2 will only trigger the collect/flush for the specific reader/exporter.

jack-berg: Right, that's the behavior I think is correct as well. I think its surprising for the java `InMemoryMetricExporter#flush()` to call collect, and will work to get that sorted out.

jack-berg (approved):

cijothomas:
> This has a dependency on the MetricExporter spec, so it should wait for #2150 to be merged.

^^ #2150 is not merged.

PR #2180, opened 2021-11-30
Define semantic conventions and instrumentation stability

tigrannajaryan:
This PR defines the rules for SDKs in all languages with regards to marking
instrumentations as "Stable" and "Unstable" and how such instrumentations
are allowed to evolve over time.

Summary of changes:

- Explain that schemas define what changes are allowed.
- Clarify that additive changes are an exception and are always allowed.
- Prohibit instrumentation changes until April 1, 2023 to allow recipients
  (backends/vendors) to start understanding Schema URLs/Schema Files.
  This gives one year advance notice, which I think should be enough for
  all parties to prepare.

### DIFF
--- a/specification/telemetry-stability.md
+++ b/specification/telemetry-stability.md
@@ -1,0 +1,98 @@
+# Telemetry Stability
+
++**Status**: [Experimental](document-status.md)
+
+<details>
+<summary>Table of Contents</summary>
+
+<!-- toc -->
+
+- [Unstable Instrumentations](#unstable-instrumentations)
+- [Stable Instrumentations](#stable-instrumentations)
+  * [Fixed Schema Telemetry Producers](#fixed-schema-telemetry-producers)
+  * [Schema-File Driven Telemetry Producers](#schema-file-driven-telemetry-producers)
+
+<!-- tocstop -->
+
+</details>
+
+This section defines stability requirements for telemetry produced by
+OpenTelemetry instrumentations.
+
+All OpenTelemetry-authored instrumentations are labeled to be either `Unstable` or `Stable`
+from the perspective of the telemetry they produce.
+
+Adding of new metrics, spans, span events or log records and adding of
+new attributes to spans, span events, log records or resources are considered
+additive, non-breaking changes and are always allowed for `Unstable` and `Stable`
+instrumentations.
+
+Other changes in the produced telemetry are regulated by the following rules.
+
+## Unstable Instrumentations
+
+Unstable telemetry-producing instrumentations (unstable instrumentations for short) SHOULD
+be clearly labeled so by any means the instrumentations authors consider idiomatic for
+their language, e.g. via version numbers, artifact names, documentation, etc.
+
+Unstable instrumentations provide no guarantees about the shape of
+the telemetry they produce and how that shape changes over time, from version to version.
+Span or metric names, attributes of any telemetry items may change without any
+restrictions. The produced telemetry MAY specify a Schema URL if the telemetry data
+conforms to a particular Schema.
+
+Unstable instrumentations authored by OpenTelemetry MAY produce telemetry that is
+not described by OpenTelemetry semantic conventions.
+
+TODO: decide if it is necessary to indicate on the wire if the produced telemetry is
+coming from an unstable instrumentation.
+
+## Stable Instrumentations
+
+Stable telemetry-producing instrumentations (stable instrumentations for short) SHOULD
+be clearly labeled so by any means the instrumentations authors consider idiomatic for
+their language, e.g. via version numbers, artifact names, documentation, etc.
+
+Stable instrumentations fall into 2 categories: fixed-schema producers and schema-file
+driven producers.
+
+Stable instrumentations authored by OpenTelemetry SHOULD NOT produce telemetry that is
+not described by OpenTelemetry semantic conventions. If, however, this rule is broken the
+instrumentations MUST NOT change such telemetry, regardless of whether they
+are fixed-schema producers or schema-file driven producers. Once the produced telemetry
+is added to the semantic conventions, changes will be allowed as described below.
+
+### Fixed Schema Telemetry Producers
+
+Instrumentations that are labeled `Stable` and do not include the Schema URL in the
+produced telemetry are called Fixed Schema Telemetry Producers.
+
+Such instrumentations are prohibited from changing any produced telemetry. If the
+specification changes over time and the semantic conventions are updated, the
+instrumentation is still prohibited from adopting the changes. If the instrumentation
+wishes to adopt the semantic convention changes it must first become a
+[Schema-File Driven Telemetry Producer](#schema-file-driven-telemetry-producers) by
+adding an appropriate Schema URL in the produced telemetry.
+
+### Schema-File Driven Telemetry Producers
+
+Stable instrumentations that include the Schema URL in the produced telemetry are
+called Schema-File Driven Telemetry Producers.
+
+Such instrumentations are prohibited from changing the produced telemetry until
+April 1, 2023 and until that date are subject to exactly the same restrictions as
+[Fixed Schema Telemetry Producers](#fixed-schema-telemetry-producers).
+
+After April 1, 2023, stable instrumentations are allowed to change the produced telemetry
+if all the following conditions are fulfilled:
+
+- The change is part of OpenTelemetry semantic conventions and is in a released
+  version of the specification,
+- The change has a corresponding
+  [published](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/overview.md#opentelemetry-schema)
+  OpenTelemetry Schema File that describes the change.
+- The produced telemetry correctly specifies the Schema URL.
+
+If the change was introduced in the semantic conventions specification before
+April 1, 2023, the instrumentations must wait until April 1, 2023 before they can adopt
+the change and begin producing the changed telemetry.
--- a/specification/telemetry-stability.md
+++ b/specification/telemetry-stability.md
@@ -88,8 +88,7 @@ if all the following conditions are fulfilled:
 
 - The change is part of OpenTelemetry semantic conventions and is in a released
   version of the specification.
-- The change has a corresponding
-  [published](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/overview.md#opentelemetry-schema)
+- The change has a corresponding [published](schemas/overview.md#opentelemetry-schema)
   OpenTelemetry Schema File that describes the change.
 - The produced telemetry correctly specifies the respective Schema URL.
 
--- a/specification/telemetry-stability.md
+++ b/specification/telemetry-stability.md
@@ -41,8 +41,8 @@ Span or metric names, attributes of any telemetry items may change without any
 restrictions. The produced telemetry MAY specify a Schema URL if the telemetry data
 conforms to a particular Schema.
 
-Unstable instrumentations authored by OpenTelemetry MAY produce telemetry that is
-not described by OpenTelemetry semantic conventions.
+Unstable instrumentations authored by OpenTelemetry MAY produce additional telemetry that
+is not described by OpenTelemetry semantic conventions.
 
 TODO: decide if it is necessary to indicate on the wire if the produced telemetry is
 coming from an unstable instrumentation.
@@ -87,11 +87,11 @@ After April 1, 2023, stable instrumentations are allowed to change the produced 
 if all the following conditions are fulfilled:
 
 - The change is part of OpenTelemetry semantic conventions and is in a released
-  version of the specification,
+  version of the specification.
 - The change has a corresponding
   [published](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/overview.md#opentelemetry-schema)
   OpenTelemetry Schema File that describes the change.
-- The produced telemetry correctly specifies the Schema URL.
+- The produced telemetry correctly specifies the respective Schema URL.
 
 If the change was introduced in the semantic conventions specification before
 April 1, 2023, the instrumentations must wait until April 1, 2023 before they can adopt
--- a/specification/versioning-and-stability.md
+++ b/specification/versioning-and-stability.md
@@ -15,8 +15,8 @@
       - [Extending Existing API Calls](#extending-existing-api-calls)
     + [SDK Stability](#sdk-stability)
     + [Contrib Stability](#contrib-stability)
-    + [NOT DEFINED: Telemetry Stability](#not-defined-telemetry-stability)
-    + [NOT DEFINED: Semantic Conventions Stability](#not-defined-semantic-conventions-stability)
+    + [Semantic Conventions Stability](#semantic-conventions-stability)
+    + [Telemetry Stability](#telemetry-stability)
   * [Deprecated](#deprecated)
   * [Removed](#removed)
   * [A note on replacing signals](#a-note-on-replacing-signals)
@@ -134,11 +134,14 @@ Languages which ship binary artifacts SHOULD offer [ABI compatibility](glossary.
 
 #### Contrib Stability
 
-**NOTE: Until telemetry stability is defined, Contrib instrumentation MUST NOT be marked as stable. See below.**
-
-Plugins, instrumentation, and other contrib packages SHOULD be kept up to date and compatible with the latest versions of the API, SDK, and Semantic Conventions.
-If a release of the API, SDK, or Semantic Conventions contains changes which are relevant to a contrib package, that package SHOULD be updated and released in a timely fashion.
-The goal is to ensure users can update to the latest version of OpenTelemetry, and not be held back by the plugins that they depend on.
+Plugins, instrumentation, and other contrib packages SHOULD be kept up to date
+and compatible with the latest versions of the API, SDK, and Semantic
+Conventions. If a release of the API, SDK, or Semantic Conventions contains
+changes which are relevant to a contrib package, that package SHOULD be updated
+and released in a timely fashion. (See limitations on instrumentation stability
+in [Telemetry Stability](telemetry-stability.md)). The goal is to ensure users can
+update to the latest version of OpenTelemetry, and not be held back by the
+plugins that they depend on.
 
 Public portions of contrib packages (constructors, configuration, interfaces) SHOULD remain backwards compatible.
 
@@ -149,20 +152,50 @@ For example, a database integration may break stability if the required database
 However, it is strongly RECOMMENDED that older contrib packages remain stable.
 A new, incompatible version of an integration SHOULD be released as a separate contrib package, rather than break the existing contrib package.
 
-#### NOT DEFINED: Telemetry Stability
+#### Semantic Conventions Stability
 
-**Telemetry stability guarantees are TBD.**
+Changes to telemetry produced by OpenTelemetry instrumentation SHOULD avoid
+breaking analysis tools, such as dashboards and alerts. To achieve this, while
+allowing the evolution of telemetry and semantic conventions, OpenTelemetry
+relies on the concept of
+[schemas](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/overview.md).
 
-Changes to telemetry produced by OpenTelemetry instrumentation SHOULD avoid breaking analysis tools, such as dashboards and alerts.
-However, it is not clear at this time what type of instrumentation changes (for example, adding additional spans and attributes) would actually cause a breaking change.
+Changes to semantic conventions in this specification are allowed, provided that
+the changes can be described by schema files. The following changes can be
+currently described and are allowed:
 
-#### NOT DEFINED: Semantic Conventions Stability
+- Renaming of attributes.
+- Renaming of metrics.
+- Renaming of span events.
 
-Telemetry stability, including semantic conventions, is not currently defined. The following practices are recommended.
+All such changes MUST be described in the OpenTelemetry
+[Schema File Format](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/file_format_v1.0.0.md)
+and published
+[here](https://github.com/open-telemetry/opentelemetry-specification/tree/main/schemas).
+For details see [how OpenTelemetry Schemas are
+published](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/overview.md#opentelemetry-schema).
 
-Semantic Conventions SHOULD NOT be removed once they are added.
-New conventions MAY be added to replace usage of older conventions, but the older conventions SHOULD NOT be removed.
-Older conventions SHOULD be marked as deprecated when they are replaced by newer conventions.
+See [Telemetry Stability](telemetry-stability.md) document for details on how
+instrumentations can use schemas to change the instrumentation they produce.
+
+In addition to the 3 types of changes described above there are certain types
+that are always allowed. Such changes do not need to be described (and are not
+described) by schema files. Here is the list of such changes:
+
+- Adding new attributes to the existing semantic conventions for resources,
+  spans, span events or log records.
+- Adding semantic conventions for new types of resources, spans, span events,
+  metrics or log records.
+
+Any other changes to semantic conventions are currently prohibited. Other types
+of changes MAY be introduced in the future versions of this specification. This
+is only allowed if OpenTelemetry introduces a new schema file format that is
+capable of describing such changes.
+
+#### Telemetry Stability
+
+For stability of telemetry produced by instrumentation see
+[Telemetry Stability](telemetry-stability.md) document.
 
 ### Deprecated
 
--- a/specification/versioning-and-stability.md
+++ b/specification/versioning-and-stability.md
@@ -139,7 +139,7 @@ and compatible with the latest versions of the API, SDK, and Semantic
 Conventions. If a release of the API, SDK, or Semantic Conventions contains
 changes which are relevant to a contrib package, that package SHOULD be updated
 and released in a timely fashion. (See limitations on instrumentation stability
-in [Telemetry Stability](telemetry-stability.md)). The goal is to ensure users can
+in [Telemetry Stability](telemetry-stability.md).) The goal is to ensure users can
 update to the latest version of OpenTelemetry, and not be held back by the
 plugins that they depend on.
 
@@ -158,13 +158,13 @@ Changes to telemetry produced by OpenTelemetry instrumentation SHOULD avoid
 breaking analysis tools, such as dashboards and alerts. To achieve this, while
 allowing the evolution of telemetry and semantic conventions, OpenTelemetry
 relies on the concept of
-[schemas](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/overview.md).
+[Telemetry Schemas](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/overview.md).
 
 Changes to semantic conventions in this specification are allowed, provided that
 the changes can be described by schema files. The following changes can be
 currently described and are allowed:
 
-- Renaming of attributes.
+- Renaming of span, metric, log and resource attributes.
 - Renaming of metrics.
 - Renaming of span events.
 
@@ -175,7 +175,7 @@ and published
 For details see [how OpenTelemetry Schemas are
 published](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/overview.md#opentelemetry-schema).
 
-See [Telemetry Stability](telemetry-stability.md) document for details on how
+See the [Telemetry Stability](telemetry-stability.md) document for details on how
 instrumentations can use schemas to change the instrumentation they produce.
 
 In addition to the 3 types of changes described above there are certain types
@@ -183,7 +183,8 @@ that are always allowed. Such changes do not need to be described (and are not
 described) by schema files. Here is the list of such changes:
 
 - Adding new attributes to the existing semantic conventions for resources,
-  spans, span events or log records.
+  spans, span events or log records. Note that adding attributes to existing metrics is
+  considered to be a breaking change.
 - Adding semantic conventions for new types of resources, spans, span events,
   metrics or log records.
 
@@ -194,7 +195,7 @@ capable of describing such changes.
 
 #### Telemetry Stability
 
-For stability of telemetry produced by instrumentation see
+For stability of telemetry produced by instrumentation see the
 [Telemetry Stability](telemetry-stability.md) document.
 
 ### Deprecated
--- a/specification/versioning-and-stability.md
+++ b/specification/versioning-and-stability.md
@@ -158,7 +158,7 @@ Changes to telemetry produced by OpenTelemetry instrumentation SHOULD avoid
 breaking analysis tools, such as dashboards and alerts. To achieve this, while
 allowing the evolution of telemetry and semantic conventions, OpenTelemetry
 relies on the concept of
-[Telemetry Schemas](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/overview.md).
+[Telemetry Schemas](schemas/overview.md).
 
 Changes to semantic conventions in this specification are allowed, provided that
 the changes can be described by schema files. The following changes can be
@@ -169,11 +169,8 @@ currently described and are allowed:
 - Renaming of span events.
 
 All such changes MUST be described in the OpenTelemetry
-[Schema File Format](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/file_format_v1.0.0.md)
-and published
-[here](https://github.com/open-telemetry/opentelemetry-specification/tree/main/schemas).
-For details see [how OpenTelemetry Schemas are
-published](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/overview.md#opentelemetry-schema).
+[Schema File Format](schemas/file_format_v1.0.0.md) and published [here](../schemas).
+For details see [how OpenTelemetry Schemas are published](schemas/overview.md#opentelemetry-schema).
 
 See the [Telemetry Stability](telemetry-stability.md) document for details on how
 instrumentations can use schemas to change the instrumentation they produce.
--- a/specification/versioning-and-stability.md
+++ b/specification/versioning-and-stability.md
@@ -169,7 +169,7 @@ currently described and are allowed:
 - Renaming of span events.
 
 All such changes MUST be described in the OpenTelemetry
-[Schema File Format](schemas/file_format_v1.0.0.md) and published [here](../schemas).
+[Schema File Format](schemas/file_format_v1.0.0.md) and published in this repository.
 For details see [how OpenTelemetry Schemas are published](schemas/overview.md#opentelemetry-schema).
 
 See the [Telemetry Stability](telemetry-stability.md) document for details on how
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,9 @@ release.
 
 ### Common
 
+- Define semantic conventions and instrumentation stability.
+  ([#2180](https://github.com/open-telemetry/opentelemetry-specification/pull/2180))
+
 ## v1.10.0 (2022-04-01)
 
 ### Context
@@ -156,7 +159,7 @@ release.
 - Add "Mapping Arbitrary Data to OTLP AnyValue" document.
   ([#2385](https://github.com/open-telemetry/opentelemetry-specification/pull/2385))
 
-## v1.9.0 (2022-02-10)
+## v1.9.0 (2021-02-10)
 
 ### Context
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -159,7 +159,7 @@ release.
 - Add "Mapping Arbitrary Data to OTLP AnyValue" document.
   ([#2385](https://github.com/open-telemetry/opentelemetry-specification/pull/2385))
 
-## v1.9.0 (2021-02-10)
+## v1.9.0 (2022-02-10)
 
 ### Context
 


specification/versioning-and-stability.md
```diff
@@ -149,20 +149,70 @@ For example, a database integration may break stability if the required database
...
+allowing the evolution of telemetry and semantic conventions, OpenTelemetry
+relies on the concept of
+[schemas](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/overview.md).
+
+Changes to semantic conventions in this specification are allowed, provided that
+the changes can be described by schema files. The following changes can be
+currently described and are allowed:
+
+- Renaming of attributes.
+- Renaming of metrics
```
pyohannes, line 165:
```suggestion
- Renaming of metrics.
```

specification/versioning-and-stability.md
```diff
@@ -149,20 +149,70 @@ For example, a database integration may break stability if the required database
...
+
+In addition to the 3 types of changes described above, adding conventions for
+new attributes, resources, metrics, spans, span events or log records is always
+allowed. Such changes do not need to be described (and are not described) by
+schema files.
+
+Any other changes to semantic conventions are currently prohibited. Other types
+of changes MAY be introduced in the future versions of this specification. This
+is only allowed if OpenTelemetry introduces a new schema file format that is
+capable of describing such changes.
```
pyohannes, line 183: Can such changes be made if the major version number is incremented?

specification/versioning-and-stability.md
```diff
@@ -149,20 +149,70 @@ For example, a database integration may break stability if the required database
...
+
+#### Telemetry Stability
+
+This section defines stability requirements for telemetry produced by
+instrumentations, such as OpenTelemetry SDKs.
```
pyohannes, line 188: I'm a bit confused by calling OpenTelemetry SDKs "instrumentations". Is "instrumentations" synonymous with [instrumentation libraries](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/glossary.md#instrumentation-library)?

specification/versioning-and-stability.md
```diff
@@ -149,20 +149,70 @@ For example, a database integration may break stability if the required database
...
+OpenTelemetry semantic conventions.
+
+Note that adding of new attributes, metrics, spans, span events or log records
+are considered additive, non-breaking changes and are always allowed, whether
+the instrumentation includes the Schema URL in the produced telemetry or no.
+
+OpenTelemetry SDKs SHOULD NOT produce telemetry that is not described by
+OpenTelemetry semantic conventions. If, however, this rule is broken the SDKs
+MUST NOT change such telemetry, regardless of whether they include the Schema
+URL in the produced telemetry or no.
```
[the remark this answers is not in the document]
tigrannajaryan: >Is this saying that if an OpenTelemetry SDK is incorrectly producing telemetry not described in semantic conventions, it must continue to do so indefinitely?

Not indefinitely. The produced telemetry can be added to the semantic conventions and after that changes to that telemetry can be described by schemas.

This only says that telemetry that for some reason did not follow the rules and is not described in the semantic conventions must be kept unchanged to avoid breaking recipients. Once it begins to follow the rules and is part of conventions it is back on track to be handled the normal way via schemas.

specification/versioning-and-stability.md
tigrannajaryan: I think the answer in theory is yes, but we already have this in the doc:
>Note that we currently have no plans for creating a major version of OpenTelemetry past v1.0.

If this plan changes we should also rethink what happens with semantic conventions.

specification/versioning-and-stability.md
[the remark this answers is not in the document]
tigrannajaryan: Added a sentence to clarify this. Let me know if it helps.

specification/versioning-and-stability.md
tigrannajaryan: Fixed.

specification/versioning-and-stability.md
tigrannajaryan: Reworded. See if it is clearer now.

jsuereth (dismissed):
ADDITIVE changes to metrics can break user alerts by creating new timeseries and adjusting thresholds.  I think we need to be a lot more careful around changes, particularly around metrics here.

tigrannajaryan:
> ADDITIVE changes to metrics can break user alerts by creating new timeseries and adjusting thresholds. I think we need to be a lot more careful around changes, particularly around metrics here.

@jsuereth I can change the PR to explicitly prohibit this but I think this contradicts with an existing understanding that attributes are optional, see this discussion https://github.com/open-telemetry/opentelemetry-specification/issues/2077

tigrannajaryan:
> ADDITIVE changes to metrics can break user alerts by creating new timeseries and adjusting thresholds. I think we need to be a lot more careful around changes, particularly around metrics here.

@jsuereth I modified the PR to disallow addition of new attributes to existing metrics. PTAL.

specification/versioning-and-stability.md
```diff
@@ -149,20 +149,76 @@ For example, a database integration may break stability if the required database
...
+and published
+[here](https://github.com/open-telemetry/opentelemetry-specification/tree/main/schemas).
+For details see [how OpenTelemetry Schemas are
+published](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/overview.md#opentelemetry-schema).
+
+In addition to the 3 types of changes described above there are certain types
+that are always allowed. Such changes do not need to be described (and are not
+described) by schema files. Here is the list of such changes:
+
+- Adding new attributes to the existing semantic conventions for resources,
```
jsuereth, line 179: For the purpose of expediency we should disallow adding new attributes to resources UNTIL we've figured out meta-issues around how that correlates with metrics, e.g. [this one](https://github.com/open-telemetry/opentelemetry-specification/issues/1782) on mapping resource into prometheus.

specification/versioning-and-stability.md
jsuereth: +1 to @tigrannajaryan's response here.   (1) Yes if we break anything it's a major version number but (2) stability-is-a-feature.   We should strive to avoid any significant breaking of our API (where API => usages of telemetry) here

specification/versioning-and-stability.md
```diff
@@ -149,20 +149,76 @@ For example, a database integration may break stability if the required database
...
+
+Instrumentations, which do not include the Schema URL in the produced telemetry
+are prohibited from changing any produced telemetry that is described by
+OpenTelemetry semantic conventions.
+
+Note that adding of new attributes, metrics, spans, span events or log records
+are considered additive, non-breaking changes and are always allowed, whether
+the instrumentation includes the Schema URL in the produced telemetry or no.
+
+OpenTelemetry instrumentations SHOULD NOT produce telemetry that is not
```
jsuereth, line 216: How does this statement align with our goal of prototyping instrumentation *prior* to specification.


Additionally, how does it align with the specification being a "minimal" standard with expansion possible on case-by-case basis (e.g. MySQL instrumentation vs. "DB" instrumentation).


I have concerns that instrumentaiton-itself is *expensive*, for us and for users.   This kind of lockdown makes sense from one point of view, but I think is highly restrictive on instrumentation authors, and puts huge bars on OTel-delivered instrumentation.

Can we somehow rephrase this to allow prototyping/experimentation within the otel project (in additon to non-otel projects) around instrumentation?

jsuereth (commented):
thanks for resolving my blocking comment!   I still have some concerns, but not as strongly blocking :)

For clarity - I agree with the sentiment and goal of this PR and approve, just nits on details/scope of what's stated here, including implications on instrumentation development.

specification/versioning-and-stability.md
pyohannes: It's somewhat clearer. I assume "instrumentation libraries that are part of OpenTelemetry" comprise all instrumentation libraries in `opentelemetry-*` repositories, like [opentelemetry-dotnet](https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src), [opentelemetry-java-instrumentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation), [opentelemetry-dotnet-contrib](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src), or [opentelemetry-cpp-contrib](https://github.com/open-telemetry/opentelemetry-cpp-contrib/tree/main/instrumentation).

specification/versioning-and-stability.md
```diff
@@ -149,20 +149,76 @@ For example, a database integration may break stability if the required database
...
+  spans, span events or log records.
+- Adding semantic conventions for new types of resources, spans, span events,
+  metrics or log records.
+
+Any other changes to semantic conventions are currently prohibited. Other types
+of changes MAY be introduced in the future versions of this specification. This
+is only allowed if OpenTelemetry introduces a new schema file format that is
+capable of describing such changes.
+
+#### Telemetry Stability
```
pyohannes, line 189: I guess this note can be removed in this PR:

> NOTE: Until telemetry stability is defined, Contrib instrumentation MUST NOT be marked as stable. See below.

specification/versioning-and-stability.md
tigrannajaryan: This is a good point. It is unfortunate that we have to disallow this. Let me think and see if there is a workaround possible.

Oberon00: Couldn't we just add new resource attributes and metric labels (or just everything new) to the schema?

specification/versioning-and-stability.md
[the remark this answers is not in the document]
kenfinnigan: Thanks!

specification/versioning-and-stability.md
tigrannajaryan: Yes, that's correct.

specification/versioning-and-stability.md
tigrannajaryan: @Oberon00 we can if we introduce such an operation in schema files. Currently it is not possible. To introduce it we need to add a new revision of schema files. It is doable if we think it is necessary. So far I refrained adding capabilities to schema files for which I did not see a clear demand.

specification/versioning-and-stability.md
tigrannajaryan: Good point. Added an exception `unless the instrumentation is clearly marked as unstable`. Does this help? Prototypes obvious are unstable so this should cover it.

specification/versioning-and-stability.md
tigrannajaryan: Good catch. Removed and added a reference from here to the Telemetry Stability section.

specification/versioning-and-stability.md
jsuereth: I think it's ok for now, we can debate if we need to lossen this. In practice, it's only as good as folks are willing to experiment :)

specification/versioning-and-stability.md
```diff
@@ -149,20 +152,76 @@ For example, a database integration may break stability if the required database
...
+
+This section defines stability requirements for telemetry produced by
+OpenTelemetry instrumentations.
+
+Instrumentations, which include the Schema URL in the produced telemetry are
+allowed to change the produced telemetry after July 1, 2022 if both of the
```
tigrannajaryan, line 198: Is this a sufficient notice? It may not be enough since implement support for schemas may take time.

trask:
Can you clarify how this applies to auto-instrumentation distributions?

Should we hide all telemetry producing auto-instrumentation in a distribution behind an experimental flag, and only enable them by default once their respective semantic conventions have reached stability and we can commit to telemetry stability?

tigrannajaryan:
> Can you clarify how this applies to auto-instrumentation distributions?
> 
> Should we hide all telemetry producing auto-instrumentation in a distribution behind an experimental flag, and only enable them by default once their respective semantic conventions have reached stability and we can commit to telemetry stability?

No, I believe it is fine to produce telemetry from auto-instrumentation. This PR only says that after you declare your auto-instrumentation "Stable" you cannot change the produced telemetry until certain conditions are met (SchemaURL in the output and wait until certain date). If the text in the PR is unclear I will fix it. Let me know which part you find unclear.

specification/versioning-and-stability.md
```diff
@@ -149,20 +152,76 @@ For example, a database integration may break stability if the required database
...
+instrumentations must wait until July 1, 2022 before they can adopt the change
+and begin producing the changed telemetry.
+
+Instrumentations, which do not include the Schema URL in the produced telemetry
+are prohibited from changing any produced telemetry that is described by
+OpenTelemetry semantic conventions.
+
+Note that adding of new attributes, metrics, spans, span events or log records
+are considered additive, non-breaking changes and are always allowed, whether
+the instrumentation includes the Schema URL in the produced telemetry or no.
```
jack-berg, line 217:
```suggestion
the instrumentation includes the Schema URL in the produced telemetry or not.
```

specification/versioning-and-stability.md
```diff
@@ -149,20 +152,76 @@ For example, a database integration may break stability if the required database
...
+
+OpenTelemetry instrumentations SHOULD NOT produce telemetry that is not
+described by OpenTelemetry semantic conventions, unless the instrumentation is
+clearly marked as unstable. If, however, this rule is broken the
```
jack-berg, line 221: How does instrumentation indicate that its unstable? Is there some sort of marker in the exported data? Could we include a special schema url that indicates instability?

specification/versioning-and-stability.md
```diff
@@ -149,20 +152,76 @@ For example, a database integration may break stability if the required database
...
+[here](https://github.com/open-telemetry/opentelemetry-specification/tree/main/schemas).
+For details see [how OpenTelemetry Schemas are
+published](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/overview.md#opentelemetry-schema).
+
+In addition to the 3 types of changes described above there are certain types
+that are always allowed. Such changes do not need to be described (and are not
+described) by schema files. Here is the list of such changes:
+
+- Adding new attributes to the existing semantic conventions for resources,
+  spans, span events or log records.
```
jack-berg, line 183: Are metrics intentionally omitted from this list? Metrics are included in a similar list on line [215](https://github.com/open-telemetry/opentelemetry-specification/pull/2180/files#diff-e7de8bed9eb1605a44c385930415011d900acccecc8b48398a75ce9fa707cdd3R215-R216) which is confusing.

tigrannajaryan: Yes, metrics are intentionally omitted. Adding a new attribute to a metric is considered a breaking change. See https://github.com/open-telemetry/opentelemetry-specification/pull/2180#pullrequestreview-820513856
I will fix line 215.

specification/versioning-and-stability.md
tigrannajaryan: Is there a need to indicate this to the receivers? Potentially lack of SchemaURL in the telemetry or version number <1.0 can be these indicators.

trask:
> > Can you clarify how this applies to auto-instrumentation distributions?
> > Should we hide all telemetry producing auto-instrumentation in a distribution behind an experimental flag, and only enable them by default once their respective semantic conventions have reached stability and we can commit to telemetry stability?
> 
> No, I believe it is fine to produce telemetry from auto-instrumentation. This PR only says that after you declare your auto-instrumentation "Stable" you cannot change the produced telemetry until certain conditions are met (SchemaURL in the output and wait until certain date). If the text in the PR is unclear I will fix it. Let me know which part you find unclear.

For individual instrumentations, we use the version number to "clearly mark" them as unstable, e.g. `0.x` or `1.x-alpha`.

For distributions (e.g. the opentelemetry javaagent), should we be marking them as "unstable" (e.g. `0.x` or `1.x-alpha`) if they emit any unstable instrumentation?

specification/versioning-and-stability.md
tigrannajaryan: Fixed.

specification/versioning-and-stability.md
tigrannajaryan: Should be fixed now, PTAL.

tigrannajaryan:
> For individual instrumentations, we use the version number to "clearly mark" them as unstable, e.g. `0.x` or `1.x-alpha`.
> 
> For distributions (e.g. the opentelemetry javaagent), should we be marking them as "unstable" (e.g. `0.x` or `1.x-alpha`) if they emit any unstable instrumentation?

The "unstable" vs "stable" labeling of a library or of a package (or individual instrumentations if that's the granularity of labeling you choose to maintain) is done based on the fact whether you are able to take the responsibility to fulfill the requirements of the "stable" labeling. What "stable" label means is described in this document.

Particularly with regards to telemetry emitted by the instrumentation this document (after accepting this PR) says that if you label the instrumentation "stable" then you must fulfil the following obligation:
- If you are not including Schema URL in the emitted telemetry then you are not allowed to modify the telemetry (with certain exceptions carved out in the doc).
- If you are including Schema URL then you are allowed to modify the telemetry after certain date, provided that they are described by a published Schema File (with the same exceptions as above applicable).

If you are not labeling the instrumentation as "stable" then you free to do whatever you want. Any changes are fine.

trask:
> The "unstable" vs "stable" labeling of a library or of a package (or individual instrumentations if that's the granularity of labeling you choose to maintain) is done based on the fact whether you are able to take the responsibility to fulfill the requirements of the "stable" labeling. What "stable" label means is described in this document.

Thanks @tigrannajaryan. Given that we don't want to mark the entire javaagent distribution as "unstable", it sounds like we will need to disable all "unstable" auto-instrumentation in the distribution by default, and only enable them (by default) once their respective semantic conventions have reached stability and we can commit to the telemetry stability described in this document.

jmacd (approved):
There are a number of questions about metrics here. The Views support in each metrics SDK should show it's possible to perform attribute removal safely (correctly, and relatively easily compared with doing so after-the-fact). This topic came up in #2208. I expect we will eventually be able to support metrics schema translation via Views and relax these restrictions, but it's good to be conservative for now.

jsuereth:
> Thanks @tigrannajaryan. Given that we don't want to mark the entire javaagent distribution as "unstable", it sounds like we will need to disable all "unstable" auto-instrumentation in the distribution by default, and only enable them (by default) once their respective semantic conventions have reached stability and we can commit to the telemetry stability described in this document.

I think we need room here for the Java Agent *itself* to be stable with stable instrumentation but bundle in unstable components that users can make use.  

Specifically, I think we need to walk a fine tightrope of the following "features" for users:

- We need to provide users stability, and a GA offering.  New users won't leverage our offerings if they're not marked stable, and "churn" in instrumentation will drive away our existing users.
- We need the ability to experiment across the huge SCOPE that is instrumentation.   As an example, there was a discussion in the Java instrumentation SiG about DB/ORM span instrumentation not being super useful and a first cut for users (not even on by default in some of the vendor releases).

I think *if* this restriction made it extremely clear to users what is considered stable and what may change in the Java agent, with clear enable/disable flags that'd be ideal.   However, I still feel like we're missing the configuration/enablement story for how the instrumentation SiG can continue experimenting/building instrumentation.  @trask is it clear to you what you would do here?

trask:
> @trask is it clear to you what you would do here?

Yes, I believe we would hide all unstable auto-instrumentations in the OpenTelemetry Javaagent behind an experimental flag.

tigrannajaryan:
> > @trask is it clear to you what you would do here?
> 
> Yes, I believe we would hide all unstable auto-instrumentations in the OpenTelemetry Javaagent behind an experimental flag.

@trask This may be the right approach regardless of this PR, right?

trask:
> > Yes, I believe we would hide all unstable auto-instrumentations in the OpenTelemetry Javaagent behind an experimental flag.
> 
> @trask This may be the right approach regardless of this PR, right?

It may be. Though I'm pretty sure all the vendors are going to enable this experimental flag in their downstream distribution/extension right away, as the OpenTelemetry Javaagent is not very useful when it emits no telemetry (and we currently have no stable semantic conventions, therefore no stable telemetry producing auto-instrumentation).

iNikem:
Is it a requirement for ALL telemetry that auto-instrumentation produces and wants to declare stable to have a corresponding semantic convention in the spec? E.g. some library-specific attributes in addition to the established db semantic convention - do they need all be described in the spec?

iNikem:
And related question to the one above: if some language-specific (as in "library exists only for one language") auto-instrumentation e.g. for some DB client wants to provide stable attributes both from DB semantic convention and some library-specific ones, which schema should this client use? We want to reference official Otel schema for DB semantic convention AND we want some schema for library-specific attributes.

tigrannajaryan:
> > > Yes, I believe we would hide all unstable auto-instrumentations in the OpenTelemetry Javaagent behind an experimental flag.
> > 
> > 
> > @trask This may be the right approach regardless of this PR, right?
> 
> It may be. Though I'm pretty sure all the vendors are going to enable this experimental flag in their downstream distribution/extension right away, as the OpenTelemetry Javaagent is not very useful when it emits no telemetry (and we currently have no stable semantic conventions, therefore no stable telemetry producing auto-instrumentation).

@trask If every vendor is going to enable the experimental flag then we should consider that that is the normal situation and not really "experimental". If that's what is going to happen I think the declarations we make in this spec become useless. The point of declaring the stability is to help real users to avoid breakage because instrumentations change. 

It seems like you are saying instrumentations are not ready to be stable, they are still in a phase where they need to experiment. To me this sounds like we have a couple choices:
1. Declare the entire artifact (e.g. java jar) that contains multiple instrumentations as unstable. I think instrumentation stability is essential part of stability guarantees. If we as artifact cannot provide the guarantees then let's label it honestly.
2. Change the granularity of stable/unstable labelling. Instead of declaring the entire java-instrumentation jar as stable or unstable let's apply the stable/unstable label to individual instrumentations, e.g. `instrumentation/apache-httpclient/` is stable while `instrumentation/cassandra/` is not.
3. Perhaps an alternate approach is to split the artifact into 2 parts: stable and unstable. The stable jar contains stable instrumentations only. As some instrumentation becomes stable move it from the unstable jar to stable.
There are probably other approaches possible. I do not think the approach that "we will declare everything unstable, then expect every vendor to enable all unstable instrumentation by default in their distro" works. To me it sounds like we don't want to take the responsibility and actually provide stability guarantees users expect.

iNikem:
> 2\. Change the granularity of stable/unstable labelling. Instead of declaring the entire java-instrumentation jar as stable or unstable let's apply the stable/unstable label to individual instrumentations, e.g. `instrumentation/apache-httpclient/` is stable while `instrumentation/cassandra/` is not.

That is more or less what we probably will do. We already explicitly declare that changes in telemetry are still possible until we have stable semantic convention. When http semantic convention is declared stable, we will say that "these and these instrumentation are now stable and here is schema file for them. All remaining are still unstable".

tigrannajaryan:
> That is more or less what we probably will do. We already explicitly declare that changes in telemetry are still possible until we have stable semantic convention. When http semantic convention is declared stable, we will say that "these and these instrumentation are now stable and here is schema file for them. All remaining are still unstable".

This sounds good to me. If we go with this approach then the wording proposed in this PR only applies to the parts that are declared stable. The unstable parts are free to change.

iNikem:
> 

@tigrannajaryan Can you please clarify this?

tigrannajaryan:
> > 
> 
> @tigrannajaryan Can you please clarify this?

I think to move forward with this PR we need to clarify how unstable instrumentations can work without being stifled by these requirements. For this we are going to:
1. Define how to indicate unstable instrumentation in the emitted telemetry (on the wire). I will work on this.
1. Come up with a way to correctly label instrumentation libraries (stable/unstable or alpha/beta/stable). @jsuereth is going to work on this.

tigrannajaryan:
@jsuereth @trask @iNikem @jack-berg 

I spent a bit of time thinking about unstable instrumentations.

I believe that we need to allow unstable instrumentations to change over time without being subject to the restrictions imposed by schemas.

I was initially thinking that we need to somehow indicate that the telemetry is emitted by such unstable instrumentation by having a specific way to record this fact on the wire. However, I no longer think so. I believe it is sufficient for unstable instrumentations to simply omit the Schema URL on the wire. For receivers of such telemetry the fact that the Schema URL is omitted is an indication that the telemetry is not subject to the concept of schemas. There may be multiple reasons for such omission, unstable instrumentation being one of the possible reasons. I do not see a need for receivers of the telemetry to be able to differentiate unstable instrumentations from instrumentations that omit the Schema URL on the wire for some other reason. (If you see such a reason please tell).

Because of this I suggest to slightly modify this PR in the following way:

### Unstable instrumentations
- Do not include Schema URL in the emitted telemetry.
- Authors of unstable instrumentations are free to modify the instrumentation, provided that the instrumentation is clearly labeled as unstable.
- Receivers of telemetry emitted by unstable instrumentations cannot make any assumptions about the schema of the telemetry since the Schema URL is not present on the wire.


### Stable instrumentations
- If you are not including Schema URL in the emitted telemetry then you are not allowed to modify the telemetry (with certain exceptions carved out in the doc).
- If you are including Schema URL then you are allowed to modify the telemetry after certain date, provided that they are described by a published Schema File (with the same exceptions as above applicable).

### What is Stable vs Unstable Instrumentation

We need a clear labeling approach that should be adopted by all instrumentation. It should be possible to label as stable or unstable each individual library in our instrumentation packages (that may include more than one library). The labels should be clearly communicated to the users of such instrumentation, the labels are an essential part of the instrumentation documentation. Based on how each individual library is labeled they must follow the Schema-related rules described above. @jsuereth I assume you will work on this labelling approach and will propose it in a separate PR or OTEP.

--

Let me know what you all think.

iNikem:
I am confused about what receivers of telemetry should think about its stability when telemetry does NOT have schema associated with it. Both stable and unstable instrumentations may produce such data. Receiver cannot reason about it.

Would it be better to make schema a requirement for stable instrumentations?

tigrannajaryan:
> I am confused about what receivers of telemetry should think about its stability when telemetry does NOT have schema associated with it. Both stable and unstable instrumentations may produce such data. Receiver cannot reason about it.

@iNikem Why do the receivers need to reason about it? What difference does it make for receivers?

iNikem:
> > I am confused about what receivers of telemetry should think about its stability when telemetry does NOT have schema associated with it. Both stable and unstable instrumentations may produce such data. Receiver cannot reason about it.
> 
> @iNikem Why do the receivers need to reason about it? What difference does it make for receivers?

I think what I am confused about is that I, as an author, can omit schema url from my telemetry both for experimental instrumentation (that I can change however I want) and for stable instrumentation (that I cannot change in almost any way).

And I don't understand why do we need stable instrumentations without schema? Why do we allow that?

tigrannajaryan:
> And I don't understand why do we need stable instrumentations without schema? Why do we allow that?

If you want to declare your instrumentation stable but did not yet have a chance to add support for schemas to your instrumentation it should be OK to do to so temporarily, until you have time to add the support for schemas. If we do not allow this then the support for schemas becomes a pre-requisite for anyone to ever release any stable instrumentation and I don't want to create this limitation for everyone in Otel. I want to ask for a less restrictive limitation: it is OK to declare your instrumentation stable even if you don't support schemas yet, but if you do so please make sure you do not change the emitted telemetry. I think it is an acceptable limitation and allows instrumentation authors here at Otel to choose their own priorities, and decide when exactly they want to start supporting schemas.

Maybe I am being overly cautious but I want to avoid potentially creating roadblocks for lots of contributors here at Otel. I prefer to introduce support for schemas gradually were possible.

iNikem:
I see your reason. I do think you are too cautious, but I will not pick this battle :) Other than this particular question, I like and support your proposal in https://github.com/open-telemetry/opentelemetry-specification/pull/2180#issuecomment-1006846264

trask:
@tigrannajaryan thanks 👍

I think the last remaining item that we need clarity on from the javaagent perspective is:

> It should be possible to label as stable or unstable each individual library in our instrumentation packages (that may include more than one library) ... @jsuereth I assume you will work on this labelling approach and will propose it in a separate PR or OTEP.

It sounds like as long as we document the varying stability of each individual instrumentation, then we don't have to declare the entire javaagent as "unstable", but looking forward to this being clarified in the follow-up PR/OTEP.

tigrannajaryan:
> It sounds like as long as we document the varying stability of each individual instrumentation, then we don't have to declare the entire javaagent as "unstable", but looking forward to this being clarified in the follow-up PR/OTEP.

Yes, I think we should be able to release a "stable" javaagent where some subset of individual instrumentation libraries are marked as "unstable". Kind of like how Kubernetes marks individual features as "alpha", but which does not mean that the entire product is "alpha".

tigrannajaryan:
Not stale, need time to update this.

tigrannajaryan:
@trask @jsuereth @iNikem @jack-berg I made substantial changes to this PR, separated Unstable vs Stable instrumentations with clear rules for each category.

Please have a look.

trask (approved):

specification/telemetry-stability.md
```diff
@@ -0,0 +1,92 @@
...
+## Unstable Instrumentations
+
+Unstable telemetry-producing instrumentations (unstable instrumentations for short) SHOULD
+be clearly labeled so by any means the instrumentations authors consider idiomatic for
+their language, e.g. via version numbers, artifact names, documentation, etc.
+
+Unstable instrumentations provide no guarantees about the shape of
+the telemetry they produce and how that shape changes over time, from version to version.
+Span or metric names, attributes of any telemetry items may change without any
+restrictions. All produced telemetry MUST have an unspecified Schema URL.
```
Aneurysm9, line 39: Why this restriction?  If telemetry is unstable but conforms to a scheme, why can it not declare that?

tigrannajaryan: My thinking was that "conforms to a schema" and "unstable" are mutually exclusive. If it conforms to a schema then it falls into the "Schema-File Driven Telemetry Producers" bucket. These are prohibited from emitting telemetry that is not described by semantic convention OR if they do so they are prohibited from changing such telemetry over time. Essentially this rules out the possibility of any unstable telemetry to be a "Schema-File Driven Telemetry Producer".

So, the presence of the Schema URL in the emitted telemetry is a signal to the recipient that the producer is schema-file driven and the recipient can make certain assumptions about it. If we allow unstable instrumentations to also specify Schema URL this assumption is no longer valid and we have to indicate by some other means that the telemetry is unstable. 

Anyway, "Unstable, but conform to a schema" sounds like a contradiction to me. Perhaps I am wrong and there is a use case for this?

tsloughter: Maybe the use of the term unstable and stable is confusing here? Since one would consider a schema version of 0.x.y to be unstable, but in a different way.

I don't know of a better term off the top of my head, just a thought.

specification/telemetry-stability.md
```diff
@@ -0,0 +1,92 @@
...
+instrumentations MUST NOT change such telemetry, regardless of whether they
+are fixed-schema producers or schema-file driven producers. Once the produced telemetry
+is added to the semantic conventions, changes will be allowed as described below.
+
+### Fixed Schema Telemetry Producers
+
+Instrumentations that are labeled `Stable` and do not include the Schema URL in the
+produced telemetry are called Fixed Schema Telemetry Producers.
+
+Such instrumentation are prohibited from changing any produced telemetry. If the
```
pirgeo, line 64:
```suggestion
Such instrumentations are prohibited from changing any produced telemetry. If the
```

specification/telemetry-stability.md
```diff
@@ -0,0 +1,92 @@
...
+specification changes over time and the semantic conventions are updated the
```
pirgeo, line 65:
```suggestion
specification changes over time and the semantic conventions are updated, the
```

specification/telemetry-stability.md
```diff
@@ -0,0 +1,92 @@
...
+### Schema-File Driven Telemetry Producers
+
+Stable instrumentations that include the Schema URL in the produced telemetry are
+called Schema-File Driven Telemetry Producers.
+
+Such instrumentations are prohibited from changing the produced telemetry until
+April 1, 2023 and until that date are subject to exactly the same restrictions as
+[Fixed Schema Telemetry Producers](#fixed-schema-telemetry-producers).
+
+After April 1, 2023 Sgable instrumentations are allowed to change the produced telemetry
```
pirgeo, line 80:
```suggestion
After April 1, 2023, stable instrumentations are allowed to change the produced telemetry
```
Unsure if I got this right.

specification/telemetry-stability.md
```diff
@@ -0,0 +1,92 @@
...
+if all the following conditions are fulfilled:
+
+- The change is part of OpenTelemetry semantic conventions and are in a released
```
pirgeo, line 83:
```suggestion
- The change is part of OpenTelemetry semantic conventions and is in a released
```

specification/telemetry-stability.md
```diff
@@ -0,0 +1,92 @@
...
+  version of the specification,
+- The change has a corresponding
+  [published](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/overview.md#opentelemetry-schema)
+  OpenTelemetry Schema File that describes the change.
+- The produced telemetry correctly specifies the Schema URL.
+
+If the change was introduced in the semantic conventions specification before
+April 1, 2023 the instrumentations must wait until April 1, 2023 before they can adopt
```
pirgeo, line 91:
```suggestion
April 1, 2023, the instrumentations must wait until April 1, 2023 before they can adopt
```

specification/telemetry-stability.md
tigrannajaryan: Fixed.

specification/telemetry-stability.md
tigrannajaryan: Fixed.

specification/telemetry-stability.md
tigrannajaryan: Fixed.

specification/telemetry-stability.md
tigrannajaryan: Fixed.

specification/telemetry-stability.md
Aneurysm9: I think it is possible for an instrumentation's implementation to be unstable while it is still using a schema and exporting likely stable telemetry using a schema.  I don't think it is necessary to prohibit emitting a schema version until the implementation reaches v1.0.

specification/telemetry-stability.md
```diff
@@ -0,0 +1,92 @@
+# Telemetry Stability
+
```
arminru, line 2:
```suggestion

**Status**: [Experimental](document-status.md)

```

specification/telemetry-stability.md
tigrannajaryan: Changed the wording to allow unstable instrumentation to specify Schema URL. Also added a TODO to decide if we need to define how unstable instrumentation is indicated on the wire.

trask (approved):

specification/telemetry-stability.md
```diff
@@ -0,0 +1,98 @@
...
+## Unstable Instrumentations
+
+Unstable telemetry-producing instrumentations (unstable instrumentations for short) SHOULD
+be clearly labeled so by any means the instrumentations authors consider idiomatic for
+their language, e.g. via version numbers, artifact names, documentation, etc.
+
+Unstable instrumentations provide no guarantees about the shape of
+the telemetry they produce and how that shape changes over time, from version to version.
+Span or metric names, attributes of any telemetry items may change without any
+restrictions. The produced telemetry MAY specify a Schema URL if the telemetry data
```
iNikem, line 41: If unstable instrumentation uses schema, is it required to record all its changes in that schema and make only those changes that can be described in its schema file?

specification/telemetry-stability.md
```diff
@@ -0,0 +1,98 @@
...
+conforms to a particular Schema.
+
+Unstable instrumentations authored by OpenTelemetry MAY produce telemetry that is
+not described by OpenTelemetry semantic conventions.
+
+TODO: decide if it is necessary to indicate on the wire if the produced telemetry is
```
iNikem, line 47: I still think that the presence of schema url on the wire could serve as a good indicator of telemetry stability. Why exactly do we need "Fixed Schema Telemetry Producers"?

specification/telemetry-stability.md
tigrannajaryan: Unstable instrumentation can change any time and it is not required that these changes are reflected in schema files. That's the point of unstable instrumentation: they can break the schema rules.

specification/telemetry-stability.md
tigrannajaryan: "Fixed Schema Telemetry Producers" are instrumentations which want to declare themselves "Stable" but do not yet support schemas. This essentially is another way of saying: "if you want to change emitted telemetry you must support the concept of schemas and must include schema URL in the emitted telemetry".

tigrannajaryan:

Aneurysm9 (approved):

specification/telemetry-stability.md
```diff
@@ -0,0 +1,98 @@
...
+be clearly labeled so by any means the instrumentations authors consider idiomatic for
+their language, e.g. via version numbers, artifact names, documentation, etc.
+
+Unstable instrumentations provide no guarantees about the shape of
+the telemetry they produce and how that shape changes over time, from version to version.
+Span or metric names, attributes of any telemetry items may change without any
+restrictions. The produced telemetry MAY specify a Schema URL if the telemetry data
+conforms to a particular Schema.
+
+Unstable instrumentations authored by OpenTelemetry MAY produce telemetry that is
```
arminru, line 44:
```suggestion
Unstable instrumentations authored by OpenTelemetry MAY produce additional telemetry that is
```
Otherwise it sounds as if it was explicitly allowed to diverge from the spec although a semantic convention is already defined for the instrumented technology/framework/area.

specification/telemetry-stability.md
```diff
@@ -0,0 +1,98 @@
...
+not described by OpenTelemetry semantic conventions.
+
+TODO: decide if it is necessary to indicate on the wire if the produced telemetry is
+coming from an unstable instrumentation.
```
arminru, line 48: This could be indicated by the version number (-alpha, -beta, 0.x) used for the instrumentation library (tracer/meter name)?

specification/telemetry-stability.md
```diff
@@ -0,0 +1,98 @@
...
+
+Such instrumentations are prohibited from changing the produced telemetry until
+April 1, 2023 and until that date are subject to exactly the same restrictions as
+[Fixed Schema Telemetry Producers](#fixed-schema-telemetry-producers).
+
+After April 1, 2023, stable instrumentations are allowed to change the produced telemetry
+if all the following conditions are fulfilled:
+
+- The change is part of OpenTelemetry semantic conventions and is in a released
+  version of the specification,
```
arminru, line 90:
```suggestion
  version of the specification.
```

specification/telemetry-stability.md
```diff
@@ -0,0 +1,98 @@
...
+- The change has a corresponding
+  [published](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/overview.md#opentelemetry-schema)
+  OpenTelemetry Schema File that describes the change.
+- The produced telemetry correctly specifies the Schema URL.
```
arminru, line 94:
```suggestion
- The produced telemetry correctly specifies the respective Schema URL.
```

specification/versioning-and-stability.md
```diff
@@ -134,11 +134,14 @@ Languages which ship binary artifacts SHOULD offer [ABI compatibility](glossary.
...
-
-Plugins, instrumentation, and other contrib packages SHOULD be kept up to date and compatible with the latest versions of the API, SDK, and Semantic Conventions.
-If a release of the API, SDK, or Semantic Conventions contains changes which are relevant to a contrib package, that package SHOULD be updated and released in a timely fashion.
-The goal is to ensure users can update to the latest version of OpenTelemetry, and not be held back by the plugins that they depend on.
+Plugins, instrumentation, and other contrib packages SHOULD be kept up to date
+and compatible with the latest versions of the API, SDK, and Semantic
+Conventions. If a release of the API, SDK, or Semantic Conventions contains
+changes which are relevant to a contrib package, that package SHOULD be updated
+and released in a timely fashion. (See limitations on instrumentation stability
+in [Telemetry Stability](telemetry-stability.md)). The goal is to ensure users can
```
arminru, line 142:
```suggestion
in [Telemetry Stability](telemetry-stability.md).) The goal is to ensure users can
```

specification/versioning-and-stability.md
```diff
@@ -149,20 +152,50 @@ For example, a database integration may break stability if the required database
...
 A new, incompatible version of an integration SHOULD be released as a separate contrib package, rather than break the existing contrib package.
 
-#### NOT DEFINED: Telemetry Stability
+#### Semantic Conventions Stability
+
+Changes to telemetry produced by OpenTelemetry instrumentation SHOULD avoid
+breaking analysis tools, such as dashboards and alerts. To achieve this, while
+allowing the evolution of telemetry and semantic conventions, OpenTelemetry
+relies on the concept of
+[schemas](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/overview.md).
```
arminru, line 161:
```suggestion
[Telemetry Schemas](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/overview.md).
```

specification/versioning-and-stability.md
```diff
@@ -149,20 +152,50 @@ For example, a database integration may break stability if the required database
...
+
+Changes to semantic conventions in this specification are allowed, provided that
+the changes can be described by schema files. The following changes can be
+currently described and are allowed:
+
+- Renaming of attributes.
```
arminru, line 167:
```suggestion
- Renaming of span, metric and log attributes.
```

specification/versioning-and-stability.md
```diff
@@ -149,20 +152,50 @@ For example, a database integration may break stability if the required database
...
+- Renaming of span events.
+
+All such changes MUST be described in the OpenTelemetry
+[Schema File Format](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/file_format_v1.0.0.md)
+and published
+[here](https://github.com/open-telemetry/opentelemetry-specification/tree/main/schemas).
+For details see [how OpenTelemetry Schemas are
+published](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/overview.md#opentelemetry-schema).
+
+See [Telemetry Stability](telemetry-stability.md) document for details on how
```
arminru, line 178:
```suggestion
See the [Telemetry Stability](telemetry-stability.md) document for details on how
```

specification/versioning-and-stability.md
```diff
@@ -149,20 +152,50 @@ For example, a database integration may break stability if the required database
...
 
-**Telemetry stability guarantees are TBD.**
+In addition to the 3 types of changes described above there are certain types
+that are always allowed. Such changes do not need to be described (and are not
+described) by schema files. Here is the list of such changes:
 
-Changes to telemetry produced by OpenTelemetry instrumentation SHOULD avoid breaking analysis tools, such as dashboards and alerts.
-However, it is not clear at this time what type of instrumentation changes (for example, adding additional spans and attributes) would actually cause a breaking change.
+- Adding new attributes to the existing semantic conventions for resources,
+  spans, span events or log records.
```
arminru, line 186:
```suggestion
  spans, span events or log records.
  Note that adding attributes to existing metrics is NOT considered non-breaking.
```
One might miss this subtle difference between the two points and I think it's worth pointing this out explictly.

specification/versioning-and-stability.md
```diff
@@ -149,20 +152,50 @@ For example, a database integration may break stability if the required database
...
+is only allowed if OpenTelemetry introduces a new schema file format that is
+capable of describing such changes.
 
-Telemetry stability, including semantic conventions, is not currently defined. The following practices are recommended.
+#### Telemetry Stability
 
-Semantic Conventions SHOULD NOT be removed once they are added.
-New conventions MAY be added to replace usage of older conventions, but the older conventions SHOULD NOT be removed.
-Older conventions SHOULD be marked as deprecated when they are replaced by newer conventions.
+For stability of telemetry produced by instrumentation see
```
arminru, line 197:
```suggestion
For stability of telemetry produced by instrumentation see the
```

arminru (approved):
This looks good to me for a first iteration. Thank you, Tigran!
The open issue about indicating (in)stability can be tackled in a separate follow-up PR as this is still marked experimental.

specification/telemetry-stability.md
tigrannajaryan: Added.

specification/telemetry-stability.md
tigrannajaryan: Fixed.

specification/versioning-and-stability.md
tigrannajaryan: Fixed.

specification/versioning-and-stability.md
tigrannajaryan: Added.

specification/versioning-and-stability.md
tigrannajaryan: Fixed.

specification/versioning-and-stability.md
tigrannajaryan: Added in a slightly re-worded form.

specification/versioning-and-stability.md
tigrannajaryan: Fixed.

specification/telemetry-stability.md
tigrannajaryan: Yes, that is a possibility.

specification/telemetry-stability.md
```diff
@@ -0,0 +1,98 @@
...
+## Stable Instrumentations
+
+Stable telemetry-producing instrumentations (stable instrumentations for short) SHOULD
+be clearly labeled so by any means the instrumentations authors consider idiomatic for
+their language, e.g. via version numbers, artifact names, documentation, etc.
+
+Stable instrumentations fall into 2 categories: fixed-schema producers and schema-file
+driven producers.
+
+Stable instrumentations authored by OpenTelemetry SHOULD NOT produce telemetry that is
```
dyladan, line 59: Does this mean an instrumentation should not provide any excess span attributes for instance? Perhaps we should have an extension namespace like `ext.my.custom.attribute` which are specifically not covered by stability guarantees?

jack-berg (approved):

specification/telemetry-stability.md
trask: this has been our interpretation in java instrumentation, and we hide all excess span attributes behind a flag, e.g. `otel.instrumentation.grpc.experimental-span-attributes`

tigrannajaryan: >Does this mean an instrumentation should not provide any excess span attributes for instance?

Yes, if the instrumentation is declared Stable.

>Perhaps we should have an extension namespace like ext.my.custom.attribute which are specifically not covered by stability guarantees?

This is an interesting suggestion. This way Stable instrumentations will still have a way to continue experimenting more freely without being bound by semantic convention acceptable into specification and Schema File release cycles. If we see the demand we can allow this exception.

specification/versioning-and-stability.md
```diff
@@ -149,20 +152,51 @@ For example, a database integration may break stability if the required database
...
 A new, incompatible version of an integration SHOULD be released as a separate contrib package, rather than break the existing contrib package.
 
-#### NOT DEFINED: Telemetry Stability
+#### Semantic Conventions Stability
+
+Changes to telemetry produced by OpenTelemetry instrumentation SHOULD avoid
+breaking analysis tools, such as dashboards and alerts. To achieve this, while
+allowing the evolution of telemetry and semantic conventions, OpenTelemetry
+relies on the concept of
+[Telemetry Schemas](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/overview.md).
```
reyang, line 161: Change to relative link.

reyang (changes requested):
There are several absolute links which need to be changed to relative (please feel free to dismiss it once the links are fixed).

specification/telemetry-stability.md
dyladan: I can tell you for certain there has been demand in javascript. Also, it's not clear how strictly we're meant to enforce these types of things in contrib repositories for example. In JS at least we have given a lot of control over the packages to the person or org that contributed them in terms of which features are included and when releases are made. Are we meant to be enforcing things like this, or do these restrictions only apply to "official" instrumentations released through the official non-contrib repositories?

tigrannajaryan:
@dyladan 

> I don't see anything about how to handle cases around early prototypes.

Is "Unstable Instrumentations" section insufficient? The prototypes are supposedly unstable and can be governed by the rules in this section I assume. Are you looking for something else?

> First implementations of instrumentations may not be feature complete, and may not want to strictly adhere even to the schema unstable promise. In alpha versions, breaking changes are an expected part of the development process.

Assuming "first implementation" is unstable this seems to be covered by "Unstable Instrumentations", which allows any changes to the emitted telemetry.


> I also don't see anything about how to handle the case where there is no semantic convention for a particular usecase.

I think this is already covered in the document:
```
Unstable instrumentations authored by OpenTelemetry MAY produce additional telemetry that
is not described by OpenTelemetry semantic conventions.
```
and
```
Stable instrumentations authored by OpenTelemetry SHOULD NOT produce telemetry that is
not described by OpenTelemetry semantic conventions. If, however, this rule is broken the
instrumentations MUST NOT change such telemetry, regardless of whether they
are fixed-schema producers or schema-file driven producers. Once the produced telemetry
is added to the semantic conventions, changes will be allowed as described below.
```
Is this unclear or not specific enough?

dyladan:
> @dyladan
> 
> > I don't see anything about how to handle cases around early prototypes.
> 
> Is "Unstable Instrumentations" section insufficient? The prototypes are supposedly unstable and can be governed by the rules in this section I assume. Are you looking for something else?
> 
> > First implementations of instrumentations may not be feature complete, and may not want to strictly adhere even to the schema unstable promise. In alpha versions, breaking changes are an expected part of the development process.
> 
> Assuming "first implementation" is unstable this seems to be covered by "Unstable Instrumentations", which allows any changes to the emitted telemetry.
> 
> > I also don't see anything about how to handle the case where there is no semantic convention for a particular usecase.
> 
> I think this is already covered in the document:
> 
> ```
> Unstable instrumentations authored by OpenTelemetry MAY produce additional telemetry that
> is not described by OpenTelemetry semantic conventions.
> ```
> 
> and
> 
> ```
> Stable instrumentations authored by OpenTelemetry SHOULD NOT produce telemetry that is
> not described by OpenTelemetry semantic conventions. If, however, this rule is broken the
> instrumentations MUST NOT change such telemetry, regardless of whether they
> are fixed-schema producers or schema-file driven producers. Once the produced telemetry
> is added to the semantic conventions, changes will be allowed as described below.
> ```
> 
> Is this unclear or not specific enough?

Hmm no I think I must have just misread. I was reading quickly and guess I wasn't careful enough and got the unstable and stable sections mashed together in my head.

specification/telemetry-stability.md
```diff
@@ -0,0 +1,97 @@
...
+  * [Schema-File Driven Telemetry Producers](#schema-file-driven-telemetry-producers)
+
+<!-- tocstop -->
+
+</details>
+
+This section defines stability requirements for telemetry produced by
+OpenTelemetry instrumentations.
+
+All OpenTelemetry-authored instrumentations are labeled to be either `Unstable` or `Stable`
```
alanwest, line 22: What does OpenTelemetry-authored mean? Does it mean all instrumentation sourced from a repo in the open-telemetry org? Can there be any exclusions to this - e.g., instrumentations in contrib repositories?

dyladan: > What does OpenTelemetry-authored mean? Does it mean all instrumentation sourced from a repo in the open-telemetry org? Can there be any exclusions to this - e.g., instrumentations in contrib repositories?

That's what I was getting at here: https://github.com/open-telemetry/opentelemetry-specification/pull/2180#discussion_r836607529

tigrannajaryan: >What does OpenTelemetry-authored mean? Does it mean all instrumentation sourced from a repo in the open-telemetry org? 

Yes, that's essentially what it means.

>Can there be any exclusions to this - e.g., instrumentations in contrib repositories?

Do we need any exclusions? We already have a form of exclusion, we call it "Unstable Instrumentation". If this is insufficient and we need another exclusion rule I can add it.

alanwest: > Do we need any exclusions?

I'm not sure yet. Like @dyladan mentions with the JS community, it is also the .NET community's intent to give the original contributor a fair amount of control over packages in the contrib repo. Some of these folks are not even members of the OpenTelemetry org much less well indoctrinated in the requirements of otel instrumentation. Might just be a matter of setting appropriate expectations with these folks.

I plan to bring this topic up in this week's .NET SIG meeting (tomorrow). Will share any suggestions or concerns.

tigrannajaryan: OK, we can keep this topic open. It does not prevent merging of the PR. The status of this document is still experimental so we can add other exclusions as necessary.

tigrannajaryan:
@Oberon00 please take another look.

tigrannajaryan:
@open-telemetry/python-maintainers can you please take a look? This affects instrumentations in all languages. It would be good to have your approval if you agree with this approach.

reyang (approved):

alanwest (approved):

specification/telemetry-stability.md
alanwest: .NET discussed this a bit. Given that releasing packages is in the hands of maintainers of the contrib repo, it's reasonable for them to gate unstable/stable at this point in time.

specification/telemetry-stability.md
jmacd: > excess span attributes

I am having trouble reconciling this discussion with the statements below, allowing changes to semantic conventions in [versioning and stability of semantic conventions below (this PR)](https://github.com/open-telemetry/opentelemetry-specification/pull/2180/files#diff-e7de8bed9eb1605a44c385930415011d900acccecc8b48398a75ce9fa707cdd3R182), where addition of span attributes is always allowed.

tigrannajaryan: >addition of span attributes is always allowed.

@jmacd Addition of attributes to _semantic conventions_ for existing Spans, Resources, Logs is always allowed and considered a non-breaking change (not that I can think of an actual semantic convention that involves a Span, but anyway). Such addition does not require a new Schema version.

However, including a span attribute in the telemetry, where such attribute is not defined in the semantic conventions is strongly discouraged (with a SHOULD NOT clause) for Stable instrumentations. If there is a need like this then the attribute first has to be added to semantic conventions and then it can be added to the telemetry.

tigrannajaryan:
@dyladan if the PR looks good to you please approve on behalf of JS SIG. I want to have an approval from each (or at least most) language SIG.

dyladan (approved):
We discussed this at the JS meeting last wednesday and the other @open-telemetry/javascript-maintainers brought up the same points as have been brought up by myself, @alanwest, and others around instrumentation which is not in the semantic conventions. There will always be some instrumentations which for whatever reason are not covered by the semantic conventions whether they have too small of a userbase, are too specific, or provide some custom functionality. This PR seems to prevent those instrumentations from ever being marked as "stable" which will limit their adoption.

From a maintenance standpoint it is not too big of a burden for maintainers to gate stable on the guarantees outlined here, but I believe the restrictions will prevent people from upstreaming contributions to the contrib repo unless we have some way to provide custom functionality that isn't covered by the semantic conventions.

As mentioned earlier, this is still marked experimental so isn't blocking, but I want to make sure the point isn't forgotten before this is finalized and made stable.

tigrannajaryan:
I believe so far we have approvals from JS, Go, Java, .Net maintainers.

trask:
> There will always be some instrumentations which for whatever reason are not covered by the semantic conventions whether they have too small of a userbase, are too specific, or provide some custom functionality. This PR seems to prevent those instrumentations from ever being marked as "stable" which will limit their adoption.

@tigrannajaryan I thought there were two options for instrumentation that are not covered by semantic conventions to go stable:

* Fixed Schema Telemetry Producers - decide not to evolve the schema via the standard schema mappings
* Schema-File Driven Telemetry Producers - publish your own schema file (we have discussed doing this in Java for these one-off instrumentations that do not have a future in the specification repo)

tigrannajaryan:
> > There will always be some instrumentations which for whatever reason are not covered by the semantic conventions whether they have too small of a userbase, are too specific, or provide some custom functionality. This PR seems to prevent those instrumentations from ever being marked as "stable" which will limit their adoption.
> 
> @tigrannajaryan I thought there were two options for instrumentation that are not covered by semantic conventions to go stable:
> 
> * Fixed Schema Telemetry Producers - decide not to evolve the schema via the standard schema mappings
> * Schema-File Driven Telemetry Producers - publish your own schema file (we have discussed doing this in Java for these one-off instrumentations that do not have a future in the specification repo)

@trask that is correct. And for both options they should also follow this rule:
```
Stable instrumentations authored by OpenTelemetry SHOULD NOT produce telemetry that is not described by OpenTelemetry semantic conventions.
```
I think @dyladan's concern is that this is too much limitation and the rules are too rigid for instrumentations contributed by non-core maintainers.

trask:
> And for both options they should also follow this rule:
> 
> 
> > Stable instrumentations authored by OpenTelemetry SHOULD NOT produce telemetry that is not described by OpenTelemetry semantic conventions.

oh, I missed this, thx! I agree this is too limiting for the reason @dyladan mentions:

> There will always be some instrumentations which for whatever reason are not covered by the semantic conventions whether they have too small of a userbase, are too specific, or provide some custom functionality

carlosalberto (approved):

specification/telemetry-stability.md
```diff
@@ -0,0 +1,97 @@
...
+from the perspective of the telemetry they produce.
+
+Adding of new metrics, spans, span events or log records and adding of
+new attributes to spans, span events, log records or resources are considered
```
cijothomas, line 26: is it intentional that "adding of new attributes to metrics" was not listed?

jack-berg: Yes, see this [note](https://github.com/open-telemetry/opentelemetry-specification/pull/2180/files#diff-e7de8bed9eb1605a44c385930415011d900acccecc8b48398a75ce9fa707cdd3R183-R184).

tigrannajaryan: Yes, adding new attributes to metrics is a breaking change and is intentionally not listed.

cijothomas: Thanks for clarifying.

robertlaurin (approved):

fbogsany (approved):

specification/telemetry-stability.md
srikanthccv: There are instrumentations where we provide callback/hook for end-user to customise the attributes added to span along with standard attributes. This strict restriction would mean instrumentation can't be stable?

tigrannajaryan: I think the "SHOULD" clause allows such behavior. By calling a hook you hand over the responsibility to the end user. If  they deviate from expected behavior that's on them.

srikanthccv (approved):

lzchen (approved):

tigrannajaryan:
> I believe so far we have approvals from JS, Go, Java, .Net maintainers.

In addition we now have approvals from Ruby and Python maintainers.

I believe this is good enough and we can move forward with this approach. Merging the PR.

PR #2182, opened 2021-12-01
Clarify log name field

### LINKED ISSUE #2074: Is Name field necessary in Log Data Model
The log [data model](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md) currently has a Name field.

It is not entirely clear why this field is needed.

Some of the example mappings provided use it this way:
- Syslog - MSGID. Could be an Attribute["syslog.msgid"] instead to match how PROCID is handled.
- SignalFx Events - EventType. Could be Attributes["com.splunk.signalfx.event_type"] to match how Category is handled.
- CloudTrail Log Event - errorCode. Unclear where we put this. Attributes["cloudtrail.error_code"]?
- Google Cloud Logging - log_name. This appears to be incorrectly designed, see https://github.com/open-telemetry/opentelemetry-specification/issues/1808

In most of these cases it appears that it could be placed in Attributes. The unclear one is CloudTrail Log Event and probably Google Cloud Logging.

Can we come up with clear justifications of why the Name field must exist and why for example Body is not enough?

jack-berg:
Clarifies that log name field is a low cardinality event classifier. 

Resolves #2074.

### DIFF
--- a/specification/logs/data-model.md
+++ b/specification/logs/data-model.md
@@ -183,7 +183,7 @@ SpanId         |Request span id.
 TraceFlags     |W3C trace flag.
 SeverityText   |The severity text (also known as log level).
 SeverityNumber |Numerical value of the severity.
-Name           |Short low cardinality event classifier.
+Name           |Short low cardinality event type.
 Body           |The body of the log record.
 Resource       |Describes the source of the log.
 Attributes     |Additional information about the event.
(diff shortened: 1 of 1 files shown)


specification/logs/data-model.md
```diff
@@ -183,7 +183,7 @@ SpanId         |Request span id.
 TraceFlags     |W3C trace flag.
 SeverityText   |The severity text (also known as log level).
 SeverityNumber |Numerical value of the severity.
-Name           |Short event identifier.
+Name           |Short low cardinality event classifier.
```
tigrannajaryan, line 186: All good except I am not sure "classifier" works better than "identifier".

jack-berg: I don't feel too strongly about this, but identifier has connotations of uniqueness, which is conflict with low cardinality.

Also considered "Show low cardinality event categorization".

djaglowski: I would slightly prefer "classifier" over "identifier" because of the cardinality-related connotations they carry. 

We might also consider using "type" or "kind".

tigrannajaryan: I also like "type" a bit more. Not a strong opinion.

tigrannajaryan (approved):

trask (approved):

kumoroku (approved):

Issue #1875, opened 2021-08-18
Log record receipt timestamp


pmm-sumo:
**What are you trying to achieve?**

Currently, logs data model describes a single (optional) [`Timestamp` field](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-timestamp) which describes when the event occurred

Unlike for e.g. spans, there are at least three ways a log record can have timestamp associated:

1. It can be set by a library supporting OTLP natively
2. It can be parsed out from the log Body or Attributes (e.g. using [timestamp operator](https://github.com/open-telemetry/opentelemetry-log-collection/blob/main/docs/types/timestamp.md) for filelogreceiver)
3. It can be set to receipt timestamp (which is a good fallback when the others are not available or fail)

While the first two can be considered mutually exclusive (if the record had timestamp already assigned, it does not make much sense to attempt parsing log body for it), this is not true with the last one. However, there's only one field where all of them can fit. 

This brings several problems, such as:
* Inability to fallback to receipt timestamp when parsing yielded incorrect value (also, no easy ability to find records with potential timestamp parsing errors)
* No way to to tell using what method the timestamp was assigned to the record 
* Lack of data to measure delay between log occurrence and log being collected

E.g. currently `filelogreceiver` sets `Timestamp` to receipt time - by default, or parsed timestamp - when timestamp operator is used. This makes it impossible to tell later if the log timestamp was parsed or not.

**Proposed solution**

Perhaps we could add another field to the log record which would store the receipt timestamp (`ReceiptTimestamp`). It would be filled in the record on the first receiver that retrieved it and then passed through. 

`Timestamp` field use would be limited only to storing the result of timestamp parsing or the value set explicitly in an OTLP-compatible library.

**Alternatives**

* the receipt timestamp could be instead stored as an optional attribute
* we could keep single Timestamp but add flags that would describe how it was created (also, attributes describing which parsing rule/template/regex was picked)

pmm-sumo:
@bogdandrutu as an example case, consider following log:

```
2021-08-19T11:55:15.498-0600	info	service/collector.go:242	Loading configuration..
```

While this is a common format, lets assume that the parsing rule missed the timezone information and the timestamp was parsed as: `2021-08-19 11:55:15.498 UTC` (while it should have been `2021-08-19 17:55:15.498 UTC`). 

[...]

PR #2183, opened 2021-12-02
Clarify the semantics of the Log Timestamp field

tigrannajaryan:
Contributes to https://github.com/open-telemetry/opentelemetry-specification/issues/1875

This is part 1 of the change. Part 2 will add the observed timestamp field. See the issue
for the discussion and the description of the source vs observed timestamps.

PR #2184, opened 2021-12-02
Add ObservedTimestamp to the Log Data Model

tigrannajaryan:
Resolves https://github.com/open-telemetry/opentelemetry-specification/issues/1875

See the issue above for the detailed discussion. A summary here:

Currently, logs data model describes a single (optional) [`Timestamp` field](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-timestamp) which describes when the event occurred

Unlike for e.g. spans, there are at least three ways a log record can have timestamp associated:

1. It can be set by a library supporting OTLP natively
2. It can be parsed out from the log Body or Attributes (e.g. using [timestamp operator](https://github.com/open-telemetry/opentelemetry-log-collection/blob/main/docs/types/timestamp.md) for filelogreceiver)
3. It can be set to receipt timestamp (which is a good fallback when the others are not available or fail)

While the first two can be considered mutually exclusive (if the record had timestamp already assigned, it does not make much sense to attempt parsing log body for it), this is not true with the last one. However, there's only one field where all of them can fit. 

This brings several problems, such as:
* Inability to fallback to receipt timestamp when parsing yielded incorrect value (also, no easy ability to find records with potential timestamp parsing errors)
* No way to to tell using what method the timestamp was assigned to the record 
* Lack of data to measure delay between log occurrence and log being collected

E.g. currently `filelogreceiver` sets `Timestamp` to receipt time - by default, or parsed timestamp - when timestamp operator is used. This makes it impossible to tell later if the log timestamp was parsed or not.

**Proposed solution**

This adds another field to the log record which would store the receipt timestamp (`ObservedTimestamp`). It is filled in the record when OpenTelemetry first observes the log record. 

See also the clarification about the `Timestamp` field here https://github.com/open-telemetry/opentelemetry-specification/pull/2183

The net result is that we are aiming for 2 timestamp fields like this:

- `Timestamp` - Time when the event occurred measured by the origin clock, i.e. the time at the source. This field is optional, it may be missing if the source timestamp is unknown.
- `ObservedTimestamp` - Time when the event was observed by the collection system. For events that originate in OpenTelemetry this timestamp is typically set at the generation time and is equal to `Timestamp`. For events originating externally and collected by OpenTelemetry (e.g. using Collector) this is the time when any of OpenTelemetry's code observed the event measure by the clock of the OpenTelemetry code. This field SHOULD be set once the event is observed by OpenTelemetry.

### DIFF
--- a/specification/logs/data-model.md
+++ b/specification/logs/data-model.md
@@ -178,6 +178,7 @@ Here is the list of fields in a log record:
 Field Name     |Description
 ---------------|--------------------------------------------
 Timestamp      |Time when the event occurred.
+ObservedTimestamp|Time when the event was observed.
 TraceId        |Request trace id.
 SpanId         |Request span id.
 TraceFlags     |W3C trace flag.
@@ -196,6 +197,24 @@ Type: Timestamp, uint64 nanoseconds since Unix epoch.
 
 Description: Time when the event occurred measured by the origin clock. This
 field is optional, it may be missing if the timestamp is unknown.
+
+### Field: `ObservedTimestamp`
+
+Type: Timestamp, uint64 nanoseconds since Unix epoch.
+
+Description: Time when the event was observed by the collection system. For
+events that originate in OpenTelemetry this timestamp is typically set at the
+generation time and is equal to Timestamp. For events originating externally and
+collected by OpenTelemetry (e.g. using Collector) this is the time when the
+OpenTelemetry's code observed the event measure by the clock of the
+OpenTelemetry code. This field SHOULD be set once the event is observed by
+OpenTelemetry.
+
+For converting OpenTelemetry log data to formats that support only one timestamp
+or when receiving OpenTelemetry log data by recipients that support only one
+timestamp internally the following logic is recommended:
+
+- Use `Timestamp` if it is present, otherwise use `ObservedTimestamp`.
 
 ### Trace Context Fields
 
--- a/specification/logs/data-model.md
+++ b/specification/logs/data-model.md
@@ -203,12 +203,12 @@ field is optional, it may be missing if the timestamp is unknown.
 Type: Timestamp, uint64 nanoseconds since Unix epoch.
 
 Description: Time when the event was observed by the collection system. For
-events that originate in OpenTelemetry this timestamp is typically set at the
-generation time and is equal to Timestamp. For events originating externally and
-collected by OpenTelemetry (e.g. using Collector) this is the time when the
-OpenTelemetry's code observed the event measure by the clock of the
-OpenTelemetry code. This field SHOULD be set once the event is observed by
-OpenTelemetry.
+events that originate in OpenTelemetry (e.g. using OpenTelemetry Logging SDK)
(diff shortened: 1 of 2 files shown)


tigrannajaryan:
@pmm-sumo @djaglowski please review.

specification/logs/data-model.md
```diff
@@ -197,6 +198,24 @@ Type: Timestamp, uint64 nanoseconds since Unix epoch.
 Description: Time when the event occurred measured by the origin clock. This
 field is optional, it may be missing if the timestamp is unknown.
 
+### Field: `ObservedTimestamp`
+
+Type: Timestamp, uint64 nanoseconds since Unix epoch.
+
+Description: Time when the event was observed by the collection system. For
+events that originate in OpenTelemetry this timestamp is typically set at the
```
pmm-sumo, line 206: Not sure if this is needed but maybe we should further explain what is meant here by OpenTelemetry, e.g.:

```suggestion
events that originate in OpenTelemetry (e.g. SDK) this timestamp is typically set at the
```

pmm-sumo (approved):

specification/logs/data-model.md
```diff
@@ -197,6 +198,24 @@ Type: Timestamp, uint64 nanoseconds since Unix epoch.
...
+generation time and is equal to Timestamp. For events originating externally and
+collected by OpenTelemetry (e.g. using Collector) this is the time when the
+OpenTelemetry's code observed the event measure by the clock of the
```
kumoroku, line 209: "measured"

kumoroku (approved):

specification/logs/data-model.md
```diff
@@ -197,6 +198,24 @@ Type: Timestamp, uint64 nanoseconds since Unix epoch.
...
 field is optional, it may be missing if the timestamp is unknown.
 
+### Field: `ObservedTimestamp`
+
+Type: Timestamp, uint64 nanoseconds since Unix epoch.
+
+Description: Time when the event was observed by the collection system. For
+events that originate in OpenTelemetry this timestamp is typically set at the
+generation time and is equal to Timestamp. For events originating externally and
+collected by OpenTelemetry (e.g. using Collector) this is the time when the
```
djaglowski, line 208:
```suggestion
collected by OpenTelemetry (e.g. using Collector) this is the time when
```

djaglowski (approved):

specification/logs/data-model.md
tigrannajaryan: Added "(e.g. using OpenTelemetry Logging SDK)".

specification/logs/data-model.md
tigrannajaryan: Fixed.

djaglowski (approved):

jack-berg (approved):

tigrannajaryan:

carlosalberto (approved):

zenmoto (approved):

jmacd (approved):

PR #2191, opened 2021-12-03
Prohibit usage of retired names in semantic conventions

tigrannajaryan:
This change adds a prohibition clause that requires that no old
metric or attribute name is used for a new attribute.

This is important to ensure reversibility of schema transformation
(converting from a new version to an old version of schema).

Without this restriction the following is possible:

Schema version 1. Attribute A exists.
Schema version 2. Attribute A is renamed to B. Appropriate schema file is created.
Schema version 3. Attribute A is introduced (a completely different new attribute).

Now attempting to go from Version 3 to version 1 is impossible since it requires
renaming B to A (for the change in version 2), but a different attribute A already exists.

### DIFF
--- a/specification/common/attribute-naming.md
+++ b/specification/common/attribute-naming.md
@@ -70,9 +70,8 @@ Names SHOULD follow these rules:
 
 ## Name Reuse Prohibition
 
-It is prohibited to introduce a new attribute with a name that matches a name of
-an attribute that existed in the past but was renamed (with a corresponding
-schema file).
+A new attribute MUST NOT be added with the same name as an attribute that
+existed in the past but was renamed (with a corresponding schema file).
 
 When introducing a new attribute name check all existing schema files to make
 sure the name does not appear as a key of any "rename_attributes" section (keys
--- a/specification/common/attribute-naming.md
+++ b/specification/common/attribute-naming.md
@@ -8,6 +8,7 @@
 <!-- toc -->
 
 - [Name Pluralization guidelines](#name-pluralization-guidelines)
+- [Name Reuse Prohibition](#name-reuse-prohibition)
 - [Recommendations for OpenTelemetry Authors](#recommendations-for-opentelemetry-authors)
 - [Recommendations for Application Developers](#recommendations-for-application-developers)
 - [otel.* Namespace](#otel-namespace)
@@ -66,6 +67,16 @@ Names SHOULD follow these rules:
 - When an attribute represents a measurement,
   [Metric Name Pluralization Guidelines](../metrics/semantic_conventions/README.md#pluralization)
   SHOULD be followed for the attribute name.
+
+## Name Reuse Prohibition
+
+It is prohibited to introduce a new attribute with a name that matches a name of
+an attribute that existed in the past but was renamed (with a corresponding
+schema file).
+
+When introducing a new attribute name check all existing schema files to make
+sure the name does not appear as a key of any "rename_attributes" section (keys
+denote old attribute names in rename operations).
 
 ## Recommendations for OpenTelemetry Authors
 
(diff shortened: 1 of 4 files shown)


Oberon00:
Is it expected that lossless conversion from new to old versions works? That does not seem like a feasible thing to guarantee to me.

yurishkuro:
> Is it expected that lossless conversion from new to old versions works? That does not seem like a feasible thing to guarantee to me.

I agree, at least there needs to be some grace period after which earlier versions of the schema should be considered unreachable and the old name could be used again.

tigrannajaryan:
I agree that we probably want to have a retirement policy longer term. But until we have it this is a relatively cheap way to ensure backward conversions to old versions work.

Backward conversions are very important for backends. The backend often will understand a specific telemetry version. When Otel releases a new version the backend cannot be expected to immediately be updated and start understanding that new version. This means backwards conversions are going to be fairly typical and I think if we we can ensure they are guaranteed to work we should do it. I think this particular limitation is worth it.

Besides, it is generally a good practice to follow to avoid confusion. Again, I agree that over long periods of time likely you want to have a different retirement policy, e.g. something that is unused for years can be probably reused. Once we have that policy we can rethink the reuse prohibition rule.

iNikem (approved):

specification/common/attribute-naming.md
```diff
@@ -67,6 +68,16 @@ Names SHOULD follow these rules:
   [Metric Name Pluralization Guidelines](../metrics/semantic_conventions/README.md#pluralization)
   SHOULD be followed for the attribute name.
 
+## Name Reuse Prohibition
+
+It is prohibited to introduce a new attribute with a name that matches a name of
```
SergeyKanzhelev, line 73: Can we use MUST NOT or SHOULD NOT language here instead of prohibited.

MrAlias:
```suggestion
A new attribute MUST NOT be added with the same name as
```

Oberon00 (approved):
@arminru @dyladan, others at Dynatrace and me discussed this PR shortly. We do not really agree with the motivation, or rather, we do not have that use case yet. But we think that adding this restriction does no harm and is beneficial also for humans to avoid confusing scenarios.

jmacd (approved):

specification/common/attribute-naming.md
tigrannajaryan: Fixed.

specification/metrics/semantic_conventions/README.md
```diff
@@ -66,6 +67,16 @@ and confusion for end users. (For example, prefer `process.runtime.java.gc*` ove
 `process.runtime.gc.*`.) Measures of many operating system metrics are similarly
 ambiguous.
 
+### Name Reuse Prohibition
+
+It is prohibited to introduce a new metric with a name that matches a name of
+a metric that existed in the past but was renamed (with a corresponding
+schema file).
```
MrAlias, line 74: Can we copy the same wording from specification/common/attribute-naming.md here and in specification/trace/semantic_conventions/README.md?

tigrannajaryan: Copied.

SergeyKanzhelev (approved):

MrAlias (approved):

specification/common/attribute-naming.md
```diff
@@ -8,6 +8,7 @@
 <!-- toc -->
 
 - [Name Pluralization guidelines](#name-pluralization-guidelines)
+- [Name Reuse Prohibition](#name-reuse-prohibition)
```
arminru, line 11: Changelog please :-)

arminru (approved):

specification/common/attribute-naming.md
tigrannajaryan: Added.

PR #2194, opened 2021-12-04
Simplify Baggage handling in the OpenTracing Shim layer.

### LINKED ISSUE #2137: OpenTracing layer support for Baggage

carlosalberto:
Baggage is NOT directly associated with an OpenTelemetry `Span` directly anymore, but with the **OpenTracing Shim layer**, which greatly simplifies its handling (no more global dictionary or extra field in Spans).

This comes with the recommendation of NOT consuming the OpenTelemetry API and the OpenTracing Shim the codebase, as the `Span`'s associated `Baggage` may be lost during in-process or inter-process propagation.

I have a [prototype](https://github.com/open-telemetry/opentelemetry-java/compare/main...carlosalberto:otshim-spanshim-lock?expand=1) for this in Java.

I'd added thread-safety requirements for `Baggage` consumption, even if it doesn't exist in the original OpenTracing Specification, as those calls are thread-safe anyway in most of (all?) OpenTracing compliant tracers.

Fixes #2137

### DIFF
--- a/specification/compatibility/opentracing.md
+++ b/specification/compatibility/opentracing.md
@@ -10,7 +10,7 @@
 * [Tracer Shim](#tracer-shim)
   * [Inject](#inject)
   * [Extract](#extract)
-* [OpenTelemetry Span and SpanContext Shim relationship](#opentelemetry-span-and-spancontext-shim-relationship)
+* [Span Shim and SpanContext Shim relationship](#span-shim-and-spancontext-shim-relationship)
 * [Span Shim](#span-shim)
   * [Get Context](#get-context)
   * [Get Baggage Item](#get-baggage-item)
@@ -44,13 +44,13 @@ upstream OpenTelemetry API.
 This functionality MUST be defined in its own OpenTracing Shim Layer, not in the
 OpenTracing nor the OpenTelemetry API or SDK.
 
-The OpenTracing Shim and the OpenTelemetry API/SDK are expected to be consumed
-simultaneously in a running service, in order to ease migration from the former
-to the latter.
-
 Semantic convention mapping SHOULD NOT be performed, with the
 exception of error mapping, as described in the [Set Tag](#set-tag) and
 [Log](#log) sections.
+
+It is not recommended the OpenTracing Shim and the OpenTelemetry API are
+consumed in the same codebase, as `Baggage` may not be properly propagated. See
+[Span Shim and SpanContext Shim relationship](#span-shim-and-spancontext-shim-relationship).
 
 ## Create an OpenTracing Tracer Shim
 
@@ -93,7 +93,7 @@ Parameters:
 - A `Format` descriptor.
 - A carrier.
 
-Inject the underlying OpenTelemetry `Span` and `Bagagge` using either the explicitly
+Inject the underlying OpenTelemetry `Span` and `Baggage` using either the explicitly
 registered or the global OpenTelemetry `Propagator`s, as configured at construction time.
 
 - `TextMap` and `HttpHeaders` formats MUST use their explicitly specified `TextMapPropagator`,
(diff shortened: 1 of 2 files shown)


carlosalberto:
Ping @yurishkuro

specification/compatibility/opentracing.md
```diff
@@ -44,14 +44,14 @@ upstream OpenTelemetry API.
...
 
-The OpenTracing Shim and the OpenTelemetry API/SDK are expected to be consumed
-simultaneously in a running service, in order to ease migration from the former
-to the latter.
-
 Semantic convention mapping SHOULD NOT be performed, with the
 exception of error mapping, as described in the [Set Tag](#set-tag) and
 [Log](#log) sections.
 
+It is not recommended the OpenTracing Shim and the OpenTelemetry API are
```
yurishkuro, line 51: Isn't this too strong? I think this recommendation only stands when conditioned on "if you want to pass baggage through the application, then don't mix shim and core API".

yurishkuro (approved):

specification/compatibility/opentracing.md
carlosalberto: Makes sense, will relax it a little bit.

carlosalberto:
@yurishkuro Updated to discourage users from using the OT Shim along the OTel API _only if_ baggage is used.

yurishkuro (approved):

jmacd (approved):
I believe the OTel-Go `bridge/opentracing` goes extensively out of it's way to make the behavior being specified here work, also (a) I'm not 100% certain of this statement because the code is difficult to follow, and (b) if I'm proven correct then we ought to simplify the code based on this spec change. I'll file an issue if you agree @carlosalberto.

PR #2201, opened 2021-12-06
Clarify confusing aspects of exemplar specification.

### LINKED ISSUE #2155: The concept of exemplars is very vaguely defined in the spec
From https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#exemplars

> An exemplar is a recorded value that associates OpenTelemetry context to a metric event within a Metric. 

"recorded value" of what? A time series is already a collection of recorded values, how is exemplar related to that?

> One use case is to allow users to link Trace signals w/ Metrics.

What does that mean? Why does one need that? What problem is it trying to solve? Is it the only way to solve that problem?

> The recorded value (value)

Recorded values are being captured by the metric / time series. What is the other type of recorded value this refers to?

From https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#exemplar

> An Exemplar is a recorded Measurement that exposes the following pieces of information:

What is the relationship between this Measurement and the metric itself? When an end-user of OTel API is calling an instrument to record a metric's value, are they also providing some additional measurement that is the exemplar? The API spec never mentions anything about exemplars.

> A Metric SDK MUST provide a mechanism to sample Exemplars from measurements.

The line above just said exemplar IS a measurement. Why is it needed to be sampled? Assuming this is explained, what is this "mechanism to sample", how/where is it defined?

### LINKED ISSUE #2193: Clarify collect + reset on exemplar reservoir
FYI - This is a good callout.  There were a few clarifying things in the spec that got dropped, but we can add:

1. The collect+reset here is to account for DELTA vs. CUMULATIVE aggregation windows.   Exemplars collection should mirror aggregation.  In Java implementation, all synchronous instruments *always* report DELTA accumulations, hence collect+reset by default.
2. Exemplar preservation should be best-effort.   It might be worth calling out that's it's explicitly ok to drop exemplar memory storage under pressure.

IF you open a ticket for spec clarification I'll tackle it.

_Originally posted by @jsuereth in https://github.com/open-telemetry/opentelemetry-js/pull/2641#discussion_r760559314_

jsuereth:
Fixes #2155
Fixes #2193

- improves description of Exemplars, including use-case.
- Clarify `measurement` as coming from API, and record time.
- Clarify `reset` behavior for `num_measurements` counters.

### DIFF
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -439,15 +439,36 @@ An [Exemplar](./datamodel.md#exemplars) is a recorded
 [Measurement](./api.md#measurement) that exposes the following pieces of
 information:
 
-- The `value` that was recorded at API call time.
-- The `time` the `Measurement` was sent to the API.
+- The `value` of the `Measurement` that was recorded by the API call.
+- The `time` the API call was made to record a `Measurement`.
 - The set of [Attributes](../common/common.md#attributes) associated with the
   `Measurement` not already included in a metric data point.
-  **Note: Views may aggregate away `Attributes`, which is where this is used.**
 - The associated [trace id and span
   id](../trace/api.md#retrieving-the-traceid-and-spanid) of the active [Span
   within Context](../trace/api.md#determining-the-parent-span-from-a-context) of
   the `Measurement` at API call time.
+
+For example, if a user has configured a `View` to preserve the attributes: `X`
+and `Y`, but the user records a measurement as follows:
+
+```javascript
+const span = tracer.startSpan('makeRequest');
+api.context.with(api.trace.setSpan(api.context.active(), span), () => {
+  // Record a measurement.
+  cache_miss_counter.add(1, {"X": "x-value", "Y": "y-value", "Z": "z-value"});
+  ...
+  span.end();
+})
+```
+
+Then an examplar would consist of:
+
+- The `value` of 1.
+- The `time` when the `add` method was called
+- The `Attributes` of `{"Z": "z-value"}`, as these are not preserved in the
+  resulting metric point.
+- The trace/span id for the `makeRequest` span.
+
 
 A Metric SDK MUST provide a mechanism to sample `Exemplar`s from measurements
 via the `ExemplarFilter` and `ExemplarReservoir` hooks.
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -461,13 +461,16 @@ api.context.with(api.trace.setSpan(api.context.active(), span), () => {
 })
 ```
 
-Then an examplar would consist of:
+Then an examplar output in OTLP would consist of:
 
 - The `value` of 1.
 - The `time` when the `add` method was called
 - The `Attributes` of `{"Z": "z-value"}`, as these are not preserved in the
   resulting metric point.
 - The trace/span id for the `makeRequest` span.
+
+While the metric data point for the counter would carry the attributes `X` and
+`Y`.
 
 A Metric SDK MUST provide a mechanism to sample `Exemplar`s from measurements
 via the `ExemplarFilter` and `ExemplarReservoir` hooks.
@@ -527,8 +530,10 @@ MAY further sample beyond the `ExemplarFilter`.
 
 The "collect" method MUST return accumulated `Exemplar`s. Exemplars are expected
 to abide by the `AggregationTemporality` of any metric point they are recorded
-with. SDKs are free to decide whether "collect" should also reset internal
-storage for delta aggregation collection, or use a more optimal implementation.
+with. In other words, Exemplars reported against a metric data point SHOULD have
+occured within the start/stop timestamps of that point.  SDKs are free to decide
+whether "collect" should also reset internal storage for delta temporal
+aggregation collection, or use a more optimal implementation.
 
 `Exemplar`s MUST retain any attributes available in the measurement that
 are not preserved by aggregation or view configuration. Specifically, at a
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -531,8 +531,8 @@ MAY further sample beyond the `ExemplarFilter`.
 The "collect" method MUST return accumulated `Exemplar`s. Exemplars are expected
 to abide by the `AggregationTemporality` of any metric point they are recorded
 with. In other words, Exemplars reported against a metric data point SHOULD have
-occured within the start/stop timestamps of that point.  SDKs are free to decide
-whether "collect" should also reset internal storage for delta temporal
+occurred within the start/stop timestamps of that point.  SDKs are free to
+decide whether "collect" should also reset internal storage for delta temporal
 aggregation collection, or use a more optimal implementation.
 
 `Exemplar`s MUST retain any attributes available in the measurement that
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -469,7 +469,6 @@ Then an examplar would consist of:
   resulting metric point.
 - The trace/span id for the `makeRequest` span.
 
-
 A Metric SDK MUST provide a mechanism to sample `Exemplar`s from measurements
 via the `ExemplarFilter` and `ExemplarReservoir` hooks.
 
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -530,7 +530,7 @@ to abide by the `AggregationTemporality` of any metric point they are recorded
 with. SDKs are free to decide whether "collect" should also reset internal
 storage for delta aggregation collection, or use a more optimal implementation.
 
-`Exemplar`s MUST retain the any attributes available in the measurement that
+`Exemplar`s MUST retain any attributes available in the measurement that
 are not preserved by aggregation or view configuration. Specifically, at a
 minimum, joining together attributes on an `Exemplar` with those available
 on its associated metric data point should result in the full set of attributes
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -429,35 +429,43 @@ series and it requires further analysis.
 
 ## Exemplar
 
+Exemplars are example data points for aggregated data. They provide specific
+context to otherwise general aggregations. Exemplars allow correlation between
+aggregated metric data and the original API calls where measurements are
+recorded. Exemplars work for trace-metric correlation across any metric, not
+just those that can also be derived from `Span`s.
+
 An [Exemplar](./datamodel.md#exemplars) is a recorded
 [Measurement](./api.md#measurement) that exposes the following pieces of
 information:
 
-- The `value` that was recorded.
-- The `time` the `Measurement` was seen.
+- The `value` that was recorded at API call time.
+- The `time` the `Measurement` was sent to the API.
 - The set of [Attributes](../common/common.md#attributes) associated with the
   `Measurement` not already included in a metric data point.
+  **Note: Views may aggregate away `Attributes`, which is where this is used.**
 - The associated [trace id and span
   id](../trace/api.md#retrieving-the-traceid-and-spanid) of the active [Span
   within Context](../trace/api.md#determining-the-parent-span-from-a-context) of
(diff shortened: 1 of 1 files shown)


specification/metrics/sdk.md
```diff
@@ -429,35 +429,43 @@ series and it requires further analysis.
 
 ## Exemplar
 
+Exemplars are example data points for aggregated data. They provide specific
```
yurishkuro, line 432: Q: are they always full-fidelity data points, i.e. representing all dimensions of the original measurement?

specification/metrics/sdk.md
```diff
@@ -429,35 +429,43 @@ series and it requires further analysis.
...
 [Measurement](./api.md#measurement) that exposes the following pieces of
 information:
 
-- The `value` that was recorded.
-- The `time` the `Measurement` was seen.
+- The `value` that was recorded at API call time.
+- The `time` the `Measurement` was sent to the API.
 - The set of [Attributes](../common/common.md#attributes) associated with the
   `Measurement` not already included in a metric data point.
+  **Note: Views may aggregate away `Attributes`, which is where this is used.**
```
yurishkuro, line 446: Not clear - can views trim down attributes of the exemplars themselves? 

Rather than have a Note here I would move this example into the intro paragraph above.

specification/metrics/sdk.md
```diff
@@ -429,35 +429,43 @@ series and it requires further analysis.
 
 ## Exemplar
 
+Exemplars are example data points for aggregated data. They provide specific
+context to otherwise general aggregations. Exemplars allow correlation between
+aggregated metric data and the original API calls where measurements are
+recorded. Exemplars work for trace-metric correlation across any metric, not
+just those that can also be derived from `Span`s.
+
```
yurishkuro, line 437: I think it's useful to provide a link to the OTEP for further background / decisions.

specification/metrics/sdk.md
jack-berg: The java implementation of exemplar only includes attributes that were filtered out by the aggregator. So the combination of the metric attributes plus the exemplar attributes provides the full set originally collected. 

I believe this behavior is specified a bit [further down](https://github.com/open-telemetry/opentelemetry-specification/pull/2201/files#diff-32d5f289beed96900fa1e6e69a3ee9cc0b6503ce8883672011c44bc772c0fa2aR513-R514).

specification/metrics/sdk.md
```diff
@@ -429,35 +429,43 @@ series and it requires further analysis.
 
 ## Exemplar
 
+Exemplars are example data points for aggregated data. They provide specific
+context to otherwise general aggregations. Exemplars allow correlation between
+aggregated metric data and the original API calls where measurements are
+recorded. Exemplars work for trace-metric correlation across any metric, not
+just those that can also be derived from `Span`s.
```
jack-berg, line 436: > across any metric, not just those that can also be derived from `Span`s.

What exactly do you mean by this?

specification/metrics/sdk.md
```diff
@@ -429,35 +429,43 @@ series and it requires further analysis.
...
+
 An [Exemplar](./datamodel.md#exemplars) is a recorded
 [Measurement](./api.md#measurement) that exposes the following pieces of
 information:
 
-- The `value` that was recorded.
-- The `time` the `Measurement` was seen.
+- The `value` that was recorded at API call time.
+- The `time` the `Measurement` was sent to the API.
```
jack-berg, line 443: "call time" and "sent to the API" dont appear to be consistent with other verbiage. A possible alternative: 
```suggestion
- The `value` of the `Measurement` that was recorded by API call.
- The `time` the API call was made to record the `Measurement`.
```

specification/metrics/sdk.md
jsuereth: The simple answer is `yes`.   The longer answer (for OTLP) is we only store 'missing' dimensions from the aggregated metric data point we're associated with.

specification/metrics/sdk.md
jsuereth: Good call out, will fix.

specification/metrics/sdk.md
jsuereth: We don't do this in any of the other parts of the specification.   Is that a change we're making in expecations?

specification/metrics/sdk.md
jsuereth: @yurishkuro has a system where you can automatically attach Span-derived metric exemplar points.   Specifically, you could derive latency metrics from spans and if you had a way to correlate to a specific metric, you could generate exemplars directly from the span itself.   The OTEL Exemplar does this, but also in scenarios where your metric does NOT correlated with a Span attribtue (e.g. "memory allocations", if you can record in context)

specification/metrics/sdk.md
jsuereth: I like your second option, will clarify.

specification/metrics/sdk.md
jsuereth: Added an example, PTAL

specification/metrics/sdk.md
```diff
@@ -429,35 +429,63 @@ series and it requires further analysis.
...
+  span.end();
+})
+```
+
+Then an examplar would consist of:
 
-A Metric SDK MUST provide a mechanism to sample `Exemplar`s from measurements.
+- The `value` of 1.
+- The `time` when the `add` method was called
+- The `Attributes` of `{"Z": "z-value"}`, as these are not preserved in the
```
yurishkuro, line 468: I find this example misleading. It focuses on the storage aspects of the exemplar (which can change throughout aggregation pipeline), instead of giving a logical representation of the exemplar having all 3 attributes.

specification/metrics/sdk.md
```diff
@@ -497,7 +525,10 @@ span context and baggage can be inspected at this point.
 The "offer" method does not need to store all measurements it is given and
 MAY further sample beyond the `ExemplarFilter`.
 
-The "collect" method MUST return accumulated `Exemplar`s.
+The "collect" method MUST return accumulated `Exemplar`s. Exemplars are expected
+to abide by the `AggregationTemporality` of any metric point they are recorded
+with. SDKs are free to decide whether "collect" should also reset internal
+storage for delta aggregation collection, or use a more optimal implementation.
```
aabmass, line 531: It's not clear for me what it means for the exemplar to abide by the temporality of the metric point. Do you mean that CUMULATIVE shouldn't reset sampling state but DELTA should reset it?

I would think you'd want to reset regardless of temporality (even for CUMULATIVE) so that you send fresh exemplars instead of repeating the same exemplar as probability drops with the `SimpleFixedSizeExemplarReservoir`. I suppose the benefit of not resetting a CUMULATIVE is to get a true random sample over the process lifetime?

Would it be easier to have the interface provide both `collect()` and `collectAndReset()` and let the SDK determine when to reset?

specification/metrics/sdk.md
```diff
@@ -497,7 +525,10 @@ span context and baggage can be inspected at this point.
...
 
 `Exemplar`s MUST retain the any attributes available in the measurement that
```
jamesmoessis, line 533:
```suggestion
`Exemplar`s MUST retain any attributes available in the measurement that
```

jamesmoessis (approved):

specification/metrics/sdk.md
jmacd: You're suggesting a more detailed answer -- here's what I suggest:

Whether or not all dimensions are included in the exemplar, the exemplar fundamentally represent a raw, individual measurement whereas both Sum and Histogram (i.e., all synchronous) OTLP data points have effectively no way to record raw measurements.

specification/metrics/sdk.md
jmacd: It's very old by now, but there is one. 
https://github.com/open-telemetry/oteps/blob/main/text/metrics/0113-exemplars.md

specification/metrics/sdk.md
jmacd: Perhaps this can be clarified. "resulting metric point" below refers to one of e.g. `NumberDataPoint`, `HistogramDataPoint`, which are concrete members of the pipeline, so I think it makes sense to focus on how the protocol looks. The exemplar's attributes are nested in one of the data points, so there is no loss of information.

These are individual measurements, that is all, and the attributes (whether included or not included) are copied from actual synchronous API events. If more text is needed to explain how to merge exemplars, for example, that belongs in `datamodel.md` I think.

It's possible to add an adjusted count concept to exemplars; at that point I would call them raw samples, not exemplars. I've put a few details into https://github.com/lightstep/varopt/blob/main/README.md for how this can be done so that adjusted counts are unbiased--the result is that the samples can be used to reconstruct the aggregated metrics they are included with in a statistically meaningful way. This is a property that can be preserved through merging OTLP data points, but if we were to go this route, it would be more than exemplars.

specification/metrics/sdk.md
jmacd: > Do you mean that CUMULATIVE shouldn't reset sampling state but DELTA should reset it?

I'd like more information from Prometheus on how exemplars are being handled. I do think the `num_measurements_seen` variable should reset so that points within an interval have equal probability, but if no new exemplars arrive during  period, then the Prometheus exporter should continue to see the last exemplar.

specification/metrics/sdk.md
```diff
@@ -429,35 +429,63 @@ series and it requires further analysis.
...
+A Metric SDK MUST provide a mechanism to sample `Exemplar`s from measurements
+via the `ExemplarFilter` and `ExemplarReservoir` hooks.
+
+A Metric SDK MUST allow `Exemplar` sampling to be disabled. In this instance
 the SDK SHOULD not have overhead related to exemplar sampling.
 
 A Metric SDK MUST sample `Exemplar`s only from measurements within the context
 of a sampled trace BY DEFAULT.
 
-A Metric SDK MUST allow exemplar sampling to leverage the configuration of a
```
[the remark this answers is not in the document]
jsuereth: Not sure why it shows the line as deleted, but it's right below this one.

specification/metrics/sdk.md
jsuereth: Exemplars in prometheus (did the investigation prior to writing any of this) abide by cumulative rules here.  THey keep last-seen-per-bucket always and you cannot collect exemplars for anything but histograms (even though they're attached to samples, if you use this it breaks for any metric type but histograms).

Specifically, the prometheus client library never clears the exemplars AFAICT, which means we need to decide what to do for DELTAs (as usual) but if we align with Prometheus, it means respecting the start/stop time window.

specification/metrics/sdk.md
```diff
@@ -531,6 +562,9 @@ algorithm](https://en.wikipedia.org/wiki/Reservoir_sampling)
   end
   ```
 
+Additionally, the `num_measurements_seen` count SHOULD be reset at every
+collection cycle.
```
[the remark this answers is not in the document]
jsuereth: That's getting into weeds of impelmentation a bit.  From a specification standpoint this is about ensuring we get FRESH exemplars in a CUMULATIVE case.  If you're optimising your exemplar based on known CUMULATIVE reader-exporter pipeline, then you don't have to clean everything, and I'm trying to avoid forcing that in the specification.

specification/metrics/sdk.md
jsuereth: I'll clarify that this is talking the *output* in OTLP, vs. the logical representation.  This is *not* intended to represent internal storage, in fact we want to allow SDKs to store full attributes in exemplars and filter later for OTLP if that's faster.

specification/metrics/sdk.md
jsuereth: Updated the verbage here, PTAL

specification/metrics/sdk.md
jsuereth: I updated the wording, PTAL.

yurishkuro (approved):

carlosalberto (approved):

PR #2204, opened 2021-12-08
Clarify the span Temporal relationship diagram

adrianjgeorge:
The temporal span relationship diagram on the overview page can be a little misleading at first glance. Due to Span C being both shorter and below Span B in the diagram a first reading suggests that it is a sub span of B. When compared to the tree view you can sort out that it's not meant to be, but it adds a bit of cognitive friction that's unnecessary. To more quickly communicate intent, I've shorten Spans B and D, and pushed Span C out past B.


jmacd (approved):

specification/overview.md
```diff
@@ -140,9 +140,9 @@ Temporal relationships between Spans in a single Trace
 ––|–––––––|–––––––|–––––––|–––––––|–––––––|–––––––|–––––––|–> time
 
  [Span A···················································]
-   [Span B··············································]
-      [Span D··········································]
-    [Span C········································]
+   [Span B··········································]
+      [Span D······································]
+    [Span C············································]
```
arminru, line 145:
```suggestion
    [Span C····················································]
```
While at it you could also extend one span's duration to indicate that a child span can outlive its parent span.

arminru (approved):
Thank you!

tigrannajaryan (approved):

Issue #1874, opened 2021-08-18
Aggregating async Counter with Views dropping some attribute


cijothomas:
Lets say I have a AsyncCounter (requestCountAsync).
At time T1, its callback reported the following measurements
- measurements (100, success=true, verb=Get)	
- measurements (5, success=false, verb=Get)
we export the following:
{{requestCountAsync{Verb="Get", success=true,  "Sum:100", Temporality=Cumulative, {T0}, T1}}}
{{requestCountAsync{Verb="Get", success=false, "Sum:5", Temporality=Cumulative, {T0}, T1}}}

At T1+1 
- measurements (120, success=true, verb=Get)	
- measurements (7, success=false, verb=Get)
we export the following:
{{requestCountAsync{Verb="Get", success=true,  "Sum:120", Temporality=Cumulative, {T0}, T1+1}}}
{{requestCountAsync{Verb="Get", success=false, "Sum:7", Temporality=Cumulative, {T0}, T1+1}}}

At T1+2
- measurements (150, success=true, verb=Get)	
- measurements (10, success=false, verb=Get)
we export the following:
{{requestCountAsync{Verb="Get", success=true,  "Sum:150", Temporality=Cumulative, {T0}, T1+2}}}
{{requestCountAsync{Verb="Get", success=false, "Sum:10", Temporality=Cumulative, {T0}, T1+2}}}

and so on.

Now I configured my SDK to have a View, which says "only pick "Verb" as an attribute, and drop everything else.
Repeating the same application:
At time T1, its callback reported the following measurements
- measurements (100, success=true, verb=Get)	
- measurements (5, success=false, verb=Get)

What should we export?
1. Report last value: 
    {{requestCountAsync{Verb="Get", "Sum:5", Temporality=Cumulative, {T0}, T1}}}

OR 

2. Do "some sort of spatial aggregation" or "some merge", and report
    {{requestCountAsync{Verb="Get", "Sum:105", Temporality=Cumulative, {T0}, T1}}}

(Note: This is the "truth", as my web server has 105 requests total until now)

OR

This is undefined and upto languages..

reyang:
Discussed during the 8/19 Metrics SIG meeting, the answer is 2.

reyang:
Victor will follow up by putting a concrete example, if it turned out that we need to clarify the spec, we will turn this into a spec issue.

victlu:
## Setup

Using an Asynchronous Counter instrument.

A View configured with:
[...]

PR #2208, opened 2021-12-08
Metrics supplemental guidelines for async views

### LINKED ISSUE #2187: Clarify SumAggregation on asynchronous counter

In the SDK spec on DefaultAggregation, it suggests that for AsynchronousCounter a default SumAggregation is used for aggregation.

> Instrument Kind | Selected Aggregation
> -- | --
> Counter | Sum Aggregation
> Asynchronous Counter | Sum Aggregation
> UpDownCounter | Sum Aggregation
> Asynchrounous UpDownCounter | Sum Aggregation
> Asynchronous Gauge | Last Value Aggregation
> Histogram | Histogram Aggregation
>
> https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#default-aggregation

When implementing the SumAggregation for the asynchronous counter, I found that in the SumAggregation:

> SumAggregation informs the SDK to collect:
> The arithmetic sum of `Measurement` values.
> https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#sum-aggregation

Whilst to the asynchronous counter, the `Measurement` values mean:

> The callback function is responsible for reporting the Measurements.
> Note: Unlike Counter.Add() which takes the increment/delta value, the callback function reports the absolute value of the counter. 
> https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#asynchronous-counter

That's been said, the `Measurement` values in asynchronous counters are absolute values and not increments. To SumAggregation, the arithmetic sum of `Measurement` values of asynchronous counters doesn't seem to be meaningful.

This issue also applies to the Asynchronous UpDownCounter.

jmacd:
Fixes #2187

## Changes

Clarifies the available strategies for calculating asynchronous Views and what it means for the receiver, especially when streams "die" and restart, requiring reset detection.

Related issues #1874


jsuereth:
This looks great to me.  I think it nails options available to SDKs and provides a simple, allowable implementation, demonstrating why it works.

Should we recommend ONE option to SDKs first though, while allowing the others?

specification/metrics/supplementary-guidelines.md
```diff
@@ -302,11 +319,45 @@ If we export the metrics using **Cumulative Temporality**:
   * dimensions: {pid = `1001`, tid = `2`}, sum: `53`
   * dimensions: {pid = `1001`, tid = `3`}, sum: `5`
 
-It is quite straightforward - we just take the data being reported from the
-asynchronous instruments and send them. We might want to consider if [Resets and
-Gaps](./datamodel.md#resets-and-gaps) should be used to denote the end of a
-metric stream - e.g. thread 1 died, the thread ID might be reused by the
-operating system, and we probably don't want to confuse the metrics backend.
+The behavior in the first fuor periods is quite straightforward - we
```
legendecas, line 322:
```suggestion
The behavior in the first four periods is quite straightforward - we
```

specification/metrics/supplementary-guidelines.md
```diff
@@ -302,11 +319,45 @@ If we export the metrics using **Cumulative Temporality**:
...
+   stream and permit forgetting streams that have stopped reporting.
+
+It's OK to ignore the options to use per-stream start timestamps and
+staleness markers. The first course of action above requires no
+additional memory or code to achieve and is correct in terms of the
+data model.
+
+##### Asynchronous example: Delta temporality
 
 If we export the metrics using **Delta Temporality**:
```
legendecas, line 362: What if a dimension is not reported in one timespan, but recovers later? Should it be considered as reset, or should we keep all dimensions for Delta Temporality for an indefinite time to calculate delta values?

jsuereth: for DELTA temporality, it's ok to forget.  In practice timestamps and expected reporting intervals are used to identify drops and missing points.

CUMULATIVEs are where it's expected that you can miss a data point, but still "assume" the value of a missing point, so drops/gaps may need some special treatment understanding.

jmacd: @legendecas I tried to answer this question in the section below, so probably some reordering would help. It feels like too many words to spell out every related consequence of these decisions, but I'm happy to keep adding them :)

Your "SHOULD" question is related to @jsuereth's question above. Do we have a recommendation? My recommended solution would prioritize correctness and minimize gaps and/or reset detection, but up until now I wasn't sure this would be a profitable discussion. Here's what I would do: maintain a map of recently-used asynchronous (instrument, attributes) streams with per-stream start times.  When a stream does not report after previously reporting, staleness markers should be issued for up to 10 minutes (pull) or 1 export interval (push). After staleness has been reported, the map can forget the stream, as it will use a new start time if/when recreated. Since the state to do so is available, the SDK should then also provide duplicate-measurement detection.

Paired with Views, my preference is to disable garbage collection, which allows the exported cumulative sum to be always correct without gaps. If we give up garbage collection, then the aggregate stream's start times should be reset whenever any components drop out of the sum, which would be bad if it happened frequently. I would give up on the garbage collection but have a safety relief in place: if the entire state of any one or more asynchronous metrics becomes too much memory, it's easy to reset all streams.

The data model backs any of these decisions up. Receivers ought to be aware of monotonicity and thus detect resets, just as they would in a Prometheus system. If the receiver is additionally aware of stream start time and the SDK uses them, that's better (in terms of correct reset detection, duplicate recognition). If the receiver is additionally aware of staleness markers, that's best.

specification/metrics/supplementary-guidelines.md
```diff
@@ -302,11 +319,45 @@ If we export the metrics using **Cumulative Temporality**:
...
+   zero sum.  Receivers with reset detection are able to calculate a
+   correct rate (except for frequent restarts relative to the
+   collection interval), however the precise time of a reset will be
+   unknown.
+2. If the SDK maintains per-stream start times, it signals to the
+   receiver precisely when a stream started, making the first
+   observation in a stream more useful for diagnostics.  Receivers can
+   perform overlap detection or duplicate suppression and do not
+   require reset detection, in this case.
+3. Independent of above treaments, the SDK can add a staleness marker
```
pirgeo, line 348:
```suggestion
3. Independent of above treatments, the SDK can add a staleness marker
```

pirgeo (commented):
Thanks for the clarification! I wonder: for the Synchronous example, there are the Delta, then the Cumulative aggregation temporality sections, but for Asynchronous there is the Cumulative section first, then the Delta section. In the attribute removal section right after that, there are references to Cumulative temporality again. Do you think it would make sense to swap the Delta and Cumulative sections for the Asynchronous example?

jsuereth (approved):

CHANGELOG.md
```diff
@@ -27,6 +27,9 @@ release.
   ([#2154](https://github.com/open-telemetry/opentelemetry-specification/pull/2154))
 - Mark In-memory, OTLP and Stdout exporter specs as Stable.
   ([#2175](https://github.com/open-telemetry/opentelemetry-specification/pull/2175))
+- Add to the supplemental guidelines for metric SDK authors text about implementing
+  attribute-removal Views for asynchronous instruments.
+  ([TODO]())
```
MrAlias, line 32:
```suggestion
  ([#2208](https://github.com/open-telemetry/opentelemetry-specification/pull/2208))
```

MrAlias (approved):

jmacd:
@pirgeo to rationalize the differently-ordered temporality sections for sync vs async, 8556ff6.

legendecas (approved):
👍

carlosalberto:
@jmacd Should we reopen this PR? It had a few approvals already, etc ;)

jmacd:
Given that these are guidelines--not specs--and have several approvals, I feel this can merge. @reyang PTAL.

reyang (approved):
LGTM.

PR #2210, opened 2021-12-09
Clarify integer count instrument units

pichlermarc:
This PR clarifies the units that should be used for instruments that measure an integer count. 

The spec mentioned that

> Instruments that measure an integer count of something SHOULD use the default unit 1 (the unity) and annotations with curly braces to give additional meaning. For example {packets}, {errors}, {faults}, etc.

Which, due to the use of "and" in the language, could be interpreted that units should be written as `1{packets}`, `1{errors}` and `1{faults}`. According to [UCUM](https://ucum.org/ucum.html), however, `{packets}` semantically is equivalent to `1{packets}`.

To prevent this interpretation, this PR clarifies that the leading `1` should be omitted.

### DIFF
--- a/specification/metrics/semantic_conventions/README.md
+++ b/specification/metrics/semantic_conventions/README.md
@@ -151,5 +151,5 @@ more clarification in
 total) are dimensionless and SHOULD use the default unit `1` (the unity).
 - Instruments that measure an integer count of something SHOULD only use
 [annotations](https://ucum.org/ucum.html#para-curly) with curly braces to
-give additional meaning *without* the leading default unit (`1`). For example
-`{packets}`, `{errors}`, `{faults}`, etc.
+give additional meaning *without* the leading default unit (`1`). For example,
+use `{packets}`, `{errors}`, `{faults}`, etc.
--- a/specification/metrics/semantic_conventions/README.md
+++ b/specification/metrics/semantic_conventions/README.md
@@ -149,5 +149,6 @@ more clarification in
 - Instruments for **utilization** metrics (that measure the fraction out of a
 total) are dimensionless and SHOULD use the default unit `1` (the unity).
 - Instruments that measure an integer count of something SHOULD only use
-annotations with curly braces to give additional meaning without the leading
-default unit (`1`). For example `{packets}`, `{errors}`, `{faults}`, etc.
+[annotations](https://ucum.org/ucum.html#para-curly) with curly braces to
+give additional meaning without the leading default unit (`1`). For example
+`{packets}`, `{errors}`, `{faults}`, etc.
--- a/specification/metrics/semantic_conventions/README.md
+++ b/specification/metrics/semantic_conventions/README.md
@@ -148,8 +148,6 @@ more clarification in
 
 - Instruments for **utilization** metrics (that measure the fraction out of a
 total) are dimensionless and SHOULD use the default unit `1` (the unity).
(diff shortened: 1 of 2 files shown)


jmacd (approved):

MrAlias (approved):

specification/metrics/semantic_conventions/README.md
```diff
@@ -148,8 +148,7 @@ more clarification in
 
 - Instruments for **utilization** metrics (that measure the fraction out of a
 total) are dimensionless and SHOULD use the default unit `1` (the unity).
-- Instruments that measure an integer count of something SHOULD use the
-default unit `1` (the unity) and
+- Instruments that measure an integer count of something SHOULD only use
 [annotations](https://ucum.org/ucum.html#para-curly) with curly braces to
-give additional meaning. For example `{packets}`, `{errors}`, `{faults}`,
-etc.
+give additional meaning without the leading default unit (`1`). For example
```
arminru, line 153:
```suggestion
give additional meaning *without* the leading default unit (`1`). For example
```

specification/metrics/semantic_conventions/README.md
```diff
@@ -148,8 +148,7 @@ more clarification in
 
 - Instruments for **utilization** metrics (that measure the fraction out of a
```
arminru, line 149: CHANGELOG please :)

arminru (approved):

specification/metrics/semantic_conventions/README.md
arminru: Also I think the sentence above would read more naturally if you'd change it to the following while you're at it:

> Units should follow the [Unified Code for Units of Measure (UCUM)](http://unitsofmeasure.org/ucum.html)
(need more clarification in
...

jsuereth (approved):

specification/metrics/semantic_conventions/README.md
pichlermarc: Added in 1fee7b8.

specification/metrics/semantic_conventions/README.md
pichlermarc: Both suggestions added in 1fee7b8 :)

specification/metrics/semantic_conventions/README.md
```diff
@@ -142,14 +142,14 @@ instrument creation, but can be added if there is ambiguity.
...
 total) are dimensionless and SHOULD use the default unit `1` (the unity).
-- Instruments that measure an integer count of something SHOULD use the
-default unit `1` (the unity) and
+- Instruments that measure an integer count of something SHOULD only use
 [annotations](https://ucum.org/ucum.html#para-curly) with curly braces to
-give additional meaning. For example `{packets}`, `{errors}`, `{faults}`,
-etc.
+give additional meaning *without* the leading default unit (`1`). For example
```
pirgeo, line 154:
```suggestion
give additional meaning *without* the leading default unit (`1`). For example, use
```
would make it even more clear IMO.

pirgeo (approved):

specification/metrics/semantic_conventions/README.md
pichlermarc: Agreed, suggestion added in b48216d. Thanks!

bogdandrutu:
@arminru can you please rebase and merge this PR since you have all the right access?

PR #2215, opened 2021-12-09
Remove usage of baggage in View from initial SDK specification.

### LINKED ISSUE #2160: Remove baggage handling from metrics SDK for now
Reading through the spec in context of #2150 to check it vs Java.

I think there should be a higher level cross-signal discussion about baggage before adding it to any particular signal. Notably, in Java instrumentation we generate one `Attributes` and would expect baggage to be populated into this via instrumentation. It would still be ok if the SDK in addition provided a mechanism for this transformation but would be awkward in two places, and the UX doesn't seem fleshed out vs the usage of baggage in traces and logs. I'm worried about jumping the gun on this.

The current restriction to synchronous instruments also seems to tie the behavior of the view to the usage within the API (the view doesn't really know if data comes from synchronous or not). This seems like an unexpected coupling.

Finally, it seems unclear what it means to get "extra dimensions" from `Baggage/Context` - baggage is clear, but context?

jsuereth:
Fixes #2160

## Changes

- Remove the interaction between `View` configuration + `Baggage` labels.

This is done to allow room for isntrumentation-library specification and a possible unified Trace/Metric baggage interaction semantic that would be orthogonal to what `View`s provide.

### DIFF
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -141,11 +141,6 @@ are output by the SDK. Here are some examples when a `View` might be needed:
   for each HTTP status code). There could also be extreme scenarios in which the
   application developer does not need any dimension (e.g. just get the total
   count of all incoming requests).
-* Add additional dimension(s) from the [Context](../context/context.md). For
-  example, a [Baggage](../baggage/api.md) value might be available indicating
-  whether an HTTP request is coming from a bot/crawler or not. The application
-  developer might want this to be converted to a dimension for HTTP server
-  metrics (e.g. the request/second from bots vs. real users).
 
 The SDK MUST provide the means to register Views with a `MeterProvider`. Here
 are the inputs:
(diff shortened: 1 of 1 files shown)


cijothomas:
This doesn't prevent a user from manually accessing Baggage, and using values from it as attributes when doing `counter.Add(10, {key="foo", value = ""Baggage.GetValue("foo")})`, right?

cijothomas (approved):
Consider adding this to changelog as well.

anuraaga (approved):

jsuereth:
> This doesn't prevent a user from manually accessing Baggage, and using values from it as attributes when doing counter.Add(10, {key="foo", value = ""Baggage.GetValue("foo")}), right?

Yes.  The idea is that any `Baggage` => Signal interaction probably should span Traces + Metrics + Logs and we want to save room to figure that aspect out.

jmacd (approved):

MrAlias (approved):

PR #2221, opened 2021-12-13
Clarifications on periodic exporting metric reader

### LINKED ISSUE #2140: Clarifications on Periodic metric reader

Implementing the SDK spec in Python

**Additional context.**

Copying for reference [Periodic exporting metric reader](https://github.com/open-telemetry/opentelemetry-specification/blob/e2a4f057a56517cdc43026459e3d5a22be420147/specification/metrics/sdk.md#periodic-exporting-metricreader):

> This is an implementation of the MetricReader which collects metrics based on a user-configurable time interval, and passes the metrics to the configured Push Metric Exporter.

Since it is a `MetricReader` impl, it has a [`collect()`](https://github.com/open-telemetry/opentelemetry-specification/blob/e2a4f057a56517cdc43026459e3d5a22be420147/specification/metrics/sdk.md#collect) method. Reading between the lines, I think the intention is that periodic metric reader subclasses the pull based `MetricReader` and starts a background timer which calls `collect()` and passes the metrics to the exporter. However, `collect()` doesn't really need to be exposed at all here, and probably shouldn't be called by anything except the background timer as it can cause undesired behavior.

For example, if a periodic metric reader is configured with delta temporality and someone besides the timer calls `collect()`, the exporter will never see the delta values from that interval.

> Collect SHOULD provide a way to let the caller know whether it succeeded, failed or timed out.

What happens if collection failed? Should the reader communicate that failure with the exporter or just log a message?

aabmass:
Fixes #2140

## Changes

Clarify interaction between periodic metric reader, push exporter, and the user. Add a potential implementation option for periodic metric reader and an example sequence of events.


jsuereth (approved):

carlosalberto:
@aabmass @jsuereth Should we re-open this PR?

jmacd (approved):

carlosalberto:
cc @open-telemetry/specs-metrics-approvers

reyang (approved):

Issue #1769, opened 2021-06-21
Leaky-bucket rate limiting sampler


anuraaga:
Many current tracing solutions provide a rate limiting sampler with a leaky bucket (reservoir). Such a sampler allows serving requests up to a fixed number per second (reservoir) and if that is passed enforces a rate limit on additional requests in the second. As requests die down, the reservoir can be filled again.

Java SDK currently has an implementation for Jaeger remote sampler support

https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/RateLimitingSampler.java

I think it's a good algorithm to have in general and relatively simple. Is it possible to add it to the spec as a possible sampler type?

oertl:
I do not support adding this sampler. This sampler makes extrapolation difficult as the sampling probability cannot be determined. Furthermore, this "first come, first served" approach is not fair (which is in contrast to reservoir sampling). As example, consider the following sequence of events on the time axis
```
                       E1 E2                                       E3 E4                                       E5 E6
 ______________________|__|________________________________________|__|________________________________________|__|______
```
Obvioulsy, events E1, E3, E5 will have a higher probability to be sampled than E2, E4, E6 as the available balance (https://github.com/open-telemetry/opentelemetry-java/blob/4844b82a59c5172c6b1d7206a2b7a3a3ea5d748f/sdk/common/src/main/java/io/opentelemetry/sdk/internal/RateLimiter.java#L53) tends to be higher when making the sampling decision for E1, E3, E5.

carlosalberto:
cc @jmacd

anuraaga:
Here's the implementation from Brave. If we need some logic to distribute across the second, we could try to adapt it, I think improving the fairness is an implementation detail of a ratelimiting sampler

https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/sampler/RateLimitingSampler.java#L120

For Zipkin users, we've found the ratelimiting sampler to be very popular. Often it's a rate limit that a tracing team may give you quota for, or you're paying by the trace, so mapping a ratelimiting sampler to these restrictions is extremely intuitive. In addition, it removes the need to target a random sampler to peak load, which can be an exercise in trial / error as well as then provide undersampling / missed errors during low load, traffic patterns changing for a service is very common especially for medium scale services.

[...]

PR #2222, opened 2021-12-14
Add JaegerRemoteSampler spec

pavolloffay:
Signed-off-by: Pavol Loffay <p.loffay@gmail.com>

This PR defines `JaegerRemoteSampler` it is basically a follow-up after https://github.com/open-telemetry/opentelemetry-specification/pull/1791.

Jaeger remote sampler is already available in:
* Collector: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/jaegerreceiver#remote-sampling
* Java: https://github.com/open-telemetry/opentelemetry-java/tree/main/sdk-extensions/jaeger-remote-sampler
* Go: soon https://github.com/open-telemetry/opentelemetry-go-contrib/pull/936

cc) @jmacd @carlosalberto @willarmiros

### DIFF
--- a/specification/sdk-environment-variables.md
+++ b/specification/sdk-environment-variables.md
@@ -65,13 +65,14 @@ Known values for `OTEL_TRACES_SAMPLER` are:
 - `"parentbased_always_on"`: `ParentBased(root=AlwaysOnSampler)`
 - `"parentbased_always_off"`: `ParentBased(root=AlwaysOffSampler)`
 - `"parentbased_traceidratio"`: `ParentBased(root=TraceIdRatioBased)`
+- `"parentbased_jaeger_remote"`: `ParentBased(root=JaegerRemoteSampler)`
 - `"jaeger_remote"`: `JaegerRemoteSampler`
 - `"xray"`: [AWS X-Ray Centralized Sampling](https://docs.aws.amazon.com/xray/latest/devguide/xray-console-sampling.html) (_third party_)
 
 Depending on the value of `OTEL_TRACES_SAMPLER`, `OTEL_TRACES_SAMPLER_ARG` may be set as follows:
 
 - For `traceidratio` and `parentbased_traceidratio` samplers: Sampling probability, a number in the [0..1] range, e.g. "0.25". Default is 1.0 if unset.
-- For `jaeger_remote`: The value is a comma separated list:
+- For `jaeger_remote` and `parentbased_jaeger_remote`: The value is a comma separated list:
   - `endpoint`: the endpoint in form of `scheme://host:port` of gRPC server that serves the sampling strategy for the service ([sampling.proto](https://github.com/jaegertracing/jaeger-idl/blob/master/proto/api_v2/sampling.proto)).
   - `pollingIntervalMs`:  in milliseconds indicating how often the sampler will poll the backend for updates to sampling strategy.
   - `initialSamplingRate`:  in the [0..1] range, which is used as the sampling probability when the backend cannot be reached to retrieve a sampling strategy. This value stops having an effect once a sampling strategy is retrieved successfully, as the remote strategy will be used until a new update is retrieved.
--- a/specification/trace/sdk.md
+++ b/specification/trace/sdk.md
@@ -356,7 +356,7 @@ Optional parameters:
 
 #### JaegerRemoteSampler
 
-[Jaeger remote sampler](https://www.jaegertracing.io/docs/1.29/sampling/#collector-sampling-configuration) allows remotely controlling the sampling configuration for the SDKs. The sampling is typically configured at the collector and the SDKs actively poll for changes. The sampler uses `TraceIdRatioBased` or `RateLimiting` sampler under the hood. These samplers can be configured per whole service (a.k.a default), or per span name in a given service (a.k.a per operation).
+[Jaeger remote sampler](https://www.jaegertracing.io/docs/1.29/sampling/#collector-sampling-configuration) allows remotely controlling the sampling configuration for the SDKs. The sampling is typically configured at the collector and the SDKs actively poll for changes. The sampler uses `TraceIdRatioBased` or rate-limited sampler under the hood. These samplers can be configured per whole service (a.k.a default), or per span name in a given service (a.k.a per operation).
 
 The full Protobuf definition can be found at [jaegertracing/jaeger-idl/api_v2/sampling.proto](https://github.com/jaegertracing/jaeger-idl/blob/main/proto/api_v2/sampling.proto).
 
--- a/specification/trace/sdk.md
+++ b/specification/trace/sdk.md
@@ -23,6 +23,8 @@
     + [TraceIdRatioBased](#traceidratiobased)
       - [Requirements for `TraceIdRatioBased` sampler algorithm](#requirements-for-traceidratiobased-sampler-algorithm)
     + [ParentBased](#parentbased)
+    + [JaegerRemoteSampler](#jaegerremotesampler)
+      - [Configuration](#configuration)
 - [Span Limits](#span-limits)
 - [Id Generators](#id-generators)
 - [Span processor](#span-processor)
@@ -351,6 +353,20 @@ Optional parameters:
 |present|true|false|`remoteParentNotSampled()`|
 |present|false|true|`localParentSampled()`|
 |present|false|false|`localParentNotSampled()`|
+
+#### JaegerRemoteSampler
+
+[Jaeger remote sampler](https://www.jaegertracing.io/docs/1.29/sampling/#collector-sampling-configuration) allows remotely controlling the sampling configuration for the SDKs. The sampling is typically configured at the collector and the SDKs actively poll for changes. The sampler uses `TraceIdRatioBased` or `RateLimiting` sampler under the hood. These samplers can be configured per whole service (a.k.a default), or per span name in a given service (a.k.a per operation).
+
+The full Protobuf definition can be found at [jaegertracing/jaeger-idl/api_v2/sampling.proto](https://github.com/jaegertracing/jaeger-idl/blob/main/proto/api_v2/sampling.proto).
+
+##### Configuration
+
+Following configuration properties should be available when creating the sampler:
+
+* endpoint - collector address with running service with sampling manager
+* polling interval - polling interval for getting configuration from remote
+* initial sampler - initial sampler that is used before the first configuration is fetched
 
 ## Span Limits
 


specification/trace/jaeger-remote-sampler.md
```diff
@@ -0,0 +1,134 @@
...
+  // Per operation/span (name) sampling configuration.
+  repeated OperationSamplingStrategy perOperationStrategies = 3;
+  // Default upper bound for rate limiting sampler.
+  double defaultUpperBoundTracesPerSecond = 4;
+}
+
+// SamplingStrategyResponse defines sampling configuration for a service.
+message SamplingStrategyResponse {
+  // Default sampler to use - either ParentBased(root=TraceIdRatioBased) or RateLimiting.
+  SamplingStrategyType strategyType = 1;
```
yurishkuro, line 53: This might be confusing - is the type meant to indicate the choice of one of the samplers below? If so, why does the enum not have per-operation? I think the interpretation is a bit more convoluted.

specification/trace/jaeger-remote-sampler.md
```diff
@@ -0,0 +1,134 @@
...
+  rpc GetSamplingStrategy(SamplingStrategyParameters) returns (SamplingStrategyResponse) {
+    option (google.api.http) = {
+      post: "/api/v2/samplingStrategy"
+      body: "*"
+    };
+  }
+}
+```
+
+### Example configuration
```
yurishkuro, line 79: What is the objective of having this here? This configuration's interpretation is somewhat complicated.

specification/trace/jaeger-remote-sampler.md
```diff
@@ -0,0 +1,134 @@
+# Jaeger remote sampler
+
+[Jaeger remote sampler](https://www.jaegertracing.io/docs/1.29/sampling/#collector-sampling-configuration) allows remotely specify sampling configuration for SDKs. The sampling is typically configured at the Jaeger or OpenTelemetry collector and SDKs actively poll for changes. The sampler uses `ParentBased(root=TraceIdRatioBased)` or `RateLimitting` sampler. These samplers can be configured as default, per service or span name in a given service.
```
yurishkuro, line 3:
```suggestion
[Jaeger remote sampler](https://www.jaegertracing.io/docs/1.29/sampling/#collector-sampling-configuration) allows remotely controlling the sampling configuration for the SDKs. The sampling is typically configured at the Jaeger or OpenTelemetry collector and the SDKs actively poll for changes. The sampler uses `ParentBased(root=TraceIdRatioBased)` or `RateLimitting` sampler under the hood. These samplers can be configured per the whole service (a.k.a. "default"), or per span name in a given service (a.k.a. "per operation").
```

yurishkuro: Shouldn't rate limiting also be wrapped in parent-based?

willarmiros:
cc @bhautikpip as well

specification/trace/jaeger-remote-sampler.md
pavolloffay: The objective is to show how users configure the sampler on the collector.

yurishkuro: But it's not part of the remote sampler configuration protocol. Sampler asks the backend "gimme strategy for my service X". There could be different implementations of how the backend decides on what to return, and this example does not really explain _how_ those decisions are made (e.g. default sampler having operations section means there are certain overlays happening in the backend).

specification/trace/jaeger-remote-sampler.md
pavolloffay: I think that is better

specification/trace/jaeger-remote-sampler.md
pavolloffay: I removing it as it seems to be controversial.

specification/trace/jaeger-remote-sampler.md
```diff
@@ -0,0 +1,87 @@
...
+  rpc GetSamplingStrategy(SamplingStrategyParameters) returns (SamplingStrategyResponse) {
+    option (google.api.http) = {
+      post: "/api/v2/samplingStrategy"
+      body: "*"
+    };
+  }
+}
+```
+
+The full Protobuf definition can be found at [jaegertracing/jaeger-idl/api_v2/sampling.proto](https://github.com/jaegertracing/jaeger-idl/blob/main/proto/api_v2/sampling.proto).
```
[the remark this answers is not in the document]
yurishkuro: +1, I'd prefer to just link to the original proto in jaeger-idl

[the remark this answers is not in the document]
pavolloffay: The comments in this repo are tailored towards OTEL vocabulary.

specification/trace/jaeger-remote-sampler.md
yurishkuro: github.com/jaegertracing/jaeger-idl/pull/87

specification/trace/jaeger-remote-sampler.md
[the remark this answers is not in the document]
yurishkuro: Yes, but the only relevant part is how to interpret the strategies themselves, which we can do in plain text, e.g. "`ProbabilisticSamplingStrategy` SHOULD be implemented via `ParentBased(root=TraceIdRatioBased)` sampler"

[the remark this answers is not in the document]
pavolloffay: Fair enough. 

Alternatively, we could make a copy and add it to OTEL proto, it will make it easier for SDKs to include it.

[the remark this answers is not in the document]
yurishkuro: BTW, I don't think we need to be mentioning `ParentBased`. For people who want that behavior, it can be composed via `ParentBased(root=JaegerRemoteSampler)`, rather than `JaegerRemoteSampler` internally having a dependency on `ParentBased`.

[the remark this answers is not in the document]
pavolloffay: I think that is good point, however, we should be thinking how this will be configured in auto-instrumentation. There only sampler type and argument can be configured. 

I have added `- "parentbased_jaeger_remote": ParentBased(root=JaegerRemoteSampler)` to the SDK env variables.

specification/sdk-environment-variables.md
```diff
@@ -65,13 +65,14 @@ Known values for `OTEL_TRACES_SAMPLER` are:
...
 - `"parentbased_traceidratio"`: `ParentBased(root=TraceIdRatioBased)`
+- `"parentbased_jaeger_remote"`: `ParentBased(root=JaegerRemoteSampler)`
 - `"jaeger_remote"`: `JaegerRemoteSampler`
 - `"xray"`: [AWS X-Ray Centralized Sampling](https://docs.aws.amazon.com/xray/latest/devguide/xray-console-sampling.html) (_third party_)
 
 Depending on the value of `OTEL_TRACES_SAMPLER`, `OTEL_TRACES_SAMPLER_ARG` may be set as follows:
 
 - For `traceidratio` and `parentbased_traceidratio` samplers: Sampling probability, a number in the [0..1] range, e.g. "0.25". Default is 1.0 if unset.
-- For `jaeger_remote`: The value is a comma separated list:
+- For `jaeger_remote` and `parentbased_jaeger_remote`: The value is a comma separated list:
```
pavolloffay, line 75: @yurishkuro I have added this to address the parent based and auto-instrumentation

yurishkuro (approved):

william-tran:
There's discussion in the Otel Sampling SIG about aligning on an OTel native remote sampler, and as a user I'd still want to see that happen as I'm looking for features like matching Samplers to any attributes instead of just service and operation name. But I see this as a way to help Jaeger users transition to OTel; while addressing the need to evolve remote sampling to enable more fine grained control can continue to happen.

specification/trace/sdk.md
```diff
@@ -352,6 +354,20 @@ Optional parameters:
 |present|false|true|`localParentSampled()`|
 |present|false|false|`localParentNotSampled()`|
 
+#### JaegerRemoteSampler
+
+[Jaeger remote sampler](https://www.jaegertracing.io/docs/1.29/sampling/#collector-sampling-configuration) allows remotely controlling the sampling configuration for the SDKs. The sampling is typically configured at the collector and the SDKs actively poll for changes. The sampler uses `TraceIdRatioBased` or `RateLimiting` sampler under the hood. These samplers can be configured per whole service (a.k.a default), or per span name in a given service (a.k.a per operation).
```
jmacd, line 359: nit: OTel doesn't have any "RateLimiting" sampler specified. (We agree that it should, it just doesn't.)

jmacd (approved):

specification/trace/sdk.md
yurishkuro: huh, good catch

NB: I merged the PR adding comments to jaeger-idl/sampling.proto

bogdandrutu (approved):

specification/trace/sdk.md
pavolloffay: Is there an issue to track the progress on the rate limiting sampler? I could add it to the doc

pavolloffay:
> 'd still want to see that happen as I'm looking for features like matching Samplers to any attributes instead of just service and operation name. But I see this as a way to help Jaeger users transition to OTel; while addressing the need to evolve remote sampling to enable more fine grained control can continue to happen.

+1. This effort is mostly to enable migration. I do agree that more generic remote sampling is needed in the long run.

specification/trace/sdk.md
jmacd: Sampling SIG has discussed several rate-limited sampler designs. 

There are (1) probabilistic/adaptive approaches that can't promise a hard rate limit, (2) probabilistic/tail-based approaches that can't promise a complete trace, and (3) non-probabilistic approaches like Leaky Bucket that can't promise accurate counts.

For (1), the adaptive route is straightforward and gives accurate counts, with varying degrees of complexity the sampler probably stays under a rate limit.

For (2), @oertl has developed a consistent-probability tail-based rate-limited sampler algorithm that is worth comparing with adaptive approaches.

For (3), the description given to "zero adjusted count" and "composition of probability and non-probability samplers" in #2047 has to do with admitting this third approach. I think we should write in Leaky Bucket because it is well understood by the community.

jmacd (approved):

specification/trace/sdk.md
pavolloffay: Should I update the text to mention that the implementation is Leaky Bucket? 

In my opinion, it's fine not mentioning it.

yurishkuro: It's not a requirement of Jaeger remote sampling protocol that rate limiter is a leaky bucket.

bogdandrutu (approved):

pavolloffay:
Can somebody merge the PR? The GH actions bot wants to close it

yurishkuro:
@pavolloffay would merging it introduce a contradiction since OTEL has no rate limiting sampler?

jmacd:
> @pavolloffay would merging it introduce a contradiction since OTEL has no rate limiting sampler?

I don't mind this contradiction. In the terminology of #2047, RateLimited is some sort of non-probability Sampler and not very specific. I understand that often RateLimited is implemented by a LeakyBucket sampler, which has 2 (?) parameters. Perhaps replace "RateLimited" by "rate-limited" and file an issue to define the LeakyBucket rate-limited sampler in a future iteration?

yurishkuro:
> Perhaps replace "RateLimited" by "rate-limited"

Seems reasonable. I do think implementing rate limiter will be needed, especially to guarantee minimal level of sampling of rare endpoints, so this would be an added requirement for SDKs that want to support Jaeger's remote sampling policy format.

> often RateLimited is implemented by a LeakyBucket sampler, which has 2 (?) parameters

the algorithm has two parameters, but the sampler usually has only one, the rate of traces / sec sampled (could be fractional).

specification/trace/sdk.md
```diff
@@ -352,6 +354,20 @@ Optional parameters:
...
+
+The full Protobuf definition can be found at [jaegertracing/jaeger-idl/api_v2/sampling.proto](https://github.com/jaegertracing/jaeger-idl/blob/main/proto/api_v2/sampling.proto).
+
+##### Configuration
+
+Following configuration properties should be available when creating the sampler:
+
+* endpoint - collector address with running service with sampling manager
+* polling interval - polling interval for getting configuration from remote
+* initial sampler - initial sampler that is used before the first configuration is fetched
```
carlosalberto, line 369: I take for granted the values are exactly https://github.com/jaegertracing/jaeger-idl/blob/main/proto/api_v2/sampling.proto#L35 ? If yes, maybe worth adding the actual used values?

carlosalberto (approved):

carlosalberto:
Left a non-blocking comment, otherwise LGTM

anuraaga:
> Perhaps replace "RateLimited" by "rate-limited" and file an issue to define the LeakyBucket rate-limited sampler in a future iteration?

#1769 is the issue for a rate limiting sampler. My impression from the thread is it was rejected for addition to OTel since it doesn't support probabilistic sampling 😃

jmacd:
> My impression from the thread is it was rejected for addition to OTel since it doesn't support probabilistic sampling

Right. Probability sampling is preferable to non-probability sampling. Speaking as a vendor, we are not interested in receiving sampled spans without being able to count them. #2047 specifies how to mix these two kinds of Sampler, and I consider this to admit specifications for non-probability samplers when it merges. When that happens, the user that @yurishkuro describes, who wants to combine probability sampling with a minimum-rate could have options:

(a) A leaky bucket sampler (for minimum-rate sampling) composed with a probability sampler (for Span-to-metrics)
(a) An adaptive sampler that adjusts a simple probability sampler to maintain a soft rate limit, in which case all spans are countable but the rate limit is not guaranteed.

Given these choices, I'd choose (b) and I expect the complexity of the adaptive logic to be less than the complexity of the leaky bucket algorithm. This is why I suggested we write "rate-limited" instead of naming a specific algorithm, because naming a specific algorithm requires, well, specifying it. Do you want to write the specification for LeakyBucket? I'd prefer to write the adaptive sampler specification.

carlosalberto:
@anuraaga @jmacd should we follow the discussion either in #1769 and merge this? Or have a blocking request?

carlosalberto:
@pavolloffay This PR needs to be updated ;(

anuraaga (approved):

pavolloffay:
@carlosalberto PR rebased and renamed `RateLimitting` to rate-limited per @jmacd request.

PR #2225, opened 2021-12-15
Document the case of activating a null value in the OT Shim.

carlosalberto:
Addresses an item from #2147 

Even though this is not documented in the OpenTracing Specification, the Java, Python and DotNet APIs follow this pattern (the remaining implementations do not define a `ScopeManager`). 

For details, see https://github.com/open-telemetry/opentelemetry-java/issues/3952 

cc @yurishkuro


specification/compatibility/opentracing.md
```diff
@@ -328,19 +328,27 @@ Parameters:
...
+If the specified `Span` is null, it MUST be set to a
+[NonRecordableSpan](../trace/api.md#wrapping-a-spancontext-in-a-span)
+wrapping an invalid `SpanContext`, to signal there is no active
+`Span` nor `Baggage`.
 
 ```java
 Scope activate(Span span) {
-  SpanContextShim spanContextShim = getSpanContextShim(span);
+  if (span == null) {
+    span = new SpanShim(io.opentelemetry.api.trace.Span.getInvalid());
```
yurishkuro, line 342: Is invalid the same thing that we normally use in OTEL when the trace is not sampled? If so, it can become active span naturally and should be returned in the method below instead of null, I.e. we need to be able to distinguish the two cases.

carlosalberto: > Is invalid the same thing that we normally use in OTEL when the trace is not sampled?

That's correct -- that is, upon no active `Span`, an invalid (no-op) is returned by OTel. Will add a note just to remark this.

> If so, it can become active span naturally and should be returned in the method below instead of null, I.e. we need to be able to distinguish the two cases.

Can you elaborate?

carlosalberto: For case 1, the SDKs will return no-op Span wrapping a **valid** `SpanContext` (sampled flag = false). And as this `SpanContext` is valid, it won't be considered as a `null` Span by the OT Shim.

So, yes, `scopeManger.activeSpan() == S1` ;)

carlosalberto: @yurishkuro Added a small test case, just to verify this: https://github.com/open-telemetry/opentelemetry-java/compare/main...carlosalberto:otshim-activate-unsampled?expand=1

yurishkuro: This is why I asked "Is invalid the same thing that we normally use in OTEL when the trace is not sampled?" Right now you're recommending "NonRecordableSpan wrapping an invalid `SpanContext`", is such construct actually always possible to create?

carlosalberto: > "NonRecordableSpan wrapping an invalid SpanContext"

Yes, this construct is always available, as per the [Specification's API section](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#wrapping-a-spancontext-in-a-span).

Added a small clarification below (better be safe than sorry).

yurishkuro (approved):

jmacd (approved):

specification/compatibility/opentracing.md
anpatellll: Nice, this was my concern as well, 👍

PR #2227, opened 2021-12-16
Define the instrumentation library name for the OT Shim.

carlosalberto:
Fixes an item in #2147 

Currently the Spec says that a `Tracer` should be supplied when creating an OT Shim, although both Java and Python require a `TracerProvider` instead:

- Python uses `opentelemetry-opentracing-shim` as instrumentation library.
- Java uses `opentracingshim` as instrumentation library.
- DotNet/Go simply take a `Tracer`.

I think it would be a good thing to use a single name for the Shim (as it's the same component, just in different languages), but would love to hear opinions.

cc @yurishkuro


specification/compatibility/opentracing.md
```diff
@@ -59,7 +59,9 @@ This operation is used to create a new OpenTracing `Tracer`:
 
 This operation MUST accept the following parameters:
 
-- An OpenTelemetry `Tracer`, used to create `Span`s.
+- An OpenTelemetry `TracerProvider`, used to get a `Tracer` with
+  instrumentation name and version set to `opentracing-shim`
+  and the shim version respectively.
```
yurishkuro, line 64: I agree about passing TracerProvider. I think the note about instrumentation name `opentracing-shim` should be slightly rephrased, like "the shim factory / constructor MUST use the TracerProvider to obtain a tracer with the name `opentracing-shim`"

carlosalberto: Updated, please review.

yurishkuro (approved):

jmacd (approved):

PR #2228, opened 2021-12-17
Add a Start Span section to the OT Shim.

carlosalberto:
Addresses an item in #2147 

For some reason I had lost this section, so adding it again. Namely, the important information is:

* When specifying `Span` references, use the first one as parent, and all of them as `Link`s (this is what the OT Shims do). The reference type is added as a `opentracing.ref_type` attribute on the `Link` itself.
* Add a small `Span References` section, to clarify that the OT `ChildOf` and `FollowsFrom` reference types must not be confused with OTel `Link`s.
* Tags/attributes specified at `Span` creation MUST be set before the `Span` is actually created, in case they are passed to some internal API, e.g. the Sampling API. See this [Java issue](https://github.com/open-telemetry/opentelemetry-java/issues/3962).

Please review @jmacd @yurishkuro


specification/compatibility/opentracing.md
```diff
@@ -88,6 +90,38 @@ See OpenTracing Propagation
...
+- An optional set of [tags](#set-tag).
+- An optional explicit start timestamp, a numeric value.
+
+For OpenTracing languages implementing the [ScopeManager](#scopemanager-shim)
+interface, the folllowing parameters are defined as well:
+
+- An optional boolean specifying whether the current `Span`
+  should be ignored as automatic parent.
+
+If a set of `Span` references is specified, the first `SpanContext`
```
yurishkuro, line 108: I think this is incorrect and could lead to data loss for backends like Jaeger. Specifically:
  - the type of OT references must be preserved as some transitional attribute (eg ot.ref.type) of the links, so that for example OTEL to Jaeger converter could use it to reconstruct ref type.
  - Jaeger used to have a notion of strong parent span via dedicated parent-id field, but we moved away from it to just a list of refs. However the question of parent often comes up, eg in the UI, and the logic we used was:
    - the first reference of type child-of is used as parent, otherwise just any first reference
    - all refs are preserved

specification/compatibility/opentracing.md
```diff
@@ -402,3 +436,23 @@ Span active() {
...
+
+OpenTracing defines two types of references:
+
+* **Child Of**: The parent `Span` depends o the child `Span`
+  in some capacity.
+* **Follows From**: The parent `Span` does not depend in any
+  way on the result of their child `Span`s.
+
+OpenTelemetry does not define strict equivalent semantics for these
+references, and the **Child Of** relationship can be assumed.
```
yurishkuro, line 456: Per above, we need to preserve ot ref type. The shim must not cause a data loss.

carlosalberto:
Hey @yurishkuro - updated the text to mention that:

1) From the specified references, the first value with type CHILD_OF is the parent, and if no value with this type exists, use the first value. The remaining values are used as Links.
2) The values added as Links will have a new attribute (part of the Link itself), i.e. `opentracing.ref_type=child_of|follows_from`. The parent does not keep any attribute - should we add one for that case too?

Let me know whether this aligns with what you have in mind.

(Semantic convention generation is not working for whatever reason, will figure that out)

specification/compatibility/opentracing.md
```diff
@@ -88,6 +90,41 @@ See OpenTracing Propagation
...
+
+For OpenTracing languages implementing the [ScopeManager](#scopemanager-shim)
+interface, the folllowing parameters are defined as well:
+
+- An optional boolean specifying whether the current `Span`
+  should be ignored as automatic parent.
+
+If a list of `Span` references is specified, the first `SpanContext`
+with **Child Of** type in the entire list is used as parent, else the
+the first `SpanContext` is used as parent. The remaining values in the list
```
yurishkuro, line 110:
```suggestion
the first `SpanContext` is used as parent. All values in the list
```

carlosalberto: Aha! Interesting, I thought only the first item was added as a `Link`. Applied.

yurishkuro (approved):

yurishkuro:
keep

jmacd:
> The reference type itself is discarded (we MAY use it later on, if we define equivalent semantics).

I take it this refers to the context that is taken for the parent? Sounds like the shim StartSpan() could be extended to select the OTel SpanKind but we probably won't. :+1:

jmacd (approved):

carlosalberto:
@jmacd Actually that's an outdated part of the description (fixed now), as now we keep the reference type as a `Link` attribute (`opentracing.ref_type`).

> Sounds like the shim StartSpan() could be extended to select the OTel SpanKind but we probably won't. 

I think that, as the semantics are not strictly equivalent, we would have to spend some cycles to define this. Let me know if you think it's important to consider this.

jmacd:
> Let me know if you think it's important to consider this.

I do not! 😀

PR #575, opened 2020-04-23
Revise and extend semantic conventions for databases

arminru:
We revised and extended the semantic conventions for database client calls that were initially copied over from OpenTracing by making definitions more precise and adding technology-specific attributes.
This should lead to improved data quality and allow better identification of databases and analysis of calls made to them.

Resolves #245.

PR #2233, opened 2021-12-20
Add Azure SDK semantic conventions

lmolkova:
Adds yml spec for Azure SDK trace semantic conventions. This doc only documents the current state, it is not stable and has experimental status.

As Otel conventions stabilize, Azure SDKs will gradually onboard onto them.

- HTTP client: compliant with Otel [http conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/9216eb8ff1832ae790369e49ce326cb6bc3d1c80/semantic_conventions/trace/http.yaml) (is a valid subset of)
- Messaging: predates otel conventions
- DB -  represents an older version of Otel convention prior to https://github.com/open-telemetry/opentelemetry-specification/pull/575

### DIFF
--- a/semantic_conventions/trace/instrumentation/azure-sdk.yml
+++ b/semantic_conventions/trace/instrumentation/azure-sdk.yml
@@ -1,0 +1,129 @@
+# Version: 0.0.0
+# Status: Experimental
+# This document describes Azure SDK trace semantic conventions. For more details refer to https://azure.github.io/azure-sdk/distributed_tracing_conventions.html
+
+groups:
+  # common
+  - id: azure-sdk
+    brief: 'Describes Azure SDK spans.'
+    attributes:
+      - id: az.namespace
+        required: always
+        type: string
+        brief: '[Namespace](https://docs.microsoft.com/azure/azure-resource-manager/management/azure-services-resource-providers) of Azure service request is made against.'
+        examples: ['Microsoft.Storage', 'Microsoft.KeyVault', 'Microsoft.ServiceBus']
+
+  # public API
+  - id: azure-sdk.api
+    span_kind: internal
+    extends: azure-sdk
+    brief: 'Describes Azure SDK API calls spans.'
+    note: 'Represents public surface API calls that wrap an Azure service call.'
+
+  # http
+  - id: azure-sdk.http
+    extends: azure-sdk
+    span_kind: client
+    brief: 'Describes HTTP client spans created per HTTP request (try).'
+    note: >
+      Follows [OpenTelemetry HTTP conventions v1.6.x](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.6.x/specification/trace/semantic_conventions/http.md)
+      but omits all optional attributes, providing only `http.url` to describe destination. It adds request-id attributes supported by Azure services.
+    attributes:
+      - id: request_id
+        type: string
+        required: always
+        brief: 'Value of the [x-ms-client-request-id] header (or other request-id header, depending on the service) sent by the client.'
+        examples: ['eb178587-c05a-418c-a695-ae9466c5303c']
+      - id: service_request_id
+        type: string
+        required: always
+        brief: 'Value of the [x-ms-request-id]  header (or other request-id header, depending on the service) sent by the server in response.'
+        examples: ['3f828ae5-ecb9-40ab-88d9-db0420af30c6']
+      - id: http.method
+        type: string
+        required: always
+        sampling_relevant: true
+        brief: 'HTTP request method.'
+        examples: ["GET", "POST", "HEAD"]
+      - id: http.url
+        type: string
+        required: always
+        sampling_relevant: true
+        brief: Full HTTP request URL in the form `scheme://host[:port]/path?query[#fragment]`
+        examples: ['https://www.foo.bar/search?q=OpenTelemetry#SemConv']
+      - id: http.status_code
+        type: int
+        required:
+          conditional: If and only if one was received/sent.
+        brief: '[HTTP response status code](https://tools.ietf.org/html/rfc7231#section-6).'
+        examples: [200]
+      - id: http.user_agent
+        type: string
+        required: always
+        brief: 'Value of the [HTTP User-Agent](https://tools.ietf.org/html/rfc7231#section-5.5.3) or X-MS-UserAgent header sent by the client.'
+        examples: ['CERN-LineMode/2.15 libwww/2.17b3']
+
+  # messaging
+  - id: azure-sdk.messaging
+    brief: 'Describes Azure messaging SDKs spans.'
+    extends: azure-sdk
+    note: Implements OpenTracing MessageBus conventions https://opentracing.io/specification/conventions/
+    attributes:
+      - id: message_bus.destination
+        type: string
+        required: always
+        brief: 'Name of the messaging entity within namespace: e.g EventHubs name, ServiceBus queue or topic name.'
+        examples: ['myqueue', 'myhub']
+      - id: peer.address
+        type: string
+        brief: 'Fully qualified messaging service name.'
+        required: always
+        examples: ['myEventHubNamespace.servicebus.windows.net']
+  - id: azure-sdk.messaging.producer
+    span_kind: producer
+    extends: azure-sdk.messaging
+    brief: 'Describes producer span created per event.'
+
+  - id: azure-sdk.messaging.send
+    span_kind: client
+    extends: azure-sdk.messaging
+    brief: 'Describes send (transport call) span.'
+    note: 'Contains links to all messages contexts being sent.'
+
+  - id: azure-sdk.messaging.process
+    span_kind: consumer
+    extends: azure-sdk.messaging
+    brief: 'Describes consumption span.'
+    note: >
+      Contains links to all messages contexts being consumed. Each link has attribute `enqueuedTime` (with `long` type)
+      attribute with unix epoch time with milliseconds precision representing when message was enqueued.
+
+  # CosmosDB
+  - id: azure-sdk.cosmos
+    span_kind: client
+    brief: 'Describes Azure CosmosDB spans.'
+    note: >
+      Events with additional debug info are added for long running operations.
+      Implements https://github.com/open-telemetry/opentelemetry-specification/blob/v0.5.0/specification/trace/semantic_conventions/database.md
+    extends: azure-sdk
+    attributes:
+      - id: db.url
+        type: string
+        required: always
+        brief: 'Cosmos DB URI'
+        examples: ['https://my-cosmos.documents.azure.com:443/']
+      - id: db.statement
+        type: string
+        required: always
+        brief: 'Database statement'
+        examples: ['createContainerIfNotExists.myContainer']
+      - id: db.instance
+        type: string
+        required: always
+        brief: 'Database name'
+        examples: ['mydb']
+      - id: db.type
+        type: string
+        required: always
+        brief: 'Database type'
+        examples: ['Cosmos']
--- a/semantic_conventions/trace/instrumentation/azure-sdk.yml
+++ b/semantic_conventions/trace/instrumentation/azure-sdk.yml
@@ -4,31 +4,30 @@
 
 groups:
   # common
-  - id: azure.sdk
-    brief: 'Describes Azure SDK spans.'
+  - id: azure
+    brief: 'Describes Azure spans.'
     attributes:
       - id: az.namespace
         required: always
         type: string
         sampling_relevant: true
-        brief: '[Namespace](https://docs.microsoft.com/azure/azure-resource-manager/management/azure-services-resource-providers) of Azure service request is made against.'
+        brief: '[Namespace](https://docs.microsoft.com/azure/azure-resource-manager/management/azure-services-resource-providers) of Azure service that the request is made against.'
         examples: ['Microsoft.Storage', 'Microsoft.KeyVault', 'Microsoft.ServiceBus']
 
   # public API
   - id: azure.sdk.api
     span_kind: internal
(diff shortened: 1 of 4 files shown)


semantic_conventions/trace/instrumentation/azure-sdk.yml
```diff
@@ -0,0 +1,129 @@
+# Version: 0.0.0
+# Status: Experimental
+# This document describes Azure SDK trace semantic conventions. For more details refer to https://azure.github.io/azure-sdk/distributed_tracing_conventions.html
```
arminru, line 3: Please also
1. create/edit MD files in https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/trace/semantic_conventions where these are used so they are rendered in a human-readable format using the semconv tool
2. add any new MD files to the overview in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/README.md
3. an entry in CHANGELOG.md for your changes

semantic_conventions/trace/instrumentation/azure-sdk.yml
```diff
@@ -0,0 +1,129 @@
...
+
+groups:
+  # common
+  - id: azure-sdk
```
arminru, line 7: As per https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-naming.md underscores should be used in attribute names, not dashes.

semantic_conventions/trace/instrumentation/azure-sdk.yml
```diff
@@ -0,0 +1,117 @@
+# Version: 0.0.0
+# Status: Experimental
+# This document describes Azure SDK trace semantic conventions. For more details refer to https://azure.github.io/azure.sdk/distributed_tracing_conventions.html
+
+groups:
+  # common
+  - id: azure.sdk
```
anuraaga, line 7: Just for reference, I didn't include the `.sdk` in the aws-sdk conventions expecting that the same conventions could be used on the server side some day. I suspect that Microsoft is even further along in using OTel in various places and you may consider whether that would be helpful here

semantic_conventions/trace/instrumentation/azure-sdk.yml
```diff
@@ -0,0 +1,117 @@
...
+    brief: 'Describes Azure SDK spans.'
+    attributes:
+      - id: az.namespace
+        required: always
+        type: string
+        sampling_relevant: true
+        brief: '[Namespace](https://docs.microsoft.com/azure/azure-resource-manager/management/azure-services-resource-providers) of Azure service request is made against.'
```
anuraaga, line 14:
```suggestion
        brief: '[Namespace](https://docs.microsoft.com/azure/azure-resource-manager/management/azure-services-resource-providers) of Azure service that the request is made against.'
```

semantic_conventions/trace/instrumentation/azure-sdk.yml
```diff
@@ -0,0 +1,117 @@
...
+    span_kind: internal
+    extends: azure.sdk
+    brief: 'Describes Azure SDK API calls spans.'
+    note: 'Represents public surface API calls that wrap an Azure service call.'
+
+  # http
+  - id: azure.sdk.http
+    extends: azure.sdk
+    span_kind: client
+    brief: 'Describes HTTP client spans created per HTTP request (try).'
```
anuraaga, line 28: I think the `(try)` is referring to being for each attempt within retries, but does that actually matter for this semantic convention?

anuraaga (approved):

semantic_conventions/trace/instrumentation/azure-sdk.yml
lmolkova: good point!

semantic_conventions/trace/instrumentation/azure-sdk.yml
lmolkova: Agreed, I removed it from yml, but kept it in md file for clarity.

bogdandrutu (approved):

lmolkova:
@open-telemetry/specs-trace-approvers can someone please take a look?

lmolkova:
@arminru thanks for the review! I think I addressed your comments.

semantic_conventions/trace/instrumentation/azure-sdk.yml
```diff
@@ -0,0 +1,116 @@
+# Version: 0.0.0
+# Status: Experimental
```
arminru, line 2: Are those metadata fields used by internal tooling or intended to be used by the MD and code generator in https://github.com/open-telemetry/build-tools?

semantic_conventions/trace/instrumentation/azure-sdk.yml
```diff
@@ -0,0 +1,116 @@
...
+    attributes:
+      - id: request_id
+        type: string
+        required: always
+        brief: 'Value of the [x-ms-client-request-id] header (or other request-id header, depending on the service) sent by the client.'
+        examples: ['eb178587-c05a-418c-a695-ae9466c5303c']
+      - id: service_request_id
+        type: string
+        required: always
+        brief: 'Value of the [x-ms-request-id]  header (or other request-id header, depending on the service) sent by the server in response.'
```
arminru, line 40: Missing link (or should these be quotation marks?).

semantic_conventions/trace/instrumentation/azure-sdk.yml
```diff
@@ -0,0 +1,116 @@
...
+    span_kind: client
+    brief: 'Describes HTTP client spans created per HTTP request.'
+    note: >
+      Follows [OpenTelemetry HTTP conventions v1.8.0](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.8.0/specification/trace/semantic_conventions/http.md)
+      but omits optional attributes, providing only `http.url` to describe destination. It adds request-id attributes supported by Azure services.
+    attributes:
+      - id: request_id
+        type: string
+        required: always
+        brief: 'Value of the [x-ms-client-request-id] header (or other request-id header, depending on the service) sent by the client.'
```
arminru, line 35: Missing link (or should these be quotation marks?).

semantic_conventions/trace/instrumentation/azure-sdk.yml
```diff
@@ -0,0 +1,116 @@
+# Version: 0.0.0
+# Status: Experimental
+# This document describes Azure SDK trace semantic conventions. For more details refer to https://azure.github.io/azure.sdk/distributed_tracing_conventions.html
+
+groups:
+  # common
+  - id: azure
+    brief: 'Describes Azure spans.'
+    attributes:
+      - id: az.namespace
```
arminru, line 10: How about `prefix: azure` and `id: namespace`?

semantic_conventions/trace/instrumentation/azure-sdk.yml
```diff
@@ -0,0 +1,116 @@
+# Version: 0.0.0
+# Status: Experimental
+# This document describes Azure SDK trace semantic conventions. For more details refer to https://azure.github.io/azure.sdk/distributed_tracing_conventions.html
+
+groups:
```
arminru, line 5: All of these define an `id` but no `prefix`. I assume you'd want to set `prefix` to match your `id`s.

semantic_conventions/trace/instrumentation/azure-sdk.yml
```diff
@@ -0,0 +1,116 @@
...
+      - id: message_bus.destination
+        type: string
+        required: always
+        brief: 'Name of the messaging entity within namespace: e.g EventHubs name, ServiceBus queue or topic name.'
+        examples: ['myqueue', 'myhub']
+      - id: peer.address
+        type: string
+        brief: 'Fully qualified messaging service name.'
+        required: always
+        examples: ['myEventHubNamespace.servicebus.windows.net']
```
arminru, line 70: How about referencing `net.peer.name` from the general conventions here?

specification/trace/semantic_conventions/instrumentation/azure-sdk.md
```diff
@@ -0,0 +1,116 @@
+# Semantic conventions for Azure SDK
```
arminru, line 1: Please add your new file to the TOC (library-specific) in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/README.md.

specification/trace/semantic_conventions/instrumentation/azure-sdk.md
```diff
@@ -0,0 +1,116 @@
...
+
+**Status**: [Experimental](../../../document-status.md)
+
+This document describes tracing semantic conventions adopted by Azure SDK. Instrumentations live in Azure SDK repos and are shipped along with Azure SDK artifacts. Instrumentations cover all modern (track 2) Azure client libraries.
```
arminru, line 5: Is "track 2" something your users are familiar with or rather a note for internal devs instrumentating the SDKs?

specification/trace/semantic_conventions/instrumentation/azure-sdk.md
```diff
@@ -0,0 +1,116 @@
...
+Azure SDKs create a span for public API call (that involves communication with Azure services).
+
+- Spans representing public APIs have names following `client.method` pattern, which is language-specific.
+- For HTTP-based SDKs, public API spans have `INTERNAL` kind.
+
+[Messaging](#messaging-sdks) and [CosmosDb](#cosmosdb) sections below describe non-HTTP semantics.
+
+## HTTP Client
+
+Azure SDK implements a valid subset of [OpenTelemetry HTTP conventions v1.8.0](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.8.0/specification/trace/semantic_conventions/http.md) and create a span per HTTP call (try).
```
arminru, line 45:
```suggestion
Azure SDK implements a valid subset of [OpenTelemetry HTTP conventions v1.8.0](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.8.0/specification/trace/semantic_conventions/http.md) and create a span per HTTP call (attempt).
```
Maybe?

semantic_conventions/trace/instrumentation/azure-sdk.yml
lmolkova: those are only intended for MD generator at this point, they might also be used by Java code generator in the future. The main intent is formal documentation though.

semantic_conventions/trace/instrumentation/azure-sdk.yml
lmolkova: I'm not aware of any requirement for id and prefix to match. I don't see anything [here](https://github.com/open-telemetry/build-tools/blob/main/semantic-conventions/syntax.md#semantics) and if I check existing specs, they don't always follow this pattern.  Moreover, with things like http, we don't add any new prefix to spans.

semantic_conventions/trace/instrumentation/azure-sdk.yml
lmolkova: once we'll switch to `net.peer.name` I'll use it and add a reference. But this is instrumentation exists for years and it can't be changed overnight.

specification/trace/semantic_conventions/instrumentation/azure-sdk.md
lmolkova: It's a semi-official name for modern SDKs,  Azure SDK users should be familiar with it (or can easily find out what it means).

semantic_conventions/trace/instrumentation/azure-sdk.yml
```diff
@@ -0,0 +1,117 @@
...
+        examples: ['Microsoft.Storage', 'Microsoft.KeyVault', 'Microsoft.ServiceBus']
+
+  # public API
+  - id: azure.sdk.api
+    span_kind: internal
+    extends: azure
+    brief: 'Describes Azure SDK API calls that wrap an Azure service call(s).'
+
+  # http
+  - id: azure.sdk.http
```
anuraaga, line 25: Just confirming if you meant to remove `.sdk` in all the cases too, assuming a possibility of filling in on the server.

lmolkova: this specific section describes the client side of things (e.g. populates `http.url`), if we add server, I assume it would go under `azure.<server>.http`, so I kept `sdk` intentionally. Do you see any problem with this? 

I assume if we add common attributes between server and client and declare them in `azure.http` this would still be a backward-compatible change.

semantic_conventions/trace/instrumentation/azure-sdk.yml
arminru: I think it would make sense to agree on the format and content first and have that specified in the `syntax.md` for the tool (but it looks good to me in general).

semantic_conventions/trace/instrumentation/azure-sdk.yml
arminru: That is correct. Currently you have `request_id` and `service_request_id` as standalone top-level attributes though. I assume those should be put under an `azure` prefix / namespace, or is this intentional?

semantic_conventions/trace/instrumentation/azure-sdk.yml
arminru: Ah so that's not a new suggestion that just borrows from OpenTracing semconvs but it's documenting an existing specification?

semantic_conventions/trace/instrumentation/azure-sdk.yml
Oberon00: For HTTP, we chose the same names for client and server, since the overlapping attributes (like http.url) do have the same meaning. We use the span kind to distinguish.

PR #2235, opened 2021-12-21
JSON serialization format

### LINKED ISSUE #1443: Define how OTLP data can be stored in a file
OTLP currently defines how telemetry can be serialized and transferred over the network.

There are use cases that require writing telemetry to a file and then reading it back, e.g. storing test data or ability to transfer telemetry from DMZ that only allows file transfers: https://github.com/open-telemetry/opentelemetry-collector/issues/2450

We need to define how OTLP data can be serialized and stored in a file. We should allow to store multiple batches of telemetry, then read it back. All of the telemetry signals should be supported. The file format should be optimized for append scenarios and for sequential reading scenario. 

The stored telemetry should contain metadata that describe what signal type it is. It should be possible to store telemetry of mixed signals in one file unambiguously.

Optional requirements: 
- Ability for random access of records.
- Compression of stored records.
- Support for both binary (e.g. Protobuf) and human-readable formats (e.g. JSON).

Note that experimental file exporter exists in the Collector: https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter/fileexporter that outputs records in OTLP/JSON format, however it lacks the metadata that tells the signal type and thus cannot be used in mixed signal scenarios.

atoulme:
Fixes #1443

## Changes

Adds a new experimental JSON serialization minimal format to store traces, metrics and logs as JSON on disk.


experimental/serialization/json.md
```diff
@@ -0,0 +1,75 @@
...
+
+Data is appended sequentially to the file, irrespective of the type of data or timestamp.
+
+#### Compression
+
+The file may be stored as uncompressed. Implementors are free to compress the file using gzip or tar compression algorithm as they see fit.
+
+## Example
+
+```json lines
```
tigrannajaryan, line 71: Would it be useful to show the BOM here?

tigrannajaryan: Can we have "rs", "rm" and "rl" in one json line or only one of the three is allowed?

experimental/serialization/json.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Telemetry data requirements
+
+#### Versioning
+
+Each object serialized to JSON should be identified with a version tagging the version of the serializer.
+
+The deserializer must support and maintain compatibility with versions explicitly supported by this specification.
+
+The version is serialized as a string under the key "v" of the JSON document.
```
tigrannajaryan, line 33: Given how verbose the rest of the keys in the format are should we use "version" for more clarity?

experimental/serialization/json.md
```diff
@@ -0,0 +1,75 @@
...
+{"v":"1.0.0.alpha","rs":{"resourceSpans":[{"resource":{"attributes":[{"key":"resource-attr","value":{"stringValue":"resource-attr-val-1"}}]},"instrumentationLibrarySpans":[{"instrumentationLibrary":{},"spans":[{"traceId":"","spanId":"","parentSpanId":"","name":"operationA","startTimeUnixNano":"1581452772000000321","endTimeUnixNano":"1581452773000000789","droppedAttributesCount":1,"events":[{"timeUnixNano":"1581452773000000123","name":"event-with-attr","attributes":[{"key":"span-event-attr","value":{"stringValue":"span-event-attr-val"}}],"droppedAttributesCount":2},{"timeUnixNano":"1581452773000000123","name":"event","droppedAttributesCount":2}],"droppedEventsCount":1,"status":{"deprecatedCode":"DEPRECATED_STATUS_CODE_UNKNOWN_ERROR","message":"status-cancelled","code":"STATUS_CODE_ERROR"}},{"traceId":"","spanId":"","parentSpanId":"","name":"operationB","startTimeUnixNano":"1581452772000000321","endTimeUnixNano":"1581452773000000789","links":[{"traceId":"","spanId":"","attributes":[{"key":"span-link-attr","value":{"stringValue":"span-link-attr-val"}}],"droppedAttributesCount":4},{"traceId":"","spanId":"","droppedAttributesCount":4}],"droppedLinksCount":3,"status":{}}]}]}]}}
```
tigrannajaryan, line 72: Is version number "1.0.0.alpha" coupled in any way to the OTLP Protobuf version number? What's the versioning strategy here? What are interoperability expectations between writers and readers? Should readers interpret the version in any way?

experimental/serialization/json.md
atoulme: I thought showing the BOM might confuse folks since it's just 3 bytes.
Only one of the type of resources should be allowed.

experimental/serialization/json.md
atoulme: ok with me

experimental/serialization/json.md
atoulme: The version number here is the version number of the serialization to json itself, irrespective and superseding the protobuf serialization. This means that it changes if the protobuf serialization changes, and it may also change if we change how we write the protobuf message. I would hope we'd keep tightly in line with the protobuf changes.

djaglowski (commented):
I like a lot about this design, in particular that it would provide a very accessible mechanism for implementing asynchronous workflows. It's probably never going to be the most elegant way to manage telemetry, but it's a great catchall.

There are a few minor points I would like to reconsider. 
- The shorthand keys ("v", "rs", "rm", "rl") clash with the preference for clarity demonstrated in telemetry objects themselves.
- Allowing only one of "rs", "rm", "rl" makes sense as a rule, but it would be nice if the data structure enforced this in a natural way.
- I would support a rule that each file should only contain one version and type of telemetry. Mixing versions or telemetry types could be problematic for a variety of reasons.
- The file is actually a [json lines](https://jsonlines.org/) file, so we might consider indicating this by suggesting the `.jsonl` file extension

I'm primarily interested in addressing these points. Probably they can be addressed with incremental changes, or not at all if that's the consensus. 

---

I will also offer a less conventional solution to the above, which we _could_ consider:
- Establish a new file extension, perhaps `.jsont` meaning "json telemetry"
- The `.jsont` file format has the following expectations:
  - Is a valid `.jsonl` file
  - The first line is a special metadata object that specifies the telemetry type and OTLP Protobuf version
  - All subsequent lines are of the format specified in the first (i.e. json objects at the `ResourceSpans | ResourceMetrics | ResourceLogs` level)
- Example `.jsont` file:
  ```json lines
  {"otlp_protobuf_version": "0.0.0", "signal_type": "metrics"}
  {"resourceMetrics":[{"resource":{"attributes":[{"key":"resource-attr","value":{"stringValue":"resource-attr-val-1"}}]},"instrumentationLibraryMetrics":[{"instrumentationLibrary":{},"metrics":[...]}]}
  {"resourceMetrics":[{"resource":{"attributes":[{"key":"resource-attr","value":{"stringValue":"resource-attr-val-2"}}]},"instrumentationLibraryMetrics":[{"instrumentationLibrary":{},"metrics":[...]}]}
  ```

atoulme:
Excellent suggestions! I'll get to work incorporating them. Thank you so much!

experimental/serialization/json.md
```diff
@@ -0,0 +1,75 @@
...
+    - [UTF-8 preamble](#utf-8-preamble)
+    - [File extension](#file-extension)
+    - [Separation between objects](#separation-between-objects)
+    - [Sequential append](#sequential-append)
+    - [Compression](#compression)
+- [Example](#example)
+
+## Overview
+
+This document describes the serialization of OpenTelemetry data as JSON objects, and their serialization to files.
```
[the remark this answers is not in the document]
atoulme: I put some words towards that. Please let me know if that works.

experimental/serialization/json.md
```diff
@@ -0,0 +1,72 @@
...
+
+#### Streaming appending
+
+Data is appended sequentially to the file, irrespective of the type of data or timestamp.
+
+This allows streaming data to the file.
+
+#### Compression
+
+The file may be stored as uncompressed. Implementors are free to compress the file using gzip or tar compression algorithm as they see fit.
```
Aneurysm9, line 64:
```suggestion
The file may be stored as uncompressed. Implementors are free to compress the file as they see fit.
```

`tar` is not a compression algorithm and there's probably no need to specify which algorithms are usable if we're talking about files at rest.

experimental/serialization/json.md
```diff
@@ -0,0 +1,72 @@
...
+    - [JSON encoding](#json-encoding)
+  - [File storage requirements](#file-storage-requirements)
+    - [JSON lines file](#json-lines-file)
+    - [Streaming appending](#streaming-appending)
+    - [Compression](#compression)
+- [Example](#example)
+
+## Overview
+
+This document describes the serialization of OpenTelemetry data as JSON objects, and their serialization to files.
```
Aneurysm9, line 19:
```suggestion
This document describes the serialization of OpenTelemetry data as JSON objects that can be stored in files.
```

experimental/serialization/json.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Requirements
+
+### Telemetry data requirements
+
+#### Versioning
+
+Each object serialized to JSON should be identified with a version tagging the version of the serializer.
```
Aneurysm9, line 27: Should this be the version of the serialization scheme?  Does the version of the serialization implementation matter?

experimental/serialization/json.md
```diff
@@ -0,0 +1,72 @@
...
+
+The deserializer must support and maintain compatibility with versions explicitly supported by this specification.
```
Aneurysm9, line 29: Does this specification define any versions of the serialization scheme?

experimental/serialization/json.md
```diff
@@ -0,0 +1,72 @@
...
+
+The version is serialized as a string under the key "version" of the JSON document.
+
+#### Resource level
+
+The JSON serialization operates at the level of `ResourceSpans`, `ResourceMetrics` and `ResourceLogs`.
+
+There is no support for serializing objects at a lower level.
```
Aneurysm9, line 37: There must be support for serializing types further down the hierarchy or these types could not be serialized.  Do you mean to say that these are the only valid types at the top level of a serialized JSON object that conforms to this spec?

atoulme: Yes indeed, that was brought up in discussions and I wanted to make it more clear.

experimental/serialization/json.md
atoulme: Yes, they're one and the same. Sorry for the confusion.

experimental/serialization/json.md
atoulme: It's implicit and it's good you're asking this - it should be very explicit this spec defines version 1.

experimental/serialization/json.md
atoulme: I have attempted to clarify, please review.

experimental/serialization/json.md
atoulme: I have attempted to clarify by creating a v1 header, please review

experimental/serialization/json.md
atoulme: I have removed this sentence for now and clarified the top level resource objects.

experimental/serialization/json.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Serialization schema v1
+
+This defines the first version of the serialization scheme.
+
+#### Resource level
+
+The JSON serialization accept top-level resource objects: `ResourceSpans`, `ResourceMetrics` and `ResourceLogs`.
+
+Resources are serialized under the key "traces", "metrics" and "logs" respectively for `ResourceSpans`, `ResourceMetrics` and `ResourceLogs`.
```
pmm-sumo, line 41: Perhaps instead of `traces`, `metrics`, `logs` those could be [`TracesData`](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto#L37), [`MetricsData`](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L37) and [`LogsData`](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/logs/v1/logs.proto#L37) respectively, for keeping consistency with proto?

experimental/serialization/json.md
```diff
@@ -0,0 +1,74 @@
...
+
+#### Streaming appending
+
+Data is appended sequentially to the file, irrespective of the type of data or timestamp.
+
+This allows streaming data to the file.
+
+#### Compression
+
+The file may be stored as uncompressed. Implementors are free to compress the file as they see fit.
```
reyang, line 66: What does this mean? e.g. how would developer A consume the file from developer B if B chose to compress the file in an unknown format?

experimental/serialization/json.md
```diff
@@ -0,0 +1,74 @@
...
+{"version":"1.0.0.alpha","metrics":{"resourceMetrics":[{"resource":{"attributes":[{"key":"resource-attr","value":{"stringValue":"resource-attr-val-1"}}]},"instrumentationLibraryMetrics":[{"instrumentationLibrary":{},"metrics":[{"name":"counter-int","unit":"1","sum":{"dataPoints":[{"attributes":[{"key":"label-1","value":{"stringValue":"label-value-1"}}],"startTimeUnixNano":"1581452772000000321","timeUnixNano":"1581452773000000789","asInt":"123"},{"attributes":[{"key":"label-2","value":{"stringValue":"label-value-2"}}],"startTimeUnixNano":"1581452772000000321","timeUnixNano":"1581452773000000789","asInt":"456"}],"aggregationTemporality":"AGGREGATION_TEMPORALITY_CUMULATIVE","isMonotonic":true}},{"name":"counter-int","unit":"1","sum":{"dataPoints":[{"attributes":[{"key":"label-1","value":{"stringValue":"label-value-1"}}],"startTimeUnixNano":"1581452772000000321","timeUnixNano":"1581452773000000789","asInt":"123"},{"attributes":[{"key":"label-2","value":{"stringValue":"label-value-2"}}],"startTimeUnixNano":"1581452772000000321","timeUnixNano":"1581452773000000789","asInt":"456"}],"aggregationTemporality":"AGGREGATION_TEMPORALITY_CUMULATIVE","isMonotonic":true}}]}]}]}}
```
reyang, line 72: What does "atInt" mean?

experimental/serialization/json.md
atoulme: Less is more - removing compression.

experimental/serialization/json.md
atoulme: Sorry, I don't see a `atInt` in the code. Did you mean `asInt`?

experimental/serialization/json.md
```diff
@@ -0,0 +1,74 @@
...
+
+The data should be encoded according to the format specified in the [OTLP/HTTP standard](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp).
+
+### File storage requirements
+
+#### JSON lines file
+
+This file is a [JSON lines file](https://jsonlines.org/), and therefore follows those requirements:
+
+* UTF-8 encoding
```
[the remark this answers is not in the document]
atoulme: It is unspecified.

experimental/serialization/json.md
atoulme: OK, sure thing. Changed.

pmm-sumo: Great, thank you!

pmm-sumo (approved):

djaglowski (commented):
> I would support a rule that each file should only contain one version and type of telemetry. Mixing versions or telemetry types could be problematic for a variety of reasons.

I want to reiterate this and expand upon it.

I'm concerned that this specification allows for mixing of telemetry types in a single file. In theory this is fine, since each line independently identifies the type of telemetry. However, I question whether a mixed file can be consumed easily. 

For example, consider the OpenTelemetry collector which presumably be a common consumer of these files. The collector allows only a single signal type per pipeline. If we were to write a receiver to ingest a these files, each instance of that receiver would be associated with a single signal type. This is awkward at best.

I also wonder whether producers of telemetry would even want to write multiple signal types to a single file. Do we even care to support this use case?

experimental/serialization/json.md
atoulme: I am going to resolve this comment, `asInt` here is an element of the JSON serialization of the data point per the serialization framework set by OLTP/JSON as referred in the spec.

atoulme:
No problem. Let's have that spelled out in the spec that one file == one type of telemetry.
However, I'd rather keep it so that each record has a version and telemetry type for simplicity sake. If we have to deal with a preamble, it will make rotating files harder. It won't be as easy to stream to a file by just adding one more element to it. Does that make sense?

djaglowski:
> No problem. Let's have that spelled out in the spec that one file == one type of telemetry. However, I'd rather keep it so that each record has a version and telemetry type for simplicity sake. If we have to deal with a preamble, it will make rotating files harder. It won't be as easy to stream to a file by just adding one more element to it. Does that make sense?

I agree the preamble suggestion is not ideal. 

I'd like to take a step back and point out that regardless of whether we use a preamble or a wrapper around each line, the two options both presume that we need additional metadata (i.e. required only by this specification) to be encoded in the file. Before we commit to this format, I'd like to clearly establish whether or not we really need these metadata, because _if not_ the specification becomes much simpler. 

As proposed, we have two pieces of metadata: 
1. signal type
2. version of this specification

If we've established that each file will only contain one signal type, then we could consider these as per-singal JSON serialization formats. In a very general sense, the structure of a json file is typically understood from context. Would it be unreasonable to expect that these files will be managed in such a way where context is not lost? (e.g. The user can read/write each signal type in different directories.)

Do we really need to version this specification? Notable to me is that [OTLP/HTTP](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp) is calling out a stability level specifically for the "JSON Format Status" that this specification relies upon. Presumably, once that format is marked Stable, this specification can likely follow quickly. So instead of managing a separate version for this specification, could we just call it Experimental and tolerate a few potential breaking changes in the near future?

If we don't need either or these metadata, then likely we do not need either a preamble or a per-line wrapper. This leaves us with:
- JSON Lines files that contain `ResourceSpans`
- JSON Lines files that contain `ResourceMetrics`
- JSON Lines files that contain `ResourceLogs`
- Maybe best practices for BOM, encoding, etc?

atoulme:
Simpler is better. I put up a version to give us some room in case things change. But I'm ok with this simpler format, until it doesn't work anymore.

atoulme:
I'm going to wait 24h - if I don't hear back, I'm going to edit the PR to reflect Dan's thinking.

pmm-sumo:
I am with Dan on the simpler approach. Also, if the version would be kept, I think it should be mapped to [OTLP version](https://github.com/open-telemetry/opentelemetry-proto/releases) (rather than introducing separate version sequence). If it's really needed, perhaps it could be included as a separate line whenever the version changes (e.g. `{"version": "0.12.0"}\n`) followed by any number of lines with actual records (or another version change).

In either case, I think it will be simpler to just leave out the version, at least for now

pmm-sumo (changes requested):
I am with Dan on the simpler approach. Also, if the version would be kept, I think it should be mapped to [OTLP version](https://github.com/open-telemetry/opentelemetry-proto/releases) (rather than introducing separate version sequence). If it's really needed, perhaps it could be included as a separate line whenever the version changes (e.g. `{"version": "0.12.0"}\n`) followed by any number of lines with actual records (or another version change).

In either case, I think it will be simpler to just leave out the version, at least for now

BTW, there are some issues with the links that docfx fails with

atoulme:
OK, we're 24h later, and it looks like we have rough consensus to simplify further. @pmm-sumo, appreciate your comment. Would it be possible to have the version be part of the OTLP JSON serialization? It might be better upstream from this work, given what I hear.

I am going to update the spec and fix the docfx issues.

pmm-sumo:
One more thought on the version. If there are some non-breaking changes, then it does not really matter (as long as the older version does not fail with new fields added or removed later). 

If the changes are breaking, then whatever reads the JSON, it needs to be able to deserialize using both the old and new schema, which also complicates a few other things (like having two proto versions and being able to do conversions). This might happen at some point, but at the same time I think that we won't have to introduce version information until breaking changes will become present. 

Again - this is in line with @djaglowski suggestion to omit version in the current version of the specification

experimental/serialization/json.md
```diff
@@ -0,0 +1,75 @@
...
+
+This document describes the serialization of OpenTelemetry data as JSON objects that can be stored in files.
+
+## Requirements
+
+### File storage requirements
+
+#### JSON lines file
+
+This file is a [JSON lines file](https://jsonlines.org/), and therefore follows those requirements:
```
aabmass, line 25: I am wondering if this spec would apply to "files" such as named pipes or stdout/stdin. If an application serializes OTLP in this format to stdout, it would need to be mixed signal.

aabmass (commented):
> I'm concerned that this specification allows for mixing of telemetry types in a single file. In theory this is fine, since each line independently identifies the type of telemetry. However, I question whether a mixed file can be consumed easily.

@djaglowski can you expand on what makes it difficult to consume a mixed file?

> I agree, with the current collector architecture, where for each pipeline there is actually a separate instance of receiver or exporter, it would be rather complex to output/feed all signals to/from a single file.

@pmm-sumo I didn't follow this either. My understanding is that you may create multiple pipelines using the same receivers (fan-out) and exporters (fan-in), and exporters/receivers may support multiple signal types.

pmm-sumo:
> > I agree, with the current collector architecture, where for each pipeline there is actually a separate instance of receiver or exporter, it would be rather complex to output/feed all signals to/from a single file.
> 
> @pmm-sumo I didn't follow this either. My understanding is that you may create multiple pipelines using the same receivers (fan-out) and exporters (fan-in), and exporters/receivers may support multiple signal types.

I realized I was mistaken. It stuck in my head that processors have single instance per pipeline, but of course this is not true for exporters and receivers which can be shared components (not that there's a lot of them, I think currently it's only otlpreceiver, opencensusreceiver and... fileexporter). So keeping single file might be not optimal if pipelines would be processed at different speeds, but then I am not sure if it should prevent us from keeping a single file in the end. I am curious to hear from @djaglowski, maybe there are other arguments against single file

experimental/serialization/json.md
atoulme: sure thing - how would you phrase it?

aabmass: I have no issue with the phrasing in this section since it's specific to file storage. It would be good to clarify elsewhere if this intended as a general streaming serialization format in addition to file storage.

But I am worried we will need mixed signals in a single stream to accomplish shell pipelining or to capture a program's telemetry from stdout.

djaglowski:
> My understanding is that you may create multiple pipelines using the same receivers (fan-out) and exporters (fan-in), and exporters/receivers may support multiple signal types.

Thanks for calling this out. I thought that a single receiver instance could only be shared across pipelines of the same signal type, but I've confirmed you are correct.

I also think your comment [here](https://github.com/open-telemetry/opentelemetry-specification/pull/2235#discussion_r803078473) makes a reasonable case that it is in fact necessary to support mixed signals.

If others agree with this, then I think we're back to having a wrapper around each line.

experimental/serialization/json.md
atoulme: I think I'd rather avoid a streaming discussion right now. Small and incremental scope, rather than trying to scope all of it out. If folks want streaming, there is OLTP support over HTTP and gRPC.

atoulme:
We might get away without a wrapper per line if we agree that they start with `ResourceLogs`, `ResourceMetrics` or `ResourceSpans`.

experimental/serialization/json.md
```diff
@@ -0,0 +1,75 @@
...
+
+This allows streaming data to the file.
+
+### Telemetry data requirements
+
+This defines the first version of the serialization scheme.
+
+The data must be encoded according to the format specified in the [OTLP/HTTP standard](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp).
+
+Only top-level objects, `ResourceSpans`, `ResourceMetrics`, and `ResourceLogs` are supported.
```
aabmass, line 44: In the example below, the actual lines are `Export(Trace|Metrics|Logs)ServiceRequest` objects.

atoulme: I'm sorry, I'm confused and do not understand what you mean.

atoulme:
I have opened https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/7840 to showcase a file receiver that follows the proposed specification. Please review and let me know if this sounds like a good approach.

experimental/serialization/json.md
aabmass: Sorry for the confusion. This json below is an [`ExportTraceServiceRequest`](https://github.com/open-telemetry/opentelemetry-proto/blob/6a41aec384e8c79a1bf92d0105fc0f5a2f864f3f/opentelemetry/proto/collector/trace/v1/trace_service.proto#L35) message:

```json
{"resourceSpans":[{"resource":{"attributes":[{"key":"resource-attr","value":{"stringValue":"resource-attr-val-1"}}]},"instrumentationLibrarySpans":[{"instrumentationLibrary":{},"spans":[{"traceId":"","spanId":"","parentSpanId":"","name":"operationA","startTimeUnixNano":"1581452772000000321","endTimeUnixNano":"1581452773000000789","droppedAttributesCount":1,"events":[{"timeUnixNano":"1581452773000000123","name":"event-with-attr","attributes":[{"key":"span-event-attr","value":{"stringValue":"span-event-attr-val"}}],"droppedAttributesCount":2},{"timeUnixNano":"1581452773000000123","name":"event","droppedAttributesCount":2}],"droppedEventsCount":1,"status":{"deprecatedCode":"DEPRECATED_STATUS_CODE_UNKNOWN_ERROR","message":"status-cancelled","code":"STATUS_CODE_ERROR"}},{"traceId":"","spanId":"","parentSpanId":"","name":"operationB","startTimeUnixNano":"1581452772000000321","endTimeUnixNano":"1581452773000000789","links":[{"traceId":"","spanId":"","attributes":[{"key":"span-link-attr","value":{"stringValue":"span-link-attr-val"}}],"droppedAttributesCount":4},{"traceId":"","spanId":"","droppedAttributesCount":1}],"droppedLinksCount":3,"status":{}}]}]}]}
```

The [OTLP/HTTP spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp-request) mentioned in this PR for traces says:

> The request body is a Protobuf-encoded `ExportTraceServiceRequest` message.

In which case I think this line should say:
```suggestion
Only top-level objects, `ExportTraceServiceRequest`, `ExportMetricsServiceRequest `, and `ExportLogsServiceRequest` are supported.
```

aabmass: If you want to divorce this spec from the RPC part of the protocol, there is also an equivalent [`TracesData`](https://github.com/open-telemetry/opentelemetry-proto/blob/6a41aec384e8c79a1bf92d0105fc0f5a2f864f3f/opentelemetry/proto/trace/v1/trace.proto#L37) message can you use here instead.

pmm-sumo:
> The only awkward thing is you can't use a `oneof` because of the `repeated` option.

perhaps `oneof` + `TracesData`/`MetricsData`/`LogsData`?

atoulme:
> what about oneof + TracesData/MetricsData/LogsData?
Can you mix traces, logs and metrics then?

aabmass:
> @aabmass what about `oneof` + `TracesData`/`MetricsData`/`LogsData`?

if you did it with oneofs, the fields can't be repeated

```proto
message OpenTelemetryData {
    oneof data {
        ResourceMetrics resourceMetrics = 1;
        ResourceLogs resourceLogs = 2;
        ResourceSpans resourceSpans= 3;
    }
}
```

so then the json would have to look like `{"resourceSpans": {...}}` instead of `{"resourceSpans":[{...}, ...]}` which is incompatible with this proposal and the current fileexporter IIUC.

atoulme:
OK, how about we pull it back a bit in scope here. Let's decouple this spec from proto changes, so we can then open a bug for the proto changes based on what we learned here, and then iterate again on this spec. Does that sound like a plan? If yes, I'm committing the suggestion here: https://github.com/open-telemetry/opentelemetry-specification/pull/2235#discussion_r804974293 and then we can get this merged, granted that I somehow make CI pass.

aabmass:
@atoulme sounds good, we can tackle the proto design later 👍

I do think mixed signals for a given file/stream of json lines should be allowed though, unless someone can think of some downsides.

atoulme:
Full agreement on mixed signal and there’s an example showing that now.

experimental/serialization/json.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Telemetry data requirements
+
+This defines the first version of the serialization scheme.
+
+The data must be encoded according to the format specified in the [OTLP/HTTP standard](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp).
+
+Only top-level objects, `ExportTraceServiceRequest`, `ExportMetricsServiceRequest`, and `ExportLogsServiceRequest` are supported.
+
+Files must contain exactly one type of data: traces, metrics, or logs. Mixed content is not supported.
```
aabmass, line 44: This line is stale, otherwise LGTM

aabmass (approved):

experimental/serialization/json.md
atoulme: oh sorry. Fixing.

experimental/serialization/json.md
```diff
@@ -0,0 +1,82 @@
+# JSON serialization
+
```
tigrannajaryan, line 2: Add document status, similar to other spec docs.

experimental/serialization/json.md
```diff
@@ -0,0 +1,82 @@
+# JSON serialization
```
tigrannajaryan, line 1: This is specifically for file representation, not any JSON (we have OTLP/JSON which is different), so perhaps:
```suggestion
# JSON File serialization
```

experimental/serialization/json.md
```diff
@@ -0,0 +1,82 @@
...
+This file is a [JSON lines file](https://jsonlines.org/), and therefore follows those requirements:
+
+* UTF-8 encoding
+* Each line is a valid JSON value
+* The line separator is `\n`
+* The preferred file extension is `jsonl`.
+
+#### Streaming appending
+
+Data is appended sequentially to the file, irrespective of the timestamp.
```
tigrannajaryan, line 32: I think "appended sequentially" should not be a requirement. What if I have an implementation that can update a record in place?
If I understand correctly the intent here is that there is no guarantee that the timestamps in the file are monotonically increasing, so perhaps say that directly.

tigrannajaryan (approved):

pmm-sumo:
> > @aabmass what about `oneof` + `TracesData`/`MetricsData`/`LogsData`?
> 
> if you did it with oneofs, the fields can't be repeated
> 
> ```proto
> message OpenTelemetryData {
>     oneof data {
>         ResourceMetrics resourceMetrics = 1;
>         ResourceLogs resourceLogs = 2;
>         ResourceSpans resourceSpans= 3;
>     }
> }
> ```
> 
> so then the json would have to look like `{"resourceSpans": {...}}` instead of `{"resourceSpans":[{...}, ...]}` which is incompatible with this proposal and the current fileexporter IIUC.

@aabmass I meant something slightly different, i.e.:

```proto
message OpenTelemetryData {
    oneof data {
        MetricsData metricsData = 1;
        LogsData logsData = 2;
        TracesData tracesData = 3;
    }
}
```

Would it still not work (since MetricsData/LogsData/TracesData is just a [repeated field underneath](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto#L43))? I think this would effect in something like `{"tracesData": [{"resource": ..., "instrumentation_library_spans": ...}, ...]}`

experimental/serialization/json.md
djaglowski: I think this is intending to say that timestamps do _not_ need to be monotonically increasing.

The question of whether the file must be append-only is interesting. Strictly speaking, the file _format_ does not necessarily need to be concerned with _how_ the file is written and read. However, I think it's worth considering that append-only files are going to be much more streamable, in the sense that the reader and writer can operate independently like a log file writer and reader. Without this requirement, tracking file rotation and updates becomes far more difficult.

experimental/serialization/json.md
atoulme: Sure thing, done.

experimental/serialization/json.md
atoulme: OK fixed.

aabmass:
> Would it still not work (since MetricsData/LogsData/TracesData is just a [repeated field underneath](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto#L43))? I think this would effect in something like `{"tracesData": [{"resource": ..., "instrumentationLibrarySpans": ...}, ...]}`

It would work but I think the json would look like this: `{"tracesData": {"resourceSpans: [{"resource": ..., "instrumentationLibrarySpans": ...}]}}`. Maybe we should table the oneof discussion for now and open a follow up issue for parsing jsonlines as a single protobuf message

> I might be missing something but my understanding is that TracesData/etc. were [introduced](https://github.com/open-telemetry/opentelemetry-proto/pull/332) for usecases like here and are a better fit than ExportTraceServiceRequest/etc.

I agree those might be a better fit. This PR currently references [OTLP/HTTP standard](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp) which is specific to use the the `Export*ServiceRequest` messages. @pmm-sumo do you want this PR be updated to say `TracesData` etc. instead?

pmm-sumo:
> I agree those might be a better fit. This PR currently references [OTLP/HTTP standard](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp) which is specific to use the the `Export*ServiceRequest` messages. @pmm-sumo do you want this PR be updated to say `TracesData` etc. instead?

I guess the reason for referencing OTLP/HTTP is because it is the only place in specification where JSON encoding is mentioned in context of OTLP. However, this PR is actually about proposing how to serialize OTLP using JSON on disk, so I believe it does not have to stick to the notion of being strictly based on OTLP/HTTP, including ExportTracesServiceRequest and such

As for TracesData/etc. - I think it would be good to hear from @bogdandrutu who introduced it into the proto. That would be perhaps first mention of it in the specification, though I believe it would be inline with the original idea behind it

tigrannajaryan:
> I guess the reason for referencing OTLP/HTTP is because it is the only place in specification where JSON encoding is mentioned in context of OTLP. However, this PR is actually about proposing how to serialize OTLP using JSON on disk, so I believe it does not have to stick to the notion of being strictly based on OTLP/HTTP, including ExportTracesServiceRequest and such

If we don't reference OTLP/HTTP spec then at least we need to duplicate the wording we have there about JSON encoding. It has specifics which are important:

>JSON-encoded Protobuf payloads use proto3 standard defined [JSON Mapping](https://developers.google.com/protocol-buffers/docs/proto3#json) for mapping between Protobuf and JSON, with one deviation from that mapping: the trace_id and span_id byte arrays are represented as [case-insensitive hex-encoded strings](https://tools.ietf.org/html/rfc4648#section-8), they are not base64-encoded like it is defined in the standard [JSON Mapping](https://developers.google.com/protocol-buffers/docs/proto3#json). The hex encoding is used for trace_id and span_id fields in all OTLP Protobuf messages, e.g. the Span, Link, LogRecord, etc. messages.

>Note that according to [Protobuf specs](https://developers.google.com/protocol-buffers/docs/proto3#json) 64-bit integer numbers in JSON-encoded payloads are encoded as decimal strings, and either numbers or strings are accepted when decoding.

atoulme:
Great discussion on the call this morning. I am going to file issues to capture discussion on this issue and make sure we continue the work after this PR, after which it should be good to go.

atoulme:
@pmm-sumo I have filed #2390 to follow up on your comment.

tigrannajaryan:
> We should be good then, but if you like I can file an issue to ask that we carve out OTLP/HTTP even more to make it super clear what we refer to.

One possible approach is to move the paragraph that described how JSON encoding works from OTLP/HTTP document in this repo to https://github.com/open-telemetry/opentelemetry-proto/tree/main/opentelemetry/proto which is responsible for encodings. Then we can link to that from OTLP/HTTP document and from the document that this PR introduces.

carlosalberto (approved):

atoulme:
Nice! Can we merge this too?

carlosalberto:
@tigrannajaryan Can you confirm the remaining items can be done as part of #2390? Sounds like a YES 😄

tigrannajaryan:
> @tigrannajaryan Can you confirm the remaining items can be done as part of #2390? Sounds like a YES 😄

Yes, I think it is fine. 

However, if possible I would like to see more approvals on this issue before we merge it.

djaglowski (approved):
LGTM

carlosalberto:
We got one more review, and this has been open for a while now. Merging.

Issue #2245, opened 2022-01-04
Specify whether keys in attribute lists are expected to be unique


achille-roussel:
Hello,

I'm opening this issue as a follow up to a discussion in https://github.com/open-telemetry/opentelemetry-proto/pull/346#issuecomment-1004270981 where it appeared unclear whether processors of telemetry data should be designed to support repeating keys in attribute lists.

I do not know what the right answer should be, but I couldn't find a definitive statement in the [Attributes Specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md) to help guide design decisions that would depend on this information.

There is one mention of _"unique keys"_ in the [Attributes Limits](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attribute-limits), but it is unclear whether _"unique"_ refers to keys being unique or the number of keys in the attribute lists:

> * set a limit of unique attribute keys such that:
>   * for each unique attribute key, addition of which would result in exceeding the limit, SDK MUST discard that key/value pair.

Let me know if I can provide any more context that could help address this issue.

bogdandrutu:
@tigrannajaryan do you remember this?

I remember for sure that we said we have a map "semantic" there https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/resource/v1/resource.proto#L28, if that is not the case we have huge problem in the collector :))

Oberon00:
I remember (actually I just saw it by accident when I opened my Thread list) we had a short Thread on Slack 4 months ago: https://cloud-native.slack.com/archives/C01N7PP1THC/p1630920628020600?thread_ts=1630790115.020300&cid=C01N7PP1THC

At that time I wrote:

> There is https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-attributes
>
>> Setting an attribute with the same key as an existing attribute SHOULD overwrite the existing attribute's value.
>
> Which currently only applies to Span.SetAttribute. IMHO we should move that to the overall attribute definition, otherwise it doesn't even apply to https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span-creation

Would not be surprised if it is still unspecified for Span Creation, Resources, Metric attributes (at least I did not follow up at that time).

PR #2248, opened 2022-01-06
Clarify that attribute keys are unique in collections

tigrannajaryan:
Attributes keys must be unique. The key/value pair collections in the specification was always intended to model a map.

There was a recent confusion about this. This change clarifies the spec. This is an editorial change. It does not change any semantics.

Resolves https://github.com/open-telemetry/opentelemetry-specification/issues/2245

### DIFF
--- a/specification/common/common.md
+++ b/specification/common/common.md
@@ -10,6 +10,7 @@
 - [Attribute](#attribute)
   * [Attribute Limits](#attribute-limits)
     + [Exempt Entities](#exempt-entities)
+- [Attribute Collections](#attribute-collections)
 
 <!-- tocstop -->
 
@@ -105,3 +106,14 @@ attribute limits for Resources.
 Attributes, which belong to Metrics, are exempt from the limits described above
 at this time, as discussed in
 [Metrics Attribute Limits](../metrics/sdk.md#attribute-limits).
+
+## Attribute Collections
+
+[Resources](../resource/sdk.md),
+Metrics [data points](../metrics/datamodel.md#metric-points),
+[Spans](../trace/api.md#set-attributes),
+Span [Events](../trace/api.md#add-events),
+Span [Links](../trace/api.md#specifying-links)
+and [Log Records](../logs/data-model.md) may contain a collection of attributes.
+The keys in each such collection are unique, i.e. there cannot exist more than
+one key-value pair with the same key.
--- a/specification/common/common.md
+++ b/specification/common/common.md
@@ -109,11 +109,32 @@ at this time, as discussed in
 
 ## Attribute Collections
 
-[Resources](../resource/sdk.md),
-Metrics [data points](../metrics/datamodel.md#metric-points),
-[Spans](../trace/api.md#set-attributes),
-Span [Events](../trace/api.md#add-events),
-Span [Links](../trace/api.md#specifying-links)
-and [Log Records](../logs/data-model.md) may contain a collection of attributes.
-The keys in each such collection are unique, i.e. there cannot exist more than
-one key-value pair with the same key.
+[Resources](../resource/sdk.md), Metrics
+[data points](../metrics/datamodel.md#metric-points),
+[Spans](../trace/api.md#set-attributes), Span
+[Events](../trace/api.md#add-events), Span
+[Links](../trace/api.md#specifying-links) and
+[Log Records](../logs/data-model.md) may contain a collection of attributes. The
+keys in each such collection are unique, i.e. there MUST NOT exist more than one
+key-value pair with the same key. The enforcement of uniqueness may be performed
+in a variety of ways as it best fits the limitations of the particular
+implementation.
+
+Normally for the telemetry generated using OpenTelemetry SDKs the attribute
+key-value pairs are set via an API that either accepts a single key-value pair
+or a collection of key-value pairs. Setting an attribute with the same key as an
+existing attribute SHOULD overwrite the existing attribute's value. See for
+example Span's [SetAttribute](../trace/api.md#set-attributes) API.
+
+A typical implementation of [SetAttribute](../trace/api.md#set-attributes) API
+will enforce the uniqueness by overwriting any existing attribute values pending
+to be exported, so that when the Span is eventually exported the exporters see
+only unique attributes. The OTLP format in particular requires that exported
+Resources, Spans, Metric data points and Log Records contain only unique
+attributes.
+
+Some other implementations may use a streaming approach where every
+[SetAttribute](../trace/api.md#set-attributes) API call immediately results in
+that individual attribute value being exported using a streaming wire protocol.
+In such cases the enforcement of uniqueness will likely be the responsibility of
+the recipient of this data.


yurishkuro:
Who is supposed to be responsible for the uniqueness?  We cannot impose requirements on the users of the API who can call setAttribute twice with the same key. And streaming implementations of the API cannot enforce key uniqueness (e.g. via last-wins) because they do not hold state.

tigrannajaryan:
> Who is supposed to be responsible for the uniqueness? We cannot impose requirements on the users of the API who can call setAttribute twice with the same key. And streaming implementations of the API cannot enforce key uniqueness (e.g. via last-wins) because they do not hold state.

I agree, users should't be required to do that. The Trace API [already enforces](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-attributes) the behavior:

>Setting an attribute with the same key as an existing attribute SHOULD overwrite the existing attribute's value.

My understanding of this clause is that the uniqueness is enforced (perhaps eventually, but nevertheless). Streaming implementations don't necessarily need to hold state on the sender side to enforce this. They can define the semantics of the receiving side rules in a manner that results uniqueness of attributes from receiver's perspective (the "overwrite if exists" rule can be applied at the receiver end).

I was not able to find an equivalent clause in the Metrics API. It should be probably added.

specification/common/common.md
```diff
@@ -105,3 +106,14 @@ attribute limits for Resources.
...
 [Metrics Attribute Limits](../metrics/sdk.md#attribute-limits).
+
+## Attribute Collections
+
+[Resources](../resource/sdk.md), Metrics
+[data points](../metrics/datamodel.md#metric-points),
+[Spans](../trace/api.md#set-attributes), Span
+[Events](../trace/api.md#add-events), Span
+[Links](../trace/api.md#specifying-links) and
+[Log Records](../logs/data-model.md) may contain a collection of attributes. The
```
yurishkuro, line 117:
```suggestion
[Resources](../resource/sdk.md), 
Metrics [data points](../metrics/datamodel.md#metric-points),
[Spans](../trace/api.md#set-attributes), 
Span [Events](../trace/api.md#add-events), 
Span [Links](../trace/api.md#specifying-links) 
and [Log Records](../logs/data-model.md)
may contain a collection of attributes. The
```
this formatting makes this more readable

tigrannajaryan: Done.

specification/common/common.md
```diff
@@ -105,3 +106,14 @@ attribute limits for Resources.
...
+## Attribute Collections
+
+[Resources](../resource/sdk.md),
+Metrics [data points](../metrics/datamodel.md#metric-points),
+[Spans](../trace/api.md#set-attributes),
+Span [Events](../trace/api.md#add-events),
+Span [Links](../trace/api.md#specifying-links)
+and [Log Records](../logs/data-model.md) may contain a collection of attributes.
+The keys in each such collection are unique, i.e. there cannot exist more than
+one key-value pair with the same key.
```
carlosalberto, line 119: Probably adding a link here to https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-attributes would keep clear what part is responsible for enforcing it.

tigrannajaryan: Yes, line 114 links exactly to that.

iNikem (approved):

carlosalberto (approved):

bogdandrutu (approved):

pirgeo (approved):

specification/common/common.md
Oberon00, line 119: cannot
```suggestion
The keys in each such collection are unique, i.e. there MUST NOT exist more than
```

And we probably should specify that conflicts SHOULD be resolved by using the attribute set latest.

specification/common/common.md
```diff
@@ -105,3 +106,14 @@ attribute limits for Resources.
...
+
+## Attribute Collections
+
+[Resources](../resource/sdk.md),
+Metrics [data points](../metrics/datamodel.md#metric-points),
+[Spans](../trace/api.md#set-attributes),
+Span [Events](../trace/api.md#add-events),
+Span [Links](../trace/api.md#specifying-links)
+and [Log Records](../logs/data-model.md) may contain a collection of attributes.
+The keys in each such collection are unique, i.e. there cannot exist more than
```
Oberon00, line 118: cannot
```suggestion
The keys in each such collection are unique, i.e. there MUST NOT exist more than
```

And we probably should specify that conflicts SHOULD be resolved by using the attribute set latest.

carlosalberto:
@yurishkuro Please take a look again.

yurishkuro:
As I already [commented](https://github.com/open-telemetry/opentelemetry-specification/pull/2248#issuecomment-1006739642), I do not understand what this description refers to. **Something** needs to enforce this uniqueness. It can be users of the API, the implementation of the API (i.e. SDK), or the backend. OTel cannot impose requirements on the users or the backends, and SDKs may not be able to enforce uniqueness (e.g. streaming API). So what is this clause referring to?

tigrannajaryan:
> As I already [commented](https://github.com/open-telemetry/opentelemetry-specification/pull/2248#issuecomment-1006739642), I do not understand what this description refers to. **Something** needs to enforce this uniqueness. It can be users of the API, the implementation of the API (i.e. SDK), or the backend.

@yurishkuro My reading of the [current requirement](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-attributes) is that it is the responsibility of the SDK or of the exporter to enforce uniqueness. 

If it is not clear we should clarify it, but I think it can be a separate PR. This PR does not contradict that, this PR merely sets the broader expectation from the semantics perspective.

yurishkuro:
The current req says "SHOULD overwrite", so uniqueness is not guaranteed. And that's my concern with the wording in this PR is that it gives a misleading impression that it *is* guaranteed / enforced.

tigrannajaryan:
> The current req says "SHOULD overwrite", so uniqueness is not guaranteed. And that's my concern with the wording in this PR is that it gives a misleading impression that it _is_ guaranteed / enforced.

That's a good point. I think the only reason it is a SHOULD and not a MUST is because an alternate strategy is possible (drop duplicates), not because duplicates are considered acceptable. I am fine with changing it to MUST or calling our more clearly the uniqueness requirement (or just linking to the wording introduced by this PR).

Again, I think this PR is more fundamental and it reflects our intent. If the existing spec wording contradicts with this PR then that wording needs to change not our guarantee of uniqueness which this PR merely states explicitly.

yurishkuro:
> Again, I think this PR is more fundamental and it reflects our intent. 

I guess this is where we disagree. First, there is no fundamental problems for the backends (at least tracing backends) to support multiple attributes with the same key (Jaeger always allowed that), and I see it as a function of you-get-back-what-you-put-in, i.e. the users are in control of uniqueness. Second, I think there's definitely room for strict maps, but only in situations with atomic writes to the API. E.g. adding an event to span is atomic, you can't amend the event afterwards, and that gives a clear choke point to enforce uniqueness. Logging API is similar (if we had one), metrics API probably same as well. However, the Span API allows multiple writes, so as long as the user can call `span.setAttribute(k, v)`, they should not expect some magic deduping by default (although it can be opt-in at some point in the pipeline).

Concrete example: Canopy at Meta supports `block.annotate(props)` API (block == span). Props is a map so uniqueness is enforced in this atomic write, but `annotate` can be called multiple times, possibly with conflicting keys. The SDK implementation is streaming, i.e. these writes are flushed independently, no state is stored in the block/span. The storage accepts all writes as is, without fiddling with the data. So the only point to enforce uniqueness (if desired) is in the trace reader when higher-level data model is returned inflated from the raw events.

yurishkuro:
> Again, I think this PR is more fundamental and it reflects our intent. 

I guess this is where we disagree. First, there is no fundamental problems for the backends (at least tracing backends) to support multiple attributes with the same key (Jaeger always allowed that), and I see it as a function of you-get-back-what-you-put-in, i.e. the users are in control of uniqueness. Second, I think there's definitely room for strict maps, but only in situations with atomic writes to the API. E.g. adding an event to span is atomic, you can't amend the event afterwards, and that gives a clear choke point to enforce uniqueness. Logging API is similar (if we had one), metrics API probably same as well. However, the Span API allows multiple writes, so as long as the user can call `span.setAttribute(k, v)`, they should not expect some magic de-duping by default (although it can be opt-in at some point in the pipeline).

Concrete example: Canopy at Meta supports `block.annotate(props)` API (block == span). Props is a map so uniqueness is enforced in this atomic write, but `annotate` can be called multiple times, possibly with conflicting keys. The SDK implementation is streaming, i.e. these writes are flushed independently, no state is stored in the block/span. The storage accepts all writes as is, without fiddling with the data. So the only point to enforce uniqueness (if desired) is in the trace reader when higher-level data model is returned inflated from the raw events.

specification/common/common.md
jmacd: Yes, I believe this was widely discussed and agreed to this point _years ago_, we just never wrote it down clearly. When an attribute map is expressed as a list, duplicate keys should be resolved by taking the last value among the ordered list. This improves efficiency since you can let the receiver de-duplicate.

jmacd (changes requested):

jmacd:
@yurishkuro I think you can have your interpretation inside a streaming SDK without changing the spec from Tigran's position. 

OTel added an array-of-strings attribute so that you could have a multi-valued attribute. If the application intends to report a multi-valued attribute, it should built the values in memory and set them as an attribute. If you want to update the attribute to a new set of values through the span's lifetime, that also works, but the set of attributes remains a map and your attribute has multiple string values. The application is required to keep its working set in memory if it wants to change the list of values for an attribute, since it has to update them all in a single call.

The Metrics data model defines more than one way to model data, and there is an analogy to draw with tracing. Metrics defines an event model, wherein each call to a synchronous instrument is an independently observable event. A streaming metric SDK could report each event as an individual (hey let's use column compression for this!) and if you're so inclined you could see every change to a counter this way.   The data model defines a "stream" model which eliminates the individual events and aggregates them together. You can no longer see the individual updates.

For a streaming trace SDK, I would recommend you consider the `SetAttribute()` as an immutable event. Considering a string-valued attribute being set multiple times, you will see one event for every modification to the attribute. When all the events are collected, you will see an attribute that was modified several times. But at each point in time, semantically, the attribute has a single value. The consumer of your streaming SDK may see a timeline of updates to the event, and you can show all the values that were set, but I would say it's off-spec to consider the set of values that were set to be a multi-valued map.

tigrannajaryan:
> But at each point in time, semantically, the attribute has a single value.

I think this is key observation. I intentionally avoid specifying who and how enforces these semantics because there are many different places and ways to achieve it. As long as we explicitly tell that it is a requirement how exactly it is achieved can and should be left undefined to give enough leeway for implementations to do the best they can.

I will follow up with a PR for OTLP in the proto repo, specifically requiring that senders must not have duplicates in the attributes list. 

Other implementations using other protocols (e.g stream SDK with an exporter that uses an event-based protocol) can move the uniqueness enforcement requirement to the receiver/backend.

yurishkuro:
I mostly agree with what @jmacd is saying. What I see problematic is that this is clearly a very nuanced topic, as evidenced by this discussion, but the spec change makes a blunt statement that pretends that these nuances do not exist. If I were reading this spec for the first time, I would still have the exact same questions I raised here, because the wording does not answer any of them.

tigrannajaryan:
> I mostly agree with what @jmacd is saying. What I see problematic is that this is clearly a very nuanced topic, as evidenced by this discussion, but the spec change makes a blunt statement that pretends that these nuances do not exist. If I were reading this spec for the first time, I would still have the exact same questions I raised here, because the wording does not answer any of them.

This is a great feedback. Let me do another round and see if I can add more clarity and capture these nuances.

yurishkuro (approved):

tigrannajaryan:
@yurishkuro @jmacd I added some wording to address your comments. Please have another look and see if it helps.

tigrannajaryan:
CI failure doesn't seem to be related to my change in any way.

tigrannajaryan:
@jmacd PTAL.

jmacd (approved):

carlosalberto:
Tried to merge but it won't allow me, as there's one **outdated** unresolved conversation - which cannot be found now, hence cannot be resolved.

Wondering if anybody has experienced this _feature_ before?

PR #2252, opened 2022-01-07
Specify optional Exponential Histogram Aggregation, add example code in the data model

jmacd:
Part of #1935.

This protocol was released in OTLP v0.11.

Related [OTEP 149](https://github.com/open-telemetry/oteps/blob/main/text/0149-exponential-histogram.md).

## Changes

In https://github.com/open-telemetry/opentelemetry-collector/pull/4642 I introduced temporary support for printing the exponential histogram data point. I used equations added to the data model as examples here.

This document is meant to justify merging the reference implementation shown in this (unfortunately LARGE) PR: https://github.com/open-telemetry/opentelemetry-go/pull/2393

I've proposed to split it into two parts: https://github.com/open-telemetry/opentelemetry-go/issues/2501. The implementation is described in further detail in its [README](https://github.com/jmacd/opentelemetry-go/blob/77d01c9db493a0adefbb71b0f0ce89b321ab916a/sdk/metric/aggregator/exponential/README.md).

The specification changes in this PR describe the critical aspects of this reference implementation in terms of the new Aggregations' two configuration parameters: MaxSize, and RangeLimits (optional) and two requirements for its behavior.

As further motivation, there is a PR to add `receiver/statsdreceiver` support using this reference implementation, https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/6666, which is waiting for everything above to be merged.


jmacd:
@MrAlias FYI, this is meant to assist with eventually merging the reference implementation in OTel-Go.

jmacd:
The equations included in this PR are tested in the corresponding OTel-Go PR: https://github.com/open-telemetry/opentelemetry-go/pull/2502

specification/metrics/sdk.md
```diff
@@ -393,6 +401,19 @@ Aggregation](./sdk.md#explicit-bucket-histogram-aggregation).
 
 This Aggregation does not have any configuration parameters.
 
+##### Histogram Aggregation common behavior
+
+Histogram aggregations should not fill out `sum` when used with
```
jsuereth, line 406: Can we lift this for Exponential histograms?  Or is this just to try to have them be a drop-in replacement?

specification/metrics/sdk.md
```diff
@@ -406,15 +427,43 @@ This Aggregation honors the following configuration parameters:
...
+for the [Exponential Histogram Metric
+Point](./datamodel.md#exponentialhistogram), which uses an exponential
+formula to determine bucket boundaries and an integer `scale`
+parameter to control resolution.
+
+This Aggregation honors the following configuration parameters:
+
+| Key                    | Value           | Default Value | Description                                                                                                                                                                                                                                     |
+| --                     | --              | --            | --                                                                                                                                                                                                                                              |
+| MaxSize                | integer         | 320           | Maximum number of buckets in each of the positive and negative ranges, not counting the special zero bucket.                                                                                                                                    |
```
jsuereth, line 448: Is 320 buckets a reasonable default?   Would it make sense to have a default = the same number of explicit buckets, e.g 11?

I think this definitely needs to be configurable, but I'm also curious what a good default would be, and 320 seems pretty high if it's used in the same locations where we used to have 11.

jmacd:
@beorn7 I would like your feedback on this proposal. To help you consider the options, consider an OpenTelemetry SDK standing in as a Prometheus client. You have perfect control over the histogram behavior: you can choose a fixed scale factor and have variable size, or you can choose range limits, fixed size and fixed scale, and I believe in a Prometheus setting these decisions should be made up front.

As a first attempt, I outlined an exponential histogram aggregator with one mandatory setting (size) and one optional setting (range limits). The user wouldn't ever set scale directly under this proposal. What do you think?

cc/ @BrianBrazil

brianbrazil-seczetta:
@brian-brazil?

jmacd:
Yes, thank you @brianbrazil-seczetta. @brian-brazil one day we hope to add you to the OTel organization 😁

beorn7:
@jmacd Thanks for pinging me. I have trouble finding time to look at this in detail (since I'm working heads-down on the Prometheus histograms, hopefully getting them to a state where less is in flux and I can give better answers to OTel questions ;). I'm not quite sure what your request is here. In Prometheus, the instrumented binary decides what to expose, and then independent scrapers can do with it what they want (including reducing the resolution AKA increasing the bucket width if they want to store at lower resolution).

https://github.com/prometheus/client_golang/blob/70253f4dd027a7128cdd681c22448d65ca30eed7/prometheus/histogram.go#L384-L407 is how to set the scale.

https://github.com/prometheus/client_golang/blob/70253f4dd027a7128cdd681c22448d65ca30eed7/prometheus/histogram.go#L421-L440 sets the strategy how to limit bucket numbers, which, at a first glance, looks similar to what's proposed here.

Does this help? I'm sorry if I missed the point while just skimming this PR. Feel free to ask more specific questions, and I'll try my best in the time given to me.

jmacd:
Thank you @beorn7, very helpful feedback.

Compatibility note: In OTel's protocol and this document `Scale` is equal to `-SparseBucketsFactor` in your struct.

I am trying to avoid letting the user set scale directly, because it is difficult to reason about.  I was proposing that users either (a) do not set scale, or (b) configure max-size and min/max range limits, which imply a fixed scale. I see that your configuration is more flexible.

For OTel to emulate the behavior implied by your sparse-histogram settings, I will have to make an adjustment in this proposal, in probably two parts:

1. the range limits described thus far have been all-or-none.  You're describing an independent minimum limit and no maximum limit. A user could not emulate what you have using `WithRangeLimits(minimum, math.MaxFloat64)` because we want a way to set the minimum while keeping adjustable scale (i.e., without setting a maximum).
2. the option to reset a histogram and use a new start timestamp is already built into the OTel data model, but having a histogram setting to force reset when scale falls below a threshold, that could be useful.

By the way, I see you have `SparseBucketsFactor float64` . OTel limited scale to integers so that all histogram conversions are non-lossy, but I admit that's not a strong requirement. The equations in this PR could be updated to compute non-integer scales by replacing `scaleFactor = math.Ldexp(math.Log2E, scale)` with `math.Log2E * math.Exp2(scale)` and so on; the mapping functions for non-integer scale perform the same as those documented here, only that the use of non-integer scale admits lossy conversions.

OTel reviewers, if you think having the ability to directly set the scale matters, please say so.

I'll update this PR with point (1) above (i.e. make limits independent) and we can address (2) when the time comes. Thank you!

specification/metrics/sdk.md
jmacd: Sorry -- I should have documented this choice as my first proposal!

The way I came to it: Prometheus has default boundaries 0.005 through 10. OpenMetrics a.k.a. Circllhist is a popular implementation of high-res histogram that has buckets which exactly land on the Prometheus buckets.

OpenMetrics has 90 buckets per factor of 10. There are 320 OpenMetrics buckets spanning between 0.005 and 10, it's three decades (90 + 90 + 90 buckets) plus the second half of a decade (0.005 through 0.01 = 50 buckets). This means that as long as your measurements are inside the Prometheus default range, you get at least OpenMetrics resolution.

For reasons I'm not sure of (could be the same?), the NrSketch code also uses 320 buckets by default. (@yzhuge -- comment?)

I can easily be convinced to lower this. @beorn7 I wonder what you / Prometheus consider a good default size limit, or whether by default you prefer no limit on size?

specification/metrics/sdk.md
jmacd: I want them to be a drop-in replacement, I also want us to solve the problem for both :-)

specification/metrics/sdk.md
brian-brazil: > OpenMetrics has 90 buckets per factor of 10

I think you're thinking of the experimental work @beorn7 is doing, OpenMetrics doesn't have anything like this currently.

jmacd: Sorry @brian-brazil I made a huge mistake. Everywhere in this specific comment above I meant "OpenHistogram", meaning https://openhistogram.io/

yzhuge: NrSketch chose 320 buckets because:
* For dataset with contrast (maxValue divided by minValue) up to 1 million, it gives about 2% relative error for quantile calculation from a histogram. This meets precision requirement of most use cases, such as SLO.
* 320 buckets usually cost less than 1KB, when using techniques like variable size counter (as used in NrSketch) for in memory storage, with minimal cpu cost when processing it. For on disk storage or network transportation, further compression can be used. Such space cost was considered acceptable in New Relic telemetry use cases (A thousand histograms only cost around 1MB).

beorn7: Clearly, too many OpenXYZ's in the game… :laughing:

beorn7:
> Compatibility note: In OTel's protocol and this document `Scale` is equal to `-SparseBucketsFactor` in your struct.

Actually not. It's more like the `scaleFactor` in this PR. The user provides the desired growth factor between one bucket boundary and the next, and then the code picks a scale that will provide the largest growth factor that is equal or smaller than the provided `SparseBucketsFactor`. For example, `SparseBucketsFactor = 1.1` results in a `Scale` of 3 (2^2^-3 = 1.0905).
 
> I am trying to avoid letting the user set scale directly, because it is difficult to reason about.

Yeah, and that's why we do what's described above. The growth factor gives you a good intuition of the precision the histogram provides.

> 1. the range limits described thus far have been all-or-none.  You're describing an independent minimum limit and no maximum limit. A user could not emulate what you have using `WithRangeLimits(minimum, math.MaxFloat64)` because we want a way to set the minimum while keeping adjustable scale (i.e., without setting a maximum).

I guess with the same line of argument that leads to a zero bucket of finite width, one might require an "overflow bucket" and an "underflow bucket" for observations exceeding a configurable max/min value. That would be nicely symmetric, but I guess, the demand hasn't come up in practice because it is relatively easy to "accidentally" create observations very close to zero (due to floating point arithmetic precision issues, or if the observations are coming from actual physical measurements), while I would assume the cases where you accidentally create extremely large observations are much less common.

So far, we have gone for not doing "overflow/underflow buckets", but if someone has relevant need, please let me know.

> By the way, I see you have `SparseBucketsFactor float64` . OTel limited scale to integers so that all histogram conversions are non-lossy, but I admit that's not a strong requirement. The equations in this PR could be updated to compute non-integer scales by replacing `scaleFactor = math.Ldexp(math.Log2E, scale)` with `math.Log2E * math.Exp2(scale)` and so on; the mapping functions for non-integer scale perform the same as those documented here, only that the use of non-integer scale admits lossy conversions.

I think this is all a misunderstanding, see above. In OTel terms, the Prometheus histograms (in the current PoC state) always have an integer scale between -8 and 4, and histograms can always be merged precisely to a histogram with the least common resolution.

specification/metrics/sdk.md
jmacd: Coming back to @jsuereth's original question, I think the answer is that if a user really wants ~11 buckets, they'll probably keep the explicit bucket histogram. I think the default here should be at least 100, and I'm happy to negotiate. 

Put another way, I think that if the user is really concerned about the cost of a histogram, they're not going to choose a very small histogram, they're going to downgrade to "MinMaxSumCount", which is the 0-bucket explicit histogram. The choice of exponential histogram, to me, signals an interest in high resolution for reasonable cost.

I can be convinced to choose a small auto-scaling default.  Looking for other's input here. How does 32 sound?

yzhuge: I think we need a relative error of 5% or less by default. This needs scale 3, with 8 buckets per 2x range, base at 1.09, relative error at 4.3% (see https://github.com/newrelic-experimental/newrelic-sketch-java). At this scale, for a dataset of 2^n contrast (maxValue divided by minValue in dataset), we need 8*n buckets.

For 32 buckets, n=4, contrast=16, not exactly a high dynamic range dataset. 

For the **high dynamic range long tail dataset** we are targeting, like response time from 1ms to 100sec, we have a contrast of 100k (n=17, 136 buckets). Fewer buckets won't work out of the box for such dataset. I suggest a default of 160. For a few buckets more, we can cover a contrast of 1M (n=20), which should cover like 99.9% use cases. 

Actual number of buckets range for max to max/2. So for max of 160, it is 80 to 160, with average at 120, with variable type counter (4 byte per counter in most cases), this will be a memory footprint of around 500 bytes. A typical app will likely have up to a few dozen histograms, for the cost of 5 to 10 kilo bytes, very affordable for today's computer systems.

jmacd: @yzhuge Thank you for the explanation. I have come to understand that values like 320 and 160 are "good" because they can be factored into `10 * 2**K` and that makes it easy to calculate maximum contrast from K and/or calculate K from maximum contrast. For 160 (K=4), I filled in examples for max-contrasts 5.6 (K+2), 32 (K+1), 1024 (K), 2**20 (K-1), and so on. I copied your derivation for relative error, thank you.

This also applies the change I described above: MinValue and MaxValue limits are independent and both optional. The requirements have been rephrased to allow 0, 1, or 2 of these limits to be set. The scale is automatic when 0 or 1 limit are set, and fixed when 2 limits are set.

Hoping this helps, 8ed6d94.

jmacd:
By the way, thank you @beorn7 for clarifying: I understand why scaleFactor is a float now, and see no disagreements. :-)

specification/metrics/sdk.md
```diff
@@ -406,15 +429,87 @@ This Aggregation honors the following configuration parameters:
...
+Note that relative error is calculated as the half of the bucket width
+divided by the bucket midpoint, which is the same in every bucket.
+Using the bucket from [1, base), we have `(bucketWidth / 2) /
+bucketMidpoint = ((base - 1) / 2) / ((base + 1) / 2) = (base - 1) /
+(base + 1)`.
+
+##### Exponential Histogram Aggregation: Handle range limits
+
+When either the MinValue or MaxValue range limit is configured, the
+implementation is REQUIRED to reject out-of-range measurements and
```
jsuereth, line 487: I feel like we need a way to capture out-of-range here.    Is it worth adding an overflow/underflow bucket or just a single "out of range" bucket to know we're missing data in the histogram?

oertl: If there are underflow and overflow buckets, it is still possible to calculate some order statistics (e.g. median). This is not possible if there is only one out-of-range bucket.

jsuereth: I feel like this is a restriction that we could lift if we find out it's not a good idea.  I'd prefer us to be flexible here, for two reasons:

1. Even logging I think it'll be hard for users to determine there are issues with their histogram bounds.  Either the log is lost, or users don't know to look for it.
2. Picking a bad upper/lower-bound leads to thinking an application is performing well, when in fact it may be catostrophically failing.

That said, we can try out being strict and get some real world usage behind it before we change our minds.

specification/metrics/sdk.md
```diff
@@ -406,15 +429,87 @@ This Aggregation honors the following configuration parameters:
...
+
+| Key                 | Value   | Default Value | Description                                                                                                  |
+|---------------------|---------|---------------|--------------------------------------------------------------------------------------------------------------|
+| MaxSize             | integer | 160           | Maximum number of buckets in each of the positive and negative ranges, not counting the special zero bucket. |
+| MinValue (optional) | double  | _not set_     | When set, limit the absolute value of measurements to be greater than or equal to this minimum value      |
+| MaxValue (optional) | double  | _not set_     | When set, limit the absolute value of measurements to be less than or equal to this maximum value         |
+
+The default of 160 buckets is selected to establish default support
+for a high-resolution histogram able to cover a long-tail latency
+distribution from 1ms to 100s with less than 5% relative error.
```
jsuereth, line 456: Love this explanation, thank you!!

jsuereth (approved):

MrAlias (approved):

jmacd:
@yzhuge @jsuereth I agree to remove the range limits; we should treat range limits as an instrument-level concern. I am comfortable with not giving the user configuring this any controls over the histogram scale. I've updated the PR.

jmacd:
This is low priority, not stale.

jmacd:
@jack-berg Please take a look.

MrAlias (approved):

jack-berg (approved):

reyang (approved):

specification/metrics/sdk.md
```diff
@@ -415,13 +435,88 @@ This Aggregation honors the following configuration parameters:
...
+1. The maximum supported scale, generally used for single-value histogram Aggregations where scale is not otherwise constrained
+2. The largest value of scale such that no more than the maximum number of buckets are needed to represent the full range of input data in either of the positive or negative ranges.
+
+##### Exponential Histogram Aggregation: Handle all normal values
+
+Implementations are REQUIRED to accept the entire normal range of IEEE
+floating point values (i.e., all values except for +Inf, -Inf and NaN
+values).
+
+Implementations are permitted to round subnormal values away from zero
```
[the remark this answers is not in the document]
jmacd: Done.

jmacd:
@yzhuge @oertl Please take another look. Your approvals really matter here, thanks. :-)

specification/metrics/sdk.md
```diff
@@ -406,15 +429,87 @@ This Aggregation honors the following configuration parameters:
...
+Range limits aside, implementations are REQUIRED to accept the entire
+normal range of IEEE floating point values (i.e., all values except
+for +Inf, -Inf and NaN values).
+
+Implementations are permitted to round subnormal values away from zero
+to the nearest normal value when it simplifies the implementation.
+
+##### Exponential Histogram Aggregation: Maintain the ideal scale when values are not limited in both directions
+
+Implementations SHOULD adjust the histogram scale as necessary to
```
yzhuge, line 522: suggest changing this paragraph to:

Implementations SHOULD adjust the histogram scale as necessary to maintain the best resolution possible, within the constraint of maximum size (max number of buckets), unless both MinValue and MaxValue range limits are set.  Best resolution (highest scale) is achieved when the number of positive or negative
range buckets exceeds half the maximum size, such that
increasing scale by one would not be possible given the size
constraint.

yzhuge (approved):
Just some editorial suggestions

specification/metrics/datamodel.md
```diff
@@ -727,19 +742,57 @@ correct rounding for the negative indices.  This may be written as:
...
+    return math.Exp(index * inverseFactor), nil
+```
+
+Implementations are expected to verify that their mapping function and
+inverse mapping function are correct near the lowest and highest IEEE
+floating point values.  In the Golang reference implementation, for
+example, the above formula computes `+Inf` for the maximum-index bucket.
+In this case, it is appropriate to subtract `1<<scale` from the index
+and multiply the result by `2`.
+
```
oertl, line 786: Is there any proof that this workaround never leads to an infinite result? I guess the behavior is platform-dependent without strict guarantees that infinite results will never occur. I can imagine that using `pow(2, ldexp(index, -scale))` instead would have higher chances to avoid infinite values. The reason is that the intermediate result of `ldexp(index, -scale)` will be exact. So any inaccuracies will only come from the `pow` function itself. Some programming languages like C++ or Go even define `exp2(x)` which could be used instead of `pow(2, x)` and which potentially would avoid accidential infinite values by definition.

specification/metrics/sdk.md
```diff
@@ -397,6 +408,24 @@ This Aggregation informs the SDK to collect:
 - The last `Measurement`.
 - The timestamp of the last `Measurement`.
 
+#### Histogram Aggregation
+
+The Histogram Aggregation informs the SDK to select the best Histogram
+Aggregation available. i.e. [Explicit Bucket Histogram
```
aabmass, line 414: Would the best available not be the Exponential Histogram?

specification/metrics/sdk.md
```diff
@@ -410,13 +439,88 @@ This Aggregation honors the following configuration parameters:
...
+
+| Input range | Contrast | Ideal Scale | Base     | Relative error |
+|-------------|----------|-------------|----------|----------------|
+| 1ms - 4ms   | 4        | 6           | 1.010889 | 0.542%         |
+| 1ms - 20ms  | 20       | 5           | 1.021897 | 1.083%         |
+| 1ms - 1s    | 10**3    | 4           | 1.044274 | 2.166%         |
+| 1ms - 100s  | 10**5    | 3           | 1.090508 | 4.329%         |
+| 1μs - 10s   | 10**7    | 2           | 1.189207 | 8.643%         |
+
+Note that relative error is calculated as the half of the bucket width
```
aabmass, line 488:
```suggestion
Note that relative error is calculated as half of the bucket width
```

specification/metrics/sdk.md
```diff
@@ -410,13 +439,88 @@ This Aggregation honors the following configuration parameters:
...
+floating point values (i.e., all values except for +Inf, -Inf and NaN
+values).
+
+Implementations MAY round subnormal values away from zero to the
+nearest normal value.
+
+##### Exponential Histogram Aggregation: Support a minimum and maximum scale
+
+The implementation MUST maintain reasonable minimum and maximum scale
+parameters, values that the automatic scale parameter will not exceed.
```
aabmass, line 512: nit wording

```suggestion
parameters that the automatic scale parameter will not exceed.
```

aabmass:
> Specify optional Exponential Histogram Aggregation, add example code in the data model

Is this intended to be optional for SDK authors to implement? I don't see anything in the spec saying so.

specification/metrics/sdk.md
jmacd: This was a mistake -- existing text was being refactored that has since been removed.

specification/metrics/sdk.md
```diff
@@ -397,6 +408,24 @@ This Aggregation informs the SDK to collect:
...
+
+This Aggregation does not have any configuration parameters.
+
+##### Histogram Aggregation common behavior
+
+All histogram Aggregations inform the SDK to collect:
+
+- Count of `Measurement` values in population.
+- Arithmetic sum of `Measurement` values in population. This SHOULD NOT be collected when used with
+instruments that record negative measurements (e.g. `UpDownCounter` or `ObservableGauge`).
```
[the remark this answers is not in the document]
jmacd: Yes. This behavior is copied from the existing definition for explicit bucket histograms, around line 416 before.

specification/metrics/sdk.md
```diff
@@ -410,13 +439,88 @@ This Aggregation honors the following configuration parameters:
...
+either:
+
+1. The maximum supported scale, generally used for single-value histogram Aggregations where scale is not otherwise constrained
+2. The largest value of scale such that no more than the maximum number of buckets are needed to represent the full range of input data in either of the positive or negative ranges.
+
+##### Exponential Histogram Aggregation: Handle all normal values
+
+Implementations are REQUIRED to accept the entire normal range of IEEE
+floating point values (i.e., all values except for +Inf, -Inf and NaN
+values).
```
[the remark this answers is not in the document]
jmacd: I'm not sure -- I was thinking that the sdk gave general guidelines for Inf and NaN, meaning this wouldn't be a question specific to the exponential histogram in any case. The section is "Numerical Limits Handling", but it just says to handle those values. I think of Inf and NaN values as lying "outside the normal range", in any case.

How would you like to handle those values? We can certainly count them, and the Sum is defined by IEEE 754, but I'm not sure it's useful to accept Inf and NaN values in counters and histograms. For Gauges, that's a separate matter.

specification/metrics/sdk.md
jmacd: :+1: (Except the MinValue and MaxValue -- those elements were removed from this PR very early.)

specification/metrics/datamodel.md
```diff
@@ -633,19 +648,57 @@ correct rounding for the negative indices.  This may be written as:
...
 ```golang
-    scaleFactor := math.Log2E * math.Exp2(scale)
-    return int64(math.Floor(math.Log(value) * scaleFactor))
+    scaleFactor := math.Ldexp(math.Log2E, scale)
+    return math.Floor(math.Log(value) * scaleFactor)
 ```
 
 Note that in the example Golang code above, the built-in `math.Log2E`
-is defined as `1 / ln(2)`.
+is defined as the inverse natural logarithm of 2.
```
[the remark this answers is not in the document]
jmacd: Done, thanks.

specification/metrics/datamodel.md
jmacd: I do not have a proof, I just know that I was having difficulty computing the last finite boundary at all scales, and this formula helped. Using the Golang `math/big` package is one answer to finding the exact last finite boundary, but I don't know how to reason about which variation on these equations will overflow. My intuition is that different equations can't both produce accidental infinity, so trying both will always work.

specification/metrics/datamodel.md
```diff
@@ -633,19 +648,57 @@ correct rounding for the negative indices.  This may be written as:
...
+The reverse mapping function is:
+
+```golang
+    inverseFactor := math.Ldexp(math.Ln2, -scale)
+    return math.Exp(index * inverseFactor), nil
+```
+
+Implementations should verify that their mapping function and inverse
+mapping function are correct near the lowest and highest IEEE floating
+point value.  In the Golang reference implementation, for example, the
```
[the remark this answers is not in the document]
jmacd: Done!

specification/metrics/datamodel.md
jmacd: The problem, however, is that the other calculation published here has shown itself to not work near an extreme value. It seems none of them are guaranteed, and I thought providing two valid calculations would be better than zero. OTOH this points to yet another formulation where the significant and exponent are processed separately, which I'd rather not get into.

specification/metrics/sdk.md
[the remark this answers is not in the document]
jmacd: I agree with this point! I'm just not sure whether we want to have some aggregations making their own mind about NaN and Inf values and allow inconsistency. For example, if a Counter is configured with standard Sum aggregation and you see a Inf or NaN, problems will arise when say computing deltas from cumulatives.

specification/metrics/datamodel.md
yzhuge: The NrSketch SubBucketIndexer can avoid overflow at extreme values, because it uses integer math on exponent and float point only on the mantissa, which is limited between 1.0 and 2.0. See https://github.com/newrelic-experimental/newrelic-sketch-java/blob/main/Indexer.md

jmacd: Yes. I understand that there are many mathematically correct formulas we can use to compute bucket indexes and boundaries. The point I was trying to illustrate is that none of these formulas are guaranteed to work in every situation with absolute precision; what is required of authors is to understand the math libraries at their disposal and ensure a good outcome. I've listed one alternate equation that can be used (among many) to illustrate that it is easy to find another equation that shifts the input away from an extreme value if it will help with correctness.

jmacd:
@reyang The remaining question in this PR gets to a bigger question about handling NaN and Inf values in the Metrics API. Should Exponential Histograms treat NaN and Inf values any differently than Counter instruments. If no, what to do we expect? If yes, what do we expect?

reyang:
> @reyang The remaining question in this PR gets to a bigger question about handling NaN and Inf values in the Metrics API. Should Exponential Histograms treat NaN and Inf values any differently than Counter instruments. If no, what to do we expect? If yes, what do we expect?

I think none of these should block this PR.

And I think the API spec doesn't care about NaN/Inf.
The SDK spec has a clear position on NaN/Inf https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#numerical-limits-handling.

Here goes something I did for .NET (and it's still experimental)
https://github.com/open-telemetry/opentelemetry-dotnet/blob/117b70c039a9b012887e3aa29717e2d4cee6274e/test/OpenTelemetry.Exporter.Prometheus.Tests/PrometheusSerializerTests.cs#L359-L362

reyang:
@aabmass @oertl what's your take on https://github.com/open-telemetry/opentelemetry-specification/pull/2252#issuecomment-1110461601?

I'm trying to understand your position - do you think we need to discuss more here (and we should not merge the PR before you feel comfortable to sign off), or we are good to merge this PR and have separate conversation about the NaN/Inf/limits? Thanks!

jmacd:
@reyang I added two commits.

[d15ea33](https://github.com/open-telemetry/opentelemetry-specification/pull/2252/commits/d15ea330613198a93ae520d65db0b5d4f00918c2) is meant to [address a question from Slack about Prometheus' exponential histogram interoperability.
](https://cloud-native.slack.com/archives/C01NP3BV26R/p1651072850521739)

[1a473d0](https://github.com/open-telemetry/opentelemetry-specification/pull/2252/commits/1a473d0b4201b9cd8c2ac500f2c91c68d3f234bc) is meant to answer the question you're asking; I believe we have sufficiently general statements about treating Inf and NaN values, however for histogram aggregation I think we want all-or-none behavior, meaning the Sum, Count, Min, Max, and Buckets should be consistent. Since we can't have consistent results with Inf and NaN values because they do not map into valid buckets, I think histogram implementations MUST disregard these. See what you think.

wicha:
Hey Team OpenTelemetry. CTO/Cofounder of a startup that values your work and uses OpenTelemetry here. Is this going to be merged soon? We need this PR merged for some resolution changes we have internally at our company. I'd appreciate if you could merge 🙏

wicha:
> @beorn7 @oertl FYI - we plan to merge this PR by end of the week (May 13th, 2022) unless you still see blocking issues.
> 
> @wicha FYI - we'll merge the PR by end of Friday, May 13th, 2022 - unless there is blocking comment(s).

🙏 thank you!

oertl (approved):

PR #2254, opened 2022-01-09
Rename OTLP exporter env variable to be consistent

lonewolf3739:
## Changes

Looks like this slipped through reviews, all the env variables follow `OTLP_TRACES_*` pattern.


lonewolf3739:
Just noticed there is already another PR, Closing in favour of #2240

PR #2258, opened 2022-01-11
Add semantic conventions for k8s

hprajapati-splunk:
## Changes

- Added k8s events related attributes ([#6885](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/6885))

- Added k8s resource object-related attributes as there can be objects of [CRD](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#customresourcedefinitions), so to make generic semantics to represent any type of k8s objects (CRD or defaults like a pod, node, etc).


specification/resource/semantic_conventions/k8s.md
```diff
@@ -185,3 +185,39 @@ A CronJob creates Jobs on a repeating schedule.
 | `k8s.cronjob.uid` | string | The UID of the CronJob. | `275ecb36-5aa8-4c2a-9c47-d8bb681b9aff` | No |
 | `k8s.cronjob.name` | string | The name of the CronJob. | `opentelemetry` | No |
 <!-- endsemconv -->
+
+## Event
```
bogdandrutu, line 189: This is a strange thing to have in the "Resource". Should this be a type in our logs?

specification/resource/semantic_conventions/k8s.md
```diff
@@ -185,3 +185,39 @@ A CronJob creates Jobs on a repeating schedule.
...
+|---|---|---|---|---|
+| `k8s.event.uid` | string | The UID of the Event. | `275ecb36-5aa8-4c2a-9c47-d8bb681b9aff` | No |
+| `k8s.event.name` | string | The name of the Event. | `opentelemetry-depl-5c8bf76b5b-tn9h5.16a678548ba4963c` | No |
+| `k8s.event.start_time` | string | The start time of the Event. | `2021-09-20T07:50:19Z` | No |
+| `k8s.event.action` | string | The action was taken/failed regarding the related object of the Event. | `Binding` | No |
+| `k8s.event.count` | int | The number of times the Event has occurred. | `1`; `2` | No |
+| `k8s.event.source` | string | The component reporting the Event. | `kubelet` | No |
+<!-- endsemconv -->
+
+## Object
```
bogdandrutu, line 209: We have all the objects defined already, what is the purpose of this?

specification/resource/semantic_conventions/k8s.md
tigrannajaryan: +1

This needs to be a semantic convention for logs/traces. Which one more time demonstrates that we need to be able to have the same namespace (`k8s.` in this case) in the Resource and in Traces/Logs.

There is also an unrelated [open issue](https://github.com/open-telemetry/opentelemetry-specification/pull/2145#issuecomment-995818361) that needs to add a convention to `db.` namespace but cannot because of the generator tool limitations. We need to fix the limitation and have  one space for all semantic conventions, with every attributes clearly stating which data types it is applicable to.

cc @manang-splunk

specification/resource/semantic_conventions/k8s.md
hprajapati-splunk: In the Kubernetes, as you might know, there is a way to add a [CRD](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#create-a-customresourcedefinition) and we can then [make objects](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#create-custom-objects) using the same CRD, so by adding this group (`Object`), we can also cover that aspect.

specification/resource/semantic_conventions/k8s.md
hprajapati-splunk: @tigrannajaryan, @bogdandrutu, yes, it should be at the log level and I think @manang-splunk had found the [solution](https://github.com/open-telemetry/opentelemetry-specification/pull/2145#issuecomment-1012166017) to the issue mentioned above. 

So, I have pushed the code to move it under the log level.

semantic_conventions/resource/k8s.yaml
```diff
@@ -168,3 +168,24 @@ groups:
...
+      - id: name
+        type: string
+        brief: >
+          The name of the Object.
+        examples: ['opentelemetry-1']
+      - id: kind
+        type: string
+        brief: >
+          A value representing the resource.
+        examples: ['opentelemetry']
```
dmitryax, line 191: This is not the best example. I don't think this there is a k8s resource kind called `opentelemetry`. Can use `opentelemetrycollector` from https://github.com/open-telemetry/opentelemetry-operator/blob/55a0dc2e86a930ff72aac54391f369520c0d8064/README.md#getting-started

semantic_conventions/resource/k8s.yaml
```diff
@@ -168,3 +168,24 @@ groups:
...
+        examples: ['275ecb36-5aa8-4c2a-9c47-d8bb681b9aff']
+      - id: name
+        type: string
+        brief: >
+          The name of the Object.
+        examples: ['opentelemetry-1']
+      - id: kind
+        type: string
+        brief: >
+          A value representing the resource.
```
dmitryax, line 190: maybe?
```suggestion
          A value representing kind of the Object.
```

semantic_conventions/resource/k8s.yaml
```diff
@@ -168,3 +168,24 @@ groups:
         brief: >
           The name of the CronJob.
         examples: ['opentelemetry']
+
+  - id: k8s.object
+    prefix: k8s.object
+    brief: >
+      A Kubernetes Resource Object.
```
dmitryax, line 175: If we keep this namespace, the difference from other existing namespaces should be more clearly described as another way to represent any k8s object including CRD. Probably should emphasize k8s.object.kind.

semantic_conventions/logs/k8s.yaml
```diff
@@ -0,0 +1,36 @@
...
+      - id: count
+        type: int
+        brief: >
+          The number of times the Event has occurred.
+        examples: [1, 2]
+      - id: source
+        type: string
+        brief: >
+          The component reporting the Event.
+        examples: ['kubelet']
```
dmitryax, line 36: nit: missed trailing newline

semantic_conventions/resource/k8s.yaml
hprajapati-splunk: I have changed the example.

semantic_conventions/resource/k8s.yaml
hprajapati-splunk: I have updated the brief accordingly and also the description in the k8s.md.

semantic_conventions/logs/k8s.yaml
hprajapati-splunk: I have added the new line.

semantic_conventions/resource/k8s.yaml
hprajapati-splunk: I have made the suggested changes.

dmitryax (approved):
LGTM (all my comments can be marked as resolved, I cannot do that)

tigrannajaryan:
`k8s.event.*` log conventions look good to me. 

I am not sure about `k8s.object.*` conventions though. We record all other K8s resources as specific concepts, e.g. Pod is recorded as `k8s.pod.*`. The proposal to use `k8s.object.*` seems to deviate from that idea of specificity and uses a generic concept of "object" or perhaps I misunderstand the intent. Is K8s Object a distinct, specific concept in K8s, that leaves on the same level of taxonomy of concepts as Pod, Node, DaemonSet, etc?

I suggest that this PR is split into 2 parts:
- Part 1 for `k8s.event.*`. I am happy to approve it. I think this one has a good chance to be merged sooner than Part 2.
- Part 2 for `k8s.object.*`. This one I am not sure about and I think it needs more discussion (I see Bogdan also the same concern as I have).

dmitryax:
Splitting makes a lot of sense. @hprajapati-splunk can you please open two separate PRs? We can continue discussing `k8s.object.*` there

specification/logs/semantic_conventions/k8s.md
```diff
@@ -0,0 +1,23 @@
+# Kubernetes
+
+**Status**: [Experimental](../../document-status.md)
+
+## Event
+
+The Event is created when other Kubernetes resources have state changes,
+errors, or other messages that should be broadcast to the system.
+
+**type:** `k8s.event`
```
Sprakhar97, line 10: how do a consumer will differentiate between a log record and an event record when both are sent as a log?

specification/logs/semantic_conventions/k8s.md
```diff
@@ -0,0 +1,23 @@
...
+
+<!-- semconv k8s.event -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `k8s.event.uid` | string | The UID of the Event. | `275ecb36-5aa8-4c2a-9c47-d8bb681b9aff` | No |
+| `k8s.event.name` | string | The name of the Event. | `opentelemetry-depl-5c8bf76b5b-tn9h5.16a678548ba4963c` | No |
+| `k8s.event.start_time` | string | The start time of the Event. | `2021-09-20T07:50:19Z` | No |
+| `k8s.event.action` | string | The action was taken/failed regarding the related object of the Event. | `Binding` | No |
+| `k8s.event.count` | int | The number of times the Event has occurred. | `1`; `2` | No |
+| `k8s.event.source` | string | The component reporting the Event. | `kubelet` | No |
```
Sprakhar97: There are certain critical attributes like reason, severity and event description that are missing.

specification/logs/semantic_conventions/k8s.md
dmitryax: I believe events can be differentiated by having `k8s.event.` attributes

specification/logs/semantic_conventions/k8s.md
dmitryax: Right, `k8s.event.reason` is missed. @hprajapati-splunk please add it. OTel log model has severity native field, and Log body should be used for event description.

dmitryax: Hm. Actually `k8s.event.reason` is there already

specification/resource/semantic_conventions/k8s.md
joaopgrassi: Actually, I give a little bit of a try in #2145 and it fails because we pass the whole semconv folder to the generator. If we split it into multiple invocations, passing the trace/resource folders separately, then (at least in my quick test) having duplicated namespaces like `db.*` wasn't a problem. But I'm not that familiar with the tool so I cannot say much if this is a good solution.

dmitryax:
This PR is superseded by #2300 #2299 and can be closed

PR #1456, opened 2021-02-23
Add more description for symbols in Optional column

ThomsonTan:
Fixes #1455 

## Changes

Move the description of `*` for `Optional` column to the beginning of the doc where other indicators are documented and add description for the possible cell content `X` and blank cell.

Related issues #

Related [oteps](https://github.com/open-telemetry/oteps) #

PR #350, opened 2019-11-12
Fix typos in api-metrics.md

arminru:

PR #2262, opened 2022-01-12
Change golang namespace to 'go', rather than 'gc'

djaglowski:
Resolves [opentelemetry-go-contrib #350](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/350).

This was discussed briefly in the Golang SIG in December. Consensus at the time was that `go` is clearer than `gc` and sufficiently appropriate. The move away from `gc` is helpful in part because the namespace is used to mean "garbage collector" elsewhere in the semantic conventions.

Related issues #

[opentelemetry-go-contrib #1456](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/1456)

### DIFF
--- a/specification/resource/semantic_conventions/process.md
+++ b/specification/resource/semantic_conventions/process.md
@@ -78,10 +78,30 @@ Example:
 
 ### Go Runtimes
 
-| Value | Description |
-| --- | --- |
-| `go` | Go compiler |
-| `gccgo` | GCC Go frontend |
+Go Runtimes should fill in the as follows:
+
+- `process.runtime.name` - Fill in an interpretation of Go's `runtime.Compiler` constant, according to the following rule: If the value is `gc`, fill in `go`. Otherwise, fill in the exact value of `runtime.Compiler`.
+
+  This can be implemented with the following Go snippet:
+
+  ```go
+  import "runtime"
+  
+  func getRuntimeName() string {
+    if runtime.Compiler == "gc" {
+      return "go"
+    }
+    return runtime.Compiler
+  }
+  ```
+
+- `process.runtime.version` - Fill in the exact value returned by `runtime.Version()`, i.e. `go1.17`.
+- `process.runtime.description` - Fill in a description corresponding to the value of `process.runtime.name`, as defined by the table below.
+
+| `runtime.Compiler` | `process.runtime.name` | `process.runtime.description` |
+| --- | --- | --- |
+| `gc` | go | cmd/compile |
+| `gccgo` | gccgo | gccgo front end |
 
 ### Java runtimes
 
--- a/specification/resource/semantic_conventions/process.md
+++ b/specification/resource/semantic_conventions/process.md
@@ -78,11 +78,9 @@ Example:
 
 ### Go Runtimes
 
-TODO(<https://github.com/open-telemetry/opentelemetry-go/issues/1181>): Confirm the contents here
-
 | Value | Description |
 | --- | --- |
-| `gc` | Go compiler |
+| `go` | Go compiler |
 | `gccgo` | GCC Go frontend |
 
 ### Java runtimes
--- a/specification/resource/semantic_conventions/process.md
+++ b/specification/resource/semantic_conventions/process.md
@@ -97,12 +97,7 @@ Go Runtimes should fill in the as follows:
   ```
 
 - `process.runtime.version` - Fill in the exact value returned by `runtime.Version()`, i.e. `go1.17`.
-- `process.runtime.description` - Fill in a description corresponding to the value of `process.runtime.name`, as defined by the table below.
-
-| `runtime.Compiler` | `process.runtime.name` | `process.runtime.description` |
-| --- | --- | --- |
-| `gc` | go | cmd/compile |
-| `gccgo` | gccgo | gccgo front end |
+- `process.runtime.description` - Use of this field is not recommended.
 
 ### Java runtimes
 
--- a/specification/resource/semantic_conventions/process.md
+++ b/specification/resource/semantic_conventions/process.md
@@ -99,6 +99,12 @@ Go Runtimes should fill in the as follows:
 - `process.runtime.version` - Fill in the exact value returned by `runtime.Version()`, i.e. `go1.17`.
 - `process.runtime.description` - Use of this field is not recommended.
 
+| `process.runtime.name` | Description |
+| --- | --- |
+| `go` | Official Go compiler. Also known as `cmd/compile`. |
+| `gccgo` | [gccgo](https://go.dev/doc/install/gccgo) is a Go [front end for GCC](https://gcc.gnu.org/frontends.html). |
+| `tinygo` | [TinyGo](https://tinygo.org/) compiler. |
+
 ### Java runtimes
 
 Java instrumentation should fill in the values by copying from system properties.
--- a/specification/resource/semantic_conventions/process.md
+++ b/specification/resource/semantic_conventions/process.md
@@ -80,7 +80,8 @@ Example:
 
 Go Runtimes should fill in the as follows:
 
-- `process.runtime.name` - Fill in an interpretation of Go's `runtime.Compiler` constant, according to the following rule: If the value is `gc`, fill in `go`. Otherwise, fill in the exact value of `runtime.Compiler`.
+- `process.runtime.name` - Fill in an interpretation of Go's [`runtime.Compiler`](https://pkg.go.dev/runtime#Compiler) constant, according to the following rule:
+  If the value is `gc`, fill in `go`. Otherwise, fill in the exact value of `runtime.Compiler`.
 
   This can be implemented with the following Go snippet:
 
--- a/specification/resource/semantic_conventions/process.md
+++ b/specification/resource/semantic_conventions/process.md
@@ -99,6 +99,8 @@ Go Runtimes should fill in the as follows:
 - `process.runtime.version` - Fill in the exact value returned by `runtime.Version()`, i.e. `go1.17`.
 - `process.runtime.description` - Use of this field is not recommended.
 
+Examples for some Go compilers/runtimes:
+
 | `process.runtime.name` | Description |
 | --- | --- |
 | `go` | Official Go compiler. Also known as `cmd/compile`. |


bogdandrutu (approved):
:) it has a GC :))

specification/resource/semantic_conventions/process.md
```diff
@@ -78,11 +78,9 @@ Example:
 
 ### Go Runtimes
 
-TODO(<https://github.com/open-telemetry/opentelemetry-go/issues/1181>): Confirm the contents here
-
 | Value | Description |
 | --- | --- |
-| `gc` | Go compiler |
+| `go` | Go compiler |
```
tigrannajaryan, line 83: Where did `gc` come from? Is there some argument in favour of `gc` or it is just a typo?

djaglowski: I'm not certain but I assume it was meant to stand for "Go Compiler". In any case, it was initialized as a best guess and called out as [needing review](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/350).

SergeyKanzhelev, line 83: can we add information how to extract these values same as done for erlang and java before removing TODO?

djaglowski: I found [this](https://pkg.go.dev/runtime#pkg-constants), which is clearly the original source of both `gc` and `gccgo`.

djaglowski: Yes, good catch. I was hasty there. Added proposed instructions for these values.

djaglowski:
As noted in my response to [Tigran's question](https://github.com/open-telemetry/opentelemetry-specification/pull/2262#discussion_r783426533), I've learned that Go has a [`runtime.Compiler`](https://pkg.go.dev/runtime#pkg-constants) constant that is either `gc` or `gccgo`. This weakens my proposal somewhat, but I still suggest that OTel use `go` instead of `gc` in order that the namespace is recognizable to those who are not familiar with this constant in Go's `runtime` package.

I've also added exact guidelines for populating `process.runtime.name`, `process.runtime.version`, and `process.runtime.description`.

djaglowski:
cc: @open-telemetry/go-approvers

pellared (approved):

XSAM (approved):
LGTM

specification/resource/semantic_conventions/process.md
```diff
@@ -78,12 +78,30 @@ Example:
...
 ### Go Runtimes
 
-TODO(<https://github.com/open-telemetry/opentelemetry-go/issues/1181>): Confirm the contents here
+Go Runtimes should fill in the as follows:
 
-| Value | Description |
-| --- | --- |
-| `gc` | Go compiler |
-| `gccgo` | GCC Go frontend |
+- `process.runtime.name` - Fill in an interpretation of Go's `runtime.Compiler` constant, according to the following rule: If the value is `gc`, fill in `go`. Otherwise, fill in the exact value of `runtime.Compiler`.
```
pellared, line 83:
```suggestion
- `process.runtime.name` - Fill in an interpretation of Go's [`runtime.Compiler`](https://pkg.go.dev/runtime#Compiler) constant, according to the following rule:
  If the value is `gc`, fill in `go`. Otherwise, fill in the exact value of `runtime.Compiler`.
```

specification/resource/semantic_conventions/process.md
```diff
@@ -78,12 +78,30 @@ Example:
...
+  }
+  ```
+
+- `process.runtime.version` - Fill in the exact value returned by `runtime.Version()`, i.e. `go1.17`.
+- `process.runtime.description` - Fill in a description corresponding to the value of `process.runtime.name`, as defined by the table below.
+
+| `runtime.Compiler` | `process.runtime.name` | `process.runtime.description` |
+| --- | --- | --- |
+| `gc` | go | cmd/compile |
+| `gccgo` | gccgo | gccgo front end |
```
pellared, line 104:
```suggestion
| `runtime.Compiler` value | `process.runtime.name` resource | `process.runtime.description` resource |
| --- | --- | --- |
| `gc` | `go` | `cmd/compile` |
| `gccgo` | `gccgo` | `gccgo front end` |
```

specification/resource/semantic_conventions/process.md
```diff
@@ -78,12 +78,30 @@ Example:
...
+  func getRuntimeName() string {
+    if runtime.Compiler == "gc" {
+      return "go"
+    }
+    return runtime.Compiler
+  }
+  ```
+
+- `process.runtime.version` - Fill in the exact value returned by `runtime.Version()`, i.e. `go1.17`.
+- `process.runtime.description` - Fill in a description corresponding to the value of `process.runtime.name`, as defined by the table below.
```
Oberon00, line 99: IMHO a static mapping from already reported information should be avoided here. If we have no additional info to add, I think we can recommend leaving description empty for now.

specification/resource/semantic_conventions/process.md
```diff
@@ -78,12 +78,30 @@ Example:
...
+  import "runtime"
+  
+  func getRuntimeName() string {
+    if runtime.Compiler == "gc" {
+      return "go"
+    }
+    return runtime.Compiler
+  }
+  ```
+
```
pellared, line 97:
```suggestion

Examples for some Go compilers/runtimes:

| Value | Description |
| --- | --- | --- |
| `go` | Official Go compiler . Also known as `cmd/compile`. |
| `gccgo` | [gccgo](https://go.dev/doc/install/gccgo) is a Go [front end for GCC](https://gcc.gnu.org/frontends.html). |
| `tinygo` | [TinyGo](https://tinygo.org/) compiler. |
```

pellared: @djaglowski Why do you think that this table is not needed? How the use would know e.g what `gccgo` mean? I think it will be convenient to write down the known value of `process.runtime.name`

djaglowski: Ok, I've added it.

specification/resource/semantic_conventions/process.md
```diff
@@ -78,12 +78,32 @@ Example:
...
+      return "go"
+    }
+    return runtime.Compiler
+  }
+  ```
 
-| Value | Description |
+- `process.runtime.version` - Fill in the exact value returned by `runtime.Version()`, i.e. `go1.17`.
+- `process.runtime.description` - Use of this field is not recommended.
+
```
pellared, line 101:
```suggestion

Examples for some Go compilers/runtimes:
```

just to make it clear that the list is not complete

pellared (approved):

tigrannajaryan:
@open-telemetry/go-approvers @open-telemetry/go-maintainers please review.

Aneurysm9 (commented):
I do not believe this change is necessary.  This affects a value in an attribute that clearly does not relate to garbage collection.  For attribute keys created by our runtime metrics instrumentation we already use `runtime.go.*`.  Having a special case for a particular compiler does not make sense to me here.

djaglowski:
The initial intent of this PR was to support the `process.runtime.go` namespace, but this was based on a misinterpretation of the current scope of this file.

The `process.runtime` namespace is established earlier in the file, so I was looking at the Go Runtimes section as establishing `process.runtime.gc` and `process.runtime.gccgo` namespaces, under the guidelines described [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/runtime-environment-metrics.md#metric-instruments).

```
## Process runtimes

**type:** `process.runtime`

...

### Go Runtimes

| Value | Description |
| --- | --- |
| `gc` | Go compiler |
| `gccgo` | GCC Go frontend |
```

Only after feedback did I take concern of the attribute values, and yet until now I did not recognize that my initial assumption about namespaces was wrong. Anyways, I see now that `process.runtime.{environment}` namespaces are not defined here. Obviously this changes the character of the PR for me, and for anyone I may have misled.

---

> I do not believe this change is necessary. This affects a value in an attribute that clearly does not relate to garbage collection. For attribute keys created by our runtime metrics instrumentation we already use runtime.go.*. Having a special case for a particular compiler does not make sense to me here.

I assume we still want to specify exactly how `process.runtime.name` and `process.runtime.version` should be obtained in Go and that at question is only whether we should change the value `gc` to `go`. 

I think there is a case for renaming it `go` if there is a chance that Go runtimes will eventually be differentiated into separate namespaces. i.e. If we're eventually going to end up with `process.runtime.go.*` and `process.runtime.gccgo.*`, then we should probably align this value to those namespaces. Any thoughts on whether this is an eventuality?

Aneurysm9:
> I think there is a case for renaming it `go` if there is a chance that Go runtimes will eventually be differentiated into separate namespaces. i.e. If we're eventually going to end up with `process.runtime.go.*` and `process.runtime.gccgo.*`, then we should probably align this value to those namespaces. Any thoughts on whether this is an eventuality?

That actually seems like a reason for not renaming it.  If we made such a change users could know that any metric bearing `runtime.go.*` names did not distinguish them by compiler and that any with `runtime.gc.*` or `runtime.gccgo.* did.  If we wanted to rename anything in that case I'd maybe see using `runtime.go-gc.*` or similar.  

That said, I think it is all irrelevant as the `process.runtime.name` resource attribute should identify the runtime used and that attribute of any metric with a `runtime.go.*` name should be consulted to determine the particular runtime used.  I think we should avoid putting too much responsibility on metric names when we have attributes to enrich their meaning in well-defined ways.

MikeGoldsmith:
> Thank you! Let's leave this PR hanging for couple of days to give chance other @open-telemetry/proto-go-approvers to comment

I'm happy for this change to go ahead too 👍🏻

MikeGoldsmith (approved):

jmacd (approved):
Yes, "gc" is plan9-ism that probably few people in this community actually care about.

SergeyKanzhelev:
@Aneurysm9 how strong is your position? Please indicate if your comment "That actually seems like a reason for not renaming it." is blocking so we can try to find a consensus.

Aneurysm9:
If there is consensus that `go` is more appropriate than `gc` and I am the only opposing voice, then I will not stand in the way of the change.  I would prefer `gc` as it seems to me to be more correct and is certainly simpler to implement, requiring no change to our current implementations.  I also worry about setting a precedent for second-guessing the results of the `runtime.Compiler` value rather than trusting that the compiler authors know how to identify the thing they are creating.

Again, not going to block if I'm the only one that feels that way, but I would like for those that approved prior to [the comment](https://github.com/open-telemetry/opentelemetry-specification/pull/2262#issuecomment-1013434465) @djaglowski made about misinterpreting the purpose of this file to re-assess their approvals.  It seems that the PR was created for one purpose, realized to not be appropriate for that purpose, but maintained anyway. We should assure that it is being approved for the currently intended purpose.

jmacd:
@Aneurysm9 if the variable in question were named `process.runtime.compiler` then I'd agree we should not translate "gc" into "go". The semantic convention is named `runtime.runtime.name` and it feels like an appropriate definition for Go: Use the value of `runtime.Compiler` as the runtime name except for its cryptic value "gc" which indicates the standard runtime, in which case use "go".

PR #2264, opened 2022-01-13
Clarify version support for the OpenTracing compatibility.

### LINKED ISSUE #2218: Document what OpenTracing API versions are supported
We need to document what versions are supported for each language, in order to avoid confusion.

* **Java**: OpenTracing supports Java 7, whereas OpenTelemetry supports Java 8+. Users will need to update to Java 8 before using the OT Shim.
* **Python**: OpenTracing supports Python 2.x whereas OpenTelemetry supports 3.x, which as done as Python 2 is **now fully deprecated**. Likewise, users will need to update to Python 3 before using the OT Shim.
* **Go**: OpenTracing supports [1.13+](https://github.com/opentracing/opentracing-go/blob/master/.github/workflows/go.yml#L19) whereas OpenTelemetry supports [1.16+](https://github.com/open-telemetry/opentelemetry-go/blob/main/README.md#compatibility)
* **Javascript/Node**: OpenTracing supports [6+](https://github.com/opentracing/opentracing-javascript#contributing--developer-information) whereas OpenTelemetry supports [8.5+](https://github.com/open-telemetry/opentelemetry-js-api#you-will-need)
* **CPP**: Not a problem, both support C++11 or later.

cc @yurishkuro

carlosalberto:
Fixes #2218 

A rather minor clarification, but important enough for upcoming users.

cc @yurishkuro @jmacd @tedsuo


specification/compatibility/opentracing.md
```diff
@@ -63,6 +64,29 @@ is not recommended for the following scenarios:
   semantics and may lead to incorrect `Context` usage and incorrect traces.
   See [Implicit and Explicit support mismatch](#implicit-and-explicit-support-mismatch).
 
+## Version Support
```
yurishkuro, line 67:
```suggestion
## Language Version Support
```
Disambiguate, as I first thought this was about SDK versions

specification/compatibility/opentracing.md
```diff
@@ -63,6 +64,29 @@ is not recommended for the following scenarios:
...
+
+Users are encouraged to check and update their language and runtime
+components before using the Shim layer, as the OpenTelemetry APIs and SDKs
+may have higher version requirements than their OpenTracing counterparts.
+e.g. OpenTracing Python supports Python 2.6 and higher, whereas the
+OpenTelemetry Python API supports Python 3.5 and higher.
+
+For reference see the following table, listing the OpenTracing APIs
```
yurishkuro, line 75: I think it would be more useful if the table also included OTEL column, with a clarification that it is as-of Jan 2022, so that we don't have to maintain it.

carlosalberto: Fair enough, will add this as well.

carlosalberto:
Hey @alanwest @cijothomas could you confirm the case for OpenTelemetry DotNet?

specification/compatibility/opentracing.md
```diff
@@ -63,6 +64,30 @@ is not recommended for the following scenarios:
...
+components before using the Shim layer, as the OpenTelemetry APIs and SDKs
+may have higher version requirements than their OpenTracing counterparts.
+e.g. OpenTracing Python supports Python 2.6 and higher, whereas the
+OpenTelemetry Python API supports Python 3.5 and higher.
+
+For reference see the following table, listing the OpenTracing
+and OpenTelemetry APIs minimum **language** version, as of
+January 2022:
+
+| Language   | OpenTracing API  | OpenTelemetry API |
```
yurishkuro, line 79:
```suggestion
| Language   | OpenTracing API  | OpenTelemetry API |
| ---------- | ---------------- | ----------------- |
```
otherwise this won't render as a table

yurishkuro (approved):

jmacd (approved):

PR #2266, opened 2022-01-14
Initial Prometheus <-> OTLP datamodel specification.

dashpole:
This supersedes https://github.com/open-telemetry/opentelemetry-specification/pull/2017, and is meant to capture the current state of prometheus support in prometheus SDK exporters and prometheus collector receivers/exporters, rather than the desired end state.  This is done to avoid (potentially) contentious debates in the initial draft, and allow tackling these topics individually.


specification/metrics/datamodel.md
```diff
@@ -1013,6 +1018,117 @@ For comparison, see the simple logic used in
...
+| `job` | `service.name` | ... |
+| `job` | `job` | ... |
+| `instance` | `instace` | ... |
+| `instance` | `host.name` | instance is split into host:port |
+| `instance` | `port` | instance is split into host:port |
+| `__scheme__` | `scheme` | ... |
+
+### Prometheus Metric points to OTLP
+
+TODO - A bit about detecting/using start_time.
```
jmacd, line 1047: There are several documented ways to handle start time in a Prometheus receiver in the data model meant to help with this, like https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#cumulative-streams-handling-unknown-start-time

specification/metrics/datamodel.md
```diff
@@ -1013,6 +1018,117 @@ For comparison, see the simple logic used in
...
+Multiple Prometheus metrics are merged together into a single metric:
+
+* The `quantile` label on non-suffixed metrics is used to identify quantile points in summary metrics. Each Prometheus metrics produces one quantile on the resulting summary.
+* Lines with `_count` and `_sum` suffixes are used to determine the summary's count and sum.
+
+Prometheus Gauge Histogram is dropped (TBD).
+
+Prometheus Stateset is dropped (TBD).
+
+Prometheus Info is dropped (TBD).
```
jmacd, line 1073: We've been discussing using the Prometheus-specified "target" information metric. I would be happy for these `1`-valued metrics to translate into OTLP cumulative UpDownCounter points, which makes sense if we're using these to report targets. (How many `target` infos are defined right now? It's a non-monotonic sum with approximately the same semantics as the Prometheus `up`.)

specification/metrics/datamodel.md
```diff
@@ -1013,6 +1018,117 @@ For comparison, see the simple logic used in
...
+
+OpenTelemetry Gauge becomes a Prometheus Gauge.
+
+TODO: Example Gauge Conversions
+
+OpenTelemetry Sum follows this logic:
+
+- If the aggregation temporality is cumulative and the sum is monotonic,
+  then it becomes a Prometheus Sum.
+- Otherwise the Sum becomes a Prometheus Gauge.
```
jmacd, line 1085: I think we can refine this. For cumulative non-monotonic sums, Gauge is definitely correct.

For delta monotonic sums, I think the preferred recommendation should be to apply a delta-to-cumulative processor, then export as a Prometheus counter.

For delta non-monotonic sums, these can also be treated with a delta-to-cumulative processor, then exported as a Prometheus gauge. 

The start time will in both cases be the point at which the process or the collector process starts. When the process or collector restarts, it will use a new start time and reset the cumulative state. This means the Prometheus observer of a non-monotonic delta sum from OTLP will see an arbitrary offset. This is OK as long as you only use this information to compute rate information; the cumulative value itself is meaningless.  

It looks like we could summarize these rules as: convert delta temporality points into cumulative points first, then apply the rules for cumulative points to convert into Prometheus.

jmacd: This comment stems from @Aneurysm9's question here: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/7156#discussion_r787222944

alolita: @Aneurysm9 pl review and provide feedback.

alolita (commented):
Nit: Prometheus should be consistently capitalized.

specification/metrics/datamodel.md
Aneurysm9: Should we attempt to define metadata that could be used to ensure that these metrics are converted to Prometheus `Info` metrics in a round-trip?

specification/metrics/datamodel.md
Aneurysm9: I think I agree with Josh's assessment here.

specification/metrics/datamodel.md
dashpole: Thanks for the reference.  I included it in the blurb on the start time.

specification/metrics/datamodel.md
dashpole: +1 to both comments (we should use target_info, and it should be round-trippable).

However, I'd prefer to handle this as a follow-up.  My goal with this initial PR is to capture the current state without making any (potentially contentious) decisions, and then to discuss small changes from the current state individually.

specification/metrics/datamodel.md
dashpole: I think in some places (e.g. SDKs) it isn't appropriate to require prometheus exporters to perform a delta-to-cumulative conversion.  I'd like to phrase it it as being valid to either convert to cumulative temporality or to drop it.

dashpole:
> Nit: Prometheus should be consistently capitalized.

Done

specification/metrics/datamodel.md
Aneurysm9: I think SDKs still need to have state to support Prometheus exporters at all, so it should be possible to do the delta-to-cumulative conversion, but I'm fine with the option to convert or drop.  I would probably prefer SHOULD convert and MAY drop, but wouldn't oppose MAY convert or drop.

dashpole: I think we are on the same page, but maybe you can help me with phrasing.  My intent is to say something like "For client libraries, the SDK is a better place to handle delta-to-cumulative conversion, and the exporter shouldn't have to re-aggregate metrics it doesn't support."  For the collector, it might make sense to do aggregation in the exporter, but I'm less opinionated about that.

Aneurysm9: Yep, agreed there, I think.  If you mean that something in an SDK should be configured to do delta-to-cumulative before handing metrics to the exporter.  I believe that there was a concept of metrics processors that should be able to handle that and, if so, agree that it would be the better place to do it.  If not, though, I think that the exporters will still have the ability to do that since they have to keep some state.

I think the interesting cases for the collector will be when metrics come in from non-Prometheus sources, so it'll still need to be able to handle that conversion even if SDK Prometheus exporters do as well.

dashpole: Sounds like I was imagining capabilities the SDK doesn't have.  I updated this (and delta histograms) to SHOULD aggregate, but MAY drop.

specification/metrics/datamodel.md
```diff
@@ -769,7 +789,7 @@ misbehaving system. Receivers SHOULD presume a single writer was intended and
 eliminate overlap / deduplicate.
 
 Note: Identity is an important concept in most metrics systems.  For example,
-[Prometheus directly calls out uniqueness](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#metric_relabel_configs):
+[Prometheus directly calls out uniqueness](https://Prometheus.io/docs/Prometheus/latest/configuration/configuration/#metric_relabel_configs):
```
Aneurysm9, line 792:
```suggestion
[Prometheus directly calls out uniqueness](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#metric_relabel_configs):
```

Perhaps slightly overzealous `s///g` :)

specification/metrics/datamodel.md
```diff
@@ -1013,6 +1033,146 @@ For comparison, see the simple logic used in
...
+When scraping a Prometheus endpoint, Prometheus service discovery discovers attributes from scraped targets prior to scraping them, and adds these as labels on incoming metrics. In OpenTelemetry, some of these labels are converted to Resource attributes, with the rest added as Metric data stream attributes.
+
+Here is a table of the set of Prometheus labels that are lifted into Resource
+attributes when converting into OpenTelemetry.
+
+| Prometheus Label | OTLP Resource Attribute | Description |
+| -------------------- | ----------------------- | ----------- |
+| `job` | `service.name` | ... |
+| `job` | `job` | ... |
+| `instance` | `instace` | ... |
```
Aneurysm9, line 1055:
```suggestion
| `instance` | `instance` | ... |
```

specification/metrics/datamodel.md
```diff
@@ -1013,6 +1033,146 @@ For comparison, see the simple logic used in
...
+* The `le` label on non-suffixed metrics is used to identify histogram bucket boundaries. Each Prometheus line produces one bucket on the resulting histogram.
+* Lines with `_count` and `_sum` suffixes are used to determine the histogram's count and sum.
+
+#### Summaries
+
+[Prometheus Summary](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#summary) becomes an OTLP Summary.
+
+Multiple Prometheus metrics are merged together into a single metric:
+
+* The `quantile` label on non-suffixed metrics is used to identify quantile points in summary metrics. Each Prometheus metrics produces one quantile on the resulting summary.
```
Aneurysm9, line 1089:
```suggestion
* The `quantile` label on non-suffixed metrics is used to identify quantile points in summary metrics. Each Prometheus line produces one quantile on the resulting summary.
```

For consistency with the above regarding histograms.  Otherwise, `s/s metrics/s metric/`.

Aneurysm9 (approved):

specification/metrics/datamodel.md
```diff
@@ -1013,6 +1033,146 @@ For comparison, see the simple logic used in
...
 
+## Prometheus Compatibility
+
+**Status**: [Experimental](../document-status.md)
+
+This section denotes how to convert metrics scraped in the [Prometheus exposition](https://github.com/Showmax/prometheus-docs/blob/master/content/docs/instrumenting/exposition_formats.md#exposition-formats) or [OpenMetrics](https://openmetrics.io/) formats to the
+OpenTelemetry metric data model and how to create Prometheus metrics from
+OpenTelemetry metric data. Since OpenMetrics has a superset of Prometheus' types, "Prometheus" should be taken to mean "Prometheus or OpenMetrics".  "OpenMetrics" refers to OpenMetrics-only concepts.
+
+### Prometheus Labels from Service Discovery
```
bogdandrutu, line 1044: This is a "service" part of the Prometheus.  What part of "Prometheus" are we trying to be compatible with? This feels a lot like an implementation detail (the fact that we use some of the components from prometheus server in the current collector implementation).

In the https://github.com/Showmax/prometheus-docs/blob/master/content/docs/instrumenting/exposition_formats.md#exposition-formats there is no notion of "Job"/"Instance"/etc.

specification/metrics/datamodel.md
```diff
@@ -1013,6 +1033,146 @@ For comparison, see the simple logic used in
 [statsd sums](https://github.com/statsd/statsd/blob/master/stats.js#L281)
 where all points are added, and lost points are ignored.
 
+## Prometheus Compatibility
+
+**Status**: [Experimental](../document-status.md)
+
+This section denotes how to convert metrics scraped in the [Prometheus exposition](https://github.com/Showmax/prometheus-docs/blob/master/content/docs/instrumenting/exposition_formats.md#exposition-formats) or [OpenMetrics](https://openmetrics.io/) formats to the
```
bogdandrutu, line 1040: Use official prometheus github links not from forks:

https://github.com/prometheus/docs/blob/main/content/docs/instrumenting/exposition_formats.md

dashpole: whoops, fixed.

specification/metrics/datamodel.md
dashpole: You are probably right about this being too closely tied to the prometheus server (e.g. referencing service discovery).  If this were phrased as "scraped prometheus metrics must have X labels added", would that be acceptable?

At minimum, I think metrics exported via prometheus exporters (push or pull) should attach "job" and "instance".  I view them as the equivalent of semantic conventions for prom: https://prometheus.io/docs/concepts/jobs_instances/#jobs-and-instances.  It also makes federation work well, since "job" and "instance" are treated specially when [configuring federation](https://prometheus.io/docs/prometheus/latest/federation/#configuring-federation).  

"job" and "instance" might not be the right resource attributes when converting prom metrics to otel, but I'd like to start with the current state of things, and make that change in a follow-up PR.

bogdandrutu: I see, I think we should separate that in a different section. Fine to do a followup, but would like to have this document easier to read: the format is converted this way, the job/instance labels are converted this way, etc. :)

bogdandrutu: That link is also good to be present somewhere here :)

bogdandrutu (approved):

bogdandrutu:
@dashpole not sure if prometheus let users set their "job" in the metrics source, but if that is the case probably the library sdk prom exporter should convert "service.name" to "job". Also a separate PR is fine.

specification/metrics/datamodel.md
dashpole: I separated this section into Prom -> OTLP and OTLP -> Prom resource attributes sections, and removed references to service discovery or the prometheus server.

dashpole:
> not sure if prometheus let users set their "job" in the metrics source, but if that is the case probably the library sdk prom exporter should convert "service.name" to "job".

It is possible to set "job" in the metrics source, but by default it will be overwritten by service discovery.  `honor_labels` in the prom scrape config allows you to override this, but it is generally intended for use with a federated prom endpoint, rather than an application endpoint.  I'll plan to address this when addressing https://github.com/open-telemetry/opentelemetry-specification/issues/1782 in a follow-up.

specification/metrics/datamodel.md
```diff
@@ -1013,6 +1033,147 @@ For comparison, see the simple logic used in
...
+
+#### Dropped Types
+
+* [OpenMetrics GaugeHistogram](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#gaugehistogram)
+* [OpenMetrics StateSet](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#stateset)
+* [OpenMetrics Info](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#info)
+
+#### Start Time
+
+Prometheus Cumulative metrics do not include the start time of the metric. When converting Prometheus Counters to OTLP, the `process_start_time_metric` may be used to provide the start time. If the start time is not provided by that mechanism, Prometheus conversion follows [Cumulative streams: handling unknown start time](#cumulative-streams-handling-unknown-start-time).
```
brian-brazil, line 1084: This is not correct in the general case, a counter may have been created before the processes started in the case of an exporter or may be started after the process if a new set of labels comes along for a metric. These are both quite common.

specification/metrics/datamodel.md
```diff
@@ -1013,6 +1033,147 @@ For comparison, see the simple logic used in
...
+
+OpenTelemetry Metric Attributes are converted to [Prometheus labels](https://Prometheus.io/docs/concepts/data_model/#metric-names-and-labels).  In Prometheus, metric labels must match the following regex: `[a-zA-Z_:]([a-zA-Z0-9_:])*`.  Metrics
+from OpenTelemetry with unsupported Attribute names should replace invalid characters with the `_` character. This
+may cause ambiguity in scenarios where multiple similar-named attributes share invalid characters at the same
+location.  In such unlikely cases, either value may be used.
+
+#### Exemplars
+
+Exemplars on OpenTelemetry Histograms should be converted to Prometheus exemplars. Exemplars on other OpenTelemetry metric types are dropped.  For Prometheus push exporters, multiple exemplars are permitted on each bucket, so all exemplars are converted.  For Prometheus pull endpoints, only a single exemplar can be added to each bucket, so the largest exemplar from each bucket is chosen.  If no exemplars exist on a bucket, the highest exemplar from a lower bucket should be used, even if it is a duplicate of another bucket.  Prometheus Exemplars should use the `trace_id` and `span_id` keys for the trace and span IDs, respectively.
```
legendecas, line 1169: > Exemplars on OpenTelemetry Histograms should be converted to Prometheus exemplars. Exemplars on other OpenTelemetry metric types are dropped.

I didn't find that the OpenTelemetry exemplars are relevant to the aggregation type. Does this mean that we need notes on https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#exemplar to say that exporters need to know the relevant aggregation type of an exemplar?

dashpole: I'm primarily basing that statement on how exemplars are stored in the [OTLP v1 metrics proto](https://github.com/open-telemetry/opentelemetry-proto/blob/8ba33cceb4a6704af68a4022d17868a7ac1d94f4/opentelemetry/proto/metrics/v1/metrics.proto), which stores exemplars within each metric types' data point.  But based on my reading of the [View API](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#view), the exemplar reservoir exists _within a view_, which has an instrument type. I think that is why the exemplar section itself doesn't mention aggregations or metric types.

specification/metrics/datamodel.md
dashpole: Thanks.  I've removed `process_start_time_seconds` from this section.  I've also opened https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/7411 to track removing this behavior from the collector.

specification/metrics/datamodel.md
legendecas: Sorry I'm a bit confused, is this "OpenTelemetry Histogram" referring to the instrument type rather than "Histogram Aggregation" in the context of exporters? To my understanding, exporters just export aggregated data points, and instrument type is not very useful since their aggregation type can be configured by views API freely.

dashpole: Sorry, I should have said aggregation instead of instrumentation type there, which i've corrected.  By OpenTelemetry Histogram, I'm actually referencing the [Histogram Data Point Type](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#histogram) above.  I've added links to that and other data point types to make it clearer.

specification/metrics/datamodel.md
```diff
@@ -1013,6 +1033,147 @@ For comparison, see the simple logic used in
...
+
+[Prometheus Counter](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#counter) becomes an OTLP Sum.
+
+#### Gauges
+
+[Prometheus Gauge](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#gauge) becomes an OTLP Gauge.
+
+#### Unknown-typed
+
+[Prometheus Unknown](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#unknown) becomes an OTLP Gauge.
```
jsuereth, line 1056: Should we preserve in anyway the notion that this was originally an "unknown" value?

jsuereth (approved):

specification/metrics/datamodel.md
dashpole: That would be good to do, but I don't think there are any simple solutions.  Since the topic will require more discussion, I'd prefer to stick with the current state, and tackle the topic as a follow-up.

jsuereth: Can you open a bug (either in Prometheus WG or Otel Metrics Data model) about being able to round-trip an "unknown" type preserving its "unknown"-ness?

dashpole: https://github.com/open-telemetry/wg-prometheus/issues/69

specification/metrics/datamodel.md
```diff
@@ -1013,6 +1033,147 @@ For comparison, see the simple logic used in
...
+
+#### Exemplars
+
+[Prometheus Exemplars](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#exemplars) may be attached to Prometheus Histogram bucket metrics, which should be converted to exemplars on OpenTelemetry histograms.  The Trace ID and Span ID should be retrieved from the `trace_id` and `span_id` label keys, respectively.
+
+#### Resource Attributes
+
+When scraping a Prometheus endpoint, resource attributes must be added to the scraped metrics to distinguish them from metrics from other Prometheus endpoints.  In particular, `job` and `instance`, [as defined by Prometheus](https://Prometheus.io/docs/concepts/jobs_instances/#jobs-and-instances), are needed to ensure Prometheus exporters can disambiguate metrics as [described below](#resource-attributes-1).
+
+The following resource attributes must be added to scraped metrics, and may not be added as metric attributes:
```
MrAlias, line 1094: Can we make this statement normative?

Is the second clause intended to be a requirement?

```suggestion
The following attributes MUST be associated with scraped metrics as resource attributes, and MUST NOT be added as attributes to the metric instruments themselves:
```

specification/metrics/datamodel.md
```diff
@@ -1013,6 +1033,147 @@ For comparison, see the simple logic used in
...
+
+| OTLP Resource Attribute | Description |
+| ----------------------- | ----------- |
+| `service.name` | The configured name of the service that the target belongs to |
+| `job` | Identical to `service.name` |
+| `instance` | The <host>:<port> of the target's URL that was scraped. |
+| `host.name` | `instance` is split into <host>:<port> |
+| `port` | `instance` is split into <host>:<port> |
```
MrAlias, line 1102:
```suggestion
| `host.name` | The <host> portion of `instance` |
| `port` | The <host> portion of `instance` if it exists |
```

specification/metrics/datamodel.md
```diff
@@ -1013,6 +1033,147 @@ For comparison, see the simple logic used in
...
+When scraping a Prometheus endpoint, resource attributes must be added to the scraped metrics to distinguish them from metrics from other Prometheus endpoints.  In particular, `job` and `instance`, [as defined by Prometheus](https://Prometheus.io/docs/concepts/jobs_instances/#jobs-and-instances), are needed to ensure Prometheus exporters can disambiguate metrics as [described below](#resource-attributes-1).
+
+The following resource attributes must be added to scraped metrics, and may not be added as metric attributes:
+
+| OTLP Resource Attribute | Description |
+| ----------------------- | ----------- |
+| `service.name` | The configured name of the service that the target belongs to |
+| `job` | Identical to `service.name` |
+| `instance` | The <host>:<port> of the target's URL that was scraped. |
+| `host.name` | `instance` is split into <host>:<port> |
```
MrAlias, line 1101: [Other semantic conventions distinguish between `host.name` and `host.ip`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#general-network-connection-attributes), should the same distinction be made here?

MrAlias: Should we try to reuse existing semantic conventions here instead of defining new values?

specification/metrics/datamodel.md
```diff
@@ -1013,6 +1033,147 @@ For comparison, see the simple logic used in
...
+
+#### Histograms
+
+[OpenTelemetry Histograms](#histogram) with a cumulative aggregation temporality becomes a metric family with the following:
+
+- A single `{name}_count` metric denoting the count field of the histogram.
+  All attributes of the histogram point are converted to Prometheus labels.
+- `{name}_sum` metric denoting the sum field of the histogram, reported
+  only if the sum is positive and monotonic. All attributes of the histogram
+  point are converted to Prometheus labels.
```
MrAlias, line 1128: Prometheus labels values can only be strings, OpenTelemetry attributes can be strings, integers, or floating points, or arrays of these. For the compatibility of different language implementations of this it seems like we need a specification defining how to convert these types to a string.

Can we do so here, or in a follow on PR?

If we want to tackle this in a follow on, can we note that deficiency here?

specification/metrics/datamodel.md
```diff
@@ -1013,6 +1033,147 @@ For comparison, see the simple logic used in
...
+- A series of `{name}` metric points that contain all attributes of the
+  histogram point recorded as labels.  Additionally, a label, denoted as `le`
+  is added denoting a bucket boundary, and having its value be the stringified
+  floating point value of bucket boundaries, starting form lowest to highest,
```
MrAlias, line 1132:
```suggestion
  floating point value of bucket boundaries, starting from lowest to highest,
```

specification/metrics/datamodel.md
```diff
@@ -1013,6 +1033,147 @@ For comparison, see the simple logic used in
...
+  and all being non-negative.  The value of each point is the sum of the count
```
MrAlias, line 1133: Does this suggest negative buckets are dropped? Do the sum and count metrics need to be updated to account for this?

specification/metrics/datamodel.md
```diff
@@ -1013,6 +1033,147 @@ For comparison, see the simple logic used in
...
+  of all histogram buckets up the the boundary reported in the `le` label.
+  These points will include a single exemplar that falls within `le` label and
+  no other `le` labelled point.
```
MrAlias, line 1136: Is the `le="+Inf"` bucket required?

specification/metrics/datamodel.md
```diff
@@ -1013,6 +1033,147 @@ For comparison, see the simple logic used in
...
+#### Dropped Data Points
+
+* [ExponentialHistogram](#exponentialhistogram)
+
+#### Metric Attributes
+
+OpenTelemetry Metric Attributes are converted to [Prometheus labels](https://Prometheus.io/docs/concepts/data_model/#metric-names-and-labels).  In Prometheus, metric labels must match the following regex: `[a-zA-Z_:]([a-zA-Z0-9_:])*`.  Metrics
+from OpenTelemetry with unsupported Attribute names should replace invalid characters with the `_` character. This
+may cause ambiguity in scenarios where multiple similar-named attributes share invalid characters at the same
+location.  In such unlikely cases, either value may be used.
```
MrAlias, line 1165: Allowing either value to be use will result in different implementations producing potentially producing different results. Would it make more sense to drop both in such a case?

MrAlias (commented):
The conversions statements here are not made using the [normative language conventions this specification is defined to use](https://github.com/open-telemetry/opentelemetry-specification#notation-conventions-and-compliance). Can we update these statements to make clear requirements and recommendations about these conversions?

specification/metrics/datamodel.md
dashpole: Yes and Yes. Since this describes how to convert between data models, i'm going to omit referring to instruments.

specification/metrics/datamodel.md
dashpole: This is under discussion here: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/7081.  Until a decision is made, i'm hoping to stick with what the collector currently does.  I agree with reusing existing conventions, and would like to handle that in a follow-up.

specification/metrics/datamodel.md
dashpole: They should not be dropped, but we need to omit the _sum metric if there are negative buckets. We also need to handle missing _sum metrics on histograms in the collector.  I've opened https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/7546 to track that.

specification/metrics/datamodel.md
dashpole: Done

specification/metrics/datamodel.md
dashpole: It is.  I've clarified that.

specification/metrics/datamodel.md
dashpole: Thinking about it more, I'm not sure either option (using either or dropping both) is safe, since both can result in two metric streams colliding.  Should we concatenate the values together with a standard (e.g. `;`) separator instead?

dashpole:
> Can we update these statements to make clear requirements and recommendations about these conversions?

Done.

specification/metrics/datamodel.md
dashpole: Let's do it as a follow-up.  For now, I think it is reasonable to say that non-string attributes should be replaced with "".

specification/metrics/datamodel.md
```diff
@@ -1013,6 +1033,143 @@ For comparison, see the simple logic used in
...
+- A series of `{name}` metric points that contain all attributes of the
+  summary point recorded as labels.  Additionally, a label, denoted as
+  `quantile` is added denoting a reported quantile point, and having its value
+  be the stringified floating point value of quantiles (between 0.0 and 1.0),
+  starting from lowest to highest, and all being non-negative.  The value of
+  each point is the computed value of the quantile point.
+
+#### Dropped Data Points
+
+The following OTLP data points must be dropped:
```
MrAlias, line 1152:
```suggestion
The following OTLP data points MUST be dropped:
```

specification/metrics/datamodel.md
```diff
@@ -1013,6 +1033,143 @@ For comparison, see the simple logic used in
...
+
+The following attributes MUST be associated with scraped metrics as resource attributes, and MUST NOT be added as metric attributes:
+
+| OTLP Resource Attribute | Description |
+| ----------------------- | ----------- |
+| `service.name` | The configured name of the service that the target belongs to |
+| `job` | Identical to `service.name` |
+| `instance` | The <host>:<port> of the target's URL that was scraped. |
+| `host.name` | `instance` The <host> portion of `instance` |
+| `port` | `instance` The <port> portion of `instance` |
```
MrAlias, line 1106: nit: is including `instance` intentional?

```suggestion
| `host.name` | The <host> portion of `instance` |
| `port` | The <port> portion of `instance` |
```

specification/metrics/datamodel.md
MrAlias: > Thinking about it more, I'm not sure either option (using either or dropping both) is safe, since both can result in two metric streams colliding. Should we concatenate the values together with a standard (e.g. `;`) separator instead?

Good point. I think concatenation is a valid approach as long as we include a recommended order (i.e. lexicographical ordered concatenation).

MrAlias (approved):

specification/metrics/datamodel.md
dashpole: Aside: What we do today in the PRW exporter is buggy: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/7547

gracewehner (approved):

specification/metrics/datamodel.md
```diff
@@ -1013,6 +1033,143 @@ For comparison, see the simple logic used in
...
+
+An [OpenTelemetry Gauge](#gauge) MUST be converted to a Prometheus Gauge.
+
+#### Sums
+
+[OpenTelemetry Sums](#sums) follows this logic:
+
+- If the aggregation temporality is cumulative and the sum is monotonic,
+  it MUST be converted to a Prometheus Sum.
+- If the aggregation temporality is delta and the sum is monotonic, it SHOULD be converted to a cumulative temporality and become a Prometheus Sum, or it MAY be dropped.
```
anuraaga, line 1121: Since dropped is `MAY`, what is expected in the case it isn't dropped, reporting the delta number? FWIU, this isn't compatible with Prometheus, it seems like this (and probably related bits in other instrument types) is supposed to be something like this (MUST drop if can't make it cumulative)

```
If the aggregation temporality is cumulative
  - If the sum is monotonic, it MUST be converted to a Prometheus SUM
  - Otherwise, it MUST be converted to a Prometheus Gauge
Otherwise
  - The value SHOULD be converted to cumulative temporality and follow the rules above
  - Otherwise, it MUST drop the value
```

specification/metrics/datamodel.md
dashpole: How does this look:

> In such unlikely cases, if multiple key-value pairs are converted to have the same Prometheus key, the values MUST be concatenated together, separated by `;`, and ordered by the lexicographical order of the original keys.

That would ensure ordering is stable across time.

specification/metrics/datamodel.md
dashpole: You are correct.  I've updated it to match that structure.

specification/metrics/datamodel.md
quentinmit: "may be started after the process if a new set of labels comes along for a metric" does not make the process start time wrong; the start time identifies "A time window (of (start, end]) time for which the Sum was calculated."

If the counter doesn't exist until a certain time, its value before that time is necessarily 0. A counter value represents the number of events that have occurred in a time window, so you can extend that time window if you know no events happened in the additional time window.

brian-brazil: > If the counter doesn't exist until a certain time, its value before that time is necessarily 0

Not so, it could have existed and been reset. Such as if a process restarted, or much far commonly if it were reset during the lifetime of a process.

quentinmit: If the process restarted, it has a new `process_start_time_seconds` so it's still correct to use that metric as the counter's start time. If you're manually destroying and recreating the same counter with the same name and labels, that's the same case as "a counter may have been created before the processes started in the case of an exporter" - it's a counter not tied to the process lifetime. That case is definitely possible (though I'm not sure why one would do that unless they're implementing an exporter); I was only talking about the second case you mentioned where new label values come into existence after a process's start.

It's not correct in the general case to use `process_start_time_seconds` as the start time of counters, but it is correct in the 95% or even 99% case of a normal in-process Prometheus counter.

specification/metrics/datamodel.md
```diff
@@ -1013,6 +1033,140 @@ For comparison, see the simple logic used in
...
+
+- A single `{name}_count` metric denoting the count field of the histogram. All attributes of the histogram point are converted to Prometheus labels.
+- `{name}_sum` metric denoting the sum field of the histogram, reported only if the sum is positive and monotonic. The sum is positive and monotonic when all buckets are positive. All attributes of the histogram point are converted to Prometheus labels.
+- A series of `{name}` metric points that contain all attributes of the histogram point recorded as labels.  Additionally, a label, denoted as `le` is added denoting the bucket boundary. The label's value is the stringified floating point value of bucket boundaries, ordered from lowest to highest. The value of each point is the sum of the count of all histogram buckets up the the boundary reported in the `le` label. These points will include a single exemplar that falls within `le` label and no other `le` labelled point.  The final bucket metric MUST have an `+Inf` threshold.
+
+OpenTelemetry Histograms with Delta aggregation temporality SHOULD be aggregated into a Cumulative aggregation temporality and follow the logic above, or MAY be dropped.
```
anuraaga, line 1132: MUST be dropped here too?

dashpole: fixed

specification/metrics/datamodel.md
dashpole: Discussed at the sig meeting today.  It should be allowable to use process start time, but it must not be enabled by default.

carlosalberto:
This looks like it's ready to go. Please confirm there's no blocker at this moment @brian-brazil @legendecas @anuraaga @jmacd

anuraaga (approved):

specification/metrics/datamodel.md
```diff
@@ -1013,6 +1033,140 @@ For comparison, see the simple logic used in
...
+
+#### Exemplars
+
+[Exemplars](#exemplars) on OpenTelemetry Histograms SHOULD be converted to Prometheus exemplars. Exemplars on other OpenTelemetry data points MUST be dropped.  For Prometheus push exporters, multiple exemplars are able to be added to each bucket, so all exemplars SHOULD be converted.  For Prometheus pull endpoints, only a single exemplar is able to be added to each bucket, so the largest exemplar from each bucket MUST be used, if attaching exemplars.  If no exemplars exist on a bucket, the highest exemplar from a lower bucket MUST be used, even though it is a duplicate of another bucket's exemplar.  Prometheus Exemplars MUST use the `trace_id` and `span_id` keys for the trace and span IDs, respectively.
+
+#### Resource Attributes
+
+In SDK Prometheus (pull) exporters, all resource attributes MUST be dropped, and MUST NOT be not attached as labels. The scraper of the endpoint is expected to discover resource attributes of the endpoint it is scraping.
```
legendecas, line 1166:
```suggestion
In SDK Prometheus (pull) exporters, all resource attributes MUST be dropped, and MUST NOT be attached as labels. The scraper of the endpoint is expected to discover resource attributes of the endpoint it is scraping.
```

legendecas (approved):
👍nice works!

specification/metrics/datamodel.md
```diff
@@ -1013,6 +1033,140 @@ For comparison, see the simple logic used in
...
+
+The following Prometheus types MUST be dropped:
+
+* [OpenMetrics GaugeHistogram](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#gaugehistogram)
+* [OpenMetrics StateSet](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#stateset)
+* [OpenMetrics Info](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#info)
+
+#### Start Time
+
+Prometheus Cumulative metrics do not include the start time of the metric. When converting Prometheus Counters to OTLP, conversion MUST follow [Cumulative streams: handling unknown start time](#cumulative-streams-handling-unknown-start-time) by default. Conversion MAY offer configuration, disabled by default, which allows using the `process_start_time_seconds` metric to provide the start time.
```
jmacd, line 1088: I wonder if we should provide more guidance on this. See https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/7411#issuecomment-1035671933

specification/metrics/datamodel.md
```diff
@@ -1013,6 +1033,140 @@ For comparison, see the simple logic used in
...
+
+#### Gauges
+
+An [OpenTelemetry Gauge](#gauge) MUST be converted to a Prometheus Gauge.
+
+#### Sums
+
+[OpenTelemetry Sums](#sums) follows this logic:
+
+- If the aggregation temporality is cumulative and the sum is monotonic, it MUST be converted to a Prometheus Sum.
```
jmacd, line 1119:
```suggestion
- If the aggregation temporality is cumulative and the sum is monotonic, it MUST be converted to a Prometheus Counter.
```

specification/metrics/datamodel.md
```diff
@@ -1013,6 +1033,140 @@ For comparison, see the simple logic used in
...
+
+#### Dropped Data Points
+
+The following OTLP data points MUST be dropped:
+
+* [ExponentialHistogram](#exponentialhistogram)
+
+#### Metric Attributes
+
+OpenTelemetry Metric Attributes MUST be converted to [Prometheus labels](https://Prometheus.io/docs/concepts/data_model/#metric-names-and-labels).  String Attribute values are converted directly to Metric Attributes, and non-string Attribute values SHOULD be replaced with an empty string.  Prometheus metric label keys are required to match the following regex: `[a-zA-Z_:]([a-zA-Z0-9_:])*`.  Metrics from OpenTelemetry with unsupported Attribute names MUST replace invalid characters with the `_` character. This may cause ambiguity in scenarios where multiple similar-named attributes share invalid characters at the same location.  In such unlikely cases, if multiple key-value pairs are converted to have the same Prometheus key, the values MUST be concatenated together, separated by `;`, and ordered by the lexicographical order of the original keys.
```
jmacd, line 1158: > non-string Attribute values SHOULD be replaced with an empty string

This sounds dangerous. Shouldn't we convert non-string values to string, at least as an option?

specification/metrics/datamodel.md
dashpole: Anything in particular you want to call out?  Right now i've added:`Using `process_start_time_seconds` is only correct when all counters on the target start after the process and are not reset while the process is running.`

specification/metrics/datamodel.md
dashpole: Agreed.  That is unfortunately the current behavior today: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/7878. It looks like we are required to use the json representation of arrays in the [attribute spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attribute), so we should follow that here.  I'm wondering if we should just point at that spec here?  I.e. `String Attribute values are converted directly to Metric Attributes, and non-string Attribute values MUST be converted to string attributes following the [attribute specification](link).`  We would just need to expand the attribute spec to specify how to convert non-string primitives to strings...

jsuereth: Yeah, that was a TODO we had deferred for the prometheus exporter specification, actually.  I can pull up the bugs for it, but you'll want to NOT drop the attribtues and instead specify string mapping.

dashpole: Changed to convert to string according to attributes spec.

dashpole: https://github.com/open-telemetry/opentelemetry-specification/pull/2343

PR #2269, opened 2022-01-18
[DoNotMerge] service.name can default to the main program module name

pellared:
I propose to NOT MERGE this PR until [the feature](https://github.com/open-telemetry/opentelemetry-dotnet/issues/2781) it is implemented in [OTel .NET SDK](https://github.com/open-telemetry/opentelemetry-dotnet). 

However, **I would like to first get feedback if such implementation can be acceptable from the specification perspective**.

## Changes

It may be convenient to use [`Assembly.EntryPoint`](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.getentryassembly) the `service.name` resource if it is not explicitly provided. It is especially useful when the processes are run via `dotnet run` which currently sets the process name as `dotnet` instead of the name of the assembly being run.

Other use-cases where it could be handy are:
- running Python scripts like: `python3 myprogam.py`
- running Java program via JAR: `java my.program.jar`

Related issue https://github.com/open-telemetry/opentelemetry-dotnet/issues/2781

FYI @open-telemetry/dotnet-approvers @pjanotti @cijothomas

### DIFF
--- a/semantic_conventions/resource/service.yaml
+++ b/semantic_conventions/resource/service.yaml
@@ -12,8 +12,10 @@ groups:
         note: >
           MUST be the same for all instances of horizontally scaled services.
           If the value was not specified, SDKs MUST fallback to `unknown_service:` concatenated
-          with [`process.executable.name`](process.md#process), e.g. `unknown_service:bash`.
-          If `process.executable.name` is not available, the value MUST be set to `unknown_service`.
+          with [`process.executable.name`](process.md#process), e.g. `unknown_service:bash`
+          or the name of the main program module like .NET entry assembly name, e.g.
+          `unknown_service:MyProgram`. If none of these values is not available, the value MUST
+          be set to `unknown_service`.
         examples: ["shoppingcart"]
       - id: namespace
         type: string
--- a/specification/resource/semantic_conventions/README.md
+++ b/specification/resource/semantic_conventions/README.md
@@ -70,7 +70,9 @@ as specified in the [Resource SDK specification](../sdk.md#sdk-provided-resource
 | `service.instance.id` | string | The string ID of the service instance. [3] | `627cc493-f310-47de-96bd-71410b7dec09` | No |
 | `service.version` | string | The version string of the service API or implementation. | `2.0.0` | No |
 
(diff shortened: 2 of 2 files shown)


iNikem:
I think this proposal makes sense, but requires better explanation what does `main program module` mean. Probably with some examples as well.

tigrannajaryan:
I like the spirit of this change.

What I believe it says is that the SDKs should not be forced to set the service name to `unknown_service:$process.executable.name` if they have a better name available (e.g. the assembly name for .NET apps).

I would like the PR to go even further and say that if the SDK has a good way to get the service name from somewhere (i.e. the assembly name) then that is preferable to use. `unknown_service:$process.executable.name` should be last resort.

yurishkuro:
> I would like the PR to go even further and say that if the SDK has a good way to get the service name from somewhere (i.e. the assembly name) then that is preferable to use.

If the SDK has some built-in capability of deriving a service name, I think it should be implemented as a configurable service name resolver. I don't like having an unchangeable behavior in the SDK. Assembly name might be ok for some people, but there are many pieces of infrastructure that make use of the notion of a service name, and I would rather telemetry use that shared notion.

semantic_conventions/resource/service.yaml
```diff
@@ -12,8 +12,10 @@ groups:
         note: >
           MUST be the same for all instances of horizontally scaled services.
           If the value was not specified, SDKs MUST fallback to `unknown_service:` concatenated
-          with [`process.executable.name`](process.md#process), e.g. `unknown_service:bash`.
-          If `process.executable.name` is not available, the value MUST be set to `unknown_service`.
+          with [`process.executable.name`](process.md#process), e.g. `unknown_service:bash`
+          or the name of the main program module like .NET entry assembly name, e.g.
+          `unknown_service:MyProgram`. If none of these values is not available, the value MUST
```
SergeyKanzhelev, line 17: > If none of these values is not

let's avoid double negatives

SergeyKanzhelev:
the implementation of defaulting to certain value ideally needs to be done via resolver as @yurishkuro suggested so the `Merge` rules would apply correctly: https://github.com/open-telemetry/opentelemetry-specification/blob/da74730bf835010229a9612c281f052e26553218/specification/resource/sdk.md#detecting-resource-information-from-the-environment It may be expected to have unknown value in some environment without defaulting to the process name

PR #2270, opened 2022-01-18
Specify the treatment of duplicate instruments and observations

### LINKED ISSUE #2226: Meter and instrument identity

In the metrics API spec it is said for `MeterProvider.getMeter`:

> Implementations MUST NOT require users to repeatedly obtain a Meter again with the same name+version+schema_url to pick up configuration changes. This can be achieved either by allowing to work with an outdated configuration or by ensuring that new configuration applies also to previously returned Meters.
> https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#get-a-meter

It also suggests that:

> It is unspecified whether or under which conditions the same or different Meter instances are returned from this function.

So, as long as SDK propagates the configurations to all meters created without another `getMeter` call, it should be spec-compliant for either returning the same or a new Meter instance for every name+version+schema_url pair.

However, for the instrument name in API spec, it is said:

> Meter implementations MUST return an error when multiple Instruments are registered under the same Meter instance using the same name.
> https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument

It is not clear when the SDK always returning a new Meter instance for every name+version+schema_url pair, if the instrument name should be resolved in the Meter instance, or it should be checked SDK-wide with the same meter name+version+schema_url? If it is resolved within a single Meter instance, users could create multiple meter instances with the same meter name+version+schema_url and create an instrument with the same name in these meters. In that case, I'm not sure how the unique instrument name constraints can effectively work, since the exporters may observe multiple same name instruments associated with identical instrumentation libraries.

### LINKED ISSUE #2249: Clarify whether multiple callbacks are allowed for async instruments

Clarify behavior of metrics API / SDK when multiple callbacks are registered to the same instrument. The [api spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument) says: 

> Meter implementations MUST return an error when multiple Instruments are registered under the same Meter instance using the same name.

In java, we interpret that as meaning that you can't register conflicting instruments under the same name. However, if the instrument type, name, unit, and description match, we'll return the previously registered instrument.

For example, in the following snippet, no error is generated and `counter1` and `counter` two are the instrument. 
```
    LongCounter counter1 = meter.counterBuilder("foo").build();
    LongCounter counter2 = meter.counterBuilder("foo").build();
```

This makes sense for synchronous instruments, but its not clear what to do in the async case. Currently, if multiple callbacks are registered for the same instrument, we silently ignore registrations after the first. This isn't great, and minimally we should log an error when subsequent callbacks are registered. However, there's also an argument to made that registering multiple callbacks should be allowed and should result in each being invoked.

This has been raised in [Issue #4051](https://github.com/open-telemetry/opentelemetry-java/issues/4051) and we believe it needs raising in the spec for resolution.

In particular, we want to make sure that our plans to publish a stable version of the metrics API would not be impacted by the spec resolution of this issue. 

**Additional context.**

[Issue #4005](https://github.com/open-telemetry/opentelemetry-java/issues/4005) aims to log a warning when multiple callbacks are registered, and is implemented in [Issue #4020](https://github.com/open-telemetry/opentelemetry-java/pull/4020).

jmacd:
Fixes #2226 
Fixes #2249

## Changes

This changes the recommendation to return an error for duplicate instrument registration into a case-by-case which permits identically-registered instruments and specifies which fields are considered.  Implementations are not required to return the same instances in these cases, only to return valid instruments.  

Note that this changes a specified error into a non-error condition. We view this as OK because the semantics are preserved and the user should not be able to detect the difference.

This adds notes to the SDK specification to remind the implementation that despite this flexibility, they are still required to follow the single-writer rule.

## Additional info

This was discussed in the 1/18/22 OTel Specification SIG.

### DIFF
--- a/specification/metrics/api.md
+++ b/specification/metrics/api.md
@@ -288,6 +288,16 @@ Please note that the term _synchronous_ and _asynchronous_ have nothing to do
 with the [asynchronous
 pattern](https://en.wikipedia.org/wiki/Asynchronous_method_invocation).
 
+When asynchronous instruments are used with duplicate registrations,
+meaning to have more than one callback provided, the order of callback
+execution is unspecified.  Callers SHOULD avoid making duplicate
+observations from asynchronous instrument callbacks.  However, if a
+user accidentally makes duplicate observations, meaning to provide
+more than one value for a given asynchronous instrument and attribute
+set, the caller SHOULD expect the last-observed value to be the one
+recorded.  This is sometimes referred to as "last-value wins" for
+asynchronous instruments.
+
 ### Counter
 
 `Counter` is a [synchronous Instrument](#synchronous-instrument) which supports
--- a/specification/metrics/api.md
+++ b/specification/metrics/api.md
@@ -180,14 +180,33 @@ will have the following information:
 * An optional `unit` of measure
 * An optional `description`
 
-Instruments are associated with the Meter during creation, and are identified by
-the name:
+<a name="duplicate-instrument-handling"></a>
 
-* Meter implementations MUST return an error when multiple Instruments are
-  registered under the same Meter instance using the same name.
-* Different Meters MUST be treated as separate namespaces. The names of the
-  Instruments under one Meter SHOULD NOT interfere with Instruments under
-  another Meter.
+Instruments are associated with the Meter during creation and are
+identified by instrument name.  Duplicate registration of
+identically-named instruments within a Meter is treated as follows:
+
+* If the registration is semantically identical, meaning to have the
+  same `kind` and `unit` as a previously registered instrument of the
+  same `name`, the implementation MUST return a valid instrument.
+  It is unspecified whether or under which conditions the same or different
+  instrument instance will be returned as a result of duplicate registration.
+* If the registration is semantically not identical, the
+  implementation MUST return a registration error to the user when the
+  instrument is constructed.
+
+When determining whether a duplicate registration is valid,
+language-level features such the distinction between integer and
+floating point numbers SHOULD be considered.  Equivalently,
+implementations SHOULD NOT support an application in emitting both
+integer and floating point numbers for the same metric inside the same
+instrumentation library.
+
+<a name="instrument-namespace"></a>
+
+Different Meters MUST be treated as separate namespaces. The names of the
+Instruments under one Meter SHOULD NOT interfere with Instruments under
+another Meter.
 
 <a name="instrument-naming-rule"></a>
 
@@ -238,6 +257,11 @@ instrument. It MUST be treated as an opaque string from the API and SDK.
   support more Unicode [Planes](https://en.wikipedia.org/wiki/Plane_(Unicode)).
 * It MUST support at least 1023 characters. [OpenTelemetry
   API](../overview.md#api) authors MAY decide if they want to support more.
+
+Note the `description` property of an instrument is explicitly
+disregarded when considering duplicate registration, because it is not
+semantic in nature.  Implementations SHOULD use any of the provided
+`description` values when emitting metrics that had duplicate registrations.
 
 Instruments can be categorized based on whether they are synchronous or
 asynchronous:
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -975,4 +975,6 @@ single-writer rule.
 As another example, users are encouraged not to make duplicate
 observations from asynchronous instrument callbacks.  However,
 implementations MUST NOT violate the single-writer rule even when
-users make duplicate observations.
+users make duplicate observations.  This is also covered in the
+[supplemental guidelines for handling asynchronous instrument
+views](#asynchronous-example-attribute-removal-in-a-view).
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -952,3 +952,22 @@ called concurrently.
 
 **MetricExporter** - `ForceFlush` and `Shutdown` are safe to be called
 concurrently.
+
+## Data model requirements
+
+The implementation is required to respect the OpenTelemetry Metrics
+data model [Single Writer](datamodel.md#single-writer) requirement.
+This rule stipulates that the implementation MUST avoid creating
+duplicate output streams from a given SDK.
+
+This rule is the basis of the output-name uniqueness check in for
+[Views](#view) above, and it also constrains how duplicate instrument
+registration is handled.
+
+For example, the implementation is not required to return the
+identical instrument when a duplicate instrument is registered, but
+assuming it does allow separate instances to co-exist, the
+implementation MUST eliminate the duplication at a later point using
+the [natural merge function](#opentelemetry-protocol-data-model) for
+those data points, as otherwise it would risk violating the
+single-writer rule.
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -971,3 +971,8 @@ implementation MUST eliminate the duplication at a later point using
 the [natural merge function](#opentelemetry-protocol-data-model) for
 those data points, as otherwise it would risk violating the
 single-writer rule.
+
+As another example, users are encouraged not to make duplicate
+observations from asynchronous instrument callbacks.  However,
+implementations MUST NOT violate the single-writer rule even when
+users make duplicate observations.


specification/metrics/api.md
```diff
@@ -264,6 +288,16 @@ Please note that the term _synchronous_ and _asynchronous_ have nothing to do
 with the [asynchronous
 pattern](https://en.wikipedia.org/wiki/Asynchronous_method_invocation).
 
+When asynchronous instruments are used with duplicate registrations,
+meaning to have more than one callback provided, the order of callback
+execution is unspecified.  Callers SHOULD avoid making duplicate
```
mateuszrzeszutek, line 293: Does this mean that registering multiple callbacks in a single async instrument is officially supported? This is related to issue #2249 - it is currently unclear whether this behavior is allowed; and the Java SDK rejects duplicate callbacks and only calls the first one.

jmacd: That was my intention, yes. There's a historical perspective on this: OpenCensus permitted duplicate instrument registration and it was part of the early thinking for synchronous instruments. The only reason duplicate-identical instruments ever seemed contentious to me is due to asynchronous instruments. The question is whether users are harmed or helped by allowing this, and I didn't have an answer to this question.

I believe #2249 is suggesting the answer ought to favor allowing duplicate registration and multiple callbacks. I have come to adopt this position as well, but it is entirely driven by thinking this is best for users, i.e., that it creates less confusion. This is independent of the question of whether and how an SDK is required to treat duplicate observations. That is, even with duplicate callbacks the SDK should be doing something to prevent duplicate observations from being double-counted.

mateuszrzeszutek: > I believe #2249 is suggesting the answer ought to favor allowing duplicate registration and multiple callbacks. I have come to adopt this position as well, but it is entirely driven by thinking this is best for users, i.e., that it creates less confusion.

Yes, I can agree with that -- having recently implemented a bridge from another metrics API to OTel, I can confirm that not being able to register multiple callbacks just forces the instrumentation author to manage that themselves. And that can get unwieldy and hacky.

mateuszrzeszutek (approved):

specification/metrics/api.md
```diff
@@ -180,14 +180,33 @@ will have the following information:
...
+* If the registration is semantically not identical, the
+  implementation MUST return a registration error to the user when the
+  instrument is constructed.
+
+When determining whether a duplicate registration is valid,
+language-level features such the distinction between integer and
+floating point numbers SHOULD be considered.  Equivalently,
+implementations SHOULD NOT support an application in emitting both
+integer and floating point numbers for the same metric inside the same
+instrumentation library.
```
carlosalberto, line 203: under (belonging to?) the same `Meter`? See #2276

specification/metrics/api.md
```diff
@@ -180,14 +180,33 @@ will have the following information:
...
+
+<a name="instrument-namespace"></a>
+
+Different Meters MUST be treated as separate namespaces. The names of the
+Instruments under one Meter SHOULD NOT interfere with Instruments under
+another Meter.
```
jack-berg, line 209: Is this compatible with the following phrasing describing meters? 

> It is unspecified whether or under which conditions the same or different Meter instances are returned from this functions.

If its undefined whether or not a meter provider must return the same or a separate instance of a meter for the same name, then an implementation might end up with two different meters for the same name. This would allow instruments to have the same name, but different type and unit under meters which have the same name.

We could resolve this by adding language that clarified that the identity of the meter is tied to the meter name, similar to the new phrasing here "Instruments are associated with the Meter during creation and are
identified by instrument name".

specification/metrics/sdk.md
```diff
@@ -952,3 +952,29 @@ called concurrently.
...
+
+This rule is the basis of the output-name uniqueness check in for
+[Views](#view) above, and it also constrains how duplicate instrument
+registration is handled.
+
+For example, the implementation is not required to return the
+identical instrument when a duplicate instrument is registered, but
+assuming it does allow separate instances to co-exist, the
+implementation MUST eliminate the duplication at a later point using
+the [natural merge function](#opentelemetry-protocol-data-model) for
```
arminru, line 971:
```suggestion
the [natural merge function](datamodel.md#opentelemetry-protocol-data-model) for
```
This is in the other file.

specification/metrics/sdk.md
```diff
@@ -952,3 +952,29 @@ called concurrently.
...
+those data points, as otherwise it would risk violating the
+single-writer rule.
+
+As another example, users are encouraged not to make duplicate
+observations from asynchronous instrument callbacks.  However,
+implementations MUST NOT violate the single-writer rule even when
+users make duplicate observations.  This is also covered in the
+[supplemental guidelines for handling asynchronous instrument
+views](#asynchronous-example-attribute-removal-in-a-view).
```
arminru, line 980:
```suggestion
views](supplementary-guidelines.md#asynchronous-example-attribute-removal-in-a-view).
```

specification/metrics/sdk.md
```diff
@@ -952,3 +952,29 @@ called concurrently.
 
 **MetricExporter** - `ForceFlush` and `Shutdown` are safe to be called
 concurrently.
+
+## Data model requirements
```
arminru, line 956: You'll need to update the ToC to make the check pass (`make markdown-toc`).

jmacd:
From zoom chat during the 1/25/22 Spec SIG:

Jack Berg to Everyone (8:37 AM)
> Couldn’t find a good moment to interject, but the duplicate callbacks change actually resolves a recurring outstanding question with the java SDK. Java allows you to obtain the same instrument from the same meter multiple times, provided that the instrument name, type, unit, are compatible. This tracks and makes sense for synchronous instruments, but doesn’t work for async unless multiple callbacks are allowed.

Daniel Dyla (Dynatrace) to Everyone (8:38 AM)
> Ditto JavaScript ^

specification/metrics/api.md
```diff
@@ -180,14 +180,33 @@ will have the following information:
...
+  It is unspecified whether or under which conditions the same or different
+  instrument instance will be returned as a result of duplicate registration.
+* If the registration is semantically not identical, the
+  implementation MUST return a registration error to the user when the
+  instrument is constructed.
+
+When determining whether a duplicate registration is valid,
+language-level features such the distinction between integer and
+floating point numbers SHOULD be considered.  Equivalently,
+implementations SHOULD NOT support an application in emitting both
```
reyang, line 201: Perhaps "MUST NOT"?

specification/metrics/api.md
```diff
@@ -239,6 +258,11 @@ instrument. It MUST be treated as an opaque string from the API and SDK.
 * It MUST support at least 1023 characters. [OpenTelemetry
   API](../overview.md#api) authors MAY decide if they want to support more.
 
+Note the `description` property of an instrument is explicitly
+disregarded when considering duplicate registration, because it is not
+semantic in nature.  Implementations SHOULD use any of the provided
```
reyang, line 263:
```suggestion
semantic in nature.  Implementations MAY use any of the provided
```

specification/metrics/api.md
```diff
@@ -180,14 +180,33 @@ will have the following information:
...
+identified by instrument name.  Duplicate registration of
+identically-named instruments within a Meter is treated as follows:
+
+* If the registration is semantically identical, meaning to have the
+  same `kind` and `unit` as a previously registered instrument of the
+  same `name`, the implementation MUST return a valid instrument.
+  It is unspecified whether or under which conditions the same or different
+  instrument instance will be returned as a result of duplicate registration.
+* If the registration is semantically not identical, the
+  implementation MUST return a registration error to the user when the
```
pichlermarc, line 195: I am wondering if instrument registration is considered part of the initialization. :thinking: 
I am worried that this might be a bit unwieldy to implement with the requirement not to throw after initialization in [this spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md#basic-error-handling-principles).

> The API or SDK may fail fast and cause the application to fail on initialization, e.g. because of a bad user config or environment, but MUST NOT cause the application to fail later at run time, e.g. due to dynamic config settings received from the Collector.

specification/metrics/api.md
```diff
@@ -180,14 +180,33 @@ will have the following information:
...
+language-level features such the distinction between integer and
+floating point numbers SHOULD be considered.  Equivalently,
+implementations SHOULD NOT support an application in emitting both
+integer and floating point numbers for the same metric inside the same
+instrumentation library.
+
+<a name="instrument-namespace"></a>
+
+Different Meters MUST be treated as separate namespaces. The names of the
+Instruments under one Meter SHOULD NOT interfere with Instruments under
```
MrAlias, line 208: This recommendation would be better declared as a requirement. There does not seem to be a situation where there is a valid reason to ignore this recommendation and doing so would present compatibility issues. If one OTel implementation allows this interference a user will not be able to implement an application in that language the a same as another because they cannot use the same instruments.

```suggestion
Instruments under one Meter MUST NOT interfere with Instruments under
```

specification/metrics/api.md
```diff
@@ -264,6 +288,16 @@ Please note that the term _synchronous_ and _asynchronous_ have nothing to do
...
 pattern](https://en.wikipedia.org/wiki/Asynchronous_method_invocation).
 
+When asynchronous instruments are used with duplicate registrations,
+meaning to have more than one callback provided, the order of callback
+execution is unspecified.  Callers SHOULD avoid making duplicate
+observations from asynchronous instrument callbacks.  However, if a
+user accidentally makes duplicate observations, meaning to provide
+more than one value for a given asynchronous instrument and attribute
+set, the caller SHOULD expect the last-observed value to be the one
+recorded.  This is sometimes referred to as "last-value wins" for
```
MrAlias, line 298: Make the normative statement be a positive one about OTel implementation behavior, not user expectations.

```suggestion
set, the last-observed value observed SHOULD be the one
recorded.  This is sometimes referred to as "last-value wins" for
```

Also, is there a valid reason to not use the last value observed? Should this be a requirement and not a recommendation? Not using the last value observed would lead to inconsistent behavior across OTel implementations. Meaning having this as a requirement would make sense.

specification/metrics/sdk.md
```diff
@@ -952,3 +952,29 @@ called concurrently.
...
+
+The implementation is required to respect the OpenTelemetry Metrics
```
MrAlias, line 958:
```suggestion
The implementation is REQUIRED to respect the OpenTelemetry Metrics
```

specification/metrics/api.md
ocelotl, line 195: > ... the implementation MUST return a registration error...

Is the use of "return an error" language-specific? I know that there are languages whose error handling consists of returning errors, but others raise exceptions.

Is the intention of this part of the specification to indicate something must me returned or that the error can be handled in a way that better suits every implementation?

ocelotl, line 195: What is a registration error?

specification/metrics/api.md
```diff
@@ -180,14 +180,33 @@ will have the following information:
...
+  same `kind` and `unit` as a previously registered instrument of the
+  same `name`, the implementation MUST return a valid instrument.
+  It is unspecified whether or under which conditions the same or different
+  instrument instance will be returned as a result of duplicate registration.
+* If the registration is semantically not identical, the
+  implementation MUST return a registration error to the user when the
+  instrument is constructed.
+
+When determining whether a duplicate registration is valid,
+language-level features such the distinction between integer and
```
ocelotl, line 199:
```suggestion
language-level features such as the distinction between integer and
```

specification/metrics/api.md
ocelotl: What does this mean?

> the identity of the meter is tied to the meter name

If 2 or more different meters may have the same name then how can the meter name be used to identify a meter?

specification/metrics/api.md
ocelotl: Is the "official" way of registering multiple callbacks going to be multiple registration of asynchronous instruments? If multiple callbacks are going to be supported and the user wants them, shouldn't the user be allowed to simply to add more than one callback when creating an asynchronous instrument?

ocelotl (commented):
Left some questions :v:

specification/metrics/api.md
jack-berg: > If 2 or more different meters may have the same name then how can the meter name be used to identify a meter?

I should be able to obtain a meter by the same name multiple times. And from the meter (or meters - it shouldn't matter whether the same exact instance is returned) I should be able to obtain the same instrument multiple times.

Using an example, the following should work without any errors, and with all calls to `add(..)` recording against the same metric stream:
```
meterProvider.get("my-meter").counterBuilder("my-counter").build().add(10);
meterProvider.get("my-meter").counterBuilder("my-counter").build().add(10);

Meter myMeter =  meterProvider.get("my-meter");
myMeter.counterBuilder("my-counter").build().add(10);
myMeter.counterBuilder("my-counter").build().add(10);
```

specification/metrics/api.md
reyang: Instrument registration is not part of initialization because libraries can get loaded after the initialization is done, and the library being loaded could register more instruments.

specification/metrics/api.md
```diff
@@ -180,14 +180,33 @@ will have the following information:
...
-  another Meter.
+<a name="duplicate-instrument-handling"></a>
+
+Instruments are associated with the Meter during creation and are
+identified by instrument name.  Duplicate registration of
+identically-named instruments within a Meter is treated as follows:
+
+* If the registration is semantically identical, meaning to have the
+  same `kind` and `unit` as a previously registered instrument of the
+  same `name`, the implementation MUST return a valid instrument.
```
reyang, line 191: "the implementation MUST return a valid instrument" will this be a breaking change given the API spec is already stable?

specification/metrics/api.md
aabmass: @ocelotl I agree this would be awkward. I think it is not meant to be the official way, but is just how multiple registrations should be handled as a specific case. See #2281, specifically "The API MUST provide a way to associate instruments with `callback` functions responsible for reporting the associated".

specification/metrics/api.md
```diff
@@ -180,14 +180,33 @@ will have the following information:
...
+  It is unspecified whether or under which conditions the same or different
+  instrument instance will be returned as a result of duplicate registration.
+* If the registration is semantically not identical, the
```
anuraaga, line 194: I am worried that being too strict here will cause more harm than good for users in the end.

https://github.com/open-telemetry/opentelemetry-specification/issues/2226#issuecomment-1025412191

anuraaga: Resolving as since this is just an aspect of the OTel data model itself there isn't much better we can do here

jmacd:
All, I've read through the feedback above and will work to clarify. Note that in #2226 there is now a developing discussion about how to handle duplicate-conflicting instrument registration. I'm willing to adjust the SDK specification to "tolerate" duplicate-conflicting instrument registration by treating instrument kind as part of metric identify for the purposes of the SDK; the specification would be to emit a warning and pass-through streams of both type. Please continue this discussion in #2226 until the objection raised by @anuraaga is settled. Thank you.

jmacd:
Reviewers, I am closing this as noted https://github.com/open-telemetry/opentelemetry-specification/issues/2226#issuecomment-1028243274

The feedback has been well taken and will be addressed. I will RE-OPEN this PR when the new draft is ready.

specification/metrics/api.md
jmacd: Take this to #2317 
I believe we can change a condition formerly specified as error into correct behavior. To me, the error state was effectively unspecified.

specification/metrics/api.md
jmacd: Fixed in #2317

specification/metrics/api.md
jmacd: Eliminated in #2317 😀

specification/metrics/api.md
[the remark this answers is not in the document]
jmacd: Removed in #2317

specification/metrics/api.md
jmacd: See #2317. There's still an error, it's just a question of where it gets surfaced.

specification/metrics/api.md
jmacd: On this note please see #2318. I removed this paragraph in a future iteration, simply stating at this point in the document that the behavior is unspecified. I view this being unspecified is a problem, but I don't want to fix it here.

specification/metrics/api.md
jmacd: Fixed in #2317

specification/metrics/api.md
jmacd: Following the discussion in #2226, I've updated this part of the proposal in a new PR please see #2317.

specification/metrics/api.md
jmacd: I've changed this detail in #2317. I think it's fine to be strict about description, as long as we're tolerating duplicate identities.

specification/metrics/api.md
jmacd: See #2317, major changes.

PR #2271, opened 2022-01-18
Drop log Name field

### LINKED ISSUE #2074: Is Name field necessary in Log Data Model

jack-berg:
Fixes #2074.

## Changes

Drop the Log `name` field.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,8 @@ release.
   ([#2184](https://github.com/open-telemetry/opentelemetry-specification/pull/2184))
 - Change mapping for log_name of Google Cloud Logging.
   ([#2092](https://github.com/open-telemetry/opentelemetry-specification/pull/2092))
+- Drop Log name.
+  field ([#2271](https://github.com/open-telemetry/opentelemetry-specification/pull/2271))
 
 ### Resource
 
--- a/specification/logs/data-model.md
+++ b/specification/logs/data-model.md
@@ -23,7 +23,6 @@
       - [Error Semantics](#error-semantics)
       - [Displaying Severity](#displaying-severity)
       - [Comparing Severity](#comparing-severity)
-    - [Field: `Name`](#field-name)
     - [Field: `Body`](#field-body)
     - [Field: `Resource`](#field-resource)
     - [Field: `Attributes`](#field-attributes)
@@ -184,7 +183,6 @@ SpanId         |Request span id.
 TraceFlags     |W3C trace flag.
 SeverityText   |The severity text (also known as log level).
 SeverityNumber |Numerical value of the severity.
-Name           |Short low cardinality event type.
 Body           |The body of the log record.
 Resource       |Describes the source of the log.
 Attributes     |Additional information about the event.
@@ -402,15 +400,6 @@ comparisons `SeverityNumber` field should be used. `SeverityNumber` can be
 compared to another `SeverityNumber` or to numbers in the 1..24 range (or to the
 corresponding short names).
 
-### Field: `Name`
-
-Type: string.
-
-Description: Short low cardinality event type that does not contain varying
-parts. `Name` describes what happened (e.g. "ProcessStarted"). Recommended to be
-no longer than 50 characters. Typically used for filtering and grouping purposes
-in backends. This field is optional.
-
 ### Field: `Body`
 
 Type: any.
@@ -585,7 +574,7 @@ this data model.
     <td>MSGID</td>
     <td>string</td>
     <td>Defines the type of the event. Part of event source identity. Example: "TCPIN"</td>
-    <td>Name</td>
+    <td>Attributes["syslog.msgid"]</td>
   </tr>
   <tr>
     <td>STRUCTURED-DATA</td>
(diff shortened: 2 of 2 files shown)


tigrannajaryan (approved):
The mappings that previously used the Name field may require more discussion, but because in this document the mappings are _Examples_ I am fine with merging this PR as is and if necessary future PRs can make further changes to the example mappings.

After this is merged we need a corresponding change in the proto repository and it will be followed by a change in the Collector.

@open-telemetry/collector-approvers @open-telemetry/collector-contrib-approvers please review and comment if you believe your receiver/exporter absolutely needs this field to stay for some reason and cannot substitute it by an attribute with a format-specific semantic convention.

djaglowski (approved):

anuraaga (approved):

dmitryax (approved):

tigrannajaryan:
@jkwatson @pmm-sumo please review.

jkwatson (approved):

pmm-sumo (approved):

tigrannajaryan:
Let's keep this open for a few more days, don't merge yet. I want to make sure more people are aware of the change.

bogdandrutu:
Not sure about the impact, but I think the example of using "Attributes" to model different "schemas" for the logs is not very robust. I am ok removing it, but I would feel strong against using attributes to define "schema"

tigrannajaryan:
> Not sure about the impact, but I think the example of using "Attributes" to model different "schemas" for the logs is not very robust. I am ok removing it, but I would feel bad about using attributes to define "schema"

@bogdandrutu which example are you referring to? I am also confused by the reference to "schemas". What do you mean by this?

tigrannajaryan:
@open-telemetry/specs-logs-approvers please review.

bogdandrutu:
@tigrannajaryan was referring to the issue.. as I said I think this is fine in general, but we should carefully think before making the standard of encoding different types of events/entries as logs and identify using attributes

tigrannajaryan:
> @tigrannajaryan was referring to the issue.. as I said I think this is fine in general, but we should carefully think before making the standard of encoding different types of events/entries as logs and identify using attributes

Got it. I have been thinking about this for a while now and it seems to fit nicely in how we understand the attributes are used how presence of attributes can indicate a "type" of the recorded information. For example let's say I want to record information about events happening in the browser. Instead of using the `Name` field I could for example have a semantic convention for an attribute named `browser.event_type` or `browser.dom.event_type` the presence of which indicates that it is a browser event and the value is a string that tells what exactly the event is.

Or for example Otel Collector may decide to emit events about its state, for example `otel.collector.state_change` with a value of "started" can indicate that the Collector is started and in that same log record also record a bunch other attributes that describe the Collector.

Essentially every domain (such as browser) needs to come up with a reasonable semantic convention for the event they want to describe.

cijothomas (approved):

zenmoto (approved):

himanshpal:
The log name is really helpful since its always going to be low cardinality field & this improves the Search by a large factor compared to doing LIKE on Body for systems which store logging data in columnar format (parquet) & query via athena/presto.

Removing this field moves the responsibility back to lib consumer to make sure all attributes minimum have something similar in all lang sdk's.

tigrannajaryan:
@himanshpal the problem with the existing Name field is that the community has not been able to come up with a clear definition of what it is supposed to reflect. The semantics are not clear and people have different opinions about what it is supposed to contain.

For your particular use case if you know what you want to record as the "name" you can do that as an attribute.

If in the future we manage to form a consensus around what the "name" means we do not preclude the possibility of adding the field back.

kumoroku (approved):

jmacd (approved):

jack-berg:
> Removing this field moves the responsibility back to lib consumer to make sure all attributes minimum have something similar in all lang sdk's.

@himanshpal one of the problems with `name` is that it lacks an unambiguous mapping for all known log and event formats, as argued [here](https://github.com/open-telemetry/opentelemetry-specification/issues/2074#issuecomment-996867783). So while its nice to have a low cardinality identifier, it's not particularly useful or guaranteed to be low cardinality unless the mapping semantics are well established. 

Instead, each log format can establish its own [mapping conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#appendix-a-example-mappings). Folks doing analysis / search can use any type / classifiers available for format of logs they're analyzing.

tigrannajaryan:
We have enough approvals, 2 business days since the last comment. Merging.

Issue #2203, opened 2021-12-06
Is it possible to rename "instrumentation library" to "instrumentation scope" everywhere?


tigrannajaryan:
This initially came from a [discussion](https://github.com/open-telemetry/opentelemetry-specification/pull/1236#issuecomment-970464780) about introducing Logger Name. 

While discussing the issue during the spec SIG meeting we came to a conclusion that the intent for "instrumentation library" was actually to have an "instrumentation scope".

The "instrumentation library" is currently used throughout the spec API and SDK. We want to explore if it is possible to rename "instrumentation library" to "instrumentation scope" without breaking existing stability guarantees.

Note: this should be only done if it is not a breaking change. We must continue honouring our stability guarantees.

Places to check:
- [x] https://github.com/open-telemetry/opentelemetry-java/issues/3961 (*This is a breaking change for Java*).
- [x] https://github.com/open-telemetry/opentelemetry-go/issues/2431 (*This is a breaking change for Go*).
- [x] https://github.com/open-telemetry/opentelemetry-python/issues/2310
- [x] https://github.com/open-telemetry/opentelemetry-dotnet/issues/2725
- [x] https://github.com/open-telemetry/opentelemetry-php/issues/504
- [x] https://github.com/open-telemetry/opentelemetry-cpp/issues/1117
- [x] https://github.com/open-telemetry/opentelemetry-ruby/issues/1052
- [x] https://github.com/open-telemetry/opentelemetry-js/issues/2656 (*This is a breaking change for JS*).
- [x] https://github.com/open-telemetry/opentelemetry-erlang/issues/326
- [x] https://github.com/open-telemetry/opentelemetry-swift/issues/283
- [x] https://github.com/open-telemetry/opentelemetry-rust/issues/675
- [x] https://github.com/open-telemetry/opentelemetry-proto/issues/345
- [x] https://github.com/open-telemetry/opentelemetry-collector/issues/4516

In this spec:
- https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/non-otlp.md#instrumentationlibrary
- https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/overview.md#otlp-support
- https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#tracer-creation
- https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#additional-span-interfaces
[...]

PR #2276, opened 2022-01-19
Clarify that Trace/Meter are associated with Instrumentation Scope

tigrannajaryan:
This is a slightly different take on https://github.com/open-telemetry/opentelemetry-specification/issues/2203

Instead of renaming instrumentation library to instrumentation scope everywhere
this PR suggests targetted editing of wording of the Trace/Meter obtaining API
to allow not just instrumentation library but other instrumentation scopes to be
used as a parameter.

This change does not force renaming of API parameters and is not a breaking change.
We consider it a clarification of usage to match the intent that we had for the "name"
field.

If this PR is accepted there will be a follow up PR that will suggest to rename
InstrumentationLibrary* messages in OTLP proto to InstrumentationScope* message.
Such a change will not be a breaking change for the OTLP wire format and is acceptable.

If this PR is accepted we will also close https://github.com/open-telemetry/opentelemetry-specification/pull/1236
since it will be no longer necessary. The logger name will be recorded as the
instrumentation scope, done in a separate PR that clarifies the behavior in https://github.com/open-telemetry/oteps/blob/main/text/logs/0150-logging-library-sdk.md

Resolves https://github.com/open-telemetry/opentelemetry-specification/issues/2307

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,7 +48,8 @@ release.
 - Change description and default value of OTEL_EXPORTER_JAEGER_ENDPOINT env var
   to point to the correct HTTP port and correct description of OTEL_TRACES_EXPORTER
   ([#2333](https://github.com/open-telemetry/opentelemetry-specification/pull/2333)).
-- Clarify that Tracer is associated with Instrumentation Scope
+- Introduce the concept of Instrumentation Scope to replace/extend Instrumentation
+  Library. The Tracer is now associated with Instrumentation Scope
   ([#2276](https://github.com/open-telemetry/opentelemetry-specification/pull/2276)).
 
 ### Metrics
@@ -80,7 +81,8 @@ release.
   ([#2282](https://github.com/open-telemetry/opentelemetry-specification/pull/2282))
 - Clarified wildcard and predicate support in metrics SDK View API.
   ([#2325](https://github.com/open-telemetry/opentelemetry-specification/pull/2325))
-- Clarify that Meter is associated with Instrumentation Scope
+- Introduce the concept of Instrumentation Scope to replace/extend Instrumentation
+  Library. The Meter is now associated with Instrumentation Scope
   ([#2276](https://github.com/open-telemetry/opentelemetry-specification/pull/2276)).
 
 ### Logs
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,15 @@ release.
 
 ### Traces
 
+- Introduce the concept of Instrumentation Scope to replace/extend Instrumentation
+  Library. The Tracer is now associated with Instrumentation Scope
+  ([#2276](https://github.com/open-telemetry/opentelemetry-specification/pull/2276)).
+
 ### Metrics
+
+- Introduce the concept of Instrumentation Scope to replace/extend Instrumentation
+  Library. The Meter is now associated with Instrumentation Scope
+  ([#2276](https://github.com/open-telemetry/opentelemetry-specification/pull/2276)).
 
 ### Logs
 
@@ -48,9 +56,6 @@ release.
 - Change description and default value of OTEL_EXPORTER_JAEGER_ENDPOINT env var
   to point to the correct HTTP port and correct description of OTEL_TRACES_EXPORTER
   ([#2333](https://github.com/open-telemetry/opentelemetry-specification/pull/2333)).
-- Introduce the concept of Instrumentation Scope to replace/extend Instrumentation
-  Library. The Tracer is now associated with Instrumentation Scope
-  ([#2276](https://github.com/open-telemetry/opentelemetry-specification/pull/2276)).
 
 ### Metrics
 
@@ -81,9 +86,6 @@ release.
   ([#2282](https://github.com/open-telemetry/opentelemetry-specification/pull/2282))
 - Clarified wildcard and predicate support in metrics SDK View API.
   ([#2325](https://github.com/open-telemetry/opentelemetry-specification/pull/2325))
-- Introduce the concept of Instrumentation Scope to replace/extend Instrumentation
-  Library. The Meter is now associated with Instrumentation Scope
-  ([#2276](https://github.com/open-telemetry/opentelemetry-specification/pull/2276)).
 
 ### Logs
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,8 @@ release.
 - Change description and default value of OTEL_EXPORTER_JAEGER_ENDPOINT env var
   to point to the correct HTTP port and correct description of OTEL_TRACES_EXPORTER
   ([#2333](https://github.com/open-telemetry/opentelemetry-specification/pull/2333)).
+- Clarify that Tracer is associated with Instrumentation Scope
+  ([#2276](https://github.com/open-telemetry/opentelemetry-specification/pull/2276)).
 
 ### Metrics
 
@@ -78,6 +80,8 @@ release.
   ([#2282](https://github.com/open-telemetry/opentelemetry-specification/pull/2282))
 - Clarified wildcard and predicate support in metrics SDK View API.
   ([#2325](https://github.com/open-telemetry/opentelemetry-specification/pull/2325))
+- Clarify that Meter is associated with Instrumentation Scope
+  ([#2276](https://github.com/open-telemetry/opentelemetry-specification/pull/2276)).
 
 ### Logs
 
--- a/spec-compliance-matrix.md
+++ b/spec-compliance-matrix.md
@@ -158,6 +158,7 @@ Disclaimer: this list of features is still a work in progress, please refer to t
 | `MeterProvider` allows a `Resource` to be specified.                                                                                                                         |          | +  |  +   |    |    +   |      |        |     |      |     |   +  |       |
 | A specified `Resource` can be associated with all the produced metrics from any `Meter` from the `MeterProvider`.                                                            |          | +  |  +   |    |    +   |      |        |     |      |     |   +  |       |
 | The supplied `name`, `version` and `schema_url` arguments passed to the `MeterProvider` are used to create an `InstrumentationLibrary` instance stored in the `Meter`.       |          | +  |  +   |    |    +   |      |        |     |      |     |   -  |       |
+| The supplied `name`, `version` and `schema_url` arguments passed to the `MeterProvider` are used to create an `InstrumentationScope` instance stored in the `Meter`.         |          |    |      |    |        |      |        |     |      |     |      |       |
 | Configuration is managed solely by the `MeterProvider`.                                                                                                                      |          | +  |  +   |    |    +   |      |        |     |      |     |   +  |       |
 | The `MeterProvider` provides methods to update the configuration                                                                                                             | X        | -  |  -   |    |    +   |      |        |     |      |     |   +  |       |
 | The updated configuration applies to all already returned `Meter`s.                                                                                                          | if above | -  |  -   |    |    -   |      |        |     |      |     |   +  |       |
@@ -292,6 +293,7 @@ Note: Support for environment variables is optional.
 | Service name mapping                                                           |          | +  | + | +  | +           | +    | +      | +   | +    | +   | +    | +     |
 | SpanKind mapping                                                               |          | +  | + | +  | +           | +    | +      | +   | +    | +   | +    | +     |
 | InstrumentationLibrary mapping                                                 |          | +  | + | -  | +           | +    | -      | +   | +    | +   | +    | +     |
+| InstrumentationScope mapping                                                   |          |    |   |    |             |      |        |     |      |     |      |       |
 | Boolean attributes                                                             |          | +  | + | +  | +           | +    | +      | +   | +    | +   | +    | +     |
 | Array attributes                                                               |          | +  | + | +  | +           | +    | +      | +   | +    | +   | +    | +     |
 | Status mapping                                                                 |          | +  | + | +  | +           | +    | +      | +   | +    | +   | +    | +     |
@@ -305,6 +307,7 @@ Note: Support for environment variables is optional.
 | Service name mapping                                                           |          | +  | + |    | +           | +    | -      |     |      | +   | +    | +     |
 | Resource to Process mapping                                                    |          | +  | + |    | +           | +    | -      |     | +    | -   | +    | -     |
 | InstrumentationLibrary mapping                                                 |          | +  | + |    | +           | +    | -      |     | +    | -   | +    | -     |
+| InstrumentationScope mapping                                                   |          |    |   |    |             |      |        |     |      |     |      |       |
 | Status mapping                                                                 |          | +  | + |    | +           | +    | -      |     | +    | +   | +    | +     |
 | Error Status mapping                                                           |          | +  | + |    | +           | +    | -      |     | +    | +   | +    | -     |
 | Events converted to Logs                                                       |          | +  | + |    | +           | +    | -      |     | +    | -   | +    | +     |
--- a/spec-compliance-matrix.md
+++ b/spec-compliance-matrix.md
@@ -20,6 +20,7 @@ formats is required. Implementing more than one format is optional.
 | Create TracerProvider                                                                            |          | +   | +    | +   | +      | +    | +      | +   | +    | +   | +    | +     |
 | Get a Tracer                                                                                     |          | +   | +    | +   | +      | +    | +      | +   | +    | +   | +    | +     |
 | Get a Tracer with schema_url                                                                     |          | +   | +    |     |        |      |        |     |      | +   |      |       |
+| Associate Tracer with InstrumentationScope                                                       |          |     |      |     |        |      |        |     |      |     |      |       |
 | Safe for concurrent calls                                                                        |          | +   | +    | +   | +      | +    | +      | +   | +    | +   | +    | +     |
 | Shutdown (SDK only required)                                                                     |          | +   | +    | +   | +      | +    | +      | +   | +    | +   | +    | +     |
 | ForceFlush (SDK only required)                                                                   |          | +   | +    | -   | +      | +    | +      | +   | +    | +   | +    | +     |
@@ -80,6 +81,7 @@ formats is required. Implementing more than one format is optional.
 | [SpanLimits](specification/trace/sdk.md#span-limits)                                             | X        | +   | +    |     | +      | +    | +      | +   |      | -   |      | +     |
 | [Built-in `SpanProcessor`s implement `ForceFlush` spec](specification/trace/sdk.md#forceflush-1) |          |     | +    |     | +      | +    | +      | +   | +    | +   |      |       |
 | [Attribute Limits](specification/common/common.md#attribute-limits)                              | X        |     | +    |     |        |      | +      | +   |      |     |      |       |
+| Fetch InstrumentationScope from ReadableSpan                                                     |          |     |      |     |        |      |        |     |      |     |      |       |
 
 ## Baggage
 
@@ -102,6 +104,7 @@ Disclaimer: this list of features is still a work in progress, please refer to t
 | `get_meter` accepts name, `version` and `schema_url`.                                                                                                                        |          | +  |  +   |    |    +   |      |        |     |      |     |   -  |       |
 | When an invalid `name` is specified a working `Meter` implementation is returned as a fallback.                                                                              |          | +  |  +   |    |    -   |      |        |     |      |     |   -  |       |
 | The fallback `Meter` `name` property keeps its original invalid value.                                                                                                       | X        | -  |  -   |    |    -   |      |        |     |      |     |   -  |       |
+| Associate `Meter` with `InstrumentationScope`.                                                                                                                               |          |    |      |    |        |      |        |     |      |     |      |       |
 | The meter provides functions to create a new `Counter`.                                                                                                                      |          | +  |  +   |    |    +   |      |        |     |      |     |   +  |       |
 | The meter provides functions to create a new `AsynchronousCounter`.                                                                                                          |          | +  |  +   |    |    +   |      |        |     |      |     |   +  |       |
 | The meter provides functions to create a new `Histogram`.                                                                                                                    |          | +  |  +   |    |    +   |      |        |     |      |     |   +  |       |
--- a/specification/glossary.md
+++ b/specification/glossary.md
@@ -155,7 +155,7 @@ Synonyms: *Instrumenting Library*.
 ### Instrumentation Scope
 
 A logical unit of the application code with which the emitted telemetry can be
-associated with. It is typically the developer's choice to decide what denotes a
+associated. It is typically the developer's choice to decide what denotes a
 reasonable instrumentation scope. The most common approach is to use the
 [instrumentation library](#instrumentation-library) as the scope, however other
 scopes are also common, e.g. a module, a package, or a class can be chosen as
--- a/specification/glossary.md
+++ b/specification/glossary.md
@@ -163,7 +163,10 @@ the instrumentation scope.
 
 If the unit of code has a version then the instrumentation scope is defined by
 the (name,version) pair otherwise the version is omitted and only the name is
-used.
+used. The name or (name,version) pair uniquely identify the logical unit of the
+code that emits the telemetry. A typical approach to ensure uniqueness is to use
+fully qualified name of the emitting code (e.g. fully qualified library name or
+fully qualified class name).
 
 The instrumentation scope is used to obtain a
 [Tracer or Meter](#tracer-name--meter-name).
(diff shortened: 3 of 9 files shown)


bogdandrutu (approved):

yurishkuro (dismissed):

specification/glossary.md
```diff
@@ -151,11 +152,31 @@ Example: `io.opentelemetry.contrib.mongodb`.
 
 Synonyms: *Instrumenting Library*.
 
+### Instrumentation Scope
+
+A logical unit of the application code with which the emitted telemetry can be
+associated with. It is typically the developer's choice to decide what denotes a
```
GraemeF, line 158:
```suggestion
associated. It is typically the developer's choice to decide what denotes a
```

tigrannajaryan: Fixed, thanks.

carlosalberto (approved):

carlosalberto:
Overall LGTM - I'd like to get the opinion of maintainers though.

cc @tedsuo

MrAlias (approved):

jmacd (approved):

tigrannajaryan:
Let's keep this open for at least one week. I want to make sure lots of folks see this.

tigrannajaryan:
We should also mention this in the Maintainers and Spec SIG meetings next week.

bogdandrutu:
@tigrannajaryan I created an entry for the next week maintainer agenda meeting, and add this as a discussion point. I will not be there.

jack-berg (approved):

tigrannajaryan:
@open-telemetry/specs-approvers I won't be able to attend the Maintainers meeting today. If you are in the meeting please ask maintainers to have a look / be aware of this change.

specification/metrics/api.md
```diff
@@ -108,8 +108,9 @@ The `MeterProvider` MUST provide the following functions:
 This API MUST accept the following parameters:
 
 * `name` (required): This name must identify the [instrumentation
-  library](../overview.md#instrumentation-libraries) (e.g.
-  `io.opentelemetry.contrib.mongodb`). If an application or library has built-in
+  scope](../glossary.md#instrumentation-scope) (e.g.
+  `io.opentelemetry.contrib.mongodb`), such as the instrumentation library,
```
arminru, line 112: Please keep the glossary reference for instrumentation library so readers can look it up.

specification/trace/api.md
```diff
@@ -106,22 +106,24 @@ The `TracerProvider` MUST provide the following functions:
 
 This API MUST accept the following parameters:
 
-- `name` (required): This name must identify the [instrumentation library](../overview.md#instrumentation-libraries)
-  (e.g. `io.opentelemetry.contrib.mongodb`).
-  If an application or library has built-in OpenTelemetry instrumentation, both
+- `name` (required): This name must identify the [instrumentation
+  scope](../glossary.md#instrumentation-scope) (e.g.
```
arminru, line 110: Please keep the glossary reference for instrumentation library so readers can look it up.

djaglowski (approved):

specification/metrics/api.md
```diff
@@ -108,8 +108,9 @@ The `MeterProvider` MUST provide the following functions:
...
+  package, module or class name. If an application or library has built-in
```
arminru, line 113: If one would like to record both the instrumentation library (e.g. `io.opentelemetry.contrib.mongodb`) and the "generic" instrumentation scope (e.g. `MyAppThatsUsingMongoDb` or `com.example.mongodbclient` - both would be applicable now with this definition), they couldn't do that but would need to decide for either, right?

How can telemetry consumers (people looking at the data other than the instrumentation developer, or automated analyzers in backends) distinguish which choice was made and how the value should be interpreted?

tigrannajaryan: >If one would like to record both the instrumentation library (e.g. io.opentelemetry.contrib.mongodb) and the "generic" instrumentation scope (e.g. MyAppThatsUsingMongoDb or com.example.mongodbclient - both would be applicable now with this definition), they couldn't do that but would need to decide for either, right?

Correct, it will not be possible.

This was the exact debate that we had in [LoggerName](https://github.com/open-telemetry/opentelemetry-specification/pull/1236) discussion where I argued that adding LoggerName as an _additional_ field is important because we may want to capture the scope such as class name in _addition_ to the library name. I don't think I was able to convince people, so I am now back to this opposite proposal.

dyladan, line 113: Expanding the definition of meter name seems quite dangerous to me. Right now it means something very specific (library name), but if we add this then it could be quite confusing to the user. According to this wording, I could easily just put the class name, losing the information of which library generate the metric.

At the very least, the list should mention "fully qualified class name" or something similar to make it clear that the instrumentation library name is still important here.

A better solution would be to say that in the context of this API, "instrumentation scope" is _always_ the instrumentation library name.

dyladan: Additionally, the list doesn't make it clear if package, module, or class refer to the instrumenting library or the instrumented library. Again, I would prefer to keep the definition restricted here to the instrumentation library name.

This might be fixed by changing the wording to something like "library name, package, module, or class of the instrumentation library" or similar

specification/trace/api.md
dyladan: I have the same complaints here that I do with regards to changes to meter name

specification/metrics/api.md
tigrannajaryan: > Expanding the definition of meter name seems quite dangerous to me. Right now it means something very specific (library name), but if we add this then it could be quite confusing to the user. According to this wording, I could easily just put the class name, losing the information of which library generate the metric.

This is precisely what I would like us to decide on. Yes, that's exactly the intent of this PR, to make it possible to specify _arbitrary_ string as the instrumentation scope. It is not longer limited to instrumentation library name. I would like the community to either decide that this dilution of the concept is acceptable or reject it. If we as a community do not believe this is the right way I want to see people blocking this PR.

tigrannajaryan:
@open-telemetry/specs-metrics-approvers @open-telemetry/specs-trace-approvers please review.

specification/metrics/api.md
dyladan: Even if we don't add logger name separately, it is important not to lose the information of the instrumentation library name. If it is going to be expanded, it needs to be expanded in a way that retains this information (some sort of fully qualified, additional field, etc).

dyladan (changes requested):
Expanding the definition of a field that was already declared stable such that it is now possible to lose the originally intended information contained in the field  (by using a class name instead of the actual library name) breaks the field.

- This is breaking for receivers who depend on the instrumentation library name being contained in this field, with no way to recover this information
- Logger name will not always refer to the instrumentation library as this field was intended. Actually, it is extremely _unlikely_ to refer to the instrumentation library and would rather refer to the thing being instrumented.

tigrannajaryan:
@dyladan to clarify, your position is that we should not make this change for Tracer/Meter name because it is a breaking change. You think that it is OK to assign a different meaning to the Logger name field, i.e. accept that it is not going to be consistent with Tracer/Meter name in its semantics.

dyladan:
> @dyladan to clarify, your position is that we should not make this change for Tracer/Meter name because it is a breaking change. You think that it is OK to assign a different meaning to the Logger name field, i.e. accept that it is not going to be consistent with Tracer/Meter name in its semantics.

Yes. I am against removing the accepted meaning from the tracer and meter name. I would be ok with changing the name of the field for tracer and meter name if the same meaning was retained. This would allow it to be used with a different meaning for logger name. For example, I would say we could call it instrumentation scope if we said for tracer and meter instrumentation scope is always instrumentation name.

WRT the motivation for this PR, I was not involved in the discussion around logger name, but it seems to me that it should be its own field. A single library can and often does create many loggers with many different names, and the names don't always mean the same thing even within the same library (sometimes they refer to the component being monitored, but sometimes they refer to the type of data being monitored, or even the destination the logs are meant to be delivered to). Trying to shoehorn it into an existing field feels like a mistake to me.

specification/metrics/api.md
Aneurysm9: I agree with Dan that this seems to be changing the scope of what might be an "instrumentation scope" to the point where it might be impossible to derive any consistent meaning from it.  The prior definition of "instrumentation library" was clear and well-defined and a change away from that at this point feels wrong.

Aneurysm9:
> WRT the motivation for this PR, I was not involved in the discussion around logger name, but it seems to me that it should be its own field. A single library can and often does create many loggers with many different names, and the names don't always mean the same thing even within the same library (sometimes they refer to the component being monitored, but sometimes they refer to the type of data being collected, or even the destination the logs are meant to be delivered to). Trying to shoehorn it into an existing field feels like a mistake to me.

I'm still not clear why this data needs to be defined as a top-level, independent field.  As noted here there can be a lot of different meanings when making the concept as general as "scope".  Can this be handled through attributes and semantic conventions?  In my experience with loggers identification of the scope typically happens through the addition of an attribute with a relevant name such as `component` or `scope`.

tigrannajaryan:
> it seems to me that it should be its own field. A single library can and often does create many loggers with many different names, and the names don't always mean the same thing even within the same library

This was my position as well. I was not able to convince everyone that this is the right approach, particularly @bogdandrutu was against it. Bogdan's reasoning was that instrumentation library name was intended to capture logger name as well. This PR is attempting to solidify that reasoning.

The reason I created this PR is that I want the Spec SIG to rule on one of these 3 approaces:
1. The instrumentation library name is actually meant to be the instrumentation scope, which can be an arbitrary string. In that case I do not need a separate logger name.
2. Or instrument library name is actually just the library name and not an arbitrary string that we can call a "scope". In that case I do need a separate logger name.
3. Logging is inherently different in this subject mater and cannot be consistent with Tracing/Metrics. We should not expect Tracer, Meter and Logger names to have the same semantics.

The current state is neither 1 nor 2 nor 3 and it is blocking the Logging SIG from making progress.

I need us all to choose and settle on one of these approaches (or perhaps there is another way, but I don't see it).

reyang (approved):

tigrannajaryan:
Discussed this in Spec SIG meeting and a new idea was brought up: Supply instrumentation library _and_ additional scope (with finer granularity) to obtainer Tracer/Meter/Logger, this would also work if a future Logs API would be added
Would be non-breaking in the API similar to schema_url.

I need some time to think about the feasibility of this idea. If it works this PR will be discarded.

yurishkuro:
given the concerns with breaking instrumentation library semantic I think an additional (not replacement) scope dimension is the right path

bogdandrutu:
Also, I hate that I did not push more in https://github.com/open-telemetry/oteps/pull/84 to call that "component", but lots of people were against that :(

specification/glossary.md
```diff
@@ -27,6 +27,7 @@ Some other fundamental terms are documented in the [overview document](overview.
   * [Exporter Library](#exporter-library)
   * [Instrumented Library](#instrumented-library)
   * [Instrumentation Library](#instrumentation-library)
+  * [Instrumentation Scope](#instrumentation-scope)
```
bogdandrutu, line 30: Add the "library" as one of the types for scope? or maybe replace it for the moment?

specification/glossary.md
```diff
@@ -151,11 +152,31 @@ Example: `io.opentelemetry.contrib.mongodb`.
...
+A logical unit of the application code with which the emitted telemetry can be
+associated. It is typically the developer's choice to decide what denotes a
+reasonable instrumentation scope. The most common approach is to use the
+[instrumentation library](#instrumentation-library) as the scope, however other
+scopes are also common, e.g. a module, a package, or a class can be chosen as
+the instrumentation scope.
+
+If the unit of code has a version then the instrumentation scope is defined by
+the (name,version) pair otherwise the version is omitted and only the name is
+used.
```
bogdandrutu, line 166: Based on today discussion, make sure we preserve the "uniqueness" suggestion/requirement.

bogdandrutu (approved):

bogdandrutu:
@dyladan can you take another look at this after today's discussion?

specification/glossary.md
```diff
@@ -151,11 +152,31 @@ Example: `io.opentelemetry.contrib.mongodb`.
 
 Synonyms: *Instrumenting Library*.
 
+### Instrumentation Scope
+
+A logical unit of the application code with which the emitted telemetry can be
+associated. It is typically the developer's choice to decide what denotes a
+reasonable instrumentation scope. The most common approach is to use the
+[instrumentation library](#instrumentation-library) as the scope, however other
+scopes are also common, e.g. a module, a package, or a class can be chosen as
```
dyladan, line 161: If a module/package/class is used it should still be globally unique or fully qualified.

specification/metrics/api.md
dyladan: I still feel the same way I did last week. I thought we had resolved at the spec meeting last week to add an additional scope field and retain library name which was fine with me. This week's spec discussion seemed to indicate that isn't the case anymore though?

dyladan (commented):
This proposal hasn't really changed since last week so I have the same concerns I did then.

specification/glossary.md
bogdandrutu: https://github.com/open-telemetry/opentelemetry-specification/pull/2276#discussion_r797018366

specification/metrics/api.md
bogdandrutu: Yes, we discussed this, that the idea of having 2 strings on the wire is fine, but asking users to provide 2 strings will be very confusing. I think we all agreed that the provided string SHOULD be unique as you pointed.

specification/metrics/api.md
tigrannajaryan: Done.

specification/trace/api.md
tigrannajaryan: Done.

specification/glossary.md
tigrannajaryan: Done.

specification/glossary.md
tigrannajaryan: We have appropriate references from scope to library in the text. I do not see a good way to restructure the TOC, let's keep as is.

specification/glossary.md
tigrannajaryan: Done.

tigrannajaryan:
All, I pushed a new [commit](https://github.com/open-telemetry/opentelemetry-specification/pull/2276/commits/5e7bbf35ec7e1f9d3b61fe21e0007b410ecc07d4) with some additional changes based on the discussions we had so far. What's new in the commit:
- The scope name is explicitly said to be unique (to match previous implicit expectation that library name is unique)
- Added `otel.scope.*` attributes for non-OTLP formats. Kept `otel.library.*` attributes for backwards compatibility reasons, but deprecated and will be removed according to our normal guarantees for "Stable" labels.
- Add InstrumentationScope to ReadableSpan. Kept InstrumentationLibrary as an alias, but deprecated it.

If this PR gets accepted the following additional PRs will be made:
- Change InstrumentationLibrary to InstrumentationScope or to simply Scope in OTLP proto repo. Not breaking wire format.
- Optional: add arbitrary additional `attributes` to InstrumentationScope, with corresponding semantic conventions to record any extra data about the scope we may want. Additive change.

tigrannajaryan:
@bogdandrutu @carlosalberto @jmacd @djaglowski @MrAlias @reyang @jack-berg please have another look, particularly at the new [commit](https://github.com/open-telemetry/opentelemetry-specification/pull/2276/commits/5e7bbf35ec7e1f9d3b61fe21e0007b410ecc07d4). I did not dismiss all approvals since I don't think the new commit radically changes the PR, but nevertheless it would be good to have another look.

specification/trace/api.md
```diff
@@ -106,11 +106,11 @@ The `TracerProvider` MUST provide the following functions:
 
 This API MUST accept the following parameters:
 
-- `name` (required): This name must identify the [instrumentation
-  scope](../glossary.md#instrumentation-scope) (e.g.
-  `io.opentelemetry.contrib.mongodb`), such as the instrumentation library,
-  package, module or class name. If an application or library has built-in
-  OpenTelemetry instrumentation, both
+- `name` (required): This name SHOULD uniquely identify the
+  [instrumentation scope](../glossary.md#instrumentation-scope) , such as the
```
jack-berg, line 110: #nit: extra space

jack-berg (approved):

jmacd (approved):

specification/trace/api.md
tigrannajaryan: Fixed.

mtwo:
FYI, @dyladan mentioned on the maintainer call that he just got back from vacation and will be able to review this tomorrow

dyladan (approved):
Looks good to me

arminru:
> Did we change anything in the `markdownlint` step? It fails on a file that I didn't change.

@tigrannajaryan A breaking change was introduced in markdownlint-cli, #2320 should fix this.

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ release.
 - Add support for probability sampling in the OpenTelemetry `tracestate` entry and
   add optional specification for consistent probability sampling.
   ([#2047](https://github.com/open-telemetry/opentelemetry-specification/pull/2047))
+- Clarify that Tracer is associated with Instrumentation Scope
```
arminru, line 20: This should reflect that an additional term was introduced and the former occurrences were changed to this new term, it's not just a clarification.
Especially the ReadableSpan representation and non-OTLP exporters need to be changed accordingly which might otherwise be overlooked.

specification/trace/sdk_exporters/non-otlp.md
```diff
@@ -21,15 +21,23 @@ over the generic rules defined in this document.
...
-| OpenTelemetry InstrumentationLibrary Field | non-OTLP Key |
-| ------------------- | --- |
-| `InstrumentationLibrary.name`|`otel.library.name`|
-| `InstrumentationLibrary.version`|`otel.library.version`|
+| OpenTelemetry InstrumentationScope Field | non-OTLP Key | Notes |
+| ------------------- | --- | --- |
+| `InstrumentationScope.name`|`otel.scope.name`|since 1.9.0|
+| `InstrumentationScope.version`|`otel.scope.version`|since 1.9.0|
+
+The following deprecated aliases are also reported with exact same values for
```
arminru, line 34:
```suggestion
The following deprecated aliases MUST also be reported with exact same values for
```

specification/glossary.md
```diff
@@ -151,11 +152,34 @@ Example: `io.opentelemetry.contrib.mongodb`.
...
+the instrumentation scope.
+
+If the unit of code has a version then the instrumentation scope is defined by
+the (name,version) pair otherwise the version is omitted and only the name is
+used. The name or (name,version) pair uniquely identify the logical unit of the
+code that emits the telemetry. A typical approach to ensure uniqueness is to use
```
arminru, line 167: The uniqueness is a normative requirement. Is it sufficient to mention this in the Glossary or should this go into the respective spec sections?

CHANGELOG.md
arminru: We should also add entries to the spec compliance matrix to track these changes across SDKs so that it will eventually be consistent.

specification/trace/sdk_exporters/non-otlp.md
tigrannajaryan: Fixed.

specification/trace/sdk.md
```diff
@@ -117,11 +117,14 @@ Thus, the SDK specification defines sets of possible requirements for
...
 
 * **Readable span**: A function receiving this as argument MUST be able to
-  access all information that was added to the span,
-  as listed [in the API spec](api.md#span-data-members).
-  In particular, it MUST also be able to access
-  the `InstrumentationLibrary` and `Resource` information (implicitly)
-  associated with the span.
+  access all information that was added to the span, as listed
+  [in the API spec](api.md#span-data-members). In particular, it MUST also be
+  able to access the `InstrumentationScope` [since 1.9.0] and `Resource`
```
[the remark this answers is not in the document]
tigrannajaryan: We already have similar references, e.g. here https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#get-a-tracer

They use the form `[since x.y.z]`. I think this is clear enough, there is no need for additional verbosity.

CHANGELOG.md
tigrannajaryan: Added entries to compliance matrix.

specification/glossary.md
tigrannajaryan: The requirement is also listed in respective API section.

specification/trace/sdk_exporters/non-otlp.md
```diff
@@ -21,15 +21,23 @@ over the generic rules defined in this document.
...
 pairs associated with the Span using the following mapping:
 
-| OpenTelemetry InstrumentationLibrary Field | non-OTLP Key |
-| ------------------- | --- |
-| `InstrumentationLibrary.name`|`otel.library.name`|
-| `InstrumentationLibrary.version`|`otel.library.version`|
+| OpenTelemetry InstrumentationScope Field | non-OTLP Key | Notes |
+| ------------------- | --- | --- |
+| `InstrumentationScope.name`|`otel.scope.name`|since 1.9.0|
+| `InstrumentationScope.version`|`otel.scope.version`|since 1.9.0|
```
arminru, line 32: Just noticed one more reference to the previous term here:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-naming.md#otel-namespace

tigrannajaryan: Fixed.

tigrannajaryan:

pmm-sumo (approved):

tigrannajaryan:
@carlosalberto if we are planning the 1.9.0 release in the next couple day I suggest we hold this PR until the release and merge after.

specification/trace/sdk_exporters/non-otlp.md
```diff
@@ -21,15 +21,23 @@ over the generic rules defined in this document.
...
+
+The following deprecated aliases MUST also be reported with exact same values for
+backward compatibility reasons:
+
+| non-OTLP Key | Alias for | Notes |
+| --- | --- | --- |
+|`otel.library.name`|`otel.scope.name`|deprecated since since 1.9.0|
+|`otel.library.version`|`otel.scope.version`|deprecated since since 1.9.0|
```
bogdandrutu, line 40: Add a note since when.

specification/trace/sdk_exporters/non-otlp.md
```diff
@@ -21,15 +21,23 @@ over the generic rules defined in this document.
...
+| ------------------- | --- | --- |
+| `InstrumentationScope.name`|`otel.scope.name`|since 1.9.0|
+| `InstrumentationScope.version`|`otel.scope.version`|since 1.9.0|
+
+The following deprecated aliases MUST also be reported with exact same values for
+backward compatibility reasons:
+
+| non-OTLP Key | Alias for | Notes |
+| --- | --- | --- |
+|`otel.library.name`|`otel.scope.name`|deprecated since since 1.9.0|
```
arminru, line 39: Duplicate "since" and as per your last comment you'll need to bump this to v1.10

specification/trace/sdk_exporters/non-otlp.md
tigrannajaryan: There is a note next to each attribute. Do you want something additional?

lzchen (approved):

spec-compliance-matrix.md
```diff
@@ -20,6 +20,7 @@ formats is required. Implementing more than one format is optional.
 | Create TracerProvider                                                                            |          | +   | +    | +   | +      | +    | +      | +   | +    | +   | +    | +     |
```
arminru, line 20: There are still a few items referencing InstrumentationLibrary in the Metrics and Exporter sections (out of reach for directly commenting on them).

tigrannajaryan:
@open-telemetry/specs-approvers @open-telemetry/specs-trace-approvers @open-telemetry/specs-metrics-approvers  This is the last chance to object to this PR. I will merge if I don't hear any objections.

jmacd (approved):

MovieStoreGuy (approved):

specification/trace/api.md
```diff
@@ -106,22 +106,24 @@ The `TracerProvider` MUST provide the following functions:
...
+  (null or empty string) is specified, a working Tracer implementation MUST be
+  returned as a fallback rather than returning null or throwing an exception,
+  its `name` property SHOULD be set to an **empty** string, and a message
+  reporting that the specified value is invalid SHOULD be logged. A library,
+  implementing the OpenTelemetry API *may* also ignore this name and return a
+  default instance for all calls, if it does not support "named" functionality
+  (e.g. an implementation which is not even observability-related). A
+  TracerProvider could also return a no-op Tracer here if application owners
+  configure the SDK to suppress telemetry produced by this library.
 - `version` (optional): Specifies the version of the instrumentation library (e.g. `1.0.0`).
```
jack-berg, line 127: Shouldn't `version`'s description have been updated as part of this PR?

arminru: Good catch, thanks! I created #2371 to track it since comments on merged PRs are often overlooked.

Issue #1929, opened 2021-09-15
Need for a trace data model specification


jmacd:
**What are you trying to achieve?**

The metrics and logs signals contain a file that describes the data model independent of the OTLP protocol. This is a place where we describe details that are too complex to put in a simple comment, providing a reference manual for use of the data.

The developing specification for probability sampling requires additions to the trace data model, and there is no appropriate file location to put this content.

We should create a file that explains what a span is, what the fields conceptually mean, and what constraints to expect on their fields (e.g., start time should be before end time).

**Additional context.**

Metrics data model: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md

Logs data model: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md

jmacd:
https://github.com/open-telemetry/oteps/pull/175 adds proposed text to add to such a specification.

yurishkuro:
+1

PR #2278, opened 2022-01-19
Allow adding links after span creation

### LINKED ISSUE #454: Please (re)-allow recording links after Span creation time
It seems that during https://github.com/open-telemetry/opentelemetry-specification/pull/258/, Adding links after Span creation was removed due to complexities around samplers. While I do appreciate the complexity, I wanted to highlight a scenario where this feature is needed and suggest a potential solution for how this can be accommodated, partially addressing the sampling decision issue.

## Example Scenario

Consider a system that waits for several events and processes them as a batch, for example for a subscription in a real-time system. These are stored in a `Queue` and each event has its OWN injected `SpanContext`. These are then consumed in a single `flush()` Operations.

### First event arrives
`Queue=[E1(SpanContext1)]` and a `flush()` is scheduled for the Subscription

### Several events arrive in the meantime
`Queue=[E1(SpanContext1),E2(SpanContext2),E3(SpanContext3)]`

### `flush()` is invoked
A Span gets created for the flushing of the events. In this case, I consider that the first event triggered the flush and hence makes sense to be the parent. So I might do something like:

```cpp
message = queue.pop();
span = tracer.CreateSpan(name="flush", parent=message.span_context)
do {
  do_something(message);
  span.AddLink(message.span_context); // This might not be needed since the first is also parent
  if ( /* early escape condition */ ) { // Lets say downstream is full
    if (!queue.empty()) {
      enqueueNextFlush(); // still messages to flush
    }
    break; 
  }
} while ( message = queue.pop() );
```

Let's say the escape condition was triggered after the second message... I expect the following Spans:
`Span1(name="flush", parent="SpanContext1", links=["SpanContext1", "SpanContext2"])`
`Span2(name="flush", parent="SpanContext3", links=["SpanContext3"])`

### Additional Notes:
In this case, let's say that our events (E1, E2) are "compressed/conflated" and it no longer makes sense to track them independently. 

Also, it is not desired for the queue to be consumed ahead of processing (and for that matter, the escape condition may not yet be known until the processing).

### Consider only the `parent` as guaranteed at creation.

The parent is necessary for `traceId` propagation already, and inheriting the default sampling decision would normally make sense purely from the parent. We can, by convention, recommend that sampling decisions should only consider this link in the sampling decision, although nothing will prevent implementations from using links if they are available in the API.

### Allow late links with some limitations.

The limitations can be clearly documented, such as the inability to consider them for sampling decisions and any future use case that is affected by this behaviour.

### Consider mutable sampling decision

This suggestion goes way beyond the scope of this feature, but its worth considering its impact on this requirement.

Consider the ability for tracers to "reconsider" the sampling decision as Span events (in this case any change to the span, e.g. operation_name, attributes, events, links are changed). This ultimately creates further complexities with propagation which may enforce constraints such as changes to sampling properties only takes effect from that point onwards. I'm happy to explore this further in a separate issue.

pyohannes:
Fixes #454

In a discussion in the specification SIG, it was deemed feasible to revise a further decision about not allowing adding links to spans after span creation.

Adding links after span creation allows to create more meaningful traces in certain scenarios (several of those scenarios were identified by the messaging workgroup). Best-effort sampling based on links is still possible, only based on links added at creation time (similar to attributes).

## Changes

This moves the section about adding span links from "Span creation" to "Span operations", removing mentions of the limitation that links can only be added at creation time. The name `AddLink` is proposed, wording is kept consistent with the corresponding section for adding attributes.

### DIFF
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -25,11 +25,11 @@
 - [Span](#span)
   * [Span Creation](#span-creation)
     + [Determining the Parent Span from a Context](#determining-the-parent-span-from-a-context)
-    + [Specifying links](#specifying-links)
   * [Span operations](#span-operations)
     + [Get Context](#get-context)
     + [IsRecording](#isrecording)
     + [Set Attributes](#set-attributes)
+    + [Add Links](#add-links)
     + [Add Events](#add-events)
     + [Set Status](#set-status)
     + [UpdateName](#updatename)
@@ -284,7 +284,7 @@ the entire operation and, optionally, one or more sub-spans for its sub-operatio
 - A start timestamp
 - An end timestamp
 - [`Attributes`](../common/common.md#attributes)
-- A list of [`Link`s](#specifying-links) to other `Span`s
+- A list of [`Link`s](#add-links) to other `Span`s
 - A list of timestamped [`Event`s](#add-events)
 - A [`Status`](#set-status).
 
@@ -369,8 +369,14 @@ The API MUST accept the following parameters:
   The API documentation MUST state that adding attributes at span creation is preferred
   to calling `SetAttribute` later, as samplers can only consider information
   already present during span creation.
+- [`Link`s](../overview.md#links-between-spans) - an ordered sequence of links.
+  Additionally, these links may be used to make a sampling decision as
+  noted in [sampling description](sdk.md#sampling). An empty collection will be
+  assumed if not specified.
 
-- `Link`s - an ordered sequence of Links, see API definition [here](#specifying-links).
+  The API documentation MUST state that adding links at span creation is
+  preferred to calling `AddLink` later, as samplers can only consider information
+  already present during span creation.
 - `Start timestamp`, default to current time. This argument SHOULD only be set
   when span creation time has already passed. If API is called at a moment of
   a Span logical start, API user MUST NOT explicitly set this argument.
@@ -404,30 +410,6 @@ If there is no `Span` in the `Context`, the newly created `Span` will be a root 
 A `SpanContext` cannot be set as active in a `Context` directly, but by
 [wrapping it into a Span](#wrapping-a-spancontext-in-a-span).
 For example, a `Propagator` performing context extraction may need this.
-
-#### Specifying links
-
-During `Span` creation, a user MUST have the ability to record links to other
-`Span`s. Linked `Span`s can be from the same or a different trace -- see [Links
-between spans](../overview.md#links-between-spans). `Link`s cannot be added after
-Span creation.
-
-A `Link` is structurally defined by the following properties:
-
-- `SpanContext` of the `Span` to link to.
-- Zero or more [`Attributes`](../common/common.md#attributes) further describing
-  the link.
-
-The Span creation API MUST provide:
-
-- An API to record a single `Link` where the `Link` properties are passed as
-  arguments. This MAY be called `AddLink`. This API takes the `SpanContext` of
-  the `Span` to link to and optional `Attributes`, either as individual
-  parameters or as an immutable object encapsulating them, whichever is most
-  appropriate for the language. Implementations MAY ignore links with an
-  [invalid](#isvalid) `SpanContext`.
-
-Links SHOULD preserve the order in which they're set.
 
 ### Span operations
 
@@ -496,6 +478,38 @@ attributes"](semantic_conventions/README.md) that have prescribed semantic meani
 Note that [Samplers](sdk.md#sampler) can only consider information already
 present during span creation. Any changes done later, including new or changed
 attributes, cannot change their decisions.
+
+#### Add Links
+
+A `Span` MUST have the ability to add `Link`s associated with it. Linked
+`Span`s can be from the same or a different trace (see
+[Links between spans](../overview.md#links-between-spans)).
+
+A `Link` is structurally defined by the following properties:
+
+- `SpanContext` of the `Span` to link to.
+- Zero or more [`Attributes`](../common/common.md#attributes) further describing
+  the link.
+
+The Span interface MUST provide:
+
+- An API to record a single `Link` where the `Link` properties are passed as
+  arguments. This MAY be called `AddLink`. This API takes the `SpanContext` of
+  the `Span` to link to and optional `Attributes`, either as individual
+  parameters or as an immutable object encapsulating them, whichever is most
+  appropriate for the language. Implementations MAY ignore links with an
+  [invalid](#isvalid) `SpanContext`.
+
+The Span interface MAY provide:
+
+- An API to set multiple `Link`s at once, where the `Link`s are passed in a
+  single method call.
+
+Links SHOULD preserve the order in which they're set.
+
+Note that [Samplers](sdk.md#sampler) can only consider information already
+present during span creation. Any `Link`s added later cannot change their
+decisions.
 
 #### Add Events
 
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -481,7 +481,7 @@ attributes, cannot change their decisions.
 
 #### Add Links
 
-A `Span` MUST have the ability to add `Link`s associated with it. Linked
+A `Span` SHOULD have the ability to add `Link`s associated with it. Linked
 `Span`s can be from the same or a different trace (see
 [Links between spans](../overview.md#links-between-spans)).
 
@@ -491,7 +491,7 @@ A `Link` is structurally defined by the following properties:
 - Zero or more [`Attributes`](../common/common.md#attributes) further describing
   the link.
 
-The Span interface MUST provide:
+The Span interface SHOULD provide:
 
 - An API to record a single `Link` where the `Link` properties are passed as
   arguments. This MAY be called `AddLink`. This API takes the `SpanContext` of
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -505,7 +505,7 @@ The Span interface MAY provide:
 - An API to set multiple `Link`s at once, where the `Link`s are passed in a
   single method call.
 
-Links SHOULD preserve the order in which they're set.
+Spans SHOULD preserve the order in which Links are set.
 
 Note that [Samplers](sdk.md#sampler) can only consider information already
 present during span creation. Any `Link`s added later cannot change their
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ release.
 
 ### Traces
 
+- Introduce the concept of Instrumentation Scope to replace/extend Instrumentation
+  Library. The Tracer is now associated with Instrumentation Scope
+  ([#2276](https://github.com/open-telemetry/opentelemetry-specification/pull/2276)).
 - Add documentation REQUIREMENT for adding attributes at span creation.
   ([#2383](https://github.com/open-telemetry/opentelemetry-specification/pull/2383)).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,8 @@ release.
 - Introduce the concept of Instrumentation Scope to replace/extend Instrumentation
   Library. The Tracer is now associated with Instrumentation Scope
   ([#2276](https://github.com/open-telemetry/opentelemetry-specification/pull/2276)).
-- Add documentation REQUIREMENT for adding attributes at span creation.
-  ([#2383](https://github.com/open-telemetry/opentelemetry-specification/pull/2383)).
+- Allow adding links after span creation, add new `AddLink` method.
+  ([#2278](https://github.com/open-telemetry/opentelemetry-specification/pull/2278))
 
 ### Metrics
 
@@ -49,6 +49,9 @@ release.
 - Add `OTEL_EXPORTER_JAEGER_PROTOCOL` environment variable to select the protocol
   used by the Jaeger exporter.
   ([#2341](https://github.com/open-telemetry/opentelemetry-specification/pull/2341))
+- Add documentation REQUIREMENT for adding attributes at span creation.
+  ([#2383](https://github.com/open-telemetry/opentelemetry-specification/pull/2383)).
+
 ### Metrics
 
 - Initial Prometheus <-> OTLP datamodel specification
@@ -181,14 +184,9 @@ release.
   variable to point to the correct HTTP port and correct description of
   `OTEL_TRACES_EXPORTER`.
   ([#2333](https://github.com/open-telemetry/opentelemetry-specification/pull/2333))
-<<<<<<< HEAD
-=======
 - Add `OTEL_EXPORTER_JAEGER_PROTOCOL` environment variable to select the protocol
   used by the Jaeger exporter.
   ([#2341](https://github.com/open-telemetry/opentelemetry-specification/pull/2341))
-- Allow adding links after span creation, require new `AddLink` method.
-  ([#2278](https://github.com/open-telemetry/opentelemetry-specification/pull/2278))
->>>>>>> 477857e (Add changelog and adapt spec compliance matrix)
 
 ### Metrics
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ release.
 
 ### Traces
 
+- Add documentation REQUIREMENT for adding attributes at span creation.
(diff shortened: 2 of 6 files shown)


bogdandrutu:
My 2cents: By approving this PR we are allowing instrumentation to use the new API, instead of trying to pass all the links at start time. This will cause instrumentations to not even consider grouping all the links at the start which will mean in lots of cases the instrumentation will produce incomplete traces since the sampler is not working in this case.

We are opening a pandora box without thinking of implications, and this may cause troubles on short to medium term until a proper solution for the samplers to work is proposed.

pyohannes:
> By approving this PR we are allowing instrumentation to use the new API, instead of trying to pass all the links at start time. This will cause instrumentations to not even consider grouping all the links at the start which will mean in lots of cases the instrumentation will produce incomplete traces since the sampler is not working in this case.

That's a valid point: this might change the shape of some traces when one is using custom samplers (none of the specified samplers use links for now).

> We are opening a pandora box without thinking of implications, and this may cause troubles on short to medium term until a proper solution for the samplers to work is proposed.

I think it would be beneficial to make a decision as soon as possible (whether to allow links after creation time or not to) for the following reasons:

1. It will give us a clear framework to work with when coming up with semantic conventions. If we now decide to go with a sub-optimal solution that is based on the current limitation, we will be stuck with this sub-optimal solution (as we want to stabilize some semantic conventions this quarter), even when the spec will remove the limitation at a later point.
2. When we want to remove this limitation, it's probably better to do it know, better than later. Links are not widely used now, but probably will be used more and more with time. The longer we wait with a decision, the harder it will be to remove the limitation.

yurishkuro:
I am not a fan of how this PR makes references to "SIG discussions" and "working groups" and yet lists not a single use case. We cannot change specs without documented explanation.

pyohannes:
> This will cause instrumentations to not even consider grouping all the links at the start which will mean in lots of cases the instrumentation will produce incomplete traces since the sampler is not working in this case.

Sampling in connection with links is a very tricky thing, as I found out by implementing a link-based sampler myself. Even with the current model, it's impossible to avoid incomplete traces: how will you sample a span that has a non-sampled parent, but sampled links? The only way to _guarantee_ complete traces would be to limit adding links only to root spans of a trace.

Producing incomplete traces with link-based samplers is not a problem that is introduced with this change, although this change increases the scope of this problem.

carlosalberto:
> We are opening a pandora box without thinking of implications, and this may cause troubles on short to medium term until a proper solution for the samplers to work is proposed.

The same happens for attributes, but we allow that anyway. I feel like a 'big' warning should be put in the related sections, mentioning that instrumentation code SHOULD try its best to set links/attributes at `Span` build time.

jmacd:
I want to add a few remarks about sampling and "completeness", since that is a major topic in #2047.

> By approving this PR we are allowing instrumentation to use the new API, instead of trying to pass all the links at start time. This will cause instrumentations to not even consider grouping all the links at the start which will mean in lots of cases the instrumentation will produce incomplete traces since the sampler is not working in this case.

Challenging this statement @bogdandrutu: if there's a real problem created as a result of the new API, that is to say that many developers opt to use AddLink() instead of passing to Start(), then users will have incomplete traces and they may be motivated to ask why the instrumented code did not use the Start() API. If there's a will and a way, the right thing will happen. OTOH, the reason for this change is that sometimes there is literally not a way to know your links when you start, so we should allow this new API.

I also want to challenge the idea that anyone using links actually expects a guarantee of "completeness" when they are sampling. Actually, the term "completeness" is not defined the way it's being used here. To me, the point of a Link is to connect one trace to another, and there is no notion of completeness for trace and its linked-to-traces that we have defined (see #1929). Completeness is about whether all the spans for a given TraceID have been collected (per #2047). 

I do not expect linked-to traces to be collected with a 100% guarantee if I'm sampling. One reason to use links is to avoid huge or never-ending traces, so I am usually explicitly not interested in having them all collected. There are two directions a user could go if they want to ensure a sampled trace has its linked-to traces also collected. 

1. Non-probabilistically: if the cost of tracing is not a concern, making a non-probabilistic decision to sample a trace whenever any of its provided-at-start-time linked-to contexts is also being traced. If the cost of tracing is a concern, it's not clear how to rate-limit this decision.
2. Probabilistically: if probability sampling is being used, then each linked-to trace has an independent probability of being collected. If you have 1000 links and use 1/2 sampling for the linked-to traces, you should expect approximately 500 of them to be collected. (This, to me, is working as intended.)

jmacd:
> how will you sample a span that has a non-sampled parent, but sampled links? The only way to guarantee complete traces would be to limit adding links only to root spans of a trace.

#2047 defines a notion of sub-trace completeness that can be used here. Generally speaking, with sampling you have the potential for incompleteness, think of it as a feature not a bug. 😁

jmacd (approved):

specification/trace/api.md
```diff
@@ -494,6 +475,38 @@ Note that [Samplers](sdk.md#sampler) can only consider information already
 present during span creation. Any changes done later, including new or changed
 attributes, cannot change their decisions.
 
+#### Add Links
+
+A `Span` MUST have the ability to add `Link`s associated with it. Linked
```
yurishkuro, line 480: nit: a MUST to have a new method in a stable API is a breaking change. I suggest making it a SHOULD.

jmacd:
> several of those scenarios were identified by the messaging workgroup

It may be helpful to link to a summary of those scenarios.

> the reason for this change is that sometimes there is literally not a way to know your links when you start 

One of the reasons I've wanted to add links after a span starts is:

a. Span A and B start 
b. Span A acquires a shared resource
c. Span B tries to acquire a shared resource, but is blocked by Span A
d. Instrumentation adds a link to Span B ("blocked by")
e. Instrumentation adds a link to Span A ("blocking")

specification/trace/api.md
pyohannes: Changed from `MUST` to `SHOULD`.

pyohannes:
> The same happens for attributes, but we allow that anyway. I feel like a 'big' warning should be put in the related sections, mentioning that instrumentation code SHOULD try its best to set links/attributes at Span build time.

@carlosalberto This is mentioned (I took this over from attributes), I'm not sure if it should be emphasized more:
> Whenever possible, users SHOULD set any already known links at span creation instead of calling `AddLink` later.

pyohannes:
> > several of those scenarios were identified by the messaging workgroup

> It may be helpful to link to a summary of those scenarios.

One is briefly sketched here: https://github.com/open-telemetry/opentelemetry-specification/issues/454#issuecomment-1012340002

We also plan to present our current design for messaging instrumentation to a wider audience, this will be covered there: https://github.com/open-telemetry/opentelemetry-specification/issues/454#issuecomment-1017718100

pyohannes:
I will bring this up in the next specification SIG meeting, to discuss next steps for resolving open questions.

specification/trace/api.md
```diff
@@ -402,30 +407,6 @@ A `SpanContext` cannot be set as active in a `Context` directly, but by
 [wrapping it into a Span](#wrapping-a-spancontext-in-a-span).
 For example, a `Propagator` performing context extraction may need this.
 
-#### Specifying links
```
Aneurysm9, line 405: Should we keep this section?  I don't think that we can remove the existing API for specifying links at creation time.  Instead, we should simply remove the last sentence of the first paragraph here:

> ~~`Link`s cannot be added after Span creation~~

pyohannes: > I don't think that we can remove the existing API for specifying links at creation time.

Right, the API for creating a span is still required to accept a list of links (see above). In addition, a method for adding links after span creation is added as a _should_.

I tried to structure this in a way similar to attributes, where the situation is similar (attributes can be added at creation time and after span creation). I'm open to change this, if it's clearer in another way.

pyohannes: > Perhaps we can create a separate "Links" subsection that describes what a link is (arguably the same could be done for Events however Events don't make sense at create time)

Any other opinions on this? Otherwise, I'll go ahead the change the PR according to the suggestion.

specification/trace/api.md
bogdandrutu: @yurishkuro why is a "breaking" change? I am a bit confused about this, maybe we should define what is and what is not allowed to do.

yurishkuro: if I implement an interface and you add a method to it, my code might not compile anymore.

pyohannes: > maybe we should define what is and what is not allowed to do.

It is defined: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#api-stability

According to what's required there, I think @yurishkuro is correct that this cannot be a MUST.

specification/trace/api.md
```diff
@@ -366,8 +366,13 @@ The API MUST accept the following parameters:
...
   Whenever possible, users SHOULD set any already known attributes at span creation
   instead of calling `SetAttribute` later.
+- [`Link`s](../overview.md#links-between-spans) - an ordered sequence of links.
+  Additionally, these links may be used to make a sampling decision as
+  noted in [sampling description](sdk.md#sampling). An empty collection will be
+  assumed if not specified.
 
-- `Link`s - an ordered sequence of Links, see API definition [here](#specifying-links).
+  Whenever possible, users SHOULD set any already known links at span creation
+  instead of calling `AddLink` later.
```
MrAlias, line 375: This sentence is out of place. It is for the user of OpenTelemetry, this specification is for developers implementing OpenTelemetry APIs and SDKs.

Possibly, this should be restated as an action for the intended audience by documenting the constraint in the `AddLink` section?

specification/trace/api.md
```diff
@@ -494,6 +475,38 @@ Note that [Samplers](sdk.md#sampler) can only consider information already
...
+  parameters or as an immutable object encapsulating them, whichever is most
+  appropriate for the language. Implementations MAY ignore links with an
+  [invalid](#isvalid) `SpanContext`.
+
+The Span interface MAY provide:
+
+- An API to set multiple `Link`s at once, where the `Link`s are passed in a
+  single method call.
+
+Links SHOULD preserve the order in which they're set.
```
MrAlias, line 504:
```suggestion
Spans SHOULD preserve the order in which Links are set.
```

Right?

specification/trace/api.md
```diff
@@ -494,6 +475,38 @@ Note that [Samplers](sdk.md#sampler) can only consider information already
...
+
+Note that [Samplers](sdk.md#sampler) can only consider information already
+present during span creation. Any `Link`s added later cannot change their
+decisions.
```
MrAlias, line 508:
```suggestion
decisions. This limitation SHOULD be documented where the add links functionality is implemented.
```

specification/trace/api.md
MrAlias: > I tried to be consistent with the wording used for attributes (see above):

Ah, I see where this came from. Thanks for pointing it out.
I still think we should not copy poorly written specification. Doing so will only duplicate the original error.

> I guess if we do the suggested change, we should do it for both?

Agreed, we should try to remain consistent. Though I would make the change to the existing specification as a separate PR so as not to overload this.

bogdandrutu:
@jmacd

> One of the reasons I've wanted to add links after a span starts is:
a. Span A and B start
b. Span A acquires a shared resource
c. Span B tries to acquire a shared resource, but is blocked by Span A
d. Instrumentation adds a link to Span B ("blocked by")
e. Instrumentation adds a link to Span A ("blocking")

This is a great example, but in this case the "link" has also associated a "relation".  Which I think is important to have if we start allowing links after.

The "default" case that we had so far was batching and the `child of` relation. With this new proposal we open the possibilities to way more relations and we probably need to think about this and add that to the API probably.

As per @pyohannes picture for the "Pull scenarios", the relation between the Span and the "linked context" is no longer a `child of`, so I think we need to document that in the link. The relation can be a semantic convention, but I think is important to have it.

@pyohannes for the "Pull scenarios":

Somehow I don't like the proposed way to model this, the reason is that (independent of adding links at start/end) I would prefer to have another Span `Execute` that wraps the 2 `console.log` ops as a sibling to `Receive`. That would feel much more natural.

pyohannes:
@bogdandrutu 

> The "default" case that we had so far was batching and the child of relation. With this new proposal we open the possibilities to way more relations and we probably need to think about this and add that to the API probably.

I think in this context, even links in batching use cases modelled "child-of" relations: it's the case where one child has multiple parents.

> As per @pyohannes picture for the "Pull scenarios", the relation between the Span and the "linked context" is no longer a child of, so I think we need to document that in the link. The relation can be a semantic convention, but I think is important to have it.

Yes, links could greatly benefit from additional semantic context. In the messaging use cases, we'd always have links between one span with kind producer, and one span with kind consumer. This way semantic context could be inferred from the spans that are linked. But having it explicitly attached to a link. As it is already possible to attach attributes to a link, it's might just be a matter of coming up with semantic conventions for attributes on links?

> @pyohannes for the "Pull scenarios":
>
> Somehow I don't like the proposed way to model this, the reason is that (independent of adding links at start/end) I would prefer to have another Span Execute that wraps the 2 console.log ops as a sibling to Receive. That would feel much more natural.

As mentioned, this is independent of adding links after span creation. In the messaging workgroup, we were discussing "Process" spans (same as "Execute" spans). The aim was to come up with a model that can easily be extended by "Process" spans. However, we're not yet clear about all the details, and we'd also like to clear up whether we'd be able to create a "Receive" span as sketched above, as "Process" spans might look different in a different model.

specification/trace/api.md
pyohannes: I opened a PR for this: #2383

pyohannes: An update for attributes was merged in the separate PR #2383, the text for links is updated in this PR.

pyohannes:
@bogdandrutu @arminru

What are your recommendations regarding this PR? Do you think it makes sense to continue working on this PR, or do you think it needs to be reconsidered and made part of a larger effort?

There are lots of questions around links coming up (as people start to use them), so do you think it makes sense to write an OTEP addressing those questions, before continuing on spec PRs?

pyohannes:
We will continue discussing this issue with @bogdandrutu in one of the next messaging workgroup meetings.

pyohannes:
This is still being discussed with @bogdandrutu in the Thursday messaging workgroup.

nerochiaro:
I came here looking for a resolution of this issue, but it seems that the last significant comment was back in April

> This is still being discussed with @bogdandrutu in the Thursday messaging workgroup.

I am curious to understand if the discussion has simply stalled and has been abandoned or if there is a possibility that at some point opentelemetry will allow adding links to spans that have already been created.

pyohannes:
> I am curious to understand if the discussion has simply stalled and has been abandoned or if there is a possibility that at some point opentelemetry will allow adding links to spans that have already been created.

The discussion has stalled but is not yet settled, issue #454 is still open. @nerochiaro Maybe you could add your use case for adding span links after span creation to this issue?

nerochiaro:
I added more comments to issue #454, but in short the use case would be when using the lambda automatic instrumentation.

A span for the lambda function is created automatically by the instrumentation's function wrapper, and there is no way to tell it to add links (for example, a link to the span created by a lambda authorizer that ran before the function itself).

Oberon00:
For that use case you don't need to add links after creation, if you can tell the wrapper to add the links right away.

nerochiaro:
> For that use case you don't need to add links after creation, if you can tell the wrapper to add the links right away.

Except that, as far as I can tell, the wrapper creates the span then runs the lambda, without accepting any input.
I would have to write my own wrapper which duplicates all the code of the default one, and add code to create links to that.

(Or possibly the wrapper could accept a configuration option that specifies a "hook" function that given the event and context would generate extra links and attributes to add to the lambda, but it seems convoluted and with drawbacks. see my comment on #454 for more details.)

I'm not saying it's not feasible, but certainly not ideal, especially when lambdas running behind API Gateway with an authorizer lambda are a pretty common case.

joaopgrassi:
I removed the stale label, but I wonder if we should focus the discussion on the issue or here. If here, then maybe re-opening the PR to get attention again?

carlosalberto:
I'd recommend discussing this in the issue itself (e.g. sometimes there's more than one possible PR for a given issue, so general discussion should be in the issue itself).

Issue #2232, opened 2021-12-20
Allow to unregister/stop/destroy instruments


mateuszrzeszutek:
**What are you trying to achieve?**

I'm currently working on the micrometer->OTel bridge instrumentation in the javaagent. Micrometer offers the possibility to remove a meter from the `MeterRegistry` and stop emitting whatever metrics it used to collect. For example, suppose you use a database connection pool that's instrumented with metrics - when you close/destroy the whole pool you probably want to stop collecting any metrics associated to it (because it doesn't exist anymore). This is useful for both asynchronous instruments (since once they're registered there's no way to stop them) and synchronous instruments (you can just stop using them, but the metrics SDK will still send in the last recorded value).

**Additional context.**

Micrometer bridge PR: open-telemetry/opentelemetry-java-instrumentation#4919

CC @jsuereth

tigrannajaryan:
(Following https://github.com/open-telemetry/opentelemetry-specification/blob/main/CONTRIBUTING.md#issue-triaging process).

@jsuereth I am reassigning this to you, I believe you know the context of this better.

bogdandrutu:
@mateuszrzeszutek micrometer is an API + Some implementation(s). I am not sure why that should translate into calls to our API instead of offering an implementation of micrometer that implements the `MetricProducer`, see https://github.com/open-telemetry/opentelemetry-java/blob/44dccd6c33c7c4cb58e1cc63f0db02b3bfec1804/opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/metrics/OpenCensusMetricProducer.java#L24

/cc @jsuereth

anuraaga:
@bogdandrutu Micrometer was just one example but do you think apps should not have a way to stop reporting metrics? Dropwizard metrics in Java also has `registry.remove` and it seems idiomatic for a library that allows registering a stateful object to always allow unregistering it. While the whole `MeterProvider` could be closed, metric definitions are generally completely separate from the MeterProvider lifecycle and that probably doesn't solve that case.

mateuszrzeszutek:
There are advantages to calling OTel API in micrometer bridge instrumentation:
* Micrometer instruments like `Timer` and `DistributionSummary` can be translated to native OTel histograms; micrometer does not have a "histogram" meter by itself, it basically exports each bucket as a separate gauge.
[...]

PR #2281, opened 2022-01-20
Make multi-instrument idioms for async callbacks possible

### LINKED ISSUE #2280: Enable multi-instrument callbacks in the metrics API

```
}
```

This feels idiomatic to me. 

jmacd:
Fixes #2280 

## Changes

This changes slightly the requirement for how instruments are associated with callbacks. Whereas before the spec required one instrument per callback, this would allow multiple instruments. A motivating example is given in #2280.

Related issues #2270 #2249


jack-berg:
I believe we also need to loosen the requirement that APIs produce an error when multiple instruments are [registered with the same name](https://github.com/open-telemetry/opentelemetry-specification/blame/main/specification/metrics/api.md#L186-L187). 

Does that make sense to do in this PR or in a separate one?

jmacd:
@jack-berg this PR definitely depends on #2270.

specification/metrics/api.md
```diff
@@ -253,12 +253,13 @@ asynchronous:
...
 * Asynchronous instruments (e.g. [Asynchronous Gauge](#asynchronous-gauge)) give
-  the user a way to register callback function, and the callback function will
+  the user a way to register one or more associated callback functions, which will
   only be invoked upon collection. For example, a piece of embedded software
   could use an asynchronous gauge to collect the temperature from a sensor every
   15 seconds, which means the callback function will only be invoked every 15
-  seconds. [Measurements](#measurement) recorded by asynchronous instruments
-  cannot be associated with the [Context](../context/context.md).
+  seconds. Multiple callbacks may be associated with one instrument as a result
+  of duplicate instrument registration. [Measurements](#measurement) recorded by
```
Aneurysm9, line 261: Does anything need to be specified regarding what should happen when multiple callbacks are associated with one instrument?

jmacd: Maybe. I was hoping this is covered in #2270.

jsuereth (approved):
Thanks for tweaking this @jmacd !

jack-berg (approved):

jack-berg:
It seems natural that if multiple callbacks allowed, that there should be some way of unregister / cancel / stop callbacks #2232. But I think that can be solved separately.

MrAlias (approved):

anuraaga (approved):

specification/metrics/api.md
jmacd: See #2226 for debate.

jmacd:
Reviewers, as noted in https://github.com/open-telemetry/opentelemetry-specification/issues/2226#issuecomment-1028243274 I am closing this PR. This work will be integrated into #2270 following the discussion in #2226.

PR #2292, opened 2022-01-25
Add semantic conventions for jvm cpu metrics

jonatan-ivanov:
Initial draft for JVM runtime cpu semantic conventions.
I created a gist where you can see where these values are coming from: https://gist.github.com/jonatan-ivanov/0449619cb6f851ce67d70e47a6a1fb73

/cc @trask @jack-berg


specification/metrics/semantic_conventions/runtime-environment-metrics.md
```diff
@@ -55,5 +55,13 @@ consider, for example pthreads vs green thread implementations.
...
-|------|-------------|-------|-----------------|------------|---------------|------------------|
+All JVM metric attributes are required unless otherwise indicated.
+
+| Name                                           | Description                                              | Unit    | Unit ([UCUM](README.md#instrument-units)) | Instrument Type            | Value Type | Attribute Key | Attribute Values      |
+|------------------------------------------------|----------------------------------------------------------|---------|-------------------------------------------|----------------------------|------------|---------------|-----------------------|
+| process.runtime.jvm.cpu.time                   | The CPU time used by the process                         | seconds | `s`                                       | Asynchronous Counter       | Double     |               |                       |
+| process.runtime.jvm.cpu.usage                  | The "recent cpu usage" for the process                   | 1       | 1                                         | Gauge                      | Double     |               |                       |
```
jack-berg, line 63: I believe `usage` should be changed to `utilization` to match conventions elsewhere:
- [system metric semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/system-metrics.md
- Semantic convention [instrument naming](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/metrics/semantic_conventions#instrument-naming)

specification/metrics/semantic_conventions/runtime-environment-metrics.md
```diff
@@ -55,5 +55,13 @@ consider, for example pthreads vs green thread implementations.
...
+| process.runtime.jvm.common.pool.parallelism    | Targeted parallelism level of the common pool            | 1       | 1                                         | Gauge                      | Int32      |               |                       |
+| process.runtime.jvm.system.cpu.usage           | The "recent cpu usage" for the whole system              | 1       | 1                                         | Gauge                      | Double     |               |                       |
+| process.runtime.jvm.system.cpu.count           | Number of processors available to use                    | 1       | 1                                         | Gauge                      | Int32      |               |                       |
+| process.runtime.jvm.system.cpu.load.average.1m | Average CPU load of the whole system for the last minute | 1       | 1                                         | Gauge                      | Double     |               |                       |
```
jack-berg, line 67: I assume this will come from (OperatingSystemMXBean#getSystemLoadAverage())[https://docs.oracle.com/javase/7/docs/jre/api/management/extension/com/sun/management/OperatingSystemMXBean.html#getSystemCpuLoad()]. 

The definition of load defined in `#getSystemLoadAverage()` seems pretty esoteric. Should we include it in the semantic conventions?

specification/metrics/semantic_conventions/runtime-environment-metrics.md
```diff
@@ -55,5 +55,13 @@ consider, for example pthreads vs green thread implementations.
...
+| process.runtime.jvm.cpu.usage                  | The "recent cpu usage" for the process                   | 1       | 1                                         | Gauge                      | Double     |               |                       |
+| process.runtime.jvm.common.pool.parallelism    | Targeted parallelism level of the common pool            | 1       | 1                                         | Gauge                      | Int32      |               |                       |
+| process.runtime.jvm.system.cpu.usage           | The "recent cpu usage" for the whole system              | 1       | 1                                         | Gauge                      | Double     |               |                       |
+| process.runtime.jvm.system.cpu.count           | Number of processors available to use                    | 1       | 1                                         | Gauge                      | Int32      |               |                       |
```
jack-berg, line 66: I wonder why there isn't a general (system instrument)[https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/system-metrics.md#systemcpu---processor-metrics] metric for cpu count.

specification/metrics/semantic_conventions/runtime-environment-metrics.md
```diff
@@ -55,5 +55,13 @@ consider, for example pthreads vs green thread implementations.
...
+|------------------------------------------------|----------------------------------------------------------|---------|-------------------------------------------|----------------------------|------------|---------------|-----------------------|
+| process.runtime.jvm.cpu.time                   | The CPU time used by the process                         | seconds | `s`                                       | Asynchronous Counter       | Double     |               |                       |
+| process.runtime.jvm.cpu.usage                  | The "recent cpu usage" for the process                   | 1       | 1                                         | Gauge                      | Double     |               |                       |
+| process.runtime.jvm.common.pool.parallelism    | Targeted parallelism level of the common pool            | 1       | 1                                         | Gauge                      | Int32      |               |                       |
```
jack-berg, line 64: Does this come from `ForkJoinPool#getPoolsize()`?

specification/metrics/semantic_conventions/runtime-environment-metrics.md
jsuereth: that's a good call out.

specification/metrics/semantic_conventions/runtime-environment-metrics.md
jonatan-ivanov: Yes, I created a gist where you can see where are these values are coming from: https://gist.github.com/jonatan-ivanov/0449619cb6f851ce67d70e47a6a1fb73

I also included some explanation and examples.
In this exact case, I guess we can copy from the javadoc or refer it in the spec to make the description more concise.

specification/metrics/semantic_conventions/runtime-environment-metrics.md
jonatan-ivanov: Should be either `ForkJoinPool.commonPool().getParallelism()` or `ForkJoinPool.getCommonPoolParallelism()`.

jack-berg:
Here's a summary of the proposed metrics and their corresponding sources. Hopefully this makes it a bit easier to evaluate the value of these. This summarizes the information from @jonatan-ivanov's [gist](https://gist.github.com/jonatan-ivanov/0449619cb6f851ce67d70e47a6a1fb73).
- `process.runtime.jvm.cpu.time` - [com.sun.management.OperatingSystemMXBean#getProcessCpuTime()](https://docs.oracle.com/javase/7/docs/jre/api/management/extension/com/sun/management/OperatingSystemMXBean.html#getProcessCpuTime())
- `process.runtime.jvm.cpu.utilization` - [com.sun.management.OperatingSystemMXBean#getProcessCpuLoad()](https://docs.oracle.com/javase/7/docs/jre/api/management/extension/com/sun/management/OperatingSystemMXBean.html#getProcessCpuLoad())
- `process.runtime.jvm.common.pool.parallelism` - [java.util.concurrent.ForkJoinPool#getParallelism()](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ForkJoinPool.html#getParallelism--)
- `process.runtime.jvm.system.cpu.utilization` -  [com.sun.management.OperatingSystemMXBean#getSystemCpuLoad()](https://docs.oracle.com/javase/7/docs/jre/api/management/extension/com/sun/management/OperatingSystemMXBean.html#getSystemCpuLoad())
- `process.runtime.jvm.system.cpu.count` - [java.lang.Runtime#availableProcessors()](https://docs.oracle.com/javase/7/docs/api/java/lang/Runtime.html#availableProcessors())
- `process.runtime.jvm.system.cpu.load.average.1m` - [java.lang.management.OperatingSystemMXBean#getSystemLoadAverage()](https://docs.oracle.com/javase/8/docs/api/java/lang/management/OperatingSystemMXBean.html#getSystemLoadAverage--)

jack-berg:
We discussed in the 2/24/2022 Java SIG meeting. Can java folks take a look at this and comment?

/cc @kittylyst @trask

My thoughts:
- `process.runtime.jvm.cpu.time`, `process.runtime.jvm.cpu.utilization` are obvious and we should merge those quickly.
- `process.runtime.jvm.common.pool.parallelism` seems more related to the thread pool topic we discussed. Maybe we should save it for a future PR where we declare the conventions for thread pools.
- `process.runtime.jvm.system.cpu.utilization` has overlap with the more general `system.cpu.utilization`, but we don't have access to the data required for the attributes. I think we should include this now.
- `process.runtime.jvm.system.cpu.count` I think we should incude this now.
- `process.runtime.jvm.system.cpu.load.average.1m` useful, but overlaps with `process.runtime.jvm.cpu.time` & `process.runtime.jvm.cpu.utilization`. I'd be fine skipping this for now.

trask:
I'm thinking now that it may be good to reuse `system.cpu.time`, `process.cpu.time`, and `system.cpu.count` since there aren't any JVM-specific attributes that we want to capture on them.

I have sent #2388 to clarify that these should be interpreted as container-level and not host-level metrics when emitted from inside a container.

Should we scale back this PR for now to just these two?
* `process.runtime.jvm.cpu.utilization`
* `process.runtime.jvm.system.cpu.utilization`

I noticed there's already a `system.cpu.utilization` metric (and so seems likely to be a `process.cpu.utilization` metric), but this is [defined](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/metrics/semantic_conventions#instrument-naming) in a way that I think is incompatible with what the JVM gives us:

> `system.cpu.utilization` is defined as the difference in `system.cpu.time` measurements divided by the elapsed time

so I don't think we should reuse in this case and need one under `process.runtime.jvm.*`

trask:
I think we have more clarity now from discussions in #2384 and #2388.

I think there's still one open question https://github.com/open-telemetry/opentelemetry-specification/pull/2392#issuecomment-1057222928 and then I think we'll have clarity on how to proceed.

@jonatan-ivanov thanks for your patience

specification/metrics/semantic_conventions/runtime-environment-metrics.md
```diff
@@ -57,18 +57,22 @@ consider, for example pthreads vs green thread implementations.
...
+|                                                |                                                          |         |                                           |                            |            | pool          | Name of pool [1]      |
+| process.runtime.jvm.memory.max                 | Measure of max obtainable memory                         | Bytes   | `By`                                      | Asynchronous UpDownCounter | Int64      | type          | `"heap"`, `"nonheap"` |
+|                                                |                                                          |         |                                           |                            |            | pool          | Name of pool [1]      |
+| process.runtime.jvm.cpu.time                   | The CPU time used by the process                         | seconds | `s`                                       | Asynchronous Counter       | Double     |               |                       |
```
trask, line 70: let's remove this and optimistically use `process.cpu.time` pending clarification in #2392 about using the `process.*` namespace

```suggestion
```

specification/metrics/semantic_conventions/runtime-environment-metrics.md
trask: let's remove this and optimistically use `process.cpu.count` pending clarification in #2392 about using the `process.*` namespace

```suggestion
```

specification/metrics/semantic_conventions/runtime-environment-metrics.md
```diff
@@ -57,18 +57,22 @@ consider, for example pthreads vs green thread implementations.
...
+| process.runtime.jvm.cpu.utilization            | The "recent cpu usage" for the process                   | 1       | 1                                         | Gauge                      | Double     |               |                       |
+| process.runtime.jvm.common.pool.parallelism    | Targeted parallelism level of the common pool            | 1       | 1                                         | Gauge                      | Int32      |               |                       |
```
trask, line 72: from @jack-berg above (below?)
> * `process.runtime.jvm.common.pool.parallelism` seems more related to the thread pool topic we discussed. Maybe we should save it for a future PR where we declare the conventions for thread pools.
```suggestion
```

specification/metrics/semantic_conventions/runtime-environment-metrics.md
trask: from @jack-berg

> * `process.runtime.jvm.system.cpu.load.average.1m` useful, but overlaps with `process.runtime.jvm.cpu.time` & `process.runtime.jvm.cpu.utilization`. I'd be fine skipping this for now.

I'd lean toward keeping this in the PR, especially if it's something that micrometer and others are already collecting

specification/metrics/semantic_conventions/runtime-environment-metrics.md
```diff
@@ -57,18 +57,19 @@ consider, for example pthreads vs green thread implementations.
...
+|                                                |                                                          |         |                                           |                            |            | pool          | Name of pool [1]      |
+| process.runtime.jvm.cpu.utilization            | The "recent cpu usage" for the process                   | 1       | 1                                         | Gauge                      | Double     |               |                       |
+| process.runtime.jvm.system.cpu.utilization     | The "recent cpu usage" for the whole system              | 1       | 1                                         | Gauge                      | Double     |               |                       |
+| process.runtime.jvm.system.cpu.load.average.1m | Average CPU load of the whole system for the last minute | 1       | 1                                         | Gauge                      | Double     |               |                       |
```
jack-berg, line 72: Can you change `Gauge` => `Asynchronous Gauge` for consistency with the other convetions, e.g. [system](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/system-metrics.md)?

jack-berg (approved):

specification/metrics/semantic_conventions/runtime-environment-metrics.md
trask, line 72: based on [this](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/system-metrics.md#systemos---os-specific-system-metrics) (and because it's shorter):

> An instrument for load average over 1 minute on Linux could be named `system.linux.cpu.load_1m`

```suggestion
| process.runtime.jvm.system.cpu.load_1m | Average CPU load of the whole system for the last minute | 1       | 1                                         | Gauge                      | Double     |               |                       |
```

trask (approved):

jonatan-ivanov:
Could somebody please approve the workflow so that the linters can check it?

trask:
@jonatan-ivanov #2426 will fix the failing `markdown-link-check`

anuraaga (approved):

specification/metrics/semantic_conventions/runtime-environment-metrics.md
```diff
@@ -57,18 +57,19 @@ consider, for example pthreads vs green thread implementations.
...
+| process.runtime.jvm.memory.max                 | Measure of max obtainable memory                         | Bytes   | `By`                                      | Asynchronous UpDownCounter | Int64      | type          | `"heap"`, `"nonheap"` |
+|                                                |                                                          |         |                                           |                            |            | pool          | Name of pool [1]      |
+| process.runtime.jvm.cpu.utilization            | The "recent cpu usage" for the process                   | 1       | 1                                         | Asynchronous Gauge         | Double     |               |                       |
+| process.runtime.jvm.system.cpu.utilization     | The "recent cpu usage" for the whole system              | 1       | 1                                         | Asynchronous Gauge         | Double     |               |                       |
```
trask, line 71:
```suggestion
| process.runtime.jvm.cpu.utilization            | Recent cpu utilization for the process                   | 1       | 1                                         | Asynchronous Gauge         | Double     |               |                       |
| process.runtime.jvm.system.cpu.utilization     | Recent cpu utilization for the whole system              | 1       | 1                                         | Asynchronous Gauge         | Double     |               |                       |
```

mateuszrzeszutek (approved):

specification/metrics/semantic_conventions/runtime-environment-metrics.md
```diff
@@ -57,18 +57,19 @@ consider, for example pthreads vs green thread implementations.
...
+| process.runtime.jvm.memory.max                 | Measure of max obtainable memory                         | Bytes   | `By`                                      | Asynchronous UpDownCounter | Int64      | type          | `"heap"`, `"nonheap"` |
+|                                                |                                                          |         |                                           |                            |            | pool          | Name of pool [1]      |
+| process.runtime.jvm.cpu.utilization            | Recent cpu utilization for the process                   | 1       | 1                                         | Asynchronous Gauge         | Double     |               |                       |
+| process.runtime.jvm.system.cpu.utilization     | Recent cpu utilization for the whole system              | 1       | 1                                         | Asynchronous Gauge         | Double     |               |                       |
```
trask, line 71: see recent discussion in https://github.com/open-telemetry/opentelemetry-specification/pull/2392#discussion_r833018035 (and follow-ups in #2435 and #2436)

with [2] being something like: These utilizations are not defined as being for the specific interval since last measurement (unlike `system.cpu.utilization`).

```suggestion
| process.runtime.jvm.cpu.utilization            | Recent cpu utilization for the process [2]               | 1       | 1                                         | Asynchronous Gauge         | Double     |               |                       |
| process.runtime.jvm.system.cpu.utilization     | Recent cpu utilization for the whole system [2]          | 1       | 1                                         | Asynchronous Gauge         | Double     |               |                       |
```

bogdandrutu (changes requested):
Please describe the motivation to have these metrics.

trask:
> Please describe the motivation to have these metrics.

These are metrics which are captured by Micrometer today. The Micrometer team would like to move to OpenTelemetry semantic conventions once they are stabilized, and this is part of that effort.

These are metrics which the JVM has been producing for a long time, and the JVM metrics working group believes there is a significant benefit to defining standard metric names for them, to promote interoperability in the Java ecosystem.

trask:
> Please describe the motivation to have these metrics.

@bogdandrutu does my response above address your question/concern?

trask:
@open-telemetry/technical-committee it would be helpful if we could get some guidance on how/if to move this forward, thx!

trask:
@bogdandrutu?

trask:
@open-telemetry/technical-committee is this not the right place to define JVM metrics that we want to emit? should we just document these in the Java repo and create our own "java" schema url? thx

specification/metrics/semantic_conventions/runtime-environment-metrics.md
```diff
@@ -57,21 +57,23 @@ consider, for example pthreads vs green thread implementations.
 
 All JVM metric attributes are required unless otherwise indicated.
```
bogdandrutu, line 58: What does it mean for a set of metrics to be "required"? If I enable jvm metrics I get all? All metrics are available, I just need to enable them? All are enabled by default if I use otel?

specification/metrics/semantic_conventions/runtime-environment-metrics.md
```diff
@@ -57,21 +57,23 @@ consider, for example pthreads vs green thread implementations.
...
+| process.runtime.jvm.threads.count              | Number of executing threads                              | threads | `{threads}`                               | UpDownCounter                                     | Int64      |               |                       |
+| process.runtime.jvm.classes.loaded             | Number of classes currently loaded                       | classes | `{classes}`                               | UpDownCounter                                     | Int64      |               |                       |
+| process.runtime.jvm.classes.unloaded           | Number of classes unloaded since JVM start               | classes | `{classes}`                               | UpDownCounter                                     | Int64      |               |                       |
+| process.runtime.jvm.cpu.utilization            | Recent cpu utilization for the process [2]               | 1       | 1                                         | Asynchronous Gauge                                | Double     |               |                       |
```
bogdandrutu, line 73: I understand this, +1

specification/metrics/semantic_conventions/runtime-environment-metrics.md
```diff
@@ -57,21 +57,23 @@ consider, for example pthreads vs green thread implementations.
...
+| process.runtime.jvm.system.cpu.utilization     | Recent cpu utilization for the whole system [2]          | 1       | 1                                         | Asynchronous Gauge                                | Double     |               |                       |
+| process.runtime.jvm.system.cpu.load_1m         | Average CPU load of the whole system for the last minute | 1       | 1                                         | Asynchronous Gauge                                | Double     |               |                       |
```
bogdandrutu, line 75: I am not sure "system" metrics should be under "process"... something seems off here. We have different section for process and system, see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/process-metrics.md and https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/system-metrics.md

specification/metrics/semantic_conventions/runtime-environment-metrics.md
trask: > What does it mean for a set of metrics to be "required"?

I may not understand your question here, none of these metrics are required.

> If I enable jvm metrics I get all?

The OTel Java SDK will not emit any JVM metrics.

There will be "JVM metrics" instrumentation which users can apply if they want to get these metrics.

It hasn't been decided yet which JVM metrics this instrumentation would emit by default, and which would require an additional opt-in (beyond enabling the instrumentation).

> All metrics are available, I just need to enable them? All are enabled by default if I use otel?

Let me know if these questions aren't answered by the above.

specification/metrics/semantic_conventions/runtime-environment-metrics.md
trask: Would it be ok to put them under `system.*`, and still emit them from the JVM, and let the resource attributes tell the receiver that they are coming from the JVM and not from a "collector"?

bogdandrutu: I think will be ok. I would use the instrumentation scope/name, that will also be different correct? Should we do the same for process CPU then? Still not sure we should do this instrumentation twice but looks like the java folks are really committed to duplicate this work 🤣

bogdandrutu: @trask would ne definitely good to make these metrics that are same with collector an opt in for the JVM metrics package. Since enabling them and using the collector may cause confusion, and trust me that is the standard across all our clients (to use the collector for these) and would be very bad to make these enable by default

trask: > I would use the instrumentation scope/name, that will also be different correct?

👍

> would definitely be good to make these metrics that are same with collector an opt in for the JVM metrics package. Since enabling them and using the collector may cause confusion, and trust me that is the standard across all our clients (to use the collector for these) and would be very bad to make these enable by default

I support this message, that the OTel Collector is the preferred deployment option in the OpenTelemetry ecosystem, and so defaults should be optimized for those users.

I will submit separate spec clarification(s) about these recommendations around emitting `system.*` and `process.*` metrics from a non-"collector".

I think that will address some other metrics we want to emit from the JVM (e.g. #2439), but sadly the particular system and process metrics in this PR pose one additional challenge because:

> These utilizations are not defined as being for the specific interval since last measurement (unlike `system.cpu.utilization`)

Because of this limitation, these metrics are definitely subpar to the existing `system.*` and `process.*` metrics, and we only emit them because they are what's available from the JVM. And so I wonder if we should keep these under the `jvm` metric namespace.

Or do you think we should emit under (e.g.) `system.cpu.utilization` and just call it a best-effort, even though it doesn't match the specification for that metric:

> Difference in system.cpu.time since the last measurement, divided by the elapsed time and number of CPUs

reyang:
I've removed the `Stale` tag.
Given we've got a good number of approvals from Java experts, I want to see if there is any blocker from @bogdandrutu and @jsuereth. If there is no blocker, I think we should move forward.

specification/metrics/semantic_conventions/runtime-environment-metrics.md
trask: btw, discussed this (^^^) with JVM metrics folks, and we are recommending to keep the particular system/process metrics in this PR under the jvm metric namespace since they don't match (and are subpar) to the existing metrics under the system/process metric namespace

reyang:
@jonatan-ivanov would you resolve the merge conflicts and clear the CI? Thanks!

specification/metrics/semantic_conventions/runtime-environment-metrics.md
```diff
@@ -57,21 +57,23 @@ consider, for example pthreads vs green thread implementations.
...
+|                                                |                                                          |         |                                           |                                                   |            | pool          | Name of pool [1]      |
+| process.runtime.jvm.threads.count              | Number of executing threads                              | threads | `{threads}`                               | UpDownCounter                                     | Int64      |               |                       |
+| process.runtime.jvm.classes.loaded             | Number of classes currently loaded                       | classes | `{classes}`                               | UpDownCounter                                     | Int64      |               |                       |
+| process.runtime.jvm.classes.unloaded           | Number of classes unloaded since JVM start               | classes | `{classes}`                               | UpDownCounter                                     | Int64      |               |                       |
```
reyang, line 72: I'm not a Java guy, is the "number of classes unloaded since JVM start" going to go up and \*down\*, or it should be a `Counter`?

reyang: Doesn't have to be resolved in this PR since it was introduced by other PRs.

trask: check out https://github.com/open-telemetry/opentelemetry-specification/pull/2478#discussion_r853173523

> represented as an UpDownCounter because the rate of class loading is not an important factor and the absolute number of classes loaded is an important factor

I'll send a PR to document this reasoning (and to see if there are any objections)

trask: oh wait, we changed it to Counter in #2550

jonatan-ivanov:
@reyang Would you approve the workflow to run?

reyang (approved):
LGTM.

reyang:
@jonatan-ivanov please send a follow up PR to update the changelog (which is an editorial/trivial change).

jonatan-ivanov:
@reyang Thanks! Here is the PR for the changelog entry: https://github.com/open-telemetry/opentelemetry-specification/pull/2584

PR #2297, opened 2022-01-27
Add opentracing.ref_type semantic convention.

carlosalberto:
A missing bit from the recent effort in the OT Shim, essentially a new `opentracing.ref_type` attribute that only accepts either `child_of` or `follows_from` as value.

cc @yurishkuro


jmacd (approved):

yurishkuro (approved):

specification/trace/semantic_conventions/compatibility.md
```diff
@@ -0,0 +1,36 @@
+# Semantic conventions for Compatibility components
```
arminru, line 1: Please
* reference this in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/opentracing.md so it's not overlooked by implementers
* add this new file to the TOC in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/README.md
* add your changes to the changelog

carlosalberto: Done, thanks for the reminder :D

yurishkuro (approved):

PR #2301, opened 2022-01-29
Minor - add link to collect

cijothomas:
Fixes #

## Changes

Related issues #


specification/metrics/api.md
```diff
@@ -254,7 +254,7 @@ asynchronous:
 
 * Asynchronous instruments (e.g. [Asynchronous Gauge](#asynchronous-gauge)) give
   the user a way to register callback function, and the callback function will
-  only be invoked upon collection. For example, a piece of embedded software
+  only be invoked upon [collection](../sdk.md#collect). For example, a piece of embedded software
```
SergeyKanzhelev, line 257: did you mean: 

```suggestion
  only be invoked upon [collection](sdk.md#collect). For example, a piece of embedded software
```

SergeyKanzhelev (approved):

specification/metrics/api.md
```diff
@@ -254,7 +254,7 @@ asynchronous:
 
 * Asynchronous instruments (e.g. [Asynchronous Gauge](#asynchronous-gauge)) give
   the user a way to register callback function, and the callback function will
-  only be invoked upon collection. For example, a piece of embedded software
+  only be invoked upon [collection](sdk.md#collect). For example, a piece of embedded software
```
reyang, line 257: I think the "collection" here is a general concept rather than a concrete method/function.

cijothomas: Thats right.. There is no section which describes the general concept of "collection", so I picked the closest. 
Maybe that's a good addition to the spec, so its easy to explain "when does my async instrument callback gets executed?". Will follow up later, and abandon this PR.

bogdandrutu (approved):

specification/metrics/api.md
cijothomas: I think its still okay to have this. There is no other place which talks about "when does the callbacks from observable instruments gets called".

carlosalberto: What about something along the lines of:

```
only be invoked upon collection (see the [collect](sdk.md#collect) operation in the SDK for reference)
```

carlosalberto (approved):

Issue #2118, opened 2021-11-12
Should Exemplar be enabled by default?


cijothomas:
Originally opened PR to improve wording (https://github.com/open-telemetry/opentelemetry-specification/pull/2105), but was clarified in the PR that, Exemplar was intentionally marked as a enabled-by-default feature.
Opening this issue, to discuss the question:
Should exemplar be enabled by default?

Since the feature has some perf penalty (at the minimum, check the current span from the context, and check if it is sampled in), I'd propose to make this as an opt-in feature. (I do not know if most metric backends support Exemplar features, so it'd be good to make it opt-in, as opposed to opt-out.) It was mentioned in [this comment](https://github.com/open-telemetry/opentelemetry-specification/pull/2105#issuecomment-966740133) that [Prometheus's Exemplar is an opt-in feature](https://prometheus.io/docs/prometheus/latest/feature_flags/#exemplars-storage)

yurishkuro:
+1

reyang:
Discussed during [11/11/2021 Metrics SIG](https://docs.google.com/document/d/1pdvPeKjA8v8w_fGKAN68JjWBmVJtPCpqdi9IZrd6eEo/edit#heading=h.pnuzwm7i7bx). So far the suggestion is to make it OFF by default.

@jonatan-ivanov mentioned during the SIG meeting: _I thinks it's better to be off by default, because the only popular backend that has exemplar is Prometheus, and it needs to be enabled explicitly._

jsuereth:
I'm copying over what I wrote in that PR:

TL;DR: I think focusing Exemplars on Traces and tying its default to trace sampling is the right default, and promotes one of the core tenants to OTel -> promoting correlated telemetry.

I think there's an important distinction here that's possibly missing. By Default, Exemplar sampling is on only if there exists a Sampled Trace. The default requires:

- Users have configured tracing
- Tracing is in context
- Tracing is sampling
- 
Exemplar sampling, by default, is controlled via Trace sampling. The ExemplarFilter/ExemplarResorvoir hooks are only for advanced Exemplar usage.

This is unlike other systems where Exemplar sampling may or may not correlate with the existence of Traces. I'd expect a user of just the otel metrics SDK to never see an exemplar (by default). It's only once they've also instrumented traces that they see exemplars be generated.


> I have a slightly connected question: should OTel Metrics support any Exemplar source or only OTel Tracing?

[...]

Issue #2072, opened 2021-10-28
Clarification on MetricReader registered on MeterProvider


legendecas:
**What are you trying to achieve?**

Clarify the object ownerships between metric reader and meter provider.

In the example (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#view), it is said the metric reader is constructed in place and then it is registered to the meter provider:

```python
meter_provider
    .add_view("*") # a wildcard view that matches everything
    .add_metric_reader(PeriodicExportingMetricReader(ConsoleExporter()))
```

Also, it is stated that metric reader has a `Collect` operation that triggers asynchronous instrument callbacks:

> Metric Reader Operations:
> Collect: Collects the metrics from the SDK. If there are asynchronous Instruments involved, their callback functions will be triggered.
> https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#metricreader-operations

As far as I could tell, all the instruments are owned by the meter provider, and thus the `collect` should consult the meter provider to perform async instrument callbacks. 

In this sense, it seems that the reader can be registered to multiple meter provider and the `metric_reader.collect` should perform collection on all registered meter provider:

```python
metric_reader = PeriodicExportingMetricReader(ConsoleExporter())
meter_provider
    .add_view("*")
    .add_metric_reader(metric_reader)

meter_provider_b
    .add_view("*")
    .add_metric_reader(metric_reader)
```

The problem here is that the SDK API interface suggested that we are registering the reader to the meter provider, however, the `metric_reader.collect` reverse the control ownership and the meter provider shall be registered on the metric reader and thus it is possible to call `metric_reader.collect` to perform a collection on all registered async instrument callbacks.

This sounds like a doubly linked ownership:
```
        meter_provider 
  shutdown/    |  ↑
  force_flush  |  | collect
               ↓  |
        metric reader
```

**What did you expect to see?**

The ownership should not be bidirectional.

reyang:
> In this sense, it seems that the reader can be registered to multiple meter provider and the `metric_reader.collect` should perform collection on all registered meter provider:

Currently the spec doesn't specify if the same MetricReader instance can be shared across multiple MeterProvider instances. In practice, different languages use different approach:
1. Java takes a factory instance instance, so it is impossible to pass in the same MetricReader instance to multiple MeterProvider instances.
[...]

PR #2304, opened 2022-01-31
Move metrics SDK spec to Mixed

reyang:
Supersedes #2150.

## Changes

* Mark Metrics SDK specification to Mixed, with Exemplar as Feature-freeze while other parts as Stable.

Related PR #2270 is touching the API spec post Stable release, so it might not be a blocker.

### DIFF
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -1,6 +1,6 @@
 # Metrics SDK
 
-**Status**: [Feature-freeze](../document-status.md)
+**Status**: [Stable](../document-status.md)
 
 <details>
 <summary>Table of Contents</summary>
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -182,10 +182,10 @@ are the inputs:
     outputs metric points that use aggregation temporality (e.g. Histogram,
     Sum), the SDK SHOULD handle the aggregation temporality based on the
     temporality of each [MetricReader](#metricreader) instance.
-  * **Status**: [Feature-freeze](../document-status.md) - the `exemplar_reservoir`
-    (optional) to use for storing exemplars. This should be a factory or
-    callback similar to aggregation which allows different reservoirs to be
-    chosen by the aggregation.
+  * **Status**: [Feature-freeze](../document-status.md) - the
+    `exemplar_reservoir` (optional) to use for storing exemplars. This should be
+    a factory or callback similar to aggregation which allows different
+    reservoirs to be chosen by the aggregation.
 
 In order to avoid conflicts, views which specify a name SHOULD have an
 instrument selector that selects at most one instrument. For the registration
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -182,7 +182,7 @@ are the inputs:
     outputs metric points that use aggregation temporality (e.g. Histogram,
     Sum), the SDK SHOULD handle the aggregation temporality based on the
     temporality of each [MetricReader](#metricreader) instance.
-  * **Status**: [Experimental](../document-status.md) - the `exemplar_reservoir`
+  * **Status**: [Feature-freeze](../document-status.md) - the `exemplar_reservoir`
     (optional) to use for storing exemplars. This should be a factory or
     callback similar to aggregation which allows different reservoirs to be
     chosen by the aggregation.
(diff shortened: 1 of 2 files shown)


jmacd (approved):

specification/metrics/sdk.md
```diff
@@ -416,13 +419,17 @@ instruments that record negative measurements, e.g. `UpDownCounter` or `Observab
...
+**Status**: [Stable](../document-status.md)
+
 Attributes which belong to Metrics are exempt from the
 [common rules of attribute limits](../common/common.md#attribute-limits) at this
 time. Attribute truncation or deletion could affect identity of metric time
 series and the topic requires further analysis.
 
 ## Exemplar
 
+**Status**: [Feature-freeze](../document-status.md)
```
jsuereth, line 431: Can I get a TL;DR: on why this is feature-freeze and what the criteria will be to mark this as stable?

specification/metrics/sdk.md
```diff
@@ -587,6 +594,8 @@ measurements using the equivalent of the following naive algorithm:
 
 ## MetricReader
 
+**Status**: [Stable](../document-status.md)
```
jsuereth, line 597: I want to call out that I'm not sure the language in this section of the SDK has really been updated to the extent we'd like it.

Specifically, it's ambiguous whether you need a physical `Collect` method.  We have talked about the java implementation of this [MetricReader](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/MetricReader.java) where a `collect` method is passed in by [MetricReaderFactory](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/MetricReaderFactory.java).

Given the number of implicit gotchas in this spec, I really think we need to go look at implementations of MetricReader and reword/reflesh out this section before marking it as stable.  I think feature-wise it's fine, but I don't think the way it's laid out is clear/obvious.

specification/metrics/sdk.md
reyang: I think the main reason is that many APM vendors need more time to form a better understanding, while we don't have to block the entire SDK spec https://github.com/open-telemetry/opentelemetry-specification/issues/2118#issuecomment-1015587247.

specification/metrics/sdk.md
reyang: Is there a spec issue tracking this? I can help to prepare a clarification PR.

cijothomas (approved):

specification/metrics/sdk.md
jsuereth: https://github.com/open-telemetry/opentelemetry-specification/issues/2072 best describes the fundamental issue.

@legendecas calls out that the control is inverted from expectations in how it's specified.  In Java, we used his description of how control should flow, but the specification wasn't updated to (explicitly) allow either control flow option.

Given the number of times folks have misinterpreted the discussion (or context is lost) I'd prefer us make this explicit before marking it stable.

reyang:
We should revisit this PR once #2317 is merged.

alolita:
@reyang @jsuereth Can this PR be merged given #2317 is merged.

specification/metrics/sdk.md
jmacd: How do we move forward with this issue and #2072?

jmacd: See the discussion in https://github.com/open-telemetry/opentelemetry-specification/issues/2072#issuecomment-1054873664

jmacd:
Note that https://github.com/open-telemetry/opentelemetry-specification/issues/2072 appears to be still blocking this PR, unfortunately. In https://github.com/open-telemetry/opentelemetry-specification/issues/2072#issuecomment-1054873664 I have a proposal that would also let us close #2232, I think. Please give #2072 and #2232 your attention, especially @mateuszrzeszutek, @legendecas, @jack-berg, @jsuereth, @reyang thanks.

jack-berg:
I think that having some sort of resolution of #2314 is important for stabilization. My read is that there's consensus that delta backends actually want cumulative non-monotonic sums instead of delta non-monotonic sums for sync / async up down counter instruments. Stabilizing without some resolution would create trouble for these backends without benefiting anyone (as far as I can tell).

cijothomas (approved):
re-approval after more recent PRs are merged.

carlosalberto (approved):

alanwest (approved):

mtwo (approved):

MrAlias (approved):

reyang:
This PR is blocked by a CI issue which is being addressed in #2426.

pirgeo (approved):

mtwo:
Following up from the specification call today: this is ready to merge

Issue #2065, opened 2021-10-27
Set preferred aggregation temporality for the periodic exporting metric reader via environment variable


pirgeo:
**What are you trying to achieve?**

I want to be able to set the preferred aggregation temporality for an periodic exporting metric reader via an environment variable, e.g. `OTEL_METRIC_EXPORT_PREFERRED_TEMPORALITY`. If set, This would take precedence over the preferred temporality specified in the `MetricReader`/`MetricExporter` (see [temporality override rules](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#temporality-override-rules)).

**Additional context.**

Similar to #2021/#2038, but for aggregation temporality.
Also affects the precedence created in #2031/#2032.

jack-berg:
Just to clarify, this is different than the ability to configure temporality for the OTLP exporter? I.e. [OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk_exporters/otlp.md)

pirgeo:
I was thinking about it in more broad terms, i.e. every exporter that supports that environment variable can read it and set the preferred temporality. The mechanism would be the same, but the scope would be a bit bigger. I guess setting it for each exporter individually is also a way to go, and gives you more granular control. These two paths do not interfere with each other, I'd use the more specific (`OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY`) over the more general variable (`OTEL_METRIC_EXPORT_PREFERRED_TEMPORALITY`).

I'd envision this as a more general guidance that exporters that don't want to specify their own env var could use.

PR #2314, opened 2022-02-04
Clarify preferred aggregation temporality; give Views a selection strategy

jmacd:
## Changes

Clarify "preferred aggregation temporality" and introduce a mechanism to configure, via View, whether the preferred aggregation temporality is honored.

Sum, Histogram, and ExpontentialHistogram data points add a new "strategy" allowing them to honor or override the Reader/Exporter's preference. UpDownCounter instruments (sync and async), in particular, will override the exporter preference and choose cumulative by default. A view configuration can change this.

Note: a hypothetical strategy to override the preference and choose delta is not introduced, though it conceptually makes sense. This could be added later if anyone finds a use.

~I propose the name "Stateless" as the name of a preference that requires no long term state, which means to use Delta temporality for synchronous instruments and leave asynchronous instruments in their natural Cumulative form.~ In a future PR, I will propose adding a new preference called Stateless, described as:

```
- **Stateless**: The Metric Exporter prefers Delta temporality for
synchronous instruments and Cumulative temporality for asynchronous
instruments to avoid long-term memory.
```

Part of #2065


specification/metrics/sdk.md
```diff
@@ -622,25 +623,66 @@ The SDK SHOULD provide a way to allow `MetricReader` to respond to
...
+A configured `MetricReader` instance MUST support reading the
+preferred aggregation temporality for synchronous instruments.
+
+An optional aggregation temporality preference "Delta" MAY be
+supported by implementations, making it possible for users to convert
+data points from asynchronous `UpDownCounter` and `Counter`
+instruments into Delta aggregation temporality inside their SDK.
+
+SDK support for converting Cumulative to Delta aggregation is optional
+because it is not considered widely useful; in metrics data formats
```
pirgeo, line 673: Not sure if this is true, but I probably just haven't worked with enough backends yet to have a strong opinion here. To me it feels as if there is value in providing that transformation, especially since we have the possibility to do so (at a cost, yes, but that's ultimately the users decision). I am not sure I like the notion of asking users to transform things that are not gauges into gauges just because we have always done it that way, but maybe others can chime in and share their thoughts on this.

specification/metrics/sdk.md
```diff
@@ -622,25 +623,66 @@ The SDK SHOULD provide a way to allow `MetricReader` to respond to
...
+that emphasize Delta aggregation temporality such as StatsD, it is
+conventional to convert cumulative values into `Gauge` data points,
+for example, and this is not typically encountered for `Histogram`
+measurements.
+
+Users that wish to convert Cumulative into Delta aggregation
```
pirgeo, line 679: Just a side thought: does this work in environments with multiple instances of the collector?

pirgeo (commented):
Thanks for clarifying this. I generally like it, I am just not entirely bought into the idea that cumulative->delta conversion is something that most people would not want (see my comment there)

specification/metrics/sdk.md
```diff
@@ -622,25 +623,66 @@ The SDK SHOULD provide a way to allow `MetricReader` to respond to
...
+  This preference is so named because it avoids long-term Cumulative 
+  state in the metrics SDK.
+
+If the preferred temporality is explicitly specified then the SDK
+SHOULD respect that, otherwise use Cumulative.
+
+A configured `MetricReader` instance MUST support reading the
+preferred aggregation temporality for synchronous instruments.
+
+An optional aggregation temporality preference "Delta" MAY be
```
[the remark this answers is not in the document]
jsuereth: I think this is where some disagreement comes in.

Java currently supports Cumulative->Delta conversion in the SDK (only for async instruments + DELTA exporters).   I'm not sure what .NET does here, but I think there's disagreement on whether we require this, or even *allow* it.


Practically:

1. CUMULATIVE->DELTA transition requires additional memory to record past history.
2. CUMULATIVE->DELTA works on our built-in aggregations but may be serious strain on custom Aggregation once we allow it.
3. We have backends in this community which ONLY support DELTA metrics, so not having CUMULATIVE->DELTA in the SDK limits metrics available to some vendors within the community.

For me, the third point swayed me to make sure CUMULATIVE->DELTA was first class.

specification/metrics/sdk.md
```diff
@@ -622,25 +623,66 @@ The SDK SHOULD provide a way to allow `MetricReader` to respond to
...
+supported by implementations, making it possible for users to convert
+data points from asynchronous `UpDownCounter` and `Counter`
+instruments into Delta aggregation temporality inside their SDK.
+
+SDK support for converting Cumulative to Delta aggregation is optional
+because it is not considered widely useful; in metrics data formats
+that emphasize Delta aggregation temporality such as StatsD, it is
+conventional to convert cumulative values into `Gauge` data points,
```
jsuereth, line 675: For StatsD -

- I think we should be converting "timer" metrics into Histograms (but this is a collector thing not an SDK thing)
- StatsD counter ~= synchronous UpDownCounter in OTEL.
- StatsD sets ~= DELTA Counter in OTEL (Is this what you would convert to Gauge?)
- StatsD gauge ~= OTEL Gauge
- StatsD "sampled" counter =>  Do we have an equivalent in OTEL?  

Specifically, I think a StatsD receiver in OTEL *can* produce a better model than just converting most things to `Gauge` data points (even if that's conventional).

specification/metrics/sdk.md
jsuereth: You need to guarantee the "single-writer-principle" i.e. there can only be ONE collector involved in writing any particular timeseries.

Another way to phrase this -

It works in multi-collector instances as long as EVERY metric with the same attributes goes to the same collector.

specification/metrics/sdk.md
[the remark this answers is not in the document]
cijothomas: > Java currently supports Cumulative->Delta conversion in the SDK (only for async instruments + DELTA exporters). I'm not sure what .NET does here

Same in .NET.

specification/metrics/sdk.md
jmacd: You're not contradicting my proposal. StatsD has no concept of a cumulative Counter or UpDownCounter, so if you were reporting those kinds of numbers through StatsD you would've chosen Gauge already, or you would've converted them to Delta temporality yourself (and used StastD Counters, which strictly speaking do allow negative counts).

jmacd:
In https://github.com/open-telemetry/opentelemetry-go/pull/2350, I wrote the PR "removes a bunch of complexity that the Metrics SIG group decided not to specify for the default SDK." 

Maintaining Cumulative-to-Delta conversion in the SDK has a definite cost and as far as I know is not useful to very many. We are debating a scenario where a user has a non-OTLP protocol that supports Delta measurements and Gauge measurements but does not support Cumulative measurements. The precedent we know about (StatsD) would use Gauge for these measurements. I polled the NewRelic engineers during that conversation, who have a protocol that is strictly based on Deltas, and they confirmed that they would convert Gauges to Cumulatives. @jack-berg could you confirm?

The test question we should answer is if you have a number like Heap Usage, which is modeled as an OTel UpDownCounter, and you have a protocol that does Gauge and Delta, would you really like to see your Heap Usage turned into a Delta? I believe the answer is no.

We had this discussion in November before the current Spec was written, before I removed Subtract() support from the OTel-Go implementation. I had implemented Subtract() for Sum points but not for Histogram points, and the total complexity for this (nearly useless, IMO) feature did not warrant keeping it.

The current specification is very unclear, and to my reading never required an implementation to implement Cumulative-to-Delta translation:

```
[OpenTelemetry SDK](../overview.md#sdk)
authors MAY choose the best idiomatic design for their language:

* Whether to treat the temporality settings as recommendation or enforcement.
  For example, if the temporality is set to Delta, would the SDK want to perform
  Cumulative->Delta conversion for an [Asynchronous
  Counter](./api.md#asynchronous-counter), or downgrade it to a
  [Gauge](./datamodel.md#gauge), or keep consuming it as Cumulative due to the
  consideration of [memory
  efficiency](./supplementary-guidelines.md#memory-management)?
* Refer to the [supplementary
  guidelines](./supplementary-guidelines.md#aggregation-temporality), which have
  more context and suggestions.
```

I don't believe I'm actually changing the specification here, except to introduce the term "Stateless" to describe a preference for no changes of temporality, which is neither always Cumulative nor always Delta.

specification/metrics/sdk.md
jmacd: @pirgeo I'm stuck wondering if anyone can document a need for converting cumulative to delta in the SDK; it has to be a protocol other than OTLP and there has to be a good reason to change cumulative into delta, not just a story about symmetry. I would argue that information is lost from cumulative to delta, since you lose information about the total, and that's why no one actually wants to do this.

specification/metrics/sdk.md
jmacd: I would challenge that there's really no value in doing this.

specification/metrics/sdk.md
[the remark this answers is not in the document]
jmacd: I'm asking everyone: Why support this? The current specification is written not to require this and we debated it in November. Taking away this conversion removed a lot of SDK complexity, and just because we "can" and the meaning is well defined doesn't mean we should.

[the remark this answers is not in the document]
jmacd: I do not agree to your point 3 @jsuereth. We have polled the vendors with delta-only protocols and asked how they would treat cumulative values originating from asynchronous instruments. My understanding is that protocols that only accept Delta are design to accept Gauges for cumulative values. This is surely the case with Datadog. No one expects you to convert your memory usage into a rate before using Datadog, even though their primary input is StatsD.

jmacd:
No one has challenged the idea that "Stateless" be a good name for the preference to not modify temporality in an SDK.

There are a number of reviewers who seem to think Cumulative-to-Delta translation is needed for backends that support only Delta temporality. I emphatically do not believe this is true. In any case, if this were true it might lead to a specification change that requires SDKs support Cumulative-to-Delta, which is clearly not included in the current specification. In that case "Delta" is the natural name for a preference to convert all data into delta temporality.

You might point to the existence of the OTC cumulativetodelta processor as proof that someone wants this. The use that we know of for that code was a Prometheus-to-Cloudwatch pipeline, where the Cumulative values are first turned into Deltas and then turned into Rates. The point here is that Deltas were not the end goal of that transformation, and CloudWatch does not directly support Deltas. You can convert Cumulative to Delta to Rate data for backends that support only Gauges, but here we are discussing a very esoteric feature that belongs in a collector processor, not in the SDK.

specification/metrics/api.md
```diff
@@ -404,7 +404,7 @@ observed. [OpenTelemetry API](../overview.md#api) authors SHOULD define whether
 this callback function needs to be reentrant safe / thread safe or not.
 
 Note: Unlike [Counter.Add()](#add) which takes the increment/delta value, the
-callback function reports the absolute value of the counter. To determine the
+callback function reports the cumulative value of the counter. To determine the
 reported rate the counter is changing, the difference between successive
 measurements is used.
```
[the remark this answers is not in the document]
jmacd: Yes.

specification/metrics/sdk.md
```diff
@@ -622,25 +623,66 @@ The SDK SHOULD provide a way to allow `MetricReader` to respond to
...
+by observations that are cumulative values to begin with.  For these
+points, a change of aggregation temporality implies conversion from
+Cumulative into Delta aggregation temporality.
+
+Because of these differences, synchronous and asynchronous instruments
+are given separate treatment.  When configuring the preferred
+aggregation temporality, the implementation MUST provide at least the
+following named preferences:
+
+- "Cumulative": All data points that define the concept are exported
```
cijothomas, line 649: Before this PR, a reader could have the following options for preferred temporality:
1. Delta
    In this mode, SDK always output metrics with delta temporality. This means doing cumulative -> delta conversion, where needed. (i.e async counter, async updowncounter)
2. Cumulative
    In this mode, SDK always output metrics with cumulative temporality. This means doing delta -> cumulative conversion, where needed. (i.e counter, histogram)

This PR proposes to add a new option
3. "Stateless"
     This mode is similar to Delta, except it doesn't do cumulative->delta conversion and hence exports async counter/updowncounter value as cumulative.

This PR also makes "Delta" an optional thing, and strongly advices against it.

Is my above understanding correct?

jmacd: Yes, your understanding is correct. 
However, I believe "Delta" was already optional, from the text quoted above.

My background belief here is that the advantage of choosing Delta to begin with is that it requires no long-term state in the SDK. There's no good reason to choose Delta _in a situation where it implies additional state and cost_.

My goal was to introduce "Stateless" as a preference meaning no conversions, no state required.

jmacd:
I put this PR together mainly because:

(1) I thought we had decided against requiring Cumulative-to-Delta. I we want to change that to a requirement, we should open a new PR.
(2) I think the "stateless" preference should be called out as it's the one that requires no long term memory

There is another potential preference that my employer finds appealing, it's like Stateless except for UpDownCounter. We prefer UpDownCounter in cumulative form because otherwise we would have to read old data or maintain checkpoints to know the current value of the metric, and the current value of an UpDownCounter metric is semantically useful.

This preference could be named "StatelessMonotonic", since it is stateless except for the one synchronous non-monotonic instrument. Why prefer to use Delta for Counter and Histogram, but not UpDownCounter, you ask? The totals of a Counter or Histogram are meaningful totals, but because these counts are monotonic there is little-to-no semantic loss in the use of Deltas.

jack-berg:
> Maintaining Cumulative-to-Delta conversion in the SDK has a definite cost and as far as I know is not useful to very many. We are debating a scenario where a user has a non-OTLP protocol that supports Delta measurements and Gauge measurements but does not support Cumulative measurements. The precedent we know about (StatsD) would use Gauge for these measurements. I polled the NewRelic engineers during that conversation, who have a protocol that is strictly based on Deltas, and they confirmed that they would convert Cumulatives to Gauges. @jack-berg could you confirm?
The test question we should answer is if you have a number like Heap Usage, which is modeled as an OTel UpDownCounter, and you have a protocol that does Gauge and Delta, would you really like to see your Heap Usage turned into a Delta? I believe the answer is no.

This PR's timing has really convenient timing because I've actually just been looking into this exact scenario more closely. As mentioned, our metrics system uses delta, and we do convert both cumulative non-monotonic sum data and delta non-monotonic sum to gauges. And as you suggest, having heap usage represented by asynchronous up down counter turned into a delta is problematic, since it means that in order to determine the current heap usage at a point in time Tn in time you have to sum up all the delta values from T0 -> Tn. And this doesn't work with gauge data because when gauges undergo temporal aggregation, the last value wins semantics means that the deltas required to compute current heap usage are dropped.

I agree with this change 👍.

specification/metrics/sdk.md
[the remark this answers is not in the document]
jack-berg: As articulated [here](https://github.com/open-telemetry/opentelemetry-specification/pull/2314#issuecomment-1030378613), converting async counter and async up down counter instruments to delta before export makes things harder for New Relic.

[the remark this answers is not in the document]
jsuereth: Interesting, if we can get rid of cumulative->delta then let's do so.  I had assumed, given new relics but reports and feature requests, this was a necessary feature.

specification/metrics/api.md
```diff
@@ -404,9 +404,9 @@ observed. [OpenTelemetry API](../overview.md#api) authors SHOULD define whether
 this callback function needs to be reentrant safe / thread safe or not.
 
 Note: Unlike [Counter.Add()](#add) which takes the increment/delta value, the
-callback function reports the absolute value of the counter. To determine the
-reported rate the counter is changing, the difference between successive
-measurements is used.
+callback function reports the cumulative value of the counter. To determine the
```
pirgeo, line 407:
```suggestion
callback function reports the current value of the counter. To determine the
```

specification/metrics/api.md
```diff
@@ -893,7 +893,7 @@ observed. [OpenTelemetry API](../overview.md#api) authors SHOULD define whether
 this callback function needs to be reentrant safe / thread safe or not.
 
 Note: Unlike [UpDownCounter.Add()](#add-1) which takes the increment/delta value,
-the callback function reports the absolute value of the Asynchronous
+the callback function reports the cumulative value of the Asynchronous
```
pirgeo, line 896:
```suggestion
the callback function reports the current value of the Asynchronous
```

specification/metrics/sdk.md
```diff
@@ -622,25 +623,66 @@ The SDK SHOULD provide a way to allow `MetricReader` to respond to
...
+registration, etc.) time.  This preference gives the user control over
+the amount of long-term memory dedicated to reading metrics.
+
+The synchronous instruments generally define data points having
+aggregation temporality (e.g., `Sum`, `Histogram`).  For these points,
+when a `MetricReader` is configured with Cumulative aggregation
+temporality, there is an implied long-term memory cost.
+
+The asynchronous `Counter` and `UpDownCounter` instruments are defined
+by observations that are cumulative values to begin with.  For these
```
pirgeo, line 640: Are they cumulative just because they are additive?

specification/metrics/sdk.md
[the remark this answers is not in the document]
pirgeo: I think I agree for the async UpDownCounter, but I am on the fence for the async Counter

specification/metrics/sdk.md
pirgeo: I am not sure I follow this. If you have delta counts, its very easy to reconstruct the total count for a specific time frame, or am I missing something?

pirgeo (commented):
After considering this in more detail, here are some more thoughts:
I am not sure if we can lump together async UpDownCounter and async Counter instruments in this PR. Lets start with the UpDownCounter:

I think what is throwing me off is that the async UpDownCounter's *current* value is called "cumulative". It is the value of something that is observed at the time the callback is called. The "cumulative" wording seems to refer to the fact that the changes are made by adding and subtracting values from the underlying value, which is later observed. The UpDownCounter itself does not have a temporality, it just reports the current value of the underlying object(s), just like a gauge does. Are values always "cumulative" just because they are created and updated in an additive fashion? To me, it does not seem to have anything to do with the export temporality.

With that being said, the UpDownCounter DOES feel very similar to a gauge. There are very fine nuances between exporting these as a gauge vs. exporting them as a non-monotonic cumulative sum (to put it into OTLP terms, and to confirm my understanding that this is what UpDownCounters would be exported as). I understand that most vendors do not care about the possibility of adding together the heap size of two processes (something using a Sum would provide out of the box) or provide ways of adding two gauges together anyway. To conclude, and to actually bring this thread back to the question about Cumulative->Delta conversion, I also agree that this conversion is probably not necessary for UpDownCounters. UpDownCounters (especially async ones) are conceptually so close to Gauges that I agree that only exporting the total value recorded by the callback (the *cumulative*/current value) makes sense.

For the Async Counter however, this is a different story, and I think we need to focus our attention on that. Consider something like CPU time or page faults. If I want Deltas in my backend, I'd want the delta to the previous value, and not the total value on every export. In this case I definitely want to be able to add CPU times together to aggregate them. In this case language like "CPU time since the last export" makes sense again (which is admittedly not entirely true for Heap size or queue length).
Therefore, I am a strong proponent of keeping the Delta conversion at least for Async Counters. Dropping support for it will render Async counter data mostly useless for delta-only backends (exporting them as gauges without calculating a delta will lead to an ever-increasing metric).

jack-berg:
FYI, @pirgeo makes a really good point about this change improving the behavior of async up down counter but making the experience worse for async counters. Still interested in this topic - taking some more time to think carefully about the various options and the repercussions.

jmacd:
Thank you @pirgeo and @jack-berg. I agree with the points raised and am willing to reconsider my position.
It sounds like three distinct preferences are real and potentially useful:

1. Cumulative: All cumulative
2. StatelessMonotonic: Delta Histogram, Delta Counter, Cumulative Async Counter, Cumulative UpDownCounter
3. DeltaMonotonic: Delta Histogram, Delta Counter, Cumulative UpDownCounter, Delta Async Counter, Cumulative UpDownCounter
 
It seems like the following preferences are less useful, potentially not needed:

4. Delta: All delta
5. FullyStateless: Delta UpDownCounter, ...

jack-berg:
@jmacd I think there might be some typos in three preferences you mention: StatelessMonotonic doesn't mention Async UpDownCounter, and DeltaMonotonic mentions CumulativeUpDownCounter twice. Here's my take on the most useful aggregation temporality semantics for a delta backend. It would be super useful if there was an aggregation temporality strategy that would produce this.

| Instrument | Export Temporality + Aggregation | Use Case | Explanation |
| ---------- | ---------- | ---------- | ---------- |
| Histogram | Delta Histogram | HTTP Server Duration | Delta histograms include use min / max, minimize memory usage in SDK |
| Sync Counter | Delta Sum | Count of processed bytes | Useful to analyze diff of bytes processed, minimize memory usage in SDK |
| Async Counter | Delta Sum | Process CPU Time | Useful to know diff in CPU time, not cumulative CPU time  |
| Sync UpDownCounter | Cumulative Sum | Items in a queue | Useful to know size of queue, not change in size of queue |
| Async UpDownCounter | Cumulative Sum | Process Memory Usage | Useful to analyze current memory usage, not change in memory usage |
| Gauge | Gauge | Temperature | Temporality not applicable |

In light of this conversation and the push to stabilize the metric SDK, I'm worried that the SDK's language around [MetricReader](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#metricreader) and [MetricExporter](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#metricexporter) may lead to SDK design that makes it difficult to allow aggregation temporality strategies like we're discussing here. 

The metric SDK spec says that MetricReader and MetricExporter have methods to access their preferred aggregation temporality. This lends itself to SDK design like java's in which [MetricExporter#getPreferredTemporality()](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/MetricExporter.java#L25-L28) returns an enum type called [AggregationTemporality](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/AggregationTemporality.java) whose values are DELTA and CUMULATIVE, and **which could not be extended to include the other strategies**. 

If we want to support these types of strategies I think it's important to adjust the SDK to rename "preferred temporality" for MetricReader and MetricExporter to something like "aggregation temporality strategy". Aggregation temporality strategies at SDK launch might only include "ALWAYS_DELTA" and "ALWAYS_CUMULATIVE", but could be extended with additional strategies later.

jmacd:
@jack-berg Thank you. Yes, I left a lot of bytes out of my summary, and you filled it in nicely. Speaking as a vendor now, we would be glad if users configured the temporality preference / strategy that you outlined.

alanwest:
> The test question we should answer is if you have a number like Heap Usage, which is modeled as an OTel UpDownCounter, and you have a protocol that does Gauge and Delta, would you really like to see your Heap Usage turned into a Delta? I believe the answer is no.

I agree that a delta non-monotonic sum is not useful here. I believe this is true irrespective of sync vs. async. That is, I can't think of an UpDownCounter use case where I'd ever want things turned into a delta.

Maybe a slightly different spin on things - instead of expanding on the notion on preferred temporality, what if UpDownCounters were exempt from temporality much like gauges. From the perspective of the OTLP data model, measurements from UpDownCounters would always result in a non-monotonic sum with temporality unspecified (if unspecified is not allowed then I guess just cumulative).

jmacd:
Yesterday, I mentioned the idea of a "6-way preference" meaning to allow the SDK to be configured with a default preference for every instrument. This was a slight mistake, it would be a "5-way preference" because one of six does not have a temporality for its default aggregation (i.e., Gauge). I've updated the text and proposed two names, "AllCumulative" and "DeltaPreferred" as straw-person proposals.

Note there is a connection with the discussion in #2317 and
https://github.com/open-telemetry/opentelemetry-specification/issues/2232#issuecomment-1041960737 about whether the user has a way to stop exporting metrics. In that setting, it seems there is a parallel concept we may call "memory preference" that matches the temporality preference. Cumulative preferences lead to memory preferences, whereas Delta preferences lead to no-memory preferences, but they are strictly independent, I think.

Note that the Stream Manipulations section in the data model is marked experimental. There's no clear instruction for how to compute Cumulative-to-Delta in the data model -- I'm not say this is difficult to find a solution to but we haven't specified it. In the context of #2317 and #2232, the question is what happens when a Counter callback stops reporting a stream that it previously reported. Do output a stale point? Do you reset the stream when it restarts? Do you remember the old value until it restarts--and output zero deltas meanwhile?

specification/metrics/api.md
jmacd: Fixed.

specification/metrics/sdk.md
jmacd: They are specified that way, at least to me. The word "Current" implies "Cumulative" for asynchronous instruments.

specification/metrics/sdk.md
```diff
@@ -624,25 +625,55 @@ The SDK SHOULD provide a way to allow `MetricReader` to respond to
...
+implementation MUST provide a mechanism that supports configuring the
+exporter's preferred temporality on the basis of instrument kind for
+the five instruments `Counter`, `Asynchronous Counter`,
+`UpDownCounter`, `Asynchronous UpDownCounter`, and `Histogram`.
+
+A configured `MetricReader` instance MUST expose data in the preferred
+aggregation temporality for points deriving from the five instruments
+that define aggregation temporality.
+
+Two well-known preferences are named, offering a simple way to express
```
jsuereth, line 658: Would it make sense to JUST provide these two options instead of per-instrument?

specification/metrics/sdk.md
```diff
@@ -624,25 +625,55 @@ The SDK SHOULD provide a way to allow `MetricReader` to respond to
...
+exporter preferences through environment variables and in
+configuration files.  The following named preferences SHOULD be
+supported:
+
+- **AllCumulative**: All data points with aggregation temporality are
+  exported using Cumulative aggregation temporality. 
+- **DeltaPreferred**: Data points from `Counter`, `Asynchronous
+  Counter`, and `Histogram` instruments use Delta aggregation
+  temporality, whereas whereas data points from `UpDownCounter` and
```
alanwest, line 667:
```suggestion
  temporality, whereas data points from `UpDownCounter` and
```

specification/metrics/sdk.md
jmacd: I thought it might be easier to specify with a general hook, but I'm easy to convince on this.

specification/metrics/sdk.md
```diff
@@ -624,25 +625,55 @@ The SDK SHOULD provide a way to allow `MetricReader` to respond to
...
+  `Asynchronous UpDownCounter` are exported with Cumualtive
```
alanwest: This design does not quite feel right. I don't think there's anywhere else in the spec where instruments are directly tied to the notion of temporality. Currently, they're only related to temporality via the type of aggregation they're configured with. I think coupling instruments to temporality may lead to problems.

Can we leave preferred temporality alone and solve this with a new aggregation? May a specialized form of the Sum aggregation which always results in a cumulative aggregation: CumulativeSum?

jack-berg:
In the 2/22/2022 Spec SIG meeting I mentioned an additional alternative to what is proposed in this PR and @alanwest's [proposal](https://github.com/open-telemetry/opentelemetry-specification/pull/2314#discussion_r812089856) for an additional aggregation:

Suppose we add a configuration parameter to the [Sum Aggregation](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#sum-aggregation) called "StrictTemporalityMode" (or something to that effect). The function of this parameter would be to determine whether or not the output temporality would strictly adhere to the the preferred temporality of the MetricReader / MetricExporter. It would default to `false`, in which case all instruments aggregated with Sum would adhere to the preferred temporality, **except** Async UpDownCounter and UpDownCounter which are more natural in cumulative form. Async UpDownCounter and UpDownCounter would always export as cumulative unless `StrictTemporalityMode=true`, which causes all them to strictly adhere to the preferred temporality. 

This would allow us to stick with the existing specification for preferred temporality, while accommodating the preference for cumulative non-monotonic sums amongst even delta backends, and while also supporting the possibility of exporting delta non-monotonic sums.

jmacd:
@jack-berg and @alanwest 

I like the idea expressed in your comments above. I explicitly want to factor in a kind of preference mentioned above and in our meetings for "stateless" aggregation, meaning whichever temporality requires no long term memory. I still see three preferences, "cumulative", "delta", and "stateless".

Then, there's the idea of the Aggregation having a temporality strategy, particularly for Sum but also applicable for histograms, the ones you've outlined are:

- Sum{strategy="preferred"}: defer to the preference of the exporter / reader
- Sum{strategy="cumulative"}: convert to cumulative temporality

The following table explains the scenario described by @jack-berg in these terms:

| Instrument          | Default Aggregation             | Stateless Temporality | Selected Temporality, Cumulative Preference | Selected Temporality, Delta Preference | Selected Temporality, Stateless Preference |   |
|---------------------|---------------------------------|-----------------------|---------------------------------------------|----------------------------------------|--------------------------------------------|---|
| Counter             | Sum{strategy="preferred"}       | Delta                 | Cumulative                                  | Delta                                  | Delta                                      |   |
| UpDownCounter       | Sum{strategy="cumulative"}      | Delta                 | Cumulative                                  | Cumulative                             | Cumulative                                 |   |
| Histogram           | Histogram{strategy="preferred"} | Delta                 | Cumulative                                  | Delta                                  | Delta                                      |   |
| Async Counter       | Sum{strategy="preferred"}       | Cumulative            | Cumulative                                  | Delta                                  | Cumulative                                 |   |
| Async UpDownCounter | Sum{strategy="cumulative"}      | Cumulative            | Cumulative                                  | Cumulative                             | Cumulative                                 |   |

jmacd:
@jack-berg in https://github.com/open-telemetry/opentelemetry-specification/pull/2314#issuecomment-1039598090

specifically 

> This lends itself to SDK design like java's in which [MetricExporter#getPreferredTemporality()](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/MetricExporter.java#L25-L28) returns an enum type called [AggregationTemporality](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/AggregationTemporality.java) whose values are DELTA and CUMULATIVE, and which could not be extended to include the other strategies.

Would it work for `getPreferredTemporality()` to return a type named `PreferredTemporality`, an enum with three values `Cumulative`, `Delta`, and `Stateless`? In my proposal https://github.com/open-telemetry/opentelemetry-specification/pull/2314#issuecomment-1057472811 I used "strategy" to describe whether the aggregation takes temporality from the exporter preference or overrides to cumulative, but I could be convinced to use different names for either of these options. 

I like this proposal because it makes the proposal Lightstep is interested also work, that is the stateless configuration for Async Counter that does not require Cumulative-to-Delta conversion. Lightstep OTel launchers would choose a Stateless preference. Users who want any other sort of configuration can override the Aggrgation using the View, which seems appropriate. How does this proposal--with three preferences and two strategies--sound?

specification/metrics/sdk.md
jmacd: Your idea with @jack-berg sounds good, the idea of having an aggregation configuration to help here.

By the way, the API spec does have a notion of "current value" for async counters vs "change value" for sync counters, so the distinction of a "natural" temporality, which leads to statelessness, is somewhat present. Thanks.

jmacd:
@open-telemetry/specs-metrics-approvers Please take a look. I've written up (and editorialized on) the suggestion by @alanwest and @jack-berg.

specification/metrics/sdk.md
```diff
@@ -372,7 +374,9 @@ The monotonicity of the aggregation is determined by the instrument type:
 | [Asynchronous Counter](./api.md#asynchronous-counter) | Monotonic |
 | [Asynchrounous UpDownCounter](./api.md#asynchronous-updowncounter) | Non-Monotonic |
 
-This Aggregation does not have any configuration parameters.
+This Aggregation the following configuration parameters:
```
alanwest, line 377:
```suggestion
This Aggregation has the following configuration parameters:
```

specification/metrics/sdk.md
```diff
@@ -672,25 +679,56 @@ The SDK SHOULD provide a way to allow `MetricReader` to respond to
...
+
+The Aggregation temporality preference of the `MetricReader` combined
+with the Aggregation temporality strategy of the `Aggregation`
+determines the exported aggregation temporality.
+
+The two supported `Aggregation` temporality strategies are:
+
+- **Cumulative**: The Aggregation Temporality will be selected as
+  Cumulative, overriding the `MetricReader` preference.
+- **Preferred**: The Aggregation Temporality will be selected
```
alanwest, line 715: I do not have a use case in mind, but does it make sense to have a `Delta` strategy just for completeness?

alanwest:
@jmacd Thank you for the deep thought you've put into to this effort.

I just want to make sure I'm reading your table correctly ... I'm not following the distinction between the "Stateless" and "Selected Temporality, Stateless Preference" columns. Is the former simply communicating which temporality results in no need for the SDK to maintain state - whereas the "Selected Temporality, Stateless Preference" column represents the actual effect of the Stateless Temporality setting?

Or put another way, setting PreferredTemporality=Stateless means the SDK would actually need to maintain state for synchronous UpDownCounters?

specification/metrics/sdk.md
alanwest: I'd be content with its omission without a practical use case imagined.

reyang:
We've discussed this during the Friday triage meeting. This PR is considered as "very important and nice to have, but NOT necessarily blocking the stable release of the SDK spec".

@jmacd will scope down this PR, leaving "stateless" out. We can take this change in the initial SDK spec release.

jmacd:
I removed a single line from this PR:

```
- **Stateless**: The Metric Exporter prefers Delta temporality for
synchronous instruments and Cumulative temporality for asynchronous
instruments to avoid long-term memory.
```

PTAL.

specification/metrics/sdk.md
```diff
@@ -672,25 +679,53 @@ The SDK SHOULD provide a way to allow `MetricReader` to respond to
...
+  according to the `MetricReader` preference.
+
+Three `MetricReader` preferences are available:
```
carlosalberto, line 718:
```suggestion
Two `MetricReader` preferences are available:
```

carlosalberto (approved):

jmacd:
@reyang pointed out an inconsistency in the OTLP exporter SDK spec which refers to the aggregation temporality, not to the preference. I consider this to be an accidental omission. See the fix: c5a5a74

specification/metrics/api.md
```diff
@@ -501,7 +501,7 @@ pattern](https://en.wikipedia.org/wiki/Asynchronous_method_invocation) and
 See the [general requirements for asynchronous instruments](#asynchronous-instrument-api).
 
 Note: Unlike [Counter.Add()](#add) which takes the increment/delta value, the
-callback function reports the absolute value of the counter. To determine the
+callback function reports the current value of the counter. To determine the
```
bogdandrutu, line 504: Understand the change, but would prefer this in a separate PR since is about just "clarification" for API.

specification/metrics/sdk.md
reyang: I think it could be useful in unit tests (e.g. InMemoryExporter) but I'm fine not having it for now (it can be additive change at any time).

reyang (approved):
LGTM.

specification/metrics/sdk.md
```diff
@@ -672,25 +679,53 @@ The SDK SHOULD provide a way to allow `MetricReader` to respond to
 idiomatic approach, for example, as `OnForceFlush` and `OnShutdown` callback
 functions.
 
+### Preferred aggregation temporality
```
bogdandrutu, line 682: I think this is too complicated for our users. I would like to think of a simpler solution.

Can we have our `MetricReader` support a configuration option that says exactly for each type of instrument what to produce?

I feel that we complicate things here because we have this "notion" of `preferred`, because of these we are trying to say "you may prefer delta, but you don't know what you want, I will give you cumulative for this".

I think this is too complicated, have we consider that every "exporter" instead of having a "preferred" to actually come with a concrete temporality <type, temporality> list so they have full control, and we don't try to be smarter like "Preferred".

alanwest (approved):

specification/metrics/sdk_exporters/otlp.md
```diff
@@ -1,4 +1,4 @@
-# OpenTelemetry Metrics Exporter - OTLP
+1# OpenTelemetry Metrics Exporter - OTLP
```
alanwest, line 1:
```suggestion
# OpenTelemetry Metrics Exporter - OTLP
```

specification/metrics/sdk.md
jmacd: > Can we have our MetricReader support a configuration option that says exactly for each type of instrument what to produce?

That was one of the initial proposals, the "5-way" function as it has been called in the review thread above. I think that's fine as a means of configuring the exporters that support both options. If we accept the 5-way function, which seems easy for the in-memory exporter, and fix the output temporality for the console and OTLP exporters to always cumulative, then:

(1) vendors that want always-delta or sometimes-delta will provide implementations of the 5-way function they prefer
(2) Does this mean we can erase the concept of preferred temporality entirely?

jmacd:
I've been convinced that we should eliminate the concept of preferred aggregation temporality, see https://github.com/open-telemetry/opentelemetry-specification/pull/2314#discussion_r819984229

jmacd:
See https://github.com/open-telemetry/opentelemetry-specification/pull/2404

Issue #607, opened 2020-05-14
Specify the use of synchronous instruments from Observer instrument callbacks


jmacd:
The Metrics API has not adopted an asynchronous interface for reporting multiple values per interval. To report a distribution of multiple values in an interval requires the use of a synchronous ValueRecorder. The asynchronous ValueObserver supports only one value per interval.

The argument that led to this is that Observers are primarily meant as an optimization, to lower the cost of reporting. There is no lowering of cost to be had from buffering a number of measurements to be periodically observed. Thus, the argument goes, you should just use a synchronous instrument, i.e., `ValueRecorder`.

There are still reasons why a user may encounter data that has been observed (with good reason, i.e., as an optimization) and contains multiple points. An example is the Golang GC stats, which contain a slice of pause times. These are buffered by the runtime and an observer could therefore observe multiple garbage collection cycles in one callback. 

The same argument could be used to argue for use of `Counter` and `UpDownCounter` from within an asynchronous callback. There are no standard delta-oriented asynchronous instruments: users should use synchronous instruments in the rare case of observing one or more deltas in a callback.

There is nothing in the specification preventing the use of synchronous instruments from asynchronous callbacks, but two questions must be answered:

1. Synchronous instruments require a context.  Proposal: Observer instruments explicitly run in an empty context. When using a synchronous instrument inside an observer callback, the context MUST be empty.
2. Will synchronous instrument events be collected in the same interval as the observer callbacks run? Proposal: Asynchronous observers MUST run before collecting synchronous instrument data for a collection interval, to ensure that synchronous observations are collected in the same time period as the observers they are made from.

I like this proposal because it gives users a perfect solution for well-known rare cases in which an asynchronous instrument wants to report delta and/or distribution values.

jmacd:
A small change is needed to support this in Go:
https://github.com/open-telemetry/opentelemetry-go/pull/725

Here is the location where this functionality will be useful:
https://github.com/open-telemetry/opentelemetry-go-contrib/blob/1cc32bc3cdc6c76e46fb6fe8fb1e0e080d6cc488/plugins/runtime/runtime.go#L268

jmacd:
I would add something about there being no special relationship between observer callbacks and synchronous instruments. There's no way to register the synchronous instrument being "observed", so this should be used as a side-effect related to other observations.

[...]

PR #2317, opened 2022-02-04
Instrument identity, registration conflicts, multi-callback registration, unregister support

### LINKED ISSUE #2226: Meter and instrument identity

### LINKED ISSUE #2232: Allow to unregister/stop/destroy instruments

### LINKED ISSUE #2280: Enable multi-instrument callbacks in the metrics API

jmacd:
Fixes #2226 
Fixes #2232 
Fixes #2280 
Supercedes #2270
Supercedes #2281 

## Changes

See the discussion in #2226.
Defines "duplicate" and "identical" for Meter and Instrument concepts. 
Permits duplicate instrument registration as a first-class feature, including the associated async callbacks
Clarifies that multi-instrument callback idioms are permitted.
Specifies support for unregistering async callbacks.
Specifies a warning for duplicate instrumentation registration conflicts (i.e., same name, not identical)
Specifies that SDKs are permitted to pass-through the semantic conflict.
Specifies that consumers are permitted to reject such data.


specification/metrics/api.md
```diff
@@ -175,19 +180,50 @@ Instruments are used to report [Measurements](#measurement). Each Instrument
...
+implementation MUST return a valid Instrument in every case.
+
+It is unspecified whether or under which conditions the same or
+different Instrument instance will be returned as a result of
+duplicate instrument registration.  Instrument identity, and the terms
+_identical_ and _distinct_ applied to Instruments in this
+specification describes Instrument instances that have identical
+and/or distinct values for all four of their `name`, `kind`, `unit`,
+and `description` attributes.  Language-level features such the
+distinction between integer and floating point numbers SHOULD be
```
jsuereth, line 204: Nit: This sentence feels off.

Language-level features such *as* the distinction...

specification/metrics/api.md
```diff
@@ -175,19 +180,50 @@ Instruments are used to report [Measurements](#measurement). Each Instrument
...
+considered as identifying, as well.
+
+The implementation MUST aggregate data from identical Instruments
+together in its export pipeline.
+
+When through duplicate registration more than one distinct Instrument
```
jsuereth, line 210: Nit:  drop the "through duplicate registration", it's implied and makes this sentence hard to parse.

specification/metrics/api.md
```diff
@@ -264,6 +300,25 @@ Please note that the term _synchronous_ and _asynchronous_ have nothing to do
...
+implementation is REQUIRED to maintain a mapping from every regsitered
+callback to the associated instrument(s).
+
+The implementation MUST provide a way to unregister callbacks
+associated with asynchronous instruments.
+
+The implementation SHOULD reject observations that are made outside of
+a callback registered for the instrument.
+
+Callers SHOULD avoid making duplicate observations from asynchronous
```
jsuereth, line 318: I think this is the right specification level flow here for now.

Alternative:  First observation recorded wins. 
Second alternative: Last observation recorded wins.

specification/metrics/datamodel.md
```diff
@@ -263,18 +263,33 @@ can be converted directly into Timeseries, and share the same identity
...
+
+Consumers of OpenTelemetry metrics data streams MAY reject data
+containing the multiple point kinds for identical `name`, `Resource`,
+and Meter details, even though the [Single-Writer](#single-writer)
+rule is only broken when distinct `Attribute` values are involved.
+
+__Note: The OpenTelemetry SDK specification is written to explicitly
+permit duplicate instrumentation registration, which means the
+potential to produce the error state.  This is considered to be in the
+user's best interest.  Although the SDK has permission to pass-through
```
jsuereth, line 287: Maybe make a note about error conflict logs, and we should have a SHOULD requirement about providing information on how to correct the issue via Views if that's the mode we're going with.

specification/metrics/api.md
jmacd: This wasn't a change, just a consolidation of existing text below.

There's a lot packed into this tiny discussion, I want to keep it separate. The most easy thing to do for implementations is to apply the natural merge function, which _could_ be defined as the correct outcome if we decided to. Then, the statement would simply be that identical observations will be merged by their natural merge function. If you observe a Sum twice, you'll see the total. If you observe a Gauge twice, you'll see the last value.

I have reasons why this definition might be better, but I don't want to talk about them now.

specification/metrics/datamodel.md
jmacd: Oh yeah I meant to add that-- wonderful suggestion.

specification/metrics/api.md
jmacd: Fixed.

specification/metrics/api.md
jmacd: Fixed,

specification/metrics/datamodel.md
jmacd: 73d3c03

specification/metrics/api.md
jmacd: See #2318. I'd like this to be specified as last-value-wins, ideally, but it's separate from this PR.

specification/metrics/datamodel.md
```diff
@@ -263,18 +263,34 @@ can be converted directly into Timeseries, and share the same identity
 characteristics for a Timeseries. A metric stream is identified by:
 
 - The originating `Resource`
+- The metric's associated Meter details (`name`, `version`, and `schema_url`)
 - The metric stream's `name`.
 - The attached `Attribute`s
-- The metric stream's point kind.
+- The metric stream's point kind (`Sum`, `Gauge`, `Histogram` `ExponentialHistogram`)
```
jack-berg, line 269: Missing `Summary`

jack-berg (approved):
I think this is great 👍 👍

specification/metrics/api.md
```diff
@@ -175,19 +180,49 @@ Instruments are used to report [Measurements](#measurement). Each Instrument
...
+identical Meters, denoted _duplicate instrument registration_, the
+implementation MUST return a valid Instrument in every case.
+
+It is unspecified whether or under which conditions the same or
+different Instrument instance will be returned as a result of
+duplicate instrument registration.  Instrument identity, and the terms
+_identical_ and _distinct_ applied to Instruments in this
+specification describes Instrument instances that have identical
+and/or distinct values for all four of their `name`, `kind`, `unit`,
+and `description` attributes.  Language-level features such as the
```
pichlermarc, line 203: In #2270 description was not part of instrument identity. Is this on purpose?

specification/metrics/api.md
```diff
@@ -264,6 +299,25 @@ Please note that the term _synchronous_ and _asynchronous_ have nothing to do
 with the [asynchronous
 pattern](https://en.wikipedia.org/wiki/Asynchronous_method_invocation).
 
+Considering duplicate registration of asynchronous instruments, the
+implementation MUST support multiple callbacks registered per
+identical instrument.
+
+The API MAY provide support for registration of a single callback
+having multiple asynchronous instruments, where idiomatic.  The
+implementation is REQUIRED to maintain a mapping from every regsitered
```
pichlermarc, line 308:
```suggestion
implementation is REQUIRED to maintain a mapping from every registered
```

specification/metrics/api.md
jmacd: Yes, it seems simpler than to create new exceptions anyway. See https://github.com/open-telemetry/opentelemetry-specification/issues/2226#issuecomment-1029527138

specification/metrics/api.md
jmacd: Fixed, thanks.

specification/metrics/datamodel.md
jmacd: Fixed.

specification/metrics/api.md
pichlermarc: Sounds good! 👍

specification/metrics/api.md
```diff
@@ -264,6 +299,25 @@ Please note that the term _synchronous_ and _asynchronous_ have nothing to do
...
+
+The API MAY provide support for registration of a single callback
+having multiple asynchronous instruments, where idiomatic.  The
+implementation is REQUIRED to maintain a mapping from every registered
+callback to the associated instrument(s).
+
+The implementation MUST provide a way to unregister callbacks
+associated with asynchronous instruments.
+
+The implementation SHOULD reject observations that are made outside of
```
legendecas, line 314: This is an interesting point, +1 from me. But how do we define "outside of a callback"? like, say, for a single callback, when SDK initiates a new invocation of the callback, should an observation from a previous invocation of the callback be considered as "outside of a callback"?

pichlermarc (approved):
LGTM with one nit. :slightly_smiling_face:

mateuszrzeszutek (approved):

specification/metrics/datamodel.md
```diff
@@ -263,18 +263,34 @@ can be converted directly into Timeseries, and share the same identity
 characteristics for a Timeseries. A metric stream is identified by:
 
 - The originating `Resource`
+- The metric's associated Meter details (`name`, `version`, and `schema_url`)
 - The metric stream's `name`.
 - The attached `Attribute`s
-- The metric stream's point kind.
+- The metric stream's point kind (`Sum`, `Gauge`, `Histogram` `ExponentialHistogram`, `Summary`)
```
MrAlias, line 269:
```suggestion
- The metric stream's point kind (e.g. `Sum`, `Gauge`, `Histogram` `ExponentialHistogram`, `Summary`)
```

MrAlias (approved):

specification/metrics/api.md
dyladan: For clarification this can happen if a callback takes a long time to complete and is still making observations when the next callback starts

jmacd: Good question, this needs to be better specified.

In a Golang setting, I would pass a `context.Context` with the timeout set during Collect.

1. Use `context.WithTimeout(...)`,
  - in a new goroutine for each instrumentation library in parallel,
  - for each registered callback in sequence,
  - execute the callback
2. In Collect(), wait for the timeout
  - allow canceled contexts to continue running asynchronously
  - do not accept observations on these contexts after the timeout
3. Collect synchronous instruments after asynchronous instruments
  - this keeps the synchronous data fresh in case the async callback(s) hit a deadline
  - question re: use of histogram instruments in asynchronous callbacks (e.g., https://github.com/open-telemetry/opentelemetry-specification/issues/607)
4. Process and pass data to the exporter

The SDK spec already has a lot of language about completing or aborting within timeouts, and states:

> `Collect` SHOULD provide a way to let the caller know whether it succeeded, failed or timed out.

The API spec has this for every asynchronous instrument:

> The callback function SHOULD NOT take indefinite amount of time. If multiple independent SDKs coexist in a running process, they MUST invoke the callback function(s) independently.

The API spec also states this, although it reads like an unstated SDK specification:

> The API MUST treat observations from a single callback as logically taking place at a single instant, such that when recorded, observations from a single callback MUST be reported with identical timestamps.

@legendecas @dyladan WDYT?

specification/metrics/sdk.md
```diff
@@ -952,3 +969,26 @@ called concurrently.
...
+
+The implementation has permission to pass-through violations of the
+semantic "error state" described in the [OpenTelemetry Metrics data
+model](datamodel.md#opentelemetry-protocol-data-model).  This is to
+address the exceptional case of duplicate instrument registration
+conflicts, which are tolerated with warnings in the API.
+
+Duplicate instrument registration conflicts can produce this semantic
+error state, which may lead to violations of the single-writer rule.
+SDKs are permitted to allow these conflicts to take place to avoid
```
jsuereth, line 991: Is it worth calling out that views are meant to resolve these conflicts?

I actually think that because we have the ability to resolve this with loosened view restrictions, we *should* require the error message to help users resolve these conflicts with views.

Specifically, I expect in all instances where we let this conflict through, backend errors will replace SDK errors and possibly be harder to diagnose.

jsuereth (approved):
I think this approach to handling the underlying issues works.  I would like to see us push further on view resolution because I suspect allowing single writer conflicts in the SDK will only, inevitably, cause more user pain than prevention.

Additionally we see precedence in other SDKs enforcing stricter modelling than what is allowed by backends.  

I'd prefer if our SDK was a bit more principled, but our backends are flexible.  That said, I don't think there's a perfect solution here.

specification/metrics/sdk.md
```diff
@@ -414,6 +416,21 @@ instruments that record negative measurements, e.g. `UpDownCounter` or `Observab
 - Min (optional) `Measurement` value in population.
 - Max (optional) `Measurement` value in population.
 
+### Resolving duplicate instrument registration conflicts
+
+Implementations are required by the API specification to tolerate
```
ocelotl, line 421: `tolerate` here is used to refer to something in the API which is not referred in the API as `tolerate`, this can cause confusion.

jmacd:
Reviewers, please refer to the question by @legendecas and following discussion: https://github.com/open-telemetry/opentelemetry-specification/pull/2317#discussion_r800084569

specification/metrics/api.md
jack-berg: The java implementation takes a lock when calling async callbacks that doesn't release until the callback completes. This means there is no way to report observations for a previous invocation, since subsequent collects will be blocked until callbacks from previous collect all complete. It may be a good idea to limit the execution time of callbacks, but we haven't done that yet, and it does introduce more complexity.

specification/metrics/sdk.md
jmacd: @jsuereth do lines 421-435 help?

I am worried that your request leads to a lot of ambiguity, implementation work is needed to prove this out. We may need to distinguish which kind of conflicts can be solved with Views and which cannot. 

If two instruments have conflicting descriptions, everything else identical, there's no way with Views to select one of them, but we can override description, and anyway the data model explicitly excludes Description from identifying a stream, so two ways the problem could be solved:

1. Rename one of the instruments
2. Override the description, to silence the conflicting registration error

The first  is simple because there's no question about whether a merge function needs to be applied.

The second seems more like a special case than a general rule. Are there any other cases where a View can solve a conflict without renaming?

(Related If synchronous and asynchronous forms of the same Counter are created, the two will not be aggregated together by some sort of natural merge function _after preferred temporality is applied_--at least I hope--because that sounds to me like double counting. In this case, should the SDK pass through both temporalities of metric?)

My assumption is that when Views are configured, the user will no longer see a conflict/warning, so the checking has to take place after Views are applied, and then: the View recommendation SHOULD be issued for the instrument that was under construction when the conflict arose (i.e., because an original registration "won first")? 

@jsuereth all of these questions seem to arise **because** we said a recipe SHOULD be provided; seems like a lot more work, would you be willing to put this in an issue and

weyert:
Does this mean both get the option of a callback?

meter.addcallback and asyncGauge.addCallback?

specification/metrics/api.md
reyang: > The java implementation takes a lock when calling async callbacks that doesn't release until the callback completes. This means there is no way to report observations for a previous invocation, since subsequent collects will be blocked until callbacks from previous collect all complete. It may be a good idea to limit the execution time of callbacks, but we haven't done that yet, and it does introduce more complexity.

@jack-berg is that a reentrant lock? Would it be possible for the callback to invoke itself recursively?

specification/metrics/api.md
```diff
@@ -264,6 +299,25 @@ Please note that the term _synchronous_ and _asynchronous_ have nothing to do
...
+Considering duplicate registration of asynchronous instruments, the
+implementation MUST support multiple callbacks registered per
+identical instrument.
+
+The API MAY provide support for registration of a single callback
+having multiple asynchronous instruments, where idiomatic.  The
+implementation is REQUIRED to maintain a mapping from every registered
+callback to the associated instrument(s).
+
+The implementation MUST provide a way to unregister callbacks
```
reyang, line 311: Do we want to mark this as `Experimental`, wait for prototypes across languages, then mark it as `Stable`?

specification/metrics/api.md
```diff
@@ -264,6 +299,25 @@ Please note that the term _synchronous_ and _asynchronous_ have nothing to do
 with the [asynchronous
 pattern](https://en.wikipedia.org/wiki/Asynchronous_method_invocation).
 
+Considering duplicate registration of asynchronous instruments, the
+implementation MUST support multiple callbacks registered per
+identical instrument.
+
+The API MAY provide support for registration of a single callback
+having multiple asynchronous instruments, where idiomatic.  The
+implementation is REQUIRED to maintain a mapping from every registered
```
reyang, line 308: This might need additional clarification; it sounds like we're suggesting that the callback function should be used as part of the identification.

Can a callback function be registered multiple times on a single instrument?

e.g.

```
var f = () => { /* something */ };
target.addEventListener("click", f);
target.addEventListener("click", f); // what would this line do?
target.removeEventListener("click", f);
```

or


```
var f = () => { /* something */ };
var atom1 = target.register("click", f);
var atom2 = target.register("click", f);
target.unregister(atom1); // or atom1.unregister();
target.unregister(atom2); // or atom2.unregister();
```

reyang (changes requested):
I think we need to mark it as `Experimental` and prototype it in several languages before release `Stable`.

reyang:
Need to update the changelog.

specification/metrics/api.md
```diff
@@ -175,19 +180,49 @@ Instruments are used to report [Measurements](#measurement). Each Instrument
...
-Instruments are associated with the Meter during creation, and are identified by
-the name:
+Instruments are associated with the Meter during creation. Instruments
+are identified by all of these fields.
+
+<a name="instrument-type-conflict-detection"></a>
+
+When more than one Instrument of the same `name` is created for
+identical Meters, denoted _duplicate instrument registration_, the
+implementation MUST return a valid Instrument in every case.
```
reyang, line 195: We need to take `Hint API` in the consideration if we plan to add them in the near future.

If we allow _hints_ to be specified duration instrument creation, we'll have to handle multiple calls under a given Meter with the same Instrument name but different hints.

An alternative solution might be allowing _hints_ to be provided for a given instance of instrument - as long as there is way to get such instance. For example, an async instrument creation API might choose not to return a handle/reference.

specification/metrics/api.md
legendecas: @jmacd yes, it does resolve the question. It seems the behavior of JavaScript SDK is similar to the Go's.

I just like to mention that in the API spec, the callback reentrancy is implementation-defined:

> https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#asynchronous-counter-creation
> [OpenTelemetry API](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/overview.md#api) authors SHOULD define whether this callback function needs to be reentrant safe / thread safe or not.

And the `state` of the callback is implementation-defined too:

> The API SHOULD provide some way to pass state to the callback. [OpenTelemetry API](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/overview.md#api) authors MAY decide what is the idiomatic approach (e.g. it could be an additional parameter to the callback function, or captured by the lambda closure, or something else).

So the "outside of a callback" may have different meanings across the implementations. And it would be better to explicitly document this.

jack-berg: > @jack-berg is that a reentrant lock? Would it be possible for the callback to invoke itself recursively?

Yes it is reentrant. It would be possible to invoke itself recursively, though probably not advisable.

specification/metrics/api.md
```diff
@@ -264,6 +299,25 @@ Please note that the term _synchronous_ and _asynchronous_ have nothing to do
 with the [asynchronous
 pattern](https://en.wikipedia.org/wiki/Asynchronous_method_invocation).
 
+Considering duplicate registration of asynchronous instruments, the
+implementation MUST support multiple callbacks registered per
+identical instrument.
+
+The API MAY provide support for registration of a single callback
```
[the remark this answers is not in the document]
jmacd: This is a good example because `/proc/stat` contains many numbers that could be [useful to report](https://askubuntu.com/questions/120953/exact-field-meaning-of-proc-stat). The advantages of multiple instruments per callback are:

1. Read the /proc/stat file once, not once per metric
2. Counters and gauges are temporally aligned by nature (which is why the API spec says callbacks use one timestamp)

The second point means that you could read the CPU stats, procs_blocked, procs_running, softirq state, etc all at once and the receiver can interpret those observations as being correlated, by having the same timestamp. In the context of a [multivariate metrics protocol](https://github.com/open-telemetry/oteps/pull/171), I would define the events of one callback to appear as a single "row" in the columnar encoding => a receiver can study covariance of the variables across a fleet of /proc/stat readings.

Here's how I might write this example.

```python
cpuTime = meter.create_observable_counter("system.cpu.time")
procsRunning = meter.create_observable_gauge("system.procs_running")
procsBlocked = meter.create_observable_gauge("system.procs_blocked")

def proc_stat_observer() -> Iterable[Measurement]:
    with open("/proc/stat") as procstat:
        procstat.readline()  # skip the first line
        for line in procstat:
            if line.startswith("cpu"):
                cpu, *states = line.split()
                yield cpuTime.observe(int(states[0]) // 100, {"cpu": cpu, "state": "user"})
                yield cpuTime.observe(int(states[1]) // 100, {"cpu": cpu, "state": "nice"})
            if line.startswith("procs_running"):
                var, value = line.split()
                yield procsRunning.observe(value)
            if line.startswith("procs_blocked"):
                var, value = line.split()
                yield procsBlocked.observe(value)

callback = meter.register_callback(proc_stat_observer)

def stop():
    callback.unregister()
```

specification/metrics/api.md
```diff
@@ -264,6 +299,25 @@ Please note that the term _synchronous_ and _asynchronous_ have nothing to do
...
+implementation is REQUIRED to maintain a mapping from every registered
+callback to the associated instrument(s).
+
+The implementation MUST provide a way to unregister callbacks
+associated with asynchronous instruments.
+
+The implementation SHOULD reject observations that are made outside of
+a callback registered for the instrument.
+
+Callers SHOULD avoid making duplicate observations from asynchronous
```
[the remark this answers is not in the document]
jmacd: Yes -- the issue of duplicate observations hasn't been identified as clearly as it should, though there is existing language that users SHOULD avoid it. To use your `/proc/stat` example where "cpuN" is the first line of a group of measurements used to identify measurements -- if a buggy kernel produced duplicate lines for "cpu0" let's say, we're not specifying the result.

jmacd:
@aabmass Thank you for the comments. the higher-level points are, I think: need more examples, lingering lack of clarity between the data model and the API/SDK specs. I think I can clear this up, but it will take some time.

specification/metrics/api.md
dyladan: I can say that a typical JS event emitter would treat the first example as:

```javascript
var f = () => { /* something */ };
target.addEventListener("click", f);
target.addEventListener("click", f); // duplicate registration. emitting 'click' would call f twice
target.removeEventListener("click", f);
target.removeEventListener("click", f); // remove only one of the registrations. emitting 'click' would now call f once
```

specification/metrics/sdk.md
MadVikingGod: Also this is the `API specification`, this should be written as such.  Maybe something along the lines of:
```
Implementation [MUST allow duplicate instrument registration](instrument-type-conflict-detection), View configuration offers the user....
```

specification/metrics/api.md
```diff
@@ -129,19 +129,24 @@ This API MUST accept the following parameters:
...
-`Meter` instances are returned from this functions.
-
-Implementations MUST NOT require users to repeatedly obtain a `Meter` again with
-the same name+version+schema_url to pick up configuration changes. This can be
+Meters are identified by all of these fields.  When more than one
+Meter of the same `name`, `version`, and `schema_url` is created, it
+is unspecified whether or under which conditions the same or different
+`Meter` instances are returned.  Meter identity, and the terms
+_identical_ and _distinct_ applied to Meters in this specification
+describes Meter instances that have identical and/or distinct values
```
ocelotl, line 137: I think this sentence needs to be redacted, if `and` and `or` are options, it may be read as `Meter instances that have identical and distinct values...` which sounds contradictory.

specification/metrics/api.md
```diff
@@ -175,19 +180,49 @@ Instruments are used to report [Measurements](#measurement). Each Instrument
 will have the following information:
```
ocelotl, line 180:
```suggestion
will have the following fields:
```
```suggestion
will have the following information:
```

specification/metrics/api.md
```diff
@@ -239,7 +274,7 @@ instrument. It MUST be treated as an opaque string from the API and SDK.
 * It MUST support at least 1023 characters. [OpenTelemetry
   API](../overview.md#api) authors MAY decide if they want to support more.
 
-Instruments can be categorized based on whether they are synchronous or
+Instruments are categorized based on whether they are synchronous or
```
ocelotl, line 277:
```suggestion
Instruments are categorized on whether they are synchronous or
```

specification/metrics/api.md
```diff
@@ -264,6 +299,25 @@ Please note that the term _synchronous_ and _asynchronous_ have nothing to do
 with the [asynchronous
 pattern](https://en.wikipedia.org/wiki/Asynchronous_method_invocation).
 
+Considering duplicate registration of asynchronous instruments, the
+implementation MUST support multiple callbacks registered per
+identical instrument.
+
+The API MAY provide support for registration of a single callback
+having multiple asynchronous instruments, where idiomatic.  The
```
ocelotl, line 307: `having` is a bit unspecific here.

specification/metrics/api.md
ocelotl: Also, what should happen if an instrument's callback is called after it is unregistered?

specification/metrics/api.md
```diff
@@ -264,6 +299,25 @@ Please note that the term _synchronous_ and _asynchronous_ have nothing to do
...
+
+The implementation MUST provide a way to unregister callbacks
+associated with asynchronous instruments.
+
+The implementation SHOULD reject observations that are made outside of
+a callback registered for the instrument.
+
+Callers SHOULD avoid making duplicate observations from asynchronous
+instrument callbacks, as the result of making duplicate observations
+is not specified.
```
ocelotl, line 319: Is something not being specified a reason to include a SHOULD avoid requirement? There are other things that the spec does not specify that don't have a `SHOULD NOT` requirement. Maybe there is another reason why callers SHOULD avoid making duplicate observations.

specification/metrics/api.md
```diff
@@ -396,7 +450,10 @@ The API MUST accept the following parameters:
   rule](#instrument-unit).
 * An optional `description`, following the [instrument description
   rule](#instrument-description).
-* A `callback` function.
+
+The API MUST support an idiomatic way to associate one or more
+`callback` functions with each instrument, whether through the
```
ocelotl, line 455:
```suggestion
`callback` functions with each Asynchronous Counter instrument, whether through the
```

specification/metrics/api.md
```diff
@@ -613,7 +670,10 @@ The API MUST accept the following parameters:
...
```
ocelotl, line 675:
```suggestion
`callback` functions with each Asynchronous Gauge, whether through the
```

specification/metrics/api.md
```diff
@@ -885,7 +945,10 @@ The API MUST accept the following parameters:
...
```
ocelotl, line 950:
```suggestion
`callback` functions with each Asynchronous UpDownCounter, whether through the
```

specification/metrics/datamodel.md
```diff
@@ -263,13 +263,34 @@ can be converted directly into Timeseries, and share the same identity
...
+- The point kind (e.g. `Sum`, `Gauge`, `Histogram` `ExponentialHistogram`, `Summary`)
+- The metric stream's unit
+- Intrinsic properties of the point, including temporality and monotonicity
 
 It is possible (and likely) that more than one metric stream is created per
 `Instrument` in the event model.
 
+Points with identical `name`, `Resource`, and Meter details but
+different point kind or different point kind intrinsic properties
+SHOULD be considered an "error state", due to semantic disagreement,
```
ocelotl, line 278: Who should consider this to be an "error state"? Consumers of opentelemetry metrics data streams?

If that is the case, then this paragraph is better removed since it is redundant with the next one which is specific on what to do if this happens and does not introduce an unspecific concept, "error state".

specification/metrics/api.md
reyang: @ocelotl I discussed this offline with @jmacd yesterday and he is working on clarifying these things.

specification/metrics/sdk.md
```diff
@@ -414,6 +416,36 @@ instruments that record negative measurements, e.g. `UpDownCounter` or `Observab
...
+arise, because they lead to semantic errors.  To help the user manage
+such conflicts, the implementation SHOULD warn about dupliclate
+instrument registration conflicts after applying View configuration.
+
+View configuration cannot work around every conflict, and in some
+cases the user will want to repair faulty instrumentation instead.
+When a potential conflict arises between two `Metric` identities
+having the same `name`:
+
+1. If the potential conflict involves a multiple `description`
```
jack-berg, line 439:
```suggestion
1. If the potential conflict involves multiple `description`
```

specification/metrics/sdk.md
```diff
@@ -414,6 +416,36 @@ instruments that record negative measurements, e.g. `UpDownCounter` or `Observab
...
+duplicate instrument registration to occur, regardless of conflicting
+identities.  This ensures that user data is at least delivered to the
+exporter, recognizing that such data MAY create a [semantic error
+state](datamodel.md#opentelemetry-protocol-data-model-producer-expectations)
+indicated by data having multiple `Metric` objects for a given
+`Resource` and `Scope`.
+
+The SDK is required to issue warnings when instrumentation conflicts
+arise, because they lead to semantic errors.  To help the user manage
+such conflicts, the implementation SHOULD warn about dupliclate
```
jack-berg, line 431:
```suggestion
such conflicts, the implementation SHOULD warn about duplicate
```

jmacd:
All: I'm still working on this PR, particularly the need for examples and a spelled-out API for callback registration and unregistration.

I've added new paragraphs clarifying uniqueness for `Metric` objects in the OTel data model and spelled out recommendations for how to handle duplicate-conflicting data. See e9384f0f121fdac45481f88b2d966747e51d0ec5 and 40f8ed80b68107a8614fe63556917bd0e1fb3360.

specification/metrics/datamodel.md
```diff
@@ -257,22 +257,90 @@ to map into, but is used as a reference throughout this document.
...
+objects as a basic feature, so the appearance of multiple `Metric`,
+considered a "semantic error", generally requires duplicate
+conflicting instrument registration to have occurred somewhere.
+
+Producers MAY be able to remediate the problem, depending on whether
+they are an SDK, a downstream processor, or a consumer at the
+destination:
+
+1. If the potential conflict involves a non-identifying property (i.e.,
+   `description`), the producer SHOULD choose the longer string.
```
aabmass, line 312: It says above the histogram parameters are non-identifying

> - Bucket boundaries of a `Histogram` data point
> - Scale or bucket count of a `ExponentialHistogram` data point.

How should conflicts in these be resolved?

specification/metrics/datamodel.md
```diff
@@ -257,22 +257,90 @@ to map into, but is used as a reference throughout this document.
...
+Producers SHOULD prevent the presence of multiple `Metric` identities
+for a given `name` with the same `Resource` and `Scope` attributes.
+Producers are expected to aggregate data for identical `Metric`
+objects as a basic feature, so the appearance of multiple `Metric`,
+considered a "semantic error", generally requires duplicate
+conflicting instrument registration to have occurred somewhere.
+
+Producers MAY be able to remediate the problem, depending on whether
+they are an SDK, a downstream processor, or a consumer at the
+destination:
```
aabmass, line 309: I think the note about consumers doesn't apply to the producer recommendations, unless I'm missing something

specification/metrics/api.md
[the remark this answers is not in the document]
legendecas: IIRC, the batch observer API was removed in the feature-freeze metrics API intentionally. Although, there are great use cases for it.

specification/metrics/api.md
```diff
@@ -129,19 +129,24 @@ This API MUST accept the following parameters:
...
   recorded in the emitted telemetry.
 
-It is unspecified whether or under which conditions the same or different
-`Meter` instances are returned from this functions.
-
-Implementations MUST NOT require users to repeatedly obtain a `Meter` again with
-the same name+version+schema_url to pick up configuration changes. This can be
+Meters are identified by all of these fields.  When more than one
+Meter of the same `name`, `version`, and `schema_url` is created, it
+is unspecified whether or under which conditions the same or different
```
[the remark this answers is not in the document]
dyladan: I think erlang is the poster child for that iirc. In order to have the same instance returned there needs to be some state maintained external to the function which breaks the functional design principle.

specification/metrics/api.md
jmacd: Done.

specification/metrics/api.md
jmacd: Fixed.

specification/metrics/api.md
jmacd: Well, rephrased actually. I've created a general requirements section for the API because things were being stated 3 times for the async instruments.

specification/metrics/datamodel.md
jmacd: I don't see them as conflicts, per se. They're definitely not semantic conflicts. For the exponential histogram case, merge rules are well-defined. For the explicit boundaries case, not sure -- interpolation can work -- I suppose this leads back to the same problem with consumers having no way to report semantic problems. I'd like a solution to report problems like this but not as a part of this PR.

specification/metrics/api.md
jmacd: There's a danger of this PR's scope creeping, but I've moved some of this discussion into a new section of the SDK specification. There were SDK requirements sprinkled into the API requirements at this spot. I've defined "outside the callback" there by way of requiring callbacks to be executed on behalf of a single `MetricReader`.

specification/metrics/sdk.md
jmacd: Fixed, thanks.

specification/metrics/api.md
jmacd: I still haven't added examples to this PR, sorry! Will catch up on Monday.

specification/metrics/api.md
jmacd: I've spelled out a lot more about instrument identity now, and in the data model we're left with some hard rules about what's identifying and what's not. Hints should lead to non-identifying traits, like histogram boundaries and scale for example.

specification/metrics/api.md
jmacd: See https://github.com/open-telemetry/opentelemetry-python/issues/2454

specification/metrics/datamodel.md
```diff
@@ -263,18 +263,35 @@ can be converted directly into Timeseries, and share the same identity
 characteristics for a Timeseries. A metric stream is identified by:
 
 - The originating `Resource`
+- The metric's associated Meter details (`name`, `version`, and `schema_url`)
 - The metric stream's `name`.
 - The attached `Attribute`s
-- The metric stream's point kind.
+- The point kind (e.g. `Sum`, `Gauge`, `Histogram` `ExponentialHistogram`, `Summary`)
```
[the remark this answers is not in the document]
jmacd: Fixed in a lot of detail. This was unintentional.

specification/metrics/datamodel.md
```diff
@@ -263,18 +263,35 @@ can be converted directly into Timeseries, and share the same identity
 characteristics for a Timeseries. A metric stream is identified by:
 
 - The originating `Resource`
+- The metric's associated Meter details (`name`, `version`, and `schema_url`)
```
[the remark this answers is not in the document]
jmacd: I've spelled this out in much more detail, see what you think.

specification/metrics/api.md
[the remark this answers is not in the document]
jmacd: No. The word "valid" is still used. I've added a definition for what "Valid" means in this case.

specification/metrics/api.md
jmacd: I refactored this conversation. There's a new sdk.md section on asynchronous instrument requirements.

specification/metrics/datamodel.md
jmacd: Hmm. See what you think of the new language here. I want a general-purpose inspecific way to refer to this sort of error, since there are many subtle details when you look closely.

specification/metrics/datamodel.md
jmacd: Fixed.

jmacd:
I've responded to all the feedback at this time. Still missing examples, but hopefully this brings us closer to agreeing on the requirements from the API, SDK, and data model perspectives. @open-telemetry/specs-metrics-approvers

jsuereth (approved):
Like the additions/clarifications.

specification/metrics/api.md
```diff
@@ -264,6 +303,27 @@ Please note that the term *synchronous* and *asynchronous* have nothing to do
...
+The API MUST support creation of asynchronous instruments by passing a
+single callback function to be registered to the newly created
+instrument.
+
+The API SHOULD support creation of asynchronous instruments
+by passing a variable number of callback functions, greater than or
+equal to zero, to be registered to the newly created instrument.
+
+The API SHOULD support registration of callbacks bound to
+one or more instruments outside of instrument construcotrs, provided
```
pichlermarc, line 315:
```suggestion
one or more instruments outside of instrument constructors, provided
```

specification/metrics/sdk.md
```diff
@@ -417,6 +419,48 @@ instruments that record negative measurements, e.g. `UpDownCounter` or `Observab
...
+between two non-identical `Metric` instances having the same `name`:
+
+1. If the potential conflict involves multiple `description`
+   properties, setting the `description` through a configured View
+   SHOULD avoid the warning.
+2. If the potential conflict involves instruments that can be
+   distinguished by a supported View selector (e.g., instrument type)
+   a View recipe SHOULD be printed advising the user how to avoid the
+   warning by renaming one of the conflicting instruments.
+3. Otherwise (e.g., use of multiple units), the implementation SHOULD
```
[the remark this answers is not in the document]
jmacd: Agree, though we probably still need a catch-all for conflicts that aren't resolvable by Views. In OTel-Go you might have an int64/float64 conflict: should there be a View feature to override?

jmacd:
@reyang Please see [16b1be0](https://github.com/open-telemetry/opentelemetry-specification/pull/2317/commits/16b1be0d8552da161fb85d6b3df44d76bb9450b6)

jmacd:
@open-telemetry/specs-metrics-approvers I changed the option to support multiple instrument callbacks into a MAY, since I see it as a true option for cases where it is a natural idiom and where there is user demand.  See [5e77317](https://github.com/open-telemetry/opentelemetry-specification/pull/2317/commits/5e77317ba083c11fb2ff9908843e778d6fe0fcce)

specification/metrics/api.md
```diff
@@ -1106,10 +1108,44 @@ for the interaction between the API and SDK.
...
+```Python
+# Python
+class Device:
+    """A device with one counter"""
+
+    def __init__(self, meter, x):
+        self.x = x
+        self.usage = meter.create_observable_counter(name="usage", description="count of items used")
+        self.pressure = meter.create_observable_gauge(name="pressure", description="force per unit area")
+        self.cb = meter.register_callback([self.usage, self.pressure], self.observe)
```
legendecas, line 1129: This example case seems to supersede the batch observers. +1 from me for the idea of "multiple callbacks" instead of a new interface like "batch observer" versus "observable metrics". It is simpler IMO.

specification/metrics/api.md
```diff
@@ -264,6 +303,33 @@ Please note that the term *synchronous* and *asynchronous* have nothing to do
...
+each callback.
+
+The API SHOULD support unregistration of callbacks, in order to cease
+making observations.  Note that unregistration of callbacks does not
+undo the effect of instrument registration itself; duplicate
+instrument registration conflicts SHOULD still occur for instruments
+with no registered callbacks.  The effect of unregistering a callback
+more than once is unspecified.
+
+Callback functions SHOULD NOT take an indefinite amount of time.
```
anuraaga, line 326: Since callback functions are registered by the user, should this read more like "Documentation SHOULD be provided indicating to users that callback functions should return quickly and should not make duplicate observations".

anuraaga (approved):

specification/metrics/sdk.md
```diff
@@ -417,6 +419,48 @@ instruments that record negative measurements, e.g. `UpDownCounter` or `Observab
...
+`MetricReader` during collection.
+
+The implementation SHOULD disregard the accidental use of APIs
+appurtenant to asynchronous instruments outside of registered
+callbacks in the context of a single `MetricReader` collection.
+
+The implementation SHOULD use a timeout to prevent indefinite callback
+execution.
+
+The implementation MUST complete the execution of all callbacks for a
```
legendecas, line 436: Does this statement override the following statement in the API spec?

> [OpenTelemetry API](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/overview.md#api) authors SHOULD define whether this callback function needs to be reentrant safe / thread safe or not.
> https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#asynchronous-counter-creation

If so, would it be better to remove the statement in the API spec? Or one SDK compliant with the API spec may not be compliant with the SDK spec.

specification/metrics/api.md
```diff
@@ -129,19 +129,24 @@ This API MUST accept the following parameters:
...
+is unspecified whether or under which conditions the same or different
+`Meter` instances are returned.  The term _identical_ applied to
+Meters describes instances where all identifying fields are equal.
+The term _distinct_ applied to Meters describes instances where at
+least one identifying field has a different value.
+
+Implementations MUST NOT require users to repeatedly obtain a `Meter` with
+the same identity to pick up configuration changes. This can be
 achieved either by allowing to work with an outdated configuration or by
 ensuring that new configuration applies also to previously returned `Meter`s.
```
dyladan, line 143: How do you define "working" if a meter can work with outdated configurations? If this is allowed, what is the non-working failure mode we are trying to guard against here?

specification/metrics/api.md
dyladan, line 326: Is a long-running or non-returning callback specifically considered to be undefined behavior?

specification/metrics/api.md
```diff
@@ -482,10 +544,32 @@ meter.CreateObservableCounter<UInt64>("caesium_oscillates", () => clock.GetCaesi
...
+        self.cb = meter.create_observable_counter(name="usage", description="count of items used", self.counter_callback)
+
+    def counter_callback(self, result):
+        result.Observe(self.read_counter(), {'x', self.x})
+
+    def read_counter(self):
+        return 100  # ...
+    
+    def stop(self):
+        self.cb.unregister()
```
bogdandrutu, line 570: Can we support "Closing" the instrument/unregister the entire instrument instead? I feel that is much simpler, and is needed anyway.

specification/metrics/api.md
jmacd: Correct. We may take this as a case of a bad prototype leading to a better design. 😁

specification/metrics/api.md
jmacd: I'm a little uncertain about this (preexisting) statement. @reyang, do you think we could remove "work with an outdated configuration", or would you help us clarify its itent?

aabmass (approved):
Edits LGTM, thanks Josh. The multi-instrument callback and unregistration feel a little vague in comparison to the other requirements in the spec, but I think this could be addressed in a follow up.

specification/metrics/sdk.md
jmacd: Yes, it seems to contradict. Does anyone disagree with requiring the callback function to be "reentrant safe"? I don't want to define these terms. The SDK is required to evaluate callback functions independently for separate Readers; even with a single SDK in the process the caller should expect to be called concurrently, right?

specification/metrics/api.md
jmacd: I'd like implementations to do whatever is realistic here, and don't want to over constrain them. In OTel-Go for example, in a setting where we are committed to timeouts, I would propose to allow each collection interval to use separate goroutines for collection. The export would not block on the callbacks. When the `context.Context` timeout is reached, those goroutines that have not yet finished are on their own, and any measurements they produce will be considered out-of-date on arrival.

specification/metrics/sdk.md
legendecas: > The SDK is required to evaluate callback functions independently for separate Readers; even with a single SDK in the process the caller should expect to be called concurrently, right?

Agreed. One or more metric readers can initiate concurrent collections.

specification/metrics/api.md
```diff
@@ -264,6 +303,33 @@ Please note that the term *synchronous* and *asynchronous* have nothing to do
...
 
+The API MUST support creation of asynchronous instruments by passing a
+single callback function to be registered to the newly created
+instrument.
+
+The API SHOULD support creation of asynchronous instruments
+by passing a variable number of callback functions, greater than or
+equal to zero, to be registered to the newly created instrument.
+
+The API MAY support registration of callbacks bound to
```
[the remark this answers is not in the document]
jmacd: Yes, I mean "bound" as in to say that there is always a mapping between instrument and callback. This allows the implementation to enumerate the callbacks that report for any given instrument independently of whether they are registered when the instrument is created. 

The implementation has to have a way to enumerate the callbacks that report for any given instrument because of the requirements about duplicate instrument registration; i.e., we have more than one callback per instrument.

Adding an allowance for multiple instruments changes nothing, i.e., we still have more than one callback per instrument, the only difference is that one callback may appear in the list of callbacks for multiple instruments.

This "binding" so to speak is meant to support exporting instruments with different frequency at some point in the future.

jack-berg:
> can we split that in a separate PR since the clarifications about the identity of the instruments is a great addition and can be merged immediately.

@bogdandrutu I disagree. I think resolving instrument identity requires supporting multiple callbacks, and supporting multiple callbacks really needs a solution for unregistration to be practical.

specification/metrics/api.md
```diff
@@ -264,6 +303,33 @@ Please note that the term *synchronous* and *asynchronous* have nothing to do
 with the [asynchronous
 pattern](https://en.wikipedia.org/wiki/Asynchronous_method_invocation).
 
+The API MUST support creation of asynchronous instruments by passing a
+single callback function to be registered to the newly created
```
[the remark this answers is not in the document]
jmacd: I meant this in the sense "MUST support 1 callback", which is not to preclude creation of instruments with zero callbacks (a SHOULD) or more than one callback (a SHOULD).

specification/metrics/api.md
jmacd: I think you're suggesting that to unregister or close a specific Instrument "instance" is to unregister the callback(s) that were passed to that instances constructor.

The only difference with what I've tried to specify instead is that it should be clear what we're counting (callback instances) and what lasts forever (instrument instances). It doesn't matter how many times an instrument is registered, it matters how many callbacks were registered as a result. 

The API was marked stable without a requirement that observable instruments return anything. I've made the option to unregister callbacks a true option, so that you aren't required to support unregistration of callbacks passed to instrument constructors, a separate API is permitted. So, without need for an instrument Close() or calling Unregister directly on an asynchronous instrument (whatever that means), you could add a new method, e.g.,

```Python
instrument = meter.create_observable_counter(name="usage", description="count of items used", first_callback)
# there is one callback registered
cb = instrument.register_callback(second_callback)
# there are now two callbacks registered
meter.unregister_callback(cb)
# there is now one callback registered
```
In this example, you can't unregister callbacks passed to the constructor. The API MAY provide another way to register callbacks that can be unregistered.

specification/metrics/api.md
```diff
@@ -264,6 +303,33 @@ Please note that the term *synchronous* and *asynchronous* have nothing to do
...
+The API SHOULD support creation of asynchronous instruments
+by passing a variable number of callback functions, greater than or
+equal to zero, to be registered to the newly created instrument.
+
+The API MAY support registration of callbacks bound to
+one or more instruments outside of instrument constructors, provided
+the API declaratively states which instrument(s) will be be used from
+each callback.
+
+The API SHOULD support unregistration of callbacks, in order to cease
```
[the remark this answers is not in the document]
jmacd: I am now considering a firmer requirement, as it would seem to help:
1. Callback functions passed to instrument constructors cannot be unregistered, because instruments are forever
2. APIs SHOULD support a way to register callbacks after instrument construction as a separate step
3. The API SHOULD support a way to unregister **only those callbacks that were registered after instrument construction**.

bogdandrutu:
@jack-berg 

>  I think resolving instrument identity requires supporting multiple callbacks, and supporting multiple callbacks really needs a solution for unregistration to be practical. In other words, all 3 of these topics are sort of woven together and should be solved together.

How come?

jmacd:
@jack-berg  Here's how we can (I hope quickly) proceed: https://github.com/open-telemetry/opentelemetry-specification/pull/2317#discussion_r807214698

specification/metrics/sdk.md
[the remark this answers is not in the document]
bogdandrutu: int64/float64 I would solve this by going to the most generic which is float64 :) This is actually a good example that you may want to have in the examples, because I think is a good default solution.

specification/metrics/api.md
bogdandrutu: This kind of changes the signature of the callback in a stable API document. Is this not a breaking change? How does languages like Java, C# support this?

Does this have a different signature? Maybe to be consistent with the other callbacks passed in the constructor we pass an array of "results" for all the instruments passed in the register_callback, to avoid adding of a new API called `Observe` on the instrument?

/cc @open-telemetry/java-maintainers @open-telemetry/dotnet-maintainers?

jmacd:
@jack-berg and @bogdandrutu here's the slight removal that takes multi-instrument callback support out without removing the concept of unregistration or creating the ambiguity that led Bogdan to suggest `Close()`.

[478b3cd](https://github.com/open-telemetry/opentelemetry-specification/pull/2317/commits/478b3cd84699a1289b631acbadc5004b30134a9d)

specification/metrics/api.md
```diff
@@ -305,25 +305,17 @@ Please note that the term *synchronous* and *asynchronous* have nothing to do
...
-The API SHOULD support unregistration of callbacks, in order to cease
-making observations.  Note that unregistration of callbacks does not
-undo the effect of instrument registration itself; duplicate
-instrument registration conflicts SHOULD still occur for instruments
-with no registered callbacks.  The effect of unregistering a callback
-more than once is unspecified.
+Where the API supports registration of callback functions after
+asynchronous instrumentation creation, it MUST return something (e.g.,
+a registration handle, receipt or token) to the user that supports
+undoing the effect of callback registation.
```
bogdandrutu, line 318: Would be good as a "recommendation":
1. The "late registered" callbacks have or have not the same signature as the callbacks passed in the constructor? Because you need different signature for the multi-instrument to be backwards compatible.
2. The "late registration" and "unregister" are operations on the "Instrument" or on the "Meter"?

specification/metrics/api.md
jmacd: This was given as a MAY in order to prevent breaking stable APIs. You could provide a separate instrument constructor to add this in a stable API.

bogdandrutu (commented):
Indeed I am happier to approve this now, only small recommendation suggestions would be good.

specification/metrics/api.md
jmacd: In any case, I removed multi-instrument callbacks from this PR as discussed in the tracking issue #2280. Another PR can add this easily, it seems.

specification/metrics/sdk.md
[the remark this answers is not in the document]
jmacd: Are you suggesting an automatic conflict resolution, or are you suggesting a new View feature? Scope is creeping here, let's discuss that if/when someone cares.

specification/metrics/api.md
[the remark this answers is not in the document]
jmacd: ^^^ This proposal is now applied. To support unregistration of callbacks, you'll have to register them in a separate call than the instrument constructor.

specification/metrics/api.md
```diff
@@ -988,10 +1108,44 @@ for the interaction between the API and SDK.
 * A value
 * [`Attributes`](../common/common.md#attributes)
 
+## Multi-instrument asynchronous callback example
+
+Where idiomatic, [the Metrics API supports the use of multiple
+instruments from a single registered
+callback](#asynchronous-instrument).  Provided there is runtime
+support for dynamic scope (e.g., thread-local state, context
```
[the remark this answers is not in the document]
jmacd: Sure, that's good. I'll put your version of this in the future PR. I've removed multi-instrument callbacks from this PR.

specification/metrics/api.md
jmacd: > Because you need different signature for the multi-instrument to be backwards compatible.

I believe we're talking about implementation specifics, but I would say single-instrument callback registration SHOULD use the same signature whether passed to the constructor or registered later.  A multi-instrument callback would be a different signature; this PR is meant to open the door for multi-instrument callbacks by clarifying what is meant by registration and unregistration, but it would be different signature callback functions used for multi-instrument callbacks.

jmacd: > The "late registration" and "unregister" are operations on the "Instrument"

I was referring to late registration for single-instrument callbacks, so the method to register a single-instrument callback would be an Instrument method.

The unregister method API was not specified, other than that some means be provided. In the examples, which I've updated, the registration token returned by `register_callback()` supports an `unregister()` method directly. I would also be happy with a `Meter` or `MeterProvider` interface. This is not a primary interface, not too important.

specification/metrics/sdk.md
```diff
@@ -417,6 +419,48 @@ instruments that record negative measurements, e.g. `UpDownCounter` or `Observab
 - Min (optional) `Measurement` value in population.
 - Max (optional) `Measurement` value in population.
 
+### Observations inside asynchronous callbacks
+
+Callback functions MUST be invoked in the context of a specific
```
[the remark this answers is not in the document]
jmacd: Right. I rephrased this to avoid the ambiguity.

specification/metrics/api.md
```diff
@@ -172,22 +177,56 @@ Also see the respective sections below for more information on instrument creati
...
+Instruments describes instances where all identifying fields are
+equal.  The term _distinct_ applied to Instruments describes instances
+where at least one field value is different.
+
+The implementation MUST aggregate data from identical Instruments
+together in its export pipeline.
+
+When more than one distinct Instrument is registered with the same
+`name` for identical Meters, the implementation SHOULD emit a warning
+to the user informing them of duplicate registration conflict(s).
```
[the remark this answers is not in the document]
jmacd: The "warning" is supposed to include a Views recipe if possible, otherwise yes.

specification/metrics/api.md
```diff
@@ -172,22 +177,56 @@ Also see the respective sections below for more information on instrument creati
...
+
+It is unspecified whether or under which conditions the same or
+different Instrument instance will be returned as a result of
+duplicate instrument registration.  The term _identical_ applied to
+Instruments describes instances where all identifying fields are
+equal.  The term _distinct_ applied to Instruments describes instances
+where at least one field value is different.
+
+The implementation MUST aggregate data from identical Instruments
+together in its export pipeline.
```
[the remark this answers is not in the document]
jmacd: Correct. I moved this line into sdk.md.

jmacd:
There are a number of comments in this review thread asking me to improve or clarify existing parts in the specification that are unrelated to this PR. I think we should move those changes into future PRs. That includes https://github.com/open-telemetry/opentelemetry-specification/pull/2317#discussion_r807019151 https://github.com/open-telemetry/opentelemetry-specification/pull/2317#discussion_r807013913 https://github.com/open-telemetry/opentelemetry-specification/pull/2317#discussion_r806415778 and https://github.com/open-telemetry/opentelemetry-specification/pull/2317#discussion_r806413848.

jmacd:
@open-telemetry/specs-metrics-approvers We've  This PR still combines solutions for issues #2226 and #2232. While I think #2232 could be factored apart from the core change here, they are closely linked. (e.g., https://github.com/open-telemetry/opentelemetry-specification/issues/2226#issuecomment-1023846383). Whereas before we had not specified what happens when multiple callbacks are registered, now we are saying that they're supported and you must have a way of unregistering them. I don't like the intermediate state where we say multiple callbacks are supported with no way of unregistering them.

jack-berg (approved):

specification/metrics/api.md
jmacd: https://github.com/open-telemetry/opentelemetry-specification/issues/2350

specification/metrics/sdk.md
jmacd: https://github.com/open-telemetry/opentelemetry-specification/issues/2350

specification/metrics/api.md
jmacd: https://github.com/open-telemetry/opentelemetry-specification/issues/2350

MadVikingGod (approved):
The go SDK is currently trying to rework its API to a more maintainable and idiomatic form.  This change, specifically the relaxation of the callback at registration, allows us to better match the go ecosystem.

codeboten (approved):

legendecas (approved):
👍

bogdandrutu:
@jmacd are we sure that this fixes https://github.com/open-telemetry/opentelemetry-specification/issues/2232?

jmacd:
> @jmacd are we sure that this fixes https://github.com/open-telemetry/opentelemetry-specification/issues/2232?

I tried to answer this here: https://github.com/open-telemetry/opentelemetry-specification/issues/2232#issuecomment-1041960737

I changed the PR description to state this is part of #2232 either way, would prefer not to block on the answers to what I think are independent questions about supporting API-level deletion and/or a SDK-level memory preference.

jmacd:
@bogdandrutu Do you think it's worth holding up this change for an answer to the question above? There are a number of approvals and the PR description has included "unregister support" from the beginning. It seems your question is whether adding unregister support is sufficient to address #2232, which we can debate, but do we still need to debate unregister support?

The original request in #2232 is for a way to Close an entire Meter, unregister its instruments and stop reporting its metrics. in the context of a metrics bridge. I've tried to frame this as request for unregistering callbacks and for allowing the SDK to automatically forget stale metrics.

@open-telemetry/specs-metrics-approvers Do we know of other reasons to support unregister()? I've tried to argue for this on the principle that without unregister support, users are forced to implement synchronization within callbacks on their own, which is not simple. (Unlike synchronous instruments, which are easy to stop using.)

ocelotl (approved):

jmacd:
Yesterday I removed support for multi-instrument callbacks from this PR.
Here is the PR that will add the same after this PR merges:
https://github.com/open-telemetry/opentelemetry-specification/pull/2355

specification/metrics/api.md
```diff
@@ -266,6 +302,26 @@ Please note that the term *synchronous* and *asynchronous* have nothing to do
 with the [asynchronous
 pattern](https://en.wikipedia.org/wiki/Asynchronous_method_invocation).
 
+The API MUST support creation of asynchronous instruments by passing
+zero or more callback functions to be permanently registered to the
+newly created instrument.  Callbacks functions passed to the
+instrument constructor cannot be unregistered.
```
bogdandrutu, line 308: I would form this in a way like:

```
Callbacks functions passed to the instrument constructor MAY not have support to unregistered.
```

bogdandrutu (approved):

specification/metrics/api.md
bogdandrutu: I would prefer to address this in a separate PR actually.

bogdandrutu, line 308:
```suggestion
newly created instrument.
```

bogdandrutu:
@MadVikingGod
> This change, specifically the relaxation of the callback at registration, allows us to better match the go ecosystem.

You still need to offer that.

PR #2320, opened 2022-02-07
Pin markdownlint-cli version and apply fixes

arminru:
Some breaking changes in the most recent markdownlint-cli version (see https://github.com/igorshubovych/markdownlint-cli/releases) cause all current checks to fail. This PR pins its version so this does not happen again and applies the fixes proposed by markdownlint (inconsistent emphases within one file and unescaped asterisks).


yurishkuro (approved):

bogdandrutu (approved):

tigrannajaryan:
Thanks for the fix.

specification/schemas/overview.md
```diff
@@ -239,9 +239,9 @@ field in the messages:
 - The schema_url field in the InstrumentationLibraryLogs message applies to the
   contained LogRecord messages.
 
-- If schema_url field is non-empty both in Resource* message and in the
-  contained InstrumentationLibrary* message then the value in
-  InstrumentationLibrary* message takes the precedence.
+- If schema_url field is non-empty both in Resource\*message and in the
```
chalin, line 242: This erased the space between Resource* and message. I don't believe that was intended, right?

PR #2323, opened 2022-02-07
[linters] Install tools based on versions in package.json

chalin:
I'm having issues running certain spec checks from the website because it's working from a spec submodule. This PR:

- Switches to using `npx` to run npm commands rather than hardcoded paths
- Uses `npm install` to install all necessary NPM-based tools
  - Adds a `package.json` file for this purpose -- with `markdownlint-cli` pinned to `v0.31.0` as was done in #2320

IMHO, this is a clean and conventional way to deal with NPM packages. WDYT?

/cc @austinlparker

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -14,7 +14,9 @@ jobs:
       uses: actions/checkout@v2
 
     - name: run markdownlint
-      run: make markdownlint
+      run: |
+        npm install
+        make markdownlint
 
   yamllint:
     runs-on: ubuntu-latest
@@ -37,7 +39,9 @@ jobs:
       uses: actions/checkout@v2
 
     - name: run markdown-link-check
-      run: make markdown-link-check
+      run: |
+        npm install
+        make markdown-link-check
 
   markdown-toc-check:
     runs-on: ubuntu-latest
@@ -46,7 +50,9 @@ jobs:
       uses: actions/checkout@v2
 
     - name: run markdown-toc
-      run: make markdown-toc
+      run: |
+        npm install
+        make markdown-toc
 
     - name: validate markdown-toc
       run: git diff --exit-code ':*.md' || (echo 'Generated markdown Table of Contents is out of date, please run "make markdown-toc" and commit the changes in this PR.' && exit 1)
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -13,7 +13,7 @@ jobs:
     - name: check out code
       uses: actions/checkout@v2
 
-    - name: npm install
+    - name: install dependencies
       run: npm install
 
     - name: run markdownlint
@@ -39,7 +39,7 @@ jobs:
     - name: check out code
       uses: actions/checkout@v2
 
-    - name: npm install
+    - name: install dependencies
       run: npm install
 
     - name: run markdown-link-check
@@ -51,7 +51,7 @@ jobs:
     - name: check out code
       uses: actions/checkout@v2
 
-    - name: npm install
+    - name: install dependencies
       run: npm install
 
     - name: run markdown-toc
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -13,10 +13,11 @@ jobs:
     - name: check out code
       uses: actions/checkout@v2
 
+    - name: npm install
+      run: npm install
+
     - name: run markdownlint
-      run: |
-        npm install
-        make markdownlint
+      run: make markdownlint
 
   yamllint:
     runs-on: ubuntu-latest
@@ -38,10 +39,11 @@ jobs:
     - name: check out code
       uses: actions/checkout@v2
 
+    - name: npm install
+      run: npm install
+
     - name: run markdown-link-check
-      run: |
-        npm install
-        make markdown-link-check
+      run: make markdown-link-check
 
   markdown-toc-check:
     runs-on: ubuntu-latest
@@ -49,10 +51,11 @@ jobs:
     - name: check out code
       uses: actions/checkout@v2
 
+    - name: npm install
+      run: npm install
+
     - name: run markdown-toc
-      run: |
-        npm install
-        make markdown-toc
+      run: make markdown-toc
 
     - name: validate markdown-toc
       run: git diff --exit-code ':*.md' || (echo 'Generated markdown Table of Contents is out of date, please run "make markdown-toc" and commit the changes in this PR.' && exit 1)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,11 +42,16 @@ one that is needlessly restrictive and complex.
 ### Consistency Checks
 
 The Specification has a number of tools it uses to check things like style,
-spelling and link validity. You can run all of these locally via:
+spelling and link validity. Before using the tools, run:
 
 ```bash
-npm install # get the tools
-make check  # run all checks
+npm install
+```
+
+You can perform all checks locally using this command:
+
+```bash
+make check
 ```
 
 Note: This can take a long time as it checks all links. You should use this
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,7 +45,8 @@ The Specification has a number of tools it uses to check things like style,
 spelling and link validity. You can run all of these locally via:
 
 ```bash
-make check
+npm install # get the tools
+make check  # run all checks
 ```
 
 Note: This can take a long time as it checks all links. You should use this
@@ -92,11 +93,10 @@ wide. There are tools that can do it for you effectively. Please submit proposal
 to include your editor settings required to enable this behavior so the out of
 the box settings for this repository will be consistent.
 
-To check for style violations, use
+To check for style violations, run:
 
 ```bash
-make install-markdownlint # install markdownlint if not installed
-make markdownlint # run at the root of the repo
+make markdownlint
 ```
 
 To fix style violations, follow the
@@ -109,13 +109,14 @@ you can also use the `fixAll` command of the
 
 In addition, please make sure to clean up typos before you submit the change.
 
-To check for typos, use
+To check for typos, run the following command:
 
 ```bash
-# Golang is needed for the misspell tool.
-make install-misspell
 make misspell
 ```
+
+> **NOTE**: The `misspell` make target will also fetch and build the tool if
+> necessary. You'll need [Go](https://go.dev) to build the spellchecker.
 
 To quickly fix typos, use
 
--- a/Makefile
+++ b/Makefile
@@ -5,9 +5,6 @@ PWD := $(shell pwd)
 TOOLS_DIR := ./internal/tools
 MISSPELL_BINARY=bin/misspell
 MISSPELL = $(TOOLS_DIR)/$(MISSPELL_BINARY)
-MARKDOWN_LINK_CHECK=./node_modules/.bin/markdown-link-check
-MARKDOWN_LINT=./node_modules/.bin/markdownlint
-MARKDOWN_TOC=./node_modules/.bin/markdown-toc
 
 # see https://github.com/open-telemetry/build-tools/releases for semconvgen updates
 # Keep links in semantic_conventions/README.md and .vscode/settings.json in sync!
@@ -28,15 +25,10 @@ misspell:	$(MISSPELL)
 misspell-correction:	$(MISSPELL)
 	$(MISSPELL) -w $(ALL_DOCS)
 
-$(MARKDOWN_LINK_CHECK):
-	npm install markdown-link-check
-
 .PHONY: markdown-link-check
-markdown-link-check:	$(MARKDOWN_LINK_CHECK)
-	@for f in $(ALL_DOCS); do $(MARKDOWN_LINK_CHECK) --quiet --config .markdown_link_check_config.json $$f; done
-
-$(MARKDOWN_TOC):
-	npm install markdown-toc
+markdown-link-check:
+	@if ! npm ls markdown-link-check; then npm install; fi
+	@for f in $(ALL_DOCS); do npx markdown-link-check --quiet --config .markdown_link_check_config.json $$f; done
 
 # This target runs markdown-toc on all files that contain
 # a comment <!-- tocstop -->.
@@ -47,22 +39,21 @@ $(MARKDOWN_TOC):
 #   <!-- toc -->
 #   <!-- tocstop -->
 .PHONY: markdown-toc
-markdown-toc:	$(MARKDOWN_TOC)
+markdown-toc:
+	@if ! npm ls markdown-toc; then npm install; fi
 	@for f in $(ALL_DOCS); do \
 		if grep -q '<!-- tocstop -->' $$f; then \
 			echo markdown-toc: processing $$f; \
-			$(MARKDOWN_TOC) --no-first-h1 --no-stripHeadingTags -i $$f; \
+			npx markdown-toc --no-first-h1 --no-stripHeadingTags -i $$f; \
 		else \
 			echo markdown-toc: no TOC markers, skipping $$f; \
 		fi; \
 	done
 
-$(MARKDOWN_LINT):
-	npm install markdownlint-cli@0.31.0
-
 .PHONY: markdownlint
-markdownlint:	$(MARKDOWN_LINT)
-	@for f in $(ALL_DOCS); do echo $$f; $(MARKDOWN_LINT) -c .markdownlint.yaml $$f || exit 1;	done
+markdownlint:
+	@if ! npm ls markdownlint; then npm install; fi
+	@for f in $(ALL_DOCS); do echo $$f; npx markdownlint -c .markdownlint.yaml $$f || exit 1;	done
 
 .PHONY: install-yamllint
 install-yamllint:
@@ -101,5 +92,5 @@ fix: table-generation misspell-correction
 
 # Attempt to install all the tools
 .PHONY: install-tools
-install-tools: $(MISSPELL) $(MARKDOWN_LINT) $(MARKDOWN_LINK_CHECK) $(MARKDOWN_TOC)
+install-tools: $(MISSPELL) markdownlint markdown-link-check markdown-toc
 	@echo "All tools installed"
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,10 @@ misspell-correction:	$(MISSPELL)
 
 .PHONY: markdown-link-check
 markdown-link-check:
-	@for f in $(ALL_DOCS); do npx markdown-link-check --quiet --config .markdown_link_check_config.json $$f; done
+	@for f in $(ALL_DOCS); do \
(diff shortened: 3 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -92,11 +93,10 @@ wide. There are tools that can do it for you effectively. Please submit proposal
 to include your editor settings required to enable this behavior so the out of
 the box settings for this repository will be consistent.
 
-To check for style violations, use
+To check for style violations, run:
 
 ```bash
-make install-markdownlint # install markdownlint if not installed
```
chalin, line 98: Btw, the `install-markdownlint` target doesn't exist. It must have been removed at one point.

Makefile
```diff
@@ -24,15 +21,10 @@ misspell:	$(MISSPELL)
...
-	npm install markdown-link-check
-
 .PHONY: markdown-link-check
-markdown-link-check:	$(MARKDOWN_LINK_CHECK)
-	@for f in $(ALL_DOCS); do $(MARKDOWN_LINK_CHECK) --quiet --config .markdown_link_check_config.json $$f; done
-
-$(MARKDOWN_TOC):
-	npm install markdown-toc
+markdown-link-check:
+	@if ! npm ls markdown-link-check; then npm install; fi
```
chalin, line 26: If we always run `npm install` first before any check-tool targets, then we don't need this check. I added it to maintain backwards compatibility with the behavior of the make target as it was before this PR.

If you prefer, I can remove the `@if ...` line here and for the other two targets.

package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "devDependencies": {
+    "markdown-link-check": "^3.9.3",
+    "markdown-toc": "^1.2.0",
+    "markdownlint-cli": "0.31.0"
```
chalin, line 5: As I mentioned, the version of `markdownlint-cli` is pinned here, as was done in #2320.

jmacd (approved):

yurishkuro (approved):
❤️

Makefile
```diff
@@ -24,15 +21,10 @@ misspell:	$(MISSPELL)
...
+	@for f in $(ALL_DOCS); do npx markdown-link-check --quiet --config .markdown_link_check_config.json $$f; done
```
anuraaga, line 27: I think `npx` will install the latest version of the package if someone forgot to run `npm install`, ignoring `package.json`, while `npm run` would avoid this to guarantee the version used is what's in the `package.json`. Though alternatively, what do you think about removing `package.json` / `npm install` steps and adding the version in the `npx` command, e.g. `npx markdown-link-check@3.9.3`?

joaopgrassi: Another alternative to npx is adding a script in `package.json` to run the tool locally installed. 

```
"scripts": {
    "markdown-lint": "./node_modules/.bin/markdown-link-check"
  }
  ```
Then this line would contain `npm run markdown-lint` instead of `npx..`. 

Now that we have a `package.json`, it "unlocks" this alternative.

File: Makefile (hunk 2 above)
Comment (joaopgrassi) [reply Makefile:26]: What about failing/not running if it's not installed? That's much simpler, I guess? Now that there's `npm install` in the contributing guide, I'm not sure if it's necessary to install here. It does change the existing behavior so..

File: package.json [hunk 5]
```diff
@@ -0,0 +1,7 @@
+{
+  "devDependencies": {
```
joaopgrassi, line 2: I think we should also consider saving the [lock file](https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json), to make sure everyone gets the exact same dependencies?

joaopgrassi (commented):
This is great! I just left a few small things.

Makefile
chalin: > I think `npx` will install the latest version of the package if someone forgot to run `npm install`, ignoring `package.json`,

According to my tests, unless you specify an explicit package version on the command line, `npx` will install the package version specified in `packages.json`.

> what do you think about removing `package.json` / `npm install` steps and adding the version in the `npx` command, e.g. `npx markdown-link-check@3.9.3`?

IMHO, the proper place to specify a package version is in the `packages.json` file rather than the command line. Also, `npx markdownlint` doesn't work because we need it's CLI package too:

```console
$ npx markdownlint
npm ERR! could not determine executable to run
```

So I think that the cleanest and most standard solution is to have a `packages.json` file to specify the packages used by this repo.

package.json
chalin: For the use that we're making of these packages, I'd vote against saving the lock file. It is too much of a pain to deal with merge conflicts.

Makefile
chalin: Yes, that would be simpler, and it would be the solution I prefer. I've pulled out two of the three `@if ! npm ls foo; then npm install; fi` lines, because two of them were actually unnecessary (I wasn't aware that `npx` would install a package if it wasn't already installed). I've kept the line for `markdownlint` because `npx` fails otherwise (because it doesn't know about the CLI package).

package.json
joaopgrassi: Sure. But then, it would make sense to pin the versions of all packages, no (like done for `markdownlint-cli`)? Otherwise new versions can be downloaded and we can't know the effect, specially in CI.

Makefile
chalin: @anuraaga - you're right, `npx` will fetch the latest. I prefer @joaopgrassi's suggestion of just requiring users to run `npm install`.

Makefile
```diff
@@ -24,15 +21,12 @@ misspell:	$(MISSPELL)
...
 .PHONY: markdown-link-check
-markdown-link-check:	$(MARKDOWN_LINK_CHECK)
-	@for f in $(ALL_DOCS); do $(MARKDOWN_LINK_CHECK) --quiet --config .markdown_link_check_config.json $$f; done
-
-$(MARKDOWN_TOC):
-	npm install markdown-toc
+markdown-link-check:
+	@for f in $(ALL_DOCS); do \
+		npx --no -- markdown-link-check --quiet --config .markdown_link_check_config.json $$f \
+			|| exit 1; \
```
chalin, line 28: Note that link check will now fail if the package isn't installed or on the first broken link that it finds.

package.json
chalin: Actually, I unpinned `markdownlint-cli` because `0.31.0` _is_ the latest version. I'd assume that the packages that we're using are following semver and so if there's a breaking change it'll be introduced in the next major version (or minor version in the case of `markdownlint-cli`), for which we'll need to update the package.json file.

Again, we're not building a binary, we're just checking markdown files. I don't think that we need to pin package versions (unless something is known to be broken). We can rely on semver working for us. My 2 cents.

joaopgrassi: Sure, fair enough. I just wanted to point out this, since I had my share of issues with npm packages not respecting that.

joaopgrassi (approved):
Looks good to me! I only have one concern about pinning package versions to avoid surprises (as we had a few days ago). But I'll leave that to others to consider as well.

package.json
```diff
@@ -0,0 +1,7 @@
...
+    "markdown-link-check": "^3.9.3",
+    "markdown-toc": "^1.2.0",
+    "markdownlint-cli": "^0.31.0"
```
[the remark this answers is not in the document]
chalin: > ... change is backwards compatible to the library itself, it's not for consumers ...

Oh, that totally makes sense. Ok, I've pinned it for `markdownlint-cli`.

.github/workflows/checks.yaml
```diff
@@ -14,7 +14,9 @@ jobs:
       uses: actions/checkout@v2
 
     - name: run markdownlint
-      run: make markdownlint
+      run: |
+        npm install
```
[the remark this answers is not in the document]
chalin: > Does it make sense to add a separate step to install the dependencies?

I'm not 100% sure that I understand (the scope/context of) your question, and I'm not an expert on GH workflows, but this is how I've seen other projects call `npm install`.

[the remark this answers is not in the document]
chalin: > Also, not sure it makes sense for this small thing, but we could consider caching the packages https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows

Thanks for the link. That would make sense. Maybe this optimization can be for a followup PR?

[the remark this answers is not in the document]
chalin: > Does it make sense to add a separate step to install the dependencies?

Done

chalin:
@joaopgrassi - thanks for the feedback! I've made the changes that you suggested. PTAL

.github/workflows/checks.yaml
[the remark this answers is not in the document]
joaopgrassi: Yes, sounds good 👍

.github/workflows/checks.yaml
```diff
@@ -13,6 +13,9 @@ jobs:
     - name: check out code
       uses: actions/checkout@v2
 
+    - name: npm install
```
joaopgrassi, line 16: I would call it something like "Install check dependencies" or just "Install dependencies"

package.json
joaopgrassi: Maybe we can pin all of them? Better be safe than random build failures? 🤣

joaopgrassi (approved):

.github/workflows/checks.yaml
joaopgrassi: This is the text that is shown in the Actions page
image

chalin: Done

joaopgrassi (approved):
LGTM

joaopgrassi:
Hi @chalin, I think this is ready right? @anuraaga maybe you can take another look? Would be a pity if this is closed due to inactivity

CONTRIBUTING.md
```diff
@@ -92,11 +93,10 @@ wide. There are tools that can do it for you effectively. Please submit proposal
...
-make markdownlint # run at the root of the repo
+make markdownlint
```
Oberon00, line 99: No npm install required here?

CONTRIBUTING.md
```diff
@@ -45,7 +45,8 @@ The Specification has a number of tools it uses to check things like style,
 spelling and link validity. You can run all of these locally via:
 
 ```bash
-make check
+npm install # get the tools
+make check  # run all checks
```
Oberon00, line 49: Can't `make check` still do that automatically, i.e. can we make `make` call `npm install` if required (e.g. by checking for presence of node_modules subfolder or calling npm install with options that cause it to not upgrade already installed stuff)?

chalin: Yes it _can_, but part of the point of this PR has been to separate concerns by having NPM package installation be consolidated (with packages listed in `packages.json`), with installation purposefully being a separate step from checks.

(Also, IMHO, eventually many of the Make targets could be replaced by NPM scripts. But that's a topic for another day.)

CONTRIBUTING.md
chalin: I just submitted a commit that adjust the wording from earlier, making it clear that `npm install` needs to be run first before all checks. PTAL

CONTRIBUTING.md
joaopgrassi: I think with the latest change in `CONTRIBUTING.md` is clear enough that one should first install the packages in order to run the checks. This also removes some complexity in the make file.

Makefile
```diff
@@ -28,15 +25,12 @@ misspell:	$(MISSPELL)
...
-	npm install markdown-link-check
-
 .PHONY: markdown-link-check
-markdown-link-check:	$(MARKDOWN_LINK_CHECK)
-	@for f in $(ALL_DOCS); do $(MARKDOWN_LINK_CHECK) --quiet --config .markdown_link_check_config.json $$f; done
-
-$(MARKDOWN_TOC):
-	npm install markdown-toc
+markdown-link-check:
+	@for f in $(ALL_DOCS); do \
```
yurishkuro, line 30: Not a ding on this PR, but I found it extremely annoying that the checks are done in this manner, one file at a time (paying the cost of invoking the whole npx chain every time). I think most of the tools would accept a list of files, which would make the linting faster and less verbose.

yurishkuro (approved):

CONTRIBUTING.md
Oberon00: As a user of the spec build (which includes most editors of the spec), this separation of concerns does not add any value for me, does it?
And isn't it an implementation detail whether some tool is installed via npm or some other means? The spec is not a Node.js project, so it's unreasonable to assume all supporting tools will be installed via npm.

Is it possible that this PR makes work easier for maintainers of the spec build, but makes work (ever so slightly) harder for everybody else working on the spec?

yurishkuro: @Oberon00 what would you suggest? If a casual contributor does not have a Node.js installation, then we wouldn't be able to fix that via the Makefile.

As a side note, it would be cool to enable GH Workspaces (or whatever it's called), so that casual contributors could propose PRs directly from a web interface using a VM with prebuilt tooling env.

chalin: @Oberon00 - I'm not against preserving the old behavior, which is to allow a dev to invoke any make target without having to run `npm install` beforehand. In fact, that's how I had it setup originally, e.g.:

```make
markdown-link-check:
	@if ! npm ls markdown-link-check; then npm install; fi
	...
```

For details, see this https://github.com/open-telemetry/opentelemetry-specification/pull/2323#discussion_r801084679.

> As a user of the spec build (which includes most editors of the spec), this separation of concerns does not add any value for me, does it? ... Is it possible that this PR makes work easier for maintainers of the spec build, but makes work (ever so slightly) harder for everybody else working on the spec?

IMHO, the most important contribution in this PR is the introduction of `packages.json` to replace lines like these: https://github.com/open-telemetry/opentelemetry-specification/blob/21a3191bced91e5de2396784951b9e343a1d02c0/Makefile#L60-L61

It's better to list packages and their versions (pinned or not) in a `packages.json` file so that standard tools can process them and, for example, report vulnerabilities and security issues. As you might know, there exists bots to do that automatically, scanning repos and opening issues when packages need updating -- such as [dependabot](https://github.com/dependabot).

To me its 50/50 whether a make target installs dependencies if they aren't already installed, as long as the command used is `npm install` -- i.e., that we install packages as specified in the `packages.json` file.

Up to y'all to decide. Let me know.

chalin: > As a side note, it would be cool to enable GH Workspaces

@yurishkuro et all: have you tried GitPod.io? You can open this repo using the following link, https://gitpod.io/#https://github.com/open-telemetry/opentelemetry-specification, without any further configuration and it will immediately run `make`.

I just pushed a commit (https://github.com/open-telemetry/opentelemetry-specification/pull/2323/commits/21d4389c78890b0ac0183f70cfb8dc5ad0837544) to ensure that if/when this PR is merged, then we can continue using GitPod.io without custom settings (not that it would be much work).

Makefile
chalin: (Agreed. FYI, I'm looking into an alternative tool. But that's possibly for another PR.)

chalin:
Hi all: I've restored the old target behavior by using install-checks. I see this as win-win, WDYT @Oberon00, @joaopgrassi,  @yurishkuro @carlosalberto and others? While conceptually less clean, this addresses the concerns that @Oberon00 raised. My hope is that we can now have enough of a consensus to get this PR merged.

joaopgrassi (approved):
Both approaches are fine to me. I'm no expert in Make, but the build passes and I don't see anything clearly wrong. 👍

CONTRIBUTING.md
```diff
@@ -42,7 +42,13 @@ one that is needlessly restrictive and complex.
 ### Consistency Checks
 
 The Specification has a number of tools it uses to check things like style,
-spelling and link validity. You can run all of these locally via:
+spelling and link validity. Before using the tools, run:
+
+```bash
+npm install
```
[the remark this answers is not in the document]
chalin: I'd leave this as the recommended workflow.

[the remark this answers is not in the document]
chalin: @joaopgrassi - if you agree, could yo mark this as resolved?

chalin:
Thanks for the feedback @joaopgrassi.
Is this ready to merge, before the bot closes it!?! 😉 😄

CONTRIBUTING.md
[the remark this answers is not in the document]
joaopgrassi: Yes, sounds good. I can't resolve it though, so feel free to :)

Oberon00 (approved):

yurishkuro:
I am resolving discussions and merging. Feel free to open new tickets if changes are needed, but in the current state the PR is an overall improvement.

tigrannajaryan:
@chalin  After this change I can no longer do `make markdownlint` on my Mac. I get:

```
$ make markdownlint
/Users/tnajaryan/work/repos/opentelemetry-specification
└─┬ markdownlint-cli@0.31.0
  └── markdownlint@0.25.1 

./CHANGELOG.md
npx: installed 203 in 6.874s
npx: command not found: markdownlint-cli
make: *** [markdownlint] Error 1
```

Any advice on what I need to do to make this work locally?

yurishkuro:
Indeed - why did CI pass then?

It works for me after this patch:
```diff
diff --git a/Makefile b/Makefile
index e0d64e4..755e8e6 100644
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,7 @@ markdownlint:
        @if ! npm ls markdownlint; then npm install; fi
        @for f in $(ALL_DOCS); do \
                echo $$f; \
-               npx --no -p markdownlint-cli markdownlint -c .markdownlint.yaml $$f \
+               npx --no -p markdownlint markdownlint -c .markdownlint.yaml $$f \
                        || exit 1; \
        done
```

however it seems to be re-doing npm install for each file:
```
./specification/compatibility/opencensus.md
npx: installed 203 in 4.611s
./specification/compatibility/openmetrics.md
npx: installed 203 in 4.363s
./specification/compatibility/opentracing.md
npx: installed 203 in 4.437s
./specification/context/api-propagators.md
npx: installed 203 in 4.312s
./specification/context/context.md
```

yurishkuro:
I suggest we revert this PR until it can be properly fixed

chalin:
Checking. Yes, adding an `.nvmrc` is a good idea. I can do that if you'd like.

chalin:
Indeed, possibly a Node version issue.

> @chalin After this change I can no longer do `make markdownlint` on my Mac. I get:
> 
> ```
> $ make markdownlint
> /Users/tnajaryan/work/repos/opentelemetry-specification
> └─┬ markdownlint-cli@0.31.0
>   └── markdownlint@0.25.1 
> 
> ./CHANGELOG.md
> npx: installed 203 in 6.874s
> npx: command not found: markdownlint-cli
> make: *** [markdownlint] Error 1
> ```
> 
> Any advice on what I need to do to make this work locally?

It works fine for me locally. I'm using the latest LTS version of Node:

```console
$ nvm current
v16.13.0
```

You?

yurishkuro:
Confirmed, works w/ v16, but not with v14.

tigrannajaryan:
v16 works for me too.

PR #2325, opened 2022-02-08
Clarify what does wildcard mean in the Metrics SDK View

reyang:
Discussed in the [2/28/2022 Specification SIG Meeting](https://docs.google.com/document/d/1pdvPeKjA8v8w_fGKAN68JjWBmVJtPCpqdi9IZrd6eEo/edit?pli=1#).

* Clarified what does wildcard mean.

### DIFF
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -148,9 +148,9 @@ are the inputs:
 * The Instrument selection criteria (required), which covers:
   * The `type` of the Instrument(s) (optional).
   * The `name` of the Instrument(s). [OpenTelemetry SDK](../overview.md#sdk)
-    authors MAY choose to support wildcard characters, with the
-    question mark (`?`) matching exactly one character and the asterisk
-    character (`*`) matching zero or more characters.
+    authors MAY choose to support wildcard characters, with the question mark
+    (`?`) matching exactly one character and the asterisk character (`*`)
+    matching zero or more characters.
   * The `name` of the Meter (optional).
   * The `version` of the Meter (optional).
   * The `schema_url` of the Meter (optional).
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -148,7 +148,7 @@ are the inputs:
 * The Instrument selection criteria (required), which covers:
   * The `type` of the Instrument(s) (optional).
   * The `name` of the Instrument(s). [OpenTelemetry SDK](../overview.md#sdk)
-    authors MAY choose to support `name` that contains wildcards, with the
+    authors MAY choose to support wildcard characters, with the
     question mark (`?`) matching exactly one character and the asterisk
     character (`*`) matching zero or more characters. In addition, the
     OpenTelemetry SDK authors MAY choose to support additional predicates that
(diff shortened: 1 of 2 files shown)


specification/metrics/sdk.md
```diff
@@ -147,7 +147,12 @@ are the inputs:
 
 * The Instrument selection criteria (required), which covers:
   * The `type` of the Instrument(s) (optional).
-  * The `name` of the Instrument(s), with wildcard support (optional).
+  * The `name` of the Instrument(s). [OpenTelemetry SDK](../overview.md#sdk)
+    authors MAY choose to support `name` that contains wildcards, with the
+    question mark (`?`) matching exactly one character and the asterisk
+    character (`*`) matching zero or more characters. In addition, the
+    OpenTelemetry SDK authors MAY choose to support additional predicates that
+    are idiomatic for their language (e.g. regular expressions, callbacks).
```
Aneurysm9, line 155: I'm not sure that the part starting at "In addition," belongs here.  It seems like the same could apply to other selection criteria as well.  It might fit better in the paragraph below stating that "authors MAY choose to support more criteria."

specification/metrics/sdk.md
```diff
@@ -147,7 +147,12 @@ are the inputs:
 
 * The Instrument selection criteria (required), which covers:
   * The `type` of the Instrument(s) (optional).
-  * The `name` of the Instrument(s), with wildcard support (optional).
+  * The `name` of the Instrument(s). [OpenTelemetry SDK](../overview.md#sdk)
+    authors MAY choose to support `name` that contains wildcards, with the
```
ocelotl, line 151:
```suggestion
    authors MAY choose to support wildcards, with the
```

specification/metrics/sdk.md
ocelotl, line 155:
```suggestion
    character (`*`) matching zero or more characters.
```
I agree with @Aneurysm9. That sentence is better placed as another example following `... authors MAY choose to support more criteria.`. Can we number these requirements so that it is easier to refer to them?

reyang: Good point, I've updated the PR.

reyang: I've updated the PR.

ocelotl (approved):

aabmass (approved):

Aneurysm9 (approved):

cijothomas (approved):

alanwest (approved):

jack-berg (approved):

ocelotl:
Hey, don't we need also an escape character to be defined here? What if a name includes `*` or `?`?

reyang:
> Hey, don't we need also an escape character to be defined here? What if a name includes * or ??

Instrument names have [restrictions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument), `*` and `?` are both illegal characters.

jmacd (approved):

jmacd:
If a language provides a shell `Glob()` function that is readily available, and implements the two characters specified here (? and *) but also provides _more matching behavior_, is an implementation permitted to use the existing matcher function instead of implementing exactly what is specified here? 

Otherwise, we should give an algorithm for implementing this matcher. 😁

Aneurysm9:
> Otherwise, we should give an algorithm for implementing this matcher. 😁

Is this sufficient?

```perl
$input = "foo*bar?"
$input =~ s/\?/./g
$input =~ s/\*/.*/g
matchRe = /^$input$/ # /^foo.*bar.$/
```

MrAlias (approved):

PR #2344, opened 2022-02-15
Add `make all` target to test changes locally

tigrannajaryan:
Make all performs roughly the same steps as the already existing github actions. The only
exception is yamllint which I was not able to make work on my Mac after quite some time
and had to skip. If anyone knows how to make it work on a Mac please submit a follow-up PR.

Note that the checks can run in parallel to speed up, e.g. `make -j4`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,10 @@ MARKDOWN_TOC=./node_modules/.bin/markdown-toc
 # Keep links in semantic_conventions/README.md and .vscode/settings.json in sync!
 SEMCONVGEN_VERSION=0.8.0
 
+# TODO: add `yamllint` step to `all` after making sure it works on Mac.
+.PHONY: all
+all: markdownlint markdown-link-check misspell table-check schema-check
+
 $(MISSPELL):
 	cd $(TOOLS_DIR) && go build -o $(MISSPELL_BINARY) github.com/client9/misspell/cmd/misspell
 


carlosalberto (approved):

carlosalberto:
Oh, nice, I was meant to have something like, thanks!

Makefile
```diff
@@ -13,9 +13,16 @@ MARKDOWN_TOC=./node_modules/.bin/markdown-toc
 # Keep links in semantic_conventions/README.md and .vscode/settings.json in sync!
 SEMCONVGEN_VERSION=0.8.0
 
+.PHONY: all
```
yurishkuro, line 16: add TODO to add yamllink?

Makefile
```diff
@@ -13,9 +13,16 @@ MARKDOWN_TOC=./node_modules/.bin/markdown-toc
...
+all: markdownlint markdown-link-check markdown-toc validate-markdown-toc misspell table-check schema-check
+
 $(MISSPELL):
 	cd $(TOOLS_DIR) && go build -o $(MISSPELL_BINARY) github.com/client9/misspell/cmd/misspell
 
+.PHONY: validate-markdown-toc
+validate-markdown-toc:
```
yurishkuro, line 23: I don't think this makes sense as a standalone step, only if run after `markdown-toc`

Makefile
yurishkuro: also, it looks like it's the only linter in Python, whereas there is a Node.js version https://www.npmjs.com/package/yaml-lint

Makefile
arminru: The CI runs one after the other: https://github.com/open-telemetry/opentelemetry-specification/blob/9ffad6653d7d81dcec359867b835e28b0b37362c/.github/workflows/checks.yaml#L48-L52
It could be surprising if the step called "validate" would alter the files locally but if we call it `generate-and-validate` or the like it should be fine to have it pull in the generate step.

arminru (approved):

Makefile
tigrannajaryan: TODO added.

Makefile
tigrannajaryan: Right, I just don't know what I am doing. Removed the markdown-toc steps from `all` target for now. I agree it can be surprising if it changes files. Not what you expect from a default `make` execution.

codeboten (approved):

yurishkuro (approved):

PR #2348, opened 2022-02-15
Clarify how classification by attribute value and presence works

tigrannajaryan:
This topic was discussed verbally multiple times (particularly when we were
considering whether the "component" attribute should be borrowed from
OpenTracing semantic conventions). 

This PR captures previous discussions that are not explicitly written anywhere in the spec.

### DIFF
--- a/specification/common/attribute-classification.md
+++ b/specification/common/attribute-classification.md
@@ -1,0 +1,117 @@
+# Classification via Attributes
+
+**Status**: [Experimental](../document-status.md)
+
+From Wikipedia:
+> Classification is a process related to categorization, the process in which
+> ideas and objects are recognized, differentiated and understood.
+
+Attributes are sometimes recorded to indicate the class of the entity that is
+emitting the telemetry or the class of the activity that is reported by
+telemetry, in other words to _classify_ the entity or the activity.
+
+There are usually 2 ways to classify using attributes, described below.
+
+## Classification Via Attribute Value
+
+The first classification approach is to record the class as a value of some
+attribute.
+
+For example we can record an attribute named `component` with a value of
+`database` (`component=database` for short) to indicate that the telemetry is
+from a database and `component=httpserver` to indicate that the telemetry is
+from an HTTP server. This approach allows to classify via the _value_ of the
+attribute.
+
+Note the limitation of this approach: it classifies across a dimension and we
+need to give this dimension a clear name (which is the attribute name) and allow
+recording only a single fact across this dimension (because each attribute can
+have only one value). In our example if we wanted to also record the fact that
+the emitting component is a VM (and not a physical host), we cannot use
+`component=vm` although it may be reasonable to say that a VM is a type of a
+component. With this approach when a database is running on a VM we simply
+cannot record this fact because we have to simultaneously have both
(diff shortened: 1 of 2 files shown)


specification/common/attribute-classification.md
```diff
@@ -0,0 +1,117 @@
...
+`db.system` participates both in classification by presence and in
+classification by value (it records 2 facts: that the entity is a database and
+the fact that the database type is PostgreSQL).
+
+Using `db.system=postgresql` for classification by value is a correct approach
+because _there is no further information_ necessary to be recorded specifically
+for PostgreSQL. If _there was such a need_, for example if we also needed to
+record the PostgreSQL version number then instead we would choose to record the
+fact that the database is a PostgreSQL by presence of for example attribute
+`db.postgres.version` (in which case we no longer need to store "postgresql" in
```
zenmoto, line 101: I think that it's important that a sub-type of something can act as a member of that type. Postgres is a great example- if a user builds a database dashboard that shows the health of all databases in a system, she shouldn't need to sniff out each database individually. A Postgres-specific dashboard can rely on all sorts of Postgres attributes, but that should be in addition to rather than instead of. Similarly Greenplum or TimescaleDB can provide all of the Postgres attributes plus whatever attributes make sense for their specialized use case.

carlosalberto: > A Postgres-specific dashboard can rely on all sorts of Postgres attributes, but that should be in addition to rather than instead of

+1 to this one.

carlosalberto (approved):

carlosalberto:
Thanks for documenting this (much discussed, never written down, as you mentioned) topic :)

specification/common/attribute-classification.md
tigrannajaryan: @zenmoto you have a good point. I need to think about I can rewrite the recommendations to take this into account.

specification/common/attribute-classification.md
```diff
@@ -0,0 +1,117 @@
...
+entity, so we introduce an attribute the presence of which classifies the entity
+as a database, e.g. `db.system`. The presence of this attribute indicates that
+we have a database.
+
+Next, we want to classify the type of the database, i.e. whether it is a
+PostgreSQL or MySQL. We can record the database type as the value of the
+attribute `db.system=postgresql`. Note that in this case one attribute
+`db.system` participates both in classification by presence and in
+classification by value (it records 2 facts: that the entity is a database and
+the fact that the database type is PostgreSQL).
```
alanwest, line 94: > it records 2 facts: that the entity is a database

I'm interpreting "entity" to mean the thing producing telemetry. If this is what you mean, then I think this statement could lead to confusion because, in practice, telemetry that includes a `db.system` attribute is not necessarily a database - it might just be a web server interacting with a DB.

Thus far, my sense is that most of the semantic conventions defined so far have not been successful in classifying the type of entity producing telemetry but rather the conventions classify the context in which telemetry was produced (e.g., "I'm an operation against a DB system" or "I'm a span that was generated by a service running on a VM in AWS").

tigrannajaryan:
I am converting this to a draft since I don't think it is ready yet, I see gaps that need to be addressed first.

PR #2351, opened 2022-02-15
[editorial] Add space in schemas/overview

chalin:
For context, see #2320 at https://github.com/open-telemetry/opentelemetry-specification/pull/2320/files#r807334158.
/cc @arminru

Btw, wouldn't be better to use code markdown (like `Resource*`) rather than having to escape `*` when used as plain text?

### DIFF
--- a/specification/schemas/overview.md
+++ b/specification/schemas/overview.md
@@ -239,7 +239,7 @@ field in the messages:
 - The schema_url field in the InstrumentationLibraryLogs message applies to the
   contained LogRecord messages.
 
-- If schema_url field is non-empty both in Resource\*message and in the
+- If schema_url field is non-empty both in Resource\* message and in the
   contained InstrumentationLibrary\* message then the value in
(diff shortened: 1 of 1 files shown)


tigrannajaryan (approved):

arminru (approved):
Thanks, @chalin! And yes, when referring to the name of the message as it's defined in https://github.com/open-telemetry/opentelemetry-proto, using `code` formatting would make sense. It would be nice if you could change it accordingly or otherwise I can take it as well.

carlosalberto (approved):

chalin:
> Thanks, @chalin! And yes, when referring to the name of the message as it's defined in https://github.com/open-telemetry/opentelemetry-proto, using `code` formatting would make sense. It would be nice if you could change it accordingly or otherwise I can take it as well.

Ok, how about we leave this PR as is. I have other higher priority issues to tackle at the moment, but I'll make the changes you suggest if/when I touch this file again.

PR #2353, opened 2022-02-16
Add Resource browser attributes, clarify device attributes as client-side only

martinkuba:
This is the same as https://github.com/open-telemetry/opentelemetry-specification/pull/2115, except it only adds `browser` attributes and the clarification about `device` attributes as client-side only. It does not include the addition of `app` attributes as in https://github.com/open-telemetry/opentelemetry-specification/pull/2115.

The goal of this PR is to define which attributes will be used to identify telemetry as client-side (the type). These attributes are NOT intended to describe the client-side application (name etc.).


specification/resource/semantic_conventions/browser.md
```diff
@@ -0,0 +1,20 @@
...
+**type:** `browser`
+
+**Description**: The web browser in which the application represented by the resource is running. The `browser.*` attributes MUST be used only for resources that represent applications running in a web browser. The presence of these attributes is intended to identify client-side telemetry.
+
+<!-- semconv device -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `browser.name` | string | The web browser name | `Chrome` | No |
+| `browser.version` | string | The web browser version | `90` | No |
+| `browser.platform` | string | The operating system | `Windows` | No |
```
arminru, line 14: This duplicates the `os.*` attributes defined here:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/os.md
You could instead reference those in the text to make instrumenters and consumers aware that this is relevant if one is interested in the `browser.*` attributes.

specification/resource/semantic_conventions/browser.md
```diff
@@ -0,0 +1,20 @@
...
+
+**type:** `browser`
+
+**Description**: The web browser in which the application represented by the resource is running. The `browser.*` attributes MUST be used only for resources that represent applications running in a web browser. The presence of these attributes is intended to identify client-side telemetry.
+
+<!-- semconv device -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `browser.name` | string | The web browser name | `Chrome` | No |
+| `browser.version` | string | The web browser version | `90` | No |
```
arminru, line 13: There is some overlap with the process runtime attributes. Do you think these would work for this purpose as well?
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/process.md#process-runtimes

specification/resource/semantic_conventions/browser.md
martinkuba: It is true that the platform could be represented using the `os.*` attributes. I was trying to keep the browser values together since typically they come from the same place (user-agent headers or API).

specification/resource/semantic_conventions/browser.md
martinkuba: The browser attributes provide additional meaning - that the application is running within the context of a browser. We are also intending to use the presence of the browser attributes to classify the telemetry as coming from a browser application.

specification/resource/semantic_conventions/device.md
```diff
@@ -4,7 +4,7 @@
 
 **type:** `device`
 
-**Description**: The device on which the process represented by this resource is running.
+**Description**: The device on which the process represented by this resource is running. The `device.*` attributes MUST be used only for resources that are not part of a private infrastructure, for example a mobile or IoT device.
```
arminru, line 7: A mobile phone is a private device as well usually.
What are you trying to achieve? Maybe something like "_MUST only be used for devices that are expected to be used directly by human end-users_" or the like could work to limit it to RUM-related data.

specification/resource/semantic_conventions/browser.md
arminru: If there would be two different data sources for retrieving this information (e.g. if you were running on a JVM or had access to the file system) it could make sense to report both separately but if the data is retrieved only from the browser anyway, I think using the existing attribute makes more sense. This would also allow existing filters and such to be used without having to extend these to take both attributes into account. WDYT?

specification/resource/semantic_conventions/browser.md
arminru: This could also be achieved by using the existing attributes `process.runtime.name` + `process.runtime.version` and only adding an additional optional flag like `process.runtime.is_web_browser`.
Would that work for your use case?

If we expect more attributes to come then introducing a separate entity for `browser` makes sense but otherwise I think using the existing attributes as they fit perfectly instead of duplicating them would be preferable and make it less complex for telemetry consumers to use the data.
@open-telemetry/specs-approvers @open-telemetry/technical-committee, what do you think would be better?

dyladan: These attributes are also difficult to reliably determine a lot of the time outside of some user agent divination

specification/resource/semantic_conventions/browser.md
martinkuba: I am not against it; I only wanted to keep the attributes together for clarity. Do you think it should be documented somewhere that a browser-based SDK should capture the user-agent platform value in `os.platform`?

specification/resource/semantic_conventions/browser.md
martinkuba: The [`process.runtime.name`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/process.md#javascript-runtimes) value is already specified to be `browser` in the browser context (not the name of the browser). This means that we would not need the additional `is_web_browser` attribute, but we still need to capture the details about the browser vendor and version.

The example listed for `process.runtime.version` is a user-agent string. It seems strange to me to use a "version" attribute to be used for something else than just a version. I think it would be more clear to have a `user_agent` attribute if that's all that is available, and `name` + `version` attributes if the instrumentation can provide it.

specification/resource/semantic_conventions/browser.md
```diff
@@ -0,0 +1,21 @@
...
+| `browser.platform` | string | The operating system | `Windows` | No |
+| `browser.mobile` | boolean | Flag indicating a mobile device | | No |
+| `browser.user_agent` | string | Full user-agent string provided by the browser [1] | `'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/95.0.4638.54 Safari/537.36'` | No |
```
MSNev, line 16: Should also include details about the "brands" and other client side APIs (not just the SEC-CH-* headers which are only sent if requested)
https://developer.mozilla.org/en-US/docs/Web/API/User-Agent_Client_Hints_API

specification/resource/semantic_conventions/browser.md
scheler: I think we should keep the `browser.*` namespace to reliably determine the resource is a client-side resource. While process.runtime sounds like it suits the situation it's still good to be explicit with `browser` namespace as the people building browser applications don't relate so much to calling the browser as a `process`, so it will only be confusing.

specification/resource/semantic_conventions/browser.md
scheler: Again, we should use the terminology that's more common. Although many consider browser as an OS, the browser runs on an OS as well. The `os` namespace uses `name` that's equivalent to `browser.platform`, but for browser apps we could use the same terminology as used in the UserAgent Client-Hints [spec](https://wicg.github.io/ua-client-hints/), which uses the term `platform`.

specification/resource/semantic_conventions/browser.md
```diff
@@ -0,0 +1,21 @@
...
+
+All of these attributes can be provided by the user agent itself in the form of an HTTP header (e.g. Sec-CH-UA, Sec-CH-Platform, Sec-CH-Mobile, User-Agent). However, in order to support batches of data with different resources, these attributes should be used when possible.
+
+**[1]:** The user-agent value should be provided only from browsers that do not have a mechanism to retrieve name, version, and platform individually (for example from the User-Agent Client Hints API).
+<!-- endsemconv -->
```
scheler, line 21: @martinkuba how about dropping the user_agent completely from day 1 and have the SDK generate the individual client-hints from the user-agent string and provide them?

specification/resource/semantic_conventions/device.md
scheler: I agree, the term `private` is confusing here.  I suggest dropping the term human as well and use just end-users. So, how about:  'MUST be used only for resources closer to the end-users, such as mobile and IoT devices'.

specification/resource/semantic_conventions/browser.md
martinkuba: Client hints are not yet supported by all browsers, and even if they are, we still need a way to identify old browser versions that will be around for a long time. I think we should keep user-agent attribute as a fallback.

scheler: @martinkuba I only meant using the Client hints terminology. So for older browsers you use user-agent to generate different attributes that are named based on CH terminology. Of course, only wherever applicable.

specification/resource/semantic_conventions/browser.md
dyladan: > FYI - Moving forward getting the browser version is going to become an asynchronous task as the UserAgent will get frozen based and the only way to reliably way get the version will be via `navigator.userAgentData.getHighEntropyValues(...)`

This is unfortunate. I was just looking the other day to see what it would take to remove the async nature of resources from javascript sdk because it makes startup extremely painful. If it's an async attribute maybe it better live as a span attribute?

martinkuba: > FYI - Moving forward getting the browser version is going to become an asynchronous task as the UserAgent will get frozen based and the only way to reliably way get the version will be via
navigator.userAgentData.getHighEntropyValues(...)

`userAgentData.brands` does include the major version of the browser. I think this should be sufficient for most cases. 

> This is unfortunate. I was just looking the other day to see what it would take to remove the async nature of resources from javascript sdk because it makes startup extremely painful. If it's an async attribute maybe it better live as a span attribute?

I wonder if the limitations of the API should drive the data model. IMO, these attributes belong on the Resource. If we really need to capture the high entropy values, could it be handled in the SDK somehow (delay assigning the resource or propose a mechanism to update the Resource)?

martinkuba:
@MSNev @dyladan I have updated the PR to match the low entropy user-agent client hints. These can be retrieved synchronously, and I think are useful for most common analysis.

@arminru @scheler I have updated the description for the `device` attributes.

specification/resource/semantic_conventions/browser.md
martinkuba: Parsing the user agent string is not straight-forward, and I don't think we should put that burden on the instrumentation.

arminru:
This is pending on the discussion in #2466.

specification/resource/semantic_conventions/browser.md
```diff
@@ -0,0 +1,21 @@
+# Browser
+
+**Status**: [Experimental](../../document-status.md)
+
+**type:** `browser`
+
+**Description**: The web browser in which the application represented by the resource is running. The `browser.*` attributes MUST be used only for resources that represent applications running in a web browser. The presence of these attributes is intended to identify client-side telemetry.
```
arminru, line 7:
```suggestion
**Description**: The web browser in which the application represented by the resource is running.
The `browser.*` attributes MUST be used only for resources that represent applications running in a web browser (regardless of whether running on a mobile or desktop device).
The presence of these attributes is intended to identify client-side telemetry.
```
Should be obvious but some might benefit from this clarification.

specification/resource/semantic_conventions/browser.md
```diff
@@ -0,0 +1,21 @@
...
+
+<!-- semconv device -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `browser.brands` | array of strings | Array of brand name and version separated by a space. [1] | `[' Not A;Brand 99','Chromium 99',Chrome 99']` | No |
+| `browser.platform` | string | The platform on which the browser is running | `Windows` | No |
```
arminru, line 13:
```suggestion
| `browser.platform` | string | The platform on which the browser is running | `Windows`, `iOS` | No |
```
Should this be taken from `navigator.platform` or parsed from the User Agent string if UA client hints are not available?

specification/resource/semantic_conventions/browser.md
```diff
@@ -0,0 +1,21 @@
...
+| `browser.user_agent` | string | Full user-agent string provided by the browser [2] | `'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/95.0.4638.54 Safari/537.36'` | No |
+
+All of these attributes can be provided by the user agent itself in the form of an HTTP header (e.g. Sec-CH-UA, Sec-CH-Platform, User-Agent). However, in order to support sending telemetry through the Collector, these attributes should be used when possible.
```
arminru, line 16: Also proxies and other intermediaries might drop such headers even before it reaches the collector, right?

arminru:
@martinkuba

> I can update the description of the PR and perhaps remove the part about device attributes?

I think the `device` part looks good and is small and related enough to be discussed in the same PR but you could also split this up if you prefer, sure.

specification/resource/semantic_conventions/browser.md
martinkuba: I specifically meant to support batch export from the collector. But yes, the headers could also be stripped before it even gets to the collector. I will update this statement.

specification/resource/semantic_conventions/browser.md
martinkuba: Good question, the values from `navigator.platform` are different than from `navigator.userAgentData.platform` (e.g. `MacIntel` vs `macOS`, `Win32` vs `Windows`). I would propose that the `browser.platform` attribute is only used with the modern UA client hints API. The `browser.user_agent` attribute can be used as a general fallback for browsers that do not support client hints.

specification/resource/semantic_conventions/browser.md
arminru: Ah with the batching part it's much clearer, thanks for adding this.

specification/resource/semantic_conventions/browser.md
arminru: You could explicitly mention that one should *only* use the UA hints and *not* use `navigator.platform` so that the values are consistent.

specification/resource/semantic_conventions/README.md
```diff
@@ -156,6 +156,7 @@ Attributes defining a running environment (e.g. Operating System, Cloud, Data Ce
 - Deployment:
   - [Deployment Environment](./deployment_environment.md)
   - [Kubernetes](./k8s.md)
+- [Browser](./browser.md)
```
arminru, line 159: Please make sure to add an entry to CHANGELOG.md for your changes.

specification/resource/semantic_conventions/browser.md
```diff
@@ -0,0 +1,23 @@
...
+| `browser.platform` | string | The platform on which the browser is running [2] | `Windows`, `macOS`, `Android` | No |
+| `browser.user_agent` | string | Full user-agent string provided by the browser [3] | `'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/95.0.4638.54 Safari/537.36'` | No |
+
+All of these attributes can be provided by the user agent itself in the form of an HTTP header (e.g. Sec-CH-UA, Sec-CH-Platform, User-Agent). However, the headers could be removed by proxy servers, and are tied to calls from individual clients. In order to support batching through services like the Collector and to prevent loss of data (e.g. due to proxy servers removing headers), these attributes should be used when possible.
+
+**[1]:** This value is intended to be taken from the [UA client hints API](https://wicg.github.io/ua-client-hints/#interface) (navigator.userAgentData.brands).
+
+**[2]:** This value is intended to be taken from the [UA client hints API](https://wicg.github.io/ua-client-hints/#interface) (navigator.userAgentData.platform).
```
arminru: Please add something along this to make clear why this attribute is (kind of) duplicated:

> Thus it might differ from the values that would be reported for `os.type` (link to https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/os.md#operating-system).

arminru: Is there any list of expected values? This would make it easier for consumer to interpret the data beyond just displaying it, if desired.

specification/resource/semantic_conventions/browser.md
```diff
@@ -0,0 +1,23 @@
...
+
+**[3]:** The user-agent value should be provided only from browsers that do not have a mechanism to retrieve brands and platform individually from the User-Agent Client Hints API. To retrieve the value, the legacy `navigator.userAgent` API can be used.
```
arminru, line 22:
```suggestion
**[3]:** The user-agent value SHOULD be provided only from browsers that do not have a mechanism to retrieve brands and platform individually from the User-Agent Client Hints API. To retrieve the value, the legacy `navigator.userAgent` API can be used.
```

specification/resource/semantic_conventions/browser.md
```diff
@@ -0,0 +1,23 @@
...
+| `browser.platform` | string | The platform on which the browser is running [2] | `Windows`, `macOS`, `Android` | No |
+| `browser.user_agent` | string | Full user-agent string provided by the browser [3] | `'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/95.0.4638.54 Safari/537.36'` | No |
+
+All of these attributes can be provided by the user agent itself in the form of an HTTP header (e.g. Sec-CH-UA, Sec-CH-Platform, User-Agent). However, the headers could be removed by proxy servers, and are tied to calls from individual clients. In order to support batching through services like the Collector and to prevent loss of data (e.g. due to proxy servers removing headers), these attributes should be used when possible.
+
+**[1]:** This value is intended to be taken from the [UA client hints API](https://wicg.github.io/ua-client-hints/#interface) (navigator.userAgentData.brands).
+
+**[2]:** This value is intended to be taken from the [UA client hints API](https://wicg.github.io/ua-client-hints/#interface) (navigator.userAgentData.platform). The legacy `navigator.platform` API should NOT be used in order for the values to be consistent.
```
arminru, line 20:
```suggestion
**[2]:** This value is intended to be taken from the [UA client hints API](https://wicg.github.io/ua-client-hints/#interface) (navigator.userAgentData.platform).
If unavailable, the legacy `navigator.platform` API SHOULD NOT be used instead and this attribute SHOULD be left unset in order for the values to be consistent .
```

specification/resource/semantic_conventions/browser.md
```diff
@@ -0,0 +1,23 @@
...
+**Status**: [Experimental](../../document-status.md)
+
+**type:** `browser`
+
+**Description**: The web browser in which the application represented by the resource is running. The `browser.*` attributes MUST be used only for resources that represent applications running in a web browser (regardless of whether running on a mobile or desktop device).
+
+<!-- semconv device -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `browser.brands` | array of strings | Array of brand name and version separated by a space [1] | `[' Not A;Brand 99','Chromium 99',Chrome 99']` | No |
```
MSNev, line 12: Question: As this is the raw data from the Browser where "Not A;Brand" is a derived value, and the spec defines this in a manner where it "may" be different for every user (in practice today it seems to be specific per version). Do we want to just pass on as it or required that this value is "decoded"?

Eg. for this specific case should it just be "Chrome 99" and the current dev version of Edge 
`[{"brand":" Not A;Brand","version":"99"},{"brand":"Chromium","version":"102"},{"brand":"Microsoft Edge","version":"102"}]` So in this case we only really need "Microsoft Edge 102"

specification/resource/semantic_conventions/browser.md
```diff
@@ -4,18 +4,20 @@
 
 **type:** `browser`
 
-**Description**: The web browser in which the application represented by the resource is running. The `browser.*` attributes MUST be used only for resources that represent applications running in a web browser. The presence of these attributes is intended to identify client-side telemetry.
```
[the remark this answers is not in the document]
martinkuba: I took it out because we are not clear yet on how to classify client-side telemetry (https://github.com/open-telemetry/opentelemetry-specification/issues/2466). We can put it back in a separate PR if we decide that's the way to go.

specification/resource/semantic_conventions/browser.md
martinkuba: Added a paragraph that clarifies this.

specification/resource/semantic_conventions/browser.md
martinkuba: The spec says that the brands list should be treated as a unit (compared to other known lists). I think maintaining the logic to parse this in the client could be a maintenance headache.

dyladan: I agree with martin here other than a very minor payload size penalty there's no reason to throw data away

dyladan (approved):
Seems fine to me. Using newer APIs means it will continue being relevant for a longer time than the old APIs even if coverage isn't great right now (missing in at least firefox and safari). The newer APIs are also clearer than the old user agent.

specification/resource/semantic_conventions/browser.md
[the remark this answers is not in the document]
arminru: Good point! I'll resolve this conversation then.

arminru (approved):
LGTM from a general spec/semconv POV. Thank you, @martinkuba!
I'll rely on the Client instrumentation/RUM SIG to asses the suitability of the proposed values and their sources.

MSNev (approved):

martinkuba:
@yurishkuro This PR needs one more approval before it can be merged. Can you please take a look?

semantic_conventions/resource/browser.yaml
```diff
@@ -0,0 +1,41 @@
...
+        running in a web browser (regardless of whether running on a mobile or desktop device).
+    attributes:
+      - id: brands
+        type: string[]
+        brief: 'Array of brand name and version separated by a space'
+        note: >
+          This value is intended to be taken from the
+          [UA client hints API](https://wicg.github.io/ua-client-hints/#interface)
+          (navigator.userAgentData.brands).
+        examples: [" Not A;Brand 99", "Chromium 99", "Chrome 99"]
```
spencerwilson, line 16: These example values are of type string, but the attribute is of type string[]. Should it be the following?
```suggestion
        examples: [[" Not A;Brand 99", "Chromium 99", "Chrome 99"]]
```

martinkuba: I thought so too at first, but I think the `type` determines how the example is formatted. All other existing attributes of type `string[]` have examples written the same way (e.g. [process.command_args](https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/resource/process.yaml#L59))

MSNev (approved):

carlosalberto:
@open-telemetry/specs-approvers Please review before we merge ;)

PR #2363, opened 2022-02-18
Add an option for multi-instrument callbacks

### LINKED ISSUE #2280: Enable multi-instrument callbacks in the metrics API

### LINKED ISSUE #2408: Clarify how should SDK invokes async instrument callbacks

In the change of https://github.com/open-telemetry/opentelemetry-specification/pull/2361, it defines that:

> Callback functions MUST be documented as follows for the end user:
> - Callback functions SHOULD be reentrant safe. The SDK expects to evaluate callbacks for each MetricReader independently.

It sounds like the SDK can invoke the callback functions _each time_ a MetricReader initiates collection. While implementing the behavior, I found that when the async instrument is configured with multiple views with different names, a created async instrument may be mapped into multiple underlying, exported async instruments. The problems are:

1. Should the SDK call the callback _each time_ a MetricReader initiates collection, or one concurrent invocation for each MetricReader independently?

Like, we have two readers, should the SDK call the callback when reader A initiates a second-time collection when the first collection is not timed out yet?

2. Should the SDK call the callback once for an end-user registered async instrument, or for a view-ed async instrument?

jmacd:
Fixes #2280.
Fixes #2408.

## Changes

Adds optional support for multi-instrument callbacks, which are a more efficient and practical way to report measurements from a single expensive method when more than one kind of measurements are involved. The "DIY" approach here is really difficult for programmers.


jmacd:
@open-telemetry/specs-metrics-approvers I consider this to be a high-priority gap in the API specification which is causing slowness in the OTel-Go repository. Please consider this an urgent request to update the API specification.  This is independent of stabilizing the SDK specification, but one reason we haven't provided an OTel-Go SDK to help stabilize the SDK specification is that we're trying to find an API we're all happy with first. Please review this PR.

specification/metrics/api.md
```diff
@@ -1108,6 +1115,47 @@ for the interaction between the API and SDK.
...
 * [`Attributes`](../common/common.md#attributes)
 
+### Multiple-instrument callbacks
+
+Where idiomatic, [the Metrics API supports an option to use multiple
+instruments from a single registered
+callback](#asynchronous-instrument).  The key requirement for the API
+is that the association between Callbacks and Instruments be explicit
+so that the SDK knows which callbacks to execute when observing
+specific instruments.
```
MadVikingGod, line 1125:
```suggestion
The Metrics API, if idiomatic, MAY support a way to use multiple instruments
from a single registered callback.  
If this API exisists it MUST explictly associate Callbacks with the Instruments
used in the callback.  This is because <insert the reason behind the key reuirement>
```

specification/metrics/api.md
```diff
@@ -1108,6 +1115,47 @@ for the interaction between the API and SDK.
...
+class Device:
+    """A device with two instruments"""
+
+    def __init__(self, meter, proeprty):
+        self.property = property
+        self.usage = meter.create_observable_counter(name="usage", description="count of items used")
+        self.pressure = meter.create_observable_gauge(name="pressure", description="force per unit area")
+
+        # Note the two associated instruments are passed to the callback.
+        meter.register_callback([self.usage, self.pressure], self.observe)
```
bogdandrutu, line 1151: Q: What does it happen if the instruments are not created via this Meter?

specification/metrics/api.md
```diff
@@ -1108,6 +1115,47 @@ for the interaction between the API and SDK.
...
+
+    def observe(self, result):
+        usage, pressure = expensive_system_call()
+        self.usage.observe(result, usage, {'property', self.property})
+        self.pressure.observe(result, pressure, {'property', self.property})
```
bogdandrutu, line 1156: why change the order "Observe" API is on the instrument instead on the result. Should we be consistent?

If we do on the `Instrument` that is a bit strange to have that API working only for `batch` but not for `single`.

bogdandrutu (commented):
Do you have a prototype of this?

jmacd:
> Do you have a prototype of this?

The original OTel-Go prototype includes a batch-observer API. So does the proposed replacement OTel-Go API in https://github.com/open-telemetry/opentelemetry-go/pull/2587.

I even think of the original prototype API as technically spec compliant, since a "single callback" is passed to each instrument constructor. Nothing in the current specification says that a callback can't observe multiple instruments. The spec just says "be idiomatic". I'm trying to clarify that "idiomatic" includes potentially the use of multiple instruments per calblack.

specification/metrics/api.md
jmacd: It would be an error at the time of registration and/or at the time of use. By the library guidelines, these would be handled and not interrupt program execution.

specification/metrics/api.md
```diff
@@ -1108,6 +1115,47 @@ for the interaction between the API and SDK.
...
+### Multiple-instrument callbacks
+
+Where idiomatic, [the Metrics API supports an option to use multiple
+instruments from a single registered
+callback](#asynchronous-instrument).  The key requirement for the API
+is that the association between Callbacks and Instruments be explicit
+so that the SDK knows which callbacks to execute when observing
+specific instruments.
+
+Some suggestions for the API author:
```
MadVikingGod, line 1127:
```suggestion
It is RECOMMENDED that the API authors use one of the following forms:
```

specification/metrics/api.md
```diff
@@ -1108,6 +1115,47 @@ for the interaction between the API and SDK.
...
+
+* the Observable result used in multi-instrument callbacks receives an
+  additional instrument argument
```
MadVikingGod, line 1130: I don't understand this bullet, but that might be a lack of familiarity with Java's implementation.

specification/metrics/api.md
jmacd: > why change the order "Observe" API is on the instrument instead on the result. Should we be consistent?

I didn't change the order, this is just a different suggestion than the one you're thinking of.  Two python examples are given for the simple case, one where a list of measurements is returned, and one where the "result" object is used to "observe" values. It's this second example that is inconsistent with the new example, but they're not required to go together.

```
def pf_callback(result):
    # Note: in the real world these would be retrieved from the operating system
    result.Observe(8,        ("pid", 0),   ("bitness", 64))
```

I don't think of the result as the one "doing" the observation in any case, I'd prefer to see that example written as "result.Add(observed_value, attrs...)" instead of "result.Observe(value, attrs...)" (i.e., "add a value to the result being observed").

> If we do on the Instrument that is a bit strange to have that API working only for batch but not for single.

Nothing I've written would prevent an instrument from being used in multiple callbacks, both single and multi-instrument forms. The fact that the two APIs have different shape is because one has an implicit instrument and one does not. You can create one instrument, register simple callbacks to provide some observations, and register batch callbacks to provide other observations.

In fact, the reason we're emphasizing this for OTel-Go is that the multi-instrument API is better than the simple API, and if possible we think having only a multi-instrument API makes sense. In the spirit of being idiomatic for Go, if you want to have a simpler API than the one provided, you'll have to write a few extra lines of code.

specification/metrics/api.md
jmacd: The "observable result" pattern is not named well above, it's this pattern:

```
def pf_callback(result):
  result.Observe(8, ("pid", 0), ("bitness", 64))
```

In OTel-Go we've been using the context.Context as a substitute for this "result" object.

MrAlias (approved):

specification/metrics/api.md
jmacd: Done. The <insert reason> has been clarified in the new `#asynchronous-instrument-api` section above.

specification/metrics/api.md
jmacd: I've bolded the definition of the observable result pattern where it first appears, to make it clearer at this point.

specification/metrics/api.md
jmacd: Done.

specification/metrics/api.md
jmacd: I've changed the order to match the observable result pattern in order, i.e., it's

```
  result.observe(instrument, value, attributes...)
```

specification/metrics/api.md
```diff
@@ -366,11 +375,28 @@ RECOMMENDATIONS is explicitly not specified at the API level.
...
-* Return a list (or tuple, generator, enumerator, etc.) of `Measurement`s.
-* Use an observable result argument to allow individual `Measurement`s to be
-  reported.
+* Return a list (or tuple, generator, enumerator, etc.) of individual
+  `Measurement` values.
+* Pass an *Observable Result* as a formal parameter of the callback,
+  where `result.Observe()` captures individual `Measurement` values.
+
+Callbacks registered at the time of instrument creation MUST apply to
+the single instruments which is under construction.
```
bogdandrutu, line 384: Idea to investigate with maintainers: Can we actually remove this limitation? If we extend the "Result" to have 2 methods/funcs one that accepts an instrument as initial argument one that it does not, and let the user call whichever API they need, and probably error somehow if they call the non-instrument version for a callback registered to multiple instruments?

I am not sure how to cc all maintainers to this comment, so I will add all some sigs here:

@open-telemetry/go-maintainers 
@open-telemetry/java-maintainers 
@open-telemetry/python-maintainers 
@open-telemetry/dotnet-maintainers 
@open-telemetry/cpp-maintainers 
@open-telemetry/erlang-maintainers 
@open-telemetry/javascript-maintainers

bogdandrutu, line 384: Idea to investigate with maintainers: Can we actually remove this limitation? If we extend the "Result" to have 2 methods/funcs one that accepts an instrument as initial argument one that it does not, and let the user call whichever API they need, and probably error somehow if they call the non-instrument version for a callback registered to multiple instruments?

I am not sure how to cc all maintainers to this comment, so I will add all some sigs here:

@open-telemetry/go-maintainers 
@open-telemetry/java-maintainers 
@open-telemetry/python-maintainers 
@open-telemetry/dotnet-maintainers 
@open-telemetry/cpp-maintainers 
@open-telemetry/erlang-maintainers 
@open-telemetry/javascript-maintainers

bogdandrutu (commented):
LGTM, except I would like to see if we can just simply allow exactly same callbacks signature for both cases, and remove the limitation that these needs to happen after initialization.

specification/metrics/api.md
bogdandrutu: After more discussions with @jmacd, we need maintainers of stable API to tell us/show a prototype how are they planning to extend existent stable API to support this. We need to make sure we don't make already stable API ugly because of this.

specification/metrics/api.md
```diff
@@ -1063,6 +1089,49 @@ for the interaction between the API and SDK.
...
+class Device:
+    """A device with two instruments"""
+
+    def __init__(self, meter, proeprty):
+        self.property = property
+        self.usage = meter.create_observable_counter(name="usage", description="count of items used")
+        self.pressure = meter.create_observable_gauge(name="pressure", description="force per unit area")
+
+        # Note the two associated instruments are passed to the callback.
+        meter.register_callback([self.usage, self.pressure], self.observe)
```
legendecas, line 1127: It seems to me that the callback registered with `meter.create_observable_counter` and the callback registered with `meter.register_callback` have different signatures. In that case, I'd find it is confusing to have them all named "callback"s.

jmacd: I have been trying to make a distinction between "callback functions" (code) and Callbacks (concept/entity). I 

I'm not sure if it's overly subjective or not, to dictate a single form of callback function here. I believe the concept/entity are identical, even if the form of the callback function is not.

I do know this: if forced to pick a single callback API, I would choose the multi-instrument one even if it means that single-instrument callback functions are slightly more difficult to use. The performance benefits outweigh the disadvantages. I consider this to be not a problem because generally amount of code devoted to asynchronous instruments is small (unlike synchronous instruments).

@bogdandrutu has given feedback on this "forced decision", I think his opinion is that if we accept a multi-instrument callback API as in the example here, we should find a delicate way to deprecate the first form. I don't feel so strongly. 🤷

legendecas: I have no strong opinion regarding the two forms of callbacks, forced to choose one or not. I just find that we have those two forms of callback named in a way that is hard to interpret or to be learned.

dashpole (approved):

specification/metrics/api.md
MrAlias: FWIW, this is already a part of the Go API.

specification/metrics/api.md
MrAlias: I don't think an implementation would use both forms. Meaning there won't be a confusion when this is implemented.

specification/metrics/api.md
```diff
@@ -337,19 +338,27 @@ The API to construct asynchronous instruments MUST accept the following paramete
...
 newly created instrument.
 
-The API SHOULD support registration of callback functions to
+The API SHOULD support registration of `callback` functions associated with
 asynchronous instruments after they are created.
 
-Where the API supports registration of callback functions after
+Where the API supports registration of `callback` functions after
 asynchronous instrumentation creation, it MUST return something (e.g.,
 a registration handle, receipt or token) to the user that supports
```
ocelotl, line 353: Isn't this too language specific? I imagine there can be other ways to do this unregistration that are more idiomatic for different languages than using a returned object.

specification/metrics/api.md
```diff
@@ -337,19 +338,27 @@ The API to construct asynchronous instruments MUST accept the following paramete
...
-undoing the effect of callback registation.
+undoing the effect of `callback` registation.
+
+A Callback is the conceptual entity created each time a `callback`
+function is registered through an OpenTelemetry API.  
+
+Every currently registered Callback associated with an instrument MUST
+be evaluted exactly once during collection prior to reading data for
+that instrument.
```
ocelotl, line 361: Some languages may not need to create anything when a callback function is registered, they will just call that function during collection.

specification/metrics/api.md
```diff
@@ -366,11 +375,28 @@ RECOMMENDATIONS is explicitly not specified at the API level.
...
-The API MUST treat observations from a single callback as logically
+Callbacks registered after the time of instrument creation MAY be
+associated with multiple instruments.
+
+Idiomatic APIs for multiple-instrument Callbacks MUST distinguish the
+instrument associated with each observed `Measurement` value.
+
+Multiple-instrument Callbacks MUST be associated at the time of
+registration with a declared set of asynchronous instruments from the
+same `Meter` instance.  This REQUIREMENT that Instruments be
```
ocelotl, line 394: REQUIREMENT is not an RFC2119 keyword.
```suggestion
same `Meter` instance.  This requirement that Instruments be
```

specification/metrics/api.md
tsloughter: I can't think of one, but I always support making the spec less language specific :), so here to suggest simply saying the user must be able to undo registration of the specific callback after its registration. This implies that some unique reference must be available -- since you can't just say 'deregister everything registered on this instrument' to achieve the same thing.

But also having a, "for example, this could be done by returning a registratio handle, receipt or token", is useful to have, instead of a `MUST`.

specification/metrics/api.md
legendecas: > I don't think an implementation would use both forms. Meaning there won't be a confusion when this is implemented.

If so, I think the spec should clearly state that implementation must only pick one of the patterns, but not do both.

specification/metrics/api.md
tsloughter: @MrAlias the restriction is part of the Go API? I can't understand the need for the restriction. Is there one in Go?

tsloughter:
Finally got around to reading this closely and what I was not expecting was code like:

```
        result.observe(self.usage, usage, {'property', self.property})
```

Simply put, I would have expected `self.usage` to be an argument to the callback:

```
def observe(self, result, i_usage, i_pressure):
      usage, pressure = expensive_system_call()
      result.observe(i_usage, usage, {'property', self.property})
      result.observe(i_pressure, pressure, {'property', self.property})
```

In the form this PR suggests I don't see the purpose of including the instruments in the callback registration since they are used through state of the Device object.

Am I missing something?

tsloughter:
I realize the passing each instrument as an argument is not ideal, that was more to make a point about how the instrument is referenced. The type of callbacks I'd want to register to report multiple instruments would be potentially dealing with dozens of instruments.

specification/metrics/api.md
jmacd: I don't see the conceptual difficulty with having more than one form of callback. I think of callbacks as things that produce measurements, and producing measurements for a single instrument is simpler than producing measurements for more than one instrument. The user is going to know which form makes sense for a given scenario. I can see one library using both forms, but I can't see one instrument using both forms.

specification/metrics/api.md
jmacd: The use of a multi-instrument form as the primary form is what OTel-Go does. We do have only one form there, but to me having both forms would be a minor improvement.  I won't speak for the other languages, I'm just surprised at the reaction that having two forms of callback is somehow confusing. Each callback is going to be called once, regardless of the form.

specification/metrics/api.md
jmacd: Changed this to 
```
Where the API supports registration of `callback` functions after
asynchronous instrumentation creation, the user MUST be able to undo
registration of the specific callback after its registration by some means.
```

jmacd: BTW @ocelotl I can't think of an idiom that doesn't involve returning something to the user. :-)
(I'm thinking of pointers as things returned to the user, so if you're somehow creating a pointer to a function you've got a thing returned to you.)

specification/metrics/api.md
jmacd: Surely the SDK is going to store something to refer to the function that will be called. That thing, I'm arguing, is the Callback entity.

jmacd:
> In the form this PR suggests I don't see the purpose of including the instruments in the callback registration since they are used through state of the Device object.

> Am I missing something?

If a View does not select any of the instruments defined by a callback, for example, the SDK does not need to evaluate that callback. I believe there is a desire to support instruments that are collected on different intervals, that is also enabled by letting the SDK know which callbacks observe which instruments.

specification/metrics/api.md
```diff
@@ -337,19 +338,26 @@ The API to construct asynchronous instruments MUST accept the following paramete
...
 
-Where the API supports registration of callback functions after
-asynchronous instrumentation creation, it MUST return something (e.g.,
-a registration handle, receipt or token) to the user that supports
-undoing the effect of callback registation.
+Where the API supports registration of `callback` functions after
+asynchronous instrumentation creation, the user MUST be able to undo
+registration of the specific callback after its registration somehow.
+
+Every currently registered Callback associated with an instrument MUST
```
reyang, line 358:
```suggestion
Every currently registered `callback` function associated with an instrument MUST
```

specification/metrics/api.md
```diff
@@ -337,19 +338,26 @@ The API to construct asynchronous instruments MUST accept the following paramete
...
+be evaluted exactly once during collection prior to reading data for
```
reyang, line 359: What if there is timeout? What if one `callback` function failed (ideally, they shouldn't fail)?

jmacd: These are good questions, but would you agree that these questions are present with or without an option for multi-instrument callbacks?

jack-berg (approved):
I've prototyped this in java [here](https://github.com/open-telemetry/opentelemetry-java/pull/4376). I think this is a good addition which solves a valuable use case 👍 .

ocelotl (approved):

legendecas (approved):

reyang (approved):

PR #2341, opened 2022-02-14
Add OTEL_EXPORTER_JAEGER_PROTOCOL and improve description of other OTEL_EXPORTER_JAEGER_* env vars

pellared:
Follow-up of #2333

## Changes

Preview [here](https://github.com/pellared/opentelemetry-specification/blob/patch-2/specification/sdk-environment-variables.md#jaeger-exporter)

- Add the `OTEL_EXPORTER_JAEGER_PROTOCOL` environment variable to select the protocol used
  by the Jaeger exporter.
- More explicit description of the rest of the  `OTEL_EXPORTER_JAEGER_*` environment variables (e.g. when they are applicable - it also represents the current state of SDKs)

Here are the following reasons for this change

1. Enable the possibility of selecting the protocol when the Jaeger exporter is set via env var `OTEL_TRACES_EXPORTER=jaeger` (instead of supporting values like `OTEL_TRACES_EXPORTER=jaeger-udp`). 
2. This resembles [the configuration of the OTLP exporter](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md) which has `OTEL_EXPORTER_OTLP_PROTOCOL`
3. The newest OTel .NET SDK RC version supports this env var. See [here](https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Exporter.Jaeger). I feel that it is safer to standardize it (especially the supported values) before it is released
4. We want to take advantage of it in [OTel .NET AutoInstrumentation](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation). However, I believe this could be useful in other places as well.
5. Node.js SDK uses `6832` as the default Agent port because it uses a different protocol

Related issues:
- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/355

FYI (all SIG maintainers where I see that `OTEL_EXPORTER_JAEGER_ENDPOINT` is used) @open-telemetry/dotnet-maintainers @open-telemetry/dotnet-instrumentation-maintainers @open-telemetry/java-instrumentation-maintainers @open-telemetry/python-maintainers @open-telemetry/go-maintainers @open-telemetry/php-maintainers  @open-telemetry/java-maintainers @open-telemetry/ruby-maintainers @open-telemetry/rust-maintainers 

## Alternative

[...]

PR #2364, opened 2022-02-18
OTEL_EXPORTER_OTLP_PROTOCOL value MUST be one of

pellared:
Based on https://github.com/open-telemetry/opentelemetry-specification/pull/2341#discussion_r806960073

By @tigrannajaryan 

> Why does this say `MAY`? This is not a recommendation, it is a strict list. Should it say: `MUST be one of ...` instead?

As far as I checked, only the OTel Erlang SDK would violate this rule. See: https://github.com/open-telemetry/opentelemetry-erlang/search?q=OTEL_EXPORTER_OTLP_PROTOCOL

@tsloughter WDYT? Is it a blocker from your perspective?

### DIFF
--- a/specification/protocol/exporter.md
+++ b/specification/protocol/exporter.md
@@ -36,7 +36,8 @@ The following configuration options MUST be available to configure the OTLP expo
   - Default: 10s
   - Env vars: `OTEL_EXPORTER_OTLP_TIMEOUT` `OTEL_EXPORTER_OTLP_TRACES_TIMEOUT` `OTEL_EXPORTER_OTLP_METRICS_TIMEOUT`
 
-- **Protocol**: The transport protocol. Options MAY include `grpc`, `http/protobuf`, and `http/json`. See [Specify Protocol](./exporter.md#specify-protocol) for more details.
+- **Protocol**: The transport protocol. Options MUST be one of: `grpc`, `http/protobuf`, `http/json`.
+  See [Specify Protocol](./exporter.md#specify-protocol) for more details.
   - Default: `http/protobuf` [3]
   - Env vars: `OTEL_EXPORTER_OTLP_PROTOCOL` `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL`
 


carlosalberto (approved):

tigrannajaryan (approved):

jmacd (approved):

specification/protocol/exporter.md
```diff
@@ -36,7 +36,8 @@ The following configuration options MUST be available to configure the OTLP expo
   - Default: 10s
   - Env vars: `OTEL_EXPORTER_OTLP_TIMEOUT` `OTEL_EXPORTER_OTLP_TRACES_TIMEOUT` `OTEL_EXPORTER_OTLP_METRICS_TIMEOUT`
 
-- **Protocol**: The transport protocol. Options MAY include `grpc`, `http/protobuf`, and `http/json`. See [Specify Protocol](./exporter.md#specify-protocol) for more details.
+- **Protocol**: The transport protocol. Options MUST be one of: `grpc`, `http/protobuf`, `http/json`.
```
codeboten, line 39: Should `grpc` be `grpc/protobuf` for consistency? In case we ever want to provide support for a json over grpc exporter in the future or other encodings.

pellared: I feel it would be redundant (gRPC is always serializing to protobuf)

> we ever want to provide support for a json over grpc

How it would look like? I imagine having a different protobuf definition. However, in such case I rather imagine a new value like `grpc.v2`

cijothomas (approved):

specification/protocol/exporter.md
carlosalberto: Based on https://groups.google.com/g/grpc-io/c/DPgipyU0g5Y it looks like it's _possible_ to use anything other than protobuf to serialize data, but there's are no actual alternatives. Also, it's probably too late to change it (or deprecate it).

codeboten: `grpc` is probably fine for most cases anyways, this works for me.

codeboten (approved):

tsloughter:
Not a blocker for us, but we'll have to keep support for our previous values of `http_protobuf` and just remove them from documentation.

carlosalberto:
@tsloughter You mean still supporting `http_protobuf` but deprecating it, while also supporting the 'new' `http/protobuf`?

tsloughter:
@carlosalberto right.

pellared:
@carlosalberto Good to merge or do we need additional reviews?

PR #2372, opened 2022-02-22
Clarify that instrument name conflict only happens within a Meter

### LINKED ISSUE #2337: Clarify if Views must always specify Meter name

Implement View specification.

**Additional context.**

There is a [requirement in the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#:~:text=In%20order%20to,card%20instrument%20name\)) which was recently tweaked in #2306:

> the SDK MUST NOT allow Views with a specified name to be declared with instrument selectors that select more than one instrument (e.g. wild card instrument name).

The problem is every instrument selector that doesn't specify the Meter name can "select more than one instrument" because identical instruments may be created on any Meter. For example:

```py
meter_provider = MeterProvider(
    views=[View(instrument_name="foo.bar", name="baz")],
)
meter1 = meter_provider.get_meter("library1")
meter2 = meter_provider.get_meter("library2")

meter1.create_counter("foo.bar")
meter2.create_counter("foo.bar")
```

Given the spec wording, the view in the example is invalid even though it doesn't create a conflict in the output metrics. In fact, you would need to specify Meter name on any view that changes the output metric `name`. I'm guessing the intention was:

> the SDK MUST NOT allow Views with a specified name to be declared with instrument selectors that select more than one instrument **in the same Meter** (e.g. wild card instrument name).

reyang:
Fixes #2337.

Based on the [2/15/2022 Spec SIG Meeting](https://docs.google.com/document/d/1pdvPeKjA8v8w_fGKAN68JjWBmVJtPCpqdi9IZrd6eEo/edit?pli=1#heading=h.ixd3fm5gwjy1) discussion and reading through [RFC2119](https://www.rfc-editor.org/rfc/rfc2119.html).

### DIFF
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -192,7 +192,7 @@ In order to avoid conflicts, views which specify a name SHOULD have an
 instrument selector that selects at most one instrument. For the registration
 mechanism described above, where selection is provided via configuration, the
 SDK SHOULD NOT allow Views with a specified name to be declared with instrument
-selectors that select more than one instrument (e.g. wild card instrument name)
+selectors that may select more than one instrument (e.g. wild card instrument name)
 in the same Meter.
 
 The SDK SHOULD use the following logic to determine how to process Measurements
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -191,8 +191,9 @@ are the inputs:
 In order to avoid conflicts, views which specify a name SHOULD have an
 instrument selector that selects at most one instrument. For the registration
 mechanism described above, where selection is provided via configuration, the
-SDK MUST NOT allow Views with a specified name to be declared with instrument
-selectors that select more than one instrument (e.g. wild card instrument name).
+SDK SHOULD NOT allow Views with a specified name to be declared with instrument
+selectors that select more than one instrument (e.g. wild card instrument name)
+in the same Meter.
 
 The SDK SHOULD use the following logic to determine how to process Measurements
 made with an Instrument:


jmacd (approved):

cijothomas (approved):

bogdandrutu (approved):

jack-berg (approved):

specification/metrics/sdk.md
```diff
@@ -191,8 +191,9 @@ are the inputs:
 In order to avoid conflicts, views which specify a name SHOULD have an
 instrument selector that selects at most one instrument. For the registration
 mechanism described above, where selection is provided via configuration, the
-SDK MUST NOT allow Views with a specified name to be declared with instrument
-selectors that select more than one instrument (e.g. wild card instrument name).
+SDK SHOULD NOT allow Views with a specified name to be declared with instrument
```
ocelotl, line 194: This can't be enforced. When a view is declared, it is not known if it will match more than one instrument in the same meter.

specification/metrics/sdk.md
```diff
@@ -191,8 +191,9 @@ are the inputs:
...
+selectors that select more than one instrument (e.g. wild card instrument name)
+in the same Meter.
```
ocelotl, line 196:
```suggestion
A view with a specified name should match at most one instrument in the same Meter.
```

ocelotl: This should also specify which one of the instruments in the same meter that can match the view is the one that would be matched or indicate that it is not specified which one of the instruments in the same meter that can match the view is the one that would be matched.

ocelotl (changes requested):
The requirement as it is stated now can't be enforced.

specification/metrics/sdk.md
reyang: > This can't be enforced.

Here we picked the wording "SHOULD NOT" rather than "MUST NOT" so it's not enforcement.

> When a view is declared, it is not known if it will match more than one instrument in the same meter.

That's correct, here we are saying that folks should not allow wild card name or something that allows multiple matches.

specification/metrics/sdk.md
reyang: From the SDK perspective, all of them would match.
When it comes to particular exporter, things might be different - e.g. OTLP exporter (based on the data model) supports this scenario. For exporters that don't support this (e.g. Prometheus), we should spec it out in the exporter doc (e.g. that's why Prometheus Exporter spec is in its current state).

specification/metrics/sdk.md
ocelotl: Consider this situation:

``` python
meter_provider = MeterProvider(
    views=[View(instrument_name="foo.bar*", name="baz")],
)
meter1 = meter_provider.get_meter("library1")
meter2 = meter_provider.get_meter("library2")

meter1.create_counter("foo.bar")
meter2.create_counter("foo.bar")
```

Here we have a situation that will not cause conflicts in the output metrics (because the view will match instruments in different meters) but the wording of the spec will not allow it.

The case above can be easily fixed by the user by just removing the `*` in the instrument name, so maybe it is ok that the spec forbids a valid case like this one.

specification/metrics/sdk.md
```diff
@@ -191,8 +191,9 @@ are the inputs:
 In order to avoid conflicts, views which specify a name SHOULD have an
 instrument selector that selects at most one instrument. For the registration
 mechanism described above, where selection is provided via configuration, the
-SDK MUST NOT allow Views with a specified name to be declared with instrument
-selectors that select more than one instrument (e.g. wild card instrument name).
+SDK SHOULD NOT allow Views with a specified name to be declared with instrument
+selectors that select more than one instrument (e.g. wild card instrument name)
```
ocelotl, line 195:
```suggestion
selectors that may select more than one instrument (e.g. wild card instrument name)
```
It is not known when the view is declared if it will match more than one instrument in the same meter or not.

specification/metrics/sdk.md
ocelotl: Ok, I think the intention of the spec is to ban the declaration of a named view with any instrument selector that may end up matching more than one instrument in the same meter _without specifying which instrument selectors can do so_. Changing my request for changes to comment.

ocelotl (commented):
Changed from request changes to comment.

specification/metrics/sdk.md
reyang: Exactly, the intention is to not allow it for now. If later we want to allow it, we can (which is an additive change rather than breaking change).

ocelotl: All right, resolving.

ocelotl (approved):

PR #2377, opened 2022-02-23
Add table for well-known rpc.system values

trask:
## Changes

Adds a table for well-known `rpc.system` values, similar to what is done already for `db.system`.

This will make it easier to propose new `rpc.system` values, and it will also enable auto-generation of `rpc.system` constants similar to what is done already for `db.system` in some languages.

### DIFF
--- a/semantic_conventions/trace/rpc.yaml
+++ b/semantic_conventions/trace/rpc.yaml
@@ -5,10 +5,20 @@ groups:
     events: [rpc.message]
     attributes:
       - id: system
-        type: string
         required: always
-        brief: 'A string identifying the remoting system.'
-        examples: ["grpc", "java_rmi", "wcf"]
+        brief: 'A string identifying the remoting system. See below for a list of well-known identifiers.'
+        type:
+          allow_custom_values: true
+          members:
+            - id: grpc
+              value: 'grpc'
+              brief: 'gRPC'
+            - id: java_rmi
+              value: 'java_rmi'
+              brief: 'Java RMI'
+            - id: wcf
+              value: 'wcf'
+              brief: 'WCF'
       - id: service
         type: string
         required:
--- a/semantic_conventions/trace/rpc.yaml
+++ b/semantic_conventions/trace/rpc.yaml
@@ -16,9 +16,9 @@ groups:
             - id: java_rmi
               value: 'java_rmi'
               brief: 'Java RMI'
-            - id: wcf
-              value: 'wcf'
-              brief: 'WCF'
+            - id: dotnet_wcf
+              value: 'dotnet_wcf'
+              brief: '.NET WCF'
       - id: service
         type: string
         required:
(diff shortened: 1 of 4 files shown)


semantic_conventions/trace/rpc.yaml
```diff
@@ -5,10 +5,20 @@ groups:
...
+        type:
+          allow_custom_values: true
+          members:
+            - id: grpc
+              value: 'grpc'
+              brief: 'gRPC'
+            - id: java_rmi
+              value: 'java_rmi'
+              brief: 'Java RMI'
+            - id: wcf
```
Oberon00, line 19:
```suggestion
            - id: dotnet_wcf
```

specification/trace/semantic_conventions/rpc.md
```diff
@@ -72,6 +72,14 @@ Examples of span names:
 
 * [`net.peer.ip`](span-general.md)
 * [`net.peer.name`](span-general.md)
+
+`rpc.system` MUST be one of the following or, if none of the listed values apply, a custom value:
```
yurishkuro, line 76: the use of MUST doesn't work foe me in this sentence. I would phrase it: if the rpc system is one of those listed in the table below, then the respective value MUST be used, otherwise a custom value MAY be used.

semantic_conventions/trace/rpc.yaml
trask: do you think I should change the `value` also?

specification/trace/semantic_conventions/rpc.md
arminru: This is generated by the markdown generator in https://github.com/open-telemetry/build-tools/tree/main/semantic-conventions and the same for all existing semantic conventions using enums in the spec.
In essence it is saying the same thing you suggest but in a more generic way ("if none apply").

If you have a suggestion for some different generic wording please open a separate issue in the spec to discuss it, then we can update the generator and subsequently the existing markdown files.

trask: https://github.com/open-telemetry/build-tools/pull/85

semantic_conventions/trace/rpc.yaml
trask: cc @open-telemetry/dotnet-approvers @open-telemetry/dotnet-contrib-approvers

cijothomas: LGTM. @CodeBlanch for additional comment, if any.

trask: I updated the value also. Should this change be added to change log, or since it was just an "example" value previously it doesn't matter?

arminru: On the wire it won't matter but both producers and consumers now have an enum at hand when generating code from the new definition making handling easier, so adding a changelog entry would make sense.

arminru (approved):

bogdandrutu (approved):

pellared (approved):

PR #2378, opened 2022-02-24
Clarify that push metric exporters can be paired with any MetricReader

alanwest:
I'm unsure if this was intentional, but language introduced in #2221 constrained push metric exporters to being paired with a periodic exporting MetricReader. I'm proposing to relax this constraint.

### DIFF
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -785,11 +785,12 @@ preferred temporality.
 
 ### Push Metric Exporter
 
-Push Metric Exporter sends metric data it receives from a paired [periodic
-exporting MetricReader](#periodic-exporting-metricreader).  Here are some
-examples:
+Push Metric Exporter sends metric data it receives from a paired
+[MetricReader](#metricreader). Here are some examples:
 
 * Sends the data based on a user configured schedule, e.g. every 1 minute.
+  This MAY be accomplished by pairing the exporter with a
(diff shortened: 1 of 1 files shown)


specification/metrics/sdk.md
```diff
@@ -785,11 +785,12 @@ preferred temporality.
...
-Push Metric Exporter sends metric data it receives from a paired [periodic
-exporting MetricReader](#periodic-exporting-metricreader).  Here are some
-examples:
+Push Metric Exporter sends metric data it receives from a paired
+[MetricReader](#metricreader). Here are some examples:
 
 * Sends the data based on a user configured schedule, e.g. every 1 minute.
+  This MAY be accomplished by pairing the exporter with a
+  [periodic exporting MetricReader](#periodic-exporting-metricreader).
 * Sends the data when there is a severe error.
```
alanwest, line 794: If we do not allow push metric exporters to be paired with anything other than a periodic metric reader, then I think this example needs to be removed.

reyang: I think it still can be left as is, and can be achieved by having a periodic metric reader with +Infinite interval.

jmacd: For example, I suppose, we want to support push-based exporters for readers that are not periodic, such as push-based exporters that are triggered on an irregular period.

jmacd (approved):

cijothomas (approved):

PR #2379, opened 2022-02-24
Default push metric exporters to use a periodic metric reader

alanwest:
At least two SDKs differ in whether they default to using the periodic MetricReader when automatically configuring which exporter to use. This PR proposes that SDKs should default to using the periodic MetricReader when configuring the use of the stdout, in-memory, and OTLP exporters.

### Java
The autoconfigure module allows users to configure the use of either the OTLP or logging (stdout) exporter. The exporter is paired with a periodic MetricReader.

https://github.com/open-telemetry/opentelemetry-java/blob/083dca1b94c6e803579311e008848aa6b80d921b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java#L44-L67

### .NET
Users can programmatically add an exporter without explicitly declaring the MetricReader to use. The exporter is _not_ paired with a periodic MetricReader by default and requires `ForceFlush` to be called on the MeterProvider. However, you can configure it to use the periodic MetricReader.

https://github.com/open-telemetry/opentelemetry-dotnet/blob/644ba5bc14eac8919032cb27a958d1fa919a8b37/examples/AspNetCore/Startup.cs#L137-L145

---

This PR is related to #2378. If #2378 is not accepted - that is, it is intended that all push metric exporters always be paired with a periodic exporting MetricReader, then this PR is probably unnecessary.


bogdandrutu (approved):

specification/metrics/sdk_exporters/in-memory.md
```diff
@@ -6,6 +6,13 @@ In-memory Metrics Exporter is a [Push Metric
 Exporter](../sdk.md#push-metric-exporter) which accumulates metrics data in the
 local memory and allows to inspect it (useful for e.g. unit tests).
 
+If a language provides a mechanism to automatically configure a
+[MetricReader](../sdk.md#metricreader) to pair with the exporter (e.g., using
+the
+[`OTEL_METRICS_EXPORTER` environment variable](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#exporter-selection)),
```
cijothomas, line 12: use relative links

jack-berg (approved):

alanwest:
Folks during the 3/1 spec SIG meeting agreed that defaulting to using a periodic metric reader for push based is fine. Though, there was some additional discussion for defaulting the `stdout` exporter to a shorter that 60s interval - 10-15s was proposed.

Since this PR already has some approvals, I can follow up with another PR suggesting a shorter interval for stdout.

cijothomas (approved):

reyang (approved):
LGTM with changelog updated.

codeboten (approved):

PR #2380, opened 2022-02-24
Convert OpenMetrics Info and StateSet metrics to non-monotonic sums

dashpole:
Issues:

* https://github.com/open-telemetry/wg-prometheus/issues/65
* https://github.com/open-telemetry/wg-prometheus/issues/66

## Changes

Instead of dropping Info and StateSet metrics, we should convert them to gauges.  In the future, there may be reasons to want to preserve the original type (similar to https://github.com/open-telemetry/wg-prometheus/issues/69), but for now, converting to gauges is preferable to dropping them.

@Aneurysm9 @jmacd @brian-brazil @jsuereth


Aneurysm9 (approved):

specification/metrics/datamodel.md
```diff
@@ -1152,13 +1154,19 @@ Multiple Prometheus metrics are merged together into a single OTLP Summary:
 * The `quantile` label on non-suffixed metrics is used to identify quantile points in summary metrics. Each Prometheus line produces one quantile on the resulting summary.
 * Lines with `_count` and `_sum` suffixes are used to determine the summary's count and sum.
 
+#### Info
+
+An [OpenMetrics Info](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#info) metric MUST be converted to an OTLP Gauge.
```
bogdandrutu, line 1159: I always thought that these will become "Resource.Attributes", a.k.a they are describing the source of (resource that records) the metrics. Unfortunately OpenMetrics does not define the protocol, and one question that I asked multiple times was, how are these related to the rest of the metrics in the same set.

/cc @RichiH can we have a clarification on how the "Info" metric correlates with the other metrics in the same set, if they correlate at all?

specification/metrics/datamodel.md
```diff
@@ -1152,13 +1154,19 @@ Multiple Prometheus metrics are merged together into a single OTLP Summary:
...
+
+#### StateSet
+
+An [OpenMetrics StateSet](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#stateset) metric MUST be converted to an OTLP Gauge.
+
```
bogdandrutu, line 1164: We need to describe a bit how. I understand that they can, but we need to at least give some examples.

specification/metrics/datamodel.md
brian-brazil: They can themselves have labels, in which case they're only for a part of the returned metrics. So it depends.

dashpole: Info metrics may only apply to some of the metrics on an endpoint, so I don't think we can use _all_ info metrics for resource.  We can and should use the (special) "target" info metric (https://github.com/open-telemetry/opentelemetry-specification/pull/2381) for resource attributes.

specification/metrics/datamodel.md
dashpole: Info and StateSet are basically just special cases of a gauge with some additional restrictions.  For example, an info is just a gauge which always has value 1, and shouldn't change over the life of the process.  I've added descriptions to this effect, and moved this section below the gauge section.

Conversion from Prometheus gauge to OTLP gauge is fairly straightforward, but I can add examples if you would still like them.

specification/metrics/datamodel.md
```diff
@@ -1128,6 +1130,14 @@ A [Prometheus Counter](https://github.com/OpenObservability/OpenMetrics/blob/mai
 
 A [Prometheus Gauge](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#gauge) MUST be converted to an OTLP Gauge.
 
+#### Info
+
+An [OpenMetrics Info](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#info) metric MUST be converted to an OTLP Gauge. An OpenMetrics Info can be thought of as a special-case of the OpenMetrics Gauge which has a value of 1, and whose labels generally stays constant over the life of the process.
+
+#### StateSet
+
+An [OpenMetrics StateSet](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#stateset) metric MUST be converted to an OTLP Gauge. An OpenMetrics StateSet can be thought of as a special-case of the OpenMetrics Gauge which has a 0 or 1 value, and has one metric for every possible state.
```
Aneurysm9, line 1139:
```suggestion
An [OpenMetrics StateSet](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#stateset) metric MUST be converted to an OTLP Gauge. An OpenMetrics StateSet can be thought of as a special-case of the OpenMetrics Gauge which has a 0 or 1 value, and has one datapoint for every possible state.
```

If the name is constant and states are expressed as labels, it would map to a datapoint per state, right?  Or does the name of the metric get appended with the state identifier?  It's not clear from the linked description.

dashpole: You are right.  I've been using "metric" and datapoint interchangeably, when I probably shouldn't.  But since i'm describing a prometheus data point, i'm going to use the prometheus name, which is "MetricPoint"

specification/metrics/datamodel.md
bogdandrutu: The added section clarified everything. Thanks.

specification/metrics/datamodel.md
```diff
@@ -1128,6 +1130,14 @@ A [Prometheus Counter](https://github.com/OpenObservability/OpenMetrics/blob/mai
 
 A [Prometheus Gauge](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#gauge) MUST be converted to an OTLP Gauge.
 
+#### Info
```
bogdandrutu, line 1133: Should we say "except" target metricfamily?

bogdandrutu (approved):

specification/metrics/datamodel.md
dashpole: done.

dashpole:
Since this PR now references the target info metric handing in https://github.com/open-telemetry/opentelemetry-specification/pull/2381, it should wait until after that PR is merged.

jsuereth (approved):

specification/metrics/datamodel.md
```diff
@@ -1128,6 +1130,14 @@ A [Prometheus Counter](https://github.com/OpenObservability/OpenMetrics/blob/mai
...
+
+An [OpenMetrics Info](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#info) metric MUST be converted to an OTLP Gauge unless it is the "target" info metric, which is used to populate [resource attributes](#resource-attributes). An OpenMetrics Info can be thought of as a special-case of the OpenMetrics Gauge which has a value of 1, and whose labels generally stays constant over the life of the process.
```
jmacd, line 1135: May I propose that Info => UpDownCounter?
For justification, the value of 1 is intended as a count, it seems to me.  If you were removing attributes from a set of Info metrics, I would expect them to be summed. The result will convert to Gauge in the OpenMetrics/Prometheus model either way.

specification/metrics/datamodel.md
```diff
@@ -1128,6 +1130,14 @@ A [Prometheus Counter](https://github.com/OpenObservability/OpenMetrics/blob/mai
...
+
+#### StateSet
+
+An [OpenMetrics StateSet](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#stateset) metric MUST be converted to an OTLP Gauge. An OpenMetrics StateSet can be thought of as a special-case of the OpenMetrics Gauge which has a 0 or 1 value, and has one metric point for every possible state.
```
jmacd, line 1139: May I propose that Info => UpDownCounter?
Same argument as above. If you remove attributes from these metrics, the 0 or 1 value will be replaced with a sum.

specification/metrics/datamodel.md
dashpole: That makes sense to me.  I also added this explanation:

"It is converted to a Non-Monotonic Sum, rather than a Gauge, because the value of 1 is intended to be viewed as a count, which should be summed together when aggregating away labels."

dashpole: Also, this is a non-monotonic sum in the datamodel terms, right?

specification/metrics/datamodel.md
dashpole: Same as above.

specification/metrics/datamodel.md
bogdandrutu: Yes, in data model that becomes a non-monotonic SUM. which I think is the right terminology to use here, since this will happen in the collector mostly.

specification/metrics/datamodel.md
```diff
@@ -1128,6 +1130,14 @@ A [Prometheus Counter](https://github.com/OpenObservability/OpenMetrics/blob/mai
 
 A [Prometheus Gauge](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#gauge) MUST be converted to an OTLP Gauge.
 
+#### Info
+
+An [OpenMetrics Info](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#info) metric MUST be converted to an OTLP Non-Monotonic Sum unless it is the "target" info metric, which is used to populate [resource attributes](#resource-attributes). An OpenMetrics Info can be thought of as a special-case of the OpenMetrics Gauge which has a value of 1, and whose labels generally stays constant over the life of the process. It is converted to a Non-Monotonic Sum, rather than a Gauge, because the value of 1 is intended to be viewed as a count, which should be summed together when aggregating away labels.
+
+#### StateSet
```
bogdandrutu, line 1137: Crazy idea to be investigated in a separate PR. Should we consider to add this as a new data type in our data model?

bogdandrutu, line 1137: Crazy idea to be investigated in a separate PR. Should we consider to add this as a new data type in our data model?

dashpole: Opened https://github.com/open-telemetry/opentelemetry-specification/issues/2409 to discuss.

jmacd (approved):

dashpole:
I think this can be merged if everyone is happy with it

bogdandrutu (approved):

PR #2381, opened 2022-02-24
Improve conversion of resource attributes to/from prometheus

dashpole:
Fixes https://github.com/open-telemetry/opentelemetry-specification/issues/1782
Related: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/7081

### Job and Instance <-> Resources Attributes

(Discussion in https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/7081)

#### Job 

Current:`job` prom label -> `job` resource attribute -> `job` prom label.
New: `job` prom label -> `service.name` resource attribute -> `job` prom label.

This will improve the OTLP receiver (and others) -> Prometheus remote write exporter path, since `service.name` is more likely to be present than `job` on metrics not originating from prometheus receivers.

#### Instance

Current:`instance` prom label -> `instance` resource attribute -> `instance` prom label.
New: `instance` prom label -> `service.instance.id` resource attributes -> `instance` prom label.

This will improve the OTLP receiver (and others) -> Prometheus remote write exporter path, since `service.instance.id` are more likely to be present than `instance` on metrics not originating from prometheus receivers.

This will also improve the Prometheus receiver -> OTLP (and other) exporter path, since `net.host.name` +`net.host.port`  align better with semantic conventions than `host.name` + `port` or `instance`.  As discussed in the linked issue, [http metric conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md#attributes) are the best fit for these attributes.

### Other Resource Attributes <-> target_info

OpenMetrics explicitly defines a way for exporters which discover their own resource information to send metadata with a batch of Prometheus metrics: [the target_info metric](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#supporting-target-metadata-in-both-push-based-and-pull-based-systems).  This PR describes how this metric SHOULD be used, although it doesn't mandate its use.

cc @Aneurysm9 @jmacd @jsuereth @bogdandrutu @djaglowski


specification/metrics/datamodel.md
```diff
@@ -1240,9 +1238,13 @@ OpenTelemetry Metric Attributes MUST be converted to [Prometheus labels](https:/
 
 #### Resource Attributes
 
-In SDK Prometheus (pull) exporters, all resource attributes MUST be dropped, and MUST NOT be attached as labels. The scraper of the endpoint is expected to discover resource attributes of the endpoint it is scraping.
+In SDK Prometheus (pull) exporters, resource attributes SHOULD be converted to the [`target_info`](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#supporting-target-metadata-in-both-push-based-and-pull-based-systems) metric, or MUST be dropped, and MUST NOT be attached as labels to other metrics. The `target_info` metric MUST info-typed metric whose labels MUST include the resource attributes, and MUST NOT include any other labels. There MUST be at most one `target_info` metric exposed on a Prometheus endpoint.
```
Aneurysm9, line 1241:
```suggestion
In SDK Prometheus (pull) exporters, resource attributes SHOULD be converted to the [`target_info`](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#supporting-target-metadata-in-both-push-based-and-pull-based-systems) metric; otherwise, they MUST be dropped, and MUST NOT be attached as labels to other metrics. The `target_info` metric MUST be an info-typed metric whose labels MUST include the resource attributes, and MUST NOT include any other labels. There MUST be at most one `target_info` metric exposed on a Prometheus endpoint.
```

specification/metrics/datamodel.md
```diff
@@ -1177,11 +1177,9 @@ The following attributes MUST be associated with scraped metrics as resource att
 | OTLP Resource Attribute | Description |
 | ----------------------- | ----------- |
 | `service.name` | The configured name of the service that the target belongs to |
-| `job` | Identical to `service.name` |
-| `instance` | The <host>:<port> of the target's URL that was scraped. |
-| `host.name` | `instance` The <host> portion of `instance` |
-| `port` | `instance` The <port> portion of `instance` |
-| `scheme` | `http` or `https` |
+| `net.host.name` | The <host> portion of the target's URL that was scraped |
+| `net.host.port` | The <port> portion of the target's URL that was scraped |
```
Aneurysm9, line 1181: While the default value of `instance` will be the `__address__` that has host and port, relabeling rules seem to often change that, potentially in ways that make this mapping inappropriate.

dashpole: Are you worried that if someone relabels instance, that it the relabeled value won't be properly reconstructed in the exporter, or that we won't be able to always host/port attributes when scraping?

Aneurysm9: As long as we only combine `net.host.name` and `net.host.port` with a `:` when the port is non-empty, reconstruction is probably not an issue.  I'm more worried about times when the `instance` label is relabeled to something that is not a hostname and then exported through OTLP or some other exporter that doesn't convert back to `instance`.

specification/metrics/datamodel.md
```diff
@@ -1240,9 +1238,13 @@ OpenTelemetry Metric Attributes MUST be converted to [Prometheus labels](https:/
 
 #### Resource Attributes
 
-In SDK Prometheus (pull) exporters, all resource attributes MUST be dropped, and MUST NOT be attached as labels. The scraper of the endpoint is expected to discover resource attributes of the endpoint it is scraping.
+In SDK Prometheus (pull) exporters, resource attributes SHOULD be converted to the [`target_info`](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#supporting-target-metadata-in-both-push-based-and-pull-based-systems) metric; otherwise, they MUST be dropped, and MUST NOT be attached as labels to other metrics. The `target_info` metric MUST be an info-typed metric whose labels MUST include the resource attributes, and MUST NOT include any other labels. There MUST be at most one `target_info` metric exposed on a Prometheus endpoint.
```
bogdandrutu, line 1241: > There MUST be at most one `target_info` metric exposed on a Prometheus endpoint.

Is this a requirement in Prometheus as well?

specification/metrics/datamodel.md
```diff
@@ -1240,9 +1238,13 @@ OpenTelemetry Metric Attributes MUST be converted to [Prometheus labels](https:/
...
...ets.  By convention, [`job` and `instance`](https://Prometheus.io/docs/concepts/jobs_instances/#jobs-and-instances) labels distinguish targets and are expected to be present on metrics exposed on a Prometheus pull exporter (a ["federated"](https://Prometheus.io/docs/Prometheus/latest/federation/) Prometheus endpoint) or pushed via Prometheus remote-write. In the collector Prometheus exporters, the `service.name` attribute MUST be converted to the `job` metric label, and the `net.host.name` and `net.host.port` attributes, if present, MUST be converted to the `instance` label as `<net.host.name>:<net.host.port>`. If `job` and `instance` are successfully added, other resource attributes SHOULD be converted to a [`target_info`](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#supporting-target-metadata-in-both-push-based-and-pull-based-systems) metric, or MUST be dropped. The `target_info` metric is an info-typed metric whose labels MUST include the resource attributes, and MUST NOT include any other labels other than `job` and `instance`.  There MUST be at most one `target_info` metric exported for each unique combination of `job` and `instance`.
```
bogdandrutu, line 1243: The prometheus remote write is not compatible with OpenMetrics (not sure about Prometheus Pull protocol). I am a bit confused about the suggestion to use target_info and the info metric.

nit: I would split this in multiple lines.

specification/metrics/datamodel.md
dashpole: From https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#supporting-target-metadata-in-both-push-based-and-pull-based-systems :

"An exposer may include an Info MetricFamily called "target" with a **single Metric** with no labels with the metadata."

It doesn't explicitly say that an exposer can't have multiple "target" info metrics (just that they MAY have one), but clearly the intention is to have one.  I wanted to state it more clearly for our exporters.

specification/metrics/datamodel.md
dashpole: WDYT about using `service.instance.id` as instance?  I.e. `instance` prom label -> `service.instance.id` resource attribute -> `instance` prom label.  Since job + instance are expected to be unique in Prometheus, it would make sense to convert`service.name` + `service.namespace` + `service.instance.id` to `job` + `instance` when exporting to prometheus.

specification/metrics/datamodel.md
dashpole: Prometheus remote-write [protocol supports](https://github.com/prometheus/prometheus/blob/main/prompb/types.pb.go#L37) info-typed metrics. The "target" info-typed metric is a convention introduced by OpenMetrics in [its spec](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#supporting-target-metadata-in-both-push-based-and-pull-based-systems), but nothing prevents us from using the convention in our exporters.

I split lines in this paragraph.  If you'd like it done for the entire prom spec, I can open a separate PR to do that.

specification/metrics/datamodel.md
Aneurysm9: That seems reasonable to me.

dashpole: Done

specification/metrics/datamodel.md
bogdandrutu: I think the "metric" in OM is different than the "metric" in OTel. Metric in otel is metricfamily I believe in OM... Also OM Metric does not have a name, that belongs to `metricfamily`.

I would use their terminology "An exposer may include an Info MetricFamily called "target" with a single Metric with no labels with the metadata. An example in the text format might be"

specification/metrics/datamodel.md
```diff
@@ -1170,18 +1170,37 @@ Prometheus Cumulative metrics do not include the start time of the metric. When
...
 
 | OTLP Resource Attribute | Description |
 | ----------------------- | ----------- |
 | `service.name` | The configured name of the service that the target belongs to |
-| `job` | Identical to `service.name` |
-| `instance` | The <host>:<port> of the target's URL that was scraped. |
-| `host.name` | `instance` The <host> portion of `instance` |
-| `port` | `instance` The <port> portion of `instance` |
-| `scheme` | `http` or `https` |
+| `service.instance.id` | A unique identifier of the target.  By default, it should be the scraped URL |
```
Aneurysm9, line 1186:
```suggestion
| `service.instance.id` | A unique identifier of the target.  By default, it should be the <host>:<port> of the scraped URL |
```

`instance` will default to the `__address__` label, which is \<host\>:\<port\>.

Aneurysm9 (approved):

specification/metrics/datamodel.md
dashpole: You are right.  I've cleaned up the language.  OM has [`metric family`](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metricfamily) (OTel Metric) and [`metric point`](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metricpoint) (OTel datapoint).

specification/metrics/datamodel.md
```diff
@@ -1170,18 +1170,38 @@ Prometheus Cumulative metrics do not include the start time of the metric. When
...
+| `http.scheme` | `http` or `https` |
+
+In addition to the attributes above, the
+["target" info](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#supporting-target-metadata-in-both-push-based-and-pull-based-systems)
+metric family is used to supply additional resource attributes. If present,
+"target" info MUST be dropped from the batch of metrics, and all labels from
+the "target" info metric family MUST be converted to resource attributes
+attached to all other metrics which are part of the scrape. By default, label
+keys and values MUST NOT be altered (such as replacing `_` with `.` characters
+in keys).
```
bogdandrutu, line 1204: This is an important and interesting difference. We may suffer because of this, but I think we can provide a transform processor capability for this.

specification/metrics/datamodel.md
```diff
@@ -1240,9 +1260,35 @@ OpenTelemetry Metric Attributes MUST be converted to [Prometheus labels](https:/
...
+formats do not include a notion of resource, and expect metric labels to
+distinguish scraped targets. By convention, [`job` and `instance`](https://Prometheus.io/docs/concepts/jobs_instances/#jobs-and-instances)
+labels distinguish targets and are expected to be present on metrics exposed on
+a Prometheus pull exporter (a ["federated"](https://Prometheus.io/docs/Prometheus/latest/federation/)
+Prometheus endpoint) or pushed via Prometheus remote-write. In OTLP, the
+`service.name`, `service.namespace`, and `service.instance.id` triplet is
+[required to be unique](https://github.com/open-telemetry/opentelemetry-specification/blob/da74730bf835010229a9612c281f052e26553218/specification/resource/semantic_conventions/README.md#service),
+which makes them good candidates to use to construct `job` and `instance`. In
+the collector Prometheus exporters, the `service.name` and `service.namespace`
+attributes MUST be combined as `<service.namespace>;<service.name>`, or
```
bogdandrutu, line 1278: Can we use `<service.namespace>/<service.name>` "/" instead of ";"? Seems nicer ...

bogdandrutu (approved):

specification/metrics/datamodel.md
dashpole: Yeah, i'm definitely on-board with working around this with processors, etc.  We just can't do that by default, since we won't always be correct about our guess.

specification/metrics/datamodel.md
dashpole: done.  I was trying to use a less-commonly-used character to reduce collisions, but readability is probably more important.

specification/metrics/datamodel.md
```diff
@@ -1170,18 +1170,38 @@ Prometheus Cumulative metrics do not include the start time of the metric. When
...
-| `port` | `instance` The <port> portion of `instance` |
-| `scheme` | `http` or `https` |
+| `service.instance.id` | A unique identifier of the target.  By default, it should be the <host>:<port> of the scraped URL |
+
+The following attributes SHOULD be associated with scraped metrics as resource
+attributes, and MUST NOT be added as metric attributes:
+
+| OTLP Resource Attribute | Description |
+| ----------------------- | ----------- |
+| `net.host.name` | The <host> portion of the target's URL that was scraped |
```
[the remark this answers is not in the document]
dashpole: fixed

specification/metrics/datamodel.md
```diff
@@ -1170,18 +1170,38 @@ Prometheus Cumulative metrics do not include the start time of the metric. When
...
 
 | OTLP Resource Attribute | Description |
 | ----------------------- | ----------- |
 | `service.name` | The configured name of the service that the target belongs to |
-| `job` | Identical to `service.name` |
-| `instance` | The <host>:<port> of the target's URL that was scraped. |
-| `host.name` | `instance` The <host> portion of `instance` |
-| `port` | `instance` The <port> portion of `instance` |
-| `scheme` | `http` or `https` |
+| `service.instance.id` | A unique identifier of the target.  By default, it should be the <host>:<port> of the scraped URL |
```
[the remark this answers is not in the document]
dashpole: fixed

jsuereth (approved):

jmacd (approved):

PR #2386, opened 2022-02-28
Add schema file checks using the new tool

tigrannajaryan:
This uses the recently added schemas tool [0] to do a bit more
comprehensive checks of schema files. More checks can be added in
the future.

0 - https://github.com/open-telemetry/build-tools/pull/83


tools/schema_check.sh
```diff
@@ -6,10 +6,13 @@
 
 set -e
 
+BUILD_TOOL_SCHEMAS_VERSION=latest
```
arminru, line 9: We should be careful to not break the spec workflows if new images are published that break existing behavior. I'd prefer pinning the version.

tools/schema_check.sh
```diff
@@ -48,5 +51,9 @@ for file in $schemas_dir/*; do
     echo "FAILED: schema_url is not found in $file"
     exit 2
   fi
+
+  docker run -v $current_dir/../schemas:/schemas \
+  		otel/build-tool-schemas:$BUILD_TOOL_SCHEMAS_VERSION --file /schemas/$ver --version=$ver
+
```
arminru, line 57: Will this just run the tool so the output appears in the log or actually fail the build if something is incorrect?

tigrannajaryan: It fails if the schema file is incorrect. `set -e` ensures that we fail and exit on any error.

tools/schema_check.sh
tigrannajaryan: There is currently no tagged version of the tool docker image. I don't know how to make a release of https://github.com/open-telemetry/build-tools so that we have a tagged version. Do you know?

arminru: Good point, this is something that should be documented there. One just has to create Github release and add a tag in the same format as the others and it will push all containers with the version specified in the tag. Ideally we'd merge https://github.com/open-telemetry/build-tools/pull/85 and then release both at once as `v0.11.0`.

tigrannajaryan: How about we merge this PR now and then we can pin to 0.11.0 when it is released? Alternatively we can release 0.11.0 now without waiting for https://github.com/open-telemetry/build-tools/pull/85

arminru:
```suggestion
BUILD_TOOL_SCHEMAS_VERSION=0.11.0
```

There you go: https://hub.docker.com/r/otel/build-tool-schemas/tags
https://github.com/open-telemetry/build-tools/releases/tag/v0.11.0

arminru (approved):

tools/schema_check.sh
tigrannajaryan: Thanks. Pinned to 0.11.0.

tigrannajaryan:
@open-telemetry/specs-logs-approvers @open-telemetry/specs-metrics-approvers @open-telemetry/specs-trace-approvers please review.

yurishkuro (approved):

PR #2389, opened 2022-03-01
Add thread semantic attributes for logs

trask:
## Changes

Adds thread semantic attributes for logs, copied from the thread semantic attributes for spans.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,9 @@ release.
 
 ### Semantic Conventions
 
+- Add thread semantic attributes for logs.
+  ([#2389](https://github.com/open-telemetry/opentelemetry-specification/pull/2389))
+
 ### Compatibility
 
 ### OpenTelemetry Protocol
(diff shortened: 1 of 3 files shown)


arminru (approved):

tigrannajaryan:
> Adds thread semantic attributes for logs, copied from the thread semantic attributes for spans.

I don't think copying is the right approach. This creates unnecessary maintenance burden. A significant potion of attributes that can be recorded on spans can also be recorded on logs. Copying is not going to work.

We discussed this in the past and the thinking is that we need to move semantic convention to a common directory. @tedsuo had a PR which did not move forward: https://github.com/open-telemetry/opentelemetry-specification/pull/1977/files

Perhaps it is time to revive it.

bogdandrutu (commented):
Why do we not use the tool to generate these attributes from yaml?

trask:
Closing for now as there's a lot of prerequisite tooling work and spec restructuring before this PR can move forward.

bogdandrutu:
@tigrannajaryan @trask would be good to at least start a tracking issue for this.

PR #2404, opened 2022-03-05
Clarify MetricReader parameters and temporality controls

### LINKED ISSUE #2065: Set preferred aggregation temporality for the periodic exporting metric reader via environment variable

### LINKED ISSUE #2403: Preferred aggregation temporality is under-specified

Whether through Views or another mechanism, the MetricReader needs a way to dictate the aggregation temporality that is used by default. For some exporters (OTLP, Console, In-Memory), the choice is something the user might care about, particularly for OTLP where there are memory costs associated with this decision.

What did you expect to see?

Some way that a user can control temporality, if they care. The current specification does not give strict requirements for "preferred aggregation temporality". One solution was outlined in #2314, where the View allowed to specify a `strategy` which was either "preferred", "cumulative" or "delta" (hypothetically). Another solution that was discussed in that PR would to let the temporality decision be defaulted on a per-instrument basis. 

This applies to most of the exporters, makes the decision a "5-way function" for the five instruments that define the concept of aggregation temporality (i.e., the six instruments except Gauge).

This PR defines the requirements, but does not go as far as defining environment variable syntax for configuring the temporality on a per-instrument basis.

jmacd:
Fixes #2403.
Replaces #2314.
Replaces #2407.
Fixes #2065.
Part of #2072.

## Changes

Clarify the role of the MetricReader in determining the default aggregation, temporality, and default_enabled properties. Clarify the Metric Exporter's "associated" MetricReader and that when auto-configuring an exporter, the Metric Reader's defaults MAY be determined in ways specific to the exporter.

The present form of this PR has a single environment variable to control variable temporality of the OTLP exporter. Note that with 5 instruments having temporality and 2 choices, there are 32 possible settings for temporality across the instruments. I would prefer not to have 5 environment variables, because not very many of the configurations are actually expected.

Therefore, this proposal uses a single environment variable with two specified values, CUMULATIVE (the default), DELTA (where Counters, Async Counters and Histograms all use Delta, UpDownCounter and Async UpDownCounter uses Cumulative).

There's a third value I've proposed in #2314, that would be STATELESS, where synchronous instruments choose Delta and asynchronous instruments choose Cumulative. I do not think we should have 5 environment variables, because not many more than 3 out of the possible 32 combinations are actually going to happen. I'm happy to leave STATELESS out of the current proposal because (speaking as a vendor), my needs are met without it. Given that a MetricReader supports configuring the default temporality as a function of instrument kind, we're able to fill in any of the 29 unspecified temporality combinations through a programmatic interface. It means we will be able to set the STATELESS behavior in our launchers even if the environment variable does not recognize that setting.


specification/metrics/sdk.md
```diff
@@ -782,7 +784,7 @@ can run at different schedule, for example:
   pipe.
 
 `MetricExporter` SHOULD provide a way to allow `MetricReader` to retrieve its
-preferred temporality.
+preferred Aggregation Temporality on a per-instrument basis.
```
bogdandrutu, line 787: Do we need the exporter to have a "preferred temporality"? Why not have a helper func in the exporter to create both the `MetricsReader` with the `MetricsExporter` installed and correct "default views" installed in the reader?

bogdandrutu (commented):
LGTM, I think this clarifies more, but I have a small question about the need of having "preferred" temporality at all.

specification/metrics/sdk.md
jmacd: > Why not have a helper func in the exporter to create both the MetricsReader with the MetricsExporter installed and correct "default views" installed in the reader?

Are we splitting hairs? The "helper func in the exporter" that you describe sounds like an implementation detail that meets the requirement.

jmacd:
I have prepared #2407, which is an alternative to this one. This specifies more, #2407 specifies less.

specification/metrics/sdk.md
```diff
@@ -199,8 +204,11 @@ The SDK SHOULD use the following logic to determine how to process Measurements
 made with an Instrument:
 
 * Determine the `MeterProvider` which "owns" the Instrument.
-* If the `MeterProvider` has no `View` registered, take the Instrument and apply
-    the default configuration.
+* If the `MeterProvider` has no `View` registered, take the Instrument
+    and apply the default configuration for Aggregation and
+    Aggregation Temporality.  The SDK MUST support configuring the
+    Aggregation and default Aggregation Temporality on a
```
codeboten, line 210:
```suggestion
    Aggregation and Aggregation Temporality on a
```

codeboten (approved):
Just a minor comment, otherwise i think this provides a good path moving forward.

specification/metrics/sdk.md
```diff
@@ -180,10 +180,15 @@ are the inputs:
     will be used by default (TODO: once the Hint API is available, the default
     behavior should respect the Hint if it is available).
   * The `aggregation` (optional) to be used. If not provided, the SDK MUST
-    apply a [default aggregation](#default-aggregation). If the aggregation
-    outputs metric points that use aggregation temporality (e.g. Histogram,
-    Sum), the SDK SHOULD handle the aggregation temporality based on the
-    temporality of each [MetricReader](#metricreader) instance.
+    apply a [default aggregation](#default-aggregation) configurable on a
+	per-instrument basis according to the [MetricReader](#metricreader) instance.
```
ocelotl, line 184:
```suggestion
    per-instrument basis according to the [MetricReader](#metricreader) instance.
```

ocelotl: Also, should this be per-instrument or per-instrument kind?

alanwest, line 184: To clarify, are you suggesting that upon initialization the default aggregation _and_ aggregation temporality to use for each instrument kind should be configurable at the `MetricReader`?

specification/metrics/sdk.md
```diff
@@ -672,12 +680,6 @@ The SDK SHOULD provide a way to allow `MetricReader` to respond to
 idiomatic approach, for example, as `OnForceFlush` and `OnShutdown` callback
 functions.
 
-The SDK SHOULD provide a way to allow the preferred [Aggregation
-Temporality](./datamodel.md#temporality) to be specified for a `MetricReader`
-instance during the setup (e.g. initialization, registration, etc.) time. If the
-preferred temporality is explicitly specified then the SDK SHOULD respect that,
-otherwise use Cumulative.
```
alanwest, line 679: Why was this section removed?

It seems we'll need to specify that there needs to be some mechanism to configure the default aggregation temporality (and also aggregation. see https://github.com/open-telemetry/opentelemetry-specification/pull/2404/files#r822224887) to use on a per-instrument kind basis with a `MetricReader`.

alanwest:
I'm liking the direction of this PR. I just had a few questions for clarification.

#2404 alone as discussed in today's SIG meeting is not ideal - that is I feel we need to continue to maintain some way to configure aggregation temporality.

jmacd:
FYI I will work to clarify the questions posed in this PR today.

specification/metrics/sdk.md
jmacd: I have revised this text to emphasize the separation of default behaviors (a MetricReader setting) from the exporter. PTAL.

specification/metrics/sdk.md
jmacd: Revised.

specification/metrics/sdk.md
jmacd: Following the feedback from Tuesday's SIG, I've heavily revised this text. I restored the OTLP exporter's DELTA setting but defined it the way you're interested in, that is to say Cumulative for UpDownCounter but Delta otherwise.

jmacd:
@open-telemetry/specs-metrics-approvers I've expanded the scope of this PR following Tuesday's SIG meeting. Mainly, I restored the OTLP exporter environment variable that expresses what was formerly known as a "preference". Here I'm defining this as the "default aggregation temporality policy", because the default aggregation temporality is a MetricReader setting that is made "on the basis of instrument kind".

In the latest round of editing, I tried to strengthen our understanding of MetricReader. It's not an interface the way I understand it, it's more like an abstract base class for the thing that registers exporters and provides their default settings. Please take a look.

specification/metrics/sdk.md
```diff
@@ -781,9 +789,6 @@ can run at different schedule, for example:
 * Exporter D is a pull exporter which reacts to another scraper over a named
   pipe.
 
-`MetricExporter` SHOULD provide a way to allow `MetricReader` to retrieve its
-preferred temporality.
```
alanwest, line 785: I think this language is important for Prometheus - though maybe it's not quite the right language. That is, Prometheus isn't going to work if you've configured the associated MetricReader to use anything but cumulative temporality. The way I'm thinking about it is a `MetricExporter` can have a restriction - not really a preference - on the temporality it supports.

specification/metrics/sdk.md
```diff
@@ -637,12 +642,30 @@ measurements using the equivalent of the following naive algorithm:
...
+* Collecting metrics from the SDK on demand.
 * Handling the [ForceFlush](#forceflush) and [Shutdown](#shutdown) signals from
   the SDK.
 
+To construct a `MetricReader` when setting up an SDK, the caller
+SHOULD provide at least the following:
+
+* The `exporter` to use, which is a `MetricExporter` instance.
+* The default View `aggregation` (optional), a function of instrument kind.
+* The default View `temporality` (optional), a function of instrument kind.
```
alanwest, line 658: With temporality by instrument kind clearly specified as configuration for a `MetricReader`, do we need the ability to override this configuration via the View API?

specification/metrics/sdk.md
```diff
@@ -180,10 +180,15 @@ are the inputs:
     will be used by default (TODO: once the Hint API is available, the default
     behavior should respect the Hint if it is available).
   * The `aggregation` (optional) to be used. If not provided, the SDK MUST
-    apply a [default aggregation](#default-aggregation). If the aggregation
-    outputs metric points that use aggregation temporality (e.g. Histogram,
-    Sum), the SDK SHOULD handle the aggregation temporality based on the
-    temporality of each [MetricReader](#metricreader) instance.
+    apply a [default aggregation](#default-aggregation) configurable on a
+    per-instrument basis according to the [MetricReader](#metricreader) instance.
+  * The aggregation `temporality` (optional) to be used, which may be one of
```
alanwest, line 185: Do we need the `temporality` option _and_ the option to configure temporality at the `MetricReader`?

specification/metrics/sdk.md
```diff
@@ -672,26 +695,6 @@ The SDK SHOULD provide a way to allow `MetricReader` to respond to
...
-The SDK SHOULD provide a way to allow the preferred [Aggregation
-Temporality](./datamodel.md#temporality) to be specified for a `MetricReader`
-instance during the setup (e.g. initialization, registration, etc.) time. If the
-preferred temporality is explicitly specified then the SDK SHOULD respect that,
-otherwise use Cumulative.
-
-[OpenTelemetry SDK](../overview.md#sdk)
-authors MAY choose the best idiomatic design for their language:
-
-* Whether to treat the temporality settings as recommendation or enforcement.
```
alanwest, line 684: 👍 Good removal. I felt this guidance was ambiguous.

specification/metrics/sdk_exporters/in-memory.md
```diff
@@ -6,16 +6,15 @@ In-memory Metrics Exporter is a [Push Metric
...
-[MetricExporter](../sdk.md#metricexporter).
+If a language provides a mechanism to automatically configure a
+[MetricReader](../sdk.md#metricreader) to pair with the associated
+exporter (e.g., using the [`OTEL_METRICS_EXPORTER` environment
+variable](../../sdk-environment-variables.md#exporter-selection)), by
+default:
+
+* the exporter MUST be paired with a [periodic exporting
+MetricReader](../sdk.md#periodic-exporting-metricreader).
+* the exporter MUST use Cumulative default aggregation temporality.
```
alanwest, line 20: If we agreed that temporality is a `MetricReader` concern and it always defaults to cumulative temporality for all instruments, then do we need this specified at the exporter level? Question applies to all the push based exporters - in-memory, stdout, and OTLP.

specification/metrics/sdk_exporters/otlp.md
```diff
@@ -6,31 +6,32 @@ OTLP Metrics Exporter is a [Push Metric
...
+
+- the exporter MUST be paired with a [periodic exporting
+MetricReader](../sdk.md#periodic-exporting-metricreader).
+- the exporter MUST configure the default aggregation temporality on a
+  per-instrument basis using the
+  `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY` variable as described
+  below, otherwise the exporter MUST use Cumulative as the default
+  aggregation temporality for all instrument kinds.
+
+The `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY` environment variable
```
alanwest, line 29: I'm becoming convinced that this environment variable is not really what we want. I like the idea you're proposing that temporality is a property to configure at the metric reader - not the exporter. Maybe something like the following would work?:

`OTEL_METRICREADER_HISTOGRAM_TEMPORALITY`
`OTEL_METRICREADER_COUNTER_TEMPORALITY`
`OTEL_METRICREADER_ASYNCCOUNTER_TEMPORALITY`
`OTEL_METRICREADER_UPDOWNCOUNTER_TEMPORALITY`
`OTEL_METRICREADER_ASYNCUPDOWNCOUNTER_TEMPORALITY`

alanwest (commented):
Your most recent pass has clarified a lot. The primary thing I'm questioning at this point is if we need to introduce this as a view configuration vs. whether we can just declare temporality as something you can configure with your `MetricReader`

specification/metrics/sdk.md
```diff
@@ -180,10 +180,15 @@ are the inputs:
...
-    outputs metric points that use aggregation temporality (e.g. Histogram,
-    Sum), the SDK SHOULD handle the aggregation temporality based on the
-    temporality of each [MetricReader](#metricreader) instance.
+    apply a [default aggregation](#default-aggregation) configurable on a
+    per-instrument basis according to the [MetricReader](#metricreader) instance.
+  * The aggregation `temporality` (optional) to be used, which may be one of
+    "cumulative", "delta", or "default".  If the configured `aggregation` outputs
+    metric points that define aggregation temporality (e.g. Histogram,
+    Sum), then the SDK MUST export the View using the configured aggregation
+    temporality.  The "default" value means the SDK SHOULD set the default
```
legendecas, line 189: The [data-model guideline for Prometheus](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#otlp-metric-points-to-prometheus) defines that:

> OTLP Metric points to Prometheus:
> If the aggregation temporality is delta and the sum is monotonic, it SHOULD be converted to a cumulative temporality and become a Prometheus Sum

It sounds to me that MetricReader's preferred aggregation temporality should override the per-instrument basis configured aggregation temporality. However, the statement here reads to me as the per-instrument basis configured aggregation temporality should override the one preferred by MetricReader.

specification/metrics/sdk_exporters/otlp.md
alanwest: @pirgeo I'm curious if this would satisfy what you were after in #2065? It's a bit different, as you were originally suggesting this be an exporter setting. Though, given the discussion that's occurred throughout @jmad's attempts to address this issue, would you agree that the ability to configure temporality makes most sense as:

1) a `MetricReader` configuration setting, and
2) a setting that can be tailored per instrument kind?

specification/metrics/sdk.md
alanwest: To clarify, I'm suggesting that the `CumulativeOnly` setting is the only thing I can think of that is a property of a `MetricExporter`. That is, there is no such thing as an exporter with a `DeltaOnly` restriction. Every backend that I'm aware of is either cumulative only (Prometheus) or supports some fine-grained mix of temporality.

specification/metrics/sdk.md
dyladan: I think the data model guidelines suggest that if the exporter receives a delta temporality metric from the SDK it should be converted to a cumulative sum (by saving state within the exporter)

specification/metrics/sdk.md
dyladan: Configuring at the metric reader (or exporter) is needed for backends which have specific requirements for temporality. I'm less certain why the view configuration for temporality is needed. If a metric reader defines that they require delta sums and a user configures a view to make sum use a cumulative temporality, then I guess the exporter is required to save whatever state is needed to convert to the needed temporality?

specification/metrics/sdk.md
dyladan: Your suggestion (1) was at #2707 but it was closed.

jmacd:
I apologize that the state of this PR still seems ambiguous. We are dealing with several questions at once, and I think we're having some difficulty because the relationship between MetricReader and MetricExporter is not precise enough. Then, there's a question about how temporality is configured and/or whether it can be controlled in a view. FYI I agree that per-view temporality selection is not needed, but doing so doesn't resolve the question completely.

I've tried to cast the temporality decision as a MetricReader concern, in the sense that MetricReader is the thing that connects a View with an Exporter. There are two ways in which temporality also "seems like" an Exporter concern, so practically speaking it's a shared concern because (1) the exported protocol may constrain the choice, (2) any discussion about "auto-configuring" an exporter makes it look like the exporter is dictating temporality. I still think default temporality should be configured at the MetricReader--I introduced the term "Associated MetricReader" to make it clear that there's the corresponding reader is how temporality is controlled, so to auto-configure an exporter means configuring the associated MetricReader for the desired temporality.

In that sense, @alanwest I see your question about per-instrument-kind METRICREADER_ environment variables, but I think when discussing environment variables we're still talking about doing so for a specific exporter. So, I would like the environment variable to refer to the exporter of choice, even though we're configuring the associated metric reader. With that said, I don't like the idea of having 15 or 18 new environment variables for every instrument crossed with every push-based exporter. I'd be happy with a(n) environment variable(s) for only the OTLP Exporter.

Another way I've considered addressing the underlying concern is in the data model. If we specify that Non-Monotonic Sum data points SHOULD always be exposed in cumulative temporality unless a reason can be identified, and make sure this guidance overrides the default temporality selection process used for Counter and Histogram instruments.

I will update this PR to remove the View control over temporality, but this won't actually resolve the questions here. Looking for more input, thanks.

specification/metrics/sdk.md
jmacd: Yes.

specification/metrics/sdk.md
```diff
@@ -180,10 +180,15 @@ are the inputs:
...
-    apply a [default aggregation](#default-aggregation). If the aggregation
-    outputs metric points that use aggregation temporality (e.g. Histogram,
-    Sum), the SDK SHOULD handle the aggregation temporality based on the
-    temporality of each [MetricReader](#metricreader) instance.
+    apply a [default aggregation](#default-aggregation) configurable on a
+	per-instrument basis according to the [MetricReader](#metricreader) instance.
+  * The aggregation `temporality` (optional), which may be one of
+    "cumulative", "delta", or "default".  If the aggregation outputs
+    metric points that use aggregation temporality (e.g. Histogram,
+    Sum), then the SDK MUST export this View using this aggregation
```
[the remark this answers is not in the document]
jmacd: Hmm, I used "output" as the verb here. The view controls what the reader reads, and the exporter exports what it reads. Does this sound right?

specification/metrics/sdk.md
jmacd: I've tried to clarify that the defaults we're talking about happen when the exporter is automatically configured with an associated reader, e.g. as with configuration by environment variable.

specification/metrics/sdk.md
jmacd: Right, though because I still see demand for this to be controlled through the environment, and since the bigger problem IMO is the clarify around MetricReader, I decided to take all the feedback into this PR. So at this time, I've removed references to view configuration, meaning temporality can't be controlled per instrument, only per instrument kind. The debate continues in this PR around how to specify the temporality through the environment, please see the updated PR description.

specification/metrics/sdk.md
jmacd: I removed the view configuration part, thanks for feedback.

specification/metrics/sdk.md
jmacd: :grin: Thanks, I do see this paragraph as central to the problem being addressed.

specification/metrics/sdk_exporters/in-memory.md
jmacd: I removed this from each of the exporters, made cumulative the default for all readers to avoid stating the same thing three times.

specification/metrics/sdk_exporters/otlp.md
jmacd: I'm still in favor of a single environment variable, with values like CUMULATIVE, DELTA, and STATELESS. See the updated PR description.

specification/metrics/sdk.md
jmacd: You can imagine a scenario where the exporter is able to tell its constraints and the reader is allowed to configure temporality w/ the exporter's constraints factored in. I feel that we're going too far to prevent users from doing the wrong thing: we need to be sure that auto-configured exporters work correctly, but if a user programmatically configures a reader for temporality that the exporter doesn't support, the exporter should report an error (IMO, by requirement 4 in our [library guidelines](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/library-guidelines.md#requirements).

jmacd:
@open-telemetry/specs-metrics-approvers I've updated this PR based on feedback, have changed the PR description "Clarify MetricReader parameters and temporality controls". The changes: 
- View temporality controls are removed
- MetricReader defaults are called out, per-exporter temporality defaults not needed.

There is still some debate over the form of the environment variable used to select temporality for those vendors that both want to auto-configure the exporter and need to be sure of delta temporality for some instruments. See the updated PR description, thanks.

alanwest:
> I see your question about per-instrument-kind METRICREADER_ environment variables, but I think when discussing environment variables we're still talking about doing so for a specific exporter.

Yes, you're right, thank you - I hadn't thought this 100% through.

> I don't like the idea of having 15 or 18 new environment variables for every instrument crossed with every push-based exporter.

I agree this is not appetizing. My suggestion was in part due to my assumption that folks would object to defining an environment variable with the name `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY` and a setting of delta to mean "Choose Delta aggregation temporality for Counter and Histogram instruments (but choose Cumulative aggregation for UpDownCounter instruments)". Which brings us back to your previous ideas of expanding the range of settings - i.e., `Cumulative`, `DeltaPreferred`, `Stateless`, etc...

Maybe the name of the environment variable is bad and does not accurately convey what we want? Wild idea (dunno if I like it) but what if it were `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_CONFIGURATION` and we defined some way to encode the instrument kind -> temporality mapping in a single string value.

`C=D,AC=C,UDC=C,AUDC=C,H=D` means:
* Counter -> Delta
* AsyncCounter -> Cumulative
* UpDownCounter -> Cumulative
* AsyncUpDownCounter -> Cumulative
* Histogram -> Delta

> Another way I've considered addressing the underlying concern is in the data model.

I'd also be interested in entertaining this. It's kinda where I was going as well with this comment https://github.com/open-telemetry/opentelemetry-specification/pull/2314#issuecomment-1039669860.

jmacd:
>> Another way I've considered addressing the underlying concern is in the data model.
>
> I'd also be interested in entertaining this. It's kinda where I was going as well with this comment https://github.com/open-telemetry/opentelemetry-specification/pull/2314#issuecomment-1039669860.

To spell this out a little more, the idea is the following:

In `datamodel.md`, a new section on the Non-Monotonic Sum data point says essentially that these points are always preferred with Cumulative aggregation temporality if possible.  Consumers are required to "totalize" the Delta NMS into meaningful values, which should only be configured explicitly and for a good purpose.  Totalizing Delta NMS points into Cumulative NMS points later is relatively expensive, whereas totalizing inside an SDK is relatively simple.  Hence, the OTel SDKs will always output synchronous UpDownCounter and Asynchronous UpDownCounter instruments with Cumulative temporality.

This would leave us here with a question about configuring only three instruments.  Now there are only 8 combinations like C=D,AC=C,H=D, and I wonder how many of them are useful.

* CCC: "Always Cumulative"
CCD: You want your Counters Cumulative for historical reasons, but Histograms Delta b/c they're fancy new high-resolution ones?  Maybe.
* CDC: This makes no sense, two stateful translations into the opposite temporality.
* CDD: This makes no sense, same as above.
* DCC: A vendor that supports Delta or Cumulative for Counters, but wants only Cumulative Histograms?  Seems not likely.
* DCD: "Stateless"
* DDC: A vendor that supports only Delta Counters but insists on Cumulative Histograms? Seems not likely.
* DDD: "Always Delta"

To me, only three or four configurations make sense, meaning only 2 or 3 values are needed.

alanwest:
> To me, only three or four configurations make sense, meaning only 2 or 3 values are needed.

Thank you for walking through this so thoroughly - very greatly appreciated. You've convinced me and I agree.

> In `datamodel.md`, a new section on the Non-Monotonic Sum data point says...

I support this new section and I wonder if it needs to be in this PR in order to better justify the revised definition of `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY=DELTA`? That is to say, I do not feel strongly feel it is required, but I wonder if others do.

specification/metrics/sdk.md
```diff
@@ -268,7 +268,7 @@ meter_provider
 meter_provider
     .add_view("X", aggregation=SumAggregation())
     .add_view("*", attribute_keys=["a", "b"])
-    .add_metric_reader(PeriodicExportingMetricReader(AggregationTemporality.DELTA, ConsoleExporter()))
+    .add_metric_reader(PeriodicExportingMetricReader(ConsoleExporter()))
```
alanwest, line 271: Just to clarify, in your mind, it is still possible to associate a ConsoleExporter w/ a reader configured w/ delta temporality, correct? For .NET, this is important specifically for the in-memory exporter because we use it for unit tests.

specification/metrics/sdk.md
```diff
@@ -199,8 +198,10 @@ The SDK SHOULD use the following logic to determine how to process Measurements
 made with an Instrument:
 
 * Determine the `MeterProvider` which "owns" the Instrument.
-* If the `MeterProvider` has no `View` registered, take the Instrument and apply
-    the default configuration.
+* If the `MeterProvider` has no `View` registered, take the Instrument
+    and apply the default Aggregation on the basis of instrument kind
+    according to the [MetricReader](#metricreader) instance's
+    `aggregation` property.
```
pirgeo, line 204: Is the default `aggregation` enough in this case, or does the SDK need the `temporality` as well? It would just tell the SDK to report a monotonic Sum, not that its cumulative or delta. Am I misunderstanding aggregation in this context?

specification/metrics/sdk.md
```diff
@@ -637,12 +637,31 @@ measurements using the equivalent of the following naive algorithm:
...
 * Handling the [ForceFlush](#forceflush) and [Shutdown](#shutdown) signals from
   the SDK.
 
+To construct a `MetricReader` when setting up an SDK, the caller
+SHOULD provide at least the following:
+
+* The `exporter` to use, which is a `MetricExporter` instance.
+* The default output `aggregation` (optional), a function of instrument kind.  If not configured, the [default aggregation](#default-aggregation) SHOULD be used.
+* The default output `temporality` (optional), a function of instrument kind.  If not configured, the Cumulative temporality SHOULD be used.
+* The boolean property `default_enabled` (optional), which determines
```
pirgeo, line 654: I think I might still miss the point of this flag. Would this mean that if there is no explicit view defined for the Instrument, it is dropped? Isn't that the whole point of having default aggregations for each of the instruments? Or does this only apply to temporalities?

specification/metrics/sdk.md
```diff
@@ -763,6 +761,13 @@ from `MetricReader` and start a background task which calls the inherited
 implement so that they can be plugged into OpenTelemetry SDK and support sending
 of telemetry data.
 
+Metric Exporters always have an _associated_ MetricReader.  The
+`aggregation`, `temporality`, `default_enabled` properties used by the
+OpenTelemetry Metric SDK are determined when registering Metric
+Exporters through their associated MetricReader.  OpenTelemetry
+language implementations MAY support automatically configuring the
+[MetricReader](#metricreader) to use for an Exporter.
```
pirgeo, line 769: If the language supports it, it would therefore be enough to set up an `VendorExporter` and that exporter would be able to pass the relevant temporality settings through to the `MetricReader` which in turn connects the Exporter to the MeterProvider, right?

pirgeo (commented):
There are just a few things that I am not quite clear on, so I added a few clarifying questions. 
I agree that we probably do not want dozens of environment variables that describe rather exotic use-cases that might not (yet or ever) exist. If I understand correctly, the environment variable setting a default configuration can be overridden in code (by setting each of the instruments to a specific temporality), which makes me wonder if that doesn't keep the door open to add more default configurations (e.g. STATELESS) at a later stage. To me it seems that this proposal would allow us to go forward with a solution that everybody can use in the way they prefer (e.g. through the launcher mentioned in the PR description), and that is still extensible, if the community deems it necessary, so 👍 from me. 
Other than that, I think this is really great work, so thanks to @jmacd for putting so much thought into this topic.

specification/metrics/sdk.md
```diff
@@ -637,12 +637,31 @@ measurements using the equivalent of the following naive algorithm:
...
+* Collecting metrics from the SDK on demand.
 * Handling the [ForceFlush](#forceflush) and [Shutdown](#shutdown) signals from
   the SDK.
 
+To construct a `MetricReader` when setting up an SDK, the caller
+SHOULD provide at least the following:
+
+* The `exporter` to use, which is a `MetricExporter` instance.
+* The default output `aggregation` (optional), a function of instrument kind.  If not configured, the [default aggregation](#default-aggregation) SHOULD be used.
+* The default output `temporality` (optional), a function of instrument kind.  If not configured, the Cumulative temporality SHOULD be used.
```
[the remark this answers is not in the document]
alanwest: ☝️ this question. It's a much clearer way to ask what I was getting at regarding my question about the ConsoleExporter/in-memory exporter.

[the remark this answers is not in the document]
dyladan: I'm curious why this would be configured by the metric reader and not the exporter. Some exporters like OTLP support both, but other don't and this is just forcing them to implement some sort of conversion in the exporter if the user configures a temporality they don't support. What is the case for a user to provide a temporality other than that expected by the exporter?

specification/metrics/sdk.md
dyladan: aggregation and temporality is not considered here. If `default_enabled=false` and there is no view for a metric it uses the drop aggregation which is not exported and has no temporality.

My question is how does a metric ever get to a metric reader if it is not configured with a view? This seems like a configuration for the MeterProvider not for the MetricReader?

dyladan (commented):
Sorry for late review

specification/metrics/sdk.md
jmacd: There is an existing statement in the View spec, saying:

```
  * If the Instrument could not match with any of the registered `View`(s), the
    SDK SHOULD provide a default behavior. The SDK SHOULD also provide a way for
    the user to turn off the default behavior via MeterProvider (which means the
    Instrument will be ignored when there is no match). Individual
    implementations can decide what the default behavior is, and how to turn the
    default behavior off.
```

This `default_enabled` flag is my interpretation of the requirement about default behavior when no View matches.

> aggregation and temporality is not considered here.

Right, the idea is that this `default_enabled` boolean indicates whether the configured Views are in addition to the default behavior or to the exclusion of default behavior. Configuring a default aggregation to `drop` would impact all the configured views, whereas `default_enabled=false` means no view is in effect, and `default_enabled=true` means there is a default view in effect that uses the default aggregation and default temporality.

specification/metrics/sdk.md
jmacd: Yes. I would add a `WithAggregationTemporality(func(InstrumentKind) Temporality)` option, maybe. 

Would this help?

```
   WithAggregationTemporality(func (_ sdkapi.InstrumentKind) { return aggregation.DeltaTemporality. })
```

I would advise unit tests to be explicit about the temporality they want, only when you auto-configure an exporter is there a need for debate. 

My goal is that you can configure any of the available `aggregation` and `temporality` choices through the reader, this gives e.g., Lightstep the opportunity to configure the "stateless" preference even without standardizing that behavior, by passing the function it wants to the Reader while configuring the OTLP exporter.

specification/metrics/sdk.md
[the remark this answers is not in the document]
jmacd: I'm arguing that the MetricReader is a general purpose adapter for Exporters, so it has full control over  configuration. Default aggregation and temporality are needed only when automatically configuring an exporter, where there should be no potential for misconfiguration.

> this is just forcing them to implement some sort of conversion in the exporter if the user configures a temporality they don't support

I would say the opposite. Exporters should not be in the business of converting temporality, that's the job of a MetricReader. [(Requirement 4 in the library guidelines.)](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/library-guidelines.md#requirements). The fact that a MeterReader can be configured in ways that are nonsense for an exporter is not a big problem, who's going to do that? If you're configuring a Prometheus exporter, you're definitely not going to accidentally override the temporality to a delta preference. If you did, you would see errors about points that could not be exported, I hope.

specification/metrics/sdk.md
```diff
@@ -763,6 +761,13 @@ from `MetricReader` and start a background task which calls the inherited
 implement so that they can be plugged into OpenTelemetry SDK and support sending
 of telemetry data.
 
+Metric Exporters always have an _associated_ MetricReader.  The
+`aggregation`, `temporality`, `default_enabled` properties used by the
+OpenTelemetry Metric SDK are determined when registering Metric
+Exporters through their associated MetricReader.  OpenTelemetry
+language implementations MAY support automatically configuring the
```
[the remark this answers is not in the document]
jmacd: I've tried to remove "preference" from this PR, so let me restate this and see if you agree.

Auto-configured exporters SHOULD dictate their default aggregation and temporality values through exporter-specific logic. When an exporter supports variable temporality, specific accommodation for auto-configuring these values SHOULD be handled on a per-exporter basis. In the case of OTLP, we use an environment variable to control variable temporality. Do we need to configure the choice for Standard error exporters? I'm not sure this choice makes sense for the in-memory exporter anyway since there is no specification on how you access the in-memory state.

specification/metrics/sdk.md
jmacd: Right, I'm imagining that whatever code is used to automatically configure `VendorExporter`, it will also be the custom logic that creates the `MetricReader`. When a user wants to change that configuration, they'll explicitly create their own `MetricReader`.

specification/metrics/sdk_exporters/otlp.md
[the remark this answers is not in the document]
jmacd: I think "COUNT" will be misleading if it applies to histograms as well.

On the other hand, this might be the one place where "preference" makes sense. What do you think about `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE`, with values like "Delta" and "Stateless" which of course defaults to cumulative?

specification/metrics/sdk.md
jmacd: You have it right. In this proposal, the View does not have control over temporality.

specification/metrics/sdk.md
dyladan: I would interpret this as a default view. If the meter provider had a `default_enabled` flag then if it was true the default view would use default aggregation and if the flag was false the default view would use drop aggregation.

specification/metrics/sdk_exporters/otlp.md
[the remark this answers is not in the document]
dyladan: yes I think that works

specification/metrics/sdk.md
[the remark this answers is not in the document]
reyang: > I would say the opposite. Exporters should not be in the business of converting temporality, that's the job of a MetricReader.

+💯

specification/metrics/sdk.md
```diff
@@ -180,10 +180,9 @@ are the inputs:
...
     behavior should respect the Hint if it is available).
   * The `aggregation` (optional) to be used. If not provided, the SDK MUST
-    apply a [default aggregation](#default-aggregation). If the aggregation
-    outputs metric points that use aggregation temporality (e.g. Histogram,
-    Sum), the SDK SHOULD handle the aggregation temporality based on the
-    temporality of each [MetricReader](#metricreader) instance.
+    apply a [default aggregation](#default-aggregation) configurable on the
+    basis of instrument kind according to the [MetricReader](#metricreader)
+	instance.
   * The `exemplar_reservoir` (optional) to use for storing exemplars.
```
ocelotl, line 186: Does this match the changes in the change log?

specification/metrics/sdk.md
```diff
@@ -180,10 +180,9 @@ are the inputs:
     will be used by default (TODO: once the Hint API is available, the default
     behavior should respect the Hint if it is available).
   * The `aggregation` (optional) to be used. If not provided, the SDK MUST
-    apply a [default aggregation](#default-aggregation). If the aggregation
-    outputs metric points that use aggregation temporality (e.g. Histogram,
-    Sum), the SDK SHOULD handle the aggregation temporality based on the
-    temporality of each [MetricReader](#metricreader) instance.
+    apply a [default aggregation](#default-aggregation) configurable on the
+    basis of instrument kind according to the [MetricReader](#metricreader)
+	instance.
```
ocelotl, line 185: Does this match the changes in the change log?

specification/metrics/sdk.md
```diff
@@ -199,8 +198,10 @@ The SDK SHOULD use the following logic to determine how to process Measurements
 made with an Instrument:
 
 * Determine the `MeterProvider` which "owns" the Instrument.
-* If the `MeterProvider` has no `View` registered, take the Instrument and apply
-    the default configuration.
+* If the `MeterProvider` has no `View` registered, take the Instrument
+    and apply the default Aggregation on the basis of instrument kind
```
ocelotl, line 202:
```suggestion
    and apply the default aggregation on the basis of instrument kind
```

dyladan:
I feel like i've misunderstood something fundamental and the confusion i'm feeling about this PR is just my existing misunderstanding or gap in knowledge being surfaced. It is unclear to me what is supposed to be determining the aggregation or temporality used by a given metric. My intuition says that it SHOULD be the exporter as the exporter knows the most about what is supported by whatever backend it is exporting to. For example, the prometheus exporter needs cumulative sums and it wouldn't make any sense for it to receive delta sums in any scenario. But this PR seems to imply that the MetricReader is what determines aggregation and temporality and that this is somehow supposed to be configured by the user? To me this seems to rely heavily on using a launcher to configure the SDK correctly, leaving users who use vanilla OpenTelemetry to figure out a very complex topic on their own. I continue to be confused by the idea that this is controlled by the metric reader or view or user and not the exporter. If an exporter supports multiple temporalities, then it can provide a configuration to users (and a sane default), but for exporters that don't support multiple temporalities then why make it possible to misconfigure it? Maybe i've misunderstood something and this is actually very simple and i'm just not getting it, but I suspect I'm not the only person confused by this.

specification/metrics/sdk_exporters/otlp.md
```diff
@@ -6,31 +6,32 @@ OTLP Metrics Exporter is a [Push Metric
...
+If a language provides a mechanism to automatically configure a
+[MetricReader](../sdk.md#metricreader) to pair with the associated
+Exporter (e.g., using the [`OTEL_METRICS_EXPORTER` environment
+variable](../../sdk-environment-variables.md#exporter-selection)),
+then by default:
+
+- the exporter MUST be paired with a [periodic exporting
+MetricReader](../sdk.md#periodic-exporting-metricreader).
+- the exporter MUST configure the default aggregation temporality on a
+  per-instrument basis using the
```
ocelotl, line 24:
```suggestion
  per-instrument kind basis using the
```
Should it be like this?

specification/metrics/sdk_exporters/otlp.md
```diff
@@ -6,31 +6,32 @@ OTLP Metrics Exporter is a [Push Metric
...
+  aggregation temporality for all instrument kinds.
+
+The `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY` environment variable
+(case-insensitive) defines the default aggregation temporality policy
+to use on the basis of instrument kind.  The recognized values are:
+
+| Value      | Definition                                                                                                    |
+|------------|---------------------------------------------------------------------------------------------------------------|
+| CUMULATIVE | Choose Cumulative aggregation temporality for all instrument kinds.                                           |
+| DELTA      | Choose Delta aggregation temporality for Counter and Histogram instrument kinds, otherwise choose Cumulative. |
```
ocelotl, line 36: What about up down counters? Are they included in Counter?

jmacd: This is part of the reason we're blocked here. UpDownCounters shouldn't be converted to Delta temporality, which all vendors who support Delta temporality have agreed on. When I originally filed #2314 it was because I wanted to be sure that "DELTA" as the name of a preference did not imply converting UpDownCounters into delta temporality under any default circumstances. Please see:

https://github.com/open-telemetry/opentelemetry-specification/pull/2404#issuecomment-1064557650

jmacd:
@dyladan 

> the MetricReader is what determines aggregation and temporality and that this is somehow supposed to be configured by the user?

I am trying to say "MetricReader is what determines aggregation and temporality", but who is responsible for configuring the MetricReader? That depends. If the SDK is auto-configuring an exporter, the SDK is responsible. If the user is configuring an exporter, the user is responsible. 

The module that auto-configures a Prometheus exporter will be sure to provide the correct MetricReader settings for Prometheus. When the user configures a Prometheus exporter, I have trouble imagining them getting this wrong. 

> To me this seems to rely heavily on using a launcher to configure the SDK correctly

I believe you mean this for the OTLP Exporter specifically. This is why the OTLP exporter also has an environment variable, since it's the standard exporter where we agree this matters. Launcher support will not be required for the default (always cumulative) or alternate (always delta) behavior of the OTLP Exporter, and any vendor-specific or 3rd-party Exporter will be in control of the problem for themselves. For Lightstep and the desire to use a "Stateless" configuration for OTLP, this will be handled in a launcher.

dyladan:
I guess that sounds reasonable. So how would a user configure the vanilla OTLP exporter?

like this?

```typescript
provider
	.addMetricReader(
		new PeriodicExportingMetricReader(
			new OTLPMetricExporter(),
			{
				temporality: function (type) {
					if (delta_types.contains(type)) return delta;
					return cumulative;
				},
			},
		)
	)
```

or like this?

```typescript
provider
	.addMetricReader(
		new PeriodicExportingMetricReader(
			new OTLPMetricExporter({
				temporality: function (type) {
					if (delta_types.contains(type)) return delta;
					return cumulative;
				},
			}),
		)
	)
```

jmacd:
I am thinking something like this

```
provider
	.addMetricReader(
                new MetricReaderOptions().
                       withDefaultTemporality(function (kind) {
					if (delta_kinds.contains(kind)) return delta;
					return cumulative;
		       }).
                       withDefaultAggregation(function (kind) {
                                          ...
                       }).
                       withDefaultEnabled(),
		new PeriodicExportingMetricReader(
                        new PeriodicExportingMetricReaderOptions().
                                withIntervalSeconds(30).
                                withTimeoutSeconds(10),
			new OTLPMetricExporter(
                                 new OTLPExporterOptions().
                                 withEndpoint(...),
                         ),
		)
	)
```

mwear (approved):

dyladan (approved):
Very excited to get this in

dyladan:
> I am thinking something like this
> 
> ```
> provider
> 	.addMetricReader(
>                 new MetricReaderOptions().
>                        withDefaultTemporality(function (kind) {
> 					if (delta_kinds.contains(kind)) return delta;
> 					return cumulative;
> 		       }).
>                        withDefaultAggregation(function (kind) {
>                                           ...
>                        }).
>                        withDefaultEnabled(),
> 		new PeriodicExportingMetricReader(
>                         new PeriodicExportingMetricReaderOptions().
>                                 withIntervalSeconds(30).
>                                 withTimeoutSeconds(10),
> 			new OTLPMetricExporter(
>                                  new OTLPExporterOptions().
>                                  withEndpoint(...),
>                          ),
> 		)
> 	)
> ```

Glossing over language specificities this looks like the first snippet I posted and I think it looks good.

pirgeo (approved):
Thanks for clarifying the few things I wasn't sure about, very happy to get this in!

specification/metrics/sdk.md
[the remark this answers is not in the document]
dyladan: I would resolve this but don't have permission.

specification/metrics/sdk_exporters/otlp.md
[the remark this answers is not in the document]
dyladan: I would resolve this but don't have permission.

specification/metrics/sdk_exporters/otlp.md
```diff
@@ -6,31 +6,32 @@ OTLP Metrics Exporter is a [Push Metric
...
+[MetricReader](../sdk.md#metricreader) to pair with the associated
+Exporter (e.g., using the [`OTEL_METRICS_EXPORTER` environment
+variable](../../sdk-environment-variables.md#exporter-selection)),
+then by default:
+
+- the exporter MUST be paired with a [periodic exporting
+MetricReader](../sdk.md#periodic-exporting-metricreader).
+- the exporter MUST configure the default aggregation temporality on a
+  per-instrument basis using the
+  `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY` variable as described
```
jack-berg, line 25: Reading this it appears that `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY` is meant to configure the temporality for all exporters, yet the inclusion of "OTLP" in the name implies the env var applies only to OTLP exporters.

Do we want to have a general purpose env var for configuring temporality? If so, its name should omit "OTLP". If not, this section should probably be rephrased to indicate that exporter specific env vars can influence the temporality of the MetricReader, where `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY` is an example of one case.

specification/metrics/sdk.md
jack-berg: Isn't this `default_enabled=false` option just syntactic sugar for configuring a metric reader `aggregation` function that returns `drop` for all instrument types?

specification/metrics/sdk_exporters/otlp.md
```diff
@@ -6,31 +6,32 @@ OTLP Metrics Exporter is a [Push Metric
...
-section will be merged to the [OpenTelemetry Protocol
-Exporter](../../protocol/exporter.md) specification once it reaches
-[Stable](../../document-status.md)):
+If a language provides a mechanism to automatically configure a
+[MetricReader](../sdk.md#metricreader) to pair with the associated
+Exporter (e.g., using the [`OTEL_METRICS_EXPORTER` environment
+variable](../../sdk-environment-variables.md#exporter-selection)),
+then by default:
+
+- the exporter MUST be paired with a [periodic exporting
```
jack-berg, line 21: #nit:

For consistency with other lists in this doc, can we change `-` -> `*`, and capitalize the first word of each list item?

specification/metrics/sdk_exporters/otlp.md
alanwest: @jack-berg I think this was partly address in this comment https://github.com/open-telemetry/opentelemetry-specification/pull/2404/files#r825020294.

Given the relationship between metric reader and exporter, I think the config needs to be specific to each exporter.

specification/metrics/sdk_exporters/otlp.md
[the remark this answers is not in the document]
alanwest: +1 to a different name. `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` is good by me.

carlosalberto (approved):

specification/metrics/sdk.md
jmacd: I removed `default_enabled`. I think that section of the SDK spec is still ambiguous, but it's not about temporality.

specification/metrics/sdk.md
```diff
@@ -209,12 +210,12 @@ made with an Instrument:
...
-    the user to turn off the default behavior via MeterProvider (which means the
-    Instrument will be ignored when there is no match). Individual
-    implementations can decide what the default behavior is, and how to turn the
-    default behavior off.
+  * If the Instrument could not match with any of the registered
+    `View`(s), the SDK SHOULD determine whether to enable the instrument using
+    the default Aggregation, on the basis of instrument kind,
+    according to the [MetricReader](#metricreader) instance's
+    `aggregation` and `default_enabled` properties.
+>>>>>>> 5f9bb1a426f6f614e5c2ac7dddab8a28a1deeb18
```
reyang, line 218: nit: merge conflict?

specification/metrics/sdk.md
```diff
@@ -637,12 +638,27 @@ measurements using the equivalent of the following naive algorithm:
 
 ## MetricReader
 
-`MetricReader` is an interface which provides the following capabilities:
+`MetricReader` is an SDK implementation object that provides the
```
reyang, line 641: The "is an interface which" wording was copied from https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#span-processor.

I would highly suggest that we keep them consistent unless there is a strong reason not to do so.

specification/metrics/sdk_exporters/otlp.md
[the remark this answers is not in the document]
jmacd: OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE is quite long, but I've applied this suggestion 😁

specification/metrics/sdk.md
```diff
@@ -637,12 +638,27 @@ measurements using the equivalent of the following naive algorithm:
...
 
-* Collecting metrics from the SDK.
+* Collecting metrics from the SDK on demand.
 * Handling the [ForceFlush](#forceflush) and [Shutdown](#shutdown) signals from
   the SDK.
 
+To construct a `MetricReader` when setting up an SDK, the caller
+SHOULD provide at least the following:
+
+* The `exporter` to use, which is a `MetricExporter` instance.
```
reyang, line 652: This might be a problem for Java SIG, the PrometheusExporter in Java is implemented as a MetricReader IIRC.

reyang (approved):

specification/metrics/sdk.md
jmacd: Updated the changelog, thanks.

specification/metrics/sdk_exporters/otlp.md
jmacd: I've added an explicit statement to clarify this concern.

specification/metrics/sdk_exporters/otlp.md
jmacd: Fixed.

specification/metrics/sdk.md
jmacd:
```
    .add_metric_reader(PeriodicExportingMetricReader(ConsoleExporter(),
	          temporality=lambda kind: Delta if kind in [Counter, AsyncCounter, Histogram] else Cumulative)
```

specification/metrics/sdk.md
jack-berg: For what its worth, here's my interpretation:

- By default all instruments are enabled and export with a default view, whose aggregation is dictated by the instrument type. 
- If view(s) are configured and an instrument matches, the view(s)' definition replaces the default view. 
- If view(s) are configured and an instrument doesn't match, the default view is used.
- A metric reader's `aggregation` configuration influences the default view in that it can change the default aggregation type per instruments type.
-  A metric reader could configure an `aggregation` that returns `drop` for every instrument type. In this case, if views are configured and an instrument doesn't match any, `drop` aggregation will be used. If views are configured and an instrument matches, the view's aggregation will be used (which will be `drop` unless it is otherwise configured by the view). 

If you agree, I'm not sure what value we get from the `SDK SHOULD also provide a way for the user to turn off the default behavior via MeterProvider` capability. And actually, java doesn't implement this capability today (and is probably unlikely to without additional clarity / use cases).  If you want to disable all instruments by default, set the reader's `aggregation` to be `drop` for all instruments, and opt-in to the instruments you want by configuring views on a per instrument basis.

specification/metrics/sdk.md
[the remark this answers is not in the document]
jmacd: I think by the library guidelines (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/library-guidelines.md#requirements) exporters SHOULD return errors when given data they can't export, for they are meant to be merely protocol converters.

specification/metrics/sdk.md
jack-berg: You are [correct](https://github.com/open-telemetry/opentelemetry-java/blob/main/exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/PrometheusHttpServer.java#L50). Is modeling the prometheus exporter as a `MetricReader` (as described [here](https://github.com/open-telemetry/opentelemetry-specification/pull/2404/files#diff-32d5f289beed96900fa1e6e69a3ee9cc0b6503ce8883672011c44bc772c0fa2aL934-L943)) incompatible with this phrasing? 

My read is that because the metric reader is also the metric exporter, it doesn't need to the exporter to be provided. Additionally, it can accept an `aggregation` function, but because its prometheus it doesn't accept any `temporality` configuration.

specification/metrics/sdk.md
jmacd: We can continue to discuss this in a separate issue. For this PR I've removed any discussion of `default_enabled` and restored the original text "SDK SHOULD also provide a way for the user to turn off the default behavior via MeterProvider".

specification/metrics/sdk.md
jmacd: I don't see a _real_ incompatibility. If there is one, it ought to be easy to resolve. It looks like exporters sometimes are willing to give you a reader configuration to use (e.g., Prometheus) and exporters sometimes will configure a reader configuration to use (e.g., OTLP).

If you wanted to use Prometheus with a View and wanted to set DROP as the default aggregation, how would you do it? It sounds like you'd have to pass a reader parameter into the Prometheus exporter. It's fine, the Prometheus exporter just wants to self-provide the reader.

specification/metrics/sdk.md
jmacd: I don't feel that we've made a complete interface description for Readers. For example, how does the SDK interact with the Reader? How does the Reader interact with the SDK? 

Span processors are different in this regard. Span processors receive new spans through an interface. How does the Reader receive new metric updates? That's not specified, its an implementation detail of the SDK. That's the distinction I was trying to raise.

specification/metrics/sdk.md
arminru, line 185:
```suggestion
    instance.
```

specification/metrics/sdk.md
```diff
@@ -637,12 +638,27 @@ measurements using the equivalent of the following naive algorithm:
...
+* The default output `aggregation` (optional), a function of instrument kind.  If not configured, the [default aggregation](#default-aggregation) SHOULD be used.
+* The default output `temporality` (optional), a function of instrument kind.  If not configured, the Cumulative temporality SHOULD be used.
```
arminru, line 654: Please add the new changes to https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md#metrics, particularly the new mapping parameters defined here.

specification/metrics/sdk_exporters/otlp.md
```diff
@@ -6,31 +6,32 @@ OTLP Metrics Exporter is a [Push Metric
...
+MetricReader](../sdk.md#periodic-exporting-metricreader).
+* The exporter MUST configure the default aggregation temporality on the
+  basis of instrument kind using the
+  `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` variable as described
+  below, otherwise the exporter MUST use Cumulative as the default
+  aggregation temporality for all instrument kinds.
+
+The `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` environment variable
+(case-insensitive) defines the default aggregation temporality policy
+to use on the basis of instrument kind.  The recognized values are:
```
arminru, line 31:
```suggestion
defines the default aggregation temporality policy
to use on the basis of instrument kind.  The recognized (case-insensitive) values are:
```
To make clear it's about the value, not the name of the env var.

specification/metrics/sdk.md
jmacd: Discussed in today's SIG meeting. The notion of an exporter being the same entity as the reader may be idiomatic in some languages, depending on how class composition works for example. We also covered that the interface between the SDK and the Reader is explicitly unspecified, as these are implementation details.

We think, however, there _could_ potentially be a narrow interface defined between SDK and Reader in the future that allows modifying data as it is processed, by some definition. Therefore, I believe we agreed that it's OK for Exporter/Reader to be the same entity as long as the functions remain distinct. In the example where a Prometheus exporter is also a reader, this likely happens because the exporter inherits the SDK's built-in reader functionality somehow.

specification/metrics/sdk.md
jmacd: We discussed in today's SIG. We are aware of "missing interfaces" that are deliberately not specified between the SDK and the reader, those being informally:

1. a way to "Register" or "Apply" the new Reader, which gives the Reader a way to initiate collection, 
2. a way for the SDK to pass data to the Reader, which the Reader is meant to process.

Both of these are implementation details, so I propose to keep the "SDK implementation object" terminology until these interfaces are standardized.

specification/metrics/sdk.md
jmacd: Done.

specification/metrics/sdk_exporters/otlp.md
```diff
@@ -6,31 +6,31 @@ OTLP Metrics Exporter is a [Push Metric
...
+  aggregation temporality for all instrument kinds.
+
+The `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` environment variable
+defines the default aggregation temporality policy
+to use on the basis of instrument kind.  The recognized (case-insensitive) values are:
+
+| Value      | Definition                                                                                                    |
+|------------|---------------------------------------------------------------------------------------------------------------|
+| CUMULATIVE | Choose Cumulative aggregation temporality for all instrument kinds.                                           |
+| DELTA      | Choose Delta aggregation temporality for Counter, Asynchronous Counter and Histogram instrument kinds, choose Cumulative aggregation temporality for UpDownCounter and Asynchronous UpDownCounter instrument kinds. |
```
cijothomas, line 36: just to confirm - this involving doing necessary CUM -> Delta conversion for async counter, right?

cijothomas (approved):
LGTM.

specification/metrics/sdk_exporters/otlp.md
jmacd: Yes. Let me know if you think more text belongs here or elsewhere on this matter.

cijothomas: The current wording in SDK spec, being removed in this PR, takes the tone that "SDK authors MAY chose to enforce temporality or keep async Counter as CUMULATIVE ignoring the temporality preference".

This PR is silent on it (i.e whether to do CUM -> DELTA), so as with any spec, one should read it as enforcement. In short I read it as "The spec says, if someone configured DELTA, then it means DELTA for async counters, even if it means SDK has to do the CUM -> DELTA conversion".

Might be good to call it out in the changelog - previously spec allowed language SDKs to ignore "temporality", but this PR change disallows it.

Also, given that supplementary guidelines is having a note about this, I am fine with this PR as-is (except a small tweak to changelog entry).
Pasting the supplementary guideline relevant section below: 


> So here are some suggestions that we encourage SDK implementers to consider:
> 
> You probably don't want to encourage your users to do Cumulative->Delta conversion. Actually, you might want to discourage them from doing this.
> If you have to do Cumulative->Delta conversion, and you encountered min/max, rather than drop the data on the floor, you might want to convert them to something useful - e.g. [Gauge](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#gauge).

bogdandrutu: > Might be good to call it out in the changelog - previously spec allowed language SDKs to ignore "temporality", but this PR change disallows it.

I think it is very important for the SDKs to be consistent here, because there are things that you can do with delta that cannot be done with cumulative, so having random behavior between languages for same configuration is a bad experience.

> You probably don't want to encourage your users to do Cumulative->Delta conversion. Actually, you might want to discourage them from doing this.

There are good things, when you want to merge timeseries from multiple sources to convert to delta. I would not be strong in "discourage".

bogdandrutu (approved):
This PR definitely clarifies who is responsible of doing what. I like that MetricExporter does not have a "preferred temporality API" (is this true?) anymore.

jmacd:
@bogdandrutu and @cijothomas please see if [efb040f](https://github.com/open-telemetry/opentelemetry-specification/pull/2404/commits/efb040f649420a4a4fa383bb5d0f404d39973ded) helps.

>  I like that MetricExporter does not have a "preferred temporality API" (is this true?) anymore.

Yes, there's no API like that in the exporter in my understanding. However, when the SDK auto-configures an exporter there may be a need to configure temporality on a case-by-case basis, with the OTLP exporter being a prime example.

specification/metrics/sdk_exporters/otlp.md
jmacd: I've added explicit language and removed the discouraging remark about Cumulative-to-Delta. That remark predates this reasoning behind this PR, in fact we had two vendors explicitly request this.

specification/metrics/sdk.md
```diff
@@ -634,12 +635,36 @@ measurements using the equivalent of the following naive algorithm:
...
+
+The [MetricReader.Collect](#collect) method allows general-purpose
+`MetricExporter` instances to explicitly initiate collection, commonly
+used with pull-based metrics collection.  A common sub-class of
+`MetricReader`, the periodic exporting `MetricReader` SHOULD be provided
+to be used typically with push-based metrics collection.
+
+The `MetricReader` MUST ensure that data points are output in the
```
aabmass, line 659: Is this meant to be normative requirement (MUST)?

This is implying that temporality conversion should be implemented in the MetricReader, but I don't think that's completely necessary.

jmacd: Yes, this is trying to say that the purpose of the MetricReader is to provide the exporter with data in the configured temporality. An exporter could have some reason to perform its own conversion, but it would still have to have a MetricReader providing it with data, and the MetricReader will still have to obey its configuration.

specification/metrics/sdk_exporters/otlp.md
```diff
@@ -6,31 +6,31 @@ OTLP Metrics Exporter is a [Push Metric
...
+variable](../../sdk-environment-variables.md#exporter-selection)),
+then by default:
+
+* The exporter MUST be paired with a [periodic exporting
+MetricReader](../sdk.md#periodic-exporting-metricreader).
+* The exporter MUST configure the default aggregation temporality on the
+  basis of instrument kind using the
+  `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` variable as described
+  below, otherwise the exporter MUST use Cumulative as the default
+  aggregation temporality for all instrument kinds.
```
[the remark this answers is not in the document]
jmacd: This is the specification for the OTLP Exporter specifically. Only the OTLP Exporter uses `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` and only the auto-configured OTLP exporter at that (as we are in a bulleted list inside an "auto-configured" clause).

If we were to have a section here on statsd (which I bet we never will, because there's not a formal specification for statsd), then: 
- You could explicitly configure the MetricReader for any temporality decisions you want. Do you want your UpDownCounters as deltas or as cumulative gauges? Up to you!
- If the SDK were to auto-configure a statsd exporter, then it would apply a reasonable configuration for the MetricReader temporality.
- Assuming we could agree on one reasonable configuration for temporality, the auto-configured statsd exporter would be configured that way.
- If we couldn't agree on one reasonable configuration, a new variable `OTEL_EXPORTER_STATSD_METRICS_TEMPORALITY_PREFERENCE` would be needed, I suppose.

dyladan:
> @bogdandrutu and @cijothomas please see if [efb040f](https://github.com/open-telemetry/opentelemetry-specification/pull/2404/commits/efb040f649420a4a4fa383bb5d0f404d39973ded) helps.
> 
> > I like that MetricExporter does not have a "preferred temporality API" (is this true?) anymore.
> 
> Yes, there's no API like that in the exporter in my understanding. However, when the SDK auto-configures an exporter there may be a need to configure temporality on a case-by-case basis, with the OTLP exporter being a prime example.

This is the only point I'm still not completely sure on. Why would we make users configure the temporality manually with the metric reader rather than the metric reader getting temporality configuration from the exporter?

To be clear I am not trying to block or delay this PR at all with this question, merely understand the reasoning for that change.

specification/metrics/sdk_exporters/otlp.md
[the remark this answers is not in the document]
aabmass: Sorry, I missed that this was OTLP specific. SGTM

specification/metrics/sdk.md
aabmass: Maybe this is irrelevant to the spec. I meant more the other way; in the python sdk, the temporality conversion happens before the data points reach the MetricReader–the MetricReader doesn't do anything besides state the temporality it wants and the SDK does the conversion.

Is that OK? If I were implementing this from scratch, this "MUST" would make me want to hold the memory to convert temporality in the metric reader instead.

cijothomas: The wording are "MetricReader must ensure that...". The actual conversion could be done in SDK itself. 

This is my understanding. I think specifying anything more here would be getting into implementation details..

cijothomas:
> > @bogdandrutu and @cijothomas please see if [efb040f](https://github.com/open-telemetry/opentelemetry-specification/pull/2404/commits/efb040f649420a4a4fa383bb5d0f404d39973ded) helps.
> > > I like that MetricExporter does not have a "preferred temporality API" (is this true?) anymore.
> > 
> > 
> > Yes, there's no API like that in the exporter in my understanding. However, when the SDK auto-configures an exporter there may be a need to configure temporality on a case-by-case basis, with the OTLP exporter being a prime example.
> 
> This is the only point I'm still not completely sure on. Why would we make users configure the temporality manually with the metric reader rather than the metric reader getting temporality configuration from the exporter?
> 
> To be clear I am not trying to block or delay this PR at all with this question, merely understand the reasoning for that change.

There is the following wording in the spec "OpenTelemetry language implementations MAY support automatically configuring the MetricReader to use for an Exporter".
Which I interpret as - If an SDK wants to make it easier for users, it can, by allowing Reader to automatically infer the temporality from the Exporter. Spec is not prohibiting you from doing that. Neither it is enforcing to offer that capability.

Also spec is silent on "How to infer the temporality from an Exporter". I think this is intentional, as languages may have some idiomatic way of expressing it.

jmacd:
@dyladan I like @cijothomas's answer. It does feel natural for an exporter to make available its preferred temporality, but I'm not sure about making it a requirement; auto-configuring exporters would have a degree of consistency that I suspect you're after.

Since we have some exporters where there are more than one reasonable temporality configuration, I feel that forcing the exporter to provide a preference makes more work for exporters, since if the user wants to change the configuration they have to have a way to change every specific exporter's configuration. Since exporters are 1:1 with readers, it seems simpler to specify that readers support this configuration change so that every exporter doesn't have to. Still, I can see adding a mechanism to get the preferred default when auto-configuring exporters. For the OTLP Exporter, it would interpret the environment variable to decide. This sounds like a good design, not sure it has to be specified this way.

specification/metrics/sdk.md
jmacd: @cijothomas wrote:

> The actual conversion could be done in SDK itself.

@aabmass wrote:

> reads like the actual conversion is MetricReader's responsibility from a design standpoint

@reyang commented in https://github.com/open-telemetry/opentelemetry-specification/pull/2404#discussion_r826432494 about my changing `MetricReader` from an "interface" to an "SDK implementation object", which is how I addressed both of the above concerns. The reader is a part of the SDK itself, defined as the thing being responsible for receiving the correct temporality. The `MetricReader`'s primary responsibility (as a part of the SDK) is make sure that temporality conversion happens when it's needed, because Exporters should not be in the business of temporality conversion. It's an SDK implementation detail how temporality conversion happens, of course.

I'd appreciate suggestions for how this sentence could be improved!

aabmass: > The reader is a part of the SDK itself, defined as the thing being responsible for receiving the correct temporality.

That makes sense, but also the spec says that SDKs [may model pull exporters as metric readers](https://github.com/open-telemetry/opentelemetry-specification/blob/158e7ac275432d2b0f425004d96df221c23557fc/specification/metrics/sdk.md#:~:text=they%20could%20model%20the%20pull%20exporter%20as%20MetricReader) (this is what at least Java and Python do right now). In other words, MetricReader is intended to be subclassed (but collect() is considered [part of the SDK](https://github.com/open-telemetry/opentelemetry-specification/blob/158e7ac275432d2b0f425004d96df221c23557fc/specification/metrics/sdk.md#:~:text=Note%3A%20it%20is%20expected%20that%20the%20MetricReader.Collect%20implementations%20will%20be%20provided%20by%20the%20SDK%2C%20so%20it%20is%20RECOMMENDED%20to%20prevent%20the%20user%20from%20accidentally%20overriding%20it%2C%20if%20possible%20(e.g.%20final%20in%20C%2B%2B%20and%20Java%2C%20sealed%20in%20C%23).)). Maybe this is getting outside the scope of this PR.

For wording, what do you think of this?

> The SDK MUST output data points to the MetricReader that are in the configured aggregation temporality for each instrument kind.

reyang: Discussed during the 3/18/2022 triage meeting, we consider this as a nice wording change which doesn't block this PR.

#2424 is created to track this improvement.

bogdandrutu:
@cijothomas 

> Which I interpret as - If an SDK wants to make it easier for users, it can, by allowing Reader to automatically infer the temporality from the Exporter. Spec is not prohibiting you from doing that. Neither it is enforcing to offer that capability.

> Also spec is silent on "How to infer the temporality from an Exporter". I think this is intentional, as languages may have some idiomatic way of expressing it.

I would definitely not do that by having a new concept "preferred" temporality with some values that are not defined by the spec. But that is just me, I would simply expose a `otlp.NewReader()` a "factory method" to create reader + exporter in that exporter package instead.

PR #2406, opened 2022-03-07
Clarify that MetricReader has one-to-one mapping to MeterProvider

### LINKED ISSUE #2072: Clarification on MetricReader registered on MeterProvider

reyang:
Fixes #2072.


legendecas (approved):
👍

jmacd (approved):

specification/metrics/sdk.md
```diff
@@ -652,6 +652,9 @@ that SDK will update the time range - e.g. from (T<sub>n</sub>, T<sub>n+1</sub>]
 to (T<sub>n+1</sub>, T<sub>n+2</sub>] - **ONLY** for this particular
 `MetricReader` instance.
 
+The SDK MUST NOT allow a `MetricReader` instance to be registered on more than
+one `MeterProvider` instances.
```
pirgeo, line 656:
```suggestion
one `MeterProvider` instance.
```

pirgeo (approved):

specification/metrics/sdk.md
```diff
@@ -652,6 +652,9 @@ that SDK will update the time range - e.g. from (T<sub>n</sub>, T<sub>n+1</sub>]
 to (T<sub>n+1</sub>, T<sub>n+2</sub>] - **ONLY** for this particular
 `MetricReader` instance.
 
+The SDK MUST NOT allow a `MetricReader` instance to be registered on more than
+one `MeterProvider` instance.
```
carlosalberto, line 656: Does this mean the SDK has to keep track of the `MetricProvider`/`MetricReader` instances it has created?

(I'd have expected that the user MUST NOT re-use a `MetricReader` instead)

legendecas: AFAICT, the SDK already needs to keep track of metric readers to track their last collection state for aggregation, as the spec is defined that each metric reader's `collect` has no side effects on others.

reyang: > Does this mean the SDK has to keep track of the `MetricProvider`/`MetricReader` instances it has created?

Not necessarily. For example, a certain language implementation could choose to implement this by only allowing `MetricReader` to be provided via a factory method, in this way the SDK could focus on tracking the `MeterProvider`s, and each `MeterProvider` can track their `MetricReader`s.

But from concept level, YES - each MetricReader is essentially having its own state.

carlosalberto (approved):

cijothomas (approved):

PR #2407, opened 2022-03-07
Specify preferred Aggregtion and Aggregation Temporality on a per-instrument basis

### LINKED ISSUE #2403: Preferred aggregation temporality is under-specified

jmacd:
Fixes #2403

## Changes

Specifies that preferred Aggregation and Aggregation Temporality are provided on a per-instrument basis by the Exporter for the Reader's information.

This is a lesser alternative to #2404.
This is another attempt, similar to #2314.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,9 @@ release.
   ([#2317](https://github.com/open-telemetry/opentelemetry-specification/pull/2317)).
 - Clarify that expectations for user callback behavior are documentation REQUIREMENTs.
   ([#2361](https://github.com/open-telemetry/opentelemetry-specification/pull/2361)).
+- Specify that preferred Aggregation and Aggregation Temporality are provided on a 
+  per-instrument basis by the Exporter for the Reader's information.
+  ([#2407](https://github.com/open-telemetry/opentelemetry-specification/pull/2407))
 
 ### Logs
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,8 +31,8 @@ release.
 - Clarify that the periodic metric reader is the default metric reader to be
   paired with push metric exporters (OTLP, stdout, in-memory)
   ([#2379](https://github.com/open-telemetry/opentelemetry-specification/pull/2379)).
-- Specify that preferred Aggregation and Aggregation Temporality are provided on a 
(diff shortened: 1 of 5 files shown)


specification/metrics/sdk.md
```diff
@@ -782,7 +768,21 @@ can run at different schedule, for example:
   pipe.
 
 `MetricExporter` SHOULD provide a way to allow `MetricReader` to retrieve its
-preferred temporality.
+preferred Aggregation and Aggregation Temporality on a per-instrument basis.
```
ocelotl, line 771:
```suggestion
preferred Aggregation and Aggregation Temporality on a per-instrument kind basis.
```

CHANGELOG.md
```diff
@@ -28,6 +28,9 @@ release.
   ([#2317](https://github.com/open-telemetry/opentelemetry-specification/pull/2317)).
 - Clarify that expectations for user callback behavior are documentation REQUIREMENTs.
   ([#2361](https://github.com/open-telemetry/opentelemetry-specification/pull/2361)).
+- Specify that preferred Aggregation and Aggregation Temporality are provided on a 
+  per-instrument basis by the Exporter for the Reader's information.
```
reyang, line 32: per-instrument or per-instrument-kind?

jmacd:
I've updated this PR to reflect a genuinely minimal approach to the apparent problems with "preferred temporality", but I would prefer we focus on #2404.

PR #2413, opened 2022-03-10
For prometheus metrics without sums, use a sum of 0

dashpole:
Related issue: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/7546

## Changes

It is not possible to implement the spec as it is currently written, since the sum cannot be computed from buckets.

The [OpenMetrics histogram spec](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#histogram) says: 

> Negative threshold buckets MAY be used, but then the Histogram MetricPoint MUST NOT contain a sum value as it would no longer be a counter semantically.

We have two options for how to handle histograms without sums:

* Leave the sum as 0
* Drop the Histogram entirely.

Leaving the sum as 0 seems like the better option, even if it can be surprising in some cases.

cc @Aneurysm9 @jmacd

### DIFF
--- a/specification/metrics/datamodel.md
+++ b/specification/metrics/datamodel.md
@@ -1141,7 +1141,7 @@ Multiple Prometheus histogram metrics MUST be merged together into a single OTLP
 * The `le` label on non-suffixed metrics is used to identify and order histogram bucket boundaries. Each Prometheus line produces one bucket count on the resulting histogram. Each value for the `le` label except `+Inf` produces one bucket boundary.
 * Lines with `_count` and `_sum` suffixes are used to determine the histogram's count and sum.
 * If `_count` is not present, the metric MUST be dropped.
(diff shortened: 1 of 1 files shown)


MrAlias (approved):

Aneurysm9 (approved):

jmacd:
I thought we were intending to make the sum field `optional` to address this, which has been done:

https://github.com/open-telemetry/opentelemetry-proto/pull/366

dashpole:
Thanks for pointing that out, Josh.  I wasn't aware of that change, but it makes a lot of sense.  I've updated this to say the sum must be unset.

jack-berg (approved):

jmacd (approved):

cijothomas (approved):

PR #2417, opened 2022-03-15
Clarify a default-enabled behavior for Metric Views

### LINKED ISSUE #2416: Metric instruments: default-enabled clarification

The review thread for #2404 points to lack of clarity around the statement for what happens when Views are configured and none matches the instrument. See the comment thread here: https://github.com/open-telemetry/opentelemetry-specification/pull/2404#discussion_r824744621

The original text leaves a lot of room for interpretation, perhaps too much.

```
   * If the Instrument could not match with any of the registered `View`(s), the
    SDK SHOULD provide a default behavior. The SDK SHOULD also provide a way for
    the user to turn off the default behavior via MeterProvider (which means the
    Instrument will be ignored when there is no match). Individual
    implementations can decide what the default behavior is, and how to turn the
    default behavior off.
```

**Additional context.**

Discussed in the 3/15 Spec SIG. One proposal appears to be favored, which would require no new parameters. If instruments are enabled when no View matches, then users can always fill in a match-all expression to disable instruments using Views to achieve the effect of instruments being disabled-by-default.

jmacd:
Fixes #2416 

## Changes

Clarify that instruments are enabled by default, even when Views are configured. Require support for the match-all View expression having `name=*` to support disabling instruments by default.

### DIFF
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -151,7 +151,9 @@ are the inputs:
   * The `name` of the Instrument(s). [OpenTelemetry SDK](../overview.md#sdk)
     authors MAY choose to support wildcard characters, with the question mark
     (`?`) matching exactly one character and the asterisk character (`*`)
-    matching zero or more characters.
+    matching zero or more characters.  If wildcards are not supported in general, 
+	OpenTelemetry SDKs MUST specifically recognize the single `*` wildcard 
+	as matching all instruments.
   * The `name` of the Meter (optional).
   * The `version` of the Meter (optional).
   * The `schema_url` of the Meter (optional).
@@ -210,11 +212,9 @@ made with an Instrument:
       know (e.g. expose
       [self-diagnostics logs](../error-handling.md#self-diagnostics)).
   * If the Instrument could not match with any of the registered `View`(s), the
-    SDK SHOULD provide a default behavior. The SDK SHOULD also provide a way for
-    the user to turn off the default behavior via MeterProvider (which means the
-    Instrument will be ignored when there is no match). Individual
-    implementations can decide what the default behavior is, and how to turn the
-    default behavior off.
+    SDK SHOULD enable the instrument using the default aggregation and temporality.
+	Users can configure match-all Views using Drop aggregation to disable instruments 
+	by default.
 * END.
 
 Here are some examples:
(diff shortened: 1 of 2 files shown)


cijothomas (approved):
LGTM.

alanwest (approved):

specification/metrics/sdk.md
```diff
@@ -210,11 +212,9 @@ made with an Instrument:
       know (e.g. expose
       [self-diagnostics logs](../error-handling.md#self-diagnostics)).
   * If the Instrument could not match with any of the registered `View`(s), the
-    SDK SHOULD provide a default behavior. The SDK SHOULD also provide a way for
-    the user to turn off the default behavior via MeterProvider (which means the
-    Instrument will be ignored when there is no match). Individual
-    implementations can decide what the default behavior is, and how to turn the
-    default behavior off.
+    SDK SHOULD enable the instrument using the default aggregation and temporality.
+	Users can configure match-all Views using Drop aggregation to disable instruments
```
aabmass, line 216:
```suggestion
    Users can configure match-all Views using [Drop aggregation](#drop-aggregation) to disable instruments
```

aabmass (approved):

reyang (approved):
LGTM.

specification/metrics/sdk.md
dyladan: typical go developer leaving \t all over the place 😆

jack-berg (approved):

specification/metrics/sdk.md
```diff
@@ -211,11 +213,10 @@ made with an Instrument:
...
   * If the Instrument could not match with any of the registered `View`(s), the
-    SDK SHOULD provide a default behavior. The SDK SHOULD also provide a way for
-    the user to turn off the default behavior via MeterProvider (which means the
-    Instrument will be ignored when there is no match). Individual
-    implementations can decide what the default behavior is, and how to turn the
-    default behavior off.
+    SDK SHOULD enable the instrument using the default aggregation and temporality.
+    Users can configure match-all Views using [Drop aggregation](#drop-aggregation)
+    to disable instruments by default.
+* END.
```
bogdandrutu, line 219: What is this END?

mtwo:
Following up from the specification call today: this is ready to merge

specification/metrics/sdk.md
jmacd: Some kind of merge artifact.

jmacd, line 219:
```suggestion
```

PR #2422, opened 2022-03-17
Handle instrumentation scope in the Prometheus conversion spec

dashpole:
Related: https://github.com/open-telemetry/opentelemetry-specification/issues/1906

## Background

[OpenMetrics Naming+Namespacing](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-naming-and-namespaces)

## Changes

This adds "short_name" to the instrumentation scope, which would be an optional, single-word name for the scope. If present, it would be used as the metric prefix in Prometheus exporters. In prometheus receivers, the `opentelemetry_instrumentation_scope` metric would identify prefixes that can be removed from metrics, and used to reconstruct the Instrumentation scope in OTLP.

Alternatives:

* Turn the library name into the single-word metric prefix.  E.g. `goopentelemetryiocontribinstrumentationnethttpotelhttp_http_server_duration`
  * Downside: This is ugly
* Add it as a label to all metrics within the scope: {"instrumentation_scope": "go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"}
  * Downside: This isn't how OpenMetrics is designed to namespace metrics. Removing this later would be breaking for users.

cc @jmacd @Aneurysm9 @jsuereth

### DIFF
--- a/specification/metrics/datamodel.md
+++ b/specification/metrics/datamodel.md
@@ -1325,7 +1325,7 @@ OpenMetrics exemplar unless they would exceed the OpenMetrics
 
 The OpenMetrics prevents naming collisions for metrics from different libraries
 by adding a [metric namespace](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-naming-and-namespaces),
-which is a single-word prefix, to metrics. The [Instrumentation Scope Short Name](TODO)
+which is a single-word prefix, to metrics. The [Instrumentation Scope Short Name](../glossary.md#instrumentation-scope)
 is the equivalent in OpenTelemetry, and MUST be attached as a prefix to all
 metrics from an Instrumentation Scope, if the short name is present. The full
 format of metrics is `<instrumentation scope short_name>_<metric_name>`.
--- a/specification/metrics/datamodel.md
+++ b/specification/metrics/datamodel.md
@@ -1186,8 +1186,38 @@ exemplar as attributes.
 
 #### Instrumentation Scope
 
-Instrumentation Scope MUST be left unset for metrics scraped from Prometheus
-endpoints.
+Each `opentelemetry_instrumentation_scope` info metric point present in a batch
+of metrics MUST be converted to an instrumentation scope, and the
+`opentelemetry_instrumentation_scope` metric must be dropped.  Metrics in the
+batch that have a prefix equal to the `short_name` of the instrumentation scope
+have the matching prefix removed, and are added under that scope.  For example,
+the OpenMetrics metrics:
+
+```
+# TYPE opentelemetry_instrumentation_scope info
+opentelemetry_instrumentation_scope{name=go.opentelemetry.io.contrib.instrumentation.net.http.otelhttp,short_name=otelhttp,version=v0.24.0} 1
+# TYPE otelhttp_http_server_duration counter
+otelhttp_http_server_duration{...} 1
+```
+
+become:
+
+```yaml
+# within a resource_metrics
+scope_metrics:
+  scope:
+    name: go.opentelemetry.io.contrib.instrumentation.net.http.otelhttp
+    short_name: otelhttp
+    version: v0.24.0
+  metrics:
+  - name: http_server_duration
+    data:
+      sum:
+        data_points:
+        - value: 1
+```
+
+Metrics which are not found to be associated with an instrumentation scope are all placed within an empty instrumentation scope, and do not have any prefixes trimmed.
 
 #### Resource Attributes
 
@@ -1293,14 +1323,17 @@ OpenMetrics exemplar unless they would exceed the OpenMetrics
 
 #### Instrumentation Scope
 
-The OpenMetrics equivalent of Instrumentation Scope is the
-[metric namespace](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-naming-and-namespaces),
-which is a single-word prefix for metrics which identifies the source of the
-metric. However, the Instrumentation Scope Name in OpenTelemetry isn't
-generally suitable to use as a prefix because of how verbose it is.  Exporters
-MUST not attach the instrumentation scope name as a prefix or label to metrics.
-The Instrumentation Scope MAY be added as an info-type metric family with the
-name "opentelemetry_instrumentation_scope", and labels "name" and "version".
+The OpenMetrics prevents naming collisions for metrics from different libraries
+by adding a [metric namespace](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-naming-and-namespaces),
+which is a single-word prefix, to metrics. The [Instrumentation Scope Short Name](TODO)
+is the equivalent in OpenTelemetry, and MUST be attached as a prefix to all
+metrics from an Instrumentation Scope, if the short name is present. The full
+format of metrics is `<instrumentation scope short_name>_<metric_name>`.
+
+Additionally, for each Instrumentation Scope with a short name, Prometheus
+exporters SHOULD generate a metric point in the info-typed metric family named
+"otel_scope". It MUST have "name", "version", and "short_name" labels, which
+MAY be empty if the corresponding metric scope field is unset.
 
 #### Resource Attributes
 
--- a/specification/metrics/datamodel.md
+++ b/specification/metrics/datamodel.md
@@ -60,6 +60,7 @@
     + [Dropped Types](#dropped-types)
     + [Start Time](#start-time)
     + [Exemplars](#exemplars-1)
+    + [Instrumentation Scope](#instrumentation-scope)
     + [Resource Attributes](#resource-attributes)
   * [OTLP Metric points to Prometheus](#otlp-metric-points-to-prometheus)
     + [Gauges](#gauges-1)
@@ -69,6 +70,7 @@
     + [Dropped Data Points](#dropped-data-points)
     + [Metric Attributes](#metric-attributes)
     + [Exemplars](#exemplars-2)
+    + [Instrumentation Scope](#instrumentation-scope-1)
     + [Resource Attributes](#resource-attributes-1)
 - [Footnotes](#footnotes)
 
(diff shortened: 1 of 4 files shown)


specification/metrics/datamodel.md
```diff
@@ -1182,6 +1184,11 @@ retrieved from the `trace_id` and `span_id` label keys, respectively.  All
 labels not used for the trace and span ids MUST be added to the OpenTelemetry
 exemplar as attributes.
 
+#### Instrumentation Scope
+
+Instrumentation Scope MUST be left unset for metrics scraped from Prometheus
+endpoints.
```
MrAlias, line 1190: If the `opentelemetry_instrumentation_scope` info metric is present that that not be used for the instrumentation scope?

dashpole: Thats correct.  It remains a separate metric stream to transmit that a given instrumentation library was used, but it doesn't allow you to figure out which metrics it applies to.  I'm not sure it makes much sense, though, without the "short_name" in instrumentation scope.  If we had that piece, then we could reconstruct the Instrumentation Scope here.

Maybe I should just skip this intermediary specification step, and just propose the "short_name" + round-tripping through prometheus...

MrAlias: > Maybe I should just skip this intermediary specification step, and just propose the "short_name" + round-tripping through prometheus...

I'd be interested to hear what other think, but that sounds good to me.

jmacd:
We should be careful to call the OpenMetrics "metric namespace" an equivalent of OTel's instrumentation library. I don't think this is completely true, see below.

## Proposal for round-tripping Instrumentation Scope

I would like to see a treatment of instrumentation scope that could preserve the information for an OTel SDK exporting Prometheus data, which means a Prometheus scrape ideally would produce the same metrics as the SDK would have pushed. I think this is possible.

I've studied the OpenMetrics guidance on target info,

https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#supporting-target-metadata-in-both-push-based-and-pull-based-systems

and consider instrumentation scope information to be in a similar category. Suppose we use an OM info metric:

```
# TYPE otel_scope_info info
# HELP otel_scope_info Instrumentation scope metadata
otel_scope_info{name="some.dns.name/lib",version="0.1",schema_url="standard/url"} 1
```

To round-trip this correctly with an OTel scraper, an OTel SDK would specifically produce one library of metrics at a time. The OM/OTel translation would infer that metrics are organized by instrumentation scope such that the most recent `otel_scope_info` is implicitly associated with metrics in a scoped section. OTLP data scraped from other sources would leave instrumentation scope empty, as you propose.

## Argument in favor of a first-class namespace concept

OpenMetrics writes about metric namespacing "the aim is to keep to a lightweight informal approach" and has three standard instrument constructor fields: "Namespace", "Subsystem", and "Name". With two optional underscore-separated fields and the potential for "Name" itself to contain underscores, there's no way to parse an OM metric name and infer what is namespace, what is subsystem, and what is name. Given no way to distinguish these, "metric namespace" appears to be a user-level option to ensure that when the same kind of instrumentation is produced by multiple distinct instances of the same instrumentation inside a single process, the user can force them not to aggregate or display in the same queries by adding an additional prefix. This would be done as-needed when the user is aware of conflicts and has the intention to solve them using namespaces, otherwise would not be done to keep metric names short. It's

To illustrate when metric namespace appears to solve for an OM user, imagine an OTel SDK configured with two clients speaking to two separate Redis instances with different load patterns, different kinds of data, different purposes. You've got two instances of the Redis client, and two instances of the OTel instrumentation.

Can the user have a way to distinguish the two scopes that works for both OM and OTel? IMO the user needs a first-class namespace concept here, which is not what OTel's instrumentation scope is at this time. Adding an additional property or property list to the Instrumentation Scope message type could help, and we could extend the OM<->OTel translation rules, for example, as follows. Suppose the `MeterProvider` has a new method called `MeterProvider.Namespace(string)` returning a `MeterProvider` where each `Meter` has the associated namespace. OTLP consumers would be advised to recognize this by prefixing metrics names in the section when displayed and queried. 

This is what two instrumentation scopes for the same instrumentation library would look like when scraped, using the OM metric namespace:

```
# TYPE otel_scope_info info
# HELP otel_scope_info Instrumentation scope metadata
otel_scope_info{name="...",version="...",schema_url="...="firstprefix_"} 1

# TYPE firstprefix_http_requests
# HELP firstprefix_http_requests Number of HTTP requests
firstprefix_http_requests{} 1000

# TYPE otel_scope_info info
# HELP otel_scope_info Instrumentation scope metadata
otel_scope_info{name="...",version="...",schema_url="...="secondprefix_"} 1

# TYPE secondprefix_http_requests
# HELP secondprefix_http_requests Number of HTTP requests
secondprefix_http_requests{} 1000
```

dashpole:
Unfortunately, I don't think the "Proposal for round-tripping Instrumentation Scope" complies with the [OM spec](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metricfamily-1): `MetricFamilies MUST NOT be interleaved.`

I haven't actually ever seen Namespace used to distinguish multiple instances of instrumentation--people just seem to use it as a second "subsystem" ([e.g. in k8s](https://github.com/kubernetes/kubernetes/blob/33a2c50bce334467640e016f68cf19e9382ba1a7/staging/src/k8s.io/apiserver/pkg/util/flowcontrol/metrics/metrics.go#L34)), but maybe they were just using it wrong :).  Since prom doesn't have a "View" concept, all metrics i've seen are fully named (including namespace, subsystem, etc.) by the library itself.

I think this is implied by the [OM spec](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-naming-and-namespaces) as well: `the more public a library is the better namespaced its metric names should be` implies that namespacing is performed _by the library_ to keep it from colliding with other libraries.

So I think we are roughly on the same page, but need to figure out if the OM namespace is the equivalent of the instrumentation library, or if it is meant to be more than that.

@brian-brazil, we discussed this briefly at the prom WG yesterday.  Can you help clarify the purpose of the OM namespace?

jmacd:
> Unfortunately, I don't think the "Proposal for round-tripping Instrumentation Scope" complies with the [OM spec](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metricfamily-1): MetricFamilies MUST NOT be interleaved.

I see. A workaround would be to namespace the `otel_scope_info` section divider itself.

```
# TYPE firstprefix_otel_scope_info info
# HELP firstprefix_otel_scope_info Instrumentation scope metadata
firstprefix_otel_scope_info{name="...",version="...",schema_url="...,namespace="firstprefix_"} 1

# TYPE firstprefix_http_requests
# HELP firstprefix_http_requests Number of HTTP requests
firstprefix_http_requests{} 1000

# TYPE secondprefix_otel_scope_info info
# HELP secondprefix_otel_scope_info Instrumentation scope metadata
secondprefix_otel_scope_info{name="...",version="...",schema_url="...,namespace="secondprefix_"} 1

# TYPE secondprefix_http_requests
# HELP secondprefix_http_requests Number of HTTP requests
secondprefix_http_requests{} 1000
```

dashpole:
I updated this to propose the short_name.

tigrannajaryan:
> I'm eager to see us add a list of arbitrary key:values to the scope, it would give us a way to distinguish two instances of the same instrumentation library which today we cannot do.

+1 to this. There are other use cases for this key:value list in the scope too, such as differentiating the type of data emitted from the scopes that belong to different data domains, e.g. profiling data emitted as log records or client-side data emitted as log records needs to be differentiated so that it can be easily routed and processed differently in the backends. We don't have a good way to handle this today.

tigrannajaryan:
Given the [comments](https://github.com/open-telemetry/opentelemetry-specification/pull/2422#issuecomment-1077775654) about key-value lists in the Scope should the addition of ShortName be paused for now and instead we make it a semantic convention to be recorded in the key-value list?

dashpole:
Yes.  I'll close this for now, and will reopen after we've introduced the key-value list.

joaopgrassi:
@tigrannajaryan @dashpole now that we have the scope attributes, should we bring this topic back, and define the semantic convention for scope attribute(s)?

tigrannajaryan:
> @tigrannajaryan @dashpole now that we have the scope attributes, should we bring this topic back, and define the semantic convention for scope attribute(s), where `short_name` would be the first?

Yes, feel free to make a proposal (or @dashpole feel free to do it).

We may need changes to semantic convention generator tooling to support scopes.

joaopgrassi:
Unfortunately I don't have the cycles to work on this now, but I quickly tried it out and I think we don't need to change the convention generator. 

Here's what I tried:
https://github.com/open-telemetry/opentelemetry-specification/compare/main...dynatrace-oss-contrib:opentelemetry-specification:feature/scope_attributes_semconv?expand=1

We could have a document inside `specification/common/scope.md` for the common scope attributes (`short_name`) for ex. Then, if there's scope attributes specific for each signal, those can be defined inside their appropriate folder like: `semantic_conventions/metrics/scope-metrics.yaml` -> `specification/metrics/semantic_conventions/scope-metrics.md`.

dashpole:
Yes, I plan to work on this

joaopgrassi:
Do we have an issue for it? Also, do you folks have an idea on how would we name this? We already have this page that states `otel.scope.name` and `otel.scope.version`. Would `short_name` also be mapped under the same? `otel.scope.short_name`? Or would it be under a new thing with just `scope.short_name`?

tigrannajaryan:
> Do we have an issue for it? Also, do you folks have an idea on how would we name this? We already have [this page ](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.12.0/specification/common/mapping-to-non-otlp.md#instrumentationscope)that states `otel.scope.name` and `otel.scope.version`. Would `short_name` also be mapped under the same? `otel.scope.short_name`? Or would it be under a new thing with just `scope.short_name`?

I don't think anything is decided yet. I would not assume `otel.scope.*` is the right approach. Should all scope attribute names start with `otel.scope`? Is there a reason the names cannot start from the root namespace?

joaopgrassi:
Yeah I don't particularly have any answers, but was rather fishing for your ideas. I think we probably should have a new issue to discuss this topic further and lay out a good foundation for other attributes to come. I also don't think we need `otel.scope.*` as those are a bit "special".

joaopgrassi:
I created an issue #2682 so we have a forum to discuss it. :)

PR #2427, opened 2022-03-22
Remove the spec status from the metrics README.md

reyang:
Editorial change.

The Metrics `README.md` serves as a landing page with links to the specs, the doc itself is not a spec (it doesn't have any MUST/SHOULD/MAY wording).


arminru (approved):

MrAlias (approved):

yurishkuro (approved):

specification/metrics/README.md
```diff
@@ -1,7 +1,5 @@
 # OpenTelemetry Metrics
 
-**Status**: [Experimental](../document-status.md)
```
bogdandrutu, line 3: I would also remove the "programatic model" which enters into too many details, and maybe replace it with a statement about datamodel.

cijothomas (approved):

jmacd (approved):

specification/metrics/README.md
reyang: Agree, I've removed this section (I can add a data model statement in a separate PR, just to keep this PR scoped).

reyang: Unresolving this, actually this section is referenced by other docs and removing it will increase the scope of this PR. I'll investigate more.

reyang:
> We seem to have an odd divergence between signals here, e.g. specification/trace/ has no readme, while metrics/README is not only a navigation page but also includes design goals and other info, which to me makes it part of the spec.

@yurishkuro Thanks! Feedback taken. https://github.com/open-telemetry/opentelemetry-specification/pull/2427#discussion_r834951009 also suggested that we might need to consider this doc as part of the spec. I'll do more investigation.

specification/metrics/README.md
bogdandrutu: Should we start an issue? Or do a separate PR for that remove?

reyang: Tracked by #2467

Issue #2419, opened 2022-03-16
semconv:  rpc.type indicate server or client?


1046102779:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/rpc.md#attributes

How do we use semantic convention:   `rpc`  to represent  server and client?

arminru:
Hi @1046102779!
The OpenTelemetry Tracing API already defines dedicated properties for both a "Span Kind" and a "Status" on Spans.
Take a look here to learn more:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#spankind
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-status

1046102779:
> https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/rpc.md#attributes

But how does metrics use tracing's `SpanKind` and `status`?

arminru:
I misunderstood that this was about metrics, not tracing - sorry!
In this case it could make sense to introduce generic attributes for both a ~span~ "party/counterpart" kind (likely with the same values as defined for SpanKind) and an operation status so they can also be used for HTTP, FaaS, messaging and others.
What do you think? Would you like to drive spec PRs for these?

1046102779:
> In this case it could make sense to introduce generic attributes for both a ~span~ "party/counterpart" kind (likely with the same values as defined for SpanKind) and an operation status so they can also be used for HTTP, FaaS, messaging and others.
> What do you think? Would you like to drive spec PRs for these?

Yes. I think it's a general concept and it's appropriate to add to the opentelemetry specification, and I'm willing to drive this PR.

arminru:
@1046102779 I just looked over the semantic conventions you were referring to again. In your screenshot, only the common metric **attributes** are shown. The metrics (i.e., the **metric names**) themselves already distinguish between client- and server-side observations (e.g., `rpc.server.duration` vs `rpc.client.duration`), however.
Because of that, there is no need for any attribute to distinguish clients from servers, right?

If additional RPC metrics were introduced, they would stick to this naming scheme with server/client in the namespace and this would also be the only way to have disjoint observations. Otherwise you would only be able to get meaningful results when aggregating metrics if you filter or split by the respective attribute.

[...]

PR #2430, opened 2022-03-23
feature/rpc: add rpc.type and rpc.status for metrics.

1046102779:
https://github.com/open-telemetry/opentelemetry-specification/issues/2419

### DIFF
--- a/semantic_conventions/trace/rpc.yaml
+++ b/semantic_conventions/trace/rpc.yaml
@@ -4,6 +4,28 @@ groups:
     brief: 'This document defines semantic conventions for remote procedure calls.'
     events: [rpc.message]
     attributes:
+      - id: type
+        requires: always
+        brief: 'An RPC request consists of a caller and a callee, the name of the type is server or client.'
+        type:
+          allow_custom_values: false
+          members:
+            - id: server
+              value: 'server'
+            - id: client
+              value: "client"
+      - id: status
+        requires: always
+        brief: 'Status only represents network and a RPC internal execution state in both client and server, not bussiness logic.'
+        type:
+          allow_custom_value: false
+          members:
+            - id: Unset
+              value: 'Unset'
+            - id: OK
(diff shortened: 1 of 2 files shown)


trask:
hi @1046102779! I liked @arminru's https://github.com/open-telemetry/opentelemetry-specification/issues/2419#issuecomment-1074961559 proposing these as general metric attributes that can be used across all semantic conventions

semantic_conventions/trace/rpc.yaml
```diff
@@ -4,6 +4,28 @@ groups:
     brief: 'This document defines semantic conventions for remote procedure calls.'
     events: [rpc.message]
     attributes:
+      - id: type
```
reyang, line 7: For traces, the [span kind](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#spankind) already has this information.

semantic_conventions/trace/rpc.yaml
```diff
@@ -4,6 +4,28 @@ groups:
...
+        requires: always
+        brief: 'An RPC request consists of a caller and a callee, the name of the type is server or client.'
+        type:
+          allow_custom_values: false
+          members:
+            - id: server
+              value: 'server'
+            - id: client
+              value: "client"
+      - id: status
```
reyang, line 17: Same here, the [span status](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-status) has covered this.

reyang (changes requested):
I understand that the intention is to have something similar like [http.status_code](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md#attributes) and such intention makes total sense to me. However, I thinking adding this to `semantic_conventions/trace/rpc.yaml` is going to introduce problem to traces as it introduces to span kind/status?

PR #2436, opened 2022-03-24
Add process.cpu.utilization metric

trask:
## Changes

Adds `process.cpu.utilization` metric with definition similar to `system.cpu.utilization`.


bogdandrutu (commented):
LGTM from semantic perspective. I still share the concern about exposing this metric from multiple language instrumentations instead of exposing it from the collector.

trask:
> LGTM from semantic perspective. I still share the concern about exposing this metric from multiple language instrumentations instead of exposing it from the collector.

I opened #2439 to discuss this, hoping we can table the concern for this specific PR.

jmacd (approved):

arminru (approved):

bogdandrutu (approved):
Per the decision to discuss what is the expectation in terms of instrumentation in a separate issue.

PR #2440, opened 2022-03-24
Clarify how metric metadata and type suffixes are handled

### LINKED ISSUE #2437: Define OTLP -> Prometheus metric name conversion rule

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#otlp-metric-points-to-prometheus defined OTLP -> Prometheus metric conversion on instrument type, attributes, exemplars and resource attributes. But the conversion rule for instrument name is missing.

Like, [Prometheus](https://prometheus.io/docs/practices/naming/#:~:text=...should%20have%20a%20suffix%20describing%20the%20unit%2C%20in%20plural%20form.%20Note%20that%20an%20accumulating%20count%20has%20total%20as%20a%20suffix%2C%20in%20addition%20to%20the%20unit%20if%20applicable.) defines that instrument units should be concatenated to the metric name. Should this apply in the conversion of OTLP -> Prometheus?

**Additional context.**

Originally posted by @aabmass https://github.com/open-telemetry/opentelemetry-js/pull/2824#pullrequestreview-919670937

dashpole:
Fixes #2437

## Changes

Explicitly call out how to handle prometheus metric metadata and OTLP metric metadata.  Most of this was implicit, but this makes it more explicit.  This PR also clarifies that `_total` suffixes are trimmed from metrics, which is already the case in the collector.

It proposes the following changes from the current state: 

* trim `_info` suffixes from info metrics (after support for info is implemented: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/8516)
* Add unit suffixes to prometheus exporters.
* Trim unit suffixes in the prometheus receiver.

Unit suffixes are required by the [OpenMetrics spec](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#unit)

cc @aabmass @legendecas @jmacd @Aneurysm9


specification/metrics/datamodel.md
```diff
@@ -1219,6 +1229,14 @@ in keys).
 
 ### OTLP Metric points to Prometheus
 
+#### Metric Metadata
+
+The Unit of an OTLP metric point MUST be added as [OpenMetrics UNIT metadata](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metricfamily). Additionally, the unit MUST be added as a suffix to the metric name, and SHOULD use [base units](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#units-and-base-units) recommended by OpenMetrics where possible.  The unit suffix
```
aabmass, line 1234: > SHOULD use base units

Is this applicable to the conversion spec since choice of units is more to do with the instrumentation? Or are you saying that units SHOULD be converted to base units when possible?

aabmass: typo

```suggestion
The Unit of an OTLP metric point MUST be added as [OpenMetrics UNIT metadata](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metricfamily). Additionally, the unit MUST be added as a suffix to the metric name, and SHOULD use [base units](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#units-and-base-units) recommended by OpenMetrics where possible.
```

aabmass (approved):

specification/metrics/datamodel.md
dashpole: Sorry I didn't finish the sentence.  It is now complete: `The unit suffix comes before any type-specific suffixes`

dashpole: I'm saying that units SHOULD be converted to base units when possible.

specification/metrics/datamodel.md
```diff
@@ -1122,17 +1124,32 @@ OpenTelemetry metric data. Since OpenMetrics has a superset of Prometheus' types
...
+
+The [OpenMetrics TYPE metadata](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metricfamily),
+if present, MUST be used to determine the OTLP data type, and dictates
+type-specific conversion rules listed below. Metric families without type
+metadata follow rules for [unknown-typed](#unknown-typed) metrics below.
+
 #### Counters
 
-A [Prometheus Counter](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#counter) MUST be converted to an OTLP Sum with `is_monotonic` equal to `true`.
+A [Prometheus Counter](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#counter) MUST be converted to an OTLP Sum with `is_monotonic` equal to `true`.  If the counter has a `_total` suffix, it MUST be removed.
```
reyang, line 1144: What if the counter name is `foobar_total_total`? Do we end up with `foobar` (which might introduce conflicts) or `foobar_tatal`?

reyang (approved):
LGTM, there might be some corner cases such as https://github.com/open-telemetry/opentelemetry-specification/pull/2440/files#r834957165 but I guess it's probably a rare case.

legendecas (approved):

specification/metrics/datamodel.md
dashpole: You would end up with `foobar_total`.  Do you think "a `_total` suffix" is specific enough, or do you think it should be further clarified?

jmacd (approved):

specification/metrics/datamodel.md
```diff
@@ -1219,6 +1236,22 @@ in keys).
 
 ### OTLP Metric points to Prometheus
 
+#### Metric Metadata
+
+The Unit of an OTLP metric point MUST be added as
+[OpenMetrics UNIT metadata](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metricfamily).
+Additionally, the unit MUST be added as a suffix to the metric name, and SHOULD
+be converted to[base units](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#units-and-base-units)
```
MrAlias, line 1244:
```suggestion
be converted to [base units](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#units-and-base-units)
```

MrAlias (approved):

specification/metrics/datamodel.md
reyang: It's good enough 👍 (I guess there is still a low chance that people might interpret it as "keep removing until there is no `_total` suffix" but maybe fine).

Aneurysm9 (commented):
Does this need to specify how to handle metrics with the '_created' suffix?

dashpole:
> Does this need to specify how to handle metrics with the '_created' suffix?

I noticed that when I looked through the spec as well.  IIUC, that would give us the start timestamp of the metric, which would be helpful.  I still need to think through how to use it (e.g. will we generate an additional metric with a `_count` suffix in exporters even if the scraped metric didn't have one?).

reyang:
> > Does this need to specify how to handle metrics with the '_created' suffix?
> 
> I noticed that when I looked through the spec as well. IIUC, that would give us the start timestamp of the metric, which would be helpful. I still need to think through how to use it (e.g. will we generate an additional metric with a `_count` suffix in exporters even if the scraped metric didn't have one?).

This can be followed up in a separate PR if it turns out to be needed.

PR #2442, opened 2022-03-25
[editorial] Fix formatting of OTLP/HTTP's two status lines

chalin:


specification/protocol/otlp.md
```diff
@@ -320,7 +320,7 @@ The default network port for OTLP/gRPC is 4317.
 
 ### OTLP/HTTP
 
-**Binary Format Status**: [Stable](../document-status.md)
+**Binary Format Status**: [Stable](../document-status.md)<br>
```
reyang, line 323: I guess inline HTML is in general discouraged in markdown files. Could we add an empty line instead?

chalin: Note that `<br>` is used elsewhere in the spec, so I'm not setting a precedent here:

> image

While I might be able to understand the _general_ principle of discouraging use of arbitrary inline HTML, judicious use of `<br>` would seem reasonable IMHO. (Note that we _can_ write this in pure Markdown by adding trailing spaces to the line, but I personally feel that that is a Markdown syntax that is to be avoided).

If you feel strongly about it, I can remove the `<br>` and add a blank line or (my preference) turn those two lines into a list:

> image

arminru: @chalin In the tables above `<br>` is used because there is no other way (that I'd be aware of) to add a line break within a table cell. I think both a blank line or a list should be fine. I agree that adding the double trailing spaces is not ideal and some extensions would even trim these which might result in someone accidentally removing it again in their next PR.

chalin: Right. So what is your preference in this case?

chalin:
I've submitted an update, that removes the `<br>` and combines both statues into a single line like this:

> image

WDYT?

jmacd (approved):

tigrannajaryan (approved):

reyang (approved):

PR #2446, opened 2022-03-25
Move net.peer.name from http common to http client

### LINKED ISSUE #2366: Move net.peer.name from common http attributes section to just client http attributes section

Help instrumenters avoid the problem we've run into in the [opentelemetry-java-instrumentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation) project with triggering reverse name lookups (https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5297, https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5305, https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5393).

Given the problems this causes, I think it's worth moving `net.peer.name` from common http attributes section to just the client http attributes section (and add an explanation why it's not recommended for http server spans).

trask:
Fixes #2366

## Changes

Adds note that `net.peer.name` SHOULD NOT be set if capturing it would require an extra reverse DNS lookup. And moves `net.peer.name` from common http attributes to just client http attributes.


specification/trace/semantic_conventions/http.md
```diff
@@ -146,13 +144,20 @@ If set, `http.url` must be the originally requested URL,
...
+| [`net.peer.name`](span-general.md) | string | Remote hostname or similar, see note below. | `example.com` | See below |
 
 **Additional attribute requirements:** At least one of the following sets of attributes is required:
 
 * `http.url`
 * `http.scheme`, `http.host`, `http.target`
 * `http.scheme`, [`net.peer.name`](span-general.md), [`net.peer.port`](span-general.md), `http.target`
 * `http.scheme`, [`net.peer.ip`](span-general.md), [`net.peer.port`](span-general.md), `http.target`
+
+Following attributes MUST be provided **at span creation time** (when provided at all), so they can be considered for sampling decisions:
```
yurishkuro, line 158: This clause is not merely moving to another section, this is a new normative requirement.

trask: @yurishkuro I don't believe it is a new normative requirement, `net.peer.name` was already marked as `sampling_relevant: true`, and was previously under a similar section above

semantic_conventions/trace/http.yaml
```diff
@@ -125,6 +123,9 @@ groups:
     extends: http
     span_kind: client
     brief: 'Semantic Convention for HTTP Client'
+    attributes:
+      - ref: net.peer.name
+        sampling_relevant: true
```
arminru, line 128: Adding a `note:` explaining why it's not recommended for http server spans like you proposed in #2366 would be really helpful I think.

mateuszrzeszutek (approved):

semantic_conventions/trace/http.yaml
arminru: It could make sense to add this right at the place where it's defined in `general.yaml` and not only in the context of HTTP, WDYT?
Note that there is already a more extensive note in the prelude here to which you could refer:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#netname-attributes=

trask: makes sense to me, ptal

jack-berg (approved):

semantic_conventions/trace/general.yaml
```diff
@@ -47,6 +47,8 @@ groups:
         type: string
         brief: 'Remote hostname or similar, see note below.'
         examples: 'example.com'
+        note: >
+          `net.peer.name` SHOULD NOT be set if capturing it requires a reverse DNS lookup.
```
Oberon00, line 51:
```suggestion
          `net.peer.name` SHOULD NOT be set if capturing it requires a reverse DNS lookup just for telemetry purposes.
```

Not sure if that adds or resolves confusion, but what I'd like to clarify: If you do a reverse lookup anyway, it's fine to also set it as net.peer.name, but you shouldn't do a reverse lookup just in order to set it.

semantic_conventions/trace/http.yaml
arminru: LGTM with the clarification proposed in https://github.com/open-telemetry/opentelemetry-specification/pull/2446#discussion_r839600757

arminru (approved):

SergeyKanzhelev (approved):

reyang (approved):

PR #2452, opened 2022-03-29
span exporter concurrency: expand on Export() requirements

### LINKED ISSUE #2434: Batch span processor support for multiple pending exports

Higher throughput from the batch span processor when using an asynchronous exporter (especially when talking to a remote backend as opposed to a local collector).

Having discussed this with the Java folks, I believe this boils down to:

* Clarity on whether the [batch span processor](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#batching-processor) should allow multiple pending exports, and if not, I'd like to propose a new configuration parameter `maxPendingExports` in the spec to support multiple pending exports.

**Additional context**

The Java SDK's span exporter returns a promise, and the batch span processor waits up to `exportTimeoutMillis` for the promise to complete before proceeding to export the next batch.

It would be great to hear if other language SDKs have had any different interpretation and have implemented the batch span processor to already allow multiple pending exports, since that will help in planning a path forward here.

@open-telemetry/cpp-approvers @open-telemetry/dotnet-approvers @open-telemetry/erlang-approvers @open-telemetry/go-approvers @open-telemetry/javascript-approvers @open-telemetry/python-approvers @open-telemetry/ruby-approvers

tsloughter:
Fixes #2434

## Changes

Additional context around what is and what isn't required of the Span Exporter `Export()` calls.


specification/trace/sdk.md
```diff
@@ -615,15 +615,19 @@ function are typically expected to serialize and transmit the data to the
 destination.
 
 Export() will never be called concurrently for the same exporter instance.
-Export() can be called again only after the current call returns.
+Export() can be called again only after the current call returns. This does not
+mean Export() is required to block until the batch is successfully sent. What
+Export() does before returning is language and exporter dependent.
```
reyang, line 620: Do we need to change the return value section? Below is what we have today, and the `For protocol exporters this typically means that the data is sent over the wire and delivered to the destination server` statement might generate confusion with the changes proposed in this PR.

**Returns:** ExportResult:

ExportResult is one of:

* `Success` - The batch has been successfully exported.
  For protocol exporters this typically means that the data is sent over
  the wire and delivered to the destination server.
* `Failure` - exporting failed. The batch must be dropped. For example, this
  can happen when the batch contains bad data and cannot be serialized.

trask: @reyang does the note below that section cover your concern?

> Note: this result may be returned via an async mechanism or a callback, if that is idiomatic for the language implementation.

trask:
> Fixes #2434

Thanks!

For this to fully fix #2434, at least for Java, we also need to clarify the meaning of the batch span processor's `exportTimeoutMillis`. Because currently Java's batch span processor interprets `exportTimeoutMillis` as the amount of time to wait for the "promise" returned by the span exporter. And since this has default value of 30 seconds, the batch span processor basically forces synchronous behavior even when using an underlying async/concurrent exporter.

Similar to retries and concurrency, it seems like timeout may also be better handled by the underlying exporter (e.g. OTLP exporter already has `OTEL_EXPORTER_OTLP_TIMEOUT`).

This is tricky water though since batch span processor is a stable component, so definitely looking forward to more opinions here.

specification/trace/sdk.md
tsloughter: Actually, yea, I don't think the async mechanism of callback covers this. What does technically cover it is that it says it is only "typically" and not required to mean the data is sent over the wire. Which I took to mean it can simply mean the data has been queued by the exporter to be exported in the next available request (based on the limit on the number of concurrent requests the exporter can send).

tsloughter:
@trask good point. I think the timeout still makes sense but just also needs more details. It keeps the batch processor from stalling completely due to a malfunctioning exporter.

I suppose it is the same person configuring the batch processor and the exporter so maybe it is overkill to not just require the exporter have a timeout..

tsloughter:
@trask have you asked anyone else in the Java SIG about why in the spec matrix it says concurrent sending is supported?

trask:
> @trask have you asked anyone else in the Java SIG about why in the spec matrix it says concurrent sending is supported?

The Java OTLP exporters themselves supports concurrent sending (and return promises to track each export's completion). It's the batch span processor that is the problem and does the blocking before proceeding on to the next batch (because of its interpretation of `exportTimeoutMillis`).

tsloughter:
> The Java OTLP exporters themselves supports concurrent sending (and return promises to track each export's completion). It's the batch span processor that is the problem and does the blocking before proceeding on to the next batch (because of its interpretation of `exportTimeoutMillis`).

Aah, ok.

specification/trace/sdk.md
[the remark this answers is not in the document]
tsloughter: > I think you're trying to say that Exporters have control over the "Definition of Done", and that from the SDK's perspective the BSP is required to block while avoiding concurrency until the exporter says it's Done for each batch of spans.

Correct.

I can add that the Exporter should document what it does before returning from an `export` call.

As for if OTLP exporter should have an option for blocking/non-blocking, I'm not sure. My initial thought, was "no", but mainly because it adds additional complexity to implement and options for the user to sort through without a user benefit. But if there is a benefit I just can't think of at the moment that'd change my opinion :).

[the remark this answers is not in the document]
trask: > I think you're trying to say that Exporters have control over the "Definition of Done", and that from the SDK's perspective the BSP is required to block while avoiding concurrency until the exporter says it's Done for each batch of spans.

can we consider that concurrency only applies to the execution of the `export()` function?

if an exporter _does not_ support concurrency, it should block and not return from the `export()` function until it is "done"

if an exporter _does_ support concurrency, it can return early from the `export()` function and use some kind of promise or callback to notify when it is "done"

in both cases, a caller (including the BSP) would be allowed to call `export()` again as soon as soon as the `export()` function returns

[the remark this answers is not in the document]
tsloughter: > if an exporter _does not_ support concurrency, it should block and not return from the `export()` function until it is "done"

I could support including that I think. My only concern is it requires the BSP to handle both possible returns instead of just one. 

> if an exporter does support concurrency, it can return early from the export() function and use some kind of promise or callback to notify when it is "done"

Is this necessary? The processor has no need to track if an export is complete for a batch since it does not handle retries and does not limit concurrency.

But I did realize the exporter needs to be able to provide some back pressure by blocking `export()` if the limit on concurrent sends it is configured for has been reached.

Then if the BSP's export timeout is reached during an export after the exporter has switched to a blocking `export()` would result in load shedding by dropping that batch -- probably with the requirement that the max batch size has been met, allowing it to instead collect more spans if it was triggered by the interval and not the batch being full.

I'll also add that to this PR if you two think it sounds right.

The way it is now it is up to the exporter to be dropping when overloaded and it feels more appropriate that the user be able to configure the size and time limit for dropping batches in the BSP instead of the exporter, leaving the exporter configuration to focus on the amount of concurrency and the timeout for individual requests.

tsloughter:
Ugh. Thought it would notify me again before closing since I updated it after it was marked stale...

And apparently you can't re-open PRs when closed by this bot? 

Won't bother re-opening as another branch right away since there are still changes to be made but I still need more feedback to make those.

carlosalberto:
@tsloughter Just reopened this ;)

tsloughter:
@carlosalberto thanks!

tsloughter:
I will get this updated this week and try to get feedback on the updates related to back pressure.

specification/trace/sdk.md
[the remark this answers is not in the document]
tsloughter: I think this "challenge" needs to be left to a future PR and this PR should stick to the limited case of making it explicit that concurrency is allowed.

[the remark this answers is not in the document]
reyang: How about this:

By default, the SDK SHOULD NOT make reentrant invocations to `Export()` on an exporter. The SDK MAY provide a mechanism to allow exporters that have higher reentrancy guarantee (e.g. safe under concurrent calls, safe under recursive calls) to expose such capability, so the SDK can invoke `Export()` with less contentions for better performance.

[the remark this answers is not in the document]
reyang: Here goes an example https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/bb01f8167b6a49d595f8bc055f5fc3f8552c1e6f/src/OpenTelemetry.Exporter.Geneva/ReentrantExportProcessor.cs

I think if we allow exporters to explicitly express "I can be called without contention", this is how one could do it in .NET:

```csharp
internal class MyExporter : BaseExporter<Activity>
{
    [ConcurrencyMode(ConcurrencyMode.Reentrant)]
    public override ExportResult Export(in Batch<Activity> batch)
    {
        return ExportResult.Success;
    }
}
```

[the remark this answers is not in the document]
tsloughter: @reyang I think that is too specific. I don't know exactly what reentrant means in the context of a .NET method but is this meant to say that `Export()` should be able to safely be called concurrently?

Maybe it should just say that each implementation must document the required concurrency characteristics of any exporter.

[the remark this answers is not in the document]
reyang: > Maybe it should just say that each implementation must document the required concurrency characteristics of any exporter.

This sounds good to me 👍

Essentially, we're pushing the responsibility to the implementations. Given exporters are tight to the SDK (e.g. an exporter cannot target different implementations because there is no abstraction/interface).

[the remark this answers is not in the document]
tsloughter: Updated.

tsloughter:
I fought with it and don't really like it still -- largely because the result of `Export` is still said to be based on if the sending of the data was successful and it still says `Export` won't be called concurrently, which it sounds like we wanted to be more up to the implementation?

The changes started to grow when I tried to fix those parts and I decided to scrap it and just make the minimal update.

tsloughter:
Looks like the failure is dead links in other parts of the docs.

tsloughter:
Added a fix for the metrics datamodel links in a separate commit here.

specification/trace/sdk.md
```diff
@@ -614,16 +614,22 @@ Protocol exporters that will implement this
...
 
-Any retry logic that is required by the exporter is the responsibility
-of the exporter. The default SDK SHOULD NOT implement retry logic, as
-the required logic is likely to depend heavily on the specific protocol
-and backend the spans are being sent to.
+Concurrent requests and retry logic is the responsibility of the exporter. The
+default SDK's Span Processors SHOULD NOT implement retry logic, as the required
+logic is likely to depend heavily on the specific protocol and backend the spans
+are being sent to. For example, the [OpenTelemetry Protocol (OTLP)
+specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md)
```
reyang, line 631:
```suggestion
specification](../protocol/otlp.md)
```

reyang (approved):
LGTM with a changelog entry.

jack-berg (approved):

reyang:
@jmacd @trask would you review the latest proposal here? Thanks!

MrAlias (approved):

trask (commented):
@open-telemetry/java-maintainers if this is merged, would we change the OTLP exporter behaviors to match? (and going back in time, would we have changed `export()` to not return a promise?)

trask:
> @open-telemetry/java-maintainers if this is merged, would we change the OTLP exporter behaviors to match? (and going back in time, would we have changed `export()` to not return a promise?)

to clarify, I'm supportive of this PR, I think this is the key:

> Concurrent requests and retry logic is the responsibility of the exporter. The
default SDK's Span Processors SHOULD NOT implement retry logic, as the required
logic is likely to depend heavily on the specific protocol and backend the spans
are being sent to.

and given that retry logic could even include storing to disk and retrying later, which would be problematic for returning a promise anyways

jmacd (approved):

specification/trace/sdk.md
```diff
@@ -635,11 +641,10 @@ e.g. for spans in Java it will be typically `Collection<SpanData>`.
 
 ExportResult is one of:
 
-* `Success` - The batch has been successfully exported.
-  For protocol exporters this typically means that the data is sent over
-  the wire and delivered to the destination server.
-* `Failure` - exporting failed. The batch must be dropped. For example, this
-  can happen when the batch contains bad data and cannot be serialized.
+* `Success` - The batch has been successfully passed to the Exporter.
```
jkwatson, line 644: I think this is a bad change. All this would mean in java, at least, is that the method call has returned, making it completely meaningless in practice, and there would never be a Failure response unless something is very very strange.

jsuereth: What else would you want this to be?  I do agree that often you won't see a Failure response.  However if you WERE using a fixed-size buffer (or maximum for outstanding requests) for your exporter, then you would.

In practice, though, what do you expect the SDK / Processor to do with this failure.   I believe the way the specification is worded ANY response leads to dropping the batch, and all retry logic is assumed to live in the exporter.  

I guess my question here for @jkwatson -  What use cases are we dropping with this change that we'd need to recover?

jkwatson: This will effectively force all of our exporters to be "fire and forget", meaning that the BatchSpanProcessor won't be able to wait until the exporter says a batch is complete before sending another one. I think this is a bad change that will not help anyone. How does this change provide any more value than what we already have? If an exporter wants to be "fire and forget", it certainly can immediately return 'success', but if it doesn't, then we're hamstrung it with this change.

jkwatson: I specifically thinking of the "future" type response that we use in the Java interfaces. Exporters are free to complete the future however they want...they should not be forced to complete them as 'success' before they've actually tried to send anything.

jack-berg: I've removed my approval because I overlooked these changes to ExportResult.  I agree with the clarifications about export concurrency, but the changes to ExportResult removes the ability to have a synchronous exporter and buys nothing in return. It's already possible to implement a concurrent exporter using the existing semantics by returning a ExportResult of `Success` after enqueuing the records.

tsloughter: Ooh, my bad, I see now that this latest change did make it more restrictive. I didn't mean for this. I think it should still just be a requirement that Export return without blocking on sending the request.

tsloughter: @jack-berg it is not possible today to implement a concurrent exporter because enqueuing the records is not what the spec says the exporter must do. It says they have been sent when a Success is returned.

My reading of the way it is today even says Java can't call Export again until that Future has returned Success.

tsloughter: I don't know if we should even define what Export returns, esp since Java returns a Future which returns a Success/Failure. I know it says in the spec that the Sucess/Failure can be sent another way other than the function call result, but it is confusing to read when the spec talks about it not being called concurrently and then notes the return of the function may not actually mean anything was done (since it could be returning a future or calling a callback at some point later).

What about just saying that the Exporter must be given a way to signal completion of a batch being sent to the processor that is language specific? That Success/Failure continue to mean the batch has fully failed (thinking about retries by the exporter here) or has been successfully sent.

lalitb: We have been discussing this in C++ SIG,  and do support these specs change for
 - It allows exporters to do concurrent uploads. This allows for handling the loss of events where the exporter is waiting for an HTTP response, the batch log/span processor is still receiving more events and the queue in it grows quickly and full eventually. Again, the maximum concurrent uploads can be a configurable parameter so can be switched off if required.
 - There is not much the processor can do with the upload failure result (other than logging/dispatching the error), as the retry logic is also part of the exporter.

lalitb (approved):
Approving it just to vote for these changes, even though my approval doesn't count :)

specification/trace/sdk.md
tsloughter: Thanks @lalitb 

I tried again and think it still allows for everything you mention @lalitb as well as hopefully covers the problems raised in this thread.

reyang (approved):

tsloughter:
Changelog updated.

Issue #1670, opened 2021-05-04
Add semantic convention for describing graphql spans


svrnm:
**What are you trying to achieve?**
>GraphQL is an open-source data query and manipulation language for APIs, and a runtime for fulfilling queries with existing data. [...] GraphQL servers are available for multiple languages, including Haskell, JavaScript, Perl, Python, Ruby, Java, C++, C#, Scala, Go, Rust, Elixir, Erlang, PHP, R, D and Clojure. ([Source: Wikipedia](https://en.wikipedia.org/wiki/GraphQL))

There are already graphql instrumentation libraries for Node.JS and Ruby (https://opentelemetry.io/registry/?s=graphql) and as quoted above there could be many more. GraphQL has a [specification](https://spec.graphql.org/June2018/) that could be used to define span attributes. For example the Node.JS instrumentation defines the following:

```
export enum SpanAttributes {
  COMPONENT = 'graphql',
  SOURCE = 'graphql.source',
  FIELD_NAME = 'graphql.field.name',
  FIELD_PATH = 'graphql.field.path',
  FIELD_TYPE = 'graphql.field.type',
  OPERATION = 'graphql.operation.name',
  VARIABLES = 'graphql.variables.',
  ERROR_VALIDATION_NAME = 'graphql.validation.error',
}
```

The most important ones would be the operation name (query, mutatation, ...) and if available the query name.

ericmustin:
Was there any reason this stalled? The stronger we can make the semantic conventions around instrumentation attributes, the better, especially for these details which are already implemented across multiple languages?

spencerwilson:
Just want to point to these two issues folks raised in JS that involve GraphQL semconv:
- https://github.com/open-telemetry/opentelemetry-js-contrib/issues/534
- https://github.com/open-telemetry/opentelemetry-js-contrib/issues/705

I don’t see any acute reason this should be blocked, so I assume it’s stalled simply for lack of someone making a PR.

jord1e:
I can look into contributing a PR this weekend, some prior art to be looked at is the Apollo schema reporting Protobuf file (https://github.com/apollographql/apollo-server/blob/e468367d52e11f3127597e4fe920eb8294538289/packages/apollo-reporting-protobuf/src/reports.proto)

And of course the standing opentelemetry-js(-contrib) and opentelemetry-ruby conventions

[...]

PR #2456, opened 2022-03-29
Add semantic conventions for GraphQL

laurit:
## Changes

Add semantic conventions for GraphQL span name and attributes.

Related issues #1670

### DIFF
--- a/semantic_conventions/trace/instrumentation/graphql.yml
+++ b/semantic_conventions/trace/instrumentation/graphql.yml
@@ -1,0 +1,20 @@
+groups:
+  - id: graphql
+    prefix: graphql
+    brief: >
+      This document defines semantic conventions to apply when instrumenting the GraphQL Server. They map GraphQL query
+      to attributes on a Span.
+    attributes:
+      - id: graphql.operation.name
+        brief: "The name of the query being executed."
+        type: string
+        examples: 'findBookById'
+      - id: graphql.operation.type
+        brief: "The name of the operation being executed."
+        type: string
+        examples: ['QUERY', 'MUTATION', 'SUBSCRIPTION']
+      - id: graphql.source
+        brief: "The GraphQL query being executed."
+        type: string
+        note: The value may be sanitized to exclude sensitive information.
+        examples: 'query findBookById { bookById(id: ?) { name } }'
--- a/semantic_conventions/trace/instrumentation/graphql.yml
+++ b/semantic_conventions/trace/instrumentation/graphql.yml
@@ -2,8 +2,8 @@ groups:
   - id: graphql
     prefix: graphql
     brief: >
-      This document defines semantic conventions to apply when instrumenting the GraphQL Server. They map GraphQL
-      operations to attributes on a Span.
+      This document defines semantic conventions to apply when instrumenting the GraphQL implementation. They map
+      GraphQL operations to attributes on a Span.
     attributes:
       - id: operation.name
         brief: "The name of the operation being executed."
--- a/semantic_conventions/trace/instrumentation/graphql.yml
+++ b/semantic_conventions/trace/instrumentation/graphql.yml
@@ -2,8 +2,8 @@ groups:
   - id: graphql
     prefix: graphql
     brief: >
-      This document defines semantic conventions to apply when instrumenting the GraphQL Server. They map GraphQL queries
-      to attributes on a Span.
+      This document defines semantic conventions to apply when instrumenting the GraphQL Server. They map GraphQL
+      operations to attributes on a Span.
     attributes:
       - id: operation.name
         brief: "The name of the operation being executed."
@@ -13,7 +13,7 @@ groups:
         brief: "The type of the operation being executed."
         type: string
         examples: ['QUERY', 'MUTATION', 'SUBSCRIPTION']
-      - id: query
+      - id: operation.body
         brief: "The GraphQL query being executed."
         type: string
         note: The value may be sanitized to exclude sensitive information.
--- a/semantic_conventions/trace/instrumentation/graphql.yml
+++ b/semantic_conventions/trace/instrumentation/graphql.yml
@@ -2,18 +2,18 @@ groups:
   - id: graphql
     prefix: graphql
     brief: >
-      This document defines semantic conventions to apply when instrumenting the GraphQL Server. They map GraphQL query
+      This document defines semantic conventions to apply when instrumenting the GraphQL Server. They map GraphQL queries
       to attributes on a Span.
     attributes:
-      - id: graphql.operation.name
+      - id: operation.name
         brief: "The name of the query being executed."
         type: string
         examples: 'findBookById'
-      - id: graphql.operation.type
+      - id: operation.type
         brief: "The name of the operation being executed."
         type: string
         examples: ['QUERY', 'MUTATION', 'SUBSCRIPTION']
-      - id: graphql.source
+      - id: source
         brief: "The GraphQL query being executed."
         type: string
         note: The value may be sanitized to exclude sensitive information.
--- a/semantic_conventions/trace/instrumentation/graphql.yml
+++ b/semantic_conventions/trace/instrumentation/graphql.yml
@@ -6,14 +6,14 @@ groups:
       to attributes on a Span.
     attributes:
       - id: operation.name
-        brief: "The name of the query being executed."
+        brief: "The name of the operation being executed."
         type: string
         examples: 'findBookById'
       - id: operation.type
-        brief: "The name of the operation being executed."
+        brief: "The type of the operation being executed."
         type: string
         examples: ['QUERY', 'MUTATION', 'SUBSCRIPTION']
-      - id: source
+      - id: query
         brief: "The GraphQL query being executed."
         type: string
         note: The value may be sanitized to exclude sensitive information.
--- a/semantic_conventions/trace/instrumentation/graphql.yml
+++ b/semantic_conventions/trace/instrumentation/graphql.yml
@@ -24,8 +24,8 @@ groups:
               value: "subscription"
               brief: "GraphQL subscription"
         examples: ['query', 'mutation', 'subscription']
-      - id: operation.body
-        brief: "The GraphQL query being executed."
+      - id: document
+        brief: "The GraphQL document being executed."
         type: string
         note: The value may be sanitized to exclude sensitive information.
         examples: 'query findBookById { bookById(id: ?) { name } }'
--- a/semantic_conventions/trace/instrumentation/graphql.yml
+++ b/semantic_conventions/trace/instrumentation/graphql.yml
@@ -11,8 +11,19 @@ groups:
         examples: 'findBookById'
       - id: operation.type
         brief: "The type of the operation being executed."
-        type: string
-        examples: ['QUERY', 'MUTATION', 'SUBSCRIPTION']
+        type:
+          allow_custom_values: false
+          members:
+            - id: query
+              value: "query"
+              brief: "GraphQL query"
+            - id: mutation
+              value: "mutation"
+              brief: "GraphQL mutation"
+            - id: subscription
+              value: "subscription"
+              brief: "GraphQL subscription"
+        examples: ['query', 'mutation', 'subscription']
       - id: operation.body
         brief: "The GraphQL query being executed."
         type: string
--- a/specification/trace/semantic_conventions/instrumentation/graphql.md
+++ b/specification/trace/semantic_conventions/instrumentation/graphql.md
@@ -11,11 +11,11 @@ span SHOULD be named `<graphql.operation.type>`. When `<graphql.operation.type>`
 MAY be used as span name.
 
 <!-- semconv graphql -->
-| Attribute  | Type | Description  | Examples  | Required |
+| Attribute  | Type | Description  | Examples  | Requirement Level |
 |---|---|---|---|---|
-| `graphql.operation.name` | string | The name of the operation being executed. | `findBookById` | No |
-| `graphql.operation.type` | string | The type of the operation being executed. | `query`; `mutation`; `subscription` | No |
-| `graphql.document` | string | The GraphQL document being executed. [1] | `query findBookById { bookById(id: ?) { name } }` | No |
+| `graphql.operation.name` | string | The name of the operation being executed. | `findBookById` | Recommended |
+| `graphql.operation.type` | string | The type of the operation being executed. | `query`; `mutation`; `subscription` | Recommended |
+| `graphql.document` | string | The GraphQL document being executed. [1] | `query findBookById { bookById(id: ?) { name } }` | Recommended |
 
 **[1]:** The value may be sanitized to exclude sensitive information.
 
--- a/specification/trace/semantic_conventions/instrumentation/graphql.md
+++ b/specification/trace/semantic_conventions/instrumentation/graphql.md
@@ -2,15 +2,22 @@
 
 **Status**: [Experimental](../../../document-status.md)
 
-This document defines semantic conventions to apply when instrumenting the GraphQL Server. They map GraphQL queries to
-attributes on a Span.
+This document defines semantic conventions to apply when instrumenting the GraphQL Server. They map GraphQL operations
+to attributes on a Span.
+
+**Span kind:** MUST always be `INTERNAL`.
+
+The **span name** MUST be of the format `<graphql.operation.type> <graphql.operation.name>` provided that
+`graphql.operation.type` and `graphql.operation.name` are available. If `graphql.operation.name` is not available, the
+span SHOULD be named `<graphql.operation.type>`. When `<graphql.operation.type>` is not available, `GraphQL Operation`
+MAY be used as span name.
 
 <!-- semconv graphql -->
 | Attribute  | Type | Description  | Examples  | Required |
 |---|---|---|---|---|
 | `graphql.operation.name` | string | The name of the operation being executed. | `findBookById` | No |
 | `graphql.operation.type` | string | The type of the operation being executed. | `QUERY`; `MUTATION`; `SUBSCRIPTION` | No |
-| `graphql.query` | string | The GraphQL query being executed. [1] | `query findBookById { bookById(id: ?) { name } }` | No |
+| `graphql.operation.body` | string | The GraphQL query being executed. [1] | `query findBookById { bookById(id: ?) { name } }` | No |
 
 **[1]:** The value may be sanitized to exclude sensitive information.
 <!-- endsemconv -->
--- a/specification/trace/semantic_conventions/instrumentation/graphql.md
+++ b/specification/trace/semantic_conventions/instrumentation/graphql.md
@@ -8,9 +8,9 @@ attributes on a Span.
 <!-- semconv graphql -->
 | Attribute  | Type | Description  | Examples  | Required |
 |---|---|---|---|---|
-| `graphql.graphql.operation.name` | string | The name of the query being executed. | `findBookById` | No |
-| `graphql.graphql.operation.type` | string | The name of the operation being executed. | `QUERY`; `MUTATION`; `SUBSCRIPTION` | No |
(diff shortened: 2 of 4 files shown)


specification/trace/semantic_conventions/instrumentation/graphql.md
```diff
@@ -0,0 +1,16 @@
+# Semantic conventions for GraphQL Server
+
+**Status**: [Experimental](../../../document-status.md)
+
+This document defines semantic conventions to apply when instrumenting the GraphQL Server. They map GraphQL query to
+attributes on a Span.
```
trask, line 6:
```suggestion
This document defines semantic conventions to apply when instrumenting the GraphQL Server. They map GraphQL queries to
attributes on a Span.
```

specification/trace/semantic_conventions/instrumentation/graphql.md
```diff
@@ -0,0 +1,16 @@
...
+
+<!-- semconv graphql -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `graphql.graphql.operation.name` | string | The name of the query being executed. | `findBookById` | No |
+| `graphql.graphql.operation.type` | string | The name of the operation being executed. | `QUERY`; `MUTATION`; `SUBSCRIPTION` | No |
+| `graphql.graphql.source` | string | The GraphQL query being executed. [1] | `query findBookById { bookById(id: ?) { name } }` | No |
```
trask, line 13: `graphql.graphql.` -> `graphql.`

specification/trace/semantic_conventions/instrumentation/graphql.md
laurit: thanks

specification/trace/semantic_conventions/instrumentation/graphql.md
```diff
@@ -0,0 +1,16 @@
...
+
+This document defines semantic conventions to apply when instrumenting the GraphQL Server. They map GraphQL queries to
+attributes on a Span.
+
+<!-- semconv graphql -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `graphql.operation.name` | string | The name of the query being executed. | `findBookById` | No |
+| `graphql.operation.type` | string | The name of the operation being executed. | `QUERY`; `MUTATION`; `SUBSCRIPTION` | No |
+| `graphql.source` | string | The GraphQL query being executed. [1] | `query findBookById { bookById(id: ?) { name } }` | No |
```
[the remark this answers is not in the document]
laurit: I only used `graphql.source` because that is what was used in js and java (which copied from js), was hoping that someone would suggest to rename this.
As far as I can tell standard GraphQL server allows only one operation. Execution fails when input query contains multiple operations and operation name is not specified.

specification/trace/semantic_conventions/instrumentation/graphql.md
```diff
@@ -0,0 +1,16 @@
+# Semantic conventions for GraphQL Server
+
+**Status**: [Experimental](../../../document-status.md)
+
+This document defines semantic conventions to apply when instrumenting the GraphQL Server. They map GraphQL queries to
+attributes on a Span.
+
```
mateuszrzeszutek, line 7: What about the span name? Can we specify `${graphql.operation.type} ${graphql.operation.name}` as the expected span name for graphQL spans?

trask: Also a note about span kind would be good, since I think both Java and JavaScript(?) are capturing these spans as `Internal`, which could be non-obvious to spec readers.

semantic_conventions/trace/instrumentation/graphql.yml
```diff
@@ -0,0 +1,20 @@
+groups:
+  - id: graphql
+    prefix: graphql
+    brief: >
+      This document defines semantic conventions to apply when instrumenting the GraphQL Server. They map GraphQL queries
```
svrnm, line 5: >  when instrumenting the GraphQL Server.

Is this really only applicable to the server? what if someone talks to a non-instrumented graphql backend and wants to use the same semantics at the client side?

svrnm, line 5: >  when instrumenting the GraphQL Server.

Is this really only applicable to the server? what if someone talks to a non-instrumented graphql backend and wants to use the same semantics at the client side?

svrnm:
Thanks for creating this @laurit !

semantic_conventions/trace/instrumentation/graphql.yml
```diff
@@ -0,0 +1,20 @@
...
+      to attributes on a Span.
+    attributes:
+      - id: operation.name
```
yurishkuro, line 8: how is this related to the span name?

specification/trace/semantic_conventions/instrumentation/graphql.md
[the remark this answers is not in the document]
laurit: changed `graphql.query` -> `graphql.operation.body` and replaced `query` with `operation` in descriptions.

mateuszrzeszutek (approved):

specification/trace/semantic_conventions/instrumentation/graphql.md
laurit: Added span kind and span name.

specification/trace/semantic_conventions/instrumentation/graphql.md
```diff
@@ -0,0 +1,23 @@
+# Semantic conventions for GraphQL Server
+
+**Status**: [Experimental](../../../document-status.md)
+
+This document defines semantic conventions to apply when instrumenting the GraphQL Server. They map GraphQL operations
+to attributes on a Span.
+
+**Span kind:** MUST always be `INTERNAL`.
```
yurishkuro, line 8: Q: why is this a requirement to create a new span instead of annotating the existing client/server spans?

laurit: My understanding is that adding these attributes to an existing server span could make sense when you have built an api that consumes graphql. At least in the graphql java framework there is nothing that requires it to be used that way. In my opinion adding these attributes to an existing server span would not make sense if multiple graphql operations are executed during one request and when the same request performs other operations besides executing qraphql.

trask: > My understanding is that adding these attributes to an existing server span could make sense when you have built an api that consumes graphql

maybe relax the requirement to allow it to be either an INTERNAL or SERVER span?

svrnm: what's the purpose of this limitation? as per my comment above, it is possible that someone wants to use this semantic conventions in a client side request, especially when the downstream service is not instrumented/owned by a 3rd party

semantic_conventions/trace/instrumentation/graphql.yml
svrnm: I would prefer `implementation` yes. By that there are no implicit assumptions where those semantics are used.

laurit: Replaced `server` with `implementation`

specification/trace/semantic_conventions/instrumentation/graphql.md
laurit: Removed span kind.

specification/trace/semantic_conventions/instrumentation/graphql.md
```diff
@@ -0,0 +1,21 @@
+# Semantic conventions for GraphQL Server
```
trask, line 1: based on discussion that this could be used for clients too

```suggestion
# Semantic conventions for GraphQL
```

trask (commented):
does anyone have thoughts on dropping `operation` from the attribute names, e.g.

* `graphql.name`
* `graphql.type`
* `graphql.body`

trask (approved):

svrnm:
> does anyone have thoughts on dropping `operation` from the attribute names, e.g.

I would keep it, there are other _things_ in graphql that might have a name, type or body (not an expert on that, but by quickly reading the [GraphQL](https://spec.graphql.org/October2021/) spec I saw that there are also things like type definitions, fragments, response. Also, their spec could extend to contain other objects worth collecting)

While those other things are not yet specified here, one might do it in the future and naming it `graphql.name|type|body` would be limiting that effort.

specification/trace/semantic_conventions/instrumentation/graphql.md
```diff
@@ -0,0 +1,21 @@
...
+MAY be used as span name.
+
+<!-- semconv graphql -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `graphql.operation.name` | string | The name of the operation being executed. | `findBookById` | No |
+| `graphql.operation.type` | string | The type of the operation being executed. | `QUERY`; `MUTATION`; `SUBSCRIPTION` | No |
+| `graphql.operation.body` | string | The GraphQL query being executed. [1] | `query findBookById { bookById(id: ?) { name } }` | No |
+
+**[1]:** The value may be sanitized to exclude sensitive information.
```
jamesmoessis, line 20: In the currently available instrumentation, is there any sanitization? The operation body seems like something that instrumentation might not have enough context to sanitize.

laurit: There is sanitation in java instrumentation and as far as I can tell in js also.

trask:
/unstale

@yurishkuro @svrnm @jamesmoessis any other thoughts on this?

svrnm (approved):

svrnm:
@trask, thanks, I was under the assumption that this was closed already. I'm fine with it!

jamesmoessis (approved):

trask:
@open-telemetry/technical-committee any thoughts on the bar for merging this? thx

specification/trace/semantic_conventions/instrumentation/graphql.md
```diff
@@ -0,0 +1,21 @@
...
+The **span name** MUST be of the format `<graphql.operation.type> <graphql.operation.name>` provided that
+`graphql.operation.type` and `graphql.operation.name` are available. If `graphql.operation.name` is not available, the
+span SHOULD be named `<graphql.operation.type>`. When `<graphql.operation.type>` is not available, `GraphQL Operation`
+MAY be used as span name.
+
+<!-- semconv graphql -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `graphql.operation.name` | string | The name of the operation being executed. | `findBookById` | No |
+| `graphql.operation.type` | string | The type of the operation being executed. | `QUERY`; `MUTATION`; `SUBSCRIPTION` | No |
```
tigrannajaryan, line 17: Is type expected to be capitalized? Graphql spec uses lowercase. Is this a value from a finite set (enum of 3 possible values) or arbitrary values are possible?

tigrannajaryan:
@spencerwilson @ericmustin @jord1e please review.

specification/trace/semantic_conventions/instrumentation/graphql.md
tigrannajaryan: There is a way to limit the values to a finite enum, see for example https://github.com/open-telemetry/opentelemetry-specification/blob/031630c818c60666b27764a5b7a0e4ed435f55c4/semantic_conventions/trace/messaging.yaml#L102

tigrannajaryan: I think using the lowercase like Graphql spec does would be a bit more preferable.

laurit: Done

jord1e:
Looks good, but shouldn't we add tracing for parsing and validation steps? And maybe for individual fields? - (GraphQL Java has the concept of [`TrivialDataFetcher`s](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/TrivialDataFetcher.java) which can be used as a marker interface for tracing exclusion, I don't know if other GraphQL implementations have an equivalent)

laurit:
> Looks good, but shouldn't we add tracing for [_parsing_](https://spec.graphql.org/October2021/#sel-EANLDCAACC5Ev_f) and [_validation_](https://spec.graphql.org/October2021/#sec-Validating-Requests) steps? They can contribute significantly to simple (mainly parsing) and complex (mainly validation) operations.
> 
Not specifying parsing, validation and data fetchers was a deliberate choice. In java instrumentation we don't have instrumentations that would provide that level of detail, hence our GraphQL instrumentation also currently also generates only a single span that encompasses the whole operation. If requested by users we may add extra spans, possibly enabled by a configuration flag. Instrumentations in other languages might choose to handle this differently and generate more spans. This PR only aims to find some common ground by defining attributes that could be useful for all GraphQL instrumentations and the name of the top level span.

laurit:
> * This Spec PR has no real details around Error handling, which I think is important. We have some handling for it here, it might make sense to formalize this, https://github.com/open-telemetry/opentelemetry-ruby/blob/46ef191f579d23ce51763a50c01ab6a43a233adf/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb#L32-L39
> 
As far as I can tell this only handles validation errors, but at least in the java implementation validation isn't the only source for errors. If people feel that we should define an event for errors I'd suggest to use a more generic name than `graphql.validation.error`. Currently both java and javascript implementation report such errors as exceptions although they really aren't exceptions.

tigrannajaryan (approved):
LGTM.

Should we have at least one attribute that is "Required" (e.g. `graphql.operation.name` or `graphql.operation.type`)? If you know an attribute is required you can easily identify all GraphQL spans by looking for just that attibute.

It is not mandatory but it is useful to have at least one Required attribute in each set of conventions. If we don't want any to mandatory then that's ok.

laurit:
> Should we have at least one attribute that is "Required" (e.g. `graphql.operation.name` or `graphql.operation.type`)? If you know an attribute is required you can easily identify all GraphQL spans by looking for just that attibute.
> 
> It is not mandatory but it is useful to have at least one Required attribute in each set of conventions. If we don't want any to mandatory then that's ok.

The problem here is that the input query might be invalid which makes it impossible to tell what the operation type is.

jord1e:
> The problem here is that the input query might be invalid which makes it impossible to tell what the operation type is.

Would this not be fixed by introducing spans for parsing and validation?

This was actually included in my original GraphQL Java instrumentation PR and I was thinking of adding it back into the library: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4337/files#diff-6475c73d1ffe60f69dc012bbf29113111ef68d07a31fc3f15d25f6e7a904ddf1R170-R212

opentelemetry-js already has these and Apollo Gateway also has them kind of: https://www.apollographql.com/docs/federation/opentelemetry#graphql-specific-spans

We could start out with `graphql.parse`, `graphql.validate` and `graphql.execute` (in this PR `graphql` I think)

I do not know the Ruby instrumentation situation but for GraphQL Java it is definitely possible

laurit:
> > The problem here is that the input query might be invalid which makes it impossible to tell what the operation type is.
> 
> Would this not be fixed by introducing spans for parsing and validation?
> 
I haven't read through everything in the spec, but in the parts that I have read I have not noticed a pattern where instrumentation would be required to generate multiple nested or sequential spans. Introducing `parse` and `validation` spans would deviate from this. I suspect that an attempt to introduce them would start a discussion on which sub-operations are important enough to deserve their own spans and whether such spans should be generated by default. I personally don't believe that we should have spans for `parse` and `validation` in default configuration and thus would refrain from generating them unless there are guidelines that either mandate or encourage generating such spans. I'm aware that we do have GraphQL instrumentations that generate these spans. The intent of this PR is not to disallow generating such spans, but rather to find some common ground. Hopefully the question of whether and how many spans should sub-operations produce will get answered in common guidelines for writing instrumentations.

> And I definitely agree that we should have error handling
>
Do you feel that storing GraphQL errors as exceptions isn't good enough?

jord1e:

No.

tigrannajaryan:

dyladan (approved):
This looks good to me. We've had multiple requests for this in JS so I'm glad to see this near the finish line.

dyladan:
> > Should we have at least one attribute that is "Required" (e.g. `graphql.operation.name` or `graphql.operation.type`)? If you know an attribute is required you can easily identify all GraphQL spans by looking for just that attibute.
> > It is not mandatory but it is useful to have at least one Required attribute in each set of conventions. If we don't want any to mandatory then that's ok.
> 
> The problem here is that the input query might be invalid which makes it impossible to tell what the operation type is.

If the query is invalid, then possibly we should have a fallback operation name/type which is explicitly `invalid` or `failure` or similar?

jmacd (approved):

spencerwilson (approved):
Thanks for the ping, Tigran.

> I suspect that an attempt to introduce them would start a discussion on which sub-operations are important enough to deserve their own spans and whether such spans should be generated by default. ... The intent of this PR is not to disallow generating such spans, but rather to find some common ground.

I'm totally supportive of doing things iteratively, and think this PR is effective at finding a basic starting point. While I'm interested in future work that specifies semconv for particular parts of operation execution (see my issue comment: https://github.com/open-telemetry/opentelemetry-specification/issues/1670#issuecomment-997332541) —and possibly defines all those more-detailed spans to be _optional_—I think specifying those conventions is unlikely to motivate changes to what this PR defines.

Thank you, @laurit!

andrewhr:
One concern around span names following the `<graphql.operation.type> <graphql.operation.name>` pattern, operation name can be anything the user defined, and so, they're strictly speaking unbounded. AFAICT, this goes against a hard constraint for span names, correct?

specification/trace/semantic_conventions/instrumentation/graphql.md
```diff
@@ -0,0 +1,29 @@
...
+`graphql.operation.type` and `graphql.operation.name` are available. If `graphql.operation.name` is not available, the
+span SHOULD be named `<graphql.operation.type>`. When `<graphql.operation.type>` is not available, `GraphQL Operation`
+MAY be used as span name.
+
+<!-- semconv graphql -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `graphql.operation.name` | string | The name of the operation being executed. | `findBookById` | No |
+| `graphql.operation.type` | string | The type of the operation being executed. | `query`; `mutation`; `subscription` | No |
+| `graphql.operation.body` | string | The GraphQL query being executed. [1] | `query findBookById { bookById(id: ?) { name } }` | No |
```
andrewhr, line 18: What about `graphql.operation.document`, or maybe `graphql.document` instead? This fall in line with some of [GraphQL's terminology](https://spec.graphql.org/June2018/#sec-Language.Document).

Alternatively, `graphql.selection` as [selection](https://spec.graphql.org/June2018/#sec-Selection-Sets) applies at resolver level - the same field could be reused.

For the record, [JS implementation](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/1da0216180de694c15ec356d476f465811757ae4/plugins/node/opentelemetry-instrumentation-graphql/src/enums/AttributeNames.ts#L18) uses `graphql.source` for both purposes.

tigrannajaryan:
@laurit please resolve the conversations and rebase from main to prepare for merging.

@open-telemetry/specs-approvers we have the minimum required number of approvals so unless I see objections I am going to merge this.

specification/trace/semantic_conventions/instrumentation/graphql.md
laurit: Thanks for the suggestion, I believe `graphql.document` is the best name that has been suggested for this attribute this far.
Your concern about the span naming is definitely valid. Luckily these conventions are experimental and open for further refinement. Having a span name that lets users identify the operation is much more user friendly than just using operation type or `graphql.execute` as the span name.

trask:
@laurit looks like there's a merge conflict

laurit:
@trask thanks, rebased.

trask:
@open-telemetry/technical-committee I think this is ready to be merged, thx!

Issue #1773, opened 2021-06-23
Define best practices for no-code configuration


carlosalberto:
A follow up to #1751 and #1130 (which we decided to go ahed with), as pointed out by @tigrannajaryan 

* Proliferation of environment variables, and whether this can end up becoming unmanageable.
* Is having _full_ configuration via environment variables a goal? Probably not.
* Maybe we need to support configuration files? (maybe as an alternative if _full_ configuration wants to be achieved)

Because of this, we should define best practices around no-code configuration and how that will impact future environment variables.

tigrannajaryan:
### Possible Solution

I think we can do the following:
- Introduce a concept of SDK config file.
- SDK config file to have a structure that is derived from the names of env variables. Or we can do the opposite: end variable names are derived from SDK config file (the end result is the same).
- The keys in the config file will have exact correspondence (mapping) with the env var names. Ideally the mapping will be trivial so that given the env var name someone can easily figure out what's the name of the key to use in the config file to change the same setting.
- We can use declare env vars list or the config file the source of truth and generate one from another in this repo.

The SDKs will accept 2 new env variables (oh the irony):
- OTEL_CONFIG_FILE. The name of config file to load and which contains config settings to use.
- OTEL_CONFIG. The config content directly in the config file format.

The rest of env variables will also fully remain. We will need to decide on the precedence when the same setting is provided in the config file and as an env variable.

### Why This is Useful

#### For End Users

For simple cases people will likely continue to use env variables. When you need to specify lots of settings it may be easier and more maintainable to have a single config file with all the settings that you want to pass to the SDK.

#### For Spec Contributors

Having one centralized view of all settings often helps to notice inconsistencies easier and helps to have better organization, naming and grouping of settings. I have just played with some structures and immediately discovered that we are still mixing plurar and singular "metric" and "metrics" in a couple places and "span" and "traces" in some others (which we tried to avoid and fixed, but it turns out we missed a few).

### Example Config File

We can choose from a variety of formats. For example let's assume it is a yaml file and each env variable becomes a key in the yaml, where underscore corresponds to a parent-child relationship.

So for example example OTEL_EXPORTER_OTLP_ENDPOINT can be set via yaml key:
```yaml
exporter:
  otlp:
    endpoint: example.com
```

For reference I generated the full list of config settings with the names of env vars printed next to the corresponding config file key. It looks like this:

```yaml
otel:
  attribute:
    count:
[...]

PR #2461, opened 2022-03-31
Add environment variables for capturing Http headers

ashu658:
## Changes
Add environment variables to be configured for capturing http headers. 
This [spec (HTTP request and response headers)](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-request-and-response-headers) describes  how to capture the http request/response header and add them in the span. It does not specify the environment variables need to be configured to capture the headers. This change specifies these environment variables.

Env vars proposed in the PR are already supported in [Java](https://github.com/open-telemetry/opentelemetry.io/blob/b04b753c9a5c9d14899053b78bbd46bb51dfa97e/content/en/docs/instrumentation/java/automatic/agent-config.md#capturing-http-request-and-response-headers) and [Python](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py#L21) sdks. 

Related PR: https://github.com/open-telemetry/opentelemetry-specification/pull/1898  

Additional info: 
One thing that I have seen in python is that headers provided by some frameworks/libraries (e.g. wsgi based frameworks like flask) are lowercase and `-` is replaced with `_` which would mean user who configures
`OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_REQUEST="CusTOM_HEAder-1"`
will be able to capture ``custom-header-1`` from request headers for some frameworks/libraries whereas for some other frameworks he won't be able to capture this header as it does not change `-` to `_`.
(All python frameworks/libraries I saw provides case-insensitive headers but only some of them replace `-` with `_`)
Mentioning this info here as I am not sure if it might affect this spec or not.

### DIFF
--- a/specification/sdk-environment-variables.md
+++ b/specification/sdk-environment-variables.md
@@ -263,3 +263,19 @@ To ensure consistent naming across projects, this specification recommends that 
 ```
 OTEL_{LANGUAGE}_{FEATURE}
 ```
+
+## HTTP request and response headers
+
+See [HTTP request and response headers](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-request-and-response-headers)
+
+Environment variables for capturing http request and response headers:
+
+| Name                                                      | Description                                                                                                                                         |
+|-----------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------|
+| OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_REQUEST  | A comma-separated list of HTTP header names. HTTP server instrumentations will capture HTTP request header values for all configured header names.  |
+| OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_RESPONSE | A comma-separated list of HTTP header names. HTTP server instrumentations will capture HTTP response header values for all configured header names. |
+| OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_CLIENT_REQUEST  | A comma-separated list of HTTP header names. HTTP client instrumentations will capture HTTP request header values for all configured header names.  |
+| OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_CLIENT_RESPONSE | A comma-separated list of HTTP header names. HTTP client instrumentations will capture HTTP response header values for all configured header names. |
+
+For example,
+``OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_REQUEST="Custom-Header-1,Content-Type"`` will capture ``Custom-Header-1`` and ``Content-Type`` headers from request headers and add them as span attributes. The names of the added span attributes will be normalised to ``http.request.header.custom_header_1`` and ``http.request.header.content_type`` as per the [semantic convention](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-request-and-response-headers).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ release.
   ([#2276](https://github.com/open-telemetry/opentelemetry-specification/pull/2276)).
 - Add documentation REQUIREMENT for adding attributes at span creation.
   ([#2383](https://github.com/open-telemetry/opentelemetry-specification/pull/2383)).
+- Add environment variables for capturing Http headers.
+  ([#2461](https://github.com/open-telemetry/opentelemetry-specification/pull/2461)).
 
 ### Metrics
 


reyang (changes requested):
https://github.com/open-telemetry/opentelemetry-specification/pull/2461/files#r841949238

specification/sdk-environment-variables.md
```diff
@@ -263,3 +263,19 @@ To ensure consistent naming across projects, this specification recommends that
...
+
+## HTTP request and response headers
+
+See [HTTP request and response headers](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-request-and-response-headers)
+
+Environment variables for capturing http request and response headers:
+
+| Name                                                      | Description                                                                                                                                         |
+|-----------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------|
+| OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_REQUEST  | A comma-separated list of HTTP header names. HTTP server instrumentations will capture HTTP request header values for all configured header names.  |
```
jmacd, line 275: Are these header names case-sensitive? Can you add a sentence or two saying what implementations are required to do with these variables in explicit terms?

dyladan: From [[RFC 7230] Section 3.2: Header Fields](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2)

>Each header field consists of a **case-insensitive** field name followed
   by a colon (":"), optional leading whitespace, the field value, and
   optional trailing whitespace.

emphasis mine

dyladan:
> One thing that I have seen in python is that headers provided by some frameworks/libraries (e.g. wsgi based frameworks like flask) are lowercase and `-` is replaced with `_` which would mean user who configures
> `OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_REQUEST="CusTOM_HEAder-1"`
> will be able to capture `custom-header-1` from request headers for some frameworks/libraries whereas for some other frameworks he won't be able to capture this header as it does not change `-` to `_`.
> (All python frameworks/libraries I saw provides case-insensitive headers but only some of them replace `-` with `_`)
> Mentioning this info here as I am not sure if it might affect this spec or not.

This seems very Python-specific to me and depends heavily on the library being used and the idioms of the ecosystem. I think this would be considered very confusing by most non-python users.

ashu658:
> This seems very Python-specific to me and depends heavily on the library being used and the idioms of the ecosystem. I think this would be considered very confusing by most non-python users.
@dyladan thanks for the feedback! I was not sure if this might affect the spec so, mentioned it here.

yurishkuro:
@open-telemetry/technical-committee I just did a quick search and did not find any env vars of the patten `OTEL_INSTRUMENTATION_*`. Which is understandable because so far the env vars were limited to configuring the SDKs. Capturing headers is not an SDK function, it's an instrumentation function. We have already heard concerns with having too many env vars that need to be supported by all SDKs. Do we want to extend this even further to configuring instrumentation libraries?

dyladan:
> @open-telemetry/technical-committee I just did a quick search and did not find any env vars of the patten `OTEL_INSTRUMENTATION_*`. Which is understandable because so far the env vars were limited to configuring the SDKs. Capturing headers is not an SDK function, it's an instrumentation function. We have already heard concerns with having too many env vars that need to be supported by all SDKs. Do we want to extend this even further to configuring instrumentation libraries?

Discussed this briefly on the spec call today and I think the consensus was that instrumentation-specific configuration doesn't belong in the SDK spec. The greater takeaway was the need for a consistent configuration specification and mechanism to ensure we can provide these configurations and others to all instrumentations in a consistent manner and document them appropriately. I can't remember who exactly but I think someone volunteered to get the ball rolling in that area.

specification/sdk-environment-variables.md
ashu658: The spec which describes how to capture the HTTP headers https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-request-and-response-headers requires the captured header names to be normalised (lowercase, with - characters replaced by _) when added as span attributes. So, two different header fields (e.g. `custom-header` and `custom_header`) will have the same span attribute key when added as span attribute. 
I am not sure if this would be okay. Please share your thoughts.

dyladan: Looks like the proposal for that behavior was done there by @Oberon00 https://github.com/open-telemetry/opentelemetry-specification/pull/1898#issuecomment-914238509

The motivation for lowercasing is obvious enough, but it seems the motivation to transform `-` to `_` was simply so that it would comply with the naming convention of our other semantic conventions? IMO the transformation was a mistake but one that it might be too late to rectify. In any case, configuring which headers are captured is an entirely different problem than how to represent them in OTLP. I still stand by my case that the `-` to `_` conversion is confusing within this header spec. Lowercasing seems fine and an obvious enough way to comply with case insensitivity.

svrnm:
> Discussed this briefly on the spec call today and I think the consensus was that instrumentation-specific configuration doesn't belong in the SDK spec. The greater takeaway was the need for a consistent configuration specification and mechanism to ensure we can provide these configurations and others to all instrumentations in a consistent manner and document them appropriately. I can't remember who exactly but I think someone volunteered to get the ball rolling in that area.

Just for our understanding: Does this mean we should close this PR and wait for the configuration specification & mechanism to come? Assuming that this will take some time, what to do with this in the meantime, Java and Python are using it already and if other languages jump on that same topic, it would be good to have used the same variables until the config exists.

tigrannajaryan:
> Just for our understanding: Does this mean we should close this PR and wait for the configuration specification & mechanism to come? Assuming that this will take some time, what to do with this in the meantime, Java and Python are using it already and if other languages jump on that same topic, it would be good to have used the same variables until the config exists.

I think we need to at least figure out some answers to these general questions so that all instrumentations follow a consistent configuration approach:
- Should instrumentations be configurable via env variables?
- Is the same configuration settable from the code?
- Is naming of instrumentations' env variables dictated by the spec?
- Where in the spec it should live?
- What is the variable naming scheme? Do we choose a particular prefix like this PR suggests?
- How to name variables which are specific for just one instrumentation and are not generic and applicable to many instrumentations like "OTEL_INSTRUMENTATION_HTTP*" is? Should the variable names be prefixed by the instrumentation library name?

This should be ideally opened as a separate issue, discussed and after we come to an agreement on what we want to do generally, we can unblock this PR.

svrnm:
thanks @tigrannajaryan for the clarification, is #1773 the right issue to continue the conversation or is a new one doing a better job?

PR #2443, opened 2022-03-25
Remove duplicate requirement from compliance matrix

cijothomas:

PR #2462, opened 2022-03-31
Clarify view conflicts

jack-berg:
Related to #2443. 

- Views that create instrument identity conflicts SHOULD be applied.
- Views that create semantic errors should be dropped.


specification/metrics/sdk.md
```diff
@@ -210,11 +210,13 @@ made with an Instrument:
...
     criteria:
-    * Try to apply the View configuration. If there is an error or a conflict
-      (e.g. the View requires to export the metrics using a certain name, but
-      the name is already used by another View), provide a way to let the user
-      know (e.g. expose
-      [self-diagnostics logs](../error-handling.md#self-diagnostics)).
+    * Try to apply the View configuration. If applying the View results
+      in [duplicate instruments](./api.md#instrument-type-conflict-detection)
+      with the same name, the implementation SHOULD apply the View and emit a
+      warning. If it is not possible to apply the View without producing
```
alanwest, line 216: Views do not create duplicate instruments. Maybe wording it like:

"Try to apply the View configuration. As can be the case with [duplicate instruments](./api.md#instrument-type-conflict-detection), if applying the View results will result in a duplicate metric stream, the implementation SHOULD..."

alanwest:
The other thing we discussed in this week's spec meeting was the option for SDK's to fail fast when it is known a view will cause conflicts.

There's a section right above that says:

> In order to avoid conflicts, views which specify a name SHOULD have an
instrument selector that selects at most one instrument...

Could some additional more generic wording be added to this like

> For all other cases where it is known that registering a view will cause a conflicts, then SDKs MAY fail fast.

specification/metrics/sdk.md
alanwest: Should this actually be a MUST?

jack-berg: > Views do not create duplicate instruments. Maybe wording it like:

You're right - good catch. There's a section in the data model that talks about [metric identity](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#opentelemetry-protocol-data-model-producer-recommendations) specifically. Better to link to this, I think. 

> Should this actually be a MUST?

I'm not sure. The instrument identity conflict section specified that implementations "MUST create a valid Instrument in ever case". But this is a bit different - if an instrument identity conflict occurs and an SDK instead returns a noop instrument, data is lost. However, if a view produces a conflict and an SDK decides to not include the stream, it means one particular view of the data is lost. The consequences are diminished a bit.

specification/metrics/sdk.md
```diff
@@ -210,11 +213,13 @@ made with an Instrument:
 * If the `MeterProvider` has one or more `View`(s) registered:
   * For each View, if the Instrument could match the instrument selection
     criteria:
-    * Try to apply the View configuration. If there is an error or a conflict
-      (e.g. the View requires to export the metrics using a certain name, but
-      the name is already used by another View), provide a way to let the user
-      know (e.g. expose
-      [self-diagnostics logs](../error-handling.md#self-diagnostics)).
+    * Try to apply the View configuration. If applying the View results
```
cijothomas, line 216: I understand this was discussed in the last SIG meeting... I'd like to make sure the following is considered:
1. View is an SDK concept, and app owner has full control over it.
2. If app owner adds a View, which cause any conflict (instrument identity conflict or asking to do something which cause semantic errors), SDK should ignore the view. i.e proceed as if *this* view was not even added. 
3. App owners has full control on the Views, and they can fix the View to avoid conflicts/semantic errors etc.

reyang: > If app owner adds a View, which cause any conflict (instrument identity conflict or asking to do something which cause semantic errors), SDK should ignore the view. i.e proceed as if this view was not even added.

The SDK should let the app developer know, following the error handling principle.

specification/metrics/sdk.md
```diff
@@ -210,11 +210,13 @@ made with an Instrument:
...
+      in [duplicate instruments](./api.md#instrument-type-conflict-detection)
+      with the same name, the implementation SHOULD apply the View and emit a
+      warning. If it is not possible to apply the View without producing
+      semantic errors (e.g. the View sets an asynchronous instrument to use
+      the [Histogram aggregation](#histogram-aggregation)) the implementation
+      SHOULD drop the View and emit a warning.
```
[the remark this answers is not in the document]
reyang: Good point 👍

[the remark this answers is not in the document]
jmacd: @jack-berg Thank you for raising the point about semantic conflicts. I wonder if we need to be concrete about which instruments are compatible with which data points? The reason I think so is that we have not added a GaugeHistogram to our data model, and we do not currently have a way to model several reasonable outcomes where a histogram is used to aggregate values from the other instruments.

Should we allow Gauge points to be aggregated from Histogram instruments? I way to say that this could be allowed in the future, if we ever reach the point of adding a synchronous gauge instrument, the aggregation it produces will be compatible with histogram instruments.

I think the rules could be listed in terms of Instrument-to-Aggregation.

| Instrument | Compatible Aggregations |
| -- | -- |
| Counter | Monotonic Sum, Explicit/Exponential Histogram |
| UpDownCounter | Non-monotonic Sum |
| Histogram | Explicit/Exponential Histogram, Monotonic Sum |
| Async Counter | Monotonic Sum |
| Async UpDownCounter | Non-Monotonic Sum |
| Async Gauge | Gauge |

Related: https://github.com/open-telemetry/opentelemetry-proto/issues/303, https://github.com/open-telemetry/opentelemetry-proto/issues/308

specification/metrics/sdk.md
```diff
@@ -210,11 +213,13 @@ made with an Instrument:
 * If the `MeterProvider` has one or more `View`(s) registered:
   * For each View, if the Instrument could match the instrument selection
     criteria:
-    * Try to apply the View configuration. If there is an error or a conflict
-      (e.g. the View requires to export the metrics using a certain name, but
-      the name is already used by another View), provide a way to let the user
-      know (e.g. expose
-      [self-diagnostics logs](../error-handling.md#self-diagnostics)).
+    * Try to apply the View configuration. If applying the View results
+      in [conflicting metric identities](./datamodel.md#exponentialhistogram-producer-recommendations)
```
jmacd, line 217:
```suggestion
      in [conflicting metric identities](./datamodel.md#opentelemetry-protocol-data-model-producer-recommendations)
```

jmacd (approved):

specification/metrics/sdk.md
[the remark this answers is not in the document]
jack-berg: > I wonder if we need to be concrete about which instruments are compatible with which data points?

I think this is a great idea. 👍 

I brought up the problem I mentioned about asynchronous instruments not being aggregatable as histograms with the java SIG last night and @jkwatson brought up the point that perhaps someone has a use case out there where they want to have "second order" aggregations, i.e. a set of observed sums aggregated into a histogram so that their distribution could be analyzed. Because asynchronous gauges and asynchronous up down counters support negative measurements, negative values would have to be dropped when aggregating as a histogram, which further diminishes the value.

Personally, I think it'd be perfectly reasonable to use spec the Instrument-to-Aggregation table you've proposed, which disallows aggregating all the async instruments as histograms. Someone might have a real use case for second order aggregations, but we can collect that feedback and expand the available aggregations later if needed.

specification/metrics/sdk.md
jack-berg: I have a slight preference having view conflict behavior be consistent with instrument identity conflict behavior. However, I think dropping is reasonable as well, and since this is an edge case I'm more interested in ensuring consistency across implementations than what the behavior is.

specification/metrics/sdk.md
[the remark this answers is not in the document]
jmacd: This idea of "second order aggregations" is what I meant by "several reasonable outcomes where a histogram is used to aggregate values from the other instruments". 

> Personally, I think it'd be perfectly reasonable to use spec the Instrument-to-Aggregation table you've proposed, which disallows aggregating all the async instruments as histograms.

Yes, I think we should also disallow aggregating synchronous instruments as Gauges until we add support in the data model. 

I think we can accommodate forms of Histogram and Gauge that are not currently representable in the data model in a future, additive change. I started drafting this position here: https://github.com/open-telemetry/oteps/pull/156, which would be to: (1) add more data type information to the histogram, for whether the primitive measurement is a counter, updowncounter, or gauge, (2) add more data type information to the histogram, for whether the counts it expresses are monotonic or not. I don't think we should focus on this now.

specification/metrics/sdk.md
jmacd: @jack-berg is the consistent position you're describing: (1) You may fail fast when a View produces a conflict (even if the conflict results from an instrument registered at runtime after the View was registered), (2) When not failing fast, all views will be performed despite conflicts?

specification/metrics/sdk.md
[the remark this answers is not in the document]
jmacd: FWIW, the sdk spec already contains this line about Histogram aggregations, it sounds like we can move this into a new Instrument-to-Aggregation compatibility section.
```
This SHOULD NOT be collected when used with
instruments that record negative measurements (e.g. `UpDownCounter` or `ObservableGauge`).
```

specification/metrics/sdk.md
```diff
@@ -196,8 +196,11 @@ In order to avoid conflicts, views which specify a name SHOULD have an
 instrument selector that selects at most one instrument. For the registration
 mechanism described above, where selection is provided via configuration, the
 SDK SHOULD NOT allow Views with a specified name to be declared with instrument
-selectors that may select more than one instrument (e.g. wild card instrument name)
-in the same Meter.
+selectors that may select more than one instrument (e.g. wild card instrument
+name) in the same Meter. For this and other cases where registering a view will
+cause a conflict, SDKs MAY fail fast in accordance with
+initialization [error handling principles](../error-handling.md#basic-error-handling-principles)
+.
```
aabmass, line 203: nit this newline causes a space in the rendered output "initialization error handling principles ."

aabmass (approved):

tsloughter:
"applied" means this is saying that measurements should be aggregated for all Views even if they conflict?

Assuming that is the meaning, and not that "apply" is referring to "applying configuration" and replacing the previous View in the state of Views, I don't see the purpose/benefit mentioned?

But now that I realize it also says a warning is emitted when the View is "applied" it must be that applying refers to updating the state and not applying the View against a measurement (since that would be a lot of warnings)?

In that case I'd just suggest using a different word and clarifying if it is a full replacement or a merge.

tsloughter:
Ok, now I realize there might even be another case this is meaning to cover, which can't be handled at configuration time -- a no name View with use of wildcards (`?` and `*`) in the instrument name match, making it unknown at the time of view creation if any of those Views it will ever match 0, 1 or N instruments.

I'd argue the spec should define that the Views are kept in the order they are registered and matching is either first or last wins in the case there is a conflict in the resulting name.

On slack it sounds like all implementations require up front View creation, but the spec reads as Views are matched to Instruments each time a measurement is taken. I wouldn't have considered this an issue -- since the end result is the same -- but now I think the spec could be written more easily and clearly by defining Instrument and View matching this way as well.

jack-berg:
@tsloughter this PR doesn't introduce the language to "apply the View" or to log a warning if there's a conflict. That language has already been there and I think it's implied that logging doesn't refer to handling each individual measurement. 

> a no name View with use of wildcards (? and *) in the instrument name match, making it unknown at the time of view creation if any of those Views it will ever match 0, 1 or N instruments.

Yes. Additionally, a view which changes the name might match exactly one instrument, but the view's name might conflict with another instrument which has no matching views. 

This PR is all about clarifying to ensure that SDKs behave similarly in these types of situations. It resulted from discovering that the dot net and java implementations had different interpretations.

>  I'd argue the spec should define that the Views are kept in the order they are registered and matching is either first or last wins in the case there is a conflict in the resulting name.

That's roughly the position that's expressed in this [comment](https://github.com/open-telemetry/opentelemetry-specification/pull/2462#discussion_r840121877).

tsloughter:
@jack-berg ok, yea, I see "apply" was already used. But the main change appears to be, "the implementation SHOULD apply the View and emit a warning". Ignoring whether "apply" should be used here or not, I'm still not sure what I should do when implementing this to get consistent results?

jack-berg:
We spoke about this issue at the 4/5 Spec SIG. 

@jmacd brought up the point that in [instruments](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-type-conflict-detection) "MUST create a valid Instrument in every case." Suppose we were to take the [position](https://github.com/open-telemetry/opentelemetry-specification/pull/2462#discussion_r840121877) that we should ignore views that produce conflicts, and a view is registered that renames an instrument from `foo` to `bar`, and later an instrument is registered with name `bar`. A view has created a conflict, but it wasn't possible to know about that conflict until some later point when the instrument with name `bar` was registered. 

The lack of ability to know that a view will produce a conflict, and the requirement that instruments must always create valid instruments that export data suggests that the most correct / consistent behavior is for the SDK to produce all views regardless of whether they produce conflicts. 

There seemed to be some consensus on this thought at the SIG meeting, and this PR currently reflects that position.

alanwest (approved):

specification/metrics/sdk.md
```diff
@@ -210,11 +213,13 @@ made with an Instrument:
...
-      the name is already used by another View), provide a way to let the user
-      know (e.g. expose
-      [self-diagnostics logs](../error-handling.md#self-diagnostics)).
+    * Try to apply the View configuration. If applying the View results
+      in [conflicting metric identities](./datamodel.md#opentelemetry-protocol-data-model-producer-recommendations)
+      the implementation SHOULD apply the View and emit a warning. If it is not
+      possible to apply the View without producing semantic errors (e.g. the
+      View sets an asynchronous instrument to use
+      the [Histogram aggregation](#histogram-aggregation)) the implementation
+      SHOULD drop the View and emit a warning.
```
cijothomas, line 222:
```suggestion
      SHOULD drop the View and emit a warning.
```

^ Could you clarify what does it mean by "Dropping the view" ?

Did you mean "proceed as if this View did not exist" ? This would mean if there is another view (or default view) , the instrument still gets reported.
OR
"Ignore this instrument."

specification/metrics/sdk.md
cijothomas: I am okay if the preference is to have consistent behavior with instrument identify conflict behavior.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#resolving-duplicate-instrument-registration-conflicts this section has the statement "The implementation MUST aggregate data from identical Instruments together in its export pipeline."

Did you intend for the Views to also do this "aggregate together" part?

For the below example:

.AddView("name", new MetricStreamConfiguration { Name = "newname" })
.AddView("name", new MetricStreamConfiguration { Name = "newname" })
var instrument1 = meter.CreateCounter<long>("name");

will instrument1 result in a two metric streams (of same name and hence, conflicting)?
OR
will instrument1 result in a single metric stream (i.e SDK "merges" 2 views into one.)

cijothomas: Another example:

.AddView("name", new MetricStreamConfiguration { Name = "newname", AttributeKeys: "key1" })
.AddView("name", new MetricStreamConfiguration { Name = "newname", AttributeKeys: "key1" })
var instrument1 = meter.CreateCounter("name");
will instrument1 result in a two metric streams (of same name, using same attributes and hence, conflicting)?
OR
will instrument1 result in a single metric stream (i.e SDK "merges" 2 views into one.)

tsloughter:
Jack reached out on slack and explained everything. I get now that this simply means conflicting metrics will be reported and that the user must intervene to resolve this error.

+1 to this solution. As was noted, the fact that a conflict could result from instrument being created at runtime that conflicts with the name of a view but Instrument API requires a valid Instrument always be returned makes it really the only viable solution.

My only concern is that at least one form of overriding views is useful: Setting up a view that drops everything and a view that doesn't and matches certain instruments only.

This would work fine with this PR as is, I just wonder if it could be confusing to give the appearance that view overriding works, when it is not an override at all. May just need some docs by implementations for SDK users.

specification/metrics/sdk.md
jack-berg: Good question. I'm inclined to think "proceed as if this View did not exist".

reyang (approved):

specification/metrics/sdk.md
jack-berg: @cijothomas sorry for the delayed reply. I don't have a strong opinion here. Right now the java implementation will aggregate views together if they are identical. If they differ in name, description, aggregation, or attributes, we produce separate streams.

cijothomas (approved):

specification/metrics/sdk.md
jack-berg: > Why is there merging of streams instead of just only having 1 view?

I think we're talking about the same thing.

reyang:
@jack-berg CI is failing, would you look into it?

specification/metrics/sdk.md
reyang: @cijothomas do you consider this conversation as resolved?
@tsloughter do you see a blocker here?

tsloughter: @reyang no, I was just trying to understand how others are implementing this and hoping for consistent phrasing so information isn't lost during discussions.

"merging streams" sounds like measurements are being handled separately for each identical view and then merged in some way at export -- as opposed to at configuration time the set of views does not contain multiple identical views, it is left with only 1.

PR #2478, opened 2022-04-08
Update JVM metrics with JMX Gatherer values.

carlosalberto:
Add a pair of metrics from the [JMX Gatherer](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/jmx-metrics/docs/target-systems/jvm.md), renamed accordingly to comply with the existing runtime metrics:

* `process.runtime.jvm.threads.count` (renamed from `jvm.threads.count`), UpDownCounter
* `process.runtime.jvm.classes.loaded` (renamed from `jvm.classes.loaded`), UpDownCounter
* `process.runtime.jvm.classes.unloaded` (new), UpDownCounter

(Left out the GC metrics as those probably belong to its own discussion)


specification/metrics/semantic_conventions/runtime-environment-metrics.md
```diff
@@ -67,6 +67,8 @@ All JVM metric attributes are required unless otherwise indicated.
 |                                      |                                     |       |                                           |                            |            | pool          | Name of pool [1]      |
 | process.runtime.jvm.memory.max       | Measure of max obtainable memory    | Bytes | `By`                                      | Asynchronous UpDownCounter | Int64      | type          | `"heap"`, `"nonheap"` |
 |                                      |                                     |       |                                           |                            |            | pool          | Name of pool [1]      |
+| process.runtime.jvm.threads.count    | Number of executing threads         | threads | `{threads}`                             | Asynchronous Gauge         | Int64      |               |                       |
+| process.runtime.jvm.classes.loaded   | Number of loaded classes            | classes | `{classes}`                             | Asynchronous Gauge         | Int64      |               |                       |
```
bogdandrutu, line 71: Both should be updowncounter, or counter (is the num classes loaded ever going down)?

carlosalberto: @bogdandrutu it's never going down ;)

trask: there are three related JVM metrics:

* `LoadedClassCount` - the number of classes that are currently loaded in the Java virtual machine
* `TotalLoadedClassCount` - the total number of classes that have been loaded since the Java virtual machine has started execution
* `UnloadedClassCount` - the total number of classes unloaded since the Java virtual machine has started execution

where `LoadedClassCount` = `TotalLoadedClassCount` - `UnloadedClassCount`

@carlosalberto it looks like the [JMX metric gatherer](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/0e6a15aff9e2134325745e975dd013847868aee0/jmx-metrics/src/main/resources/target-systems/jvm.groovy#L18-L19) is collecting `LoadedClassCount` which does go down

carlosalberto: Oh, thanks a lot for the clarification @trask

carlosalberto:
@bogdandrutu @trask Please review (check is failing but that's not related to these changes).

carlosalberto:
@bogdandrutu @trask Please review (check is failing but that's not related to these changes).

specification/metrics/semantic_conventions/runtime-environment-metrics.md
jack-berg: If the sum of loaded and unloaded is semantically meaningful (i.e. loaded + unloaded = total loaded since jvm started), perhaps the metric should be named `process.runime.jvm.classes.count`, and an attribute should reflect whether the count represents loaded or unloaded classes.

jmacd (approved):

specification/metrics/semantic_conventions/runtime-environment-metrics.md
jack-berg: @trask and I discussed this a bit more today. Specifying loaded and unloaded as attributes is actually probably not such a great idea because they're not semantically equivalent. The loaded attribute would refer to the _current_ number of loaded classes, whereas the unloaded attribute would refer to the _total_ number of unloaded classes since JVM start. 

Probably better to leave them separate. 

@carlosalberto would you mind adding a separate `process.runtime.jvm.classes.unloaded` metric? There's precedent for collecting this information in several java metric systems.

carlosalberto: Oh, definitely. Will do.

carlosalberto: Done, please review ;)

specification/metrics/semantic_conventions/runtime-environment-metrics.md
```diff
@@ -67,6 +67,9 @@ All JVM metric attributes are required unless otherwise indicated.
 |                                      |                                     |       |                                           |                                                   |            | pool          | Name of pool [1]      |
 | process.runtime.jvm.memory.max       | Measure of max obtainable memory    | Bytes | `By`                                      | UpDownCounter                                     | Int64      | type          | `"heap"`, `"nonheap"` |
 |                                      |                                     |       |                                           |                                                   |            | pool          | Name of pool [1]      |
+| process.runtime.jvm.threads.count    | Number of executing threads         | threads | `{threads}`                             | UpDownCounter                                     | Int64      |               |                       |
+| process.runtime.jvm.classes.loaded   | Number of loaded classes            | classes | `{classes}`                             | UpDownCounter                                     | Int64      |               |                       |
```
jack-berg, line 71:
```suggestion
| process.runtime.jvm.classes.loaded   | Number of classes that are currently loaded            | classes | `{classes}`                             | UpDownCounter                                     | Int64      |               |                       |
```

specification/metrics/semantic_conventions/runtime-environment-metrics.md
```diff
@@ -67,6 +67,9 @@ All JVM metric attributes are required unless otherwise indicated.
...
+| process.runtime.jvm.classes.unloaded | Number of unloaded classes          | classes | `{classes}`                             | UpDownCounter                                     | Int64      |               |                       |
```
jack-berg, line 72:
```suggestion
| process.runtime.jvm.classes.unloaded | Number of classes unloaded since the JVM start           | classes | `{classes}`                             | UpDownCounter                                     | Int64      |               |                       |
```

jack-berg (approved):
Just a small clarification on the description

specification/metrics/semantic_conventions/runtime-environment-metrics.md
jmacd: @jack-berg Would you agree that the "number of loaded classes", which surely sounds monotonic, should be represented as an UpDownCounter because the rate of class loading is not an important factor and the absolute number of classes loaded is an important factor? I.e., we do not expect this value to drop but resets (i.e., the zero value) are significant, so we use a non-monotonic sum. I agree with this reasoning, but I'm not sure it's written down.

trask (approved):
with @jack-berg's proposed clarifications

jmacd (approved):

specification/metrics/semantic_conventions/runtime-environment-metrics.md
jack-berg: Although you're right about `process.runtime.jvm.unloaded` being monotonic. As you point out it could just be a counter. I still prefer up down counter because as you point out, its more useful to analyze the absolute number of classes loaded. Also, offers symmetry with `process.runtime.jvm.loaded`. 

It is admittedly a bit whacky.

PR #2479, opened 2022-04-08
[editorial] Ensure common section has a README

chalin:
- Contributes to fixing https://github.com/open-telemetry/opentelemetry.io/issues/1280 by doing the following:
  - Promotes `common/common.md` to be the section's index file (`common/README.md`)
- Renames `#attributes` anchors to their canonical name `#attribute` (singular), under the Common page

/cc @austinlparker @tigrannajaryan @carlosalberto


tigrannajaryan:
The common.md content is probably not a great fit for a README. I think we can allow this PR since it will result in a better than currently situation on the website, but we probably will need to re-work the README after that to make the content more suitable as an introduction to what a "common" section in the specification supposedly should be.

Question: do we need a README file in every sub-directory under the "specification" directory? Most sub-directories currently don't have one.

chalin:
> The common.md content is probably not a great fit for a README. I think we can allow this PR since it will result in a better than currently situation on the website

Great!



> Question: do we need a README file in every sub-directory under the "specification" directory? Most sub-directories currently don't have one.

Yes, in order to fix

- https://github.com/open-telemetry/opentelemetry.io/issues/1280

every section will need a README. I'm available to help here: e.g., I submit a first PR that adds a README (with only a title) to each section that doesn't have such a file. That'll act as a placeholder that can be built upon via subsequent PRs. WDYT?

tigrannajaryan:
> every section will need a README. I'm available to help here: e.g., I submit a first PR that adds a README (with only a title) to each section that doesn't have such a file. That'll act as a placeholder that can be built upon via subsequent PRs. WDYT?

Yes, sounds good. Some sub-directories have good candidates for README (e.g. most that have an "overview.md" should work as README).

tigrannajaryan (approved):

reyang (approved):

PR #2480, opened 2022-04-08
Retry RESOURCE_EXHAUSTED only if the server can recover

svrakitin:
Recommend clients to retry on `RESOURCE_EXHAUSTED` only if the server can recover. The server can indicate this by signaling backpressure (by including `RetryInfo` into response).

See: https://github.com/open-telemetry/opentelemetry-collector/pull/5147#issuecomment-1091952940

I might need help with exact wording here to stay consistent with the rest of the spec.


svrakitin:
cc @tigrannajaryan

specification/protocol/otlp.md
```diff
@@ -236,6 +236,8 @@ When retrying, the client SHOULD implement an exponential backoff strategy. An
 exception to this is the Throttling case explained below, which provides
 explicit instructions about retrying interval.
 
+The client SHOULD interpret `RESOURCE_EXHAUSTED` code as retryable only if the server signals backpressure to indicate a possible recovery.
```
tigrannajaryan, line 239: We need a more precise definition here. How is the backpressure signalled? The specification must unambiguously tell how to implement this behavior in the clients.

If https://github.com/open-telemetry/opentelemetry-collector/pull/5147 is the exact behavior that we aim for then describe that directly.

tigrannajaryan (commented):
This also needs a CHANGELOG enty.
I believe this is an allowed change since it can be considered a specification bug that it did not take into account the RESOURCE_EXHAUSTED can have more than one meaning and needs a more detailed description in the spec.

specification/protocol/otlp.md
svrakitin: Sure, I omitted that, because it is the next section after this sentence.

yurishkuro:
```suggestion
The client SHOULD interpret `RESOURCE_EXHAUSTED` code as retryable only if the server signals backpressure (described below) to indicate a possible recovery.
```

tigrannajaryan: I think it would be useful to say something like this:
> The client SHOULD interpret `RESOURCE_EXHAUSTED` code as retryable only if the server signals that the recovery from resource exhaustion is possible. This is signalled by the server by returning [a status](https://godoc.org/google.golang.org/grpc/status#Status.WithDetails)
containing
[RetryInfo](https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L40). In this case the behavior of the server and the client is exactly as described in [OTLP/gRPC Throttling]() section.
If no such status is returned then the `RESOURCE_EXHAUSTED` code SHOULD be treated as non-retryable.

svrakitin: Done, I added the implementation detail and made a reference to next section.

svrakitin: Haven't seen your message before I pushed.

tigrannajaryan (approved):

SergeyKanzhelev (approved):

tigrannajaryan:
@svrakitin please resolve the conflicts.

reyang (approved):

PR #2481, opened 2022-04-08
Update java log compliance matrix

jack-berg:
Add entries for java log compliance matrix. 

Also fixes the compliance matrix, moving `Shutdown`, and `ForceFlush` methods to `LogEmitterProvider` instead of `LogEmitter`. 

Request in issue [#4349](https://github.com/open-telemetry/opentelemetry-java/issues/4349)

### DIFF
--- a/spec-compliance-matrix.md
+++ b/spec-compliance-matrix.md
@@ -208,22 +208,22 @@ formats is required. Implementing more than one format is optional.
 
 Disclaimer: this list of features is still a work in progress, please refer to the specification if in any doubt.
 
-| Feature                                                     | Optional | Go | Java | JS | Python | Ruby | Erlang | PHP | Rust | C++ | .NET | Swift |
-|-------------------------------------------------------------|----------|----|------|----|--------|------|--------|-----|------|-----|------|-------|
-| **[Logging SDK](specification/logs/logging-library-sdk.md)**| Optional | Go | Java | JS | Python | Ruby | Erlang | PHP | Rust | C++ | .NET | Swift |
-| Get LogEmitter                                              |          |    |      |    |        |      |        |     |      |     |      |       |
-| LogEmitter.Emit(LogRecord)                                  |          |    |      |    |        |      |        |     |      |     |      |       |
-| LogEmitter.Shutdown                                         |          |    |      |    |        |      |        |     |      |     |      |       |
-| LogEmitter.ForceFlush                                       |          |    |      |    |        |      |        |     |      |     |      |       |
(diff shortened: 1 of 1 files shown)


trask (approved):

reyang (approved):

spec-compliance-matrix.md
```diff
@@ -208,22 +208,22 @@ formats is required. Implementing more than one format is optional.
...
+| LogEmitter.Emit(LogRecord)                                   |          |     | +    |     |        |      |        |     |      |     |      |       |
+| SimpleLogProcessor                                           |          |     | +    |     |        |      |        |     |      |     |      |       |
+| BatchLogProcessor                                            |          |     | +    |     |        |      |        |     |      |     |      |       |
+| Can plug custom log processor                                |          |     | +    |     |        |      |        |     |      |     |      |       |
+| OTLP/gRPC exporter                                           |          |     | +    |     |        |      |        |     |      |     |      |       |
+| OTLP/HTTP exporter                                           |          |     | +    |     |        |      |        |     |      |     |      |       |
+| OTLP File exporter                                           |          |     | -    |     |        |      |        |     |      |     |      |       |
```
jkwatson, line 223: I think we do have this: https://github.com/open-telemetry/opentelemetry-java/blob/main/exporters/logging-otlp/src/main/java/io/opentelemetry/exporter/logging/otlp/OtlpJsonLoggingLogExporter.java

jack-berg: So because we log to JUL in OTLP format, and because JUL has file logging options, we qualify as having OTLP file export capabilities?

arminru (approved):

spec-compliance-matrix.md
tigrannajaryan: >When configuring the LogEmitterProvider, OTLP File exporter should be used to write logs to a file or stdout in either OTLP JSON or OTLP Protobuf binary format.

Does JUL allow the target to be Protobuf binary format?

This subject is actually marked as a TODO in the spec, we need to clarify:
>TODO: clarify how this functionality co-exists with the overlapping functionality in logging libraries that allow specifying how logs are written to a file.

IMO, this cannot be not a substitute for a proper OTLP File exporter unless it can write Protobuf binary format (or we remove the Protobuf binary format as a requirement).

Until this is clarified in the spec, I think we can merge this PR as is.

tigrannajaryan (approved):

spec-compliance-matrix.md
reyang: Not blocking this PR.

PR #2482, opened 2022-04-09
Add spec for probabilistic log sampling

atoulme:
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/9117

## Changes
Add a new specification (experimental) for probabilistic log sampling.

Related issues:
https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/9118

### DIFF
--- a/specification/logs/log-probability-sampling.md
+++ b/specification/logs/log-probability-sampling.md
@@ -1,4 +1,4 @@
-# Logs Sampling
+# Log Probability Sampling
 
 **Status**: [Experimental](../document-status.md)
 
--- a/specification/logs/log-probability-sampling.md
+++ b/specification/logs/log-probability-sampling.md
@@ -19,7 +19,7 @@
 
 ## Overview
 
-This specification defines how logs are sampled with OpenTelemetry.
+This specification defines how probability sampling works for logs in OpenTelemetry.
 
 Log sampling inherits from the logic defined in tracing sampling. By default, a log will use its trace ID to determine
 whether it should be sampled, in a similar manner as [Trace state probability sampling](../trace/tracestate-probability-sampling.md).
@@ -33,7 +33,7 @@ Log sampling can consider another attribute as the source of its sampling. In th
 When the attribute is not set, the trace ID will be used instead if present.
 
 If no explicit attribute is set or no attribute has been defined, and no trace ID is present or trace ID sampling is disabled,
-the log is not considered as part of the sampling process and must be collected.
+the log is not considered as part of the sampling process and MUST be collected.
 
 Additionally, the log record may define an attribute to be considered as a sampling priority in a similar manner to how
 traces rely on a `sampling.priority` field to set the priority of the trace sampling.
--- a/specification/logs/log-probability-sampling.md
+++ b/specification/logs/log-probability-sampling.md
@@ -1,0 +1,56 @@
+# Logs Sampling
+
+**Status**: [Experimental](../document-status.md)
+
+<details>
+<summary>Table of Contents</summary>
+
+<!-- toc -->
+
+- [Overview](#overview)
+  * [References](#references)
+  * [Definitions](#definitions)
+- [Requirements](#requirements)
+  * [Deterministic sampling source](#deterministic-sampling-source)
+
+<!-- tocstop -->
+
+</details>
+
+## Overview
+
+This specification defines how logs are sampled with OpenTelemetry.
+
+Log sampling inherits from the logic defined in tracing sampling. By default, a log will use its trace ID to determine
+whether it should be sampled, in a similar manner as [Trace state probability sampling](../trace/tracestate-probability-sampling.md).
+
+In particular, log sampling borrows and reuses the notion of a hash seed that can be explicitly set on the sampler
+to guarantee sampling is consistent with a series of collectors.
+
+Trace ID sampling may be explicitly disabled.
+
+Log sampling can consider another attribute as the source of its sampling. In that case, the attribute must be explicitly defined.
+When the attribute is not set, the trace ID will be used instead if present.
+
+If no explicit attribute is set or no attribute has been defined, and no trace ID is present or trace ID sampling is disabled,
+the log is not considered as part of the sampling process and must be collected.
+
+Additionally, the log record may define an attribute to be considered as a sampling priority in a similar manner to how
+traces rely on a `sampling.priority` field to set the priority of the trace sampling.
+
+### References
+
+* [Trace state probability sampling](../trace/tracestate-probability-sampling.md)
+
+### Definitions
+
+* Sampling source: an attribute representing the source of sampling.
+* Sampling priority: an attribute representing the priority of the log to be sampled. It is represented as percentage points, with 100 to always sample the log record, 0 to never sample the log record.  
+
+## Requirements
+
+### Deterministic sampling source
+
+The log record must use a deterministic sampling source to represent the log record while sampling.
+This is necessary to address the case where logs are sampled by multiple collectors, either as part of a highly-available setup where multiple collectors ingest a log source,
+or when collectors are placed in sequence.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ release.
 
 ### Logs
 
+- Add a log sampling specification.
+  ([#2482](https://github.com/open-telemetry/opentelemetry-specification/pull/2482))
+
 ### Resource
 
 ### Semantic Conventions


specification/logs/log-probability-sampling.md
```diff
@@ -0,0 +1,56 @@
+# Logs Sampling
```
reyang, line 1: The file name and the title of the doc doesn't match (the title of the doc is very generic, while the file name is specific to "probability sampling").

atoulme: OK fixed!

specification/logs/log-probability-sampling.md
```diff
@@ -0,0 +1,56 @@
...
+- [Requirements](#requirements)
+  * [Deterministic sampling source](#deterministic-sampling-source)
+
+<!-- tocstop -->
+
+</details>
+
+## Overview
+
+This specification defines how logs are sampled with OpenTelemetry.
```
reyang, line 22: This seems to be an overstatement. Maybe:

This specification defines how probability sampling works for logs in OpenTelemetry.

specification/logs/log-probability-sampling.md
```diff
@@ -0,0 +1,56 @@
...
+
+Log sampling inherits from the logic defined in tracing sampling. By default, a log will use its trace ID to determine
```
reyang, line 24: What if there is no trace ID / span context?

specification/logs/log-probability-sampling.md
```diff
@@ -0,0 +1,56 @@
...
+In particular, log sampling borrows and reuses the notion of a hash seed that can be explicitly set on the sampler
+to guarantee sampling is consistent with a series of collectors.
+
+Trace ID sampling may be explicitly disabled.
+
+Log sampling can consider another attribute as the source of its sampling. In that case, the attribute must be explicitly defined.
+When the attribute is not set, the trace ID will be used instead if present.
+
+If no explicit attribute is set or no attribute has been defined, and no trace ID is present or trace ID sampling is disabled,
+the log is not considered as part of the sampling process and must be collected.
```
reyang, line 36:
```suggestion
the log is not considered as part of the sampling process and MUST be collected.
```

reyang: I think the `MUST` statement is too strong. There are cases where trace is not enabled at all, and folks want to sample logs based on some sampling algorithm which doesn't take any attributes into account (e.g. take no more than 5 logs per second, take 1% of the logs).

reyang (changes requested):
I think there are many higher level questions to be answered.

https://github.com/open-telemetry/opentelemetry-specification/pull/2482/files#r850044193

@atoulme I would suggest that you bring this topic to "Specification: Sampling" SIG or the "Specification: General" SIG. You can find more info from https://github.com/open-telemetry/community#special-interest-groups.

specification/logs/log-probability-sampling.md
atoulme: fixed

specification/logs/log-probability-sampling.md
atoulme: * 5 logs per second implies a throughput filter, not based on probabilistic sampling.
* 1% of the logs overall can be achieved with probabilistic sampling.

I see what you mean with a default sampling rate regardless of the application of an attribute. However, you require an attribute value to be the source of the sampling, otherwise it is not possible to sample according to a rate.
Therefore, it is not possible to sample without considering data associated with the log record, such as an attribute.

specification/logs/log-probability-sampling.md
atoulme: As provisioned, then no sampling can be applied, unless an attribute is configured as sampling source.

atoulme:
Thank you for your help. I believe it would be great to receive a second review from someone else as well.
I am familiar with SIGs and am unsure a synchronous process will yield better direct value at this time. Please feel free to correct me, and in any case, I'd like to be prepared - would you have the list of higher level questions you mention?

reyang:
> would you have the list of higher level questions you mention?

1. What's the overall logging sampling direction.
2. How does probability sampling fit into the overall sampling direction for logging.
3. Should log sampling be affected by trace sampling by default or not (e.g. in other words, is it opt-in or opt-out).
4. If log sampling is influenced by trace sampling decision (whether by default, or opt-in), what to do when the log happened outside of a trace context.

atoulme:
I am going to try and provide clarity here as best I can. I will also invite @pmm-sumo , who bore most of this initiative and had the most informed opinions about this.

1. There is currently, as far as I know, no log sampling work taking place, and in general there has been no foray into log sampling. This is the first movement towards supporting log sampling. Please help keep me honest here.
2. Probabilistic log sampling fits as part of log sampling as the first initiative towards supporting log sampling, and is based on the hard work folks put towards probabilistic trace sampling.
3. By default, there is a conjunction between log and trace sampling. There is business benefit in sampling logs and traces in the same way, so you sample data uniformly and allow folks to correlate logs and traces. This is a business requirement expressed by @pmm-sumo . As expressed by the specification, the default behavior is to sample according to the trace ID if present.
4. The specification goes on to explain that in the absence or when opting out of trace ID sampling, the log sampling can take place if the operator of the collector defines an attribute to be used as sampling source. When that attribute holds a value, it is used as the source of the sampling algorithm and defines whether the log should be sampled. When absent, the mechanism is inoperative and the log must be sampled.

I will try and join the SIG meeting next week, though it is not easy for me to do so. Folks, please let me know what else can be done to advance this PR.

specification/logs/log-probability-sampling.md
```diff
@@ -0,0 +1,56 @@
...
+In particular, log sampling borrows and reuses the notion of a hash seed that can be explicitly set on the sampler
+to guarantee sampling is consistent with a series of collectors.
+
+Trace ID sampling may be explicitly disabled.
+
+Log sampling can consider another attribute as the source of its sampling. In that case, the attribute must be explicitly defined.
+When the attribute is not set, the trace ID will be used instead if present.
+
+If no explicit attribute is set or no attribute has been defined, and no trace ID is present or trace ID sampling is disabled,
+the log is not considered as part of the sampling process and MUST be collected.
```
pmm-sumo, line 36: I think we might mix two concepts here - sampling and filtering. 

Using Trace ID for sampling is a "trick" that allows to have consistent sampling decision without sharing sampler state. Reusing this concept for log sampling allows to achieve sampling consistent with trace probabilistic sampling (for logs which have trace id attached). So we avoid situation that there's a trace of interest that misses a log record (and the other way around).

Being able to use another attribute for consistent log record sampling might make sense if there's one that is known to have uniformly distributed values. I think it is a big requirement, not easy to be met in the logs space, where commonly one deals with multiple sources of data (system logs, application logs, etc.). It might be stil possible to assign the value in the pipeline though (e.g. "log_uuid") that could be used for the purpose.

I think that when trace id is not available, one should consider just random sampling according to the configured rate (though the implementation could assign different sampling rates depending on some other properties, like severity or record attributes)

pmm-sumo:
I think what's also worth bringing to the discussion is the answer to question "why do we need it". I believe in this case this breaks down to following reasons (there might be more):
1. When trace sampling happens, we want log sampling on logs with TraceId attached in a consistent way
2. We want to have an ability to reduce the volume of logs

I think the first one is pretty clear and the outcome of it should be extending probabilistic filtering to logs, which is largely described by this proposal.

For the second, there are more solutions available:
* filtering data (e.g. removing all logs with low severity levels)
* random sampling (essentially extending this proposal with ability to randomly sample log records when TraceId is not available)
* rule-based sampling (this can get more complex quickly, see composite policy in [tailsamplingprocessor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/tailsamplingprocessor)) - I think implementation wise it can be achieved easily on the collector side by having a filter for records that are subject of random sampling (so e.g. there might be one random sampling policy defined for log records with severity ERROR and above and another one for log records with severity INFO and WARN)

Out of the above, I believe that only the "random sampling" would be a concern of the specification as of now. The other two are (might be wrong here) implementation specific and be implemented without need for specification as OpenTelemetry Collector processors.

In any case, I think we need to have more clarity how does it affect SDK's and maybe some of the existing [OTEPS](https://github.com/open-telemetry/oteps/blob/main/text/trace/0170-sampling-probability.md). We consider TraceId-less logs as disconnected pieces of data so I believe it makes it much simpler as of now

@jmacd what do you think?

specification/logs/log-probability-sampling.md
atoulme: You cannot apply random sampling, since it would not satisfy having multiple log samplers in sequence. Otherwise, each sampler would filter randomly logs.

You're better off assigning a random value to an attribute at the start, and use that for all your log samplers.
This is convenient because it pushes the generation and distribution of the sampling source outside of this process.

pmm-sumo: > You cannot apply random sampling, since it would not satisfy having multiple log samplers in sequence. Otherwise, each sampler would filter randomly logs.

It depends how this are constructed and what's the aim. One approach is to have "cascading" samplers that further reduce the volume of logs, as needed. (e.g. Instrumentation SDK samples 10% of logs and then Agent samples 10% out of that, so efficiently 1%). In such case, this works as anticipated.

The other approach is to combine samplers with some rules (e.g. different scope or severity levels would get different sampling rates). So random sampling would be done only for log records that match certain criteria (and everything else would be passed). 

> You're better off assigning a random value to an attribute at the start, and use that for all your log samplers. This is convenient because it pushes the generation and distribution of the sampling source outside of this process.

So basically a random, uniformly distributed Log ID value?

In any case, I am not sure if we need this. For tracing this is required for consistent trace sampling, since we deal with distributed transaction that involves spans generated by different services, collected through multiple agents and so on. For logs, it is much simpler

jmacd:
(Sorry to have fallen behind on this. I will catch up.)

tigrannajaryan:
@open-telemetry/specs-logs-approvers please also have a look.

jmacd:
Several loose bits of feedback here:

1. The idea of consistent sampling between traces and logs is sound. When we have the trace context in hand, we're both able to tell when to record a log based on the trace's sampled flag.
2. The idea of using the probability associated with the trace to adjusted the count of the log event is also sound. **However** , we are not recording the `tracestate` value associated with the trace context in the log record, and it's the `p` value (or potentially others in the future) that tells us what to count. One solution would be to add tracestate to the log, which brings the log event to a total of 4 tracecontext-related fields.
3. @pmm-sumo's point that sampling is about both counting (probability sampling) and filtering is valid. We see the same in tracing, where the Sampler API is used both for choosing spans and for probability sampling. Mixing these two in a comprehensive way is still ahead of us, but the roadmap looks like: do something like Jaeger-remote sampling to configure filters, use the specified probability mechanism for counting, and these should work together. The same could be said for logging: with configurable filter mechanism you should be able to choose which logs might record _and then decide which probability to sample them at independently_.

We can also discuss ways to opt-in to other sources of randomness, subject to the question above about how to record probability if there's no tracestate. If we add a new mechanism to record the adjusted count in a log, then we can define alternative sampling strategies. I support discovery in this area, because I can see it being reasonable for example to perform consistent logs sampling based on the **SpanID** instead of the **TraceID** if the trace is not important and your only goal is to summarize logs, the result would be an all-or-none for logs of a particular span (which I think is useful) regardless of its trace.

The current mechanism specified for tracing was meant as a bridge to the point where W3C admits true randomness. That's on the verge of happening, where we'll have the option to put 56 bits (or so) of true randomness in the TraceID (I hear). If we get that far, either we add tracestate to the log record or we add another way to convey the adjusted count, and then it could be possible to switch between consistent logs sampling based on span ID or trace ID. If you can get something like that to work, then it should be equally do-able with another source of randomness (e.g., `logs.uuid`), but as you can see there are a lot of hurdles here.

PR #2485, opened 2022-04-11
Initial addition of Kafka metrics.

carlosalberto:
Initial addition of metrics Kafka metrics as reported by the [JMX gatherer](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/jmx-metrics) (leaving out the Consumer specific ones for now).

The present PR has what the JMX Gatherer has **today**, but I'd like to get feedback regarding some potential changes:

* Some of these are **pre-cooked** metrics (directly from Kafka), e.g. counts in `nanoseconds`, some of them in `milliseconds` instead, and more importantly 50th/99th percentiles. Do we 'massage' them?
* All the gauges reported here should maybe be converted from Gauge to UpDownCounter.
* Name prefix for broker-specific metrics - mostly, some of the general Kafka metrics could have a `broker` component IMO, e.g. `kafka.message.count` maybe should become `kafka.broker.message.count`?
* A few of the metrics use a `count` sufix. The metric guidelines seem to suggest use of plurals instead? e.g. rename `kafka.messages.count` to `kafka.messages` (or `kafka.total-messages`).
* Is there any preference between usage of `-` and `_` in names?

cc @jmacd


pyohannes:
It would be worth collaborating with Kafka folks on this. There are ideas on Kafka side to natively expose OTel Metrics. I'm not sure about the state of this effort, but there's a recent proposal here: https://cwiki.apache.org/confluence/display/KAFKA/KIP-714%3A+Client+metrics+and+observability#KIP714:Clientmetricsandobservability-Metricsnamingandformat

The [metric names proposed in the above document](https://cwiki.apache.org/confluence/display/KAFKA/KIP-714%3A+Client+metrics+and+observability#KIP714:Clientmetricsandobservability-Clientinstance-levelProducermetrics) seem to follow a different convention, neither `_` nor `-` are used.

carlosalberto:
@pyohannes That looks interesting, thanks for posting that! 

I think the plan is to keep the current JMX metrics around, but we should unify the used metric conventions first (then, when Kafka goes OTLP we can stop relying on JMX, etc). Will drop them a note ;)

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,61 @@
...
+| kafka.request.count                  | Asynchronous Counter | Int64      | requests | `{requests}` | The number of requests received by the broker. | `type` | `produce`, `fetch` |
+| kafka.request.failed                 | Asynchronous Counter | Int64      | requests | `{requests}` | The number of requests to the broker resulting in a failure. | `type`  | `produce`, `fetch` |
+| kafka.request.time.total             | Asynchronous Counter | Int64      | milliseconds | `ms` | The total time the broker has taken to service requests. | `type` | `Produce`, `FetchConsumer`, `FetchFollower` |
+| kafka.request.time.50p               | Asynchronous Gauge   | Double     | milliseconds | `ms` | The 50th percentile time the broker has taken to service requests. | `type` | `Produce`, `FetchConsumer`, `FetchFollower` |
+| kafka.request.time.99p               | Asynchronous Gauge   | Double     | milliseconds | `ms` | The 99th percentile time the broker has taken to service requests. | `type` | `Produce`, `FetchConsumer`, `FetchFollower` |
+| kafka.request.queue                  | Asynchronous Gauge   | Int64      | requests | `{requests}` | The number of requests in the request queue. | | |
```
jmacd, line 25:
```suggestion
| kafka.request.queue                  | UpDownCounter   | Int64      | requests | `{requests}` | The number of requests in the request queue. | | |
```
There's a pattern here, if the description starts with "The number of ..." and the type is Gauge, switch to UpDownCounter. Or, if the type is Gauge and the units are "{pseudo_units}", switch to UpDownCounter.

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,61 @@
...
+| kafka.network.io                     | Asynchronous Counter | Int64      | bytes | `by` | The bytes received or sent by the broker. | `state` | `in`, `out` |
+| kafka.purgatory.size                 | Asynchronous Gauge   | Int64      | requests | `{requests}` | The number of requests waiting in the purgatory.  | `type` | `produce`, `fetch` |
```
jmacd, line 27:
```suggestion
| kafka.purgatory.size                 | UpDownCounter   | Int64      | requests | `{requests}` | The number of requests waiting in the purgatory.  | `type` | `produce`, `fetch` |
```

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,61 @@
...
+| kafka.partition.count                | Asynchronous Gauge   | Int64      | partitions | `{partitions}` | The number of partitions in the broker.  | | |
```
jmacd, line 28:
```suggestion
| kafka.partition.count                | UpDownCounter   | Int64      | partitions | `{partitions}` | The number of partitions in the broker.  | | |
```

Does this include offline partitions?

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,61 @@
...
+| kafka.partition.offline              | Asynchronous Gauge   | Int64      | partitions | `{partitions}` | The number of offline partitions. | | |
```
jmacd, line 29:
```suggestion
| kafka.partition.offline              | UpDownCounter   | Int64      | partitions | `{partitions}` | The number of offline partitions. | | |
```

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,61 @@
...
+| kafka.partition.under_replicated     | Asynchronous Gauge   | Int64      | partition  | `{partitions}` | The number of under replicated partitions. | | |
```
jmacd, line 30:
```suggestion
| kafka.partition.under_replicated     | UpDownCounter   | Int64      | partition  | `{partitions}` | The number of under replicated partitions. | | |
```

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,61 @@
...
+| kafka.isr.operation.count            | Asynchronous Counter | Int64      | operations | `{operations}` | The number of in-sync replica shrink and expand operations. | `operation` | `shrink`, `expand` |
```
jmacd, line 31:
```suggestion
| kafka.isr.operation.count            | UpDownCounter | Int64      | operations | `{operations}` | The number of in-sync replica shrink and expand operations. | `operation` | `shrink`, `expand` |
```

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,61 @@
...
+| kafka.max.lag                        | Asynchronous Gauge   | Int64      | messages   | `{messages}`   | Max lag in messages between follower and leader replicas. | | |
```
jmacd, line 32: I wonder if there's a more-primitive metric, such as `kafka.follower.lag`, an UpDownCounter, that reports the number of messages each replica is lagging. Then the user can query this for the maximum value.

(Otherwise I agree this should be Gauge, not UpDownCounter, because you would not want to sum maximum values, the only reasonable aggregation of these is to compute the MAX over a group.)

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,61 @@
...
+| kafka.controller.active.count        | Asynchronous Gauge   | Int64      | controllers | `{controllers}` | Number of active controllers in the broker. | | |
```
jmacd, line 33:
```suggestion
| kafka.controller.active.count        | UpDownCounter   | Int64      | controllers | `{controllers}` | Number of active controllers in the broker. | | |
```

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,61 @@
...
+| kafka.leader.election.rate           | Asynchronous Counter | Int64      | elections | `{elections}` | Leader election rate (increasing values indicates broker failures). | | |
```
jmacd, line 34:
```suggestion
| kafka.leader.elections           | Counter | Int64      | elections | `{elections}` | The number of leader elections (increasing values indicates broker failures). | | |
```

 I don't like attaching "rate" to something that is a count because the rate of change is a meaningful signal from every count.

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,61 @@
...
+| kafka.unclean.election.rate          | Asynchronous Counter | Int64      | elections | `{elections}` | Unclean leader election rate (increasing values indicates broker failures). | | |
```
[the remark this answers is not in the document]
jmacd: What defines an unclean leader election? Does the total on line 34 include the unclean values on line 35?

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,61 @@
...
+| kafka.logs.flush.time.count          | Asynchronous Counter | Int64      | milliseconds | `ms` | The time it has taken to flush logs. | | |
+| kafka.logs.flush.time.50th           | Asynchronous Gauge   | Double     | milliseconds | `ms` | The 50th percentile time it has taken to flush logs. | | |
+| kafka.logs.flush.time.99th           | Asynchronous Gauge   | Double     | milliseconds | `ms` | The 99th percentile time it has taken to flush logs. | | |
```
jmacd, line 38:
```suggestion
| kafka.logs.flush.duration         | Histogram | Int64      | milliseconds | `ms` | The timing of logs flushing. | | |
```

Same remarks as for `kafka.request.duration`. Semantically a histogram, acceptable to downgrade to Summary.

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Kafka Producer Metrics
+
+**Description:** Kafka Producer level metrics.
+
+| Name                                 | Instrument         | Value type | Unit | Unit ([UCUM](README.md#instrument-units)) | Description    | Attribute Key | Attribute Values |
+| ------------------------------------ | ------------------ | ---------- | ---- | ----------------------------------------- | -------------- | ------------- | ---------------- |
+| kafka.producer.io-wait-time-ns-avg   | Asynchronous Gauge | Double     | nanoseconds | `ns` | The average length of time the I/O thread spent waiting for a socket ready for reads or writes. | client-id | `client-id` value |
```
jmacd, line 46:
```suggestion
| kafka.producer.io-wait.duration   | Histogram | Double     | nanoseconds | `ns` | The length of time the I/O thread spent waiting for a socket ready for reads or writes. | client-id | `client-id` value |
```

A zero bucket histogram defines Sum and Count which contain the average value.   This additionally allows plotting the number of IO waits as distinct from the average value.

I mean to say that semantically this is a Histogram instrument, but then in a standard configuration for Kafka it would be appropriate to configure a View that selects a zero bucket histogram to lower cost.

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,61 @@
...
+| kafka.producer.outgoing-byte-rate    | Asynchronous Gauge | Double     | bytes | `by`| The average number of outgoing bytes sent per second to all servers. | `client-id` | `client-id` value |
```
jmacd, line 47:
```suggestion
| kafka.producer.server.output  | Counter | Int64     | `{bytes}` |  The number of output bytes sent to each server. | `client-id` | `client-id` value |
```

Rates should be derived from Counters, generally, I think. Note OTel has a convention for bi-direcitonal transfer counters that end in `.io` but this is uni-directional.

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,61 @@
...
+| kafka.producer.request-latency-avg   | Asynchronous Gauge | Double     | milliseconds | `ms` | The average request latency. | `client-id` | `client-id` value |
+| kafka.producer.request-rate          | Asynchronous Gauge | Double     | requests | `{requests}` | The average number of requests sent per second. | `client-id` | `client-id` value |
```
jmacd, line 49:
```suggestion
| kafka.producer.request.duration   | Histogram | Double     | milliseconds | `ms` | The request latency. | `client-id` | `client-id` value |
```
As above, these are semantically a histogram that can be aggregated somewhere, including in a View configuration, down to the average value and count.

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,61 @@
...
+| kafka.producer.response-rate         | Asynchronous Gauge | Double     | responses | `{responses}` | Responses received per second. | `client-id` | `client-id` value |
```
jmacd, line 50:
```suggestion
| kafka.producer.responses         | Counter | Int64     | responses | `{responses}` | Responses received. | `client-id` | `client-id` value |
```

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,61 @@
...
+| kafka.producer.byte-rate             | Asynchronous Gauge | Double     | bytes | `by` | The average number of bytes sent per second for a specific topic. | `client-id` | `client-id` value |
```
jmacd, line 51:
```suggestion
| kafka.producer.topic.output             | Asynchronous Gauge | Double     | bytes | `by` | The number of bytes sent for a specific topic. | `client-id`, `topic` | `client-id` value and `topic` name |
```

Following the pattern for `kafka.producer.server.output` suggested above.

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,61 @@
...
+|                                      |                    |            |       |      |                                                                   | `topic`     | topic name        |
+| kafka.producer.compression-rate      | Asynchronous Gauge | Double     | compression rate | `{compressionrate}` | The average compression rate of record batches for a specific topic. | `client-id` | `client-id` value |
```
jmacd, line 53:
```suggestion
| kafka.producer.compression-ratio      | Asynchronous Gauge | Double     | compression rate | _unitless_ | The average compression rate of record batches for a specific topic. | `client-id`, `topic` | `client-id` value and `topic` name|
```

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,61 @@
...
+|                                      |                    |            |            |               |                                                                                                | `topic`     | topic name        |
+| kafka.producer.record-retry-rate     | Asynchronous Gauge | Double     | retry rate | `{retryrate}` | The average per-second number of retried record sends for a specific topic. | `client-id` | `client-id` value  |
+|                                      |                    |            |            |               |                                                                             | `topic`     | topic name         |
+| kafka.producer.record-send-rate      | Asynchronous Gauge | Double     | records sent rate | `{recordssentrate}` | The average number of records sent per second for a specific topic.  | `client-id` | `client-id` value  |
+|                                      |                    |            |                   |                     |                                                                      | `topic`     | topic name         |
```
jmacd, line 60: I suggest continuing to replace Gauges w/ the current average rate by Counters from which average rates can be derived. These should be errors, retries, and records sent by topic.

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,61 @@
...
+| kafka.message.count                  | Asynchronous Counter | Int64      | messages | `{messages}` | The number of messages received by the broker. | | |
+| kafka.request.count                  | Asynchronous Counter | Int64      | requests | `{requests}` | The number of requests received by the broker. | `type` | `produce`, `fetch` |
+| kafka.request.failed                 | Asynchronous Counter | Int64      | requests | `{requests}` | The number of requests to the broker resulting in a failure. | `type`  | `produce`, `fetch` |
+| kafka.request.time.total             | Asynchronous Counter | Int64      | milliseconds | `ms` | The total time the broker has taken to service requests. | `type` | `Produce`, `FetchConsumer`, `FetchFollower` |
+| kafka.request.time.50p               | Asynchronous Gauge   | Double     | milliseconds | `ms` | The 50th percentile time the broker has taken to service requests. | `type` | `Produce`, `FetchConsumer`, `FetchFollower` |
+| kafka.request.time.99p               | Asynchronous Gauge   | Double     | milliseconds | `ms` | The 99th percentile time the broker has taken to service requests. | `type` | `Produce`, `FetchConsumer`, `FetchFollower` |
```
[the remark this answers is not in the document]
dashpole: Summary definitely seems like a good backwards-compatible way of representing the items being replaced here.  I also agree that new/future instrumentation should prefer histograms to summaries.  Switching to a histogram from the current metrics would be breaking for users.

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,61 @@
...
+
+<!-- Re-generate TOC with `markdown-toc --no-first-h1 -i` -->
+
+<!-- tocstop -->
+
+## Kafka Metrics
+
+**Description:** General Kafka metrics.
+
+| Name                                 | Instrument           | Value type | Unit | Unit ([UCUM](README.md#instrument-units)) | Description    | Attribute Key | Attribute Values |
```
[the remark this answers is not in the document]
carlosalberto: Done.

specification/metrics/semantic_conventions/instrumentation/kafka.md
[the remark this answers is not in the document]
jmacd: > Switching to a histogram from the current metrics would be breaking for users.

The statement above is not an OpenTelemetry definition, "breaking" is what happens when a vendor cannot meaningfully use some data, but it's a per-vendor definition. There may be a vendor that provides a consistent user experience when Summary and Histogram data points are mixed. At the very least, the Sum, Count, Min, and Max aggregations are well defined in both cases, so vendors have options.

jmacd:
I understand now, the reason this PR and these conventions are problematic is that we have metrics data that is semantically produced from a Histogram or Summary instrument, but it has been precomputed by an external metrics system and is made available as individual Sum, Count, and quantile timeseries. As an OpenTelemetry API user, we have no way to capture a precomputed summary into OTLP as a single stream via the API, so there is a temptation to bypass the semantic definition and perpetuate the use of pre-calculated metrics. To me, it is not a good outcome if we specify individual precomputed timeseries in situations where semantically a Histogram or Summary instrument would be used, unless we also specify semantic-naming conventions for translating between OTLP Summary data points and individual timeseries.

I'm imagining optional, dedicated OTLP Exporter support for recognizing that a set of metrics are named like Summary timeseries (e.g., `X_sum`, `X_total`, `X_p50`, `X_p90` and so on) that the Exporter SHOULD recombine those series into individual Summary data points. That is, if we have both a set of naming conventions for precomputed Summary timeseries as well as a directive for Exporters to rewrite them as Summary points on the wire, then I think we have satisfied our mission; these semantic conventions will refer to Histogram semantics, receivers will pass through precomputed timeseries, and OTLP exporters will correct the problem.

specification/metrics/semantic_conventions/instrumentation/kafka.md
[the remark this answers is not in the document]
carlosalberto: Actually I haven't been able to confirm one or the another, as the docs do not throw enough light into this - however, my impression, based on running tests, is that they are proportional, i.e. `election.rate` includes the `unclean-election.rate` values. WTDY?

specification/metrics/semantic_conventions/instrumentation/kafka.md
carlosalberto: So actually I found this:

MaxLag (used for this metric): "Max lag in messages btw follower and leader replicas".
ConsumerLag (not used at all here): "Lag in messages per follower replica" (additional information includes topic/partition ids).

Sounds like the latter indeed has what you asked. Maybe then remove `kafka.max.lag` and go with `kafka.follower.lag` instead?

specification/metrics/semantic_conventions/instrumentation/kafka.md
carlosalberto: So the pattern here (as seen by what JMX reports) is that anything that is a **count-per-second** receives a `rate` in the name. With that clarification, should we still strip the `rate` part?

carlosalberto:
@jmacd Updated the PR. 

Summary is that the metrics that can be reported as Summary/Histogram will be added in a follow up PR, so I will be removing these for now.

carlosalberto:
An additional note: all the counter metrics seem to be rates (e.g. `kafka.requests.failed` comes from `FailedProduceRequestsPerSec`, and `kafka.request.count` comes from `TotalProduceRequestsPerSec`).

Existing instrumentation (e.g. DataDog) seems to use these values as `rate`, rather than `counter`; and this IBM documentation hints that the granularity is 1 second: https://www.ibm.com/docs/en/obi/current?topic=technologies-monitoring-kafka

specification/metrics/semantic_conventions/instrumentation/kafka.md
carlosalberto: Any reason to change this to an UpDownCounter? Other instrumentations report this as a rate/counter.

specification/metrics/semantic_conventions/instrumentation/kafka.md
jmacd: My understanding is that metrics systems like Prometheus and data formats / APIs like OpenTelemetry encourage counting events instead of precomputing rates, because counter data aggregates correctly and rates can be correctly computed from counters, while aggregating rates directly leads to average rates but cannot be used to compute global rates.

That is why I would strip the `.rate` and specify the convention as a Counter. I understand that current instrumentation is outputting the precomputed `.rate` of a counter. I might then write a generic semantic convention that metrics named `X.rate` for `X` with units `U` are semantically equal to be the rate of `X` with units `U/s` as a Gauge series.

This allows dashboards to be built from existing instrumentation; existing instrumentation can be migrated to use Counters; metric processors can be written to convert from `X.rate` to `X` and `X` to `X.rate` as a way to assist with migration, and so on.

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,51 @@
...
+| ------------------------------------ | ------------- | ---------- | ------ | ----------------------------------------- | -------------- | ------------- | ---------------- |
+| kafka.producer.outgoing-byte-rate    | UpDownCounter | Double     | bytes | `by`| The average number of outgoing bytes sent per second to all servers. | `client-id` | `client-id` value |
+| kafka.producer.response-rate         | UpDownCounter | Double     | responses | `{responses}` | The average number of responses received per second. | `client-id` | `client-id` value |
+| kafka.producer.byte-rate             | UpDownCounter | Double     | bytes | `by` | The average number of bytes sent per second for a specific topic. | `client-id` | `client-id` value |
+|                                      |               |            |       |      |                                                                   | `topic`     | topic name        |
+| kafka.producer.compression-rate      | UpDownCounter | Double     | compression rate | `{compressionrate}` | The average compression rate of record batches for a specific topic. | `client-id` | `client-id` value |
```
jmacd, line 43:
```suggestion
| kafka.producer.compression-ratio      | Gauge | Double     | compression ratio | `{compression}` | The average compression rate of record batches for a specific topic. | `client-id` | `client-id` value |
```

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,51 @@
...
+|                                      |               |            |                  |                     |                                                                      | `topic`     | topic name        |
+| kafka.producer.record-error-rate     | UpDownCounter | Double     | error rate | `{errorrate}` | The average per-second number of record sends that resulted in errors for a specific topic.  | `client-id` | `client-id` value |
```
jmacd, line 45:
```suggestion
| kafka.producer.record-errors.rate     | Gauge | Double     | error rate | `{errors}/s` | The average per-second number of record sends that resulted in errors for a specific topic.  | `client-id` | `client-id` value |
```

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,51 @@
...
+|                                      |               |            |            |               |                                                                                              | `topic`     | topic name        |
+| kafka.producer.record-retry-rate     | UpDownCounter | Double     | retry rate | `{retryrate}` | The average per-second number of retried record sends for a specific topic. | `client-id` | `client-id` value  |
```
jmacd, line 47:
```suggestion
| kafka.producer.record-retry.rate     | Gauge | Double     | retry rate | `{retries}/s` | The average per-second number of retried record sends for a specific topic. | `client-id` | `client-id` value  |
```

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,51 @@
...
+|                                      |               |            |            |               |                                                                             | `topic`     | topic name         |
+| kafka.producer.record-send-rate      | UpDownCounter | Double     | records sent rate | `{recordssentrate}` | The average number of records sent per second for a specific topic.  | `client-id` | `client-id` value  |
```
jmacd, line 49:
```suggestion
| kafka.producer.records-sent.rate      | Gauge | Double     | records sent rate | `{records_sent}/s` | The average number of records sent per second for a specific topic.  | `client-id` | `client-id` value  |
```

jmacd (commented):
On the topic of precomputed rate metrics, see https://github.com/open-telemetry/opentelemetry-specification/pull/2485#discussion_r866384374. 

Suggest using `-ratio` for compression ratio, which is different than all the other rates in this PR. I've suggested changing some of the UpDownCounters to Gauge when the units are unitless (as -ratio) or {things}/s (as rates derived from counters).

You see there are three `.rate` suggestions here. I would support a generic semantics convention stating that a metric named `X.rate` is the rate derived from a Counter or UpDownCounter named `X`; this would allow us to specify the conventions as Counters even though current instrumentation reports the rate.

carlosalberto:
@jmacd @reyang Please review ;)

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,51 @@
...
+
+<!-- tocstop -->
+
+## Kafka Metrics
+
+**Description:** General Kafka metrics.
+
+| Name                                 | Instrument    | Value type | Unit   | Unit ([UCUM](../README.md#instrument-units)) | Description    | Attribute Key | Attribute Values |
+| ------------------------------------ | ------------- | ---------- | ------ | -------------------------------------------- | -------------- | ------------- | ---------------- |
+| kafka.message.count                  | Counter       | Int64      | messages | `{messages}` | The number of messages received by the broker. | | |
```
jmacd, line 20:
```suggestion
| kafka.messages                 | Counter       | Int64      | messages | `{messages}` | The number of messages received by the broker. | | |
```

The semantic conventions guidelines suggest pluralizing things that are being counted, not using the "count" or "total" suffix.

carlosalberto: I like it.

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Kafka Producer Metrics
+
+**Description:** Kafka Producer level metrics.
+
+| Name                                 | Instrument    | Value type | Unit   | Unit ([UCUM](../README.md#instrument-units)) | Description    | Attribute Key | Attribute Values |
+| ------------------------------------ | ------------- | ---------- | ------ | -------------------------------------------- | -------------- | ------------- | ---------------- |
+| kafka.producer.outgoing-byte.rate    | UpDownCounter | Double     | bytes | `by`| The average number of outgoing bytes sent per second to all servers. | `client-id` | `client-id` value |
+| kafka.producer.response.rate         | UpDownCounter | Double     | responses | `{responses}` | The average number of responses received per second. | `client-id` | `client-id` value |
+| kafka.producer.byte-rate             | UpDownCounter | Double     | bytes | `by` | The average number of bytes sent per second for a specific topic. | `client-id` | `client-id` value |
```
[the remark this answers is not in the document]
jmacd: Is this correct?

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,51 @@
...
+| kafka.producer.record-error.rate     | Gauge         | Double     | error rate | `{error}s` | The average per-second number of record sends that resulted in errors for a specific topic.  | `client-id` | `client-id` value |
+|                                      |               |            |            |               |                                                                                              | `topic`     | topic name        |
+| kafka.producer.record-retry.rate     | Gauge         | Double     | retry rate | `{retries}` | The average per-second number of retried record sends for a specific topic. | `client-id` | `client-id` value  |
+|                                      |               |            |            |               |                                                                             | `topic`     | topic name         |
+| kafka.producer.record-sent.rate      | Gauge         | Double     | records sent rate | `{records_sent}` | The average number of records sent per second for a specific topic.  | `client-id` | `client-id` value  |
```
jmacd, line 50:
```suggestion
| kafka.producer.outgoing-bytes.rate   | Gauge         | Double     | bytes per second | `by`/s | The average number of outgoing bytes sent per second to all servers. | `client-id` | `client-id` value |
| kafka.producer.responses.rate        | Gauge         | Double     | responses per second | `{responses}/s` | The average number of responses received per second. | `client-id` | `client-id` value |
| kafka.producer.bytes.rate            | Gauge         | Double     | bytes per second | `by/s` | The average number of bytes sent per second for a specific topic. | `client-id` | `client-id` value |
|                                      |               |            |       |      |                                                                   | `topic`     | topic name        |
| kafka.producer.compression-ratio     | Gauge         | Double     | compression ratio | `{compression}` | The average compression ratio of record batches for a specific topic. | `client-id` | `client-id` value |
|                                      |               |            |                  |                     |                                                                      | `topic`     | topic name        |
| kafka.producer.record-error.rate     | Gauge         | Double     | error rate | `{errors}/s` | The average per-second number of record sends that resulted in errors for a specific topic.  | `client-id` | `client-id` value |
|                                      |               |            |            |               |                                                                                              | `topic`     | topic name        |
| kafka.producer.record-retry.rate     | Gauge         | Double     | retry rate | `{retries}/s` | The average per-second number of retried record sends for a specific topic. | `client-id` | `client-id` value  |
|                                      |               |            |            |               |                                                                             | `topic`     | topic name         |
| kafka.producer.record-sent.rate      | Gauge         | Double     | records sent rate | `{records_sent}/s` | The average number of records sent per second for a specific topic.  | `client-id` | `client-id` value  |
```

jmacd (approved):

specification/metrics/semantic_conventions/instrumentation/kafka.md
```diff
@@ -0,0 +1,51 @@
...
+
+<!-- tocstop -->
+
+## Kafka Metrics
+
+**Description:** General Kafka metrics.
+
+| Name                                 | Instrument    | Value type | Unit   | Unit ([UCUM](../README.md#instrument-units)) | Description    | Attribute Key | Attribute Values |
+| ------------------------------------ | ------------- | ---------- | ------ | -------------------------------------------- | -------------- | ------------- | ---------------- |
+| kafka.messages                       | Counter       | Int64      | messages | `{messages}` | The number of messages received by the broker. | | |
```
reyang, line 20: If we follow https://github.com/open-telemetry/opentelemetry-specification/pull/2525#issuecomment-1119752082, maybe this should be called "eventing.kafka.messages"? (I don't have strong opinion here)

reyang (approved):
LGTM.

PR #2495, opened 2022-04-18
Clarify MetricReader.Collect result

### LINKED ISSUE #2295: Clarify MetricReader collect result

When implementing MetricReader.Collect, the spec is said:

> Collect SHOULD provide a way to let the caller know whether it succeeded, failed or timed out.
> https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#collect

In JavaScript implementation, the timeout is only meaningful on the async instrument callback context. Synchronous instruments collection can not be interrupted or fail (if the SDK is implemented correctly). That is said, async instruments may time out or fail, but synchronous instruments always can be collected without failure. So to JavaScript implementation, it can do its best to collect metric instruments and inform the MetricReader that _some_ instruments collection failed, but here you can still get part of the result. 

However, the spec has no words on if the implementations _should_ achieve a greater availability on collection.

legendecas:
Fixes #2295

## Changes

- Add statements on MetricReader.Collect that it should provide better availability when partial errors occur.


specification/metrics/sdk.md
```diff
@@ -710,7 +710,9 @@ Instruments](./api.md#asynchronous-instrument-api) involved, their callback
 functions will be triggered.
 
 `Collect` SHOULD provide a way to let the caller know whether it succeeded,
-failed or timed out.
+failed or timed out. When the `Collect` operation fails or times out on
+part of the instruments, the SDK SHOULD return all successfully collected result
+and a failed reasons list to the caller.
```
reyang, line 715: This seems to be a high bar.

Consider the current tracing spec, for the Exporter only status is required https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#exportbatch.

Maybe use a weak wording, e.g. "MAY"?

legendecas: Thank you for the suggestion. Updated :)

reyang (approved):
LGTM.

specification/metrics/sdk.md
```diff
@@ -710,7 +710,9 @@ Instruments](./api.md#asynchronous-instrument-api) involved, their callback
 functions will be triggered.
 
 `Collect` SHOULD provide a way to let the caller know whether it succeeded,
-failed or timed out.
+failed or timed out. When the `Collect` operation fails or times out on
+part of the instruments, the SDK MAY return successfully collected results
```
cijothomas, line 714:
```suggestion
some of the instruments, the SDK MAY return successfully collected results
```

I think this is a slightly better language....

cijothomas (approved):

MrAlias (approved):

PR #2508, opened 2022-04-23
Specify how to obtain a Ruby thread's id

fallwith:
Update the span-general.md#source-code-attributes table for obtaining
thread ids and names across languages to explain how to obtain a
thread's id for Ruby.


Oberon00:
Is there always only one unique Ruby object per logical Thread? Otherwise I think this may have confusing results.

fallwith:
> Is there always only one unique Ruby object per logical Thread? Otherwise I think this may have confusing results.

Hi @Oberon00. The table row in question already had a value for the current thread's name. There is a `name` attribute available on the instance of the `Thread` class that is referenced by the table.

That instance of the `Thread` class is itself a unique object from Ruby's perspective and has a unique object id which is exposed via `object_id`.

arminru:
@open-telemetry/ruby-maintainers @open-telemetry/ruby-approvers PTAL

ahayworth (approved):
👋 Hello from @open-telemetry/ruby-approvers 👋 

Since the semantic conventions explicitly require that `thread.id` be the `Current "managed" thread ID (as opposed to OS thread ID).` - this would be the right way to go about it. 👍

arminru (approved):

reyang (approved):

PR #2510, opened 2022-04-25
Loosen requirement for a major version bump

pellared:
I think a change in support for a language/runtime version should not necessarily force a major version bump.

Examples where I agree that making a version bump would be an overkill:

- https://github.com/open-telemetry/opentelemetry-dotnet/issues/3147
- https://github.com/open-telemetry/opentelemetry-dotnet/issues/2064

However, bumping a minor version in such cases makes sense.

PS. _Maybe I wrongly understand what "a drop in support" means._

FYI @open-telemetry/dotnet-maintainers

### DIFF
--- a/specification/versioning-and-stability.md
+++ b/specification/versioning-and-stability.md
@@ -254,7 +254,7 @@ When a signal becomes stable, the version MUST be bumped to match the other stab
 
 ### Major version bump
 
-Major version bumps MUST occur when there is a breaking change to a stable interface, the removal of a deprecated signal, or a drop in support for a language or runtime version.
+Major version bumps MUST occur when there is a breaking change to a stable interface, the removal of a deprecated signal.
 Major version bumps SHOULD NOT occur for changes which do not result in a drop in support of some form.
 
 ### Minor version bump
--- a/specification/versioning-and-stability.md
+++ b/specification/versioning-and-stability.md
@@ -255,7 +255,7 @@ When a signal becomes stable, the version MUST be bumped to match the other stab
 
 ### Major version bump
 
-Major version bumps MUST occur when there is a breaking change to a stable interface, the removal of a deprecated signal.
+Major version bumps MUST occur when there is a breaking change to a stable interface or a deprecated signal is removed.
 Major version bumps SHOULD NOT occur for changes which do not result in a drop in support of some form.
 
 ### Minor version bump
@@ -288,7 +288,7 @@ Each language implementation SHOULD define
 how the removal of a supported language/runtime version
 affects its versioning.
 As a rule of thumb,
-it SHOULD take the recommended approach in the given ecosystem
+it SHOULD follow the conventions in the given ecosystem.
 
 ## Long Term Support
 
--- a/specification/versioning-and-stability.md
+++ b/specification/versioning-and-stability.md
@@ -24,6 +24,7 @@
   * [Major version bump](#major-version-bump)
   * [Minor version bump](#minor-version-bump)
   * [Patch version bump](#patch-version-bump)
+  * [Language version support](#language-version-support)
 - [Long Term Support](#long-term-support)
   * [API support](#api-support)
   * [SDK Support](#sdk-support)
@@ -280,6 +281,14 @@ The following are examples of patch fixes.
 Currently, the OpenTelemetry project does NOT have plans to backport bug and security fixes to prior minor versions of the SDK.
 Security and bug fixes MAY only be applied to the latest minor version.
 We are committed to making it feasible for end users to stay up to date with the latest version of the OpenTelemetry SDK.
+
+### Language version support
+
+Each language implementation SHOULD define
+how the removal of a supported language/runtime version
+affects its versioning.
+As a rule of thumb,
+it SHOULD take the recommended approach in the given ecosystem
 
 ## Long Term Support
 


specification/versioning-and-stability.md
```diff
@@ -254,7 +254,7 @@ When a signal becomes stable, the version MUST be bumped to match the other stab
 
 ### Major version bump
 
-Major version bumps MUST occur when there is a breaking change to a stable interface, the removal of a deprecated signal, or a drop in support for a language or runtime version.
+Major version bumps MUST occur when there is a breaking change to a stable interface, the removal of a deprecated signal.
```
cijothomas, line 257: Specifically for the .NET case... (dont know if this is similar for other languages..)
we can leave this part of spec as-is, and add a clarification/exception to this: if the underlying language/runtime version itself is going out-of-support, then Major version bump need not be done...
(The key reason OTel .NET dropped support for an underlying runtime is because that runtime version itself is unsupported.)

pellared: All languages I "know" (.NET, Go, Python, Java) have some policies regarding version support.

I am open to any proposal to clarify it. I have currently no idea how to phrase it in a clear way that will not bring any potential confusion.

pellared: It looks like OTel Python has a similar policy https://github.com/open-telemetry/opentelemetry-python/issues/2360

@open-telemetry/python-maintainers 

Similarly for OTel Go: https://github.com/open-telemetry/opentelemetry-go/pull/2679

pellared:
@open-telemetry/python-maintainers  @open-telemetry/go-maintainers

specification/versioning-and-stability.md
tigrannajaryan: I support removing this clause. I don't see why it is necessary to increment Otel major version number because we no longer support an old version of a runtime.

@tedsuo you wrote this originally, maybe you can help clarify what was the idea.

tigrannajaryan (approved):

bogdandrutu (approved):

specification/versioning-and-stability.md
dyladan: I think because a user who tests with some version expects to be able to accept minor version increases without their application breaking. If they are on a runtime which loses support, this expectation is violated.

alanwest (approved):

Rauno56 (approved):
Even though it's a common practice to increment major versions of a package when even an EOL verision of the runtime is dropped in JS, I don't think it makes sense for API and probably SDK.
Bumping in that case would hurt interoperability and adoption immensely long term.

specification/versioning-and-stability.md
dyladan: I agree different language users have very different expectations. The definition of "support" can also be different language-to-language. For example in some languages it may be impossible to even compile with an older compiler, but in another language it may simply mean it is simply not tested with that runtime.

pellared: > Perhaps the wording should not be simply removed, but clarified that "may or may not require a major bump, see specific SIG guidelines/policy"

Proposal: 

```md
### Language or runtime version support

Each language implementation SHOULD define how the removal of a supported language/runtime version
affects its versioning. As a rule of thumb, it SHOULD take the recommended approach in a given ecosystem. 
```

PS. I find it hard to read. Edits are more than welcome.

yurishkuro: lgtm

reyang (approved):

specification/versioning-and-stability.md
ocelotl: Yes, definitely. I agree with removing this part of the spec, major version bumps should happen when breaking API changes happen not when a specific version of the interpreter (or equivalent in other languages) happens.

ocelotl (approved):

specification/versioning-and-stability.md
pellared: Addressed https://github.com/open-telemetry/opentelemetry-specification/pull/2510/commits/b13196b6616fab4369ca1255ac7c504894050445

specification/versioning-and-stability.md
```diff
@@ -254,7 +255,7 @@ When a signal becomes stable, the version MUST be bumped to match the other stab
...
```
yurishkuro, line 258:
```suggestion
Major version bumps MUST occur when there is a breaking change to a stable interface or a deprecated signal is removed.
```

specification/versioning-and-stability.md
```diff
@@ -281,6 +282,14 @@ Currently, the OpenTelemetry project does NOT have plans to backport bug and sec
 Security and bug fixes MAY only be applied to the latest minor version.
 We are committed to making it feasible for end users to stay up to date with the latest version of the OpenTelemetry SDK.
 
+### Language version support
+
+Each language implementation SHOULD define
+how the removal of a supported language/runtime version
+affects its versioning.
+As a rule of thumb,
+it SHOULD take the recommended approach in the given ecosystem
```
yurishkuro, line 291:
```suggestion
it SHOULD follow the conventions in the given ecosystem.
```

specification/versioning-and-stability.md
```diff
@@ -281,6 +282,14 @@ Currently, the OpenTelemetry project does NOT have plans to backport bug and sec
 Security and bug fixes MAY only be applied to the latest minor version.
 We are committed to making it feasible for end users to stay up to date with the latest version of the OpenTelemetry SDK.
 
+### Language version support
+
+Each language implementation SHOULD define
```
yurishkuro, line 287: Could you format the paragraph more densely, without random breaks?

pellared: It is not random. I tried to separate it by clauses. I find shorter lines easier to analyze when reviewing  diffs or when adding suggestions.

reyang: @pellared I would suggest that we pick a way that can be easily supported by some auto-formatting tool.

@yurishkuro please let us know if you think is a blocking issue or not (I personally feel it's not, although I agree that it should be more dense).

reyang: @yurishkuro I'll go ahead and merge this PR. If we feel strong about the wrapping, it can be a simple follow up editorial PR.

PR #2511, opened 2022-04-25
Semantic conventions for Summary metrics

jmacd:
Part of #2485 

## Changes

#2485 deals in specifying conventional names for metrics that have been traditionally pre-calculated about a distribution of measurements. When these measurements originate in Prometheus/OpenMetrics, we have a purpose-defined data point that can be used. When these measurements originate in a system with delta temporality or outside of Prometheus, they may exist in an "expanded form" as individual counter and gauge series.

This PR establishes semantic-conventional defaults for such pre-calculated summary metrics. Using these conventions, the semantic conventions for Kafka measurements can be written as "Histogram" while existing instrumentation may still output pre-calculated metrics in one of several ways. They can be expressed using the OTLP Summary point (if cumulative temporality), the OTLP Histogram data point (if delta temporality and no percentiles), or the expanded form (multiple metrics).

### DIFF
--- a/specification/metrics/semantic_conventions/README.md
+++ b/specification/metrics/semantic_conventions/README.md
@@ -168,10 +168,66 @@ total) are dimensionless and SHOULD use the default unit `1` (the unity).
 give additional meaning *without* the leading default unit (`1`). For example,
 use `{packets}`, `{errors}`, `{faults}`, etc.
 
-### Instrument Types
+### Instrument Conventions
 
-The semantic metric conventions specification is written to use the names of the synchronous instrument types,
-like `Counter` or `UpDownCounter`. However, compliant implementations MAY use the asynchronous equivalent instead,
+Metrics semantic conventions are specificatied using the names of the synchronous instrument types,
+like `Counter` or `UpDownCounter`.  Implementations MAY use the asynchronous equivalent instead,
 like `Asynchronous Counter` or `Asynchronous UpDownCounter`.
 Whether implementations choose the synchronous type or the asynchronous equivalent is considered to be an
 implementation detail. Both choices are compliant with this specification.
+
+Semantic conventions for capturing a distribution of recorded
+measurements SHOULD specify the use of a `Histogram` instrument (e.g.,
+`http.server.duration`).  Generally, consumers of data compliant with
+this specification SHOULD support any of the OpenTelemetry metrics
+data points defined for aggregating measurements, such as the
+Explicit-boundary Histogram or the Exponential Histogram.
+
+When reporting Metrics that have been previously calculated, in the
+case of metrics conventionally specified using `Histogram`
+instruments, consumers SHOULD support receiving OpenTelemetry
+`Summary` data points that were written by the external system.
+
+### Precalculated Summary Conventions
+
+The OpenTelemetry protocol (OTLP) Summary data point includes two
+fields in common with the OpenTelemetry Histogram data points,
+including the `sum` of recorded measurements and `count` of
+observations, which allows drawing a semantic correspondence between
+the two.
+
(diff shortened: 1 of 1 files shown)


jmacd:
@jack-berg This didn't make it into the meeting today, please take a look.

specification/metrics/semantic_conventions/README.md
```diff
@@ -168,10 +168,66 @@ total) are dimensionless and SHOULD use the default unit `1` (the unity).
...
+measurements SHOULD specify the use of a `Histogram` instrument (e.g.,
+`http.server.duration`).  Generally, consumers of data compliant with
+this specification SHOULD support any of the OpenTelemetry metrics
+data points defined for aggregating measurements, such as the
+Explicit-boundary Histogram or the Exponential Histogram.
+
+When reporting Metrics that have been previously calculated, in the
+case of metrics conventionally specified using `Histogram`
+instruments, consumers SHOULD support receiving OpenTelemetry
+`Summary` data points that were written by the external system.
```
jack-berg, line 189: This implies that the the consumer is able to know whether the metric was written by an external system. 

An alternative would be to update the paragraph on 179-185 to indicate that consumers of `Histogram` instruments should accept Summaries, in addition to Explicit-boundary Histogram and Exponential Histogram.

specification/metrics/semantic_conventions/README.md
```diff
@@ -168,10 +168,66 @@ total) are dimensionless and SHOULD use the default unit `1` (the unity).
...
+specification.  Semantic-conventional metrics specified for
+`Histogram` instruments MAY be expressed using Summary data points
+instead, as they have equivalent semantic interpretation, when they
+originate from OpenMetrics instrumentation.
+
+Receivers and exporters that handle precalculated metrics that cannot
+use an existing OpenTelemetry or OpenMetrics data point definition MAY
+use an expanded form containing multiple metrics to convey
+precalculated information about a distribution of measurements named
+`H`.
```
jack-berg, line 209: Do you know of any system that would take advantage of this expanded form?

specification/metrics/semantic_conventions/README.md
```diff
@@ -168,10 +168,66 @@ total) are dimensionless and SHOULD use the default unit `1` (the unity).
...
+observations, which allows drawing a semantic correspondence between
+the two.
+
+The Summary data point is defined exactly as in the OpenMetrics
+specification.  Semantic-conventional metrics specified for
+`Histogram` instruments MAY be expressed using Summary data points
+instead, as they have equivalent semantic interpretation, when they
+originate from OpenMetrics instrumentation.
+
+Receivers and exporters that handle precalculated metrics that cannot
```
dashpole, line 205: I think we need to treat receivers and exporters differently here.  Receivers (e.g. Prometheus receiver) isn't going to look for metrics with the naming patterns below, since it has an established naming convention for summaries already.  I expect other receivers that can receive summary-like data will have different conventions that the receiver will need to adapt to.

For exporters, the idea of converting summaries into individual timeseries in a consistent way makes more sense to me as a thing we could specify.

dashpole (commented):
I buy that vendors can provide compatibility between Histograms and Summaries to assist in migrations from legacy instrumentation, but I don't think we need to make statements about histogram/summary compatibility within OTel, as the data types are not currently translatable without data loss.  I would prefer:

1. Expanding the summary type to accommodate additional sources of quantile-based aggregations (e.g. with avg, min, max, etc.)
2. Defining conventions (or a set of conventions) for converting from summary metrics to other types (a series of gauges and counters) for exporters which don't natively support summaries.
3. Leave conversion from individual series to summaries up to individual receivers.

jmacd:
This needs more consideration.

PR #2512, opened 2022-04-25
Specify /v1/logs path in a few places where it was missing

tigrannajaryan:
It is already specified at https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp-request
but I think for completeness we want it everywhere the other
signals paths are mentioned.


arminru (approved):

tigrannajaryan:
This is an editorial change, does not require extensive review. Merging.

PR #2515, opened 2022-04-26
Add app and os semantic conventions

ladd:
## Changes

- Add `app` and `os` semantic conventions for describing language, timezone, region, and configuration version.


tigrannajaryan:
Please describe some use cases to understand if this belong to OpenTelemetry semantic conventions or no. Why one would want to record these as attributes?

ladd:
> Please describe some use cases to understand if this belong to OpenTelemetry semantic conventions or no. Why one would want to record these as attributes?

Our use case is production troubleshooting for end user applications. Specific types of questions we'd like to be able to answer:

- For a specific trace, does the user have a stale configuration version?
- Does the user's language setting match a supported language in the service?
- For a service computing shipping time, is user's local timezone correctly accounted for?
- In aggregate, what is the distribution of `country`, `language`, `timezone`, `version` values?

yurishkuro (commented):
this looks more like `user.agent.*` attributes, `configuration.` is too broad.

ladd:
> this looks more like `user.agent.*` attributes, `configuration.` is too broad.

I don't see any existing definition of `user.agent` -- are you proposing I create that?

specification/resource/semantic_conventions/configuration.md
```diff
@@ -0,0 +1,24 @@
...
+
+**type:** `configuration`
+
+**Description:** Any configurable software component.
+
+<!-- semconv configuration -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `configuration.country` | string | An ISO country code [1] | `US` | No |
+| `configuration.language` | string | A BCP 47 language tag [2] | `en-US`; `zh-Hant-CN` | No |
```
reyang, line 13: Would you clarify what does this mean?
For example, there could be the operating system level locale setting, and each user session could have its own locale, and a process can have different locale. Even inside a process, a thread can choose to have different locale. https://docs.microsoft.com/windows/win32/api/winnls/nf-winnls-setthreadlocale

ladd: This is intended to be application-level language setting, which would normally be derived from OS-level language setting.

A few notes of nuance here:

- Language is often separated from Locale (Windows, macOS, iOS, and Android make this distinction, and probably others)
- Language may have multiple options in preference order -- this is intended to reflect the user's "top choice".

reyang: Thanks @ladd!
I feel this is under specified. E.g. in .NET each thread can have different "culture" which affects which language to be used while formatting an exception https://www.csharpcodi.com/csharp-examples/System.Globalization.CultureInfo.GetCultureInfo(string,%20string)/.

ladd:
> I think the name is way too generic, and the actual semantic is under specified.

Thanks for the feedback! I'm mostly thinking about this from an end-user application perspective. How would you feel about this being specified in terms of application configuration? `app.configuration.language` or `application.configuration.language` for example? I could create some example mapping from OS default APIs, but there will be some aspects that may be up to the discretion of the implementation.

reyang:
> How would you feel about this being specified in terms of application configuration?

+1 👍

I think this can be modeled as different schemas (e.g. app.lang, os.lang, etc.) pointing to the same type definition (e.g. what is a `language`).

ladd:
Reworked this PR into separate `app` and `os` definitions.

semantic_conventions/resource/os.yaml
```diff
@@ -52,10 +52,31 @@ groups:
...
+        brief: 'An IANA Time Zone Database name or NormalizedCustomID value'
+        note: >
+          Identifies the timezone of the operating system instance, using a timezone
+          identifier from the IANA [Time Zone Database](https://www.iana.org/time-zones).
+          This may be derived from OS APIs, such as `TimeZone.current` on iOS or
+          `TimeZone.getDefault()` on Android. Systems without a means of converting to
+          IANA identifiers may use Java's
+          [NormalizedCustomID](https://docs.oracle.com/javase/8/docs/api/java/util/TimeZone.html#NormalizedCustomID)
+          format.
+        examples: ['America/Los_Angeles', 'UTC', 'GMT-08:00']
```
scheler, line 79: Timezone is a changing property of OS and therefore it doesn't fit into a Resource under the current semantics of Resource being immutable.

Same with Languages, as the user can change his preferences. That doesn't mean the source of telemetry, identified by the Resource, has changed.

specification/resource/semantic_conventions/app.md
```diff
@@ -0,0 +1,22 @@
...
+**Description:** The end-user application.
+
+<!-- semconv app -->
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `app.config_version` | string | A configuration version [1] | `42`; `5.6.2`; `808c8a73edef058f84e4b51da57c07779e48ed94` | No |
+| `app.lang` | string | A BCP 47 language tag [2] | `en-US`; `zh-Hant-CN` | No |
+| `app.region` | string | An ISO country code or a UN M49 code [3] | `US`; `419` | No |
+| `app.version` | string | The version string of the application. | `6.5.0` | No |
+
```
scheler, line 16: We tried this before. As Client SIG, we tried asking for `app.name` as the attribute for identify the application name and make `service.name` optional which is currently mandatory. The term service isn't so applicable for client apps. However, the request to make `service.name` optional was turned down and we were suggested to use service.name for naming the client-apps as well. In other words, the `service` namespace is a synonym for app and hence these attributes should be under `service` namespace.  cc: @tedsuo

scheler (changes requested):

semantic_conventions/resource/os.yaml
Oberon00: This is somewhat fuzzy. The hostname can also change, for example. But if it's uncommon enough for it to change, it may make sense to just use the "initial" value at the time of resource detection.

Timezone, language, do seem tricky though, e.g. servers may handle requests from different users with different settings.

ladd: I'd be happy to define it as being collected at initial time. Different users doesn't seem like a huge issue to me: this is defined as the running OS context. So if you care about distinct users, you would collect telemetry directly from the user's device. Conversely, if you care about server-side settings, you collect there.

specification/resource/semantic_conventions/app.md
ladd: I'm OK using `service.version` for this, if that's the existing consensus.

PR #2469, opened 2022-04-05
Remove alternative attribute sets from HTTP semantic conventions

lmolkova:
Fixes #2114

Depends on #2522

## Changes

Here's a revision of what current instrumentations do (https://github.com/open-telemetry/opentelemetry-specification/issues/2114#issuecomment-1079473468) and discussions on required/optional. 

The TLDR: 
Require HTTP instrumentation to be consistent: 
- define one set of required attributes for server and client, clarify how hostname should be populated
- clarify what required and optional attributes mean
  - required: MUST provide if available
  - optional opt-in: MUST populate by default (if available), MAY allow disabling
  - optional opt-out: MUST NOT populate by default, MAY allow enabling
- categorize all attributes based on it

Proposal:

**Client:  the only set of attributes clients should support is: `http.url`, `net.peer.name`, `net.peer.port`**
1. require `http.url`, `net.peer.name` (use IP if only IP is avaialble), `net.peer.port` (when not default) 
2. Make `net.peer.ip` attribute optional (opt-in)

**Server: the only set of attributes servers should support is: `http.scheme`, `net.host.name`, `net.host.port`, `http.target`, `http.route`**
1. Require `net.host.name`, ask for it's value to be the best available server name with specific priority (server-name, host header, ip address)
2. Require (when available) `http.route` - it's an essential attribute for both server traces and metrics
3. Remove `http.host` - host header should be populated (opt-in) using `http.request.header.host`
4. Remove `http.server_name` - if it's available it's populated in `net.host.name`
5. Make `http.url` optional (opt-in)
6. Make `net.host.ip` attribute optional (opt-in)

Related issues #2028

Issue #2114, opened 2021-11-10
Optional vs required attributes in HTTP conventions and beyond


lmolkova:
Based on instrumentation SIG meeting discussion

## Problem

Optional attributes in semantic conventions leave the decision to populate them on the instrumentation library. 
However, this decision should be in the user/backend/distro hands: this is where you know what information you'd like to see and pay for.
So one library that doesn't populate e.g. `http.request_content_length` (let's put aside [the need to have it on spans](https://github.com/open-telemetry/opentelemetry-specification/issues/2028)) makes the decision for all backends and users of this library.

## Proposal

**TL;DR: Instrumentation should collect all useful attributes. When some attributes are truly optional (i.e. rarely needed), there should be a configuration option for them. This issue does not try to address configuration.**

- Instrumentation should populate all generally useful attributes (and we should probably clean them up at least for HTTP).
- When there is a need for optional attributes (e.g. because the collection is too expensive and/or attributes are not generally useful),  instrumentation libraries should provide configuration options and populate them in opt-in mode 
  - is seems the current approach to leave it up to language+per group of attributes (e.g. [http headers](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-request-and-response-headers)) might not scale well, but we can address it separately
- The consumption side (distro/backend/user) should decide which attributes to export. E.g. exporter can ignore `http.scheme` if it has no use for it. Nothing new here.

### What does it mean for HTTP (example)
This is a general semantic convention discussion, but if we go through this exercise for HTTP, here's what I suggest. 

**All of the below is just an example of this approach, we can discuss each of those suggestions separately from this issue**

1. Let's pick one set of common attributes to make collection consistent (perhaps a specific set worth a separate discussion) 
  - `http.url` for clients
  - e.g. `http.scheme`, `net.host.name`, `net.host.port`, `http.target` for servers
    - it boils down to unification on `http.host` `net.host.name` and `http.server_name`
    - if scheme is only there to reconstruct url - do we *really* need it?

2. It leaves us with a few more generally-useful attributes:
- client and server:
  - `http.method`
  - `http.status_code`
- server only:
  - `http.route` - I assume if your web framework supports it, you'd want it
  - `http.server_name` - it seems it's supported by a few instrumentations and if it is supported, it's important and generally-useful
    
[...]

PR #2522, opened 2022-05-03
Define attribute requirement levels

lmolkova:
Extracted general attribute requirement level from  https://github.com/open-telemetry/opentelemetry-specification/pull/2469

## Changes

Adds terminology to be used by semantic convention to declare attribute presence requirements.

We currently have 4 vaguely defined levels across semantic conventions:
- `always` - all instrumentations MUST provide
- `conditional` which we use to specify when an attribute should be provided (but we use it inconsistently on optional and required attributes)
  e.g. [rpc.service](
https://github.com/open-telemetry/opentelemetry-specification/blob/031630c818c60666b27764a5b7a0e4ed435f55c4/semantic_conventions/trace/rpc.yaml#L27-20) is recommended (but not required?)     
  ```yml
     required:
       conditional: "No, but recommended"
  ```
  or [http.retry_count](https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/trace/http.yaml#L114) that is optional and conditional
  ```yml
     required:
       conditional: If and only if a request was retried.
  ```

  or [db.name](https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/trace/database.yaml#L179) which is required if available
  ```yml
     required:
       conditional: Required, if applicable.
  ```
- none, which means optional
- requires explicit enablement (e.g. [HTTP headers](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-request-and-response-headers))

**This PR adds and clarifies the terminology:** 
- `required` - MUST provide
-  ~~`conditional` (naming suggestions are welcome)~~ `conditionally required` - MUST IF \<condition\>
- `recommended` - SHOULD provide
- ~~`opt-in`~~ `optional` - MAY provide (when enabled)

Here's the corresponding tooling change: https://github.com/open-telemetry/build-tools/pull/92

Related issues #2114.

### DIFF
--- a/specification/common/attribute-requirement-level.md
+++ b/specification/common/attribute-requirement-level.md
@@ -18,11 +18,10 @@ _This section applies to Log, and Metric, Resource, and Span describes requireme
 Following attribute requirement levels are specified:
 
 - **Required**. All instrumentations MUST populate the attribute. Semantic convention defining required attribute expects that an absolute majority instrumentation libraries and applications are able to efficiently retrieve and populate it, can ensure cardinality, security, and other requirements specific to signal defined by the convention. `http.method` is an example of a required attribute.
+_Note: Consumers of the telemetry may use `Required` attributes to identify specific semantic convention(s) the given telemetry item follows._
 
-- **Conditional**. All instrumentations SHOULD add the attribute when instrumented entity supports corresponding feature and the attribute value can be [efficiently retrieved and populated](#performance-suggestions). Semantic convention assigning `Conditional` level on an attribute, SHOULD clarify when attribute is expected to be populated.
-`http.route` is an example of a conditional attribute: is widely supported by HTTP web frameworks, but some low-level HTTP server implementations do not support it.
-_Note: For producers of telemetry `Required` and `Conditional` levels are semantically the same (under the assumption that `Required` attributes are always available). However, consumers may use this distinction to identify conventions or validate telemetry._
-< TODO need a better name, took 'conditional' from schema definition>
+- **Conditionally required**. All instrumentations MUST add the attribute when given condition is satisfied and the attribute value can be [efficiently retrieved and populated](#performance-suggestions). Semantic convention of a `Conditionally required` level of an attribute MUST clarify the condition under which the attribute is expected to be populated.
+`http.route` is an example of a conditionally required attribute to be populated when instrumented HTTP framework provides route information for the instrumented request. Some low-level HTTP server implementations do not support routing and corresponding instrumentations can't provide the attribute.
 
 - **Recommended**. Instrumentations SHOULD add the attribute by default if it's readily available and can be [efficiently populated](#performance-suggestions). Instrumentation MAY allow explicit configuration to disable recommended attributes.
 
@@ -30,15 +29,14 @@ _Note: For producers of telemetry `Required` and `Conditional` levels are semant
 
 The requirement level for attribute is defined by semantic conventions depending on attribute availability across instrumented entities, performance, security, and other factors. When defining requirement levels, semantic conventions MUST take into account signal-specific requirements. For example, Metric attributes that may have high cardinality can only be defined with **Opt-in** level.
 
-Semantic convention that refers to an attribute from another (e.g. general) semantic convention MAY modify the requirement level within its own scope. Otherwise, requirement level from referred semantic convention applies.
+Semantic convention that refers to an attribute from another semantic convention MAY modify the requirement level within its own scope. Otherwise, requirement level from referred semantic convention applies.
+For example, [Database semantic convention](../trace/semantic_conventions/database.md) references `net.transport` attribute defined in [General attributes](../trace/semantic_conventions/span-general.md) with `Conditionally required` level on it.
 
-Instrumentations that decide not to populate `Conditional` or `Recommended` attributes due to performance, security, privacy, or other consideration by default, SHOULD use the **Opt-in** requirement level on them if the attributes are logically applicable.
+Instrumentations that decide not to populate `Conditionally required` or `Recommended` attributes due to performance, security, privacy, or other consideration by default, SHOULD use the **Opt-in** requirement level on them if the attributes are logically applicable.
 
 ## Performance suggestions
 
-In some cases instrumented entities do not have `Conditional` or `Recommended` attribute value readily available and retrieval can be expensive. Instrumentations SHOULD NOT perform expensive resource-consuming operation to obtain attribute values. Corresponding attributes SHOULD NOT be populated by default.
-
-Here are several examples of expensive operations:
+Here are several examples of expensive operations to be avoided by default:
 
 - DNS lookup to populate `net.peer.name` if only IP address is available to the instrumentation.
 - forcing `http.route` calculation before HTTP framework calculates it
--- a/specification/common/attribute-requirement-level.md
+++ b/specification/common/attribute-requirement-level.md
@@ -18,7 +18,7 @@ _This section applies to Log, and Metric, Resource, and Span describes requireme
 Following attribute requirement levels are specified:
 
 - **Required**. All instrumentations MUST populate the attribute. Semantic convention defining required attribute expects that an absolute majority instrumentation libraries and applications are able to efficiently retrieve and populate it, can ensure cardinality, security, and other requirements specific to signal defined by the convention. `http.method` is an example of a required attribute.
-_Note: Consumers of telemetry may want to know if telemetry item follows specific semantic convention. They can rely on presence of `Required` attribute defined by such convention. For example, `db.system` attribute on a span can be used as an indication that span follows database semantics._
+_Note: Consumers of telemetry can detect if telemetry item follows specific semantic convention by checking presence of a `Required` attribute defined by such convention. For example, `db.system` attribute on a span can be used as an indication that span follows database semantics._
 
 - **Conditionally Required**. All instrumentations MUST add the attribute when given condition is satisfied and the attribute value can be [efficiently retrieved and populated](#performance-suggestions). Semantic convention of a `Conditionally Required` level of an attribute MUST clarify the condition under which the attribute is expected to be populated.
 `http.route` is an example of a conditionally required attribute to be populated when instrumented HTTP framework provides route information for the instrumented request. Some low-level HTTP server implementations do not support routing and corresponding instrumentations can't populate the attribute.
--- a/specification/common/attribute-requirement-level.md
+++ b/specification/common/attribute-requirement-level.md
@@ -1,0 +1,43 @@
+# Attribute Requirement Level
+
+**Status**: [Experimental](../document-status.md)
+
+<details>
+<summary>Table of Contents</summary>
+
+<!-- toc -->
+
+- [Performance considerations](#performance-considerations)
+
+<!-- tocstop -->
+
+</details>
+
+_This section applies to Resource, Span, Log, and Metric and describes requirement levels for attributes._
+
+Following attribute requirement levels are specified:
+
+- **Required**. All instrumentations MUST populate the attribute. Semantic convention defining required attribute expects that an absolute majority instrumentation libraries and applications are able to efficiently retrieve and populate it, can ensure cardinality, security, and other requirements specific to signal defined by the convention. `http.method` is an example of required attribute.
+
+- **Conditional**. All instrumentations MUST add the attribute when instrumented entity supports corresponding feature and the attribute value can be efficiently retrieved and populated (see [Performance Considerations](#performance-considerations)).
+For example, `http.route` is widely supported by HTTP web frameworks, but some low-level HTTP server implementations do not support it.
+_Note: For producers of telemetry `Required` and `Conditional` levels are semantically the same (under the assumption that `Required` attributes are always available). However, consumers may use this distinction to identify conventions or validate telemetry._
+< TODO need a better name, took 'conditional' from schema definition>
+
+- **Recommended**. Instrumentations SHOULD add the attribute by default if it's readily available and can be efficiently populated. Instrumentation MAY allow explicit configuration to disable recommended attributes.
+
+- **Opt-in**. Instrumentations SHOULD NOT add the attribute by default. Instrumentation SHOULD populate the attribute if and only if user configures instrumentation to do so. Instrumentation that doesn't support configuration MUST NOT populate **Opt-in** attributes.
+
+The requirement level for attribute is defined by semantic conventions depending on attribute availability across instrumented entities, performance, security, and other factors. When defining requirement levels, semantic conventions MUST take into account signal-specific requirements. For example, Metric attributes that may have high cardinality can only be defined with **Opt-in** level.
+
+Semantic convention that refers to attribute from another (e.g. general) semantic convention defaults to initial requirement level and MAY modify the requirement level in scope of a child convention.
+
+Instrumentation that cannot populate attribute according to defined requirement level due to performance, security, privacy, or other consideration, MUST use **Opt-in** requirement level on it.
+
+## Performance considerations
+
+In some cases instrumented entities do not have attribute value readily available and retrieval can be expensive. For example, `net.peer.name` may require DNS lookup if only IP address is available to the instrumentation.
+
+Instrumentations MUST NOT perform expensive resource-consuming operation to obtain attribute values. Corresponding attributes MUST NOT be populated by default.
+
+Basic local operations such as string formatting, escaping or arithmetic operations are allowed.
--- a/specification/common/attribute-requirement-level.md
+++ b/specification/common/attribute-requirement-level.md
@@ -19,25 +19,27 @@ Following attribute requirement levels are specified:
 
 - **Required**. All instrumentations MUST populate the attribute. Semantic convention defining required attribute expects that an absolute majority instrumentation libraries and applications are able to efficiently retrieve and populate it, can ensure cardinality, security, and other requirements specific to signal defined by the convention. `http.method` is an example of required attribute.
 
-- **Conditional**. All instrumentations MUST add the attribute when instrumented entity supports corresponding feature and the attribute value can be efficiently retrieved and populated (see [Performance Considerations](#performance-suggestions)).
+- **Conditional**. All instrumentations SHOULD add the attribute when instrumented entity supports corresponding feature and the attribute value can be [efficiently retrieved and populated](#performance-suggestions).
 For example, `http.route` is widely supported by HTTP web frameworks, but some low-level HTTP server implementations do not support it.
 _Note: For producers of telemetry `Required` and `Conditional` levels are semantically the same (under the assumption that `Required` attributes are always available). However, consumers may use this distinction to identify conventions or validate telemetry._
 < TODO need a better name, took 'conditional' from schema definition>
 
-- **Recommended**. Instrumentations SHOULD add the attribute by default if it's readily available and can be efficiently populated. Instrumentation MAY allow explicit configuration to disable recommended attributes.
+- **Recommended**. Instrumentations SHOULD add the attribute by default if it's readily available and can be [efficiently populated](#performance-suggestions). Instrumentation MAY allow explicit configuration to disable recommended attributes.
 
 - **Opt-in**. Instrumentations SHOULD NOT add the attribute by default. Instrumentation SHOULD populate the attribute if and only if user configures instrumentation to do so. Instrumentation that doesn't support configuration MUST NOT populate **Opt-in** attributes.
 
 The requirement level for attribute is defined by semantic conventions depending on attribute availability across instrumented entities, performance, security, and other factors. When defining requirement levels, semantic conventions MUST take into account signal-specific requirements. For example, Metric attributes that may have high cardinality can only be defined with **Opt-in** level.
 
-Semantic convention that refers to attribute from another (e.g. general) semantic convention defaults to initial requirement level and MAY modify the requirement level in scope of a child convention.
+Semantic convention that refers to an attribute from another (e.g. general) semantic convention MAY modify the requirement level within its own scope. Otherwise, requirement level from referred semantic convention applies.
 
-Instrumentation that cannot populate attribute according to defined requirement level due to performance, security, privacy, or other consideration, MUST use **Opt-in** requirement level on it.
+Instrumentation that cannot populate `Conditional` or `Recommended` attribute due to performance, security, privacy, or other consideration, MUST use **Opt-in** requirement level on it.
 
 ## Performance suggestions
 
-In some cases instrumented entities do not have attribute value readily available and retrieval can be expensive. For example, `net.peer.name` may require DNS lookup if only IP address is available to the instrumentation.
+In some cases instrumented entities do not have `Conditional` or `Recommended` attribute value readily available and retrieval can be expensive. Instrumentations SHOULD NOT perform expensive resource-consuming operation to obtain attribute values. Corresponding attributes SHOULD NOT be populated by default.
 
-Instrumentations SHOULD NOT perform expensive resource-consuming operation to obtain attribute values. Corresponding attributes SHOULD NOT be populated by default.
+Here are several examples of expensive operations:
 
-Basic local operations such as string formatting, escaping or arithmetic operations are allowed.
+- DNS lookup to populate `net.peer.name` if only IP address is available to the instrumentation.
+- forcing `http.route` calculation before HTTP framework calculates it
+- reading response stream to find `http.response_content_length` when `Content-Length` header is not available
--- a/specification/common/attribute-requirement-level.md
+++ b/specification/common/attribute-requirement-level.md
@@ -20,7 +20,7 @@ Following attribute requirement levels are specified:
 - **Required**. All instrumentations MUST populate the attribute. Semantic convention defining required attribute expects that an absolute majority instrumentation libraries and applications are able to efficiently retrieve and populate it, can ensure cardinality, security, and other requirements specific to signal defined by the convention. `http.method` is an example of a required attribute.
 _Note: Consumers of telemetry can detect if telemetry item follows specific semantic convention by checking presence of a `Required` attribute defined by such convention. For example, `db.system` attribute on a span can be used as an indication that span follows database semantics._
 
-- **Conditionally Required**. All instrumentations MUST add the attribute when given condition is satisfied and the attribute value can be [efficiently retrieved and populated](#performance-suggestions). Semantic convention of a `Conditionally Required` level of an attribute MUST clarify the condition under which the attribute is expected to be populated.
+- **Conditionally Required**. All instrumentations MUST add the attribute when given condition is satisfied. Semantic convention of a `Conditionally Required` level of an attribute MUST clarify the condition under which the attribute is expected to be populated.
 `http.route` is an example of a conditionally required attribute to be populated when instrumented HTTP framework provides route information for the instrumented request. Some low-level HTTP server implementations do not support routing and corresponding instrumentations can't populate the attribute.
 
 - **Recommended**. Instrumentations SHOULD add the attribute by default if it's readily available and can be [efficiently populated](#performance-suggestions). Instrumentation MAY offer a configuration option to disable recommended attributes.
@@ -32,7 +32,7 @@ The requirement level for attribute is defined by semantic conventions depending
 Semantic convention that refers to an attribute from another semantic convention MAY modify the requirement level within its own scope. Otherwise, requirement level from referred semantic convention applies.
 For example, [Database semantic convention](../trace/semantic_conventions/database.md) references `net.transport` attribute defined in [General attributes](../trace/semantic_conventions/span-general.md) with `Conditionally Required` level on it.
 
-Instrumentations that decide not to populate `Conditionally Required` or `Recommended` attributes due to performance, security, privacy, or other consideration by default, SHOULD use the **Optional** requirement level on them if the attributes are logically applicable.
+Instrumentations that decide not to populate `Conditionally Required` or `Recommended` attributes due to [performance](#performance-suggestions), security, privacy, or other consideration by default, SHOULD use the **Optional** requirement level on them if the attributes are logically applicable.
 
 ## Performance suggestions
 
--- a/specification/common/attribute-requirement-level.md
+++ b/specification/common/attribute-requirement-level.md
@@ -17,9 +17,9 @@ _This section applies to Log, and Metric, Resource, and Span describes requireme
 
 Following attribute requirement levels are specified:
 
-- **Required**. All instrumentations MUST populate the attribute. Semantic convention defining required attribute expects that an absolute majority instrumentation libraries and applications are able to efficiently retrieve and populate it, can ensure cardinality, security, and other requirements specific to signal defined by the convention. `http.method` is an example of required attribute.
+- **Required**. All instrumentations MUST populate the attribute. Semantic convention defining required attribute expects that an absolute majority instrumentation libraries and applications are able to efficiently retrieve and populate it, can ensure cardinality, security, and other requirements specific to signal defined by the convention. `http.method` is an example of a required attribute.
 
-- **Conditional**. All instrumentations SHOULD add the attribute when instrumented entity supports corresponding feature and the attribute value can be [efficiently retrieved and populated](#performance-suggestions).
+- **Conditional**. All instrumentations SHOULD add the attribute when instrumented entity supports corresponding feature and the attribute value can be [efficiently retrieved and populated](#performance-suggestions). Semantic convention assigning `conditional` level on an attribute, SHOULD clarify when attribute is expected to be populated.
 For example, `http.route` is widely supported by HTTP web frameworks, but some low-level HTTP server implementations do not support it.
 _Note: For producers of telemetry `Required` and `Conditional` levels are semantically the same (under the assumption that `Required` attributes are always available). However, consumers may use this distinction to identify conventions or validate telemetry._
 < TODO need a better name, took 'conditional' from schema definition>
@@ -32,7 +32,7 @@ The requirement level for attribute is defined by semantic conventions depending
 
 Semantic convention that refers to an attribute from another (e.g. general) semantic convention MAY modify the requirement level within its own scope. Otherwise, requirement level from referred semantic convention applies.
 
-Instrumentation that cannot populate `Conditional` or `Recommended` attribute due to performance, security, privacy, or other consideration, MUST use **Opt-in** requirement level on it.
+Instrumentations that decide not to populate `Conditional` or `Recommended` attributes due to performance, security, privacy, or other consideration by default, SHOULD use the **Opt-in** requirement level on them if the attributes are logically applicable.
 
 ## Performance suggestions
 
--- a/specification/common/attribute-requirement-level.md
+++ b/specification/common/attribute-requirement-level.md
@@ -1,4 +1,4 @@
-# Attribute Requirement Level
+# Attribute Requirement Levels for Semantic Conventions
 
 **Status**: [Experimental](../document-status.md)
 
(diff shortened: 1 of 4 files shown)


specification/common/attribute-requirement-level.md
```diff
@@ -0,0 +1,43 @@
...
+
+- **Opt-in**. Instrumentations SHOULD NOT add the attribute by default. Instrumentation SHOULD populate the attribute if and only if user configures instrumentation to do so. Instrumentation that doesn't support configuration MUST NOT populate **Opt-in** attributes.
+
+The requirement level for attribute is defined by semantic conventions depending on attribute availability across instrumented entities, performance, security, and other factors. When defining requirement levels, semantic conventions MUST take into account signal-specific requirements. For example, Metric attributes that may have high cardinality can only be defined with **Opt-in** level.
+
+Semantic convention that refers to attribute from another (e.g. general) semantic convention defaults to initial requirement level and MAY modify the requirement level in scope of a child convention.
+
+Instrumentation that cannot populate attribute according to defined requirement level due to performance, security, privacy, or other consideration, MUST use **Opt-in** requirement level on it.
```
reyang, line 35: _Instrumentation that cannot populate attribute according to defined requirement level due to ..._ - is this essentially making all the Required attributes `SHOULD` instead of `MUST`?

specification/common/attribute-requirement-level.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Performance considerations
+
+In some cases instrumented entities do not have attribute value readily available and retrieval can be expensive. For example, `net.peer.name` may require DNS lookup if only IP address is available to the instrumentation.
+
+Instrumentations MUST NOT perform expensive resource-consuming operation to obtain attribute values. Corresponding attributes MUST NOT be populated by default.
```
reyang, line 41: I think performance considerations are more like suggestions/recommendations:

```suggestion
Instrumentations SHOULD NOT perform expensive resource-consuming operation to obtain attribute values. Corresponding attributes MUST NOT be populated by default.
```

specification/common/attribute-requirement-level.md
```diff
@@ -0,0 +1,43 @@
...
+
+<!-- toc -->
+
+- [Performance considerations](#performance-considerations)
+
+<!-- tocstop -->
+
+</details>
+
+_This section applies to Resource, Span, Log, and Metric and describes requirement levels for attributes._
```
reyang, line 16: minor: consider alphabetical sorting

```suggestion
_This section applies to Log, Metric, Resource and Span and describes requirement levels for attributes._
```

specification/common/attribute-requirement-level.md
lmolkova: MUST is still more appropriate - it's semantic convention responsibility to not require something that's not available. But I assume there might be edge-case instrumentation that can't populate specific required attributes. E.g. out-of-process instrumentations may lack some important detail. 
If it's the case and it can't populate the required attribute, such instrumentations should not be even considered as following the specific semantic convention defining this required attribute. 

So I believe it's a better choice to make spec more strict rather than relaxing it for all instrumentations to accommodate rare edge cases.

If over time we see that edge case is more common than we thought, the semantic convention would need to switch the requirement level to conditional or recommended. At this point, it'll go through the normal versioning (potentially we'll need to define new transformation rules to the level).

specification/common/attribute-requirement-level.md
```diff
@@ -0,0 +1,43 @@
...
+< TODO need a better name, took 'conditional' from schema definition>
+
+- **Recommended**. Instrumentations SHOULD add the attribute by default if it's readily available and can be efficiently populated. Instrumentation MAY allow explicit configuration to disable recommended attributes.
+
+- **Opt-in**. Instrumentations SHOULD NOT add the attribute by default. Instrumentation SHOULD populate the attribute if and only if user configures instrumentation to do so. Instrumentation that doesn't support configuration MUST NOT populate **Opt-in** attributes.
+
+The requirement level for attribute is defined by semantic conventions depending on attribute availability across instrumented entities, performance, security, and other factors. When defining requirement levels, semantic conventions MUST take into account signal-specific requirements. For example, Metric attributes that may have high cardinality can only be defined with **Opt-in** level.
+
+Semantic convention that refers to attribute from another (e.g. general) semantic convention defaults to initial requirement level and MAY modify the requirement level in scope of a child convention.
```
tigrannajaryan, line 33: This sentence is not clear to me. What does "defaults" mean? What is a "child convention"?

specification/common/attribute-requirement-level.md
```diff
@@ -0,0 +1,43 @@
...
+
+Instrumentation that cannot populate attribute according to defined requirement level due to performance, security, privacy, or other consideration, MUST use **Opt-in** requirement level on it.
+
+## Performance suggestions
```
tigrannajaryan, line 37: Do considerations in this section apply only to "Conditional" level? Does it apply to "Required" or "Recommended"? It is currently referenced only from "Conditional" level. If it is applicable only to "Conditional" level, move it to the definition of "Conditional" or make applicability more clear here.

specification/common/attribute-requirement-level.md
lmolkova: clarified in the doc

specification/common/attribute-requirement-level.md
lmolkova: I removed the exception for `Required` attributes. 
Semantic convention authors should take all the conditions into account when declaring an attribute as `Required` and then it becomes a `MUST` for instrumentation authors to follow.

reyang (approved):
LGTM.

specification/common/attribute-requirement-level.md
```diff
@@ -0,0 +1,45 @@
...
+
+<!-- tocstop -->
+
+</details>
+
+_This section applies to Log, and Metric, Resource, and Span describes requirement levels for attributes._
+
+Following attribute requirement levels are specified:
+
+- **Required**. All instrumentations MUST populate the attribute. Semantic convention defining required attribute expects that an absolute majority instrumentation libraries and applications are able to efficiently retrieve and populate it, can ensure cardinality, security, and other requirements specific to signal defined by the convention. `http.method` is an example of required attribute.
```
Oberon00, line 20:
```suggestion
- **Required**. All instrumentations MUST populate the attribute. Semantic convention defining required attribute expects that an absolute majority instrumentation libraries and applications are able to efficiently retrieve and populate it, can ensure cardinality, security, and other requirements specific to signal defined by the convention. `http.method` is an example of a required attribute.
```

specification/common/attribute-requirement-level.md
```diff
@@ -0,0 +1,45 @@
...
+
+- **Conditional**. All instrumentations SHOULD add the attribute when instrumented entity supports corresponding feature and the attribute value can be [efficiently retrieved and populated](#performance-suggestions).
```
Oberon00, line 22: Can we specify that for "conditional" attributes it should (must?) be clarified in the semantic  conventions in which cases they apply/are required? I think this is even currently the case for the old "conditional" attributes.

specification/common/attribute-requirement-level.md
```diff
@@ -0,0 +1,45 @@
...
+
+- **Opt-in**. Instrumentations SHOULD NOT add the attribute by default. Instrumentation SHOULD populate the attribute if and only if user configures instrumentation to do so. Instrumentation that doesn't support configuration MUST NOT populate **Opt-in** attributes.
+
+The requirement level for attribute is defined by semantic conventions depending on attribute availability across instrumented entities, performance, security, and other factors. When defining requirement levels, semantic conventions MUST take into account signal-specific requirements. For example, Metric attributes that may have high cardinality can only be defined with **Opt-in** level.
+
+Semantic convention that refers to an attribute from another (e.g. general) semantic convention MAY modify the requirement level within its own scope. Otherwise, requirement level from referred semantic convention applies.
+
+Instrumentation that cannot populate `Conditional` or `Recommended` attribute due to performance, security, privacy, or other consideration, MUST use **Opt-in** requirement level on it.
```
Oberon00, line 35: I think this does not make sense for "conditional". For example, how should http.route be populated by some low-level HTTP server implementation (to reuse your example), even if the user explicitly requests it?

```suggestion
Instrumentations that decide not to populate `Conditional` or `Recommended` attributes due to performance, security, privacy, or other consideration by default, SHOULD use the **Opt-in** requirement level on them if the attributes are logically applicable.
```

Oberon00:
Generally looks very sensible to me 👍, my only concern is about the MUST for transforming conditional/recommended into opt-in.

specification/common/attribute-requirement-level.md
lmolkova: Good question, I don't think it makes sense for `http.route` unless instrumentation lets the user manually set a route in some callback (e.g. based on regex). 
I was thinking about something like `net.peer.name` DNS lookup. Instrumentation can do it more or less efficiently if it caches DNS lookup results. Doing it by default is not a great idea, but if a user enables it knowing the risks, it'd improve telemetry quality.

Oberon00: It sounds like something that may often work but not always, and also nothing will break if the instrumentation does not implement it at all. So I think a SHOULD-requirement instead of a MUST here would be enough

specification/common/attribute-requirement-level.md
```diff
@@ -0,0 +1,45 @@
...
+
+<!-- tocstop -->
+
+</details>
+
+_This section applies to Log, and Metric, Resource, and Span describes requirement levels for attributes._
+
+Following attribute requirement levels are specified:
+
+- **Required**. All instrumentations MUST populate the attribute. Semantic convention defining required attribute expects that an absolute majority instrumentation libraries and applications are able to efficiently retrieve and populate it, can ensure cardinality, security, and other requirements specific to signal defined by the convention. `http.method` is an example of a required attribute.
```
tigrannajaryan, line 20: I am a bit worried by how strongly this is worded. When designing a semantic convention that needs to be implemented in 10 languages and potentially in dozens of libraries It is hard to be so confident that you understand everything about this attribute and you are certain that this is possible to implement. What happens if in some obscure library it turns out to be impossible to get this data? Does it mean that library cannot support this set of semantic conventions at all or they will just omit this particular required attribute?

I wonder if we need to change `MUST` to `SHOULD` here to recognize the fact that the understanding of semantic convention's author is always going to be imperfect and the developer who implements the convention ultimately should be allowed to make the last call.

specification/common/attribute-requirement-level.md
lmolkova: Make sense, I use SHOULD clause now

specification/common/attribute-requirement-level.md
lmolkova: @tigrannajaryan 
I understand your concern. I'd like to explain my arguments for MUST and explore this path first.

- the specs we have now are implemented in 10 languages and we can be confident marking specific attributes as required
- we need a process that makes semantic convention owned by OTel stable and implementation of it in a few languages/libraries should be a prereq. I don't think we need 10, but a diverse set of 3-5.
- new semantic conventions can start with everything recommended or conditional and gradually make requirements stricter - it should not be breaking
- SHOULD, despite its description, at least in what I see, is not perceived strong enough
-  If a rare edge-case instrumentation that can't provide a bare-minimum level of attributes should not even be considered as following a specific convention
- consumers of telemetry may want to know what convention(s) this item represents (Azume Monitor needs to for sure). Having a MUST required attribute would help with that

Overall I agree that it makes semantic convention authors' life harder, but I believe it's a good thing that helps instrumentations, vendors, and users.

reyang: > I wonder if we need to change `MUST` to `SHOULD` here to recognize the fact that the understanding of semantic convention's author is always going to be imperfect and the developer who implements the convention ultimately should be allowed to make the last call.

+1, IMHO nothing is really `MUST` in semantic convention.

lmolkova: There are quite a few things that come with `MUST` in semantic conventions currently and I don't see any active discussions on changing it.

carlosalberto: > A rare edge-case instrumentation that can't provide a bare-minimum level of attributes should not even be considered as following a specific convention

For me this is the compromise we should be ready to have if/when such case appears. Is this safe enough? 

(And we should definitely instrument representative enough libraries in a few languages to make sure this kind of components are indeed **commonly** available)

carlosalberto (approved):

carlosalberto:
Overall LGTM - let's address @tigrannajaryan concern though. Maybe we can discuss it in today's Spec call?

specification/common/attribute-requirement-level.md
Oberon00: I think the semantic conventions could be more useful for consumers of telemetry if there are at least some MUST requirements. If some instrumentation cannot follow the MUST requirements, it does indeed not conform to the semantic conventions. That does not mean it's useless, and there is no MUST requirement for following semantic conventions. It just means that the instrumentation may have issues with some telemetry consumers.

bogdandrutu:
Summary of the SIG discussion:
1. The "conditional" level is equivalent of SHOULD;
2. The "required" level is equivalent of MUST;

Bogdan's feedback:
1. Find a better name for "conditional";
2. Almost all current (already defined) "required" semantic conventions are "conditional";
3. @lmolkova Recommended in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt) is "SHOULD" and by their definition looks like "conditional" here.
4. @tigrannajaryan per [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt) REQUIRED == MUST, I do believe that we should not use required everywhere like we do right now, but I would not make the change REQUIRED == SHOULD as suggested.

Bogdan's questions:
1. If the "summary" is correct, why not using directly the words in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt)?

tigrannajaryan:
> 3\. @lmolkova Recommended in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt) is "SHOULD" and by their definition looks like "conditional" here.

I think Conditional is different here because it has a "condition" associated with it.

specification/common/attribute-requirement-level.md
```diff
@@ -0,0 +1,45 @@
...
+
+- **Conditional**. All instrumentations SHOULD add the attribute when instrumented entity supports corresponding feature and the attribute value can be [efficiently retrieved and populated](#performance-suggestions). Semantic convention assigning `Conditional` level on an attribute, SHOULD clarify when attribute is expected to be populated.
+`http.route` is an example of a conditional attribute: is widely supported by HTTP web frameworks, but some low-level HTTP server implementations do not support it.
+_Note: For producers of telemetry `Required` and `Conditional` levels are semantically the same (under the assumption that `Required` attributes are always available). However, consumers may use this distinction to identify conventions or validate telemetry._
```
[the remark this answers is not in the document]
lmolkova: > I am not sure what the purpose of this sentence is. It seems like it can confuse more than help.

I clarified it and moved to `Required` level, I wanted to emphasize that required attributes are something backends can rely upon.

lmolkova:
> Find a better name for "conditional";

How would everyone feel about `Conditionally required`? 


@bogdandrutu 

> If the "summary" is correct, why not using directly the words in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt)?

The summary is mostly correct, but not full:
- `Required` == `MUST` is just fine
- `Conditionally required` == MUST IF \<CONDITION\>, not really a SHOULD. If such an attribute is missing, telemetry loses a lot in quality
  - `http.status_code` MUST be provided if the response was received
  - `net.host.port` MUST be set if port is not default.
  - `net.transport` for DB MUST be set for inproc databases  https://github.com/open-telemetry/opentelemetry-specification/blob/e6b292a08b4d32cff3e5d3bbedbdc7ffb0a9637c/semantic_conventions/trace/database.yaml#L231 
  - `http.route` MUST if available
  - [At least one of `process.executable.name`, `process.executable.path`, `process.command`, `process.command_line` or `process.command_args` is required](https://github.com/open-telemetry/opentelemetry-specification/blame/9faecf44d3edf6bc1433b01e7064c24f0fa5e4ff/specification/resource/semantic_conventions/process.md#L47)

    I still believe MUST works better here, but it doesn't really fit into RFC language.
    
- `Recommended` == SHOULD,  unless disabled. Telemetry is mostly fine without it (e.g. `http.user_agent` could be redundunt on internal calls)
- `Opt-in` == SHOULD IF explicitly enabled

Considering these nuances RFC levels would not be descriptive enough.

lmolkova:
I went through existing semantic conventions we have YAML for and here's what I believe this change would mean for them: 
https://github.com/lmolkova/opentelemetry-specification/commit/8f53b12788fc9a3c200ae5097db6379a6615442f - It would only change the attributes that we currently classify as `required conditionally`, but they are actually recommended.

**We have 26 `required: conditional` attributes, which are mostly used to:**
- describe `constraints.any_of` mostly for `net.peer.name/ip` substitutes
- to skip if default (port, success response, empty string, etc)
- used with if available/applicable: `db.name`, `db.statement` (not disabled), `net.transport`, `http.status_code`
- `faas.invoked_region` comes with [MUST if clause](https://github.com/open-telemetry/opentelemetry-specification/blob/3a82056c292cc649bd45e14c33e9d72882e74fe6/semantic_conventions/trace/faas.yaml#L186-L193)
- `messaging.destination_kind` - static enum

**We have 26 `required: always` attributes and I believe they are safe to keep**, the details:
- comes a lot on static things: `os.type`, `service.name` - also has [MUST in desription](  https://github.com/open-telemetry/opentelemetry-specification/blob/63da3f466323935fcc78b9c955ba75b89d9332d9/semantic_conventions/resource/service.yaml#L7-L16), `webengine_resource.name`, `db.system`, `messaging.system`, `rpc.system`
- things defined in other specs: `http.method`, `rpc.method`, `cloudevents.<many>` (required in the CloudEvent spec)
- Specific implementations that can confiently require things:  `db.mongodb.collection`, `messaging.rocketmq` has a few required fields 
- faas:
  - seems fine: `faas.name` (resource) and `fass.invoked_name` (span) - function name, `faas.time`, `faas.trigger`, `faas.invoked_provider` (cloud provider)
  - **[might need to relax requirement, but apply to special-case convention and are probably fine]** `faas.document.collection`, `faas.document.operation`, `faas.document.time`

So I believe this is pretty minor change for current conventions and we already have good classification there, just some luck of terminology.

bogdandrutu:
> Required == MUST is just fine

+1

> `Conditionally Required` == MUST IF <CONDITION>, not really a SHOULD. If such an attribute is missing, telemetry loses a lot in quality

+1 I like the name.

> Recommended == SHOULD, unless disabled. Telemetry is mostly fine without it (e.g. http.user_agent could be redundunt on internal calls)
> Opt-in == SHOULD IF explicitly enabled

The RFC defines "recommended" as "should", without any caveat like "unless disabled". Are we mixing 2 "dimensions" here: 1. for devs to make enough effort to make these available in the instrumentation; 2. the capability for users of enabling/disabling these?

I would actually consider to change "Opt-In" to "OPTIONAL", and say that instrumentation plugins must offers way to "opt-out" for recommended attributes, and to "opt-in' for "optional" attributes. the required and conditional required cannot be disabled, this way we separate the 2 concerns about what is available and what is configurable for the user.

> Considering these nuances RFC levels would not be descriptive enough.

I agree, but I would use references to RFC when explain them, since is easier to understand.

lmolkova:
@bogdandrutu thank you, I believe I addressed your feedback.

@tigrannajaryan, LMK what you think on SHOULD/MUST discussion

specification/common/attribute-requirement-level.md
```diff
@@ -0,0 +1,43 @@
...
+
+- **Conditionally Required**. All instrumentations MUST add the attribute when given condition is satisfied and the attribute value can be [efficiently retrieved and populated](#performance-suggestions). Semantic convention of a `Conditionally Required` level of an attribute MUST clarify the condition under which the attribute is expected to be populated.
+`http.route` is an example of a conditionally required attribute to be populated when instrumented HTTP framework provides route information for the instrumented request. Some low-level HTTP server implementations do not support routing and corresponding instrumentations can't populate the attribute.
+
+- **Recommended**. Instrumentations SHOULD add the attribute by default if it's readily available and can be [efficiently populated](#performance-suggestions). Instrumentation MAY offer a configuration option to disable recommended attributes.
+
+- **Optional**. Instrumentations MAY populate the attribute. Instrumentation SHOULD populate the attribute if and only if user configures instrumentation to do so. Instrumentation that doesn't support configuration MUST NOT populate **Optional** attributes.
```
tigrannajaryan, line 28: I think the very first sentence is unnecessary (unless I misunderstand the intent).
```suggestion
- **Optional**. Instrumentation SHOULD populate the attribute if and only if user configures instrumentation to do so. Instrumentation that doesn't support configuration MUST NOT populate **Optional** attributes.
```

tigrannajaryan:
Looks mostly good to me. I have one question remaining: can "Recommended" also have a condition (examples show it can)? If that is the case why do we not have "Conditionally Recommended" as a separate level? If we don't want that then why "Required" and "Conditionally Required" are separate levels?

lmolkova:
@tigrannajaryan 

> I have one question remaining: can "Recommended" also have a condition (examples show it can)? If that is the case why do we not have "Conditionally Recommended" as a separate level?

Recommended attributes come with SHOULD, which implies there could be some reasons when instrumentations would skip them. The conditions still exist and if we want to be very formal, we can introduce an additional level for it. 

If feels through it would not help much to have this distinction:
- instrumentations use their best judgment when and how condition applies and if there are other factors. 
- backends can't expect the attribute anyway and should build experiences with this in mind

The tooling might benefit from it and I added an optional description on recommended attributes, which seems to be enough.
If it proves wrong at some point, we can add one more level.

The key difference with `Required` \ `Conditionally Required` is that `Required` attribute presence is guaranteed by semantic convention authors. Backends can rely on the presence of such attributes.
With `Conditionally Required`, backends can't rely on the presence, but without them, the user experience would be subpar. Instrumentations need to do everything reasonable they can to populate them. And, I believe, we need more structure and formality to specify conditions than for `Recommended`.

specification/common/attribute-requirement-level.md
bogdandrutu: Closing this per the discussion in the main thread. @tigrannajaryan feel free to open if you feel otherwise.

bogdandrutu:
@lmolkova my last question is related to `conditionally required`. Should those be "conditional" on performance? I somehow felt that that is the "recommended" level. Do you think we can simplify and say "conditionally required" are not conditional on performance (similar to required)?

specification/common/attribute-requirement-level.md
```diff
@@ -0,0 +1,43 @@
...
+<!-- tocstop -->
+
+</details>
+
+_This section applies to Log, and Metric, Resource, and Span describes requirement levels for attributes._
+
+Following attribute requirement levels are specified:
+
+- **Required**. All instrumentations MUST populate the attribute. Semantic convention defining required attribute expects that an absolute majority instrumentation libraries and applications are able to efficiently retrieve and populate it, can ensure cardinality, security, and other requirements specific to signal defined by the convention. `http.method` is an example of a required attribute.
+_Note: Consumers of the telemetry may use `Required` attributes to identify specific semantic convention(s) the given telemetry item follows._
```
reyang, line 21: What does "identify" mean? (or maybe consider removing the note from this PR, and use a separate PR to address it later)

Oberon00: Identify that a particular semantic convention is used at all. E.g. currently you can check for the presence of the required `db.system` attribute to identify all spans that involve database operations.

tigrannajaryan: We probably need to make it clear that a set of semantic conventions is typically associated with an entity (e.g. database) and the presence of a "Required" attribute from that set (e.g. `db.system`) is how you can identify that the telemetry is about that entity.
It doesn't have to be in this PR, can be done later.

tigrannajaryan:
> The key difference with `Required` \ `Conditionally Required` is that `Required` attribute presence is guaranteed by semantic convention authors. Backends can rely on the presence of such attributes.

This is good. We need to make this more explicit in the spec. Can be done in a separate PR.

lmolkova:
> @lmolkova my last question is related to conditionally required. Should those be "conditional" on performance? I somehow felt that that is the "recommended" level. Do you think we can simplify and say "conditionally required" are not conditional on performance (similar to required)?

@bogdandrutu 

good point, there are many conditions, and the most popular is the availability of the attribute value, not perf. Specific conventions can be more specific. I removed perf from the `conditionally required` definition.

specification/common/attribute-requirement-level.md
```diff
@@ -0,0 +1,43 @@
...
+
+The requirement level for attribute is defined by semantic conventions depending on attribute availability across instrumented entities, performance, security, and other factors. When defining requirement levels, semantic conventions MUST take into account signal-specific requirements. For example, Metric attributes that may have high cardinality can only be defined with **Optional** level.
+
+Semantic convention that refers to an attribute from another semantic convention MAY modify the requirement level within its own scope. Otherwise, requirement level from referred semantic convention applies.
+For example, [Database semantic convention](../trace/semantic_conventions/database.md) references `net.transport` attribute defined in [General attributes](../trace/semantic_conventions/span-general.md) with `Conditionally Required` level on it.
+
+Instrumentations that decide not to populate `Conditionally Required` or `Recommended` attributes due to [performance](#performance-suggestions), security, privacy, or other consideration by default, SHOULD use the **Optional** requirement level on them if the attributes are logically applicable.
```
bogdandrutu, line 35: @lmolkova this also needs to be updated correct? for `Conditionally Required` the instrumentation cannot decide to not populate due to other reason than the condition correct?

bogdandrutu (approved):
Approved with one the last comment that I hope is clear :)

specification/common/attribute-requirement-level.md
lmolkova: thanks! I updated it - I believe we need to give an idea of what happens if the condition is *not* met.

bogdandrutu (approved):

specification/common/attribute-requirement-level.md
bogdandrutu: +1

specification/common/attribute-requirement-level.md
```diff
@@ -0,0 +1,46 @@
...
+
+<!-- toc -->
+
+- [Performance suggestions](#performance-suggestions)
+
+<!-- tocstop -->
+
+</details>
+
+_This section applies to Log, and Metric, Resource, and Span describes requirement levels for attributes._
```
tigrannajaryan, line 16:
```suggestion
_This section applies to Log, Metric, Resource, and Span, and describes requirement levels for attributes._
```

specification/common/attribute-requirement-level.md
```diff
@@ -0,0 +1,46 @@
...
+
+Following attribute requirement levels are specified:
```
tigrannajaryan, line 18:
```suggestion
The following attribute requirement levels are specified:
```

specification/common/attribute-requirement-level.md
```diff
@@ -0,0 +1,46 @@
...
+
+- **Required**. All instrumentations MUST populate the attribute. Semantic convention defining required attribute expects that an absolute majority instrumentation libraries and applications are able to efficiently retrieve and populate it, can ensure cardinality, security, and other requirements specific to signal defined by the convention. `http.method` is an example of a required attribute.
```
tigrannajaryan, line 20:
```suggestion
- **Required**. All instrumentations MUST populate the attribute. Semantic convention defining a Required attribute expects that an absolute majority of instrumentation libraries and applications are able to efficiently retrieve and populate it, can ensure cardinality, security, and other requirements specific to signal defined by the convention. `http.method` is an example of a Required attribute.
```

specification/common/attribute-requirement-level.md
```diff
@@ -0,0 +1,46 @@
...
+_Note: Consumers of telemetry can detect if telemetry item follows specific semantic convention by checking presence of a `Required` attribute defined by such convention. For example, `db.system` attribute on a span can be used as an indication that span follows database semantics._
```
tigrannajaryan, line 21:
```suggestion
_Note: Consumers of telemetry can detect if telemetry item follows a specific semantic convention by checking the presence of a `Required` attribute defined by such convention. For example, the presence of `db.system` attribute on a span can be used as an indication that the span follows database semantics._
```

specification/common/attribute-requirement-level.md
```diff
@@ -0,0 +1,46 @@
...
+
+- **Conditionally Required**. All instrumentations MUST add the attribute when given condition is satisfied. Semantic convention of a `Conditionally Required` level of an attribute MUST clarify the condition under which the attribute is expected to be populated.
+`http.route` is an example of a conditionally required attribute to be populated when instrumented HTTP framework provides route information for the instrumented request. Some low-level HTTP server implementations do not support routing and corresponding instrumentations can't populate the attribute.
+
+- **Recommended**. Instrumentations SHOULD add the attribute by default if it's readily available and can be [efficiently populated](#performance-suggestions). Instrumentation MAY offer a configuration option to disable recommended attributes.
```
tigrannajaryan, line 26:
```suggestion
- **Recommended**. Instrumentations SHOULD add the attribute by default if it's readily available and can be [efficiently populated](#performance-suggestions). Instrumentations MAY offer a configuration option to disable Recommended attributes.
```

specification/common/attribute-requirement-level.md
```diff
@@ -0,0 +1,46 @@
...
+
+- **Optional**. Instrumentation SHOULD populate the attribute if and only if user configures instrumentation to do so. Instrumentation that doesn't support configuration MUST NOT populate **Optional** attributes.
```
tigrannajaryan, line 28:
```suggestion
- **Optional**. Instrumentations SHOULD populate the attribute if and only if the user configures the instrumentation to do so. Instrumentation that doesn't support configuration MUST NOT populate **Optional** attributes.
```

specification/common/attribute-requirement-level.md
```diff
@@ -0,0 +1,46 @@
...
+
+The requirement level for attribute is defined by semantic conventions depending on attribute availability across instrumented entities, performance, security, and other factors. When defining requirement levels, semantic conventions MUST take into account signal-specific requirements. For example, Metric attributes that may have high cardinality can only be defined with **Optional** level.
+
+Semantic convention that refers to an attribute from another semantic convention MAY modify the requirement level within its own scope. Otherwise, requirement level from referred semantic convention applies.
```
tigrannajaryan, line 32:
```suggestion
Semantic convention that refers to an attribute from another semantic convention MAY modify the requirement level within its own scope. Otherwise, requirement level from the referred semantic convention applies.
```

specification/common/attribute-requirement-level.md
```diff
@@ -0,0 +1,46 @@
...
+For example, [Database semantic convention](../trace/semantic_conventions/database.md) references `net.transport` attribute defined in [General attributes](../trace/semantic_conventions/span-general.md) with `Conditionally Required` level on it.
+
+When condition on `Conditionally Required` attribute is not satisfied and there is no requirement to populate attribute, semantic conventions MAY provide special instructions on how to handle it. If no instructions are given and if instrumentation can populate the attribute, instrumentation SHOULD use the **Optional** requirement level on the attribute.
```
tigrannajaryan, line 35:
```suggestion
When the condition on `Conditionally Required` attribute is not satisfied and there is no requirement to populate attribute, semantic conventions MAY provide special instructions on how to handle it. If no instructions are given and if instrumentation can populate the attribute, instrumentation SHOULD use the **Optional** requirement level on the attribute.
```

specification/common/attribute-requirement-level.md
```diff
@@ -0,0 +1,46 @@
...
+For example, `net.peer.name` is `Conditionally Required` by [Database convention](../trace/semantic_conventions/database.md) when available. When only `net.peer.ip` is available,  instrumentation can do DNS lookup, cache and populate `net.peer.name` but only if user explicitly enables instrumentation to do so considering performance issues DNS lookup introduces.
```
tigrannajaryan, line 36:
```suggestion
For example, `net.peer.name` is `Conditionally Required` by [Database convention](../trace/semantic_conventions/database.md) when available. When only `net.peer.ip` is available,  instrumentation can do a DNS lookup, cache and populate `net.peer.name` but only if user explicitly enables instrumentation to do so considering performance issues the DNS lookup introduces.
```

tigrannajaryan (approved):
LGTM, great job.
I left some minor spelling comments.

specification/common/attribute-requirement-level.md
```diff
@@ -0,0 +1,46 @@
+# Attribute Requirement Level
+
+**Status**: [Experimental](../document-status.md)
+
+<details>
+<summary>Table of Contents</summary>
+
+<!-- toc -->
+
+- [Performance suggestions](#performance-suggestions)
```
Oberon00, line 10: Editorial: The toc looks strange. Maybe add (a) subheading(s) for the main section?

specification/common/attribute-requirement-level.md
```diff
@@ -0,0 +1,46 @@
+# Attribute Requirement Level
```
Oberon00, line 1:
```suggestion
# Attribute Requirement Levels for Semantic Conventions
```

specification/common/attribute-requirement-level.md
```diff
@@ -0,0 +1,46 @@
...
+
+<!-- toc -->
+
+- [Performance suggestions](#performance-suggestions)
+
+<!-- tocstop -->
+
+</details>
+
+_This section applies to Log, Metric, Resource, and Span, and describes requirement levels for attributes._
```
Oberon00, line 16:
```suggestion
_This section applies to Log, Metric, Resource, and Span, and describes requirement levels for attributes defined in semantic conventions._
```

specification/common/attribute-requirement-level.md
```diff
@@ -0,0 +1,46 @@
...
+
+Semantic convention that refers to an attribute from another semantic convention MAY modify the requirement level within its own scope. Otherwise, requirement level from the referred semantic convention applies.
+For example, [Database semantic convention](../trace/semantic_conventions/database.md) references `net.transport` attribute defined in [General attributes](../trace/semantic_conventions/span-general.md) with `Conditionally Required` level on it.
+
+When the condition on `Conditionally Required` attribute is not satisfied and there is no requirement to populate attribute, semantic conventions MAY provide special instructions on how to handle it. If no instructions are given and if instrumentation can populate the attribute, instrumentation SHOULD use the **Optional** requirement level on the attribute.
+For example, `net.peer.name` is `Conditionally Required` by [Database convention](../trace/semantic_conventions/database.md) when available. When only `net.peer.ip` is available,  instrumentation can do a DNS lookup, cache and populate `net.peer.name` but only if user explicitly enables instrumentation to do so considering performance issues the DNS lookup introduces.
```
Oberon00, line 36:
```suggestion
For example, `net.peer.name` is `Conditionally Required` by [Database convention](../trace/semantic_conventions/database.md) when available. When only `net.peer.ip` is available,  instrumentation can do a DNS lookup, cache and populate `net.peer.name` but only if user explicitly enables instrumentation to do so, considering performance issues the DNS lookup introduces.
```

Oberon00 (approved):

specification/common/attribute-requirement-level.md
lmolkova: Good point, it'd also help us to have anchor links to levels. I did it and moved things around. I did not change any text.

trask (approved):

pyohannes (approved):

tigrannajaryan:
@lmolkova please resolve the conflicts.

lmolkova:
@tigrannajaryan done, thank you!

jamesmoessis (approved):
This is super valuable work, thankyou for getting this through!

PR #2524, opened 2022-05-05
Add semantic conventions for nginx metrics

nslaughter:
## Changes

- Add nginx metrics semantic conventions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,8 +54,7 @@ release.
 - Add database connection pool metrics semantic conventions
   ([#2273](https://github.com/open-telemetry/opentelemetry-specification/pull/2273)).
 - Add nginx metrics semantic conventions
-  ([#TODO](https://github.com/open-telemetry/opentelemetry-specification/pull/TODO)).
-  TODO(@nslaughter) add PR # here when you have it
+  ([#2524](https://github.com/open-telemetry/opentelemetry-specification/pull/2524)).
 
 ### Logs
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ release.
   ([#2495](https://github.com/open-telemetry/opentelemetry-specification/pull/2495)).
 - Add database connection pool metrics semantic conventions
   ([#2273](https://github.com/open-telemetry/opentelemetry-specification/pull/2273)).
+- Add nginx metrics semantic conventions
+  ([#TODO](https://github.com/open-telemetry/opentelemetry-specification/pull/TODO)).
+  TODO(@nslaughter) add PR # here when you have it
 
 ### Logs
 
--- a/specification/metrics/semantic_conventions/nginx-metrics.md
+++ b/specification/metrics/semantic_conventions/nginx-metrics.md
@@ -17,5 +17,5 @@ This document defines how to apply semantic conventions when instrumenting NGINX
 | nginx.requests              | Counter       | Int64      | requests    | `{requests}`                              | The total number of requests made to the server since it started. | | |
 | nginx.connections_accepted  | Counter       | Int64      | connections | `{connections}`                           | The total number of accepted client connections.   | | |
 | nginx.connections_handled   | Counter       | Int64      | connections | `{connections}`                           | The total number of handled connections. Generally, the parameter value is the same as nginx.connections.accepted unless some resource limits have been reached (for example, the worker.connections limit) | | |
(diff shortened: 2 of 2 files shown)


carlosalberto:
So the `nginx.requests` can easily be abstracted as `proxy.requests` or similar.

About the other ones, I'm not entirely sure it's possible to _easily_ abstract them - is there a clear, common distinction between what is considered an accepted/handled/current connection among different proxies/servers? Maybe there is, but maybe not. In that case, do we _try_ to 'unify't them? Not really sure that's a way I'd go.

carlosalberto:
So the `nginx.requests` can easily be abstracted as `proxy.requests` or similar.

About the other ones, I'm not entirely sure it's possible to _easily_ abstract them - is there a clear, common distinction between what is considered an accepted/handled/current connection among different proxies/servers? Maybe there is, but maybe not. In that case, do we _try_ to 'unify't them? Not really sure that's a way I'd go.

nslaughter:
Thank you for the feedback @MrAlias and @carlosalberto. I've taken a closer look and I'd like to get more specific feedback regarding the approach. The questions are at the end.

## What metrics are commonly supported by proxies?

I looked at stats for HAProxy and Envoy to get a broader idea of commonly computed metrics as a reference for what OTEL users may require. 

Here's a table of how like metrics are provided...

| nginx                 | envoy                         | haproxy                              |
|-----------------------|-------------------------------|--------------------------------------|
| .requests             | .http.downstream_rq_completed | .Sessions.Frontend.Cum.HTTP_requests |
| .connections_accepted | .http.downstream_cx_total     | .Sessions.Frontend.Cum.Connections   |
| .connections_handled  | ?                             | ?                                    |
| .connections_current  | .http.downstream_cx_active    | .Backend.Cur.Connections             |

The breakdown of current connections by state is apparently not supported by Envoy or HAProxy. I also noted that the equivalent of .connections_handled could be calculated from stats provided by the other two.

## How could we make the semantic conventions more useful and general?

After reviewing the feedback I see that `nginx.requests` could be `otel.http.server.active_requests`. And `nginx.connections_current` could be `otel.system.network.connections`. That makes sense to me assuming it's appropriate to use system metrics for services.

Does the commonality of `.connections_accepted` as a cumulative metric suggest that we should also support that in the conventions?

carlosalberto:
Hey @nslaughter - thanks for the update!

> The breakdown of current connections by state is apparently not supported by Envoy or HAProxy

IMHO we should not try too hard to unify metrics - instead, add metrics for the most common cases, and let specific components add their very-specific ones. In this case we could perhaps define total/active connections as common metrics, and have nginx define an additional one for its `.connections_handled`?

(Btw, a general metric for requests is fine - probably a new one, as the existing one is for **active** requests, whereas envoy's metric seems to be for **completed** ones)

Opinion @MrAlias ?

nslaughter:
Thanks for the help, @tedsuo.

Regarding... 
> * `nginx.connections_current` should map to `http.server.active_requests` not `system.network.connections`

Mapping something called `connections` to something called `requests` is confusing to me. My understanding is that they would map under the original HTTP model, but with persistent connections, pipelining, and concurrency beyond HTTP 1.1 we can only expect the `active_connections <= active_requests`. That said, I'm neither expert in networking nor these conventions.

> * small fyi, we don't currently use `otel` as a namespace prefix.

Thanks for pointing that out. I introduced that confusion hoping to clarify the distinction when side-by-side with something called `nginx` otherwise that's the namespace I would have expected as it is in collector-contrib. Will fix it.

> One larger point: NGINX is currently adding native integration of OTel. I will notify them and make sure we get their input; we should avoid a situation where we define something in the spec which doesn't match what they are actually building. :)

Thanks for the awareness and comms on this. Much appreciated.

carlosalberto:
@tedsuo Any reason to not go with `proxy.*` instead? Just trying to be on the conservative side (and avoid collusions 😄 ) 
 
 Not a string opinion, just considering our options.

gramidt:
Hi All - 

This PR was just brought to my attention, and I wanted to let you know that it's being circulated to the corresponding NGINX teams within F5.

carlosalberto:
Thanks @gramidt !

Other than that part of the review, circling back, I'd like to propose:

1. Metrics are dead obvious (e.g. requests) exist under `proxy.*`, e.g. `proxy.requests`.
2. Metrics for operations that are NOT dead obvious go under `proxy.<component>.metric`, e.g. `proxy.nginx.connections.handled`.
3. We _eventually_ merge some of the ones in 2 as common metrics.

The reason for this is that trying to _audit_ all new metrics may severely slow us down overall - and as they say, "perfect is the enemy of good" (or done).

nslaughter:
Closing this for the time... Will revisit this with a proposal reflecting some feedback so concensus can coalesce -> then move from towards a PR.

PR #2526, opened 2022-05-06
Changes to PHP values

yuktea:
- Associate Tracer with InstrumentationScope
- Default values for `service.name`


brettmc (approved):
Maintainer approval

PR #2529, opened 2022-05-10
Add feature flagging semantic conventions

beeme1mr:
Resolves https://github.com/open-telemetry/opentelemetry-specification/issues/2532

## Changes

Added semantic conventions for representing feature flags as spans. This can be used to determine the impact a feature has on a request.

## Related issues
- https://github.com/open-telemetry/opentelemetry-specification/issues/2532
- https://github.com/open-feature/spec/issues/30


tigrannajaryan:
This needs an issue in this repository (a reference to an issue in another repository is useful but not sufficient) and then follow the [issue workflow](https://github.com/open-telemetry/opentelemetry-specification/blob/main/CONTRIBUTING.md#issue-triaging). Ideally this should have happened before the PR is created so that we know we want it in the OpenTelemetry.

tigrannajaryan:
> This needs an issue in this repository (a reference to an issue in another repository is useful but not sufficient) and then follow the [issue workflow](https://github.com/open-telemetry/opentelemetry-specification/blob/main/CONTRIBUTING.md#issue-triaging). Ideally this should have happened before the PR is created so that we know we want it in the OpenTelemetry.

You can wait a bit and see if there are approvers commenting on this PR who think the PR is a good idea then we can move forward without an issue.

beeme1mr:
I just noticed that you clarified this in [a PR](https://github.com/open-telemetry/opentelemetry-specification/pull/2530). I'll open an issue tomorrow. Thanks!

reyang:
May I get a short summary regarding why feature-flagging needs to be part of semantic conventions? Is there an example (I read through the linked [issue](https://github.com/open-feature/spec/issues/30) and couldn't find a clear explanation)?

dyladan:
> May I get a short summary regarding why feature-flagging needs to be part of semantic conventions? Is there an example (I read through the linked [issue](https://github.com/open-feature/spec/issues/30) and couldn't find a clear explanation)?

If feature flag keys and values are consistently reported, it will enable splitting traces and metrics by these values for usecases like A/B testing, feature rollouts, and similar. In order for this to work it must be consistent across all microservices, potentially in multiple languages and/or using multiple feature flag services.

dyladan:
> > If feature flag keys and values are consistently reported, it will enable splitting traces and metrics by these values for usecases like A/B testing, feature rollouts, and similar. In order for this to work it must be consistent across all microservices, potentially in multiple languages and/or using multiple feature flag services.
> 
> If the user has System X calling into Cloud Provider Y which in turn calls back to System Z, and they use different A/B testing / feature flag system, what would be the expectation? Or this PR is specific about `open-features` and it's compatible systems, not about other systems?

They would just report span attributes consistent with the feature flag system they're using. The attributes are general enough to be used with any feature flag system. I'm not sure I see what the problem is.

dyladan:
This is not stale

beeme1mr:
Marking this as a draft until #2532 is resolved.

beeme1mr:
Still waiting for an official response on #2532

tigrannajaryan:
I removed the Stale label to give more time for people to express their support at https://github.com/open-telemetry/opentelemetry-specification/issues/2532

semantic_conventions/trace/feature-flag.yaml
```diff
@@ -0,0 +1,49 @@
...
+    prefix: feature_flag.provider
+    extends: feature_flag
+    brief: Semantic convention for the provider performing the flag evaluation.
+    attributes:
+      - id: name
+        type: string
+        required: always
+        brief: The name of the provider performing the flag evaluation.
+        examples: ["Flag Manager"]
+      - id: management_url
```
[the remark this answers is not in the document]
joaopgrassi: Also: What is the use of such URL? Would it be relevant to know this URL in each evaluated flag? I feel this is something the user of the flag provider should know already? Isn't a management portal the same for all flags a customer have?

semantic_conventions/trace/feature-flag.yaml
```diff
@@ -0,0 +1,49 @@
...
+    attributes:
+      - id: variant
+        type: string
+        brief: The name associated with the evaluated value.
+        examples: ["reverse"]
+        required:
+          conditional: >
+            A variant should be used if it is available. If the variant is
+            present, feature_flag.evaluated.value should be omitted.
+      - id: value
```
joaopgrassi, line 41: Are there any privacy/security risks in adding the value as a span attribute? I'm not an expert in this space so take it with a pinch of salt.

semantic_conventions/trace/feature-flag.yaml
```diff
@@ -0,0 +1,49 @@
...
+        brief: The URL used to manage the feature flag in the provider.
+        examples: ["http://localhost:4200/flags/1"]
+  - id: feature_flag.evaluated
+    prefix: feature_flag.evaluated
+    extends: feature_flag
+    brief: Semantic convection for the values evaluated by provider.
+    attributes:
+      - id: variant
+        type: string
+        brief: The name associated with the evaluated value.
```
joaopgrassi, line 35: I'm not entirely sure if I understand what this attribute means. Is it something like an "alias" for the evaluated value? Could you maybe point me to any documentation or more examples of how this would look like in reality?

specification/trace/semantic_conventions/feature-flags.md
```diff
@@ -0,0 +1,53 @@
...
+
+- [General Attributes](#general-attributes)
+- [Provider Attributes](#provider-attributes)
+- [Evaluated Attributes](#evaluated-attributes)
+
+<!-- tocstop -->
+
+**Span kind:** MUST always be `INTERNAL`.
+
+Span `name` should be a combination of the provider name and the name of the function used when evaluating the flag, separated by a hyphen. For example, `unleash - isEnabled` or `my flag eval - getBoolValue`.
```
joaopgrassi, line 17: What about instead of the provider name, using `OpenFeature` in the span name? If we have the provider in an attribute already, I wonder what is best 🤔

joaopgrassi (changes requested):

semantic_conventions/trace/feature-flag.yaml
```diff
@@ -0,0 +1,49 @@
...
+        brief: The unique identifier of the feature flag.
+        examples: ["show-new-logo"]
+  - id: feature_flag.provider
+    prefix: feature_flag.provider
+    extends: feature_flag
+    brief: Semantic convention for the provider performing the flag evaluation.
+    attributes:
+      - id: name
+        type: string
+        required: always
```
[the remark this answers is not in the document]
beeme1mr: Sure, that makes sense to me. I'll make this change.

semantic_conventions/trace/feature-flag.yaml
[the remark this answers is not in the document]
beeme1mr: The original intent was to provide quick access to view/modify a feature flag directly from a trace. However, this seems unprecedented in the semantic conventions. I'll remove this attribute for now.

semantic_conventions/trace/feature-flag.yaml
beeme1mr: You could think of it as an alias. This is partically useful when the value is large (i.e. object, complete sentence or paragraph) or the meaning may be unclear (i.e. object, numeric values).

Please see the OpenFeature glossary for more information: https://github.com/open-feature/spec/blob/main/specification/glossary.md#variant

specification/trace/semantic_conventions/feature-flags.md
beeme1mr: I'm open to ideas for the span name but I would prefer to avoid using `OpenFeature`. Currently, the proposal is vendor and tool agnostic.

semantic_conventions/trace/feature-flag.yaml
beeme1mr: A feature flag value should never contain sensitive information. However, the administrator of a feature flag system would need to ensure this doesn't happen.

semantic_conventions/trace/feature-flag.yaml
toddbaert: @joaopgrassi 

The example @beeme1mr added in the convention is a good one. Many people might not intuitively understand CSS hex color values (ie: `#CC0000` is red). Reporting the variant name when possible is helpful in this regard.

This idea is not unprecedented, many vendors use the same or similar concepts: https://docs.launchdarkly.com/home/flags/variations, https://docs.getunleash.io/advanced/toggle_variants

dyladan: @toddbaert the example was different when joao originally made the comment. I think the current example is more obvious than the old one

toddbaert: @dyladan ah, thanks for pointing that out - I was away on Friday, catching up now :sweat_smile:

specification/trace/semantic_conventions/feature-flags.md
joaopgrassi: Yeah sure, now that I think better you are probably right. The important thing with the span name is it being of low cardinality. I imagine an app sticks with a provider, so in that context the span name will be of low cardinality, correct?

beeme1mr: I've updated the proposal for the span name to make it more useful while maintaining low cardinality.

semantic_conventions/trace/feature-flag.yaml
joaopgrassi: Ah yeah the hex color examples now is perfect. You can resolve this. @beeme1mr

toddbaert:
@beeme1mr 

One thing that may be worth adding is a `evaluation_reason` or something similar. This would identify _why_ a particular flag value was returned (some targeting rule was matched, a random split is configured, or a static flag value was assigned, for instance):

```
      - id: evaluation_reason
        type: string
        requirement_level: recommended
        brief: A string indicating why a particular variant was selected.
        examples: ["STATIC", "TARGETING_MATCH". "PSEUDORANDOM_SPLIT"]
```
Such a concept exists in [OpenFeature](https://github.com/open-feature/spec/blob/main/specification/flag-evaluation.md#requirement-146) as well as some vendors, for example: https://docs.launchdarkly.com/sdk/concepts/evaluation-reasons

dyladan:
> @beeme1mr
> 
> One thing that may be worth adding is a `evaluation_reason` or something similar. This would identify _why_ a particular flag value was returned (some targeting rule was matched, a random split is configured, or a static flag value was assigned, for instance):
> 
> ```
>       - id: evaluation_reason
>         type: string
>         requirement_level: recommended
>         brief: A string indicating why a particular variant was selected.
>         examples: ["STATIC", "TARGETING_MATCH". "PSEUDORANDOM_SPLIT"]
> ```
> 
> Such a concept exists in [OpenFeature](https://github.com/open-feature/spec/blob/main/specification/flag-evaluation.md?rgh-link-date=2022-07-11T20%3A07%3A14Z#requirement-146) as well as some vendors, for example: [docs.launchdarkly.com/sdk/concepts/evaluation-reasons](https://docs.launchdarkly.com/sdk/concepts/evaluation-reasons)

I see a couple problems with this (although none that can't be overcome).

1. I can't see what obvious value that provides. Knowing that a flag is `true` because it targets specific users is possible from just the key and the data stored in your feature flag system.
2. I think this might be tough to implement in a way that makes sense across all tools/vendors. In order for the reason to be meaningful, each reason would have to have a definition and all vendors would have to agree on them. There is also a lot of potential crossover between different reasons. For example, if I split 50% of my traffic based on the hash of the user's email, is that `TARGETING_MATCH` or is that `PSEUDORANDOM_SPLIT`?

I think in this case I would advocate to get the basic functionality already in the PR merged, then see what users ask for when they start actually consuming telemetry data.

toddbaert:
> 1. I can't see what obvious value that provides. Knowing that a flag is `true` because it targets specific users is possible from just the key and the data stored in your feature flag system.

This is true, but it might be tedious to cross reference such configuration when viewing telemetry data. The `reason` can help make it obvious that some specific targeting occurred "at a glance", perhaps in a situation when a static value was expected (maybe somebody forgot to remove some specific targeting that was added for debug purposes).

Vendors thought this information was important enough to include in their API, so I think it could be important for the semantic conventions.

> 2. I think this might be tough to implement in a way that makes sense across all tools/vendors. In order for the reason to be meaningful, each reason would have to have a definition and all vendors would have to agree on them. There is also a lot of potential crossover between different reasons. For example, if I split 50% of my traffic based on the hash of the user's email, is that `TARGETING_MATCH` or is that `PSEUDORANDOM_SPLIT`?

I'm not sure that there needs to be a precise consensus on the possible values of this field across all providers for it to be valuable, but I see your point. Perhaps a like of consistency here is a dealbreaker :shrug:

dyladan:
> > 1. I can't see what obvious value that provides. Knowing that a flag is `true` because it targets specific users is possible from just the key and the data stored in your feature flag system.
> 
> This is true, but it might be tedious to cross reference such configuration when viewing telemetry data. The `reason` can help make it obvious that some specific targeting occurred "at a glance", perhaps in a situation when a static value was expected (maybe somebody forgot to remove some specific targeting that was added for debug purposes).

¯\_(ツ)_/¯ the value there seems like a stretch to me, but maybe I'm wrong.

> Vendors thought this information was important enough to include it in their API, so I think it could be important for the semantic conventions.

This is definitely not reason enough on its own. There are all sorts of things available from host/process/library APIs that we don't include in semconv.

Like I said. I think all of these things can be overcome, but it is a non-breaking addition and can always be included later if it becomes important.

semantic_conventions/trace/feature-flag.yaml
joaopgrassi: Fair enough. Feel free to resolve this.

dyladan (approved):

specification/trace/semantic_conventions/feature-flags.md
```diff
@@ -0,0 +1,22 @@
+# Semantic conventions for Feature Flags
+
+**Status**: [Experimental](../../document-status.md)
+
+**Span kind:** MUST always be `INTERNAL`.
+
+The **span name** MUST be of the format `<feature_flag.provider_name> <feature_flag.flag_key>` provided that
```
joaopgrassi, line 7:
```suggestion
The **span name** SHOULD be of the format `<feature_flag.provider_name> <feature_flag.flag_key>` provided that
```
I feel this needs to be a SHOULD, since it says: if the `provider_name` is available. If provider name is not readily available then it can't fulfill the must requirement here right?

specification/trace/semantic_conventions/feature-flags.md
```diff
@@ -0,0 +1,22 @@
+# Semantic conventions for Feature Flags
+
```
joaopgrassi, line 2: I feel like there's something missing here. Maybe some sort of "introduction" to these conventions. Perhaps some small paragraph stating why this is useful, or in which scenarios this would be helpful? 

You can look at some of these examples maybe for inspiration: 
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions
/faas.md
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/process-metrics.md

Maybe even the `brief` you have in the `yml` is enough?

> This semantic convention defines the attributes used to represent a feature flag evaluation as a span.

>Added semantic conventions for representing feature flags as spans. This can be used to determine the impact a feature has on a request.

joaopgrassi (approved):
Just small nits, but overall this looks good!

specification/trace/semantic_conventions/feature-flags.md
beeme1mr: Done! Let me know if you think a more detailed description is required.

joaopgrassi: Looks good! Thanks for adding it.

beeme1mr:
Hi @tigrannajaryan and @SergeyKanzhelev, do you have any additional questions or concerns I can address?

specification/trace/semantic_conventions/feature-flags.md
```diff
@@ -0,0 +1,25 @@
...
+**Status**: [Experimental](../../document-status.md)
+
+This document describes how a feature flag evaluation can be represented as a span in
```
tigrannajaryan, line 5: I am not sure I understand the proposal. I would expect a feature flag and its value to be recorded as an attribute on all spans that are affected by the feature. 
However, it seems like the proposal is to record the feature flag value as a separate span. Can you explain how this is intended to be used?

beeme1mr: The current proposal is to add a dedicated span for each feature flag evaluation. This was done because multiple feature flags can be evaluated during a single operation. Also, the evaluations themselves can have a impact on performance. Having a dedicated span solves both of those issues.

Is it possible to add attributes to every affected span? If it is, would it be possible to address the challenges outlined above?

tigrannajaryan: >Is it possible to add attributes to every affected span?

I think it should be possible using a [SpanProcessor](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#span-processor), but it will apply to all spans emitted via the particular Tracer/TracerProvider (I don't know if it is desirable).

>The current proposal is to add a dedicated span for each feature flag evaluation. This was done because multiple feature flags can be evaluated during a single operation. 

If there are multiple feature flags then perhaps use a pair of attributes (one for flag names, one for values) with collinear arrays as values or one attribute per feature flag, e.g. `feature_flag.<flag_name>`.

>Also, the evaluations themselves can have a impact on performance. Having a dedicated span solves both of those issues.

Do you mean fetching the value of the feature flag can be slow and thus doing it once is preferable? I would expect any feature flag library worth its salt to be close to zero-cost for fetching the flag values since that's what I need in the regular code anyway.

Having a special span just for feature flags looks weird to me. Spans are for representing operations. What operation does this special span represent? What would be the duration of the span? I assume 0, in that case it is not a span, it is an event.

This approach is using a span as a storage for metadata about other operations, which is not what a span is supposed to be. Arguably a log record would be better suited for this, since log records at least are by design generic data structures that hold information about arbitrary events (and a "feature flags state" can be a valid event to record).

beeme1mr: > This approach is using a span as a storage for metadata about other operations, which is not what a span is supposed to be.

Flag evaluation values can be modified during runtime (e.g. a feature flag can be updated to release a new feature, decoupled from a code deployment) which, in my opinion, makes a feature flag evaluation an operation.
The operation just happens to typically run in-process, but conceptually it's similar to a call to a DB or an external API. 

Most feature flagging tools perform flag evaluation based on incoming context and a remotely configured ruleset. This typically happens in-process or over a local socket. It should execute quickly but the span duration is not 0.

While not defined in the spec, there are official OTel sensors that create spans for in-process code execution. For example, the [express sensor](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts) creates middleware spans and the [nestjs sensor](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-nestjs-core/src/instrumentation.ts) creates controller spans. I believe that feature flags could be handled in a similar manner.

dyladan: A span processor is not the right tool for this since the flag evaluation happens in the context of a request and may vary transaction to transaction. Colinear arrays may work but would require backend support in order to query making it annoying or impossible to use with existing backends. Putting the flag name in the key might work but only if the flag is guaranteed to be called a single time. If it is called multiple times the value would be overwritten

plantfansam: > Is it possible to add attributes to every affected span? If it is, would it be possible to address the challenges outlined above?

This sounds extremely useful to me (and could be enabled by some flavor of https://github.com/open-telemetry/oteps/pull/207), but I worry about how many attributes might be generated. Any given code path could have an arbitrary number of feature flags on it, and we might end up hitting attribute count limits (and generating a bunch of redundant data). Trace viewers might get overwhelmed by the sheer volume of attributes.

One alternative might be to leave the feature flag span open until whatever conditional it enables has concluded. So we could represent this in Ruby as follows:

```ruby
Tracer.in_span("evaluate my cool feature span", 
  attributes: { 
    "feature_flag.evaluated_variant" => FeatureFlagService.lookup(:my_cool_feature),
    more_semconv: "goes here" 
   }) do |span|
     # Do stuff that generates more spans, which are child spans of "evaluate my cool feature span"
end
```

This would allow trace viewers to see the logical impact of a feature flag decision, and could also enable feature-flag aware analysis of raw span data.

It's possible that this is already what's being proposed, in which case I apologize for misunderstanding!

tigrannajaryan:
Removing the Stale label. It would be good to see the questions [I raised here ](https://github.com/open-telemetry/opentelemetry-specification/pull/2529#discussion_r926256652) answered by this PR.

specification/trace/semantic_conventions/feature-flags.md
beeme1mr: > Is it supposed to be a single span per trace?

No, the expectation would be that it's a span per flag evaluation. There could be many of them on a single trace.

> What happens if the flag changes during the trace?

The value at the time of flag evaluation would be captured as an attribute on the span. This would be similar to be a database query.

> Do you get multiple spans with different values?

Yes, if there are multiple flag evaluations using different flag keys and/or evaluation contexts.

> How do you know what was the value of the flag during the execution of other (regular spans)?

I'm not sure I understand the question. Could you please elaborate?

> Is feature flag span supposed to be a parent of a subtree to which it applies?

Currently, this is not the case and I'm not sure it would be possible with existing feature flagging tools. A typical flag evaluation may look similar to this:

```typescript
const isEnabled = ffClient.getBooleanValue("new-feature", false);

if (isEnabled}
  myNewFeature();
} else {
  myOldFeature();
}
```

The current proposal would create a span for the `getBooleanValue` method (referencing the example above). However, as @tigrannajaryan mentioned, this information may be better represented as an event. It looks like there's [an open PR](https://github.com/open-telemetry/opentelemetry-specification/pull/2676) to add an events API to the spec. Perhaps it would be better to update the proposal to use events instead of spans.

tigrannajaryan: The proposed approach, where feature flag values are represented as dedicated Spans create at the time of flag evaluation seems not very useful IMO. 

I would imagine when I am analyzing a trace I would want to know the value of a feature flag for the Spans which represent the actual work performed. For example a Span may be emitted by `myNewFeature()` or by `myOldFeature()` in the example above and it would be very useful to know what is the value of "new-feature" flag when that Span was emitted.

With this proposal unless it is required/guaranteed that Spans that are executed in the context of the selected feature flag are parented by the Span which represents the feature flag evaluation it seems impossible to know what the value of the flag is for the Span that I care the most about.

Maybe there are use-cases which I am not aware of where this is highly valuable. It would be useful to describe such use cases.

dyladan: @tigrannajaryan seems pretty simple to me to filter traces by spans which contain an attribute with a particular value. Having the feature flag evaluation span be a parent of other spans doesn't really make sense to me because it implies that the feature flag only affects its children which may not be the case. Feature flags often would affect sibling or even parent span timings and behavior. All @beeme1mr really needs is an event with an attribute that he can filter entire traces by, and since we don't have an events API the best way to do that is a short or zero length span. It is modeled as a span because the feature flag evaluation may take some (very small) amount of time, but the timing of the feature flag evaluation itself is really not the important information conveyed.

dyladan: In short: you can't have affected spans be children because there is no good way to know which spans are affected. IMO an event would be a better model for this, but as that's not available yet I think this is a pragmatic solution.

scheler: For feature flags implemented in browser apps or mobile apps, this is best represented as a Session level attribute, in which case there is a [proposal](https://github.com/open-telemetry/oteps/pull/208) to send them in Resource as ephemeral attributes. 

> All @beeme1mr really needs is an event with an attribute that he can filter entire traces by, and since we don't have an events API the best way to do that is a short or zero length span

There's no good way today to represent Trace level properties - do we want it to be an Event or 0-length Span?  
The same feature flag will be effective in multiple services, so if it is one span per feature flag per trace, there's a question of which service will be creating it or semantics when multiple services create spans for the same feature flag? Also, there could be additional considerations for the feature flag span when it gets lost or not sampled.

toddbaert: > For feature flags implemented in browser apps or mobile apps, this is best represented as a Session level attribute, in which case there is a https://github.com/open-telemetry/oteps/pull/208 to send them in Resource as ephemeral attributes.

I agree with this, and I think it will work for server side in some cases as well. Most flag evaluations take place in the context of a subject (an end user, for instance). Some feature-flag SDKs even go so far as to require a "user" object with each evaluation, and can take into account request-scoped data to be used in dynamic flag evaluation. In these cases, session level attributes could work well.

weyert: Do events supported timings? I think it's really beneficial how much time is spent in a trace on feature flag evaluation. In my current project I can spend quite a bit of time on evaluating feature flag like 1/3 when handling a route/API endpoint.

beeme1mr: I've put together a sample app that shows how the proposal would look on a trace. This particular trace performed two flag evaluations. They're the last two spans on the trace.

image

I've also included a [Zipkin trace](https://github.com/open-telemetry/opentelemetry-specification/files/9363064/zipkin-trace.zip) so you can explore the data yourself.

tigrannajaryan:
I am still not convinced the feature-flag data in the proposed form is useful, however I won't block the PR and if sufficient number of other approvers find it useful it can get merged.

specification/trace/semantic_conventions/feature-flags.md
scheler: @beeme1mr can you clarify on the semantics of two services participating in the trace adding spans for the same feature flag? And is there a reason why multiple feature flags are not represented in the same span? There's currently an ask to support [map type for attribute values](https://github.com/open-telemetry/opentelemetry-specification/issues/376), but in the absence of it you can use the alternative used by `http.request.headers.<key>` to represent maps, so in this case it would be `feature_flag.<flag_key>.{evaluated_variant, provider_name, etc}`.

semantic_conventions/trace/feature-flag.yaml
```diff
@@ -0,0 +1,35 @@
...
+        examples: ["Flag Manager"]
+      - id: evaluated_variant
+        type: string
+        brief: The human readable name that identifies the evaluated value.
+        examples: ["red", "blue", "green"]
+        requirement_level:
+          conditionally_required: >
+            A variant should be used if it is available. If the variant is
+            present, feature_flag.evaluated_value should be omitted.
+      - id: evaluated_value
```
scheler, line 27: Can you please clarify how a value is different from the variant?

Also, how are the values of boolean feature flags supposed to be represented?

specification/trace/semantic_conventions/feature-flags.md
beeme1mr: > can you clarify on the semantics of two services participating in the trace adding spans for the same feature flag?

This would be represented as two spans.

> And is there a reason why multiple feature flags are not represented in the same span?

These are two separate API calls that need to be represented as two separate spans.

> There's currently an ask to support [map type for attribute values](https://github.com/open-telemetry/opentelemetry-specification/issues/376), but in the absence of it you can use the alternative used by `http.request.headers.<key>` to represent maps, so in this case it would be `feature_flag.<flag_key>.{evaluated_variant, provider_name, etc}`.

Representing as two spans avoids this problem entirely.

beeme1mr: @tigrannajaryan Is is okay to resolve this thread? I feel like the original question was answered and new questions should start a new thread.

semantic_conventions/trace/feature-flag.yaml
beeme1mr: Sure! The reason to have an evaluated value attribute is because not all feature flag vendors have the concept of a variant. In that case, a stringified version of the value can be used. In the case of a boolean value without a variant, the value would be "true" or "false".

specification/trace/semantic_conventions/feature-flags.md
tigrannajaryan: Sure, feel free to resolve.

evantorrie:
Would you envisage feature flag attributes appearing as a "resource" attribute rather than just a tracing span attribute?

We sometimes run experiments with the same source code run side-by-side in 2 deployments deployment (with randomly distributed A/B style user traffic), but often with different compilation options (e.g. -O3 vs. -O2), or perhaps running on JVM 11 vs. JVM 14.  Usually this is for internal regression analysis/comparison of performance.

I could imagine creating a `feature_flag.flag_key=jvm`, `feature_flag.evaluated_variant=jdk14` for such a situation and adding them to the resource, but I don't know if this is what is intended.

Personally for our case, I thought something like a `service.variant` (or perhaps just a suffixed `service.version`) would be appropriate given that `service.*` is already associated with Resources, but this PR caught my attention since it has some similarity in the semantics.

semantic_conventions/trace/feature-flag.yaml
beeme1mr: Does that answer your question?

beeme1mr:
Hi @evantorrie, you could certainly use this proposal for analyzing A/B tests. For example, you could select what Fibonacci algorithm to run using a feature flag. The results could be added to the trace as a dedicated span. In many tools, you can search for specific span name and value to see the full trace. This would allow you to compare the various states to see how the flag impacted the request.

> Would you envisage feature flag attributes appearing as a "resource" attribute rather than just a tracing span attribute?

From my understanding, resource attributes are meant to be set when the system starts and [should be immutable](https://opentelemetry.io/docs/reference/specification/resource/sdk/). A feature flag can be dynamically changed during runtime. The value can also be different based on evaluation context (i.e. enabled for a specific user).

dyladan:
@evantorrie compile time flags or flags which never change value would probably be a reasonable addition to the resource semconv in the future, but this proposal is about flags which may change from request to request.

jmacd:
I think I agree with https://github.com/open-telemetry/opentelemetry-specification/pull/2529#issuecomment-1218160430, that this is not an area calling for semantic conventions (though I feel uncertain about where that line is drawn).

beeme1mr:
Hi @jmacd, could you please elaborate on why you think this is not an appropriate use of the semantic convention? According to the [OTel docs](https://opentelemetry.io/docs/concepts/signals/traces/#attributes), semantic attributes are simply a known naming convention for common operations. In my experience, feature flags are common in most applications nowadays. This would provide recommendations for what to call the span attributes.

Other uses cases, such as a [Feature Gate](https://github.com/open-telemetry/opentelemetry-specification/issues/2532#issuecomment-1231533979), may also benefit from a common convention.

weyert:
Yes, it helps with debugging or analysing requests. I have been using these conventions for a while know and they are really helpful. I don’t see why this aspect wouldn’t be able to be standardised. You can then correlate the data from various sources. For example, if I am tracing requests of a system that uses Launchdarkly, but also another system that uses Split.io  which creates traces or maybe only log them. How would you query these consistently?

specification/trace/semantic_conventions/feature-flags.md
```diff
@@ -0,0 +1,25 @@
...
+|---|---|---|---|---|
+| `feature_flag.flag_key` | string | The unique identifier of the feature flag. | `logo-color` | Required |
+| `feature_flag.provider_name` | string | The name of the service that performs the flag evaluation. | `Flag Manager` | Recommended |
+| `feature_flag.evaluated_variant` | string | The human readable name that identifies the evaluated value. | `red`; `blue`; `green` | Conditionally Required: [1] |
+| `feature_flag.evaluated_value` | string | A string representation of the evaluated value. | `ff0000`; `0000ff`; `00ff00` | Conditionally Required: [2] |
+
+**[1]:** A variant should be used if it is available. If the variant is present, feature_flag.evaluated_value should be omitted.
+
+**[2]:** The value should only be used if the variant is not available. How the value is represented as a string should be determined by the implementer.
+<!-- endsemconv -->
```
arminru, line 25: Please add an example on how a span like this would look like in conjunction with a parent and a sibling span. You can illustrate the structure like we did here, for example: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#examples
Maybe you could add an incoming request for which handling an FF needs to be evaluated before a follow-up request can be made to serve the incoming one?

beeme1mr: I've added an example. Is this what you had in mind?

specification/trace/semantic_conventions/feature-flags.md
```diff
@@ -0,0 +1,25 @@
+# Semantic conventions for Feature Flags
+
+**Status**: [Experimental](../../document-status.md)
+
+This document describes how a feature flag evaluation can be represented as a span in
+OpenTelemetry. The span can be used to determine the impact a feature has on a request.
+
```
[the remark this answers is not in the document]
beeme1mr: Added a motivation section.

specification/trace/semantic_conventions/feature-flags.md
```diff
@@ -0,0 +1,25 @@
+# Semantic conventions for Feature Flags
+
+**Status**: [Experimental](../../document-status.md)
+
+This document describes how a feature flag evaluation can be represented as a span in
+OpenTelemetry. The span can be used to determine the impact a feature has on a request.
```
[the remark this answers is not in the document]
beeme1mr: Added as section explaining how feature flags can be dynamically evaluated.

specification/trace/feature-flags.md
```diff
@@ -0,0 +1,76 @@
...
+- [Convention](#convention)
+- [Example](#example)
+
+<!-- tocstop -->
+
+## Motivation
+
+Features flags are commonly used in modern applications to decouple feature
+releases from deployments. Many feature flagging tools support the ability to
+update flag configurations in near real-time. They also commonly allow rulesets
```
arminru, line 20: Maybe add something like "from a remote (feature flag management) server" to indicate that they're not updated locally.

specification/trace/feature-flags.md
```diff
@@ -0,0 +1,76 @@
...
+to be defined that return value based on contextual information. For example, a
```
arminru, line 21:
```suggestion
to be defined that return values based on contextual information. For example, a
```

specification/trace/semantic_conventions/feature-flags.md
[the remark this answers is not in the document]
arminru: I think adding the example mentioned above with blocking synchronous requests to the flag manager server would also make a good case for why you'd like to instrument flag evaluations so they show up in your traces.

specification/trace/feature-flags.md
```diff
@@ -0,0 +1,76 @@
...
+
+**[2]:** The value should only be used if the variant is not available. How the value is represented as a string should be determined by the implementer.
+
+<!-- endsemconv -->
+
+## Example
+
+The following example shows how a feature flag span can be represented on a trace.
+
+In this scenario, developers are updating an endpoint which returns the nth number in the Fibonacci sequence. They want to monitor the impact of this updated algorithm on the system. After a GET request is made to `/fib`, a feature flag with the key `fib-algo-name` determines which algorithm is used during this calculation. This feature flag is evaluated by a feature flagging service called `Flag Manager`, which returns the new algorithm only for internal test users. The span `Flag Manager fib-algo-name` has the attribute `feature_flag.evaluated_value`, which represents the algorithm that was run during this request. This span information can then be used to understand the impact that a flag(s) had on a request and to determine whether a new feature should be rolled out to a more general audience.
```
arminru, line 66:
```suggestion
In this scenario, developers are updating an endpoint which returns the n-th number in the Fibonacci sequence. They want to monitor the impact of this updated algorithm on the system. After a GET request is made to `/fib`, a feature flag with the key `fib-algo-name` determines which algorithm is used during this calculation. This feature flag is evaluated by a feature flagging service called `Flag Manager`, which returns the new algorithm only for internal test users. The span `Flag Manager fib-algo-name` has the attribute `feature_flag.evaluated_value`, which represents the algorithm that was run during this request. This span information can then be used to understand the impact that a flag(s) had on a request and to determine whether a new feature should be rolled out to a more general audience.
```

arminru: This sounds good to me!
If you could add line breaks after the sentences it would make the diffs easier to review and annotate, by the way.

specification/trace/feature-flags.md
```diff
@@ -0,0 +1,76 @@
+# Semantic conventions for Feature Flags
```
arminru, line 1: This file belongs into the `semantic_conventions` folder (that's also why the link checks are breaking for document-status and the README.md).

semantic_conventions/trace/feature-flag.yaml
```diff
@@ -0,0 +1,35 @@
+groups:
+  - id: feature_flag
+    prefix: feature_flag
```
arminru, line 3:
```suggestion
    prefix: feature_flag
    type: span
```

see https://github.com/open-telemetry/opentelemetry-specification/runs/8159209071

beeme1mr:
Hi @jmacd, I've updated the proposal to include a section on the motivation for the semantic convention. Does that help make the proposal more clear?

arminru (approved):
LGTM now but please make sure that the motivation and explanation recently added addresses the concerns brought up earlier in https://github.com/open-telemetry/opentelemetry-specification/pull/2529#issuecomment-1218160430 and https://github.com/open-telemetry/opentelemetry-specification/pull/2529#issuecomment-1231939532 sufficiently.

jsuereth:
Given the new investment in Logs + Events, shouldn't Feature flags be logged as semantic events (tied to a trace) rather than masking them as Spans?  I still disagree with the motivation to encode flags into Spans.

beeme1mr:
Hi @jsuereth, is there a place to define semantics for events? I can test events to see if they'll be sufficient and update the proposal accordingly.

scheler:
@beeme1mr I am wondering if you considered decorating existing spans with the attributes of feature flags. I quickly looked at the previous comments here and could not determine if you considered it.  I am curious how it compares with creating additional spans per service per feature flag.

dyladan:
> @beeme1mr I am wondering if you considered decorating existing spans with the attributes of feature flags. I quickly looked at the previous comments here and could not determine if you considered it. I am curious how it compares with creating additional spans per service per feature flag.

Decorating the existing span with attributes is possible and we considered it but the question of what to do if multiple flags are evaluated or if a flag is evaluated multiple times isn't an easy one to answer. Span events are likely a better option in that case, but then you lose the duration of the feature flag call which may itself be an asynchronous operation.

>Given the new investment in Logs + Events, shouldn't Feature flags be logged as semantic events (tied to a trace)

I am not at all clear when this is preferable to using span events. are span events to be deprecated in favor of this? if not, we need to work on answering that question at a spec and docs level or many people will be confused why we have both

>I still disagree with the motivation to encode flags into Spans.

@jsuereth Is this is because you think the duration of the flag evaluation isn't relevant?

weyert:
> @jsuereth Is this is because you think the duration of the flag evaluation isn't relevant?

It's pretty helpful to see how much time is spend on flag evaluation as part of a request/task execution :)

dyladan:
not stale

beeme1mr:
Hi @jsuereth, I've updated the motivation section to highlight why a span may be preferred over an event. Does the additional context change your opinion? If not, could you expand on why an event would be preferred over a span? It's not clear to me when one would be preferred over another.

carlosalberto (approved):

beeme1mr:
Hey @jmacd @jsuereth @tigrannajaryan... I tried to address your feedback in the document, please also see the comments above and let me know if there is anything left to address.

specification/trace/semantic_conventions/README.md
```diff
@@ -23,6 +23,7 @@ The following semantic conventions for spans are defined:
 * [FaaS](faas.md): For [Function as a Service](https://en.wikipedia.org/wiki/Function_as_a_service) (e.g., AWS Lambda) spans.
 * [Exceptions](exceptions.md): For recording exceptions associated with a span.
 * [Compatibility](compatibility.md): For spans generated by compatibility components, e.g. OpenTracing Shim layer.
+* [Feature Flags](feature-flags.md): For spans that represent a feature flag evaluation.
```
arminru, line 26: Please make sure to add an entry in CHANGELOG.md (Unreleased > Semantic Conventions).

beeme1mr: Done, thanks!

yurishkuro (commented):
I have a bit of general question on the naming of the attributes.
  1. `feature_flag.flag_key`- why not just `feature_flag.key`?
  2. `feature_flag.evaluated_variant`
    * why `evaluated`, what other options are there?
    * why `variant`? It's a confusing name, why not go with more obvious `string_value`?
  3. `feature_flag.evaluated_value` -> `feature_flag.value` ?

toddbaert:
@yurishkuro I think I prefer all of your proposed changes with the exception of:

> >    * why `variant`? It's a confusing name, why not go with more obvious `string_value`?

The variant is specifically _not_ a value. It's a semantic "label" for the value. The value may be a JSON object, base64 encoded string, or some other payload which is unintelligible to humans. The variant is a way for humans to understand the value in this case. The `value` may also be large and unsuitable to log. This is another reason the variant is useful.

Many SaaS feature flag vendors have the same or similar concepts. Frequently in the vendor case, when creating flag values in a UI you give a "name" or "label" to them. This is what most of our providers (the adapter layer) map into the `variant`.

beeme1mr:
@yurishkuro, thanks for your suggestions. I also agree with @toddbaert that all the changes make sense with the exception of the variant change.

Here are a few examples of the term being used by feature flag vendors:

- https://docs.launchdarkly.com/home/flags/variations
- https://docs.harness.io/article/1j7pdkqh7j-create-a-feature-flag#how_feature_flag_variations_work
- https://help.amplitude.com/hc/en-us/articles/360061687311-Working-with-feature-flags-and-feature-rollouts#create-a-new-flag

I'll update the proposal.

yurishkuro:
@beeme1mr what I am noticing is the vendors you linked use "variant" in the context of "variations", but they never use the word "variant" to refer to the name/description/label, they instead use those words. 

So perhaps `s/variant/label/`

yurishkuro:
fair enough, I think it will be good to link to open-feature from this semantic convention.

yurishkuro:
We had a brief discussion on TC chat about this. One big remaining concern is about statements like this:

> The following semantic convention defines how feature flags can be represented as a span in OpenTelemetry.

This is a big deviation from the existing scope of semantic conventions which are primarily aimed at defining conventions for span attributes, not for the structure of the trace. I did not find the motivation for "a span" particularly convincing. Yes, there are frameworks where retrieving the value of a feature flag may require blocking IO, which doesn't strike me as a particularly good design, because it's mixing the control plane concerns of propagating feature flags with the data plane of request execution. But there are plenty of other feature flag frameworks that do better than that, by cleanly separating control and data plane concerns. So to require a span for recording feature flags in all cases does not seem like an adequate approach. It also seems fairly detrimental to the bigger purpose of recording feature flags - to do data analysis of app performance partitioned by the feature flags. If they were recorded on the server span directly, I could ask questions from the backend "find traces where service=A endpoint=B feature_x=Z", which is much more difficult support if feature_x is not actually recorded on the server span, but on some internal children spans.

My concrete recommendation:
  * remove "inner span" requirements. If the feature flag framework is designed such that it makes blocking IOs to retrieve feature values, it is free to create child spans to mark its own execution, but that's a problem orthogonal to the purpose of semantic conventions for recording feature flags
  * consider different format for recording feature flags, e.g. a map. Requiring an inner span is like a cheat code that gives you independent scope where to record one feature flag, but the same convention does not work if you need to record many feature flags on a single span.

jsuereth:
Sorry for radio silence, but I spent a bit of time reading the OpenFeature specification and trying to wrap my head around it (and then how this ties to OTEL).

So, let me address changes made to this PR and my current problem with it:

1. *It ties a timeline to a Feature-Flag.*
   - Yes, you can record how long a feature flag takes to compute
   - However, aren't you computing more than one feature flag at a time?  Span name has feature flag in it, so you'll just have a ton of spans?
   - If you were truly using this semconv to track how long feature flag computation takes to process, why is their one span name per-feature flag?  Are you planning to instrument EVERY feature flag lookup in the system, or just write a span when the computation of flags happens?   What will you do if you compute once + cache flags as they are looked up?
   - TL;DR; if understanding feature-flag computation time was the goal, then this seems like a poor semconv to track it effectively.  You're blending two concepts together that I think are disjoint.
2. *Interaction with other signals.*
   Why is it only traces interact with feature flags?  What if we wanted to include feature flags as labels on metrics or logs?
   I feel like there's an interaction w/ OTEL baggage here (across signals) that is missing in our specification that would enable a LOT of useful feature-flag based telemtery tracking.  E.g. could I (at the SDK level) ensure a feature flag is added to metrics exported (e.g. latency) to do A:B comparison there?  Why just traces as special?
3. *Feature flag is in span name*
    I'd definitely prefer to see it use Attributes.   Encoding in Name since like an odd decision here, unless you're focused only on backends that can search by existence of sub-spans.  In which case, why don't we try to find a way to get feature flags *on spans* themselves in some way via a baggage-attach-like mechanism.

Overall, I think it'd be much simpler to start by defining a log structure for dumping feature flags *or* defining a mechanism to attach feature flags as attributes to spans in a consistent way that we can (in the future) propogate with baggage or some other contextual hook.

Specifically, I think it should be the case that we can acheive something like:

```
{pseudo-code}

my_request_handler() {
  feature_ctx = resolve_features();
  add_my_features_to_current_otel_scope_for_this_function(feature_ctx);
  
  do_thing_x():
  my_counter.increment();
  
  do_thing_y_with_span();
}
```

Basically a way to add *contextual* attributes to instrumentation that come *not* via instrument configuration, but instead from current request environment (i.e. some kind of direct hook between [Evaluation Context](https://docs.openfeature.dev/docs/specification/sections/evaluation-context) <-> OTel Context and then the ability to configure an SDK to auto-pull attributes from OTEL context for ANY signal.

The current proposal seems pretty limited only to trace and rigid in requirements for how it works and how backends need to deal with it, or which ones would support it.

beeme1mr:
Hi @jsuereth @yurishkuro, thanks for the feedback. It seems like the general consensus is that creating a new span per flag evaluation is not the correct approach. How would you feel about attaching events to the current span? I can experiment with that approach and update the proposal if eventing seems like a better option.

Regarding interacting with other signals, I would be interested in exploring what's available. Metrics would be great but I fear this could cause a metric explosion by introducing too many dimensions. That's just an assumption and I would be happy to work with someone more familiar with metrics in OTel to see what would be possible.

yurishkuro:
I personally don't think events are a good abstraction for feature flags. Events happen at a point in time, that's not how feature flags work, they tend to apply to the whole span, and therefore make more sense as span attributes (using events because they provide a separate namespace for attributes is not a good reason). I would rather see feature flags as an array or a map.

beeme1mr:
Okay, that makes sense. I'll give it a try.

yurishkuro:
@beeme1mr yes, but from this PR is sounds like you need more complexity for each feature flag than just key=value. Or not? The PR mentions variant and raw value, are both required?

Also, `f"feature_flag.{flag_id}"` seems like it may not produce a valid attribute name (e.g. can be dots in the flag_id, we don't know).

MovieStoreGuy (approved):
It looks like there is no outstanding issues for this PR and I am happy for this to be the first iteration of the feature flag semantics.

MovieStoreGuy:
Hey @beeme1mr ,

Could you please resolve the conflict?

beeme1mr:
Hi @yurishkuro, I looked into using an array or map on the current span but it doesn't appear to be possible. Arrays are allowed [according to the spec](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/common#attribute) but may only contain primitive values. Also, it's not possible to access attributes already defined on the current span. That would make it difficult to update the array as feature flags are evaluated.

After experimenting with spans, events, and attributes, I believe that events are the best fit. I feel this way because events are associated with the current span, contain their own attributes, and are queryable in tools like Jaeger.  Also, since flags are evaluated throughout a request and only affect the system after they're evaluated, the point in time nature of an event is appropriate. 

Please let me know what you think. If you're okay with an event, I'll update the proposal accordingly.

yurishkuro:
I don't have strong objection to using events for this, but that does not answer @jsuereth's question 2 (https://github.com/open-telemetry/opentelemetry-specification/pull/2529#issuecomment-1261493987): how would it work with other telemetry signals, given that span events are only captured in traces.

dyladan:
> I don't have strong objection to using events for this, but that does not answer @jsuereth's question 2 ([#2529 (comment)](https://github.com/open-telemetry/opentelemetry-specification/pull/2529#issuecomment-1261493987)): how would it work with other telemetry signals, given that span events are only captured in traces.

IMO the question of multiple-signals is out of scope here. Saying how to represent feature flags on traces doesn't preclude a representation for metrics. If, in the future, there is some specified way to propagate feature flag data through the transaction so that it could be used for multiple signals, this semconv would still apply. IMO what's discussed there is a _much_ larger topic than just feature flags and a general mechanism should be proposed in a separate PR (or even OTEP since it seems like quite an impactful feature).

Without such a mechanism, here are some ways the feature flag data in this PR could be used in other signals:

- Trace-derived metrics could use attributes from traces (including event attributes) as labels if the developer so chose
- An exemplar trace/span on a metric data point could have feature flag data. I'm not aware of any backend that would let you filter metrics by exemplar attributes but such a feature isn't unfeasible.
- Span events could be converted to a regular event/log with a trace context in a processing step

There may be other ways as well.

beeme1mr:
Hi @yurishkuro, I've updated the proposal to use events. I believe this addresses the majority of the concerns that have been brought up so far on this thread.

I would like to see if we can get this proposal approved before extending to other signals. Fortunately, it looks like it should be straightforward to use the same structure for an equivalent logging semantic convention.

beeme1mr:
Hi @tigrannajaryan, I noticed in the most recent TC meeting notes that you would prefer to have the attributes defined more generally. Looking at the current spec, I could replicate what the [exception semantic convention](https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/exception.yaml) looks like and simply reference to those attributes in both the [trace](https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/trace/trace-exception.yaml) and [log](https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/logs/log-exception.yaml) convention. Is that what you had in mind? I'm happy to update the proposal but I want to make sure we're on the same page before investing time in the wrong direction.

jsuereth:
@beeme1mr I can speak to a lot of the TC conversation.  I've been a user of feature-flag tracking/analysis for quite some time, relying on it (the general feature, not your specification) heavily on my time working on Google Assistant.

Effectively, we're asking ourselves a lot of questions on feature-flag tracking right now that we feel aren't (well) answered in this semantic convention, and likely really need some work demonstrating + agreement across vendors to embed in OTEL.

- As a user of feature flagging, do I need to use traces?  Can I track things with just logs?  What's the interaction with metrics?
  - From my experience, considering "live" metrics as out of scope makes a lot of sense, due to cardinality restrictions and issues.  Having a separate system designed + tuned for feature-tracking metrics is generally necessary.
  - Requiring a distributed trace for feature-flag tracking may be a high bar for some.  I'd like to see what a log-only, or log + trace complementary solution looks like here.
- What's important to model w/ feature flags?
   - Right now you need to search for Events w/ `feature_flag.key=foo,feature_flag.variant=bar` and do roll-up metrics with custom code.   Should we aim for a model, e.g. with better interaction with the span->metrics feature of the collector?  What about other features of OTEL?  Should we be concerned with PII or redaction?
   - Would it make more sense to encode feature flags as attributes on spans? e.g. `feature.foo=bar`.  You may have already discussed why not, but it'd be important to capture this somewhere.
- Is there an OpenSource feature-flag tracking solution? Do you have a reference architecture you're targeting with this SemConv?  Does that architecture resonate across different solutions/vendors/OSS options?


Fundamentally, I think this deserves an OTEP for a few reasons:

- It should capture why we're not encoding in alternative manners
- It should capture how feature flags (and feature flag tracking) interacts with various OTEL technologies (e.g. otel collector span-metrics processor)
- It should capture a reference architecture and applicable backends
- It should denote the overall user experience on feature tracking and whether they *need* a trace backend, or a log-based solution is good enough (I'm most comfortable with log-based feature tracking myself, and it's unclear how trace + logs would interact with the current proposal)


At a minimum, if you include the following in the description, and update the PR to address these questions that'd be ideal.

jsuereth:
> Another similar alternative which was brought up was to use either an array or map to represent multiple flags, but this doesn't work because the existing list/map cannot be read using the API meaning it would be near impossible to update the attribute if another flag was read later in the span.

Which API are you talking about here?

> There are several open source feature flagging solutions, but the one that most directly is relevant here is likely the OpenFeature project, which is a CNCF project. Similar to OpenTelemetry it represents a fairly strong industry consensus with most of the major feature flagging vendors collaborating on the spec and implementations.

I'm aware of OpenFeature.  What I'm asking is what should OpenFeature <-> OTel integration look like and that means what should the signal look like within otel.

> To me this seems way beyond the scope of what is trying to be achieved here. The proposal is only trying to capture the existing behavior of feature flags. You seem to be suggesting it should also cover the processing and analysis of the data; something that OTel in general has been explicitly not covering in the spec.

I'm treating this similar to Profiling, Baggage, or other important signals that need to be encoded within OTEL.   I see Feature Flagging as hugely important and beneficial, but I don't want to rush an encoding, particularly one as Traces, when it's not clear that's how this problem is solved in real life.

Again, my experience shows this problem being solved mostly with logs and log-based metrics.  So I find it curious that's not accounted for here.

Semantic conventions are about a shared vocabulary for understanding a production system.  If this was proposing semantic conventions around how to propagate traces through message passing systems, I'd also be asking detailed questions to ensure the modeling is well thought out and works across many implementations of message systems and backends.  I know the Messaging SiG spend a significant amount of time and discussion around this problem to reach consensus.

If that consensus *has* been reached here for Feature Flag encoding in Otel, that's great.  I don't see documented evidence, and I also don't see this proposal working with systems I know that track feature flags for observability.

I think there's two paths forward here:

1. We get a proposal for feature flags that makes it clear how the encoding can be used for general feature-flag observability features.  Something I'd argue is akin to HTTP semconv or Message-Passing semconv in scope.  If we want this to be a standard, let's put in the effort on the OTel side.
2. Don't make this an OTEL standard, just do it as part of the Feature library.  At some point, there will be a weight of evidence that this is the only viable way to represent features in observability, and OTel will adopt it.

The proposal, as-is, limits the future direction of OTel + Feature flagging in a way I'd like to make sure we've really thought through the implications.

> IMO this also points to a lack of clarity as to when the log/event signal should be used vs when a span event should be used.

Highly agree.  This is something I can take back the TC as I think this proposal walks into the thick of that debate.  cc @tigrannajaryan  who I believe wrote guidance on this topic already.

dyladan:
man i wish github had threads :/ sorry for such a long message here

> > Another similar alternative which was brought up was to use either an array or map to represent multiple flags, but this doesn't work because the existing list/map cannot be read using the API meaning it would be near impossible to update the attribute if another flag was read later in the span.
> 
> Which API are you talking about here?

There does not exist an API to read the attributes of the current span, so in order to add multiple feature flags at different times through the span you would have to keep a map or array reference in your own code (really a problem if multiple libraries are used that aren't aware of each other).

> > There are several open source feature flagging solutions, but the one that most directly is relevant here is likely the OpenFeature project, which is a CNCF project. Similar to OpenTelemetry it represents a fairly strong industry consensus with most of the major feature flagging vendors collaborating on the spec and implementations.
> 
> I'm aware of OpenFeature. What I'm asking is what should OpenFeature <-> OTel integration look like and that means what should the signal look like within otel.

I think I see what you're saying here. I think this PR is actually not trying to be OpenFeature specific. It does reference OF but only for definitions. This proposal provides value on its own and doesn't preclude a deeper integration in the future. I don't think it should block this PR.

> > To me this seems way beyond the scope of what is trying to be achieved here. The proposal is only trying to capture the existing behavior of feature flags. You seem to be suggesting it should also cover the processing and analysis of the data; something that OTel in general has been explicitly not covering in the spec.
> 
> I'm treating this similar to Profiling, Baggage, or other important signals that need to be encoded within OTEL. I see Feature Flagging as hugely important and beneficial, but I don't want to rush an encoding, particularly one as Traces, when it's not clear that's how this problem is solved in real life.

I'm not sure I think feature flagging will rise to the level of profiling or baggage which have their own OTLP data model and OTel API respectively. Feature flagging data is definitely important for analysis usecases but I think it will fit neatly into the signals we have.

> Again, my experience shows this problem being solved mostly with logs and log-based metrics. So I find it curious that's not accounted for here.

The biggest issue with logs/events is honestly that it wouldn't be useable today. The API spec is experimental and implementations range from spotty to completely missing. Given the signals that have maturity today, using span events is a pragmatic solution.

Even if there was a mature logs/events client, this proposal deals mostly with _dynamic_ feature flags which have a direct impact on the current span. There might actually be 2 different types of feature flags to deal with here. Feature flags which affect the whole service/process which are likely best represented as logs, and feature flags which only affect the current operation which span events describe nicely.

> Semantic conventions are about a shared vocabulary for understanding a production system. If this was proposing semantic conventions around how to propagate traces through message passing systems, I'd also be asking detailed questions to ensure the modeling is well thought out and works across many implementations of message systems and backends. I know the Messaging SiG spend a significant amount of time and discussion around this problem to reach consensus.
> 
> If that consensus _has_ been reached here for Feature Flag encoding in Otel, that's great. I don't see documented evidence, and I also don't see this proposal working with systems I know that track feature flags for observability.

I don't think a consensus has been reached. There have been a bunch of proposals and the PR has evolved a few times based on those proposals. The existing PR describes feature flags as a span event, but @beeme1mr a few days ago suggested that it might be better modeled off the existing exception semconv, which can be modeled as a span event or as a log, whichever the instrumentation author deems most appropriate. His last comment seems to be asking if that would satisfy the latest set of concerns (before yours were raised).

> I think there's two paths forward here:
> 
> 1. We get a proposal for feature flags that makes it clear how the encoding can be used for general feature-flag observability features.  Something I'd argue is akin to HTTP semconv or Message-Passing semconv in scope.  If we want this to be a standard, let's put in the effort on the OTel side.
> 2. Don't make this an OTEL standard, just do it as part of the Feature library.  At some point, there will be a weight of evidence that this is the only viable way to represent features in observability, and OTel will adopt it.

This is not something I can address but really needs the original author.

> The proposal, as-is, limits the future direction of OTel + Feature flagging in a way I'd like to make sure we've really thought through the implications.

I'm not sure I see how the current proposal limits future plans. Can you give an example which would be prevented by this proposal?

> > IMO this also points to a lack of clarity as to when the log/event signal should be used vs when a span event should be used.
> 
> Highly agree. This is something I can take back the TC as I think this proposal walks into the thick of that debate. cc @tigrannajaryan who I believe wrote guidance on this topic already.

I'd appreciate it if you would. Another part of this debate which seems relevant in a similar manner is the lack of guidance on how to report telemetry outside of semantic conventions. If, for example, @beeme1mr implemented (2), then it would possibly collide with a future OTel spec which would be bad for everyone. It would be ideal if we could have some namespace for attributes/events which explicitly marked them as non-semconv or not-yet-semconv.

jsuereth:
@dyladan Thanks for the well thought out response.  I'm going to miss today's Spec SiG due to having my car inspected, but I'd be happy to talk this out in a higher bandwidth forum and then come back here with our latest round of thinking to make progress.

TL;DR; Let's not let some of the skeletons in our closet (Event vs. Logs, SemConv vs. Documented stable instrumentation) block progress on feature-flag tracking.

tigrannajaryan:
> > Highly agree. This is something I can take back the TC as I think this proposal walks into the thick of that debate. cc @tigrannajaryan who I believe wrote guidance on this topic already.
> 
> I'd appreciate it if you would.

I am not sure a guidance about using standalone logs vs span events exists. I agree it is useful/important to have. It possible is going revolve around these:
- Do you need to record an event that doesn't fit Span Event data model? Then use LogRecord. (Typically applies to external log/event data sources).
- Do you need to record an event that happened where the there is no known Span Context? The use LogRecord.
- Do you need to record an event that clearly happen in a context of Span? Then use a Span Event.

There is probably more that we can say, these are just from the top of my head and unfortunately these do not seem to be very helpful for deciding where to put the feature-flags.


> It would be ideal if we could have some namespace for attributes/events which explicitly marked them as non-semconv or not-yet-semconv.

We have a bit of guidance on this: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-naming.md#recommendations-for-application-developers

beeme1mr:
Hi @tigrannajaryan, thanks, that helps. Based on those guidelines, I think a span event would be the best fit because the primary benefit of capturing feature flag metadata in OTel would be to have it correlate with a transaction. Many popular feature flagging tools allow for the return value to be determined based on transaction context. Understanding how a feature flag(s) impacted a request is the primary goal.

Defining semantics for a LogRecord may also make sense. It may be an ideal fit for feature flags that aren't tied to a specific transaction but I would expect the span event and log semantic convention to look very similar. It would then be up to the consumer to decide if logs or span events are the better fit.

Quickly look at the currently defined semantic conventions, it looks like exceptions are the only example of this being done. Do you think this would be a good approach here as well? I can update the proposal if that sounds like the right approach.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/exception.yaml

jsuereth:
> Quickly look at the currently defined semantic conventions, it looks like exceptions are the only example of this being done. Do you think this would be a good approach here as well? I can update the proposal if that sounds like the right approach.

Yes. I think taking a similar modelling approach to exceptions here is likely the right call for how to deal with Trace + Logs.

weyert:
My main worry as a production user OpenFeature and Opentelemetry is the limited support for Logs and Events in some tracing backends. For example, Google Trace hasn't improved their tracingUI for over a year and I not sure how the events/logs would work with such a solution. 

Wouldn't we want to take into account what the current experience?

dyladan:
This is not stale

dyladan:
Updated to add a log version

beeme1mr:
Hi @tigrannajaryan and @jsuereth, does the recent changes made by @dyladan address your concerns?

specification/trace/semantic_conventions/feature-flags.md
```diff
@@ -0,0 +1,52 @@
...
+
+<!-- toc -->
+
+- [Motivation](#motivation)
+- [Overview](#overview)
+- [Convention](#convention)
+
+<!-- tocstop -->
+
+## Motivation
```
jsuereth, line 15: Can we link / x-link to this motivation on the log markdown as well?

jsuereth (approved):
This answers my main concerns on log/trace usage.  Just a minor nit / visibility concern left.

specification/trace/semantic_conventions/feature-flags.md
dyladan: Just pushed an update. PTAL

dyladan:
Can someone with permission please run the link check again? It failed on jsonlines.org which is not a dead link

dyladan:
Link check is passing locally and appears to be on a file we haven't touched

dyladan:
@tigrannajaryan @SergeyKanzhelev I think all the feedback has been addressed and there are 5 green approvals. I think this can probably be merged

PR #2534, opened 2022-05-11
Consolidate tools directories, move things to internal/tools

bogdandrutu:


Makefile
```diff
@@ -85,7 +85,7 @@ table-check:
 
 .PHONY: schema-check
 schema-check:
-	cd tools && ./schema_check.sh
+	cd $(TOOLS_DIR) && ./schema_check.sh
```
yurishkuro, line 88: I would recommend calling `$(TOOLS_DIR)/schema_check.sh` and then in the script `root_dir=$PWD`, to minimize the back & forth

reyang (approved):

Makefile
bogdandrutu: Done.

yurishkuro (approved):

PR #2538, opened 2022-05-13
Clarify async instrument callback identity

### LINKED ISSUE #2408: Clarify how should SDK invokes async instrument callbacks

legendecas:
Fixes #2408

## Changes

> Callbacks registered after the time of instrument creation MAY be associated with multiple instruments.

- Clarify the identity of the statement "one callback associated with an instrument MUST be evaluated exactly once".


reyang (approved):

reyang:
@jmacd @jsuereth need your help to review this PR, thanks!

specification/metrics/api.md
```diff
@@ -355,9 +355,10 @@ Where the API supports registration of `callback` functions after
 asynchronous instrumentation creation, the user MUST be able to undo
 registration of the specific callback after its registration by some means.
 
-Every currently registered Callback associated with an instrument MUST
-be evaluted exactly once during collection prior to reading data for
-that instrument.
+Every registered Callback associated with a set of instruments MUST
+be evaluated exactly once during collection prior to reading data for
+that instrument. The evaluation is identified using a combination of the
+callback itself, and the instruments set.
```
jack-berg, line 361: It's not clear to me what exactly is being clarified with this language.

legendecas: As mentioned in the OP, callbacks can be associated with multiple instruments as stated by the spec. So the current statement "one callback associated with an instrument MUST be evaluated exactly once" is not clear if one callback is associated with multiple instruments, and how many times it should be evaluated for each collection.

This PR tries to explicitly add an identity statement for that: the evaluation is identified with the callback itself, and its associated instruments set.

jack-berg: I see. That was always implied to me since if a multiple instrument callback was called multiple times during collection (once for each associated instrument), it would defeat the purpose of multiple instrument callbacks. 

From an implementation standpoint, java achieves this by sequentially evaluating every callback that has a reader (i.e. the reader has a view for the instrument that is not noop). After all callbacks are evaluated, all the instruments are read.

jmacd: I see what's being questioned and clarified. I have trouble parsing the last of the replacement sentences.

It would have been easier to state that all callbacks are executed once per collection per reader per Meter. However, if you have views in place so that only one metric instrument is exporting, then it should be possible _not_ to evaluate some of the callbacks, right? That is why I wrote it the way it's written, "... be evaluated exactly once during collection ...", so that no matter how many of the instruments are being read, if at least one of the instruments is being read the callback has to be executed exactly once. To execute it once per instrument would not be exactly once.

I'm OK improving this, but I'd like to find more clear words than "The evaluation is identified using a combination of the callback itself, and the instruments set."

legendecas: @jmacd thank you for the suggestion. However, I'm having trouble following the idea of

> However, if you have views in place so that only one metric instrument is exporting, then it should be possible not to evaluate some of the callbacks, right?

would you mind elaborating on this problem?

legendecas:
@open-telemetry/specs-metrics-approvers this PR is still valid but I don't have the permission to re-open it. I'd appreciate it if anyone could help me to reopen it. Thank you <3.

specification/metrics/api.md
```diff
@@ -355,9 +355,10 @@ Where the API supports registration of `callback` functions after
 asynchronous instrumentation creation, the user MUST be able to undo
 registration of the specific callback after its registration by some means.
 
-Every currently registered Callback associated with an instrument MUST
-be evaluted exactly once during collection prior to reading data for
-that instrument.
+Every registered Callback associated with a set of instruments MUST
+be evaluated exactly once during collection prior to reading data for
+that instrument. The evaluation is identified using a combination of the
```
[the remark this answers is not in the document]
legendecas: Done.

specification/metrics/api.md
legendecas: > I'm OK improving this, but I'd like to find more clear words than "The evaluation is identified using a combination of the callback itself, and the instruments set."

I tried to reword the statement but found out that the amendment may have an identical meaning to the preceding statements. So I just removed this and left the PR as fixing the word "instrument" to "instrument set".

legendecas:
@jmacd @jsuereth would you mind taking a look at this again? thank you!

legendecas:
this is not stale.

@jmacd @jsuereth would you mind taking a look at this again? thank you!

jmacd (approved):

PR #2556, opened 2022-05-19
Add network metrics to process semantic conventions

codeboten:
## Changes

Adding network input/output as a process semantic convention. This could be used to report some of the metrics in the redis semconv pr

https://github.com/open-telemetry/opentelemetry-specification/pull/2547/files#r876330741


jmacd (approved):

carlosalberto (approved):

carlosalberto:
Probably I like `in` / `out` better, but that's just me ;) Overall LGTM.

specification/metrics/semantic_conventions/process-metrics.md
```diff
@@ -34,6 +34,7 @@ Below is a table of Process metric instruments.
...
 | `process.disk.io`         | Counter                                           | By    | Disk bytes transferred.                                                                                                             | `direction` SHOULD be one of: `read`, `write`                                                                                                                                                   |
+| `process.network.io`      | Counter                                           | By    | Network bytes transferred.                                                                                                          | `direction` SHOULD be one of: `input`, `output`                                                                                                                                                 |
```
bogdandrutu, line 37: Do we also have "device"? similar with `system.network.io`?

reyang (approved):

specification/metrics/semantic_conventions/process-metrics.md
```diff
@@ -34,6 +34,7 @@ Below is a table of Process metric instruments.
...
```
arminru, line 38: CHANGELOG please 🙂

arminru (approved):

specification/metrics/semantic_conventions/process-metrics.md
arminru, line 37:
```suggestion
| `process.network.io`      | Counter                                           | By    | Network bytes transferred.                                                                                                          | `direction` SHOULD be one of: `in`, `out`                                                                                                                                                 |
```

@carlosalberto good point! We already use in and out here and should stay consistent:

https://github.com/open-telemetry/opentelemetry-specification/blob/0c793b450098e8d7a739e575897707f7db08d8ce/specification/metrics/semantic_conventions/instrumentation/kafka.md?plain=1#L23

https://github.com/open-telemetry/opentelemetry-specification/blob/ef5dc5f4d1d187ae44d1e2a1c3938e074d17bfdc/specification/metrics/semantic_conventions/system-metrics.md?plain=1#L58-L59

carlosalberto: @reyang For network I think I slightly prefer `in` and `out` - also we are already using `in`/ `out` as pointed out by @arminru 

For local operations we can stick to `read` and `write` though (they also sound better for local context).

carlosalberto: @bogdandrutu Could we perhaps add it later if/as needed? I have seen a few Collector receivers (and similar components, e.g. DDog integrations) reporting the in/out bytes, without specifying which device. 

My impression overall is that, from a system perspective (`system.network.io`) it's more common to get the device info, and not so common for process level metrics.

carlosalberto: Ping @bogdandrutu  ;)

specification/metrics/semantic_conventions/process-metrics.md
codeboten: Added

specification/metrics/semantic_conventions/process-metrics.md
codeboten: Updated to `in`/`out`

bogdandrutu: I am fine if this information is not easily accessible.

tigrannajaryan:
I think this PR also violates our guidelines (along with other already existing metrics). See https://github.com/open-telemetry/opentelemetry-specification/issues/2589
I am not blocking the PR since we may prefer to change the guidelines (I don't know).

tigrannajaryan:
I think we can go ahead and merge this. https://github.com/open-telemetry/opentelemetry-specification/issues/2589 needs to be resolved regardless and when it is solved all affected semantic conventions (including this one) may need to be updated.

cijothomas (approved):

MrAlias (approved):

carlosalberto:
Any concerns regarding the resulting labels (`receive` and `transmit`)? FWIW, they are also used in this PR: https://github.com/open-telemetry/opentelemetry-specification/pull/2617 (it's _almost_ always good to have uniformity)

If no concern arises, I will merge this PR.

reyang:
> @arminru I was going to update whichever of this or #2617 was merged last with the changes to ensure consistency. I don't have a strong preference one way or the other. This PR already has enough approvals to be merged. I would suggest we merge this and I can update #2617 accordingly. WDYT?

+1 on making the update in #2617.

PR #2568, opened 2022-05-20
Add trace semantic conventions attributes for capturing application layer protocols

### LINKED ISSUE #2548: Add semantic convention attributes for capturing the application layer protocol

Currently, the `net.transport` attribute allows capturing the transport layer protocol used (like TCP, UDP, or IP). However, there is no generic attribute to capture the application layer protocol (like AMQP, DNS, SSH, NTP). Where it is currently captured, it is done so via attributes specific to certain areas, e. g. as `messaging.protocol`, or somehow obfuscated in `http.flavor`.

Having a general way to capture the application layer protocol would allow cleaning up HTTP and messaging semantic conventions and allow for more alignment between the different areas.

**Additional context.**

I would propose to add two new attributes:
`net.protocol.name`
`net.protocol.version`

The first would capture the name of an application layer protocol according to the OSI Model, the second would capture a version of the protocol. The naming is consistent with ECS, where [`network.protocol`](https://www.elastic.co/guide/en/ecs/current/ecs-network.html#field-network-protocol) is used.

Those general attributes could be used to replace the existing `messaging.protocol`, `messaging.protocol_version`, and `http.flavor`.

If that sounds feasible, I'm happy to submit a PR.

pyohannes:
Fixes #2548

## Changes

This PR proposed the two additional attributes `net.procol.name` and `net.protocol.version`. These allow capturing the name and version of the application layer protocol used (in addition to the name of the transport protocol captured in `net.transport`). The naming is consistent with ECS, where [network.protocol](https://www.elastic.co/guide/en/ecs/current/ecs-network.html#field-network-protocol) is used.

Those general attributes could be used to replace the existing attributes `messaging.protocol` and `messaging.protocol_version` by the proposed generic attributes, which can also be reused for other use cases.

### DIFF
--- a/semantic_conventions/trace/general.yaml
+++ b/semantic_conventions/trace/general.yaml
@@ -33,6 +33,14 @@ groups:
               brief: 'Something else (non IP-based).'
         brief: >
           Transport protocol used. See note below.
+      - id: protocol.name
+        type: string
+        brief: 'Application layer protocol used. The value SHOULD be normalized to lowercase.'
+        examples: ['amqp', 'http', 'nntp']
+      - id: protocol.version
+        type: string
+        brief: 'Version of the application layer protocol used.'
+        examples: '0.9.1'
       - id: peer.ip
         type: string
         brief: >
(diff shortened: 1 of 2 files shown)


reyang:
Need to follow https://github.com/open-telemetry/opentelemetry-specification/blob/main/CONTRIBUTING.md#proposing-a-change.

PR #2571, opened 2022-05-21
Supplementary guidance for metrics additive property

reyang:
## Changes

Explained the additive property, overflow/underflow, dynamic range, IEEE 754 double saturation, IEEE 754 subnormal, etc.

This PR doesn't change the spec, it provides additional guidance and examples to help people better understand the metrics specifications.

@jmacd @jsuereth FYI


specification/metrics/supplementary-guidelines.md
```diff
@@ -79,6 +79,128 @@ Here is one way of choosing the correct instrument:
...
+  | Earth       | 5.97e24         |
+
+* The voltage of battery cells can be added up if the batteries are connected in
+  series. However, if the batteries are connected in parallel, it makes no sense
+  to add up the voltage values anymore.
+
+In OpenTelemetry, each [Instrument](./api.md#instrument) implies the additive
+property. For example, [Counter](./api.md#counter) is additive while
+[Asynchronous Gauge](./api.md#asynchronous-gauge) is non-additive.
+
```
yurishkuro, line 114: Do you want to add a section header here? The discussion below seems unrelated to instruments being additive.

yurishkuro (approved):

specification/metrics/supplementary-guidelines.md
```diff
@@ -79,6 +79,128 @@ Here is one way of choosing the correct instrument:
...
+  | Mercury     | 3.30e23         |
+  | Mars        | 6.42e23         |
+  | Venus       | 4.87e24         |
+  | Earth       | 5.97e24         |
+
+* The voltage of battery cells can be added up if the batteries are connected in
+  series. However, if the batteries are connected in parallel, it makes no sense
+  to add up the voltage values anymore.
+
+In OpenTelemetry, each [Instrument](./api.md#instrument) implies the additive
```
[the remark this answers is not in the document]
reyang: I've improved the wording, thanks for the catch!

specification/metrics/supplementary-guidelines.md
reyang: Good point, I've added sections headers and made small adjustments to the wording (to match with the newly introduced section headers).

specification/metrics/supplementary-guidelines.md
```diff
@@ -79,6 +82,139 @@ Here is one way of choosing the correct instrument:
...
+  add up to `226.2`, but this value has no practical meaning.
+
+  | Host Name | Temperature (F) |
+  | --------- | --------------- |
+  | MachineA  | 58.8            |
+  | MachineB  | 86.1            |
+  | MachineC  | 81.3            |
+
+* The mass of planets is additive, the value `1.18e25` (`3.30e23 + 6.42e23 +
+  4.87e24 + 5.97e24`) means the total mass of terrestrial planets in the solar
```
yurishkuro, line 100: Better rephrase this because this is not the _total_ mass, only 4 planets

specification/metrics/supplementary-guidelines.md
```diff
@@ -79,6 +82,139 @@ Here is one way of choosing the correct instrument:
 
 ### Additive property
 
+In OpenTelemetry a [Measurement](./api.md#measurement) encapsulates a value and
+a set of [`Attributes`](../common/README.md#attribute). Depending on the nature
+of the measurements, they can be additive, non-additive or somewhere in the
+middle. Here are some examples:
```
yurishkuro, line 88: Which instrument would be "in the middle"?

specification/metrics/supplementary-guidelines.md
```diff
@@ -79,6 +82,139 @@ Here is one way of choosing the correct instrument:
...
+In OpenTelemetry, each [Instrument](./api.md#instrument) implies whether it is
+additive or not. For example, [Counter](./api.md#counter) is additive while
+[Asynchronous Gauge](./api.md#asynchronous-gauge) is non-additive.
+
+#### Numeric type selection
+
+For Instruments which take increments and/or decrements as the input (e.g.
+[Counter](./api.md#counter) and [UpDownCounter](./api.md#updowncounter)), the
+underlying numeric types (e.g. signed integer, unsigned integer, double) being
+used have direct impact on the dynamic range, precision and how the data is
```
yurishkuro, line 123:
```suggestion
underlying numeric types (e.g., signed integer, unsigned integer, double) 
have direct impact on the dynamic range, precision, and how the data is
```

specification/metrics/supplementary-guidelines.md
```diff
@@ -79,6 +82,139 @@ Here is one way of choosing the correct instrument:
...
+
+* During (T<sub>0</sub>, T<sub>1</sub>], we reported `131,108`.
+* During (T<sub>1</sub>, T<sub>2</sub>], we reported `375,463`.
+* During (T<sub>2</sub>, T<sub>3</sub>], we reported `832,019`.
+* During (T<sub>3</sub>, T<sub>4</sub>], we reported `1,257,308`.
+* During (T<sub>4</sub>, T<sub>5</sub>], we reported `1,860,103`.
+* ...
+* During (T<sub>n</sub>, T<sub>n+1</sub>], we reported `9,007,199,254,325,789`.
+* During (T<sub>n+1</sub>, T<sub>n+2</sub>], we reported `9,007,199,254,740,992`.
+* During (T<sub>n+2</sub>, T<sub>n+3</sub>], we reported `9,007,199,254,740,992`.
```
yurishkuro, line 199: Feels like we should have a prescribed convention for instrumentation to handle this use case better, eg by resetting.

yurishkuro (approved):

specification/metrics/supplementary-guidelines.md
reyang: What about "combined mass of terrestrial planets"? https://en.wikipedia.org/wiki/Terrestrial_planet

specification/metrics/supplementary-guidelines.md
reyang: In OpenTelemetry Metrics API, we don't have a way to precisely describe such situation. This was discussed a year ago during the API design discussion and considered as a very corner case, and in such case, the developer can use gauge.

specification/metrics/supplementary-guidelines.md
```diff
@@ -79,6 +82,139 @@ Here is one way of choosing the correct instrument:
...
+* During (T<sub>2</sub>, T<sub>3</sub>], we reported `116`.
+* During (T<sub>3</sub>, T<sub>4</sub>], we reported `128`.
+* During (T<sub>4</sub>, T<sub>5</sub>], we reported `128`.
+* During (T<sub>5</sub>, T<sub>6</sub>], we reported `173`.
+* ...
+* During (T<sub>n</sub>, T<sub>n+1</sub>], we reported `1,872`.
+* During (T<sub>n+1</sub>, T<sub>n+2</sub>], we reported `35`.
+* During (T<sub>n+2</sub>, T<sub>n+3</sub>], we reported `76`.
+
+In the above case, the backend system could tell that there was likely a system
```
cijothomas, line 145:
```suggestion
In the above case, a backend system could tell that there was likely a system
```

specification/metrics/supplementary-guidelines.md
```diff
@@ -79,6 +82,139 @@ Here is one way of choosing the correct instrument:
...
+* During (T<sub>m</sub>, T<sub>m+1</sub>], we reported `32,758`.
+* During (T<sub>m+1</sub>, T<sub>m+2</sub>], we reported `32,762`.
+* During (T<sub>m+2</sub>, T<sub>m+3</sub>], we reported `-32,738`.
+* During (T<sub>m+3</sub>, T<sub>m+4</sub>], we reported `-32,712`.
+
+In the above case, the backend system could tell that there was an integer
+overflow during (T<sub>m+2</sub>, T<sub>m+3</sub>], so it has chance to adjust
+the data to:
+
+* (T<sub>m+2</sub>, T<sub>n+3</sub>] : `32,798` (32,762 + 36).
```
cijothomas, line 163:
```suggestion
* (T<sub>m+2</sub>, T<sub>m+3</sub>] : `32,798` (32,762 + 36).
```

specification/metrics/supplementary-guidelines.md
```diff
@@ -79,6 +82,139 @@ Here is one way of choosing the correct instrument:
...
+* (T<sub>m+3</sub>, T<sub>n+4</sub>] : `32,824` (32,762 + 62).
```
cijothomas, line 164:
```suggestion
* (T<sub>m+3</sub>, T<sub>m+4</sub>] : `32,824` (32,762 + 62).
```

jack-berg (approved):

specification/metrics/supplementary-guidelines.md
reyang: I've replaced "total" with "combined".

specification/metrics/supplementary-guidelines.md
```diff
@@ -79,6 +82,139 @@ Here is one way of choosing the correct instrument:
...
+* During (T<sub>3</sub>, T<sub>4</sub>], we reported `128`.
+* During (T<sub>4</sub>, T<sub>5</sub>], we reported `128`.
+* During (T<sub>5</sub>, T<sub>6</sub>], we reported `173`.
+* ...
+* During (T<sub>n</sub>, T<sub>n+1</sub>], we reported `1,872`.
+* During (T<sub>n+1</sub>, T<sub>n+2</sub>], we reported `35`.
+* During (T<sub>n+2</sub>, T<sub>n+3</sub>], we reported `76`.
+
+In the above case, a backend system could tell that there was likely a system
+restart (and the counter was reset to zero) during (T<sub>n+1</sub>,
```
tigrannajaryan, line 146: This is unclear. How can the system tell the restart happened? Is it because the counter value is smaller than the previous reported value? What if the restart happened but the counter managed to quickly ramp up to greater than previous value? We no longer can tell a restart happened, right?

cijothomas: I think the key here is "likely". Not with 100% certainty.

specification/metrics/supplementary-guidelines.md
```diff
@@ -79,6 +82,139 @@ Here is one way of choosing the correct instrument:
...
+
+* During (T<sub>0</sub>, T<sub>1</sub>], we reported `70`.
+* During (T<sub>1</sub>, T<sub>2</sub>], we reported `115`.
+* During (T<sub>2</sub>, T<sub>3</sub>], we reported `116`.
+* During (T<sub>3</sub>, T<sub>4</sub>], we reported `128`.
+* During (T<sub>4</sub>, T<sub>5</sub>], we reported `128`.
+* During (T<sub>5</sub>, T<sub>6</sub>], we reported `173`.
+* ...
+* During (T<sub>n</sub>, T<sub>n+1</sub>], we reported `1,872`.
+* During (T<sub>n+1</sub>, T<sub>n+2</sub>], we reported `35`.
```
dashpole, line 142: In order to be consistent with the [data model](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#sums), this point (and subsequent points) needs to have a start time between `T<sub>n+1</sub>` and `T<sub>n+1</sub>`, since the value 35 describes the number of transactions between the overflow time and `T<sub>n+1</sub>`.  It wouldn't be accurate to report 35 transaction between `T<sub>0</sub>` and `T<sub>n+1</sub>`, when the actual number is 1,907.  If the number of events for a time window is too large to represent, we should choose a smaller time window (i.e. reset), rather than provide an incorrect value.

specification/metrics/supplementary-guidelines.md
```diff
@@ -79,6 +82,139 @@ Here is one way of choosing the correct instrument:
...
+
+Let's take an example: a 16-bit signed integer is used to count the committed
+transactions in a database, reported as cumulative sum every 15 seconds:
+
+* During (T<sub>0</sub>, T<sub>1</sub>], we reported `70`.
+* During (T<sub>1</sub>, T<sub>2</sub>], we reported `115`.
+* During (T<sub>2</sub>, T<sub>3</sub>], we reported `116`.
+* During (T<sub>3</sub>, T<sub>4</sub>], we reported `128`.
+* During (T<sub>4</sub>, T<sub>5</sub>], we reported `128`.
+* During (T<sub>5</sub>, T<sub>6</sub>], we reported `173`.
```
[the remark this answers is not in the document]
reyang: Good callout, let me make the change.

specification/metrics/supplementary-guidelines.md
```diff
@@ -79,6 +82,139 @@ Here is one way of choosing the correct instrument:
...
+  | Mars        | 6.42e23         |
+  | Venus       | 4.87e24         |
+  | Earth       | 5.97e24         |
+
+* The voltage of battery cells can be added up if the batteries are connected in
+  series. However, if the batteries are connected in parallel, it makes no sense
+  to add up the voltage values anymore.
+
+In OpenTelemetry, each [Instrument](./api.md#instrument) implies whether it is
+additive or not. For example, [Counter](./api.md#counter) is additive while
```
bogdandrutu, line 115: add "UpDownCounter" as an example here as being additive.

specification/metrics/supplementary-guidelines.md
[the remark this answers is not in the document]
reyang: I've made the update, PTAL.

specification/metrics/supplementary-guidelines.md
reyang: Added.

specification/metrics/supplementary-guidelines.md
reyang: The correct way is to use the start time rather than telling whether the number decreased or not. I've rewritten this section, please take a look.

specification/metrics/supplementary-guidelines.md
reyang: Thanks, I've rewritten this section https://github.com/open-telemetry/opentelemetry-specification/pull/2571#discussion_r880795973.

specification/metrics/supplementary-guidelines.md
```diff
@@ -79,6 +82,139 @@ Here is one way of choosing the correct instrument:
...
+
+Imagine we keep the database running:
+
+* During (T<sub>0</sub>, T<sub>m+1</sub>], we reported `32,758`.
+* During (T<sub>0</sub>, T<sub>m+2</sub>], we reported `32,762`.
+* During (T<sub>0</sub>, T<sub>m+3</sub>], we reported `-32,738`.
+* During (T<sub>0</sub>, T<sub>m+4</sub>], we reported `-32,712`.
+
+In the above case, the backend system could tell that there was an integer
+overflow during (T<sub>m+2</sub>, T<sub>m+3</sub>] (because the start time
```
dashpole, line 162: Should we be adjusting the start time in these overflow cases as well?

specification/metrics/supplementary-guidelines.md
```diff
@@ -79,6 +82,139 @@ Here is one way of choosing the correct instrument:
...
+For Instruments which take increments and/or decrements as the input (e.g.
+[Counter](./api.md#counter) and [UpDownCounter](./api.md#updowncounter)), the
+underlying numeric types (e.g., signed integer, unsigned integer, double) have
+direct impact on the dynamic range, precision, and how the data is interpreted.
+Typically, integers are precise but have limited dynamic range, and might see
+overflow/underflow. [IEEE-754 double-precision floating-point
+format](https://wikipedia.org/wiki/Double-precision_floating-point_format) has a
+wide dynamic range of numeric values with the sacrifice on precision.
+
+##### Integer
```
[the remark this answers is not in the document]
reyang: The overflow / underflow part should also apply to UpDownCounter.

For example, a power meter could support both increment (when power being consumed) and decrement (when the solar panels are sending power back to the grid).

The reported number could go down until it underflows and becomes positive:
-1, -3, -8, -10, ...., -32765, +32762, +32755, ...

specification/metrics/supplementary-guidelines.md
reyang: Good question. I think we can, but don't have to. Whether we should encourage it or not - I don't have a strong opinion right now.

For example - for systems that can easily detect overflow/underflow without big perf penalty, it might be a viable solution. If performing such detection would cause big perf overhead (e.g. extra check, extra contention), we probably want to discourage it.

specification/metrics/supplementary-guidelines.md
```diff
@@ -79,6 +82,139 @@ Here is one way of choosing the correct instrument:
...
+
+* The voltage of battery cells can be added up if the batteries are connected in
+  series. However, if the batteries are connected in parallel, it makes no sense
+  to add up the voltage values anymore.
+
+In OpenTelemetry, each [Instrument](./api.md#instrument) implies whether it is
+additive or not. For example, [Counter](./api.md#counter) is additive while
+[Asynchronous Gauge](./api.md#asynchronous-gauge) is non-additive.
+
+#### Numeric type selection
```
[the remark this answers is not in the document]
reyang: Based on my previous experience working on OpenCensus/OpenTelemetry Python, I would suggest that we cast the integers to 64-bit signed integers in the OTLP exporter:
* `2^63` maps to `-(2^63-1)`
* `-2^64` maps to `0`

I don't recommend float because the precision issue.

specification/metrics/supplementary-guidelines.md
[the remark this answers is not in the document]
aabmass: I might be missing something, but it's perfectly valid for non-monotonic sum to jump from -32765 to +32762 without there being an underflow?

specification/metrics/supplementary-guidelines.md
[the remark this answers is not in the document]
aabmass: I agree that probably makes the most sense assuming backends can correct for overflow

specification/metrics/supplementary-guidelines.md
```diff
@@ -79,6 +82,139 @@ Here is one way of choosing the correct instrument:
...
+  | Mercury     | 3.30e23         |
+  | Mars        | 6.42e23         |
+  | Venus       | 4.87e24         |
+  | Earth       | 5.97e24         |
+
+* The voltage of battery cells can be added up if the batteries are connected in
+  series. However, if the batteries are connected in parallel, it makes no sense
+  to add up the voltage values anymore.
+
+In OpenTelemetry, each [Instrument](./api.md#instrument) implies whether it is
```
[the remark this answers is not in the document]
aabmass: SGTM feel free to resolve

aabmass (approved):

specification/metrics/supplementary-guidelines.md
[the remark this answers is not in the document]
aabmass: > one has to add some big numbers during a single reporting cycle

So the backend should know that because there was a big jump, it was likely an underflow and not that the user did `up_down_counter.add(65527`?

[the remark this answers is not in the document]
reyang: > So the backend should know that because there was a big jump, it was likely an underflow and not that the user did `up_down_counter.add(65527`?

Yep.

The user won't be able do `up_down_counter.add(65527)` for 16-bit signed int because the only valid value should be within `[-32768, 32767]`; however, the user can call `add(10000)` several times during a collection/exporting cycle and cause the same problem.

If the intention is to be able to take big delta values (e.g. `add(10000)`), a proper dynamic range should be used (e.g. int32, int64 or even double).

If the intention is to take lots (high frequency) of small values (e.g. calling `add(10)` thousands of times per minute), either a bigger dynamic range, or a finer granularity collection/exporting cycle could help.

specification/metrics/supplementary-guidelines.md
reyang: Good point. To keep the PR small, I can address this in a follow up PR.

specification/metrics/supplementary-guidelines.md
reyang: This discussion also covered a bit about "in the middle" situation https://github.com/open-telemetry/opentelemetry-specification/pull/2571#discussion_r880771804.

specification/metrics/supplementary-guidelines.md
```diff
@@ -79,6 +82,139 @@ Here is one way of choosing the correct instrument:
...
+additive or not. For example, [Counter](./api.md#counter) and
+[UpDownCounter](./api.md#updowncounter) are additive while [Asynchronous
+Gauge](./api.md#asynchronous-gauge) is non-additive.
+
+#### Numeric type selection
+
+For Instruments which take increments and/or decrements as the input (e.g.
+[Counter](./api.md#counter) and [UpDownCounter](./api.md#updowncounter)), the
+underlying numeric types (e.g., signed integer, unsigned integer, double) have
+direct impact on the dynamic range, precision, and how the data is interpreted.
```
noahfalk, line 124: Can we codify how the implementers choice of datatype impacts what values can be stored for intermediate aggregates, on the wire, and long-term persistent aggregates? We are telling them there is impact and the examples give suggestions of what the impacts might be, but extrapolating these examples still left me with ambiguity. For example what parts of these examples are guarantees that all OpenTelemetry compliant implementations would/would not overflow and what parts were implementation specific choices how one particular SDK and backend might behave?

specification/metrics/supplementary-guidelines.md
```diff
@@ -79,6 +82,139 @@ Here is one way of choosing the correct instrument:
...
+* During (T<sub>0</sub>, T<sub>1</sub>], we reported `70`.
+* During (T<sub>0</sub>, T<sub>2</sub>], we reported `115`.
+* During (T<sub>0</sub>, T<sub>3</sub>], we reported `116`.
+* During (T<sub>0</sub>, T<sub>4</sub>], we reported `128`.
+* During (T<sub>0</sub>, T<sub>5</sub>], we reported `128`.
+* During (T<sub>0</sub>, T<sub>6</sub>], we reported `173`.
+* ...
+* During (T<sub>0</sub>, T<sub>n+1</sub>], we reported `1,872`.
+* During (T<sub>n+2</sub>, T<sub>n+3</sub>], we reported `35`.
+* During (T<sub>n+2</sub>, T<sub>n+4</sub>], we reported `76`.
```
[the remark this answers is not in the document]
reyang: > However explicit notions of a time range I think only show up in the SDK which makes me think this report is SDK->Exporter or Exporter->Back end?

Yep, `report` here is referring to the cumulative sum from the SDK output.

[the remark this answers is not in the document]
reyang: On Ln132-133 "count the committed transactions in a database, reported as cumulative sum every 15 seconds" - "count" is the action from API perspective (count transactions by calling `counter.Add(increment)`), `report` is the result from the SDK (after aggregation).

specification/metrics/supplementary-guidelines.md
```diff
@@ -79,6 +82,139 @@ Here is one way of choosing the correct instrument:
 
 ### Additive property
 
+In OpenTelemetry a [Measurement](./api.md#measurement) encapsulates a value and
```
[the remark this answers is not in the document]
reyang: There is no direct agreement between instrumenter and the backend, as there are several roles involved - instrumenter (folks who use the API), application owner (who configures the SDK and exporter), backend (time-series database + query engine) and UX (query client + UI).

I think the UI is using the type of the metrics stream (e.g. sum, gauge, histogram, summary, etc.) plus additional metadata (e.g. the name of the metrics stream if it's a well known thing as defined by the semantic convention, unit) to provide the experience.

The way I look at the supplementary guideline - it is trying to provide additional context which might help to give clarity, or help folks to think about things that they haven't thought about previously, or just calling out that something is hard/complex without providing a simple answer.

specification/metrics/supplementary-guidelines.md
reyang: There is no guarantee - as called out in the first line of this doc ("Note: this document is NOT a spec, it is provided to support the Metrics API and SDK specifications, it does NOT add any extra requirements to the existing specifications.")

However, I do think it'll be helpful to give some suggestions (not requirements) to certain situations - e.g. JavaScript might not support int64, Python might treat integers as BigInt.

PR #2586, opened 2022-05-27
Use consistent file name for data-model.md

tigrannajaryan:
We use (mostly) snake-case for file names in the repository.
I renamed datamodel.md to data-model.md to be consistent.


reyang (approved):
LGTM.

specification/README.md
```diff
@@ -22,7 +22,7 @@
 - Data Specification
   - [Semantic Conventions](overview.md#semantic-conventions)
   - [Protocol](protocol/README.md)
-    - [Metrics](metrics/datamodel.md)
+    - [Metrics](metrics/data-model.md)
     - [Logs](logs/data-model.md)
```
arminru, line 26: Same thing here.

arminru (commented):
Do you think we might have any incoming links that would break?

arminru:
@chalin are there any changes that would need to be reflected on the website following this renaming?

carlosalberto:
LGTM (prior to a review from @chalin as it may impact the docs in the OTel website).

Other than that, we should have an 'official' suggestion somewhere saying "this-is-my-file.md" format is preferred (and also consider updating other related items, e.g. should `semantic_conventions/` become `semantic-conventions/`?

tigrannajaryan:
> Do you think we might have any incoming links that would break?

It is possible. I am not sure there is a way to prevent it.

reyang:
> > Do you think we might have any incoming links that would break?
> 
> It is possible. I am not sure there is a way to prevent it.

One possible way is to keep the existing file but shrink it to:
```markdown
This file has been moved to ...
```

Anyways I think we shouldn't try to solve this problem. GitHub has permanent links, if folks use a non-permanent link, the default expectation is that it could break.

tigrannajaryan:
@chalin can you please confirm that we are not breaking anything at opentelemetry.io by this renaming?

bogdandrutu (approved):

chalin:
> @chalin looks like you are Away/OOO. Will merge the PR now, please follow up after you're back, thanks!

Thanks for the original pings and apologies for the delayed response. Yes, a path aliases needed to be added, which is done in:

- #2870

PR #2594, opened 2022-06-01
Adopt attribute requirement levels in semantic conventions

lmolkova:
Updates existing conventions to use attribute requirement levels introduced in #2522


semantic_conventions/trace/database.yaml
```diff
@@ -294,8 +294,8 @@ groups:
       - id: table
         type: string
         tag: call-level-tech-specific-cassandra
-        required:
-          conditional: Recommended if available.
+        requirement_level:
+          recommended: When available.
```
Oberon00, line 298: IMHO we can simplify this to

```suggestion
        requirement_level: recommended
```

since this is the default for "recommended".

(and other locations)

Oberon00 (approved):

bogdandrutu (commented):
Reading this PR, just scares me how many things are "required". Do we need an audit of these, or we already checked that we really "require" this?

One last thing, in a separate PR: Can we document if changing "requirement_levels" is considered backwards compatible, and if yes between which levels?

lmolkova:
> Reading this PR, just scares me how many things are "required". Do we need an audit of these, or we already checked that we really "require" this?

across all specs we only  have 26 attributes that are required (some are duplicated in traces and metrics):

Most of `required` attributes look reasonable to me, I created issues where I believe we can relax the level:
- resources: os.type, serivce.name, faas.name, webengine.name
- span:
  - http: method
  - rpc: method, status code, technology
  - cloudevents - can be relaxed - https://github.com/open-telemetry/opentelemetry-specification/issues/2600
  - db: system and name
  - messaging: system. destination - under [rewrite](https://github.com/open-telemetry/oteps/blob/8e663f0b0d9d8b8ad0372797f844058ace4643de/text/trace/0192-messaging-semantic-conventions-spec.md#attributes) anyway
    - RocketMq extensions requires a few attributes
  - faas - needs revision - https://github.com/open-telemetry/opentelemetry-specification/issues/2599
    - defines several required attributes for specific trigger

We also have ~25 `conditionally_required` attributes which seem reasonable.

lmolkova:
> One last thing, in a separate PR: Can we document if changing "requirement_levels" is considered backwards compatible, and if yes between which levels?

Great point, here it is https://github.com/open-telemetry/opentelemetry-specification/pull/2601

lmolkova:
@bogdandrutu let me know if you see #2600 and #2599 are preconditions without which this PR can't be merged. 
From my side, I'd prefer to merge this one first, and then I'll send small follow-up PRs for the above issues.

lmolkova:
@open-telemetry/specs-metrics-approvers can you please take a look?

bogdandrutu (approved):
Discuss that @lmolkova will review "required" attributes in a different PR.

lmolkova:
@jack-berg @lzchen @MrAlias @cijothomas @justinfoote @jkwatson can someone please review on behalf of @open-telemetry/specs-metrics-approvers ?

MrAlias (approved):

semantic_conventions/trace/instrumentation/aws-sdk.yml
```diff
@@ -31,7 +31,7 @@ groups:
     attributes:
       - ref: db.system
         brief: "The value `dynamodb`."
-        required: always
```
jack-berg, line 34: I noticed that in this file there are are attributes listed below for `dynamodb.shared` which don't contain a `required` field. Is there a default requirement level for these?

semantic_conventions/trace/database.yaml
```diff
@@ -215,20 +215,20 @@ groups:
         examples: ['findAndModify', 'HMSET', 'SELECT']
       - ref: net.peer.name
         tag: connection-level
-        required:
-          conditional: See below.
+        requirement_level:
+          conditionally_required: See below.
```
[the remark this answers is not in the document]
Oberon00: Something that is below in the markdown text from which the table is referenced. 😕  Not optimal, but not changed in this PR.

semantic_conventions/trace/instrumentation/aws-sdk.yml
Oberon00: Previously, the default was "not required". I think now it is "recommended", right @lmolkova?
I think there is an argument for changing the semantic convention generator to require specifying a requirement level.

semantic_conventions/resource/process.yaml
```diff
@@ -11,35 +11,35 @@ groups:
         examples: [1234]
       - id: executable.name
         type: string
-        required:
-          conditional: "See below"
+        requirement_level:
+          conditionally_required: See below
```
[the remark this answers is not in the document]
lmolkova: I agree it seems strange, the way it works is that there is a paragraph in the md file that talks about different attribute alternatives and covers multiple attributes:

> At least one of `process.executable.name`, `process.executable.path`, `process.command`, `process.command_line` or `process.command_args` is required to allow back ends to identify the executable.

While applied to all of these attributes. 

We use `constraints.any_of` in other specs to express when one of the attributes is required, so I changed process yaml to also leverage `any_of` and the condition to be a bit more explicit. Hope it helps.

semantic_conventions/trace/instrumentation/aws-sdk.yml
lmolkova: Yes, this is right, `recommended` is the default one and I believe the semantics of it didn't change: it's the attribute instrumentation SHOULD provide by default. 

@Oberon00 I'm not sure about requiring to specify required level - md files will have "Recommended" explicitly generated there, so I don't see a good reason to, but happy to learn if there are any.

semantic_conventions/trace/database.yaml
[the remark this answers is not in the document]
lmolkova: updated to `See alternative attributes below.`, In markdown it looks like this:

!-- semconv db(tag=connection-level) -->
|---|---|---|---|---|
| [`net.peer.ip`](span-general.md) | string | Remote address of the peer (dotted decimal for IPv4 or [RFC5952](https://tools.ietf.org/html/rfc5952) for IPv6) | `127.0.0.1` | Conditionally Required: See alternative attributes below. |
| [`net.peer.name`](span-general.md) | string | Remote hostname or similar, see note below. [1] | `example.com` | Conditionally Required: See alternative attributes below. |

**[1]:** `net.peer.name` SHOULD NOT be set if capturing it would require an extra DNS lookup.

....

**Additional attribute requirements:** At least one of the following sets of attributes is required:

* [`net.peer.name`](span-general.md)
* [`net.peer.ip`](span-general.md)

reyang:
> > @jack-berg @lzchen @MrAlias @cijothomas @justinfoote @jkwatson can someone please review on behalf of @open-telemetry/specs-metrics-approvers ?
> 
> I need to be taken off that list.

@jkwatson I've removed you from the list. THANK YOU for helping metrics API/SDK specs to the shape!

reyang (approved):

reyang (approved):

cijothomas (approved):

lmolkova:
@open-telemetry/technical-committee can someone please merge?

PR #2602, opened 2022-06-02
Add semantic convention attributes for capturing the application layer protocol

pyohannes:
Fixes https://github.com/open-telemetry/opentelemetry-specification/issues/2548

### Changes

This PR proposed the two additional attributes `net.app.protocol.name` and `net.app.protocol.version`. These allow capturing the name and version of the application layer protocol used (in addition to the name of the transport protocol captured in net.transport). The naming is consistent with ECS, where [network.protocol](https://www.elastic.co/guide/en/ecs/current/ecs-network.html#field-network-protocol) is used.

Those general attributes can be used to capture information about application layer protocols for which no dedicated semantic conventions exist (e. g. AMQT, or MQTT).

@Oberon00 proposed using a new [`net.app` namespace for this](https://github.com/open-telemetry/opentelemetry-specification/issues/2548#issuecomment-1140923341).

### DIFF
--- a/semantic_conventions/trace/general.yaml
+++ b/semantic_conventions/trace/general.yaml
@@ -33,6 +33,14 @@ groups:
               brief: 'Something else (non IP-based).'
         brief: >
           Transport protocol used. See note below.
+      - id: protocol.name
+        type: string
+        brief: 'Application layer protocol used. The value SHOULD be normalized to lowercase.'
+        examples: ['amqp', 'http', 'nntp']
+      - id: protocol.version
+        type: string
+        brief: 'Version of the application layer protocol used.'
+        examples: '0.9.1'
       - id: peer.ip
         type: string
         brief: >
--- a/semantic_conventions/trace/general.yaml
+++ b/semantic_conventions/trace/general.yaml
@@ -33,14 +33,14 @@ groups:
               brief: 'Something else (non IP-based).'
         brief: >
           Transport protocol used. See note below.
-      - id: protocol.name
+      - id: app.protocol.name
         type: string
         brief: 'Application layer protocol used. The value SHOULD be normalized to lowercase.'
-        examples: ['amqp', 'http', 'nntp']
-      - id: protocol.version
+        examples: ['amqp', 'http', 'mqtt']
+      - id: app.protocol.version
         type: string
         brief: 'Version of the application layer protocol used.'
-        examples: '0.9.1'
+        examples: '3.1.1'
       - id: peer.ip
         type: string
         brief: >
--- a/semantic_conventions/trace/general.yaml
+++ b/semantic_conventions/trace/general.yaml
@@ -39,8 +39,12 @@ groups:
         examples: ['amqp', 'http', 'mqtt']
       - id: app.protocol.version
         type: string
-        brief: 'Version of the application layer protocol used.'
+        brief: 'Version of the application layer protocol used. See note below.'
         examples: '3.1.1'
+        note: >
+          `net.app.protocol.version` refers to the version of the protocol used and might be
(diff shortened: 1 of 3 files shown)


semantic_conventions/trace/general.yaml
```diff
@@ -33,6 +33,14 @@ groups:
               brief: 'Something else (non IP-based).'
         brief: >
           Transport protocol used. See note below.
+      - id: app.protocol.name
+        type: string
+        brief: 'Application layer protocol used. The value SHOULD be normalized to lowercase.'
+        examples: ['amqp', 'http', 'mqtt']
+      - id: app.protocol.version
+        type: string
+        brief: 'Version of the application layer protocol used.'
```
Oberon00, line 42:
```suggestion
        brief: 'Version of the application layer protocol used (might be different from the protocol client's version).'
```

Not sure if that adds clarity or confusion though, your call.

Oberon00 (approved):

semantic_conventions/trace/general.yaml
pyohannes: It took me some time to understand the motivation behind it, but if other reviewers think it adds clarity, I'm fine adding it.

semantic_conventions/trace/general.yaml
```diff
@@ -33,6 +33,14 @@ groups:
               brief: 'Something else (non IP-based).'
         brief: >
           Transport protocol used. See note below.
+      - id: app.protocol.name
+        type: string
+        brief: 'Application layer protocol used. The value SHOULD be normalized to lowercase.'
+        examples: ['amqp', 'http', 'mqtt']
+      - id: app.protocol.version
```
lmolkova, line 40: would this mean that `http.flavor` needs to be removed in favor of a combination of `net.app.protocol.name` = `http`
and `net.app.protocol.version` = what was in `http.flavor` ?

it can happen as a separate PR along with schema attribute renames, but I'd like to understand the direction.

lmolkova (approved):

semantic_conventions/trace/general.yaml
pyohannes: @Oberon00 suggested to first introduce the new `app.protocol.*` attributes, and then discuss replacing existing attributes in separate PRs/issues.

So, I think the `app.protocol.*` attributes can be used to replace `http.flavor`, but that should be proposed and discussed separately once this PR is merged.

semantic_conventions/trace/general.yaml
blumamir: I am not sure what you meant, is "protocol client" the package that is used as the client for this operation?
For example, [axios in npm](https://www.npmjs.com/package/axios/v/0.27.2) is currently version 0.27.2 but the http it produces is 1.1, so this value should be `1.1` and not `0.27.2`?

specification/trace/semantic_conventions/span-general.md
```diff
@@ -35,6 +35,8 @@ the `net.peer.*` properties of a client are equal to the `net.host.*` properties
 | Attribute  | Type | Description  | Examples  | Required |
```
arminru, line 35: Please add an entry in CHANGELOG.md

arminru (approved):

semantic_conventions/trace/general.yaml
arminru: @pyohannes To make it easier to understand you could also make it as `note` where you can add a more detailed explanation and maybe an example like @blumamir brought up.

specification/trace/semantic_conventions/span-general.md
pyohannes: I added a changelog entry.

reyang (approved):

PR #2604, opened 2022-06-02
Add rpc.grpc.status_code attribute to rpc metric semantic conventions

### LINKED ISSUE #2603: Include rpc.grpc.status_code attributes in rpc metrics

I'd like to see a breakdown of rpc metrics by status code, but the rpc metric semantic conventions [attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/rpc.md#attributes) don't contain any sort of indication of the status of the rpc, akin to `http.status_code`. 

I expected to see grpc have `rpc.grpc.status_code` for symmetry with the [grpc trace semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/rpc.md#grpc-attributes).

jack-berg:
Resolves #2603.


specification/metrics/semantic_conventions/rpc.md
```diff
@@ -112,4 +113,38 @@ For remote procedure calls via [gRPC][], additional conventions are described in
...
+| `11` | OUT_OF_RANGE        |
+| `12` | UNIMPLEMENTED       |
+| `13` | INTERNAL            |
+| `14` | UNAVAILABLE         |
+| `15` | DATA_LOSS           |
+| `16` | UNAUTHENTICATED     |
+<!-- endsemconv -->
+
+[gRPC]: https://grpc.io/
+
```
arminru, line 149:
```suggestion
```
duplicate

specification/metrics/semantic_conventions/rpc.md
```diff
@@ -3,9 +3,10 @@
 **Status**: [Experimental](../../document-status.md)
```
arminru, line 3: Please add an entry in CHANGELOG.md

specification/metrics/semantic_conventions/rpc.md
```diff
@@ -112,4 +113,38 @@ For remote procedure calls via [gRPC][], additional conventions are described in
...
 `rpc.system` MUST be set to `"grpc"`.
 
+### gRPC Attributes
+
+Below is a table of attributes that SHOULD be included on client and server RPC measurements when `rpc.system` is `"grpc"`.
+
+<!-- semconv rpc -->
+| Attribute                                                                         | Type | Description                                                                                                  | Examples | Required |
+|-----------------------------------------------------------------------------------|------|--------------------------------------------------------------------------------------------------------------|----------|----------|
+| [`rpc.grpc.status_code`](../../trace/semantic_conventions/rpc.md#grpc-attributes) | int  | The [numeric status code](https://github.com/grpc/grpc/blob/v1.33.2/doc/statuscodes.md) of the gRPC request. | `0`      | Yes      |
```
trask, line 123: interesting, I was surprised to notice this is required (and that it is required on spans too) as we don't capture `rpc.grpc.status_code` in Java instrumentation when an exception is thrown. any idea if we are doing something wrong, or if we should request to change this from required to non-required? (or conditionally required if its available)

trask (approved):

specification/metrics/semantic_conventions/rpc.md
jack-berg: Yeah that's a good point. `rpc.grpc.status_code` is at the same level of importance as `http.status_code`, which is [optional](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md#attributes) for metrics, and [required](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#common-attributes) for tracing "if and only if one was received". 

In either case, we should capture it in java instrumentation. 

I'm favor of reducing the requirement to optional for alignment with http.

jack-berg: On second thought, `http.status_code` is required if one was sent / received. In gRPC, my experience is that a status code is always returned, regardless of network errors that may prevent the underlying http response from being returned. In these cases I believe [status code](https://grpc.github.io/grpc/core/md_doc_statuscodes.html) `UNKNOWN` is often used.

PR #2614, opened 2022-06-10
Define net.sock attributes and clarify logical net.peer|host.name attributes

### LINKED ISSUE #2607: Should net.peer.* describe socket-level peer or "logical" peer (for HTTP and in general)?
(from discussion with @lmolkova in https://github.com/open-telemetry/opentelemetry-specification/pull/2469#issuecomment-1145539176)

In the HTTP protocol, the peer you connect to can be unclear if an HTTP-level proxy is used. In that case, an application opens a TCP connection to the configured proxy host and port, and sends an n HTTP request with a full URL as target to it (`GET http://hostname.example.com/some?thing=1 HTTP/1.1`).

In this case, it is not clear whether net.peer.name should be the hostname from the requested URL (hostname.example.com) or of the HTTP proxy the application sends the HTTP request to.

One thing that seems clear is that neet.peer.name and net.peer.ip should be consistent which each other. If net.peer.name is set to a logical peer distinct from the socket-level peer, net.peer.ip would have to be left unset and there would currently be no way to store network-level connection information.

There is also an issue with gRPC which (usually?) goes over HTTP/2, but the RPC conventions *require* at least one of net.peer.ip or net.peer.name. Clearly, in absence of any other attribute to hold it, the destination hostname would be more useful than the hostname of an HTTP proxy.

An open question to me is, whether this problem is specific to HTTP (and HTTP-based protocols), or if this is common also in other protocols.

* Option 1: Leave net.peer.* ambiguous, make it deprecated. Depending on the semantic convention used, it might be the logical peer, the socket-level peer or (usually) both. Introduce separate `net.sock.*` and `net.app.*` namespaces to have unambiguous options.
* Option 2: Define net.peer.* to be either one or the other, introduce only the other namespace.
* Option 3: Try to stay as close to the status quo as possible: Define net.peer.* to be the network-level peer. Protocols that have a logical peer canonically identified by a DNS-hostname that might be different from the network-level peer should have a special attribute that defines the role of this peer (http.host which already exists, rpc.endpoint, db.server_name, ...)
* Option 4: ("Do nothing") Leave net.peer.* ambigous, accept that only one of network-level or logical peer can be set on a span in most semantic conventions (e.g. HTTP server conventions have both net.peer.ip and http.client_ip, while rpc client conventions only have net.peer.ip/name)

Option 1 would be the easiest. Option 3 would be the most complex, also for processing on the backend, but would be future-proof also in case some protocol defines multiple peers with different roles that you might connect to with the same span. That seems unlikely, so I don't think Option 3 is very useful.

Option 2 is similar to Option 1 and would be preferable if we find that net.peer.* is overwhelmingly used for either network-level or logical peer information, instead of being used for both in different instrumentations/languages. Choosing option 2 would need a survey of some sort to find how feasible it is and choose one or the other option.

lmolkova:
Fixes #2607

## Changes

1. Introduces socket-level attributes: 
   - BREAKING: `net.peer.ip` ->  `net.sock.peer.addr`, `net.host.ip` ->  `net.sock.host.addr`
   - BREAKING: for pure socket-level instrumentation, `net.peer.name` is renamed to `net.sock.peer.addr`. For other instrumentations, `net.peer.name` semantics didn't change.
3. Clarifies `net.peer|host.name`  and port attributes as logical ones representing final destination.
4. Clarifies socket-level proxy scenario

Related issues #2469

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ release.
 - Add experimental `OTEL_EXPORTER_OTLP_DEFAULT_HISTOGRAM_AGGREGATION` variable for
   configuring default histogram aggregation of OTLP metric exporter
   ([#2619](https://github.com/open-telemetry/opentelemetry-specification/pull/2619)).
+- Clarify async instrument callback identity.
+  ([#2538](https://github.com/open-telemetry/opentelemetry-specification/pull/2538)).
 
 ### Logs
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,6 @@ release.
 - Add `http.*.*.size` metric semantic conventions for tracking size of requests
   / responses for http servers / clients
   ([#2588](https://github.com/open-telemetry/opentelemetry-specification/pull/2588)).
-- BREAKING: rename `net.peer.ip` to `net.sock.peer.addr` and `net.host.ip` to `net.sock.host.addr`.
 - BREAKING: rename `net.peer.ip` to `net.sock.peer.addr`, `net.host.ip` to `net.sock.host.addr`,
   `net.peer.name` to `net.sock.peer.name` for socket-level instrumentation.
   Define socket-level attributes and clarify logical peer and host attributes meaning.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,9 @@ release.
 - Add `http.*.*.size` metric semantic conventions for tracking size of requests
   / responses for http servers / clients
   ([#2588](https://github.com/open-telemetry/opentelemetry-specification/pull/2588)).
+- BREAKING: rename `net.peer.ip` to `net.sock.peer.addr` and `net.host.ip` to `net.sock.host.addr`.
+  Define socket-level attributes and clarify logical peer and host attributes meaning.
+  ([#2594](https://github.com/open-telemetry/opentelemetry-specification/pull/2594))
 
 ### Compatibility
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,8 @@ release.
   / responses for http servers / clients
   ([#2588](https://github.com/open-telemetry/opentelemetry-specification/pull/2588)).
 - BREAKING: rename `net.peer.ip` to `net.sock.peer.addr` and `net.host.ip` to `net.sock.host.addr`.
+- BREAKING: rename `net.peer.ip` to `net.sock.peer.addr`, `net.host.ip` to `net.sock.host.addr`,
+  `net.peer.name` to `net.sock.peer.name` for socket-level instrumentation.
   Define socket-level attributes and clarify logical peer and host attributes meaning.
   ([#2594](https://github.com/open-telemetry/opentelemetry-specification/pull/2594))
 
--- a/schemas/1.12.0
+++ b/schemas/1.12.0
@@ -2,6 +2,12 @@ file_format: 1.0.0
 schema_url: https://opentelemetry.io/schemas/1.12.0
 versions:
   1.12.0:
+    spans:
+      changes:
+        - rename_attributes:
+            attribute_map:
+              net.peer.ip: net.sock.peer.addr
+              net.host.ip: net.sock.host.addr
   1.11.0:
   1.10.0:
   1.9.0:
--- a/semantic_conventions/trace/general.yaml
+++ b/semantic_conventions/trace/general.yaml
@@ -103,7 +103,9 @@ groups:
         examples: '192.168.0.1'
       - id: sock.host.port
         type: int
-        brief: 'Local socket peer port (if defined for the address family).'
+        brief: 'Local socket port number.'
+        requirement_level:
+          recommended: If defined for the address family and if different than `net.host.port` and if `net.sock.host.addr` is set.
         examples: 35555
       - id: host.connection.type
         type:
--- a/semantic_conventions/trace/general.yaml
+++ b/semantic_conventions/trace/general.yaml
@@ -6,18 +6,12 @@ groups:
     attributes:
       - id: transport
         type:
-          allow_custom_values: false
+          allow_custom_values: true
           members:
             - id: ip_tcp
               value: "ip_tcp"
             - id: ip_udp
               value: "ip_udp"
-            - id: ip
-              value: "ip"
-              brief: 'Another IP-based protocol'
-            - id: unix
-              value: "unix"
-              brief: 'Unix Domain socket. See below.'
             - id: pipe
               value: "pipe"
               brief: 'Named or anonymous pipe. See note below.'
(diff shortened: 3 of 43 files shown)


lmolkova:
@Oberon00 I made the first stab at `net.sock` attributes. 

One thing I didn't figure out yet is the relationship between `net.transport` and `net.sock.family`. 

Transport currently is a closed set, address families are not (and different sets on win and linux).
Transport defines TCP and UDP, address family differentiates IPv4 and IPv6.
Transport has `pipe`  and it's not a socket.

I'd propose the following on '`net.transport`:
- remove or deprecate `net.transport` `ip` and `unix` values
- add `socket` value

Then `net.sock.family` can be used to understand if it's an IP-based protocol, Unix, or perhaps Bluetooth when `net.transport = socket`. Or if addr family is `AF_INET6`,  `net.transport` explains if it tcp or udp. 

LMK what you think.

Oberon00:
I didn't have time to think about the net.transport question yet, but I noticed one thing:

There is no replacement for net.peer.name yet, and also the definition in the "net.*.name" section has not changed yet to address the logical peer change:

> For net.peer.name, this should be the name of logical remote destination that was used to look up the IP address that was connected to (i.e., matching net.peer.ip if that one is set)

I also think that net.peer.ip does not make sense anymore, there should probably only be net.sock.peer.ip.

lmolkova:
> I also think that net.peer.ip does not make sense anymore, there should probably only be net.sock.peer.addr.

I think we wanted to use it instead of `http.client_ip`, but thinking more about it, perhaps reverse proxy and `x-forwarded-for` and friends are so HTTP-specific that `http.client_ip` would make more sense than generic `net.peer.ip`.

lmolkova:
@Oberon00 I think I addressed your feedback. I'm not sure if renaming `net.peer.name` is necessary - even though we work on experimental spec, we should minimize breaking changes. Currently, the breaking change is minimal (`net.peer.ip` -> `net.sock.peer.addr` and `net.host.ip` ->  `net.sock.host.addr`) and spec is clear on socket vs logical attributes plus proxy scenarios,

Oberon00:
I'm not talking about renaming of net.peer.name, but about adding an additional attribute, e.g. net.sock.peer.name that would then correspond to net.sock.peer.addr (net.peer.name does not necessarily do that anymore).

lmolkova:
@Oberon00  what would `net.sock.peer.name` mean - DNS name corresponding for `net.sock.peer.addr` ? Like proxy host name if one's available?

Oberon00:
> @Oberon00 what would net.sock.peer.name mean - DNS name of net.sock.peer.addr ? Like proxy host name if one's available?

Exactly. This is how I understood the net.peer.ip/name attributes originally. net.sock.peer.name would then be mostly IP-specific, just like net.peer.ip currently is.

As I understand it, this PR will cause the following "reshuffling":

* net.peer.name: Logical peer name, but not of the peer service (which is peer.service) but of something that could theoretically be "connected" to, typically a DNS-resolvable hostname  (even if just from some private network), but might be something else. Depending on the address family used, a low-level 2non-logical" net.peer.name becomes net.peer.sock.addr or net.peer.sock.name.
* net.peer.ip -> net.peer.sock.addr, always, and only available for low-level peer, not for logical one if different.

lmolkova:
@Oberon00 thank you for the feedback! I believe I addressed it and this PR now is ready for review.

CHANGELOG.md
```diff
@@ -23,6 +23,9 @@ release.
   ([#2602](https://github.com/open-telemetry/opentelemetry-specification/pull/2602)).
 - Adopt attribute requirement levels in semantic conventions
   ([#2594](https://github.com/open-telemetry/opentelemetry-specification/pull/2594))
+- BREAKING: rename `net.peer.ip` to `net.sock.peer.addr` and `net.host.ip` to `net.sock.host.addr`.
```
Oberon00, line 26:
```suggestion
- BREAKING: rename `net.peer.ip`, and `net.peer.name` in some usages, to `net.sock.peer.addr` and `net.host.ip` to `net.sock.host.addr`.
```

semantic_conventions/trace/general.yaml
```diff
@@ -48,31 +42,67 @@ groups:
...
-        brief: 'Remote port number.'
-        examples: [80, 8080, 443]
+        brief: Remote socket peer port (if defined for the address family).
+        examples: 16456
+      - id: sock.family
+        type:
+          allow_custom_values: true
+          members:
+            - id: inet
+              value: 'AF_INET'
```
Oberon00, line 67:
```suggestion
              value: 'inet'
```

If we want to reference the Unix C define name, we can add it to brief.

semantic_conventions/trace/general.yaml
```diff
@@ -48,31 +42,67 @@ groups:
...
+              brief: "IPv4 address"
+            - id: inet6
+              value: 'AF_INET6'
```
Oberon00, line 70:
```suggestion
              value: 'inet6'
```

semantic_conventions/trace/general.yaml
```diff
@@ -48,31 +42,67 @@ groups:
...
+              brief: "IPv6 address"
+            - id: unix
+              value: 'AF_UNIX'
```
Oberon00, line 73:
```suggestion
              value: 'unix'
```

semantic_conventions/trace/general.yaml
```diff
@@ -48,31 +42,67 @@ groups:
...
+              brief: "Unix domain socket path/"
+        requirement_level:
+          conditionally_required: if different than `AF_INET` and if any of `net.sock.peer.addr` or `net.sock.host.addr` are set
```
Oberon00, line 76:
```suggestion
          conditionally_required: if different than `inet` and if any of `net.sock.peer.addr` or `net.sock.host.addr` are set
```

lmolkova:
@Oberon00 I believe I resolved your feedback, can you please take another look?

Oberon00 (approved):
I wonder if we also need net.host.sock.*?

lmolkova:
@Oberon00 

> I wonder if we also need net.sock.host.*?

I defined `net.sock.host.addr` and  `net.sock.host.port` and briefly described them [here](https://github.com/open-telemetry/opentelemetry-specification/blob/39ec7eeb6f6804dd8294864d0873171d5eb73994/specification/trace/semantic_conventions/span-general.md#host) Do you think something important is missing?

semantic_conventions/trace/general.yaml
```diff
@@ -6,18 +6,12 @@ groups:
     attributes:
```
Oberon00, line 6: I think at this point it would make sense to split net.* into a separate file (also the markdown)

semantic_conventions/trace/http.yaml
```diff
@@ -114,10 +114,17 @@ groups:
...
           recommended: If and only if a request was retried.
         examples: 3
-      - ref: net.peer.ip
-        sampling_relevant: true
-      - ref: net.peer.port
-        sampling_relevant: true
+      - ref: net.sock.peer.addr
+      - ref: net.sock.peer.port
+        requirement_level:
+          recommended: If different than `net.peer.port` and if `net.sock.peer.addr` is set.
```
Oberon00, line 120: I think this (and similar ones) should go to the primary definition.

semantic_conventions/trace/general.yaml
```diff
@@ -48,31 +42,67 @@ groups:
...
+              value: 'inet'
+              brief: "IPv4 address"
+            - id: inet6
+              value: 'inet6'
+              brief: "IPv6 address"
+            - id: unix
+              value: 'unix'
+              brief: "Unix domain socket path/"
+        requirement_level:
+          conditionally_required: if different than `inet` and if any of `net.sock.peer.addr` or `net.sock.host.addr` are set
```
Oberon00, line 76: I think this requirement is incompatible with the schema. Previously, we did not differentiate inet and inet6 in net.peer.ip. Now if the schema renames this to net.peer.sock.addr, and it contains an IPv6, then this would result in bad data.

I don't know how to handle this nicely, but some not so nice alternatives:
- Only have `inet`, rely on addr format to distinguish. If no addr is set, then we don't know.
- Add inet_any.
- Specify that if not set, inet6 vs inet must be determined by looking at the address

Maybe the last one would be quite alright:
```suggestion
          conditionally_required: if neither `inet` nor `inet6` and if any of `net.sock.peer.addr` or `net.sock.host.addr` are set.
```

semantic_conventions/trace/general.yaml
lmolkova: Yes, it makes sense, Can I do it in a separate PR? I want to unblock the HTTP - it's been open for a while.

semantic_conventions/trace/http.yaml
lmolkova: Apparently, requirement level and msg references don't work - fixing it here https://github.com/open-telemetry/build-tools/pull/102

lmolkova:
@carlosalberto can you please take a look?

lmolkova:
@open-telemetry/specs-approvers can someone please take a look? It unblocks https://github.com/open-telemetry/opentelemetry-specification/pull/2469

semantic_conventions/trace/http.yaml
Oberon00: IIRC this was originally intentional, but I don't see any benefit of that design now.

specification/metrics/semantic_conventions/http-metrics.md
```diff
@@ -52,7 +52,7 @@ and whether they should be on server, client, or both types of HTTP metric event
...
 | `net.peer.port`    | `client`            | see [1] in [attribute alternatives](#attribute-alternatives) | See [general network connection attributes](../../trace/semantic_conventions/span-general.md#general-network-connection-attributes)                                                                                 |
-| `net.peer.ip`      | `client`            | see [1] in [attribute alternatives](#attribute-alternatives) | See [general network connection attributes](../../trace/semantic_conventions/span-general.md#general-network-connection-attributes)                                                                                 |
+| `net.sock.peer.addr`      | `client`            | see [1] in [attribute alternatives](#attribute-alternatives) | See [general network connection attributes](../../trace/semantic_conventions/span-general.md#general-network-connection-attributes)                                                                                 |
```
trask, line 55: attribute alternatives is referenced here, but ip/addr was deleted from attribute alternatives below

specification/trace/semantic_conventions/database.md
```diff
@@ -42,21 +42,28 @@ Some database systems may allow a connection to switch to a different `db.user`,
...
 | `db.user` | string | Username for accessing the database. | `readonly_user`; `reporting_user` | Recommended |
-| [`net.peer.ip`](span-general.md) | string | Remote address of the peer (dotted decimal for IPv4 or [RFC5952](https://tools.ietf.org/html/rfc5952) for IPv6) | `127.0.0.1` | Conditionally Required: See alternative attributes below. |
-| [`net.peer.name`](span-general.md) | string | Remote hostname or similar, see note below. [1] | `example.com` | Conditionally Required: See alternative attributes below. |
-| [`net.peer.port`](span-general.md) | int | Remote port number. | `80`; `8080`; `443` | Conditionally Required: [2] |
-| [`net.transport`](span-general.md) | string | Transport protocol used. See note below. | `ip_tcp` | Conditionally Required: [3] |
+| [`net.peer.name`](span-general.md) | string | Name of the database host. [1] | `example.com` | Conditionally Required: See alternative attributes below. |
+| [`net.peer.port`](span-general.md) | int | Logical remote port number | `80`; `8080`; `443` | Conditionally Required: [2] |
```
trask, line 46: are these only for client now? if so, maybe we should add column to the table to identify client and/or server similar to http metrics attributes table:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md#attributes

semantic_conventions/trace/general.yaml
```diff
@@ -45,34 +39,72 @@ groups:
...
+        type: int
+        brief: 'Logical local port number, preferably the one that the peer used to connect'
+        examples: 8080
+      - id: sock.host.addr
+        type: string
+        brief: Local socket address. Useful in case of a multi-IP host.'
+        examples: '192.168.0.1'
+      - id: sock.host.port
+        type: int
+        brief: 'Local socket peer port (if defined for the address family).'
```
trask, line 106:
```suggestion
        brief: 'Local socket port number.'
        requirement_level:
          recommended: If defined for the address family and if different than `net.host.port` and if `net.sock.host.addr` is set.
```

trask (approved):
fyi we've prototyped this change in Java instrumentation: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/6268

lmolkova:
@trask thanks a lot for the feedback and the prototype!

@carlosalberto, can you please take another look?  Thanks!

jmacd (approved):

jmacd:
@open-telemetry/specs-approvers Last call for feedback.

specification/trace/semantic_conventions/span-general.md
```diff
@@ -37,12 +43,17 @@ the `net.peer.*` properties of a client are equal to the `net.host.*` properties
...
 | `net.app.protocol.version` | string | Version of the application layer protocol used. See note below. [1] | `3.1.1` | Recommended |
-| `net.peer.ip` | string | Remote address of the peer (dotted decimal for IPv4 or [RFC5952](https://tools.ietf.org/html/rfc5952) for IPv6) | `127.0.0.1` | Recommended |
-| `net.peer.port` | int | Remote port number. | `80`; `8080`; `443` | Recommended |
-| `net.peer.name` | string | Remote hostname or similar, see note below. [2] | `example.com` | Recommended |
-| `net.host.ip` | string | Like `net.peer.ip` but for the host IP. Useful in case of a multi-IP host. | `192.168.0.1` | Recommended |
-| `net.host.port` | int | Like `net.peer.port` but for the host port. | `35555` | Recommended |
-| `net.host.name` | string | Local hostname or similar, see note below. | `localhost` | Recommended |
+| `net.sock.peer.name` | string | Remote socket peer name. | `proxy.example.com` | Recommended: [2] |
+| `net.sock.peer.addr` | string | Remote socket peer address (IPv4 or IPv6 for internet protocols, path for local communication,
```
tigrannajaryan, line 47: The formatting of this line is broken. Check the rendered markdown.

lmolkova: great catch, fixed!

schemas/1.12.0
```diff
@@ -2,6 +2,12 @@ file_format: 1.0.0
 schema_url: https://opentelemetry.io/schemas/1.12.0
 versions:
   1.12.0:
+    spans:
+      changes:
+        - rename_attributes:
+            attribute_map:
+              net.peer.ip: net.sock.peer.addr
+              net.host.ip: net.sock.host.addr
```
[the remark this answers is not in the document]
lmolkova: > I assume the reason why we don't have a line for net.peer.name here is because its semantics didn't change for most case but it was renamed for specific entities only.

Yes and because we didn't have socket-level semantics clearly defined before.

I double-checked the document status on each document changed and all of the conventions are experimental (to be fair, we don't have non-experimental semantic conventions yet).

tigrannajaryan (approved):

reyang (approved):

Issue #2589, opened 2022-05-30
Is direction dimension a mistake in semantic conventions?


tigrannajaryan:
We have metrics that have a "direction" attribute, for example [disk metrics](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/system-metrics.md#systemdisk---disk-controller-metrics).

This seems to be in contradiction to the [general guidelines](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/README.md#general-guidelines) about attributes
>["As a rule of thumb, aggregations over all the attributes of a given metric SHOULD be meaningful,"](https://prometheus.io/docs/practices/naming/#metric-names) as Prometheus recommends.

I am not sure what aggregation is meaningful for "direction" attribute.

Should the attribute be removed and the metrics split into 2, one for read and one for write?

It does look pretty natural to express this dimension as an attribute, it is low cardinality, it groups 2 very related measurements. Is this an indication that the guidelines need to be adjusted or this is an exception from the rule? Or perhaps there is a meaningful aggregation of "direction" that I just don't see?

reyang:
+1, also in most cases the removal of this dimension would result in meaningless value (e.g. what does a "total read/write bytes" mean for a disk?)

In addition, many disk/NICs have extra operations beyond read and write (I just took a screenshot from my laptop):

![image](https://user-images.githubusercontent.com/17327289/171229825-33cf90bb-6d03-4238-8a81-3e62df3c04ba.png)

tigrannajaryan:
> @tigrannajaryan does the aggregation not represent "total" bytes send/received over the network? Is that not meaningful?

Yes, probably meaningful for network (e.g. total bytes/throughput can be interesting to know as a percentage of bidirectional bandwidth available for half-duplex physical connections).

Is it also meaningful for disks? I am not sure what's the meaning of total combined bytes written and read for disk. Sure, both are bytes flowing from/to the same device, maybe that's useful somehow?

I am not against keeping the `direction` if we think it is meaningful (but see also @reyang comment which shows we do not fully understand the allowed values for the dimension).

bertysentry:
I have mixed feelings about this. If you take the network traffic example, you can calculate the bandwidth utilization with something like: `max(rate(system.network.io)) / hw.network.speed`, where the `max()` function applies to each direction of the `system.network.io` metric, for a full-duplex link.

[...]

PR #2617, opened 2022-06-15
add metrics to replace metrics with `direction`

codeboten:
Fixes https://github.com/open-telemetry/opentelemetry-specification/issues/2589

## Changes

This PR looks to address the concerns voiced in #2589. It adds new metrics that include the direction in their names, and marks existing metrics as deprecated.

A few questions I have in mind:
1. Are there other dimensions that may also need to be addressed? I'm not sure if it makes sense to have a `protocol` dimension for network metrics that ends up combining UDP and TCP as one is stateless and the other stateful.
2. Can the change in this PR be added in a schema to assist end users?


tigrannajaryan:
I am not a metrics expert and don't have an opinion on whether this is the right approach, however if we go ahead with this I would like to be able to model this change in the Schema. We need to add a new type of transform in the Schema file, essentially an ability to split a metric into multiple metrics by eliminating a dimension.

I will wait a bit until we get approvals on this PR and if we are going ahead with it I will make the corresponding proposal to add the new transform to Schemas.

Also, @codeboten can you clarify how this impacts the Collector? Do we already emit these metrics in the Collector? If yes, then we are going to change Collector implementation too as a result of this PR.

codeboten:
> Also, @codeboten can you clarify how this impacts the Collector? Do we already emit these metrics in the Collector? If yes, then we are going to change Collector implementation too as a result of this PR.

@tigrannajaryan yes this will impact the hostmetrics receiver. If this change goes it, I will go ahead and make the changes as needed.

Any thoughts on whether a similar change to this one should be proposed around separating `protocol`?

jack-berg (approved):

reyang (approved):

tigrannajaryan:
> Any thoughts on whether a similar change to this one should be proposed around separating `protocol`?

it seems like `protocol` is a valid dimension. I think aggregation of `system.network.connections` by `protocol` is meaningful. You can sum and get the total number of connections regardless of the protocol.

tigrannajaryan:
@dmitryax please check if this change also requires corresponding changes to https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/signalfxexporter

codeboten:
> it seems like `protocol` is a valid dimension. I think aggregation of `system.network.connections` by `protocol` is meaningful. You can sum and get the total number of connections regardless of the protocol.

This is true, but aggregating protocols over `state` would be meaningless, as it would exclude UDP

tigrannajaryan:
> This is true, but aggregating protocols over `state` would be meaningless, as it would exclude UDP. If you wanted to get all connections with the state `open`, you may think you're getting all connections, but really you'd only see tcp connections.

It seems to be even worse than that.

The `state` is barely defined in our spec, it is just a link to Wikipedia article about tcp which is not nearly enough precision for our spec. Depending on how the `state` is properly defined it may or may not be meaningful to aggregate across (`open` is not a valid value even for tcp connections). It probably warrants its own PR.

trask (approved):

dmitryax:
> @dmitryax please check if this change also requires corresponding changes to https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/signalfxexporter

Yes, we need to add another translation rule to combine `system.disk.io.read` and `system.disk.io.write` into `system.disk.io.total`. That shouldn't be complicated

bertysentry:
Is there somewhere an active discussion on this topic? I'd like to see some arguments backing the proposed change. Issue #2589 is merely a question from @tigrannajaryan.

carlosalberto:
@bertysentry We can discuss them tomorrow in the Spec call, which happens at 8am Pacific. Would you be interested in joining?

jmacd (approved):
I support this change to the general semantic conventions provided the schema translation formats and tools are updated to handle conversion. 

My opinion has been influenced by the widespread practice of keeping bidirectional metrics separate, though it does seem like a new semantic convention is emerging (e.g., the use of ".read" and ".write" suffixes).

carlosalberto:
@bertysentry We have it in our Google Calendar, and the call itself happens through Zoom ;)  See https://calendar.google.com/calendar/u/0/r?cid=Z29vZ2xlLmNvbV9iNzllM2U5MGo3YmJzYTJuMnA1YW41bGY2MEBncm91cC5jYWxlbmRhci5nb29nbGUuY29t

reyang:
Discussed during the Jun 28, 2022 Specification SIG Meeting, folks agreed to go with this direction.

tigrannajaryan:
@codeboten now that https://github.com/open-telemetry/opentelemetry-specification/pull/2653 is merged can you please add a schema file 1.13 with the transformation that describes this PR? We need to publish it with the next spec release.

PR #2618, opened 2022-06-16
cloudevents: relax requirement level  for type and version

### LINKED ISSUE #2600: CloudEvents: relax requirement levels
While CloudEvents spec requires most of the required attributes to be present on the event, not all of them may be required for telemetry purposes. 

`event_source` + `event_id` would uniquely identify the event, so let's keep them `required`.

For others requirement level could potentially be relaxed:
~~- [`event_subject` is optional](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#subject) per CloudEvents spec~~
- `event_spec_version` is probably mostly static and rarely changed. Making it `recommended` or `optional` seem to be more reasonable
- `event_type`: can it become `recommended`? 

[Recommended](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-requirement-level.md#recommended) attributes would still be populated by default unless there are strong reasons not to. Instrumentation may provide a way to disable them explicitly.

lmolkova:
Fixes #2600

## Changes

Changes requirement level of `cloudevents.event_spec_version` and `cloudevents.event_type` to recommended. While they are required [per CloudEvents spec](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md) to route and process messages, requiring corresponding attributes on spans is not necessary.

### DIFF
--- a/semantic_conventions/trace/cloudevents.yaml
+++ b/semantic_conventions/trace/cloudevents.yaml
@@ -20,13 +20,11 @@ groups:
         examples: ['https://github.com/cloudevents', '/cloudevents/spec/pull/123', 'my-service' ]
       - id: event_spec_version
         type: string
-        requirement_level: required
         brief: >
           The [version of the CloudEvents specification](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#specversion) which the event uses.
         examples: '1.0'
       - id: event_type
         type: string
-        requirement_level: required
         brief: >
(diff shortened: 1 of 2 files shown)


jmacd (approved):

joaopgrassi (approved):
Thanks @lmolkova !

lmolkova:
@open-telemetry/specs-trace-approvers can someone  please take a look?

arminru (approved):

PR #2619, opened 2022-06-16
Config option enabling exponential histogram as default histogram aggregation

### LINKED ISSUE #2471: Option to make exponential histogram default aggregation for OTLP exporter
Once exponential histograms are a stable part of the metrics SDK, they'll likely be a better default histogram aggregation option than explicit bucket histograms, since they can dynamically adjust to best capture the distribution of measurements that has been recorded.

We should add an option to the [OTLP Metrics Exporter](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk_exporters/otlp.md) to change its default histogram aggregation from explicit bucket to exponential.

Discussed in #2429.

jack-berg:
Resolves #2471.

opentelemetry-java has an exponential histogram implementation, but its currently quite obscure / difficult to use because the exponential histograms aren't yet stable in the spec. As discussed in #2471 and #2429, we removed the "best available histogram" aggregation anticipating later adding an option like this to make it easy for users to opt in to exponential histograms. I'd like to add the configuration option now so that opentelemetry-java users can more easily use exponential histograms, and provide any feedback needed to stabilization of the specification.

### DIFF
--- a/specification/metrics/sdk_exporters/otlp.md
+++ b/specification/metrics/sdk_exporters/otlp.md
@@ -1,5 +1,9 @@
 # OpenTelemetry Metrics Exporter - OTLP
 
+**Status**: [Mixed](../../document-status.md)
+
+## Overview
+
 **Status**: [Stable](../../document-status.md)
 
 OTLP Metrics Exporter is a [Push Metric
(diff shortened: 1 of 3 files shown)


specification/metrics/sdk_exporters/otlp.md
```diff
@@ -1,5 +1,9 @@
 # OpenTelemetry Metrics Exporter - OTLP
 
+**Status**: [Mixed](../../document-status.md)
```
jack-berg, line 3: Wasn't quite sure how best to indicate that the document is stable _except_ one experimental config option. Decided to add "Overview" and "Additional Configuration" sections with different statuses. Open to alternatives.

jmacd (approved):

specification/metrics/sdk_exporters/otlp.md
```diff
@@ -19,18 +23,31 @@ variable](../../sdk-environment-variables.md#exporter-selection)),
...
-to use on the basis of instrument kind.  The recognized (case-insensitive) values are:
+The recognized (case-insensitive) values for `OTEL_EXPORTER_OTLP_DEFAULT_HISTOGRAM_AGGREGATION` are:
 
-| Value      | Definition                                                                                                    |
-|------------|---------------------------------------------------------------------------------------------------------------|
-| CUMULATIVE | Choose Cumulative aggregation temporality for all instrument kinds.                                           |
-| DELTA      | Choose Delta aggregation temporality for Counter, Asynchronous Counter and Histogram instrument kinds, choose Cumulative aggregation temporality for UpDownCounter and Asynchronous UpDownCounter instrument kinds. |
+* `explicit_bucket_histogram`:
+  Use [Explicit Bucket Histogram Aggregation](../sdk.md#explicit-bucket-histogram-aggregation).
+* `exponential_hisotgram`: Use [Exponential Histogram Aggregation](../sdk.md#exponential-histogram-aggregation).
```
reyang, line 53: minor: do we want to call it `exponential_bucket_histogram` (and perhaps update the SDK spec wording)?

reyang, line 53:
```suggestion
* `exponential_histotgram`: Use [Exponential Histogram Aggregation](../sdk.md#exponential-histogram-aggregation).
```

reyang (approved):
LGTM with a changelog entry.

specification/metrics/sdk_exporters/otlp.md
jack-berg: I couldn't think of any reason to justify this inconsistency and have updated the language here and in `sdk.md` 👍

specification/metrics/sdk_exporters/otlp.md
arminru: SGTM, maybe with "General" instead of "Overview" (which sounds more like an editorial introduction rather than the general normative section).

specification/metrics/sdk_exporters/otlp.md
```diff
@@ -19,18 +23,32 @@ variable](../../sdk-environment-variables.md#exporter-selection)),
...
-MetricReader](../sdk.md#periodic-exporting-metricreader).
+  MetricReader](../sdk.md#periodic-exporting-metricreader).
 * The exporter MUST configure the default aggregation temporality on the
   basis of instrument kind using the
   `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` variable as described
-  below, otherwise the exporter MUST use Cumulative as the default
-  aggregation temporality for all instrument kinds.
+  below.
+* The exporter MUST configure the default aggregation on the basis of instrument kind using
+  the `OTEL_EXPORTER_OTLP_DEFAULT_HISTOGRAM_AGGREGATION` variable as described below.
```
arminru, line 32:
```suggestion
  the `OTEL_EXPORTER_OTLP_DEFAULT_HISTOGRAM_AGGREGATION` variable as described below
  if it is implemented.
```
Otherwise it would be a breaking change for existing SDKs by making the additional config introduced here mandatory.

specification/metrics/sdk_exporters/otlp.md
```diff
@@ -19,18 +23,32 @@ variable](../../sdk-environment-variables.md#exporter-selection)),
...
+The recognized (case-insensitive) values for `OTEL_EXPORTER_OTLP_DEFAULT_HISTOGRAM_AGGREGATION` are:
 
-| Value      | Definition                                                                                                    |
-|------------|---------------------------------------------------------------------------------------------------------------|
-| CUMULATIVE | Choose Cumulative aggregation temporality for all instrument kinds.                                           |
-| DELTA      | Choose Delta aggregation temporality for Counter, Asynchronous Counter and Histogram instrument kinds, choose Cumulative aggregation temporality for UpDownCounter and Asynchronous UpDownCounter instrument kinds. |
+* `explicit_bucket_histogram`:
+  Use [Explicit Bucket Histogram Aggregation](../sdk.md#explicit-bucket-histogram-aggregation).
+* `exponential_bucket_histogram`:
+  Use [Exponential Bucket Histogram Aggregation](../sdk.md#exponential-bucket-histogram-aggregation).
```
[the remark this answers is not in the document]
jack-berg: The histogram in the name and value does feel repetitive, but its being used in different context so I think it makes sense to repeat it. In the name, histogram is referencing instruments. In the value, histogram is referencing aggregations. While probably unlikely, you could imagine other aggregation options being added here that aren't histograms.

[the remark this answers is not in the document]
pirgeo: That makes sense! Fine as is then.

arminru (approved):

Issue #2581, opened 2022-05-25
[Common] Spec inconsistency with proto definition of Attributes


MSNev:
The proto definition used for attributes uses "KeyValue" which can have a value of AnyValue, however, the spec definition of "Attribute" only allows for primitives or an Array of primitives, this is inconsistent.

What did you expect to see?
- They should be defined in the same way or have their own distinct definitions

**Additional context.**

The [Spec Attribute definition](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/common#attribute)

Trace [proto attribute definition](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto#L225-L237)
```protobuf
  // attributes is a collection of key/value pairs. Note, global attributes
  // like server name can be set using the resource API. Examples of attributes:
  //
  //     "/http/user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/71.0.3578.98 Safari/537.36"
  //     "/http/server_latency": 300
  //     "abc.com/myattribute": true
  //     "abc.com/score": 10.239
  //
  // The OpenTelemetry API specification further restricts the allowed value types:
  // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attributes
  // Attribute keys MUST be unique (it is not allowed to have more than one
  // attribute with the same key).
  repeated opentelemetry.proto.common.v1.KeyValue attributes = 9;
```

And the [Logs spec also uses Attributes](https://github.com/open-telemetry/opentelemetry-specification/issues/2581)

Logs [proto attribute definition](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/logs/v1/logs.proto#L198-L201)
```protobuf
  // Additional attributes that describe the specific event occurrence. [Optional].
  // Attribute keys MUST be unique (it is not allowed to have more than one
  // attribute with the same key).
  repeated opentelemetry.proto.common.v1.KeyValue attributes = 6;
```

With [Proto KeyValue](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/common/v1/common.proto#L61-L66) defined as taking AnyValue
```protobuf
// KeyValue is a key-value pair that is used to store Span attributes, Link
// attributes, etc.
message KeyValue {
  string key = 1;
[...]

PR #2620, opened 2022-06-16
[Common] Clarify Attribute support based on stable wire protocol definition of Attributes #2581

MSNev:
Fixes #
#2581 [Common] Spec inconsistency with proto definition of Attributes

## Changes

- Clarifies the Attribute support as defined by Proto `AnyValue` and clarifies Attribute Collections.
- Clarifies and identifies how nested attributes support (as defined by Proto) should be supported.

~Related issues #~
~#376 Support ~map values and~ nested values for attributes #376~


MSNev:
Not stale

spec-compliance-matrix.md
```diff
@@ -63,6 +63,7 @@ formats is required. Implementing more than one format is optional.
 | Double floating-point type                                                                       |          | +   | +    | +   | +      | +    | +      | -   | +    | +   | +    | +     |
 | Signed int64 type                                                                                |          | +   | +    | +   | +      | +    | +      | -   | +    | +   | +    | +     |
 | Array of primitives (homogeneous)                                                                |          | +   | +    | +   | +      | +    | +      | +   | +    | +   | +    | +     |
+| `nested` Attribute value                                                                         | x        |     |      | +   |        |      |        |     |      |     |      |       |
```
MSNev, line 66: On re-reviewing the JS code, this is not technically supported by the typescript type definitions.
 So will remove

Aneurysm9:
> Clarifies the Attribute support as defined by Proto AnyValue and clarifies Attribute Collections.
> Clarifies and identifies how nested attributes support (as defined by Proto and some SDK's) should be supported.

To the extent that this would apply to attributes for traces and metrics I think this goes beyond clarification and changes the specification.  To my knowledge there are no trace/metric SDK implementations that allow nested attributes.

MSNev:
> > Clarifies the Attribute support as defined by Proto AnyValue and clarifies Attribute Collections.
> > Clarifies and identifies `how nested attributes support (as defined by Proto and some SDK's) should be supported.`
> 
> To the extent that this would apply to attributes for traces and metrics I think this goes beyond clarification and changes the specification. To my knowledge there are no trace/metric SDK implementations that allow nested attributes.

If you read the spec changes, I believe that I have not specified that trace/metric implementations have to implement or allow for nested attributes. And even the PR comment from above I am stating how it `should` be supported not that it `must` or `is`

What specific or additional clarification would you suggest?

specification/common/README.md
```diff
@@ -31,6 +31,9 @@ An `Attribute` is a key-value pair, which MUST have the following properties:
   - A primitive type: string, boolean, double precision floating point (IEEE 754-1985) or signed 64 bit integer.
   - An array of primitive type values. The array MUST be homogeneous,
     i.e., it MUST NOT contain values of different types.
+  - It MAY also contain another valid Attribute value resulting in `nested` Attributes
```
arminru, line 34:
```suggestion
  - **For the [Logs](../logs) signal only:** It MAY also contain another valid Attribute value resulting in `nested` Attributes
```
This would document the status quo and thus resolve the current inconsistency in the spec pointed out in https://github.com/open-telemetry/opentelemetry-specification/issues/2581.
Adding support for nested attributes also for other signals should be proposed and discussed separately. This would be a breaking change as trace and metrics consumers do not have to expect such data as per OTel Traces and Metrics `v1.x`.

specification/common/README.md
```diff
@@ -52,6 +55,17 @@ indices that are kept in sync (e.g., two attributes `header_keys` and `header_va
 both containing an array of strings to represent a mapping
 `header_keys[i] -> header_values[i]`).
 
+`nested` Attribute support MAY be protocol, signal, language, SDK or Exporter
+specific as not all SDKs, exporters or backends support the storing or sending
```
arminru, line 59:
```suggestion
`nested` Attributes are only supported by the [Logs](../logs) signal.
Only certain protocols, languages, SDKs or Exporters support the storing or sending
```

specification/common/README.md
```diff
@@ -81,6 +95,13 @@ If an SDK provides a way to:
 - set a limit of unique attribute keys such that:
   - for each unique attribute key, addition of which would result in exceeding
     the limit, SDK MUST discard that key/value pair.
+- set a limit on the supported depth of nested attributes:
+  - when this limit is exceeded it MAY provide a configurable strategy for handling
+    items beyond this limit which MAY include:
+    - Dropping the excessive depth,
+    - Replacing the value with a relevant token indicating values have been truncated,
+    - Converted to a string representation, which MUST also comply with the value
+      length limits;
```
arminru, line 104: Having a configurable limit like this does not document the current status but would be a new feature and should be proposed in a separate issue and PR.

specification/common/README.md
```diff
@@ -101,6 +122,7 @@ followed by the global limit default value.
 
 * `AttributeCountLimit` (Default=128) - Maximum allowed attribute count per record;
 * `AttributeValueLengthLimit` (Default=Infinity) - Maximum allowed attribute value length;
+* `AttributeNestedLengthLimit` (Default=10; if supported) - Maximum allowed number of nested attribute value;
```
arminru, line 125: Having a configurable limit like this does not document the current status but would be a new feature and should be proposed in a separate issue and PR.

spec-compliance-matrix.md
```diff
@@ -63,6 +63,7 @@ formats is required. Implementing more than one format is optional.
 | Double floating-point type                                                                       |          | +   | +    | +   | +      | +    | +      | -   | +    | +   | +    | +     |
 | Signed int64 type                                                                                |          | +   | +    | +   | +      | +    | +      | -   | +    | +   | +    | +     |
 | Array of primitives (homogeneous)                                                                |          | +   | +    | +   | +      | +    | +      | +   | +    | +   | +    | +     |
+| `nested` Attribute value                                                                         | x        |     |      |     |        |      |        |     |      |     |      |       |
```
arminru, line 66:
```suggestion
```
see comment above

CHANGELOG.md
```diff
@@ -38,6 +38,10 @@ release.
 
 - Introduce Instrumentation Scope Attributes
   ([#2579](https://github.com/open-telemetry/opentelemetry-specification/pull/2579))
+- Clarify Attribute Value definition to match Protobuf AnyValue implementation
+  ([#2581](https://github.com/open-telemetry/opentelemetry-specification/issues/2581))
+- Support ~~map values and~~ nested values for attributes (via AnyValue)
+  ([#376](https://github.com/open-telemetry/opentelemetry-specification/issues/376))
```
arminru, line 44:
```suggestion
```
see comment above

spec-compliance-matrix.md
MSNev: Removed from compliance matrix

specification/common/README.md
MSNev: Reworked to relocate and highlight that this is only support by Logs (and ProtoBuf)

specification/common/README.md
MSNev: Removed

specification/common/README.md
arminru: I think it would be better to indicate right here in the list that this only applies to logs as the "See ..." below is easy to overlook. WDYT?

specification/common/README.md
```diff
@@ -52,6 +57,12 @@ indices that are kept in sync (e.g., two attributes `header_keys` and `header_va
 both containing an array of strings to represent a mapping
 `header_keys[i] -> header_values[i]`).
 
+`nested` Attributes are also how consumers MAY provide custom attributes that
+are domain specific and have no defined semantic conventions beyond the specific
+name used for the signal.
```
arminru, line 62: What's meant by "specific name"? The log attribute name (key)?

specification/common/README.md
```diff
@@ -116,6 +127,36 @@ Attributes, which belong to Metrics, are exempt from the limits described above
...
+This should be used as a guide only, when a specific area is not listed
+it does not infer that `nested` attributes are or MUST be supported.
+
+| Area                 | Optional  |
+|----------------------|-----------|
+| Span                 | -         |
+| Metrics              | -         |
+| Logs                 |           |
+| ProtoBuf (AnyValue)  |           |
+
```
arminru, line 159: I find this table confusing. To me it looks like "For Span and Metrics it is NOT optional and for Logs and ProtoBuf it is undefined whether it is optional".
How about this?
| Signal               | Supported |
|----------------------|-----------|
| Span                 | no        |
| Metrics              | no        |
| Logs                 | yes       |

I don't think it's necessary to mention again in the table that the protobuf AnyValue would support it.

specification/common/README.md
```diff
@@ -116,6 +127,36 @@ Attributes, which belong to Metrics, are exempt from the limits described above
...
 
+### Nested Attributes
+
+Only certain protocols, languages, SDKs or Exporter support the storing or sending
+of a hierarchy of Attributes. When not supported those cases they SHOULD implement
+a strategy to handle how nested Attributes are stored or transmitted, like using
+a JSON-encoded string value as highlighted for protocols above. The depth of the
+any resulting hierarchy SHOULD be limited and MUST not include recursive references.
+
+While the at the ProtoBuf protocol level the [`AnyValue`](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/common/v1/common.proto#L28-L67)
```
arminru, line 138: I hope this does not change again as it's a core message and considered stable but please use a permalink (you can just press `y` on your keyboard to have `main` replaced with the respective commit hash), just in case.

specification/common/README.md
MSNev: Understanding the concern about it being missed, but I'd still prefer to link out rather than repeat.

I'll move the "see" to the beginning.

specification/common/README.md
MSNev: "beyond the specific name used for the signal" is meant to call out that some signals (Logs) and more specifically Events will define a specific name to allow any Event to support passing user / application (domain) specific structure.

While vendors may also have their own (additional) semantic conventions which define their own domain specific "events"

specification/common/README.md
MSNev: Changed to yes/no and removed ProtoBuf

specification/common/README.md
```diff
@@ -31,6 +33,10 @@ An `Attribute` is a key-value pair, which MUST have the following properties:
   - A primitive type: string, boolean, double precision floating point (IEEE 754-1985) or signed 64 bit integer.
   - An array of primitive type values. The array MUST be homogeneous,
     i.e., it MUST NOT contain values of different types.
+  - It MAY (see [Nested Attribute Support](#nested-attribute-support)) also
+    contain another valid Attribute value resulting in `nested` Attributes
+    and the referenced Attribute MUST not be recursive. i.e. The value of the
+    key-value pair will itself be a key-value pair.
```
mateuszrzeszutek, line 39: It is unclear to me what impact this has on the API -- since you included this in a general Attribute specification, does that mean that you MAY pass nested attributes to a metric instrument?

Extending the generic Attribute definition has impact on all signals (and I suppose at least some SDKs share Attribute/Attributes classes/modules between signals); if you need to introduce nested attribute support in logs, then how about adding it directly to logs instead? Or introducing a separate "nested attribute collection" just for logs' sake.

specification/common/README.md
```diff
@@ -116,6 +128,34 @@ Attributes, which belong to Metrics, are exempt from the limits described above
...
+support.
+
+This should be used as a guide only, when a specific area is not listed
+it does not infer that `nested` attributes are or MUST be supported.
+
+| Area                 | Optional  |
+|----------------------|-----------|
+| Span                 | no        |
+| Metrics              | no        |
+| Logs                 | yes       |
```
mateuszrzeszutek, line 157: This table is kinda confusing, I am not sure what "optional = no" is supposed to mean in the context of spans/metrics.

Consider using the requirement keywords in the table instead

```suggestion
| Area                 | Nested attributes support |
|----------------------|---------------------------|
| Span                 | MUST NOT                  |
| Metrics              | MUST NOT                  |
| Logs                 | MAY                       |
```

arminru: I agree, to me this currently reads as "For Spans and Metrics it is not optional, i.e., a MUST requirement".

specification/common/README.md
arminru: I'm still confused by the sentence, sorry.
Consumers (processors/exporters/backends) don't provide any attributes, they consume them. Should it say producers (instrumentation/receivers) here?

> some signals (Logs) and more specifically Events will define a specific name to allow any Event to support passing user / application (domain) specific structure.

So that would be the attribute name then, right?

specification/common/README.md
MSNev: As per the existing released ProtoBuf definition someone CAN already construct and send nested attributes and they WILL be successfully transported. And depending on the receiver / backend this is something that needs to be handled, even if that handling strategy buy converting or dropping the values.

Additional, part of this clarification is driven by the fact that SDK's share Attribute/Attributes classes/modules, so that when the SDK's start implementing the Logs signal they WILL run into this confusion. Which would force them to create a situation where Attributes !== Attributes and therefore they will need to introduce additional mappings which is just additional complexity across the entire API surface rather than defining it as a "possibility" and therefore redirecting the required code changes to "handle" this situation to the relevant signal exported (protocol - i.e.. the point at which the Attributes are converted to the wire format).

So for Span and metrics the strategy would be that is already defined as per
> For protocols that do not natively support non-string values, non-string values SHOULD be represented as JSON-encoded strings. For example, the expression int64(100) will be encoded as 100, float64(1.5) will be encoded as 1.5, and an empty array of any type will be encoded as [].

And for Logs it would be supported

This then removes the need and complexity of having multiple Attribute definition.

specification/common/README.md
MSNev: > they consume them

That is the key point, and also where they should be handled.
When an exporter hits this situation it needs to handle it as it will occur either via Logs / events or some application developer passing in an or constructing their own exported / OTLP exporter (this has already occurred) as it won't be limited to just code produced by the OTel Sdk's.

specification/common/README.md
MSNev: Will update the "Optional" to "Supported"

Oberon00:
Please give this PR a more descriptive title. I think the "inconsistency" issue has been explained. It seems to me this is not solving an inconsistency but proposing a new feature.

MSNev:
> Please give this PR a more descriptive title. I think the "inconsistency" issue has been explained. It seems to me this is not solving an inconsistency but proposing a new feature.

Renamed.

I don't believe that this PR is identifying any new functionality, it is attempting to properly document what Attribute support is today. 
- The protocol SUPPORTS nesting for Attributes (and has for around 2 years)
- Logs SUPPORTS and requires nested attributes

While Spans and Metrics also use "Attributes" their definition of Attributes does not support nesting (from at the spec level), which if not clarified is going to cause confusion not just at the consumer level (using the API vs Protocol) but also at the implementation level when SDK's start to support the proper definition of the Logs Attributes

Oberon00:
> I don't believe that this PR is identifying any new functionality, it is attempting to properly document what Attribute support is today.

Look at any SDK and you will see that this is not a clarification, but an actual new (optional) feature.

MSNev:
> In practice, you are proposing a new feature. This is not supported today.

That is because none of the SDK's (that I'm aware of) are currently SPEC compliant for LOGS, once they start to correctly implement Attributes for Logs everyone will hit this issue.

Oberon00:
What should be "clarified" then, is that Log Attributes are not the general attribute types. You have a clear definition of a general "Attribute" here which does not allow nesting.
You have a clear definition of the "attributes" field on log records at https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-attributes, which is of type `map<string, any>`, and you also have a definition of `any` that explicitly allows nesting at https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#type-any

MSNev:
Not Stale

MSNev:
This last (forced) push is just to bring the PR up to date and mark as not stale. Forced push was required as git refused to do a normal push.

specification/common/README.md
```diff
@@ -116,6 +128,34 @@ Attributes, which belong to Metrics, are exempt from the limits described above
...
+
+This should be used as a guide only, when a specific area is not listed
+it does not infer that `nested` attributes are or MUST be supported.
+
+| Area                 | Supported  |
+|----------------------|------------|
+| Span                 | no         |
+| Metrics              | no         |
+| Logs                 | yes        |
+
```
MSNev, line 158: Add scopes and resources as not supported

specification/common/README.md
```diff
@@ -31,6 +33,10 @@ An `Attribute` is a key-value pair, which MUST have the following properties:
   - A primitive type: string, boolean, double precision floating point (IEEE 754-1985) or signed 64 bit integer.
   - An array of primitive type values. The array MUST be homogeneous,
     i.e., it MUST NOT contain values of different types.
+  - It MAY (see [Nested Attribute Support](#nested-attribute-support)) also
+    contain another valid Attribute value resulting in `nested` Attributes
+    and the referenced Attribute MUST not be recursive. i.e. The value of the
```
tigrannajaryan, line 38: @MSNev  What if we say something like this:
"- In addition for LogRecord attributes [Nested Attributes] are supported."

This makes it clear that it only applies to LogRecords and it is clear that it does not change semantics of attributes of any other data types.

specification/common/README.md
tigrannajaryan: Perhaps delete this paragraph. I don't think it helps to achieve what this PR wants to achieve.

specification/common/README.md
```diff
@@ -116,6 +128,34 @@ Attributes, which belong to Metrics, are exempt from the limits described above
 at this time, as discussed in
 [Metrics Attribute Limits](../metrics/sdk.md#attribute-limits).
 
+### Nested Attributes
+
+Only certain protocols, languages, SDKs or Exporter support the storing or sending
+of a hierarchy of Attributes. When not supported those cases they SHOULD implement
+a strategy to handle how nested Attributes are stored or transmitted, like using
+a JSON-encoded string value as highlighted for protocols above. The depth of the
+any resulting hierarchy SHOULD be limited and MUST not include recursive references.
```
tigrannajaryan, line 137: I would drop this. If above we say that LogRecords attribute can be nested then it is up to each languages to figure out how to support it.

The specifics of how to convert nested attribute values to non OTLP formats do not belong to this document. They should be in mapping-to-non-otlp.md.

specification/common/README.md
```diff
@@ -116,6 +128,34 @@ Attributes, which belong to Metrics, are exempt from the limits described above
...
+
+While the at the ProtoBuf protocol level the [`AnyValue`](https://github.com/open-telemetry/opentelemetry-proto/blob/157bedf2d42bd2c2c9a35e355839a48f99ce06ee/opentelemetry/proto/common/v1/common.proto#L28-L67)
+supports `nested` attributes via the `KeyValueList` at the specification level
+this does not infer that `nested` attributes are or should be fully supported for
+every protocol, signal, language, SDK or Exporter due to different environment,
+runtime and storage concerns.
```
tigrannajaryan, line 143: Same here, I don't think the paragraph is necessary.

specification/common/README.md
```diff
@@ -116,6 +128,34 @@ Attributes, which belong to Metrics, are exempt from the limits described above
...
+
+#### Nested Attribute Support
+
+This section provides a single compliance lookup of current `nested` attribute
+support.
+
+This should be used as a guide only, when a specific area is not listed
+it does not infer that `nested` attributes are or MUST be supported.
```
tigrannajaryan, line 151: I suggest to delete this and list all areas explicitly to make sure there is no misreading possible.

specification/common/README.md
```diff
@@ -147,3 +187,8 @@ Some other implementations may use a streaming approach where every
 that individual attribute value being exported using a streaming wire protocol.
 In such cases the enforcement of uniqueness will likely be the responsibility of
 the recipient of this data.
+
+An Attribute Collection does not infer a hierarchy or `nested` collection of
+Attributes, it represents a single flat definition which MAY be represented
+logically as an array, list, set or map as `nested` Attributes support is
+provided by the Attribute Value.
```
tigrannajaryan, line 194: I do not quite understand what this intends to say.

MSNev: This is attempting to state that Attribute Collections does not mean nested attributes, when I first read this section (because I came from the Protobuf perspective) I "assumed" that a Collection meant nesting. So just attempting to clarify so it's clear so someone elses doesn't make the same assumption.

specification/common/README.md
MSNev: Reworded a little

specification/common/README.md
MSNev: Reworded

specification/common/README.md
MSNev: Moved to mapping-to-non-otlp.md

specification/common/README.md
MSNev: Reworded

specification/common/README.md
MSNev: Removed

specification/common/README.md
MSNev: Reworded a little as well

specification/common/README.md
```diff
@@ -31,8 +32,15 @@ An `Attribute` is a key-value pair, which MUST have the following properties:
   - A primitive type: string, boolean, double precision floating point (IEEE 754-1985) or signed 64 bit integer.
   - An array of primitive type values. The array MUST be homogeneous,
     i.e., it MUST NOT contain values of different types.
+  - When specified via semantic conventions it MAY (see [Nested Attributes](#nested-attributes))
```
Oberon00, line 35: This is not a clarification but allows a new feature.

Oberon00 (changes requested):

specification/common/README.md
MSNev: This line explicitly calls out to see the Nested Attribute support which clarifies that this is NOT automatically supported by this line but must be explicitly defined by the semantic conventions (and is supported by Logs).

Oberon00: But it allows semantic conventions to add them, as written now, doesn't it?

MSNev:
Not stale as this is part of a larger issue like #2888

MSNev:
> Closed as inactive. Feel free to reopen if this PR is still being worked on.

Don't have the ability (permissions) to re-open...

PR #2627, opened 2022-06-21
FaaS conventions: make faas.document.time and faas.time attributes recommended

### LINKED ISSUE #2599: Faas trace attributes: requirement levels need revision
[faas trace semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/faas.md)  define multiple attributes as required and some of these choices seem controversial:
- `faas.document.time` - it's required, but it's likely the same time as span start time. Is this attribute really needed? Should it be `optional` ?
- `faas.time` - same concern - it should match span start time, why another one is needed and can it be `optional`?
- `faas.invoked_name` span attribute that SHOULD match `faas.name` resource attribute, seems like duplication, why  `faas.invoked_name` is needed and can it be `optional`?
- `invoked_provider` and `invoked_region` - should they be resource attributes (they are static)? should they be `recommended` instead?

lmolkova:
Fixes #2599

## Changes
`faas.document.time` and `faas.time` levels are changed from `requirement` to `recommended` as  span start time can be used instead when high precision is not so important.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ release.
   ([#2456](https://github.com/open-telemetry/opentelemetry-specification/pull/2456))
 - Change `cloudevents.event_spec_version` and `cloudevents.event_type` level from `required` to `recommended`
   ([#2618](https://github.com/open-telemetry/opentelemetry-specification/pull/2618))
+- Change `faas.document.time` and `faas.time` level from `required` to `recommended`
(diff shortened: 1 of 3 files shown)


lmolkova:
@open-telemetry/specs-trace-approvers can someone please review?

carlosalberto (approved):

PR #2633, opened 2022-06-28
Change the inclusivity of exponential histogram bounds

### LINKED ISSUE #2611: Prometheus and OTel High Resolution Histograms incompatibilities
In the recent [SIG Spec call](https://youtu.be/cRghV918KWk?t=1038) on 31st May, we discussed the incompatibilities in spec of Prometheus Sparse High Resolution Histograms and OTel Exponential Histograms. 

I've documented the differences and how things work in this doc here: https://docs.google.com/document/d/11z8hc_T4RHPHaXuWiETxGa81Ix89dBV5jh9byh9ySgE/edit

Please add your suggestions and comments to the doc and we will be meeting on June 22 (8am-9am PT) in the [Prometheus WG meeting](https://docs.google.com/document/d/19bnXziPn2MZ9wO6684UoI4D-LCjGL5bTJkGhux29bx8/edit#) to discuss this. Please feel free to join that meeting if interested.

jmacd:
Fixes #2611 (for the most part)

## Changes

The OpenTelemetry OTLP v0.11 exponential histogram defined lower-inclusive histogram buckets. The rationale for this decision is that the computations involved are simpler than for upper-inclusive histogram buckets, relatively speaking, due to the nature of IEEE 754's floating point representation.

The Prometheus project is strongly opposed to the use of lower-inclusive boundaries for histogram buckets because of historical precedent. In brief, there are existing modes of querying (existing) histogram data that are defined with upper-inclusive boundaries => it is possible to make exact inferences using existing histogram data => future histograms should preserve this mode of inference.

The Prometheus project has (with limits) adopted the base-2 exponential boundary structure defined in OTLP v0.11, so the only major difference between theirs and OpenTelemetry's is this issue. The proposed changes that will bring harmony in this space are: (1) this PR, (2) a [demonstration of the aggregator change (OTel-Go)](https://github.com/open-telemetry/opentelemetry-go/pull/2982), and (3) [a comments-only change in the protocol](https://github.com/open-telemetry/opentelemetry-proto/pull/409).

The counter argument to making this change can be summarized: 
1. It is slightly simpler to use lower-inclusive definitions
2. Exactness, even for power of two queries, is unwelcome overhead
3. Exactness is not possible for non-power-of-two queries
4. Negative-range numbers have opposite inclusivity in both proposals, counter to the claim of consistency.

The argument in favor of this proposal can be summarized:
1. Prometheus will be happy (even without exactness)
2. OTel never required exactness, so it's not a breaking change
3. Checking for exact for powers-of-two is cheap
4. ~Table-lookup implementations become less relatively-complex.~
 
Please see the changes to the mapping functions prototyped in https://github.com/open-telemetry/opentelemetry-go/pull/2982.

CC: @gouthamve @beorn7 @RichiH.


reyang (approved):
LGTM.

jmacd:
>  Isn't it rather the opposite? The floating-point representations of values 1.0 and 1.00001 have the same exponent, therefore it is easier to map them to the same bucket than to put them in different buckets.

I think it's established that lower-inclusive boundaries have less "absolute" complexity, that's demonstrated in these changes.  I suspect your point is that even a table-lookup implementation has to do more work here, because it requires a new test for `significand==0`--it's the same in every one of these methods, we're adding an explicit test for power-of-two values and subtracting 1 from the calculated index.

As I recall, @oertl your O(1) table-lookup implementation used 2 comparisons per call. I believe your point is that now it will require 3 comparisons per call?

I was trying to find a silver lining here: the complexity difference between table-lookup and log()-based methods is shrinking, even as they become more complex.

oertl:
@jmacd 
> As I recall, @oertl your O(1) table-lookup implementation used 2 comparisons per call. I believe your point is that now it will require 3 comparisons per call?

Yes, but maybe it is possible to avoid additional branching for powers of two by decrementing the floating-point input value first (e.g., using `Math.nextDown` in Java). Anyway, I do not see how table-lookup implementations would become "less relatively-complex" with exclusive lower bounds, which is one of your arguments in favor of this proposal.

jmacd:
@oertl I struck that sentence about relative complexity. (It didn't land well!)

You're right that `nextAfter` can change the boundary condition between one and another inclusivity (Prometheus made this suggestion, too, but it looks like equal complexity to me--it costs more than one additional comparison at least.)

I looked back at the table lookup mapping function I had prototyped, which I derived from yours ([included in this draft PR](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/1353)). To get a rough idea, the program to generate the table of constants is about the same size as the library that performs the lookup and the reverse mapping function.  The library to perform the lookup/reverse-mapping is about the same size of, but already more complex than the logarithm-based mapping function. To me the change you would make in a table lookup function will leave an implementation with about the same complexity as before. On the other hand, the change to be made in a logarithm or exponent-based lookup function leaves behind substantially more complexity than it had before, and this is what I meant by relative complexity. The logarithm implementation suffers a lot, the table lookup implementation suffers a little.

jack-berg (approved):
Working on the implementation but left some comments.

specification/metrics/data-model.md
```diff
@@ -599,14 +599,14 @@ downscale) without introducing error.
 #### Exponential Buckets
 
 The ExponentialHistogram bucket identified by `index`, a signed
-integer, represents values in the population that are greater than or
-equal to `base**index` and less than `base**(index+1)`.  Note that the
-ExponentialHistogram specifies a lower-inclusive bound while the
-explicit-boundary Histogram specifies an upper-inclusive bound.
+integer, represents values in the population that are greater than
+`base**index` and less than or equal to `base**(index+1)`.
```
beorn7, line 603: Just as a side note: If you changed this definition to "greater than `base**(index-1)` and less than or equal to `base**index`, you would save a whole bunch of `- 1` terms below. (Not saying that you should do it. This is really just an internal implementation detail. But during the discussion, it was mentioned that the "upper inclusive" way was unnatural because you need those `- 1` terms. But in reality, it depends on the definition of `index`. (And fittingly, in Prometheus, the bucket with id==0 is the one with an upper bound of 1.))

specification/metrics/data-model.md
```diff
@@ -650,6 +650,18 @@ that have been rounded to zero.
 
 #### Producer Expectations
 
+Producers MAY use an inexect mapping function because in the general
```
beorn7, line 653: inexect → inexact

beorn7:
Sorry for the separate comments rather than doing one review.

I think this looks fine from the Prometheus perspective. Many thanks for doing this!

specification/metrics/data-model.md
```diff
@@ -716,21 +726,30 @@ func GetExponent(value float64) int32 {
...
     // Note: Frexp() rounds submnormal values to the smallest normal
     // value and returns an exponent corresponding to fractions in the
-    // range [0.5, 1), whereas we want [1, 2), so subtract 1 from the
+    // range [0.5, 1), whereas we want (1, 2], so subtract 1 from the
     // exponent.
-    _, exp := math.Frexp(value)
+    frac, exp := math.Frexp(value)
+	if frac == 0.5 {
+   		// Special case for power-of-two boundary: subtract one.
+	    exp--
```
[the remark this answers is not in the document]
beorn7: But then the doc comment should be "subtract two instead of one". ;)

specification/metrics/data-model.md
```diff
@@ -767,18 +793,42 @@ compute the bucket index.  A multiplicative factor equal to `2**scale
 example:
 
 ```golang
+// MapToIndex for any scale.
+func MapToIndex(value float64) int {
     scaleFactor := math.Ldexp(math.Log2E, scale)
-    return math.Floor(math.Log(value) * scaleFactor)
+    return math.Ceil(math.Log(value) * scaleFactor) - 1
```
jack-berg, line 799: It wasn't obvious to me what the `scalefactor` is doing here and why it works. Here's how I reasoned through it:

1. The goal is to find the `index` for the value. We know that `base = 2**(2**(-scale))`, and that a given `index` has boundaries of `base**index < values <= base**(index+1)`. 
2. Logs prove useful here. We know that "log<sub>base</sub>(value) = index", and can bring it to a whole number while respecting our boundaries by using "index = ceil(log<sub>base</sub>(value)) - 1".
3. Its inconvenient to take the log using a real number base, so let's try to change the formula to use natural log (which is built into most programming language's math library) and using `scale` instead of `base`. 
4. We know that log<sub>x</sub>y = log<sub>n</sub>y / log<sub>n</sub>x, where the `n` is any base. We can use that to change our formula to "index = ceil(log<sub>2</sub>(value)/log<sub>2</sub>(base)) - 1".
5. We know that `base = 2**(2**(-scale))`, so "log<sub>2</sub>(base) = 2^(-scale)", or  "log<sub>2</sub>(base) = 1/(2^scale)", or "2^scale = 1/log<sub>2</sub>(base)". We see the "1/log<sub>2</sub>(base)" term in our equation, and can replace it with "2^scale" to get "index = ceil(log<sub>2</sub>(value)*2^scale) - 1".
6. We can again change the base of our logs from base 2 to base e to get "index = ceil(ln(value)*2^scale/ln(2)) - 1", and there we see that we can extract a constant "2^scale/ln(2)" as our `scalefactor`. 
7. Bringing it all together we have "index = ceil(ln(value)*scalefactor) - 1", where "scalefactor = 2^scale/ln(2)".

Deriving the scale factor helped improve my understanding during implementation and I believe could help others develop better intuition and avoid bugs.

specification/metrics/data-model.md
jmacd: > (And fittingly, in Prometheus, the bucket with id==0 is the one with an upper bound of 1.)

Is this a mistake? I was under the impression that we were dealing with a boundary-condition difference, not an off-by-one-bucket difference combined with a boundary difference. I thought that if OpenTelemetry make the change in this PR, we would have identical bucket numbering schemes. @gouthamve can you clarify this matter? This was not clear at this [bookmark](https://docs.google.com/document/d/11z8hc_T4RHPHaXuWiETxGa81Ix89dBV5jh9byh9ySgE/edit#bookmark=id.qbepdrorwhde) in your design document, it looked like Prometheus could equally use the `math.Ceil(math.Log(x) * scaleFactor)-1` formula, but according to @beorn7's statement here, the formula that Prometheus can use would be `math.Ceil(math.Log(x) * scaleFactor)`. 

I do not want to make the compromise if it leaves us with an off-by-one condition, and I feel we are not receiving coherent guidance. Should OTel be using `math.Ceil(math.Log(x) * scaleFactor`? That is both a boundary change AND an correction by 1.

gouthamve: Hi, so this is an internal detail imo. You could implement this as is, with the current indexing and it would be completely fine and okay.

I was vaguely aware that Prometheus indexing and OTel indexing is off by one but I dismissed it as I thought it was an implementation detail. I was more focused on making sure that the conversion from one to the other is easy. The algorithm on how to encode doesn't need to be *exactly* same and is indeed not the same given Prometheus does a sparse representation and OTel does a dense representation. I thought the indexing being off by one wouldn't change anything when it comes to converting from one to the other.

>  feel we are not receiving coherent guidance

This was not my intent but I can see how it comes across, and I am sorry about that 😬. 

> Should OTel be using `math.Ceil(math.Log(x) * scaleFactor`?

No, I don't think it is required tbh. It was a suggestion on simplifying things, but I think it can be ignored if it is too big a change.

beorn7: > I thought that if OpenTelemetry make the change in this PR, we would have identical bucket numbering schemes.

To be clear, I was completely unaware that identical bucket numbering schemas were a goal here.

jmacd: Still, we were led into thinking that Prometheus was using an equation equivalent to `math.Ceil(math.Log(x) * scaleFactor) - 1` and that the only difference was boundary inclusivity and zero tolerance. Given this new information, I do not want this change in its current form--this outcome was only a good one if it gave us identical numbering. It would lead to confusion and entirely predictable errors if the rules for converting between the two histograms require adding or subtracting 1 from bucket numbers.

@open-telemetry/specs-metrics-approvers I think we are better off making a breaking change to remedy this. As the thread above shows, using `math.Ceil(math.Log(x) * scaleFactor)` (without the -1) returns bucket 0 for value 1, which equals the OTLP v0.11 formulation because the off-by-one and the inclusivity change counteract each other.

jmacd:
As explained in https://github.com/open-telemetry/opentelemetry-specification/pull/2633#discussion_r914269139, I no longer think this is a good idea.

jmacd:
As explained in https://github.com/open-telemetry/opentelemetry-specification/issues/2611#issuecomment-1178119261, I still consider this change acceptable, but the text had some errors. I will re-open this after correcting the mistakes.

specification/metrics/data-model.md
jmacd: I have documented why I prefer `math.Ceil(...)-1` as compared with `math.Ceil(...)` here: https://github.com/open-telemetry/opentelemetry-specification/issues/2611#issuecomment-1178119261. This is not an implementation detail, it would be a breaking change for us at this point.

The `math.Ceil()-1` formulation is preserves the property that the right-shift operator can be used to downscale the histogram, whereas in a `math.Ceil()` formulation, downscaling the histogram requires a complicated adjustment, you need to write `((oldIndex-1)>>downscale)+1` in a whole bunch of places.

In short, the original comment ("you would save a whole bunch of - 1 terms below") is misleading.

specification/metrics/data-model.md
```diff
@@ -704,10 +716,8 @@ double-width floating point values have indices in the range
 `[-1074, -1023]`.  This may be written as:
```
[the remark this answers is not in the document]
jmacd: The minimum normal value stays the same, but its index becomes one smaller yes.

specification/metrics/data-model.md
jmacd: I've added a derivation for the index, left deriving the scaling factor an exercise :grin:. I found this derivation originally from @yzhuge and had it written in the source, but you're right it belongs here.

specification/metrics/data-model.md
```diff
@@ -791,12 +841,19 @@ reference implementation, for example, the above formula computes
...
+}
 ```
 
+In the Golang reference implementation, for example, the above formula
+does not accurately compute the lower boundary of the minimum-index
+bucket (it is a subnormal value).  In this case, it is appropriate to
+add `1<<scale` to the index and divide the result by `2`.
+
 *Note that floating-point to integer type conversions have been
 omitted from the code fragments above, to improve readability.*
```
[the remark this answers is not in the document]
jmacd: I added some text to clarify why there are different mapping functions being given.

specification/metrics/data-model.md
```diff
@@ -743,13 +762,20 @@ by `-scale`.  Note that because of sign extension, this shift performs
 correct rounding for the negative indices.  This may be written as:
```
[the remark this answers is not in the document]
jmacd: Fixed.

specification/metrics/data-model.md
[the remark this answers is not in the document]
jmacd: I untabified. :grin:

specification/metrics/data-model.md
```diff
@@ -720,40 +727,59 @@ func GetExponent(value float64) int32 {
         // unless value is zero.
         rawExponent -= int64(bits.LeadingZeros64(rawSignificand) - 12)
```
reyang, line 728:
```suggestion
        rawExponent -= int64(bits.LeadingZeros64(rawSignificand - 1) - 12)
```

specification/metrics/data-model.md
```diff
@@ -708,10 +717,8 @@ double-width floating point values have indices in the range
...
 ```golang
-// GetExponent extracts the normalized base-2 fractional exponent.
-// Let the value be represented as `1.significand x 2**exponent`,
-// this returns `exponent`.  Not defined for 0, Inf, or NaN values.
-func GetExponent(value float64) int32 {
+// MapToIndexScale0 computes a bucket index at scale 0.
+func MapToIndexScale0(value float64) int32 {
     rawBits := math.Float64bits(value)
     rawExponent := (int64(rawBits) & ExponentMask) >> SignificandWidth
     rawSignificand := rawBits & SignificandMask
```
reyang, line 724: Consider changing to "fragment" to avoid confusion - for _normal_ numbers, this value represents `normalized significand - 1` instead of significand.

```suggestion
    rawFragment := rawBits & FragmentMask
```

specification/metrics/data-model.md
reyang: FYI - here goes the prototype in OpenTelemetry .NET:

Implementation: https://github.com/open-telemetry/opentelemetry-dotnet/blob/720cbdd2f8172569861e23783d8a06b6d5950f30/src/OpenTelemetry/Metrics/ExponentialBucketHistogram.cs#L105

Test: https://github.com/open-telemetry/opentelemetry-dotnet/blob/720cbdd2f8172569861e23783d8a06b6d5950f30/test/OpenTelemetry.Tests/Metrics/ExponentialBucketHistogramTest.cs#L45-L55

specification/metrics/data-model.md
jmacd: Thanks, great point. I added more comments to explain what's happening.

specification/metrics/data-model.md
jmacd: Thank you!  The logic I had written was not correct for subnormal values. I had not tested the upper-inclusive form of this function, the "correct subnormals" variation.

specification/metrics/data-model.md
```diff
@@ -765,51 +813,100 @@ boundary `0x1p-1024`.
...
+calculate the exactly corect bucket index.  Since it is relatively
+simple to check for exact powers of two, implementations SHOULD
+apply such a special case:
 
 ```golang
+// MapToIndex for any scale, exact for powers of two.
+func MapToIndex(value float64) int {
+    // Special case for power-of-two values.
+    if frac, exp := math.Frexp(value); frac == 0.5 {
+        return ((exp-1) << scale) - 1
```
reyang, line 865: Nit:

```suggestion
        return ((exp - 1) << scale) - 1
```

reyang (approved):
LGTM. I've [prototyped this in OpenTelemetry .NET](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3462), as I can tell, it is working very well. @jmacd thank you for great work! 👏

jmacd:
@open-telemetry/specs-approvers Please approve this PR that was requested by the Prometheus developers as it will improve collaboration between these groups.

gouthamve (approved):
LGTM!

dashpole (approved):
LGTM

cijothomas (approved):

PR #2635, opened 2022-06-29
[editorial] Add Hugo front matter to selected pages

chalin:
- Contributes to https://github.com/open-telemetry/opentelemetry.io/issues/1467

/cc @austinlparker

### DIFF
--- a/specification/overview.md
+++ b/specification/overview.md
@@ -1,3 +1,7 @@
+<!---
+weight: 1
+--->
+
 # Overview
 
 <details>
--- a/specification/README.md
+++ b/specification/README.md
@@ -1,3 +1,14 @@
+<!---
+linkTitle: Specification
+no_list: true
+cascade:
+  body_class: otel-docs-spec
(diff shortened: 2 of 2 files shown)


specification/overview.md
```diff
@@ -1,3 +1,7 @@
+<!---
+weight: 1
+--->
+
```
arminru, line 4: Would it be possible to add a comment that this is needed for the website? I could imagine someone passing by in a few months thinking this is some leftover/accidental comment and then removing it.

arminru (approved):

specification/overview.md
chalin: Sure, how about something like this:

arminru: @chalin Yes I think this would be helpful!

arminru: Sure!

chalin: - Done via #2651

Issue #390, opened 2019-12-16
Default SDK Configuration mechanisms


jkwatson:
The default SDK should have a specified mechanism for configuration. This issue is not to track the specification of what is configurable, but rather *how* configuration should be made available to SDK users.

In general, the underlying mechanism should be programmatic. Every piece of the default SDK that requires configuration should have that configuration exposed to code. Declarative options (eg. JSON, YAML, env vars, etc) can then be built on top of that programmatic mechanism, as appropriate to the language of the SDK.

jkwatson:
Note: I am working on a PR to provide the skeleton of this specification.

mwear:
Ruby has a streamlined way to configure the SDK. I think it is similar to what you are proposing here. Take a look at this PR for examples: https://github.com/open-telemetry/opentelemetry-ruby/pull/171.

PR #2636, opened 2022-06-29
Add support for partial success in an OTLP export response

### LINKED ISSUE #2454: Add ability for OTLP response to signal partial acceptance of data
Each OTLP request can carry a batch of data some of which may be acceptable for the receiver and some may be unacceptable. Today the receiver cannot signal this situation to the sender. The receiver has to either accept the entire batch or reject the entire batch. Typically implementations accept the entire batch and then drop the unwanted portion of the data. This works from the receiver's perspective however the sender remains uniformed that something may be wrong.

A typical situation when this can happen is when OTLP logs are used for carrying different types of events, e.g. profiling events, client-side browser events, generic application logs, etc. These data is sent to the backend where these various event types may correspond to different product features that the end user may or may not be entitled to. So, for example if the user is entitled to generic application logs but is not entitled to profiling events the batch needs to be processed and profiling events need to be dropped from it.

While this approach works it leaves the senders uninformed which complicates the troubleshooting.

We need a way for OTLP to signal in its response that the request was partially accepted. This will allow the sender to surface the problem to the user (e.g. it can be a log message in the Collector's OTLP exporter).

One possible approach is to return counters of accepted and dropped data in the response, e.g.
```protobuf
message ExportLogsServiceResponse {
  int64 accepted_log_records = 1;
  int64 dropped_log_records = 2;
  // may also add counters for Resources and Scopes.
}
```
If necessary a finer granularity information about individual records may be also added to the response to further aid the troubleshooting, e.g. add a message explaining the reason for dropping or why the data is considered bad.

joaopgrassi:
Fixes #2454 

## Changes

This PR adds guidance around how servers can signal a partial success to clients. It indicates in which situations servers can respond with a partial success and how the fields introduced in [opentelemetry-proto #390](https://github.com/open-telemetry/opentelemetry-proto/pull/390) should be populated.

I also refactored the `OTLP/gRPC Response` section a bit, to make it consistent with the HTTP one. The main inconsistency I found was that the gRPC section did not have clear sections for `Success`, `Failure` as the HTTP does. Now both share the same sections and the wording should be consistent between the two.

Related
Proto change: https://github.com/open-telemetry/opentelemetry-proto/pull/390


joaopgrassi:
Even though the [Proto PR](https://github.com/open-telemetry/opentelemetry-proto/pull/390) is not merged, I thought of already opening the spec change, to kick start discussions. 

cc @tigrannajaryan @jmacd @jack-berg @reyang

jmacd (approved):

jmacd:
@open-telemetry/specs-metrics-approvers please take a look. The corresponding change in OTLP has been approved.

specification/protocol/otlp.md
```diff
@@ -145,16 +149,62 @@ was not delivered.
...
 #### OTLP/gRPC Response
 
-The server may respond with either a success or an error to the requests.
+The response MUST be the appropriate serialized Protobuf message (see below for
+the specific message to use in the [Success](#success),
+[Partial Success](#partial-success) and [Failure](#failures) cases).
+
+##### Success
 
 The success response indicates telemetry data is successfully processed by the
```
yurishkuro, line 158: "processed" seems very vague. I think producers care the most about the chain of custody, i.e. that the data was handed over to the server who is now responsible for this data, subject to its SLOs (including SLOs for data loss).

If we phrase it in terms of accepting responsibility, then the next paragraph about "processing at a later point" becomes irrelevant.

specification/protocol/otlp.md
```diff
@@ -145,16 +149,62 @@ was not delivered.
...
+The success response may also be used when telemetry data processing is done
+at a later point. In such cases, the server may want to return
+immediately to ensure low-latency communication with clients.
+
+If the server receives an empty request (a request that does not carry
 any telemetry data) the server SHOULD respond with success.
 
-Success response is returned via
+On success, the server response MUST be a Protobuf-encoded
+[Export*ServiceResponse](https://github.com/open-telemetry/opentelemetry-proto)
```
yurishkuro, line 169:
```suggestion
[Export<signal>ServiceResponse](https://github.com/open-telemetry/opentelemetry-proto)
```

specification/protocol/otlp.md
```diff
@@ -145,16 +149,62 @@ was not delivered.
...
+message (`ExportTraceServiceResponse` for traces,
+`ExportMetricsServiceResponse` for metrics and
+`ExportLogsServiceResponse` for logs).
+
+The server SHOULD respond with success no sooner than after successfully
```
yurishkuro, line 174:
```suggestion
The server SHOULD respond with success only after successfully
```
(remove double negative)

specification/protocol/otlp.md
```diff
@@ -145,16 +149,62 @@ was not delivered.
...
+
+The server MUST leave the `partial_success` field unset
+in case of a successful response.
+
+##### Partial Success
+
+If the processing of the request is partially successful
+(i.e. when the server accepts only parts of the data and rejects the rest), the
+server response MUST be a Protobuf-encoded
 [Export*ServiceResponse](https://github.com/open-telemetry/opentelemetry-proto)
```
yurishkuro, line 185:
```suggestion
[Export<signal>ServiceResponse](https://github.com/open-telemetry/opentelemetry-proto)
```

specification/protocol/otlp.md
```diff
@@ -145,16 +149,62 @@ was not delivered.
...
+`ExportMetricsServiceResponse` for metrics and
+`ExportLogsServiceResponse` for logs).
+
+Additionally, the server MUST initialize the `partial_success` field
+(`ExportTracePartialSuccess` message for traces,
+`ExportMetricsPartialSuccess` message for metrics and
+`ExportLogsPartialSuccess` message for logs), and it MUST set the respective
+`accepted_spans`, `accepted_data_points` or `accepted_log_records` field with
+the number of spans/data points/log records it accepted. In case the server
+rejected everything, the `accepted_*` field MUST be set to `0`.
```
yurishkuro, line 196: if it rejected everything, how is it a "success"?

specification/protocol/otlp.md
```diff
@@ -382,8 +432,9 @@ numbers or strings are accepted when decoding.
 
 #### OTLP/HTTP Response
 
-Response body MUST be the appropriate serialized Protobuf message (see below for
```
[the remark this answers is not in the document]
joaopgrassi: There are a few nuances, for ex, for HTTP, we indicate which `Content-Type` should be used for the response, compression, which status code to return. Such things don't apply to gRPC.. so I was not sure what makes sense. Having a combined section and doing "If HTTP, return xyz header" doesn't sound good. Another point is if someone is implementing a gRPC server, they don't don't have to read the HTTP stuff and vice-versa. 

I'm open to improve it, but I'm not sure to what extent. It could also be a follow up to make the review easier? Open to suggestions :)

specification/protocol/otlp.md
joaopgrassi: Doesn't this change the meaning a bit? Before it was implied that the server could respond with success in two moments:
1. after validating
2. after validating and processing. 

But not sooner than `1`. With the change, now one could interpret that as soon as the server validates the message, it needs to respond.

specification/protocol/otlp.md
joaopgrassi: Fixed in [57734b1](https://github.com/open-telemetry/opentelemetry-specification/pull/2636/commits/57734b10340b19168700805363aa3d7bdeeae72a)

specification/protocol/otlp.md
joaopgrassi: Fixed in [57734b1](https://github.com/open-telemetry/opentelemetry-specification/pull/2636/commits/57734b10340b19168700805363aa3d7bdeeae72a)

specification/protocol/otlp.md
joaopgrassi: Yeah now that you mentioned I think I agree. I will try to rephrase it.

specification/protocol/otlp.md
```diff
@@ -145,16 +149,58 @@ was not delivered.
...
+
+Additionally, the server MUST initialize the `partial_success` field
+(`ExportTracePartialSuccess` message for traces,
+`ExportMetricsPartialSuccess` message for metrics and
+`ExportLogsPartialSuccess` message for logs), and it MUST set the respective
+`accepted_spans`, `accepted_data_points` or `accepted_log_records` field with
+the number of spans/data points/log records it accepted. In case the server
+rejected everything, the `accepted_<signal>` field MUST be set to `0`.
+
+The server MAY populate the `error_message` field with a human-readable
```
tigrannajaryan, line 194:
```suggestion
The server SHOULD populate the `error_message` field with a human-readable
```

specification/protocol/otlp.md
tigrannajaryan: Isn't this requirement superfluous? How can the server respond at all without decoding and validating the request? I think this is already implied by the fact that the server needs to respond with success/partial/failure and constructing such response does not seem to be possible unless decoding and validation is done.

joaopgrassi: Less is more, so I'm in favor of removing it. I just didn't because it was already there but I agree :)

specification/protocol/otlp.md
```diff
@@ -520,11 +604,21 @@ received yet. The client will typically choose to re-send such data to guarantee
 delivery, which may result in duplicate data on the server side. This is a
 deliberate choice and is considered to be the right tradeoff for telemetry data.
 
-### Partial Success
+#### Partial Success Retry
+
+Each server has its particularities and its way of treating data. There
```
jsuereth, line 609: This seems to be a bit verbose.  I think we should specify more explicitly here:

1. Clients, by default, will not attempt automatic retry.
2. Servers should leverage partial success when they fully understand that resending the *exact same* bundle of telemetry would lead to another error (preventing retry loops).

The rest of this commentary is nice for WHY, but the WHAT should be called out explicitly.

jsuereth (approved):

specification/protocol/otlp.md
joaopgrassi: Removed in [d9482df](https://github.com/open-telemetry/opentelemetry-specification/pull/2636/commits/d9482df28dddc4d83ae10429a2e288d40bfc7788)

jmacd (changes requested):
@tigrannajaryan has raised concerns about the use of optional. I believe we should revise this PR so that the partial_success messages always have a meaningful 0 value, which means counting dropped things not accepted things.

yurishkuro:
I still have a problem with `accepted_<signal> = 0`, and I think this problem comes from unclear motivation in the original ticket.

Errors from the server can be viewed in two separate categories: client errors and server errors.
  * Client errors mean the request or data is malformed, and the server wants to reject it. In this case there is no point for the client to resent the data. 
  * Server errors mean it's not the client fault - the server may be overloaded and unable to accept the data, or the server may be trying to write to storage synchronously and there's a hiccup there. The server may indicate whether the error with the data item is retryable or not to inform the client what it can do.

I feel like the wording in this PR is trying to ignore this complexity and use poorly defined terms like "accepted".

So I suggest to first start with a clear motivation - why does the client need to know partial success? What are the possible outcomes of such response? This will then inform how the response format should look.

jmacd:
> I feel like the wording in this PR is trying to ignore this complexity

I don't see it that way. This specification refers to conditions that are neither client errors not server errors and it does not change the interpretation of either. The data is well formed, there is neither a client error nor a server error, but there is still a problem in the data.

One easy example  -- an OTLP <signal> export request arrives with weeks-old data. We are not willing to accept such old data, so we have to drop those points. It's not a client error, not a server error, it's likely a wall clock error. We have no grounds to fail the request (it was not a client error nor a server error) -- the OTLP was also well formed, just couldn't be accepted.

My goal in supporting this proposal is that I think we MUST have a way to count the number of points that are being accepted vs. dropped. I support having a free-form error message string too, but I am not willing to include an error message for every point that is not accepted, thus I need a way to count points accepted/dropped one way or another independent of the error details string.

yurishkuro:
Your example seems quite clearly a client error, which breaks the implied contract that the dates should be within a certain time range. The server can reject it with "client error" indicator and the client will know it should not resubmit it.

Saying that it's some grey area is precisely what bothers me in the proposal. We can argue about the validity of certain server decisions, but the protocol must be unambiguous with clear path forward after receiving the partial response.

Currently the only path forward after receiving partial success is to do nothing functional. Client can log something, but it cannot resubmit the data that was not "accepted" (for reasons unknown in machine-readable way). In this case, why not just add accepted counts to the regular successful response? It's backwards compatible.

It also doesn't address the server-error situation where some items could be retried - it doesn't fit in any of the 3 response categories.

I'd still have a conceptual issue with server returning all accepted counts as zeros - it's equivalent to a failed response, so the duality indicates a design issue.

jmacd:
> breaks the implied contract that the dates should be within a certain time range.

I do not believe there is any such contract.

In any case, what would you advise if half of the data is valid (good clocks) and half of the data is out of date?

tigrannajaryan:
>Client can log something, but it cannot resubmit the data that was not "accepted" (for reasons unknown in machine-readable way). In this case, why not just add accepted counts to the regular successful response? It's backwards compatible.

I think this is what this proposal does. It adds an optional message that is an indication to the client that it needs to log an error. The message contains a human readable error text and a number that indicates how many records were successful. Both of these items (error text and successful count) seem useful for diagnosing bad situations. I would expect the client to log something like this:

"Request partially successful. Sent $sent_records, accepted $accepted_records. Error text: $error_text".

Such a message IMO can be very useful for understanding why you send X records, but somehow only Y appear on the backend.

>I'd still have a conceptual issue with server returning all accepted counts as zeros - it's equivalent to a failed response, so the duality indicates a design issue.

Not necessarily equivalent. Accepted=0 indicates the Server has a problem with individual data records (all of them). To use @jmacd's example - perhaps all data points timestamps are stale and cannot be accepted, although the data itself is understandable to the Server. As opposed to that a failed response indicates that the Server has a problem with the request as a whole (i.e. can't be decoded as Protobuf, hence returns 400 Bad Data response).

Perhaps the design issue here is that we have 2 different ways to indicate an error to the client: one using HTTP response codes and another using Protobuf messages. I think this is inevitable. We do want to return proper HTTP response codes when they applicable (i.e. 400 when request payload is impossible to decode). However, HTTP response codes alone are not sufficient to convey more nuanced situations, hence additional data must be present in the response body, which happens to a Protobug message. In theory we could say that all failure information should be conveyed via response body in the Protobuf message (i.e. always return 200 and a body with an "error=true" field inside). Unfortunately we can't do this anymore, since that would be a breaking change.

joaopgrassi:
> I feel like the wording in this PR is trying to ignore this complexity and use poorly defined terms like "accepted".

@yurishkuro I'm not sure I agree and I'm a bit confused as well. I used the word `processing` to define what the server did with the data, which you pointed out to be vague and that senders care on chain of custody, which I completely agree. That's why I changed the semantics to "accepted". Once the server accepted the data, what it does after is not to be cared by senders. Maybe I misunderstood you then? 🤔 

The others gave good examples already, but another one I could think of is a customer hitting some sort of quota/limitation on their subscription model. A customer with a subscription allowing 100gb of logs in their subscription, multiple apps are sending logs. A server might start accepting a batch of logs, and halfway stop because the quota was reached by another app. Is this a client or a server error? IMHO they are neither.

Another case: A back-end might have specific rules about what they accept for metric keys. What should a server respond in case a client sends a batch of 1000 datapoints where half of it has "invalid" metric keys and the server rejected it? You can say, ah but this is a client error, they should know the back-end format for metric keys. Ok, what if the metrics are not coming from the app itself, but from a 3rd party library the app uses, e.g. a db driver library? Today, the only way a server can signal such case is via a 400 Bad request because we have a payload to put stuff in. I honestly don't think this is a good experience for anyone. With the `partial_success` we could solve this and stop returning 400s and triggering false alarms on customers.

>It also doesn't address the server-error situation where some items could be retried - it doesn't fit in any of the 3 response categories.

This was [also discussed in the Proto PR](https://github.com/open-telemetry/opentelemetry-proto/pull/390#discussion_r871989856), and it seemed to me we had an agreement that it _might_ make sense to include more detailed and structured information about the error so they could be retry-able, but this was not a focus for now. 

The problem is, none of us could come with well-defined examples of such error cases nor a good structure. Also, this would blow up the scope of this PR and require massive work around defining semantic conventions first. What we had in these two PRs seemed like a good first iteration that would help many cases and we can always improve it and add more structure in a non-breaking manner. I even created a separated issue to discuss such extra structure/retry for metrics https://github.com/open-telemetry/opentelemetry-proto/issues/404

I'm out of ideas on how to continue with this at this point. Should we stop it altogether? Open an OTEP to discuss it?

yurishkuro (approved):

joaopgrassi:
As we agreed, I opened a new PR to change the semantics from "accepted" to "rejected". https://github.com/open-telemetry/opentelemetry-proto/pull/414.

Let me know your thoughts. If we agree on that approach I will make the changes to this PR.

jmacd:
This can be reopened if need be, but there appears to be consensus around #2696.

jmacd:
Thank you @joaopgrassi !!!!

PR #2637, opened 2022-06-30
Make clear that context is not mutable when setting a span

### LINKED ISSUE #2624: Context Interaction Section Makes it Seem like Context Should Be Mutable
Was trying to implement the 2nd part of [the Context Interaction](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#context-interaction) section for PHP.

What did you expect to see?
I expected the spec to call out that a new context would need to be created with the span that was being passed.

Instead, the spec seemed to imply that the current context should be mutated with the span that was being passed.

**Additional context.**
Here is that spec language that threw me off

"Set the currently active span to the implicit context. This is equivalent to getting the implicit context, then inserting the Span to the context."

Link to [this section of the doc](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#context-interaction)

Fwiw I'm not sure how the spec words similar discussions of context mutability elsewhere, so if this is just being consistent with that (and me being ignorant of those parts) that seems totally fair to me (and this can be closed)

Grunet:
Fixes #2624 

## Changes

Tries to clarify the sentence to make it clear that setting a span should not mutate the current context.

Related issues #

### DIFF
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -173,7 +173,7 @@ If the language has support for implicitly propagated `Context` (see
 the following functionality:
 
 - Get the currently active span from the implicit context. This is equivalent to getting the implicit context, then extracting the `Span` from the context.
-- Set the currently active span to the implicit context. This is equivalent to getting the implicit context, then inserting the `Span` to the context.
+- Set the currently active span into a new context, and make that the implicit context. This is equivalent to getting the implicit context, copying it, and then inserting the `Span` into the copy.
 
 All the above functionalities operate solely on the context API, and they MAY be
 exposed as either static methods on the trace module, or as static methods on a class
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -173,7 +173,7 @@ If the language has support for implicitly propagated `Context` (see
 the following functionality:
 
 - Get the currently active span from the implicit context. This is equivalent to getting the implicit context, then extracting the `Span` from the context.
-- Set the currently active span into a new context, and make that the implicit context. This is equivalent to getting the implicit context, copying it, and then inserting the `Span` into the copy.
+- Set the currently active span into a new context, and make that the implicit context. This is equivalent to combining the current implicit context's values with the `Span` to create a new context, which is then made the current implicit context.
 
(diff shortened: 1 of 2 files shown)


Grunet:
I'm not at all tied to the wording changes. Would very much appreciate feedback from more familiar folks

specification/trace/api.md
```diff
@@ -173,7 +173,7 @@ If the language has support for implicitly propagated `Context` (see
 the following functionality:
 
 - Get the currently active span from the implicit context. This is equivalent to getting the implicit context, then extracting the `Span` from the context.
-- Set the currently active span to the implicit context. This is equivalent to getting the implicit context, then inserting the `Span` to the context.
+- Set the currently active span into a new context, and make that the implicit context. This is equivalent to getting the implicit context, copying it, and then inserting the `Span` into the copy.
```
yurishkuro, line 176: it seems this still has the same issue though, because of "inserting into copy" wording, which is incorrect (and similar to L166) because inserting and copying happen as a single step.

Grunet: Yup good point. I switched it to "combining the current implicit context's values with the `Span` to create a new context, which is then made the current implicit context" to follow the sense of the wording at the start of the Context doc - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/README.md#overview

I also updated L166 to try and match that.

Does that seem like it's on the right track?

yurishkuro (approved):

arminru (approved):

specification/trace/api.md
```diff
@@ -173,7 +173,7 @@ If the language has support for implicitly propagated `Context` (see
 the following functionality:
 
 - Get the currently active span from the implicit context. This is equivalent to getting the implicit context, then extracting the `Span` from the context.
-- Set the currently active span to the implicit context. This is equivalent to getting the implicit context, then inserting the `Span` to the context.
+- Set the currently active span into a new context, and make that the implicit context. This is equivalent to combining the current implicit context's values with the `Span` to create a new context, which is then made the current implicit context.
```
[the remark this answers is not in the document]
Grunet: Yup, added one in

PR #2644, opened 2022-07-05
Only monotonic sum are counters (with _total)

gouthamve:
Gauges shouldn't have a `_total` attached to them according to OpenMetrics.


pirgeo:
There is a separate section on Gauges (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md#gauges), which doesn't specify the `_total`. Can Prometheus only accept monotonic sums? If yes, should non-monotonic sums be exported as gauge?

gouthamve:
Yes, a non-monotonic sum is coverted to a Gauge :)

cijothomas (approved):

srikanthccv (approved):

pirgeo (approved):

jmacd (approved):

PR #2645, opened 2022-07-05
[OM/OTLP] Use _created for StartTimeUnixNano and vice-versa

gouthamve:
## Changes

OpenMetrics has the `_created` metric which is very similar to `StartTimeUnixNano`, convert one into the other.

### DIFF
--- a/specification/metrics/data-model.md
+++ b/specification/metrics/data-model.md
@@ -1259,7 +1259,7 @@ The following Prometheus types MUST be dropped:
 
 #### Start Time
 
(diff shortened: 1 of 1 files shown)


jmacd (approved):
Nice! This appears to create work for the OTel collector's Prometheus receiver and PRW exporter, I assume you will follow up with issues to track that work.

jsuereth (approved):

gouthamve:
Hey, yep, I will create the follow up issues in collector-contrib once this is merged. I've rebased it to remove the conflict.

reyang (approved):

PR #2648, opened 2022-07-06
Fix metrics SDK table of contents

jack-berg:
Fixing a typo in the metrics SDK headings.

### DIFF
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -21,7 +21,7 @@ linkTitle: SDK
     + [Default Aggregation](#default-aggregation)
     + [Sum Aggregation](#sum-aggregation)
     + [Last Value Aggregation](#last-value-aggregation)
-      - [Histogram Aggregation common behavior](#histogram-aggregation-common-behavior)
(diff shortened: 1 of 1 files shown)


cijothomas (approved):

tigrannajaryan (approved):

specification/metrics/sdk.md
```diff
@@ -410,7 +410,7 @@ This Aggregation informs the SDK to collect:
 - The last `Measurement`.
 - The timestamp of the last `Measurement`.
 
-##### Histogram Aggregation common behavior
+#### Histogram Aggregation common behavior
```
pirgeo, line 413: Would it make sense to call this heading `Histogram Aggregation` and then put Explicit Bucket and Exponential histograms as subheadings under this one? I feel like it's easy to miss the common behavior section when looking for specifics on the Explicit Bucket histogram, for example.

MrAlias: This sounds like a good idea, but it seems like something for an additional PR.

arminru: ^ #2656 was opened to track this

PR #2653, opened 2022-07-07
Introduce "split" metric schema transformation

tigrannajaryan:
This is a new transformation type that allows to describe a change
where a metric is converted to several other metrics by eliminating
an attribute.

An example of such change that happened recently is this:
https://github.com/open-telemetry/opentelemetry-specification/pull/2617

This PR contains:
- A new "split" transformation added to the schema file format.
- Schema file format incremented from 1.0.0 to 1.1.0. This is considered
  a backward compatible change and thus the minor version number is
  incremented.
- Deleted incorrect sentence that claimed that the addition of a new
  transformation type is not backward compatible and requires a
  major version number change. This is incorrect. Addition of
  new transformation types is backward compatible according to our
  current definition of backward compatibility, which is:
  "consumers can consume version X.Y and older versions X.Z
  provided that they are aware of version X.Y".
- Deleted the requirement that a full OTEP is always necessary to
  introduce new transformation types. This seems excessive. I think
  in simple cases like this a PR directly in the spec is sufficient.

This draft PR shows how the corresponding implementation of schema file
parser in Go can be done:
https://github.com/open-telemetry/opentelemetry-go/pull/2999

This commit in my personal repo shows in more detail how the "split"
transformation can be implemented:
https://github.com/tigrannajaryan/telemetry-schema/commit/5687ddb82307cbea4407e277cf94d05f8f587ff7

### DIFF
--- a/specification/schemas/file_format_v1.1.0.md
+++ b/specification/schemas/file_format_v1.1.0.md
@@ -1,0 +1,576 @@
+# Schema File
+
+**Status**: [Experimental](../document-status.md)
+
+A Schema File is a YAML file that describes the schema of a particular version.
+It defines the transformations that can be used to convert the telemetry data
+represented in any other older compatible version of the same schema family to
+this schema version.
+
+## File Structure
+
+Here is the structure of the Schema File:
+
+```yaml
+# Defines the file format. MUST be set to 1.1.0.
+file_format: 1.1.0
+
+# The Schema URL that this file is published at. The version number in the URL
+# MUST match the highest version number in the "versions" section below.
+# Note: the schema version number in the URL is not related in any way to
+# the file_format setting above.
+schema_url: https://opentelemetry.io/schemas/1.2.0
+
+# Definitions for each schema version in this family.
+# Note: the ordering of versions is defined according to semver
+# version number ordering rules.
+versions:
+  <version_number_last>:
+    # definitions for this version. See details below.
+
+  <version_number_previous>:
+    # definitions for previous version
+    ...
+  <version_number_first>:
+    # Defines the first version.
+```
+
+Each `<version_number>` section has the following structure:
+
+```yaml
+  <version_number>:
+    all:
+      changes:
+        # sequence of transformations.
+
+    resources:
+      changes:
+        # sequence of transformations.
+
+    spans:
+      changes:
+        # sequence of transformations.
+
+    span_events:
+      changes:
+        # sequence of transformations.
+
+    metrics:
+      changes:
+        # sequence of transformations.
+
+    logs:
+      changes:
+        # sequence of transformations.
+```
+
+There are 6 sub-sections under each version definition: "all", "resources",
+"spans", "span_events", "metrics", "logs". The last 5 sub-sections in this list
+contain definitions that apply only to the corresponding telemetry data type.
+Section "all" contains definitions that apply to all types of telemetry data.
+
+Below we describe each section in detail.
+
+### all Section
+
+"all" section in the schema file defines transformations. It must contain a
+sub-section named "changes" that defines how attributes were renamed from the
+previous version to this version.
+
+The "changes" section is a sequence of transformations. Only one transformation
+is supported for section "all": "rename_attributes" transformation.
+
+"rename_attributes" transformation requires a map of key/value pairs, where the
+key is the old name of the attribute used in the previous version, the value is
+the new name of the attribute starting from this version. Here is the structure:
+
+```yaml
+    all:
+      changes:
+        - rename_attributes:
+            attribute_map:
+              # map of key/values.
+```
+
+The transformations in section "all" apply to the following telemetry data:
+resource attributes, span attributes, span event attributes, log attributes,
+metric attributes.
+
+Important: when converting from the previous version to the current version the
+transformation sequence in section "all" is performed first. After that the
+transformations in the specific section ("resources", "spans", "span_events",
+"metrics" or "logs") that correspond to the data type that is being converted
+are applied.
+
+Note that "rename_attributes" transformation in most cases is reversible. It is
+possible to apply it backwards, so that telemetry data is converted from this
+version to the previous version. The only exception is when 2 or more different
+attributes in the previous version are renamed to the same attribute in the new
+version. In that case the reverse transformation is not possible since it would
+be ambiguous. When the reverse transformation is not possible it is considered
+an incompatible change. In this case the MAJOR version number of the schema
+SHOULD be increased in the new version.
+
+### resources Section
+
+"resources" section is very similar in its structure to "all". Like section
+"all" the transformations in section "resources" may contain only
+"rename_attributes" transformation.
+
+The only difference from section "all" is that this transformation is only
+applicable to Resource data type.
+
+Here is the structure:
+
+```yaml
+    resources:
+      changes:
+        - rename_attributes:
+            attribute_map:
+              # map of key/values. The keys are the old attribute name used
+              # the previous version, the values are the new attribute name
+              # starting from this version.
+```
+
+### spans Section
+
+"spans" section in the schema file defines transformations that are applicable
+only to Span data type. It must contain a sub-section named "changes" that
+defines a sequence of actions to be applied to convert Spans from the previous
+version to this version.
+
+One transformation is supported for section "span":  "rename_attributes".
+
+#### rename_attributes Transformation
+
+This is similar to the "rename_attributes" transformation supported in "all" and
+"resource" sections. In addition it is also possible to optionally specify spans
+that the transformation should apply to. Here is the structure:
+
+```yaml
+    spans:
+      changes:
+        - rename_attributes:
+            attribute_map:
+              # map of key/values. The keys are the old attribute name used
+              # in the previous version, the values are the new attribute name
+              # starting from this version.
+
+            apply_to_spans:
+              # Optional. If it is missing the transformation is applied
+              # to all spans. If it is present the transformation is applied
+              # only to the spans with the name that is found in the sequence
+              # specified below.
+```
+
+### span_events Section
+
+"spans_events" section in the schema file defines transformations that are
+applicable only to Span's Event data type. It must contain a sub-section named
+"changes" that defines a sequence of actions to be applied to convert events
+from the previous version to this version.
+
+Two transformations are supported for section "spans_events": "rename_events"
+and "rename_attributes".
+
+#### rename_events Transformation
+
+This transformation allows to change event names. It is applied to all events or
+only to events of spans that have the specified name. Here is the structure:
+
+```yaml
+    span_events:
+      changes:
+        - rename_events:
+            name_map:
+              # The keys are old event name used in the previous version, the
+              # values are the new event name starting from this version.
+```
+
+#### rename_attributes Transformation
+
+This is similar to the "rename_attributes" transformation supported in "all" and
+"resource" sections. In addition it is also possible to optionally specify spans
+and events that the transformation should apply to (both optional conditions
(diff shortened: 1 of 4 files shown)


specification/schemas/file_format_v1.1.0.md
```diff
@@ -12,8 +12,8 @@ this schema version.
 Here is the structure of the Schema File:
 
 ```yaml
-# Defines the file format. MUST be set to 1.0.0.
-file_format: 1.0.0
+# Defines the file format. MUST be set to 1.1.0.
```
Aneurysm9, line 15: Should we retain `file_format_v1.0.0.md` for posterity?  Or should this be `file_format_v1.md` and always represent the latest definition of the `v1` schema?

specification/schemas/file_format_v1.1.0.md
```diff
@@ -353,19 +374,22 @@ change according to the following rules:
   in a backward compatible manner. "Backward compatible" in this context means
   that consumers that are aware of the new MINOR number can consume the file of
   a particular MINOR version number or of any MINOR version number lower than
-  that, provided that MAJOR version numbers are the same. Typically, this means
-  that the added setting in file format is optional and the default value of the
-  setting matches the behavior of the previous file format version.
+  that, provided that MAJOR version numbers are the same. This can happen for
+  example when:
+
+  - A new transformation type is added.
```
Aneurysm9, line 380: Do we care about forward compatibility here?  What happens when a schema transform processor that doesn't understand the new transformation type is used with a schema of this version?  Should it refuse to process? Process those transformations it does know about?

specification/schemas/file_format_v1.1.0.md
tigrannajaryan: I was on the fence about this and eventually decide against it just to make the diff easier to understand. I can restore file_format_v1.0.0.md. I think it is a bit better to preserve the history, but will make a bit more difficult to understand what has changed between the versions since github won't show the diff automatically anymore.

specification/schemas/file_format_v1.1.0.md
tigrannajaryan: This is specifically called out in the spec.

```
there is no "forward compatibility" based on MINOR version number. Consumers which support reading up to a particular MINOR version number SHOULD NOT attempt to consume files with higher MINOR version numbers.
```

So, it means it should refuse to process. The entire schema file should be refused. It should not process just the transformations it knows about since that can be dangerous and result in a wrong state.

specification/schemas/file_format_v1.1.0.md
```diff
@@ -527,6 +551,20 @@ versions:
...
+            # Example from the change done by https://github.com/open-telemetry/opentelemetry-specification/pull/2617
+            # Name of old metric to split.
+            apply_to_metric: system.paging.operations
+            # Name of attribute in the old metric to use for splitting. The attribute will be
+            # eliminated, the new metric will not have it.
+            by_attribute: direction
+            # Names of new metrics to create, one for each possible value of attribute.
+            attributes_to_metrics:
+              # If "direction" attribute equals "in" create a new metric called "system.paging.operations.in".
+              in: system.paging.operations.in
```
tigrannajaryan, line 564: Should we reverse the keys and values in this map? It would allow to specify attribute values which are non-primitives (in theory we allow arrays as attribute values).

tigrannajaryan: keys/value reversed now.

tigrannajaryan:

tigrannajaryan:
I believe we need this before 1.13 release so that the schema file for 1.13 includes the change introduced by https://github.com/open-telemetry/opentelemetry-specification/pull/2617

jsuereth:
Meta-comment - Should we be allowing semantic convention changes *prior* to defining the compatibility transformation ?  (@tigrannajaryan).  

I feel like this should have been a prerequisite to #2617 using `split`

tigrannajaryan:
> Meta-comment - Should we be allowing semantic convention changes _prior_ to defining the compatibility transformation ?

Ideally, no. However, it is a chicken and egg problem. Unless there is a known example of a change in the semantic convention I am very reluctant with introducing the transformation in the schema file since I will have a hard time justifying it.

tigrannajaryan:
> I feel like this should have been a prerequisite to #2617 using `split`

So, I guess what you are saying we could have blocked #2617 until the "split" is introduced in the schema file.

tigrannajaryan:

tigrannajaryan:
@codeboten please review.

jsuereth:
> So, I guess what you are saying we could have blocked https://github.com/open-telemetry/opentelemetry-specification/pull/2617 until the "split" is introduced in the schema file.

@tigrannajaryan Yes, that's what I'm suggesting.  It'd be important for us to FIRST discuss if "split" is a compatibility primitive we wish to support before merging a change that uses it.

specification/schemas/file_format_v1.1.0.md
```diff
@@ -0,0 +1,576 @@
...
+
+#### split Transformation
+
+This transformation splits a metric into several metrics and eliminates an attribute.
+Here is the structure:
+
+```yaml
+    metrics:
+      changes:
+        - split:
```
jsuereth, line 267: I'm a bit worried about this operation here.   It seems highly targeted at the exact use case of ONE CL.


If we think about long-term maintenance of these transformation, I think we should consider leveraging something like a stream-time language for altering metrics.


Specifically, something like:

```
metrics:
  changes:
    split:
      apply_to_metric: my.bad.metric
      transformations:
        my.bad.metric.up: 'DROP direction WHERE direction=up'
        my.bad.metric.down: 'DROP direction WHERE directoin=down'
```

It's a bit too late now to really think through metric transformations abstractly (and I know there's been a lot of discussion around this in the collector).  My hope here is that we can stave too many of these ad-hoc single-use-case transformations for something more general, as we're very likely to need it (evidence: Prometheus relabelling + rewrite rules).

jsuereth (commented):
Generally I think this is 'ok' but I'm worried about long term direction.  Additionally, do we have an implementation of this transformation available?

tigrannajaryan:
> Generally I think this is 'ok' but I'm worried about long term direction. Additionally, do we have an implementation of this transformation available?

As a proof of concept, yes, here: https://github.com/tigrannajaryan/telemetry-schema/commit/5a5d1a2a30f0a776a6a2171af38faee80b877f39

Production-grade? Not yet. Work is in progress to introduce a schematransform processor in the Collector.

specification/schemas/file_format_v1.1.0.md
tigrannajaryan: > I'm a bit worried about this operation here. It seems highly targeted at the exact use case of ONE CL.

Arguably, it is not one use case. I have seen this transformation earlier internally when looking at different schemas. The particular change here in Otel was the trigger. "split" is one of the transform types I met in the past. A more complete list can be seen here: https://github.com/tigrannajaryan/telemetry-schema/blob/main/schema/testdata/schema-example.yaml
Unsurprisingly there is also a "merge" which is the reverse of "split", plus a few others. I intentionally excluded these from the initial version of Schema Files, since I wanted a stronger evidence that Otel actually needs them.

> If we think about long-term maintenance of these transformation, I think we should consider leveraging something like a stream-time language for altering metrics.

The transformations must be easy to implement efficiently. They also need to be reversible as much as possible (to allow version downgrades). A stream-like language will be very difficult to fit into these requirements. Not impossible, but much harder. So, possible something we can add in the future as another type of transformation when it becomes a necessity.

jmacd: See also #2675.

> The transformations must be easy to implement efficiently. They also need to be reversible as much as possible

This is a good explanation for the use of a dedicated structure, to me.

jmacd (approved):

specification/schemas/file_format_v1.1.0.md
```diff
@@ -0,0 +1,576 @@
...
+            # Name of old metric to split.
+            apply_to_metric: 
+            # Name of attribute in the old metric to use for splitting. The attribute will be
+            # eliminated, the new metric will not have it.
+            by_attribute: 
+            # Names of new metrics to create, one for each possible value of attribute.
+            attributes_to_metrics:
```
jmacd, line 274: The name "attributes_to_metrics" is somehow backwards because this is a map from metric names to attribute values. I wonder if the name "metrics_from_attributes" would be more explanatory?

specification/schemas/file_format_v1.1.0.md
```diff
@@ -0,0 +1,576 @@
...
+
+### metrics Section
+
+"metrics" section in the schema file defines transformations that are applicable
+only to Metric data type. It must contain a sub-section named "changes" that
+defines a sequence of actions to be applied to convert metrics from the previous
+version to this version.
+
+Two transformations are supported for section "metrics": "rename_metrics" and
+"rename_attributes".
```
codeboten, line 222:
```suggestion
Three transformations are supported for section "metrics": "rename_metrics",
"rename_attributes" and "split".
```

specification/schemas/file_format_v1.1.0.md
```diff
@@ -0,0 +1,576 @@
...
+            # Example from the change done by https://github.com/open-telemetry/opentelemetry-specification/pull/2617
+            # Name of old metric to split.
+            apply_to_metric: system.paging.operations
+            # Name of attribute in the old metric to use for splitting. The attribute will be
+            # eliminated, the new metric will not have it.
+            by_attribute: direction
+            # Names of new metrics to create, one for each possible value of attribute.
+            attributes_to_metrics:
+              # If "direction" attribute equals "in" create a new metric called "system.paging.operations.in".
+              system.paging.operations.in: in
```
codeboten, line 563: based on the name `attributes_to_metrics` I would expect the order of the configuration here to be reversed

tigrannajaryan: Renamed to `metrics_from_attributes`.

specification/schemas/file_format_v1.1.0.md
tigrannajaryan: Done.

specification/schemas/file_format_v1.1.0.md
tigrannajaryan: Yes, renamed to `metrics_from_attributes`.

jsuereth (approved):

codeboten (approved):

specification/schemas/file_format_v1.1.0.md
tigrannajaryan: The spec [touches it here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/file_format_v1.0.0.md#order-of-transformations), but we probably need to make it clearer. I have also just filed another related issue https://github.com/open-telemetry/opentelemetry-specification/issues/2680

jack-berg (approved):

specification/schemas/file_format_v1.1.0.md
```diff
@@ -0,0 +1,576 @@
...
+              # Optional. If it is missing the transformation is applied
+              # to all metrics. If it is present the transformation is applied
+              # only to the metrics with the name that is found in the sequence
+              # specified below.
+              - system.cpu.utilization
+              - system.memory.usage
+              - system.memory.utilization
+              - system.paging.usage
+
+        - split:
```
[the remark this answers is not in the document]
tigrannajaryan: >the door is probably closed on doing anything about it.

Yes, unfortunately too late for that. It can be done in Schema File Format 2.x.

tigrannajaryan:
I think we have the necessary number of approvals but I will keep this open more a bit more in case other @open-telemetry/specs-metrics-approvers want to review.

PR #2658, opened 2022-07-11
Make periodic exporting MetricReader env vars stable

pellared:
## What

Mark `OTEL_METRIC_EXPORT_INTERVAL`, `OTEL_METRIC_EXPORT_TIMEOUT` environment variables as stable.

## Why

I think this change is representing the current state as all of these env vars are already implemented in OTel PHP, OTel Python, and partially by OTel Java.

See:
- [`OTEL_METRIC_EXPORT_INTERVAL` usage](https://github.com/search?p=1&q=org%3Aopen-telemetry+OTEL_METRIC_EXPORT_INTERVAL&type=Code)
- [`OTEL_METRIC_EXPORT_TIMEOUT` usage](https://github.com/search?p=1&q=org%3Aopen-telemetry+OTEL_METRIC_EXPORT_TIMEOUT&type=Code)

Created because of: https://github.com/open-telemetry/opentelemetry-dotnet/pull/3424

### DIFF
--- a/specification/sdk-environment-variables.md
+++ b/specification/sdk-environment-variables.md
@@ -237,7 +237,7 @@ Known values for `OTEL_LOGS_EXPORTER` are:
 
 ## Metrics SDK Configuration
 
-**Status**: [Stable](document-status.md)
+**Status**: [Mixed](../../document-status.md)
 
 | Name            | Description | Default | Notes |
 |-----------------|---------|-------------|---------|
@@ -250,6 +250,11 @@ Known values for `OTEL_METRICS_EXEMPLAR_FILTER` are:
 - `"with_sampled_trace"`: Only allow measurements with a sampled parent span in context.
 
 ### Periodic exporting MetricReader
+
+**Status**: [Stable](document-status.md)
+
+Environment variables specific for the push metrics exporters (OTLP, stdout, in-memory)
+that use [periodic exporting MetricReader](metrics/sdk.md#periodic-exporting-metricreader).
 
 | Name                          | Description                                                                   | Default | Notes |
 | ----------------------------- | ----------------------------------------------------------------------------- | ------- | ----- |
--- a/specification/sdk-environment-variables.md
+++ b/specification/sdk-environment-variables.md
@@ -237,7 +237,7 @@ Known values for `OTEL_LOGS_EXPORTER` are:
 
 ## Metrics SDK Configuration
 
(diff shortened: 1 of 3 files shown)


specification/sdk-environment-variables.md
```diff
@@ -237,7 +237,7 @@ Known values for `OTEL_LOGS_EXPORTER` are:
 
 ## Metrics SDK Configuration
 
-**Status**: [Experimental](document-status.md)
+**Status**: [Stable](document-status.md)
```
reyang, line 240: The `OTEL_METRIC_EXPORT_INTERVAL` and `OTEL_METRIC_EXPORT_TIMEOUT` for Periodic exporting MetricReader might be misleading. For traces, `OTEL_BSP_SCHEDULE_DELAY`, `OTEL_BSP_EXPORT_TIMEOUT` are used for a Batch Span Processor.

In addition, `OTEL_METRIC_EXPORT_INTERVAL` could be misleading if a configured exporter/reader is using pull mode (e.g. Prometheus) - imagine some operator trying to increase the Prometheus scraping frequency and got really puzzled by this environment variable.

pellared:
PTAL @open-telemetry/java-instrumentation-maintainers @open-telemetry/python-maintainers @open-telemetry/php-approvers @open-telemetry/wg-prometheus

specification/sdk-environment-variables.md
pellared: Do you think the names should be changed or should we just make the description+notes more precise?

reyang: > Do you think the names should be changed or should we just make the description+notes more precise?

I think description/notes.

reyang (changes requested):
The `OTEL_METRICS_EXEMPLAR_FILTER` is still in very early stage and should remain as Experimental, consider marking the "Metrics SDK Configuration" status as Mixed, and make the "Periodic exporting MetricReader" section Stable.

In addition, consider making an update in https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md#environment-variables. Doesn't have to be in this PR though.

specification/sdk-environment-variables.md
pellared: @reyang did my best to address it. PTAL

reyang (approved):

cijothomas (approved):

PR #2663, opened 2022-07-14
add definition for state attribute of system.network.connections

### LINKED ISSUE #2623: Improve the definition of `state`

Currently the definition for the `state` attribute of the `system.network.connections` metric really only applies to TCP connections. It should either be defined in a way that makes it meaningful to multiple protocols or the metric should be separated per protocol.

**Additional context.**

This issue is a follow up to the discussion here: https://github.com/open-telemetry/opentelemetry-specification/pull/2617#issuecomment-1159163579

codeboten:
Fixes #2623

## Changes
Replacing wikipedia link with set values for the state. I would add that the states listed here only apply to TCP protocol, which reinforces my thinking that the system.network.connections metric should be split by protocol. Asking reviewers to consider this.

If folks agree, I will open a follow up PR to mark `system.network.connections` as deprecated in favour of: 
* `system.network.tcp.connections`
* `system.network.udp.connections`


jmacd (approved):

reyang (approved):

specification/metrics/semantic_conventions/system-metrics.md
```diff
@@ -148,7 +148,7 @@ perf counter (similar for Writes)
...
 |                                                 |                                                                                           |               |                                                   |            | protocol      | tcp, udp, [etc.](https://en.wikipedia.org/wiki/Transport_layer#Protocols)                      |
-|                                                 |                                                                                           |               |                                                   |            | state         | [e.g. for tcp](https://en.wikipedia.org/wiki/Transmission_Control_Protocol#Protocol_operation) |
+|                                                 |                                                                                           |               |                                                   |            | state         | if specified, SHOULD be one of: close, close_wait, closing, delete, established, fin_wait_1, fin_wait_2, last_ack, listen, syn_recv, syn_sent, time_wait. A stateless protocol must not set this attribute. |
```
arminru, line 151:
```suggestion
|                                                 |                                                                                           |               |                                                   |            | state         | If specified, SHOULD be one of: close, close_wait, closing, delete, established, fin_wait_1, fin_wait_2, last_ack, listen, syn_recv, syn_sent, time_wait. A stateless protocol MUST NOT set this attribute. |
```

arminru (approved):

PR #2665, opened 2022-07-15
Add optional Zero Threshold for Exponential Histograms to the metrics data model

gouthamve:
This is a follow up of https://github.com/open-telemetry/opentelemetry-specification/issues/2611 and the [Prometheus WG call](https://docs.google.com/document/d/19bnXziPn2MZ9wO6684UoI4D-LCjGL5bTJkGhux29bx8/edit) from June 22, 2022.

## Changes

This discusses the `zero_threshold` field that exists and is used in Prometheus Exponential Histograms and it would be nice to include it in OTel Exponential Histograms as well.

You can see the reference implementation here:
https://github.com/prometheus/prometheus/blob/5937b4f5d4f658758f4aa6b9ab803f022e6e5497/model/histogram/float_histogram.go#L186-L197

https://github.com/prometheus/prometheus/blob/5937b4f5d4f658758f4aa6b9ab803f022e6e5497/model/histogram/float_histogram.go#L770-L790

Prometheus also uses it (increases `zero_threshold`) to limit the number of buckets: https://github.com/prometheus/client_golang/blob/5a321c7b05264e54d2dbc86f12d9b80d4e7a492b/prometheus/histogram.go#L773-L808

[DDSketch](https://www.vldb.org/pvldb/vol12/p2195-masson.pdf) from Datadog uses something similar to limit the number of buckets (they expand the lowest bucket, not zero bucket but same principle).

Related issues #2611


specification/metrics/data-model.md
```diff
@@ -643,14 +643,22 @@ perfect subsetting.
...
-#### Zero Count
+#### Zero Count and Zero Threshold
 
-The ExponentialHistogram contains a special `zero_count` field
-containing the count of values that are either exactly zero or within
-the region considered zero by the instrumentation at the tolerated
-level of precision.  This bucket stores values that cannot be
-expressed using the standard exponential formula as well as values
-that have been rounded to zero.
+The ExponentialHistogram contains special `zero_count` bucket and an optional
```
beorn7, line 648: …contains a special…

specification/metrics/data-model.md
```diff
@@ -643,14 +643,22 @@ perfect subsetting.
...
+`zero_threshold` fields where `zero_count` contains the count of values whose
```
beorn7, line 649: fields → field

specification/metrics/data-model.md
```diff
@@ -643,14 +643,22 @@ perfect subsetting.
...
+for the `zero_threshold` is arbitrary and not related to the scale.
+
+When `zero_threshold` is unset or `0`, this bucket stores values that cannot
+be expressed using the standard exponential formula as well as values that
+have been rounded to zero.
+
+Histograms with different `zero_threshold` can still be merged easily by
+taking the largest `zero_threshold` of all involved Histograms and merge the
+lower buckets of Histograms with a smaller `zero_threshold` into the common
+wider zero bucket. In special cases, a wider zero bucket could be used to
```
beorn7, line 660: Merging gets slightly more complicated if the larger `zero_threshold` ends up in the middle of a populated bucket of the other histogram. In this case, the `zero_threshold` needs to be increased even further to match the upper boundary of the bucket it ended up in. ("upper boundary" in terms of absolute values)

beorn7 (changes requested):

specification/metrics/data-model.md
```diff
@@ -643,14 +643,25 @@ perfect subsetting.
...
-The ExponentialHistogram contains a special `zero_count` field
-containing the count of values that are either exactly zero or within
-the region considered zero by the instrumentation at the tolerated
-level of precision.  This bucket stores values that cannot be
-expressed using the standard exponential formula as well as values
-that have been rounded to zero.
+The ExponentialHistogram contains a special `zero_count` bucket and an optional
+`zero_threshold` field where `zero_count` contains the count of values whose
+absolute value is less than or equal to `zero_threshold`. The precise value
+for the `zero_threshold` is arbitrary and not related to the scale.
```
jack-berg, line 651: It would be useful to include some notes on what implementers might expect to use as the `zero_threshold` for different strategies of computing the index. I.e. is it typically the minimum normal value? The minimum positive double value? Something else?

reyang: At minimum, `zero_threshold` must be a non-negative AND finite double (with value `0` representing UNSET).

I feel that leaving it UNSET by default might be less helpful (considering the caller might pass in a very tiny value by accident, which "taints" the entire histogram resolution until the start time got reset).

Maybe have the ` default zero_threshold` = maximum subnormal (`2.225073858507201E-308`, or `0 00000000000 1111111111111111111111111111111111111111111111111111` in binary) - so that all subnormal numbers (whether positive or negative) will go to zero_count?

jmacd: I prefer to have the default threshold be zero (unset), and I also prefer to have subnormals round up to the smallest normal value. (Personally, I think the programmer has a bug if they "accidentally" record an extremely small measurement, they should witness the consequences and adjust their code. IOW, I think zero threshold should be implemented by the programmer, i.e., use zero to mean zero).

I do not believe that protecting the user from the "taint" of only subnormal values by default is all that important, considering the taint-impact of small normal values such as 0x1p-1020. If we are going to specify a default for zero threshold, I think it should be more like 1e-9, which would avoid unrealistic timings even when measuring nanoseconds.

I'm happy with this proposal as far as the protocol and data model are concerned. @jack-berg's questions stand, however: what should we say in the SDK spec about this? The simple answer is that exponential histogram aggregations will accept an optional zero threshold parameter, this belongs in `sdk.md` around where the default max-size is specified. Separately, I think we may want to agree with Prometheus on how to specify other limits, e.g., on the value itself or on the scale used.

jmacd (approved):

specification/metrics/data-model.md
jack-berg: I'm happy enough with a default 0 threshold of zero and making it configurable parameter in the [exp histogram aggregation](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#exponential-bucket-histogram-aggregation). 

> Personally, I think the programmer has a bug if they "accidentally" record an extremely small measurement, they should witness the consequences and adjust their code. IOW, I think zero threshold should be implemented by the programmer, i.e., use zero to mean zero

Agreed. And making it a exp histogram configuration option gives users that didn't write the instrumentation a way to get around the bug by configuring views that override the default of zero. So there's an escape hatch that doesn't require code changes.

jmacd:
@gouthamve I changed the PR title to limit its scope to match what's currently reviewed. I think after this merges, we're free to add it in the protocol for OTLP v0.20. Separately, we'll update ./sdk.md to say that expo-histos will accept an option and will have a default behavior w.r.t. this new thing--this we can debate later.

PR #2670, opened 2022-07-16
Update SDK spec to the latest W3C Baggage spec version

### LINKED ISSUE #2639: Incompatibility between the W3C Baggage spec version used in the Resource spec and the Baggage spec

When defining the value for the `OTEL_RESOURCE_ATTRIBUTES` environment variable, the [resource specification links to](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#specifying-resource-information-via-an-environment-variable) the version [`fdc7a5c4f4a31ba2a36717541055e551c2b032e4`](https://github.com/w3c/baggage/blob/fdc7a5c4f4a31ba2a36717541055e551c2b032e4/baggage/HTTP_HEADER_FORMAT.md) of the W3C Baggage format, while the OpenTelemetry Baggage specification [doesn't specify any particular version](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/baggage/api.md#propagation).

This causes an incompatibly in how baggage values are handled between the two specifications because the latest W3C Baggage spec defines that:

> Any characters outside of the baggage-octet range of characters MUST be percent-encoded.
https://w3c.github.io/baggage/#value

But the version used by the Resource spec doesn't have this requirement.

The effect of this difference is that, when using a library that follows the OpenTelemetry Baggage spec (such as [`go.opentelemetry.io/otel/baggage`](https://pkg.go.dev/go.opentelemetry.io/otel/baggage)) to generate the value for the `OTEL_RESOURCE_ATTRIBUTES` environment variable, the baggage values will be encoded by the library but not decoded by the SDK.

**Additional context.**

@MrAlias noticed this difference in https://github.com/open-telemetry/opentelemetry-go/pull/2963.

lgfa29:
Fixes #2639

## Changes

Update the version of the W3C Baggage spec used by the Resource SDK for the `OTEL_RESOURCE_ATTRIBUTES` environment variable so it matches the version used by the [baggage propagation API](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/baggage/api.md?plain=1#L149).

The key difference is the requirement to percent-encode characters outside the `baggage-octate` range, and so this is now highlighted in the SDK spec. 

Full diff between the latest version and the version used before: 
https://github.com/w3c/baggage/compare/fdc7a5c4f4a31ba2a36717541055e551c2b032e4..main#diff-f65e951e116caf027fff48025b09977ad3539456579743d5c8018b9313484229


lgfa29:
I wasn't sure if this could be considered a significant change, and so I didn't update the `CHANGELOG.md` nor `spec-compliance-matrix.md`. But maybe the `OTEL_RESOURCE_ATTRIBUTES` row in the [Environment Variables](https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md#environment-variables) section should be updated since they don't percent-decode values?

jsuereth (approved):

MrAlias (approved):

reyang (approved):

lgfa29:
Thanks @jsuereth! 

> I think you should update CHANGELOG.md. For spec compliance matrix

I pushed a new commit with a CHANGELOG entry. I wasn't sure where to add it, so I placed it under the `Resource` section, but let me know if it should go somewhere else.

> I think you should ping maintainers to see if they need to update their compliance around parsing this attribute.

I'm not sure who the maintainers are, so I will open an issue on each SDK repo and link back to this PR to help maintainers keep track of the changes necessary for their own project.

PR #2672, opened 2022-07-18
Added additional code attribute for code.url

### LINKED ISSUE #2687: Added Code links to spans as an approved convention
It would be really useful for observability backends to have a consistent way to link to the code that generates spans. This would allow clickable links to span generating code, and therefore aid in debugging.

What did you expect to see?
A new attribute in the Semantic Conventions for a Trace that include a url that links to the code in a cloud system. The suggestion is `code.url` as the attribute, and guidance that it should be a valid Url.

**Additional context.**
None

martinjt:
This adds a new attribute to the `code` prefix to allow instrumentors to add links to their code in the spans. This is incredibly useful if you're using things like centrally managed git in the cloud (e.g. GitHub) to provide navigatable links to code.

fixes #2687


martinjt:
@svrnm as requested :)

arminru:
Thank you, @martinjt!
Please make sure to open an accompanying issue and add a changelog entry in your PR before you mark it as ready for review 🙂

semantic_conventions/trace/general.yaml
```diff
@@ -284,3 +284,9 @@ groups:
         brief: >
           The line number in `code.filepath` best representing the operation. It SHOULD point within the code unit named in `code.function`.
         examples: 42
+      - id: url
+        type: string
+        brief: >
+          A valid Url that points to the code that generated the span. It SHOULD contain the file and line number references
```
arminru, line 290:
```suggestion
          A valid Url that points to the code that generated the span. It SHOULD contain the file and line number references.
```

svrnm: A few thoughts:
* Following @arminru's comment below: should this call out that if the url points to a VCS, that the version of the used code should be provided and not some pointer to a moving url like "main"
* You say that it "points to the code that generated the span" and also It SHOULD contain the file and line number references, although for the line number the statement is "The line number in `code.filepath` best representing the operation." -- That's pedantic but that's not the same line number then? Maybe rephrase it to "A valid Url that SHOULD point to the file and line number references in a version control system or any other accessible system containing the code"
* I am wondering if there are alternatives to pointing to the code? Let's say you have a 3rd party solution without code access, could there be a useful URL as well?

plantfansam: > * Following @arminru's comment below: should this call out that if the url points to a VCS, that the version of the used code should be provided and not some pointer to a moving url like "main"

+1

plantfansam:
Just want to +1 this overall — seems super useful 😄

semantic_conventions/trace/general.yaml
martinjt: See, this was the dilemma, trying not to be too prescriptive about what the URL is. Does it need to be to a VCS? if it's a library, maybe that URL could be manually added to point to docs?

Should it be "If it's pointing to a VCS, it should contain the file and line number references"

martinjt: I've updated the description, let me know if that reads better.

specification/trace/semantic_conventions/span-general.md
```diff
@@ -315,4 +315,5 @@ about the span.
 | `code.namespace` | string | The "namespace" within which `code.function` is defined. Usually the qualified class or module name, such that `code.namespace` + some separator + `code.function` form a unique identifier for the code unit. | `com.example.MyHttpService` | Recommended |
 | `code.filepath` | string | The source code file name that identifies the code unit as uniquely as possible (preferably an absolute file path). | `/usr/local/MyApplication/content_root/app/index.php` | Recommended |
 | `code.lineno` | int | The line number in `code.filepath` best representing the operation. It SHOULD point within the code unit named in `code.function`. | `42` | Recommended |
+| `code.url` | string | A valid Url that points to the code that generated the span. If the Url points to a VCS, the link SHOULD contain the file and line number references, and SHOULD point to the individual revision of the code. | `https://github.com/open-telemetry/opentelemetry-specification/blob/afc47a5a7d41b3f468d2fb3fc2ddb0c7ad52fadc/specification/trace/semantic_conventions/span-general.md#L259` | Optional |
```
arminru, line 318: How is instrumentation generally supposed to provide the VCS URL? Can we provide any guidance/ implementation hints here?

arminru: Could it be more useful to just record the base URL of the VCS repository (which could be configured as a static attribute) and then have the backend combine it to the full URL using the `code.filepath` and `code.lineno`? Maybe we would need to introduce a `code.revision` then as well, however.

PR #2681, opened 2022-07-20
Allow LogProcessor to mutate log records

### LINKED ISSUE #2647: Allow logs to be mutated in LogProcessor
There's currently an inconsistency between [LogProcessor](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/logging-library-sdk.md#logprocessor) and the [SpanProcessor](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#span-processor) concept it was modeled after which severely limits the usefulness of LogProcessor. 

Multiple SpanProcessors can be registered on a TracerProvider, and each is invoked in the order they are registered. [SpanProcessor#onStart](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#onstart) accepts a read/write span object. This read/write span is different than the readable span accepted by SpanProcessor#OnEnd, which is again typically different from a third SpanData interface accepted by SpanExporter, as described [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#additional-span-interfaces). The effect of this design is that multiple processors can be registered, with upstream processors having the opportunity to enrich the read/write span with additional data extracted from context, baggage, etc, before a downstream processor batches up as SpanData and exports. 

In contrast, log processors have a single Emit method, which accepts a readable [LogData](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/logging-library-sdk.md#logdata) interface. LogData is the same interface accepted by [LogExporter](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/logging-library-sdk.md#logexporter). LogData is necessarily read only, since LogExporters need to have stable data to export. However, by having LogProcessor accept the same read only interface, there's effectively nothing useful you can do by registering multiple LogProcessors, because each processor has no ability to change the data.

I think its pretty clear that the intention of the Logging Library SDK spec is to mirror the trace SDK spec when possible. I propose that we fix this inconsistency as follows:

- Introduce a new read/write log interface to mirror read/write span. Read/write log should have a method to obtain a LogData.
- Change LogProcessor#emit to accept read/write log instead of LogData.
- The effect is that multiple LogProcessors can be registered with a LogEmitterProvider. When a log is emitted, the emit(read/write log) method of each LogProcessor is called in order of registration. Upstream LogProcessors can enrich the log with data extracted from context, baggage, etc. Downstream LogProcessor(s) (typically the last) can batch up LogData and export via LogExporter#export(List<LogData>).

jack-berg:
Resolves #2647.

The simplest possible fix would have been to simply change `LogProcessor#emit` to accept `LogRecord` instead of `LogData`, but I took the opportunity to add some additional detail to make logs more consistent with spans where possible. 

Changes include:

- Add TOC
- Rename LogProcessor#emit to LogProcessor#onEmit. This is more consistent with spans (SpanProcessor#onStart, SpanProcessor#onEnd), and avoids confusion by being different than LogEmitter#emit. 
- Define LogRecord interface, accepted by LogEmitter#emit
- Define Read/write LogRecord interface, accepted by LogProcessor#onEmit
- Define Readable LogRecord interface, accepted in batches by LogExporter#export
- Add LogExporter#forceFlush method (stubbed out for now) for consistency with SpanExporter#forceFlush


specification/logs/logging-library-sdk.md
```diff
@@ -62,45 +94,99 @@ TBD
...
   Note: some languages may opt to avoid having a LogRecord data type and instead
   use a more idiomatic builder pattern to prepare and emit a log record (see
   e.g.
-  [Java discussion](https://github.com/open-telemetry/opentelemetry-java/pull/3759#discussion_r738019425))
+  [Java discussion](https://github.com/open-telemetry/opentelemetry-java/pull/3759#discussion_r738019425)).
 
-### LogRecord
-
-See LogRecord [data model](data-model.md) for the list of fields.
+Open Question: do we need to also pass the Baggage so
```
reyang, line 159: I think no, LogProcessor should run inside the caller's context, and it have access to the general Context - which includes Baggage. If LogProcessor wants certain thing from Baggage, it can retrieve it.

reyang (approved):

specification/logs/logging-library-sdk.md
```diff
@@ -62,45 +94,99 @@ TBD
...
 Methods:
 
-- Emit(LogData). Call when a log record is ready to be processed and exported.
-- Shutdown.
-- ForceFlush.
+#### OnEmit
 
-Built-in implementations: SimpleLogProcessor, BatchLogProcessor.
+`OnEmit` is called when a log record is [emitted](#emit). This
+method is called synchronously on the thread that emitted the log record,
```
tigrannajaryan, line 175: We discussed the possibility of having an async processor today which does redaction in a background thread in the SIG meeting today. I am not sure how it is possible if the OnEmit is synchronous.

tigrannajaryan (approved):
LGTM, just one non-blocking question on sync/async behavior.

specification/logs/logging-library-sdk.md
jack-berg: The `onEmit` method is called synchronously, but it can perform work asynchronously. For example, it can synchronously add the log record to a queue, and have a background thread asynchronously poll from the queue and perform processing before passing to exporter.

specification/logs/logging-library-sdk.md
jack-berg: This was an existing open question that I've just relocated. I do agree tho and am happy enough to remove it as resolved if others agree.

specification/logs/logging-library-sdk.md
tigrannajaryan: How does this work if there are multiple processors and the first processor wants to asynchronously mutate the log records? Wouldn't this result in a race?

jack-berg: Yes that's correct. I think as a general rule for both trace and log processors, you want any asynchronous processing to happen last to avoid such races.

scheler: Wouldn't it still be an issue if there are multiple processors that run async at the end?

jack-berg: Yes.

jack-berg:
@open-telemetry/specs-logs-approvers please take a look!

jack-berg:
Pinging @open-telemetry/specs-logs-approvers. Please take a look! Otel java has already implemented and published this change [here](https://github.com/open-telemetry/opentelemetry-java/pull/4643).

specification/logs/logging-library-sdk.md
djaglowski: Do we need a resolution to this issue?

jack-berg: I think we just need to acknowledge it, not resolve it. This design is consistent with tracing, which is correct since log processing does not significantly differ from tracing. The consequence of this design is that if multiple async processors are required, the user must be more careful in their configuration, and use a pattern of the first processor doing its work and then delegating to next. Essentially, they must build the control flow into the processors rather than relying on the SDK.

djaglowski (approved):

specification/logs/logging-library-sdk.md
reyang: > I think we just need to acknowledge it, not resolve it. This design is consistent with tracing, which is correct since log processing does not significantly differ from tracing. The consequence of this design is that if multiple async processors are required, the user must be more careful in their configuration, and use a pattern of the first processor doing its work and then delegating to next. Essentially, they must build the control flow into the processors rather than relying on the SDK.

+1

bogdandrutu: I do believe that instead of LogData we probably should pass a ReadableLogRecord, to facilitate mutation by simply wrapping the previous object.

jack-berg: > I do believe that instead of LogData we probably should pass a ReadableLogRecord, to facilitate mutation by simply wrapping the previous object.

One of the core use cases here is that processors should be able to enrich the logs with context / baggage. Mutation by wrapping the previous object doesn't work unless we break the control flow pattern of traces. Specifically, the span processors are called sequentially by the SDK - its not the responsibility of the span processor to call the next in the chain. Wrapping the previous object only works if each processor is responsible for calling the next.

bogdandrutu: Then you can have an "OnCreate" hook (or other equivalent), called sync and allow mutation only during that function call. Allowing mutations async requires unnecessary sync/mutexs/etc. 

Also it s very fragile to do the "calls in order" and rely on that, since registration/unregistration of processor can happen at any moment and is 1 by 1 not as a list. So if you have a "mutating" processor I would strongly suggest users to build that as a chain processor, which accepts a next.

jack-berg: > Then you can have an "OnCreate" hook (or other equivalent), called sync and allow mutation only during that function call. Allowing mutations async requires unnecessary sync/mutexs/etc.

That's definitely an option. If such a method existed it could have a `ReadableLogRecord` response and the SDK could pass the response to the next processor. However logs current don't have any sort of clear start and end phase like traces do. Logs are just "emitted". You'd need to carefully define the start / onCreate phase since its definition dictates which fields could be mutated by processors.

Alternatively, a log processor could have a property indicating whether or not it mutates the log record. If it doesn't a lock free read only version can be provided. 

> since registration/unregistration of processor can happen at any moment

I suppose thats true since [TracerProvider MAY provide methods to update the configuration](https://github.com/open-telemetry/opentelemetry-specification/blame/main/specification/trace/sdk.md#L69) and its likely that the logger provider will follow suit. In practice, we currently don't allow reconfiguration in the java SDK. If we did, we would have to make limited guarantees about when logs / traces / metrics would receive the updated configuration since the APIs to collect those records are happening on different threads than the reconfiguration.

tigrannajaryan: >Alternatively, a log processor could return a boolean property indicating whether or not it mutates the log record. If it doesn't a lock free read only version can be provided.

Interestingly, this is the design we arrived to in the [Collector processors](https://github.com/open-telemetry/opentelemetry-collector/blob/e092fc728b724f862d37294dff03a602996f1dd2/consumer/consumer.go#L23): they declare whether they mutate data so that we know how to deal with data ownership. :-)

Issue #2501, opened 2022-04-19
OTLP exporters should add their version information as a resource attribute


codeboten:
As a result of the discussion in April 19th SIG meeting, it was suggested that it would be useful for debugging purposes that the OTLP exporter add version information to the resource attributes it emits.

bogdandrutu:
No, it is about the version of the exporter component. We decided intentionally to not add the version of the `opentelemetry-proto` to disallow backends to do business decisions based on the proto version. This information can be used only for metrics/debugging purpose.

alanwest:
Maybe worthy of a separate issue, but I think it also makes sense to add this information in the user-agent (or equivalent). Much in the same way that was done in the collector project https://github.com/open-telemetry/opentelemetry-collector/pull/3970.

PR #2684, opened 2022-07-21
add user agent to OTLP exporter spec

codeboten:
## Changes

The following PR adds a section to the OTLP exporter about setting a user agent. This was implemented in the otel collector and provides users with additional information to troubleshoot their telemetry pipelines with. This was discussed in the context of #2501. It's worth noting that the OpenTelemetry collector is currently setting a user agent in the format:

```
OpenTelemetry Collector Contrib/0.53.0 (linux/amd64) grpc-go/1.47.0
```

Additionally, existing exporters are already sending this header (using whatever the underlying library is setting), it's just not set to anything OpenTelemetry specific.

Related issues #2501


specification/protocol/exporter.md
```diff
@@ -160,4 +160,15 @@ Transient errors MUST be handled with a retry strategy. This retry strategy MUST
 
 For OTLP/HTTP, the errors `408 (Request Timeout)` and `5xx (Server Errors)` are defined as transient, detailed information about errors can be found in the [HTTP failures section](otlp.md#failures). For the OTLP/gRPC, the full list of the gRPC retryable status codes can be found in the [gRPC response section](otlp.md#otlpgrpc-response).
 
+## User Agent
+
+OpenTelemetry protocol exporters SHOULD emit a User-Agent header to at a minimum identify the exporter, the language of its implementation, and the version of the exporter. For example, the Python OTLP exporter version 1.2.3 would report the following:
+
+```
+OpenTelemetry Exporter Python/1.2.3
```
mwear, line 168: If a given language has exporters for multiple protocols (grpc, json over http, and proto over http) should the user agent reflect that?

codeboten: I did consider this, however after thinking through the common scenarios, I figured other headers would be enough to convey that information.

reyang, line 168: Does this follow a specific format so folks can expect to parse it?

reyang, line 168: "OpenTelemetry Exporter" or "OTLP Exporter"?

codeboten: I believe this should follow https://datatracker.ietf.org/doc/html/rfc7231#section-5.5.3

codeboten: I'd be happy with either. There may be a time where a similar header is added to the otel jaeger/zipkin exporters, in which case I would expect the following to make the most sense:

```
OpenTelemetry OTLP Exporter
OpenTelemetry Jaeger Exporter
OpenTelemetry Zipkin Exporter
```

carlosalberto: @codeboten Shall we add the mentioned link, for clarity purposes?

carlosalberto (approved):

specification/protocol/exporter.md
codeboten: Updated the text to show the complete example

codeboten: Added.

bogdandrutu:
I would make this an "option" maybe opt-out.

MSNev (approved):

codeboten:
> I would make this an "option" maybe opt-out.

@bogdandrutu today most (i believe at least all grpc and many http) OTLP exporters already have a user agent header, something like `grpc-go/1.47.0`, and there is no way to opt out. This PR is merely suggesting that the exporters *should* provide something more meaningful.

specification/protocol/exporter.md
```diff
@@ -160,4 +160,16 @@ Transient errors MUST be handled with a retry strategy. This retry strategy MUST
 
 For OTLP/HTTP, the errors `408 (Request Timeout)` and `5xx (Server Errors)` are defined as transient, detailed information about errors can be found in the [HTTP failures section](otlp.md#failures). For the OTLP/gRPC, the full list of the gRPC retryable status codes can be found in the [gRPC response section](otlp.md#otlpgrpc-response).
 
+## User Agent
+
+OpenTelemetry protocol exporters SHOULD emit a User-Agent header to at a minimum identify the exporter, the language of its implementation, and the version of the exporter. For example, the Python OTLP exporter version 1.2.3 would report the following:
```
jmacd, line 165: May I suggest the versions of OTLP that the exporter was compiled for, i.e., the OTLP version that was used to define the protobuf?

jmacd (approved):

reyang (approved):

specification/protocol/exporter.md
arminru: If consumed by a human, they can look up the respective SDK/exporter release and find the OTLP version it's using there, but this would not be possible vice-versa. You could, for example, have a very outdated SDK/exporter with (known) issues and a recent, fixed version of it, but both of them might use the same OTLP version (maybe OTLP 1.0.0 one day).

specification/protocol/exporter.md
```diff
@@ -160,4 +160,16 @@ Transient errors MUST be handled with a retry strategy. This retry strategy MUST
...
+
+```
+OpenTelemetry OTLP Exporter Python/1.2.3
+```
+
+The format of the header SHOULD follow [RFC 7231][rfc-7231]. The conventions used for specifying the language and version are available in the [Resource semantic conventions][resource-semconv].
```
arminru, line 171:
```suggestion
The format of the header SHOULD follow [RFC 7231][rfc-7231]. The conventions used for specifying the OpenTelemetry SDK language and version are available in the [Resource semantic conventions][resource-semconv].
```

arminru (approved):

specification/protocol/exporter.md
carlosalberto: @jmacd Is @arminru's take good enough for you? Or would you rather still have the OTLP version?

jmacd (approved):

carlosalberto:
@bogdandrutu Any further opinion on this?

specification/protocol/exporter.md
```diff
@@ -160,4 +160,16 @@ Transient errors MUST be handled with a retry strategy. This retry strategy MUST
 
 For OTLP/HTTP, the errors `408 (Request Timeout)` and `5xx (Server Errors)` are defined as transient, detailed information about errors can be found in the [HTTP failures section](otlp.md#failures). For the OTLP/gRPC, the full list of the gRPC retryable status codes can be found in the [gRPC response section](otlp.md#otlpgrpc-response).
 
+## User Agent
+
+OpenTelemetry protocol exporters SHOULD emit a User-Agent header to at a minimum identify the exporter, the language of its implementation, and the version of the exporter. For example, the Python OTLP exporter version 1.2.3 would report the following:
+
+```
+OpenTelemetry OTLP Exporter Python/1.2.3
```
tigrannajaryan, line 168: Nit: do we need `OpenTelemetry` in the user agent? OTLP seems to be sufficient.

tigrannajaryan (approved):

specification/protocol/exporter.md
jack-berg: Including "OpenTelemetry" might be helpful to differentiate exporters from OpenTelemetry published artifacts, versus 3rd party OTLP clients.

carlosalberto: That's right - IIRC that was one of the original reasons to explicitly include "OpenTelemetry".

bogdandrutu: Otel to shorter the bytes on wire? `Otel Exporter Python/1.2.3`?

OTLP is useless since that is the protocol?

specification/protocol/exporter.md
robbkidd: What do you think of including both exporter version and a separate user-agent stanza for OTLP version (for example `OpenTelemetry OTLP Exporter Python/1.2.3 OTLP/0.18.0 …`)?

This request for proto version in user_agent is in spirit with [an open OTEP](https://github.com/open-telemetry/oteps/pull/204/files). As a receiver of OTLP from a wide variety of sources not under our control, an Exporter reporting its identity _and_ protocol version somewhere in the request and outside the request body facilitates debugging transmission problems. Humans looking up build or dependency manifests feels … not observable?

tigrannajaryan: > This request for proto version in user_agent is in spirit with [an open OTEP](https://github.com/open-telemetry/oteps/pull/204/files).

I had [objections](https://github.com/open-telemetry/oteps/pull/204#discussion_r907348388) to the spirit of that OTEP that I still have, so I also object to adding the OTLP version number to the user agent string.

robbkidd: OK. I'll respond to the objections there!

specification/protocol/exporter.md
carlosalberto: @bogdandrutu I think we want to leave the door open to add this to other exporters, i.e. non OTLP exporters OTel may offer (_and_ keep things uniform). See [this comment](https://github.com/open-telemetry/opentelemetry-specification/pull/2684#discussion_r927046560)

Otherwise I think it's fine going with "OTel", i.e. "OTel OTLP Exporter Python/1.2.3".

carlosalberto: Ping @bogdandrutu - any opinion on this one?

codeboten: I've shortened the user agent to `Otel OTLP Exporter Python/1.2.3`.

@bogdandrutu my thinking around keeping OTLP in the user agent is to allow for other exporters (jaeger/zipkin) to use the same format for user agent strings, for example:

```
Otel OTLP Exporter Python/1.2.3
Otel Jaeger Exporter Python/1.2.3
Otel Zipkin Exporter Python/1.2.3
```

Thoughts?

reyang: > ```
> Otel OTLP Exporter Python/1.2.3
> ```

Regarding "Otel OTLP":
1. The current glossary is using "OTel" instead of "Otel" https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification#acronym. Should we avoid "Otel" with lowercase "t"?
2. "OTLP"'s "OTL" part stands for OpenTelemetry according to https://opentelemetry.io/docs/reference/specification/protocol/, "OTel OTLP" sounds duplicated.

jsuereth (approved):

CHANGELOG.md
```diff
@@ -79,6 +79,8 @@ release.
 
 - Add support for partial success in an OTLP export response
   ([#2696](https://github.com/open-telemetry/opentelemetry-specification/pull/2696)).
+- Add user agent to OTLP exporter specification
+  ([#2684](https://github.com/open-telemetry/opentelemetry-specification/pull/2684))
```
arminru, line 83: Please make sure this resides in the "Unreleased" section on top after rebasing onto main.

arminru (approved):

CHANGELOG.md
arminru: fixed in cb31f3a033ae281b000d0e3d8772cb270716bc8f

specification/protocol/exporter.md
codeboten: I've updated the casing for Otel -> OTel and for the reasons @arminru listed above, I think we should keep both OTel and OTLP in the user agent string. Even though it is repeating OpenTelemetry, each instance indicates different things. The first OpenTelemetry references the organization/project that's producing the exporter, the second references the protocol being emitted.

reyang (approved):
LGTM.

PR #2688, opened 2022-07-22
add schema definition for direction change

codeboten:
## Changes

This is a follow up to https://github.com/open-telemetry/opentelemetry-specification/pull/2617 now that https://github.com/open-telemetry/opentelemetry-specification/pull/2653 has been merged.

Related issues https://github.com/open-telemetry/opentelemetry-specification/issues/2589

### DIFF
--- a/schemas/1.13.0
+++ b/schemas/1.13.0
@@ -4,60 +4,83 @@ versions:
   1.13.0:
     metrics:
       changes:
+        # https://github.com/open-telemetry/opentelemetry-specification/pull/2617
         - split:
             apply_to_metric: system.paging.operations
             by_attribute: direction
             metrics_from_attributes:
                 system.paging.operations.in: in
                 system.paging.operations.out: out
+        # https://github.com/open-telemetry/opentelemetry-specification/pull/2617
         - split:
             apply_to_metric: system.disk.io
             by_attribute: direction
             metrics_from_attributes:
                 system.disk.io.read: read
                 system.disk.io.write: write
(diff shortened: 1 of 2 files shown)


codeboten:
Leaving this as a draft as schema check will fail until https://github.com/open-telemetry/build-tools/issues/106 is resolved

schemas/1.13.0
```diff
@@ -0,0 +1,82 @@
+file_format: 1.1.0
+schema_url: https://opentelemetry.io/schemas/1.13.0
+versions:
+  1.13.0:
+    metrics:
+      changes:
+        - split:
+            apply_to_metric: system.paging.operations
```
tigrannajaryan, line 8: I wonder if it makes sense to add a comment with a link to the PR which introduced the change. It would help with reviewing this schema.

schemas/1.13.0
```diff
@@ -0,0 +1,82 @@
+file_format: 1.1.0
+schema_url: https://opentelemetry.io/schemas/1.13.0
+versions:
+  1.13.0:
+    metrics:
+      changes:
```
tigrannajaryan, line 6: Do we also need to add process.disk.io and process.network.io splits?

codeboten: Yes. Thanks for catching this, added.

schemas/1.13.0
codeboten: I've added a comment above each change. I think this is a good practice to get into for future changes.

tigrannajaryan (approved):

jmacd (approved):

Issue #414, opened 2020-01-20
Define OpenTelemetry to Jaeger span Transformation


mauriciovasquezbernal:
Jaeger is one of the recommended exporters that the SDKs should support [1], however there is not a specification about how the transformation from the OpenTelemetry to the Jaeger Spans should work. 

It is already causing some confusion as the implementation in different languages is different, https://github.com/open-telemetry/opentelemetry-python/pull/367#issuecomment-575306955 & https://github.com/open-telemetry/opentelemetry-python/issues/369.

There is an open PR handling this for Zipkin https://github.com/open-telemetry/opentelemetry-specification/pull/380, probably there should also be a similar document for Jaeger.

[1] https://github.com/tigrannajaryan/opentelemetry-specification/blob/master/specification/library-guidelines.md#requirements.

andrewhsu:
from the issue triage mtg today with TC, allowing changes related to this issue for GA if editorial changes only

carlosalberto:
I'm going to close this one as we indeed added this but we removed it after some time (see #3964), i.e.

> Jaeger exporter support will be removed from OpenTelemetry in July 2023.

PR #2696, opened 2022-07-26
Add support for partial success in an OTLP export response [2]

### LINKED ISSUE #2454: Add ability for OTLP response to signal partial acceptance of data

joaopgrassi:
Fixes #2454, Closes #2636

This is a second take on #2636. It's basically the same changes, except it uses `rejected_<signal>` semantics instead of `accepted_<signal>` as discussed in the PRs and in the spec meeting on 7/19/2022.

### Changes

This PR adds guidance around how servers can signal a partial success to clients. It indicates in which situations servers can respond with a partial success and how the fields introduced in [opentelemetry-proto #414](https://github.com/open-telemetry/opentelemetry-proto/pull/414) should be populated.

I also refactored the `OTLP/gRPC Response` section a bit, to make it consistent with the HTTP one. The main inconsistency I found was that the gRPC section did not have clear sections for `Success`, `Failure` as the HTTP does. Now both share the same sections and the wording should be consistent between the two.

Related
Proto change: https://github.com/open-telemetry/opentelemetry-proto/pull/414

cc @tigrannajaryan @jmacd @jsuereth @jack-berg @reyang @yurishkuro


specification/protocol/otlp.md
```diff
@@ -520,12 +606,19 @@ received yet. The client will typically choose to re-send such data to guarantee
 delivery, which may result in duplicate data on the server side. This is a
 deliberate choice and is considered to be the right tradeoff for telemetry data.
 
-### Partial Success
+#### Partial Success Retry
+
+The partial success defined by the protocol is neither designed nor intended
```
joaopgrassi, line 611: Hi @jsuereth 

I tried adding your suggestion for this section from the previous PR (https://github.com/open-telemetry/opentelemetry-specification/pull/2636#discussion_r919136736). Let me know what you think.

specification/protocol/otlp.md
```diff
@@ -145,16 +149,59 @@ was not delivered.
...
+(`ExportTracePartialSuccess` message for traces,
+`ExportMetricsPartialSuccess` message for metrics and
+`ExportLogsPartialSuccess` message for logs), and it MUST set the respective
+`rejected_spans`, `rejected_data_points` or `rejected_log_records` field with
+the number of spans/data points/log records it rejected.
+
+A `partial_success` with the `rejected_<signal>` field holding a `0` value
+is an invalid result. In such cases, senders MUST ignore the `partial_success`
+field and handle such response from a server in the same way as defined in the
+[Success](#success) section.
```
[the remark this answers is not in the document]
joaopgrassi: Yes, this is something I would like to have as well. But with both ways (accepted/rejected) it doesn't seem to leave a space to communicate such cases. Leaving it with 0 _and_ adding a message would be something I'd like to have but I'm not sure if it does make sense.

specification/protocol/otlp.md
```diff
@@ -520,12 +606,19 @@ received yet. The client will typically choose to re-send such data to guarantee
...
+thus preventing retry loops.
 
-The protocol does not attempt to communicate partial reception success from the
-server to the client (i.e. when part of the data can be received by the server
-and part of it cannot). Attempting to do so would complicate the protocol and
-implementations significantly and is left out as a possible future area of work.
+The protocol does not attempt to define how clients should automatically retry
+a partially successful request.
+Attempting to do so would complicate the protocol and implementations
+significantly and is left out as a possible future area of work.
```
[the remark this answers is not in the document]
joaopgrassi: @jmacd Would you be in favor of removing this whole "limitation" section then? I think in the sections above is quite clear retries shouldn't be done so I'm fine in dropping this altogether. Less is more 😅.

My idea behind adding thiswas that some people expressed their desire for having some way of retrying things, but for now, this is left unspecified. So it lands under "current limitations" and it might be good to have it under section on the document.

specification/protocol/otlp.md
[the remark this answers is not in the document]
reyang: > Was it a partial success in this case? the line is quite thin

I suggest the following semantics:
* Success: you're all set for the batch of data you sent, if you try to resend any data in that batch, you're essentially wasting energy/time.
* Failure: everything in the batch of data you sent is getting dropped by the receiver, whether you should retry or not is depending on the actual type of failure
* Partial-success: anything that is not a Success AND not a Failure

If the server decided to do normalization and then take the data, there is no need for the client to resend, it should be Success (and maybe a Success can also have some indication of "suggestion" - e.g. "Hey we've received your data and you're all set. BTW we've noticed that you were sending the data via a less efficient/secure way, you should upgrade to a newer version of proto").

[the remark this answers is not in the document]
joaopgrassi: @reyang 

> it should be Success (and maybe a Success can also have some indication of "suggestion"

Right, I agree with this. But to be able to do that we need something:

1. A new field to the top-level `Export<signal>ServiceResponse` message type. E.g.  `additional_message`.

2. Change the types introduced in https://github.com/open-telemetry/opentelemetry-proto/pull/414 to not be called "partial success", so they can be used also in this situation - the server accepted everything but wants to send some information back. 

I wouldn't even consider option (2) because the `rejected_<signal>`  is useless in a success case. Option (1) is what I think we should have to keep a clear separation of things. It can also be added in a separated PR  (it's also a backwards compatible change)

specification/protocol/otlp.md
```diff
@@ -145,16 +149,59 @@ was not delivered.
 
 #### OTLP/gRPC Response
 
-The server may respond with either a success or an error to the requests.
+The response MUST be the appropriate serialized Protobuf message (see below for
+the specific message to use in the [Success](#success),
+[Partial Success](#partial-success) and [Failure](#failures) cases).
+
+##### Success
```
tigrannajaryan, line 156:
```suggestion
##### Full Success
```

specification/protocol/otlp.md
```diff
@@ -145,16 +149,59 @@ was not delivered.
...
+
+##### Partial Success
+
+If the request is only partially accepted
+(i.e. when the server accepts only parts of the data and rejects the rest), the
+server response MUST be a Protobuf-encoded
+[Export<signal>ServiceResponse](https://github.com/open-telemetry/opentelemetry-proto)
+message (`ExportTraceServiceResponse` for traces,
+`ExportMetricsServiceResponse` for metrics and
+`ExportLogsServiceResponse` for logs).
```
tigrannajaryan, line 181: Can we avoid repeating this? For example: `If the request is only partially accepted the same [Export<signal>ServiceResponse] message as in Full Success case is returned and additionally ...`

specification/protocol/otlp.md
```diff
@@ -145,16 +149,59 @@ was not delivered.
...
+
+Additionally, the server MUST initialize the `partial_success` field
+(`ExportTracePartialSuccess` message for traces,
+`ExportMetricsPartialSuccess` message for metrics and
+`ExportLogsPartialSuccess` message for logs), and it MUST set the respective
```
tigrannajaryan, line 186: We are using `Export<signal>ServiceResponse` elsewhere, so why not use it here is as well and avoid repetition.

specification/protocol/otlp.md
```diff
@@ -145,16 +149,59 @@ was not delivered.
...
-The success response indicates telemetry data is successfully processed by the
-server. If the server receives an empty request (a request that does not carry
+If the server receives an empty request (a request that does not carry
 any telemetry data) the server SHOULD respond with success.
 
-Success response is returned via
-[Export*ServiceResponse](https://github.com/open-telemetry/opentelemetry-proto)
-message (`ExportTraceServiceResponse` for traces, `ExportMetricsServiceResponse`
-for metrics, `ExportLogsServiceResponse` for logs).
+On success, the server response MUST be a Protobuf-encoded
```
tigrannajaryan, line 164: `Protobuf-encoded` is superfluous since this is about gRPC transport.

specification/protocol/otlp.md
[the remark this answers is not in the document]
tigrannajaryan: >Would you be in favor of removing this whole "limitation" section then? I think in the sections above is quite clear retries shouldn't be done so I'm fine in dropping this altogether. Less is more

I am in favour of removing the section.

I do not see a reasonable reading of the rest of the spec which results in someone deciding to make a retry after partial success because partial success response does not tell which part succeeded and which failed so there is no way to know which part to retry.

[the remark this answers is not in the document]
joaopgrassi: OK I will remove this section then.

specification/protocol/otlp.md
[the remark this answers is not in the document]
jmacd: I don't see a problem calling a situation with 0 dropped points a "partial success", so I suppose I don't see the problem with (3). I'm not inclined to block this work, however: we can leave this option open either way if we do nothing now.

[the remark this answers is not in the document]
joaopgrassi: I also don't see a problem, and I would like to be able to use it in that way, if everyone agrees. Looking forward to seeing what the others think about it.

specification/protocol/otlp.md
joaopgrassi: We use `Export<signal>ServiceResponse` + tell each individual name for signal on the "Full Success". With your comment here https://github.com/open-telemetry/opentelemetry-specification/pull/2696#discussion_r931274598 I changed to avoid the repetition of `ExportTraceServiceResponse...` in the "Partial Success". 

But for the `Export<signal>PartialSuccess` here I think it's clear to lay it out. It's only mentioned once in here (gRPC).

specification/protocol/otlp.md
joaopgrassi: Makes sense. Fixes in [55c1b61](https://github.com/open-telemetry/opentelemetry-specification/pull/2696/commits/55c1b61887f7f9649d1e179c4fe579c0569d2030)

specification/protocol/otlp.md
joaopgrassi: Good point. I followed your suggestion for both cases in [55c1b61](https://github.com/open-telemetry/opentelemetry-specification/pull/2696/commits/55c1b61887f7f9649d1e179c4fe579c0569d2030)

specification/protocol/otlp.md
joaopgrassi: Fixed in [55c1b61](https://github.com/open-telemetry/opentelemetry-specification/pull/2696/commits/55c1b61887f7f9649d1e179c4fe579c0569d2030) + internal links, toc

specification/protocol/otlp.md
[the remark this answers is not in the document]
joaopgrassi: Removed in [55c1b61](https://github.com/open-telemetry/opentelemetry-specification/pull/2696/commits/55c1b61887f7f9649d1e179c4fe579c0569d2030)

specification/protocol/otlp.md
joaopgrassi: NVM, I removed this entire section in [55c1b61](https://github.com/open-telemetry/opentelemetry-specification/pull/2696/commits/55c1b61887f7f9649d1e179c4fe579c0569d2030)

specification/protocol/otlp.md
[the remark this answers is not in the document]
joaopgrassi: @tigrannajaryan @reyang @jsuereth @jack-berg just pinging to see if any of you are opposed to @jmacd's suggestion here. I think this is the last point to settle before going forward :)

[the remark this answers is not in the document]
tigrannajaryan: @joaopgrassi @jmacd to clarify: you are suggesting that it is valid to have `ExportTracePartialSuccess` with `rejected_spans`=0 and non-empty `error_message`, right? I don't see a problem with this either. I think it should be acceptable to do so and it may be useful. I don't see how it can result in problems, so there is no reason to prohibit it.

jmacd (approved):

specification/protocol/otlp.md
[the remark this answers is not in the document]
joaopgrassi: Alright! I took the suggestion from @jmacd. Only difference is that I added it below the text explaining the `error_message`, felt weird that it was talking about it before properly introduced. Marking this thread as resolved!

joaopgrassi:
Hi all, just an update on the current state of this PR:

All discussions are solved and I think we should be good to go!

The last thing to be solved ([here](https://github.com/open-telemetry/opentelemetry-specification/pull/2696#discussion_r930249217)) was around using the partial success to convey warnings/suggestions to clients when **ALL** the data was accepted.

This is useful, for example, when the server did some massaging on the data (character normalization) and wants to send a "warning" to clients that this happened. Another usage would be to signal suggestions to clients, e.g. "hey you are sending the data via a less efficient/secure way, you should do xyz" like @reyang suggested [here](https://github.com/open-telemetry/opentelemetry-specification/pull/2696#discussion_r930388409).

Let me know your thoughts.

jack-berg (approved):

pirgeo (approved):

specification/protocol/otlp.md
```diff
@@ -145,16 +148,57 @@ was not delivered.
 
 #### OTLP/gRPC Response
 
-The server may respond with either a success or an error to the requests.
+The response MUST be the appropriate serialized message (see below for
```
tigrannajaryan, line 151: I think "serialized" is unnecessary here. There is no way to have a gRPC response that is not serialized.

specification/protocol/otlp.md
```diff
@@ -145,16 +148,57 @@ was not delivered.
...
+the number of spans/data points/log records it rejected.
+
+The server SHOULD populate the `error_message` field with a human-readable
+error message in English. The message should explain why the
+server rejected parts of the data, and might offer guidance on how users
+can address the issues.
+The protocol does not attempt to define the structure of the error message.
+
+Servers MAY also make use of the `partial_success` field to convey
+warnings/suggestions to clients when the request was fully accepted.
```
arminru, line 194:
```suggestion
warnings/suggestions to clients even when the request was fully accepted.
```
This might make it more readable.

arminru (approved):

tigrannajaryan (approved):

specification/protocol/otlp.md
joaopgrassi: Fixed in [21266d0](https://github.com/open-telemetry/opentelemetry-specification/pull/2696/commits/21266d0cd68aa00e53e9070df293e05528fc9558)

jmacd:
@joaopgrassi please rebase (or otherwise resolve "This branch is out-of-date with the base branch")?

joaopgrassi:
@jmacd rebased and also allowed edits from maintainers on the branch, in case it needs again and I'm away.

PR #2697, opened 2022-07-26
[enh][pulsar] add apache pulsar semantic_conventions

tjiuming:
## Changes

add apache pulsar semantic_conventions.

Related issues #

### DIFF
--- a/semantic_conventions/trace/messaging.yaml
+++ b/semantic_conventions/trace/messaging.yaml
@@ -262,7 +262,7 @@ groups:
       - id: producer_name
         type: string
         brief: >
-          Globally unique producer name which could have been assigned by the system or specified by the client.
+          Globally unique producer name.
         example: 'myProducer'
       - id: consumer_name
         type: string
--- a/semantic_conventions/trace/messaging.yaml
+++ b/semantic_conventions/trace/messaging.yaml
@@ -261,19 +261,18 @@ groups:
     attributes:
       - id: producer_name
         type: string
-        requirement_level: required
         brief: >
           Globally unique producer name.
         example: 'myProducer'
       - id: consumer_name
         type: string
-        requirement_level: required
         brief: >
           The name of consumer.
         example: 'myConsumer'
       - id: subscription
         type: string
-        requirement_level: required
+        requirement_level:
+          conditionally_required: If `consumer_name` has been set.
         brief: >
           Subscription name of Pulsar consumer.
         example: 'mySubscription'
--- a/semantic_conventions/trace/messaging.yaml
+++ b/semantic_conventions/trace/messaging.yaml
@@ -9,7 +9,7 @@ groups:
         type: string
         requirement_level: required
         brief: 'A string identifying the messaging system.'
-        examples: ['kafka', 'rabbitmq', 'rocketmq', 'activemq', 'AmazonSQS']
+        examples: ['kafka', 'rabbitmq', 'rocketmq', 'activemq', 'AmazonSQS', 'pulsar']
       - id: destination
         type: string
         requirement_level: required
@@ -46,7 +46,7 @@ groups:
       - id: url
         type: string
         brief: 'Connection string.'
-        examples: ['tibjmsnaming://localhost:7222', 'https://queue.amazonaws.com/80398EXAMPLE/MyQueue']
+        examples: ['tibjmsnaming://localhost:7222', 'https://queue.amazonaws.com/80398EXAMPLE/MyQueue', 'pulsar://localhost:6650']
       - id: message_id
         type: string
         brief: 'A value used by the messaging system as an identifier for the message, represented as a string.'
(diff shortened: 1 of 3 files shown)


semantic_conventions/trace/messaging.yaml
```diff
@@ -252,3 +252,76 @@ groups:
...
         brief: >
           Model of message consumption. This only applies to consumer spans.
+
+  - id: messaging.pulsar
+    prefix: messaging.pulsar
+    extends: messaging
+    brief: >
+      Attributes for Apache Pulsar
+    attributes:
+      - id: producer_name
```
lmolkova, line 262: can you please specify [attribute requirement level](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-requirement-level.md) when different than `recommended`? Or all of them are recommended?

(for all attributes added)

lmolkova (commented):
A couple of comments: 

1. Can you please also update messaging.md?
It could be done with something like `docker run -v /path/to/opentelemetry-specification/semantic_conventions:/source -v /path/to/opentelemetry-specification/specification:/dst otel/semconvgen:latest -f /source markdown -md /dst`
More info in https://github.com/open-telemetry/build-tools/tree/main/semantic-conventions

2. We have messaging SIG that works on the messaging semantic conventions [improvements](https://github.com/open-telemetry/oteps/pull/192), we [meet on Thu at 8am PST](https://github.com/open-telemetry/community).
Would you be able to join and to discuss this and the messaging spec changes we're workgin on?

semantic_conventions/trace/messaging.yaml
tjiuming: UPDATED.

tjiuming: Sorry, I cant join the meeting and it's too late for me.

tjiuming:
@codelipenghui Please also take a look

semantic_conventions/trace/messaging.yaml
```diff
@@ -252,3 +252,79 @@ groups:
...
+        type: string
+        requirement_level: required
```
lmolkova, line 264: I'd suggest limiting required attributes to essential ones only keeping telemetry costs for users in mind. E.g. producer_name can't be required on the consumer side. 

I see how this can help investigate issues between client and broker or detect non-closed clients, but still this information does not look absolutely essential for other cases, so I'd suggest the `recommended` level here and on the consumer_name.

semantic_conventions/trace/messaging.yaml
```diff
@@ -252,3 +252,79 @@ groups:
...
+        example: 'myProducer'
+      - id: consumer_name
+        type: string
+        requirement_level: required
+        brief: >
+          The name of consumer.
+        example: 'myConsumer'
+      - id: subscription
+        type: string
+        requirement_level: required
```
lmolkova, line 276: it can't be `required` on the producer side, so `conditionally_required` with the note that it applies to consumers only would make sense.

specification/trace/semantic_conventions/messaging.md
```diff
@@ -263,6 +264,46 @@ Specific attributes for Apache RocketMQ are defined below.
...
+
+| Value  | Description |
+|---|---|
+| `Durable` | Make the subscription to be backed by a durable cursor that will retain messages and persist the current position |
+| `NonDurable` | Lightweight subscription mode that does not have a durable cursor associated |
+<!-- endsemconv -->
+
 ## Examples
 
 ### Topic with multiple consumers
```
lmolkova, line 309: can you please add an example ?

semantic_conventions/trace/messaging.yaml
tjiuming: it makes sense, I'll change it

specification/trace/semantic_conventions/messaging.md
lmolkova: an example similar to one we have Kafka [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#apache-kafka-with-quarkus-or-spring-boot-example)

It would be nice to understand which spans pulsar instrumentation would create and have an example of each spans with all attributes, kind, links.

We only have example for Kafka in this spec, but examples in [database spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#examples) are very useful and it'd be nice to have something similar for each messaging system. It'd also give an idea of what to expect from pulsar in terms of general messaging attributes.

joaopgrassi:
@tjiuming just FYI, there's a group currently working on re-hauling the the existing messaging conventions. We are currently working on the existing attributes, trying to see if they make sense, how they are supposed to be populated, etc. 

Since the conventions are going to change I wonder if we should hold off on this now or still go ahead with it. For now, I added an item in our project board so we have it under our radar: https://github.com/orgs/open-telemetry/projects/20/views/1

specification/trace/semantic_conventions/messaging.md
tjiuming: @lmolkova example added, PTAL

PR #2699, opened 2022-07-26
Fix copy-paste error

foadnh:
There is an extra `}` character in the python method call. Probably it is there from copy-paste of the other method.

### DIFF
--- a/specification/metrics/api.md
+++ b/specification/metrics/api.md
@@ -709,7 +709,7 @@ API](../overview.md#api) authors might consider:
 # Python
 
 http_server_duration.Record(50, {"http.method": "POST", "http.scheme": "https"})
-http_server_duration.Record(100, http_method="GET", http_scheme="http"})
+http_server_duration.Record(100, http_method="GET", http_scheme="http")
 ```
 
 ```csharp


reyang (approved):

reyang:
Thanks @foadnh! Would you clear the CLA?

foadnh:
Thanks, @reyang; I signed CLA.

specification/metrics/api.md
```diff
@@ -709,7 +709,7 @@ API](../overview.md#api) authors might consider:
 # Python
 
 http_server_duration.Record(50, {"http.method": "POST", "http.scheme": "https"})
-http_server_duration.Record(100, http_method="GET", http_scheme="http"})
+http_server_duration.Record(100, http_method="GET", http_scheme="http")
```
Oberon00, line 712: Should authors really consider this? The keys here have underscores instead of dots, but both are actually used in semantic conventions, so you can't automatically map that. Maybe this line should be removed altogether?

MrAlias (approved):

specification/metrics/api.md
MrAlias: This example does not show how the variable names of the key-word arguments are turned into attributes. It seems assumed they are canonicalized when added telemetry.

For what it is worth, if I remember correctly, the keys to key-word arguments need to be valid python variables. Meaning they need to following python naming standards, and those standards do not allow `.` in them.

bogdandrutu: I think the discussion is different than what the PR tries to do. @Oberon00 please open an issue to discuss that, thank you :)

bogdandrutu (approved):

PR #2700, opened 2022-07-27
Clarify "automatic instrumentation" term

### LINKED ISSUE #2336: Clarify the definition of automatic instrumentation?
Should the glossary definition of Automatic Instrumentation change?

**Background**

The current definition is this:

> Refers to telemetry collection methods that do not require the end-user to write or access application code to use the OpenTelemetry APIs. Methods vary by programming language, and examples include bytecode injection or monkey patching.

I think this can be interpreted in two conflicting ways:

* It is enabled strictly through means that don't involve application code
* It is enabled through means that don't involve the use of OpenTelemetry APIs in application code

The first case would imply that automatic instrumentation _must_ come from something like an agent that runs alongside an app. The latter would imply that automatic instrumentation can also refer to enabling an instrumentation library in your application, since doing so does not utilize the OpenTelemetry APIs in your own code to generate data.

I would propose that we change the definition to something like this:

> Refers to telemetry collection done without the use of manual instrumentation via OpenTelemetry APIs. Methods vary by programming language, and examples include instrumentation libraries or instrumentation agents run side-by-side with applications

Here is why I think it could change:

* People adopting OpenTelemetry are already using a definition like this when they say Automatic Instrumentation (I see this at least once a week in my day job working for a vendor that supports OpenTelemetry ingest)
* It's not a stretch to say that instrumentation libraries generate data for me _automatically_
* The current definition seems geared towards maintainers and not users, and maintainers must absolutely make a distinction between instrumentation libraries and other means of gathering instrumentation automatically. But what's a useful distinction for maintainers is not necessarily a useful distinction for end-users
* Documentation for different languages is already using a definition like this, to varying degrees (this is easily addressed, but also a motivation for this discussion)

Alternatively, it may be helpful to have a definition like today in the spec, but have documentation intentionally use a broader definition that's inclusive of enabling an instrumentation library.

On the documentation site, for now we (myself, @austinlparker, @chalin) would prefer to [stick with the broader definition](https://github.com/open-telemetry/opentelemetry.io/issues/1042#issuecomment-1035319097) and not have to distinguish between "truly automatic" instrumentation vs. instrumentation libraries on a per-language basis (e.g., Python supports both, Go only supports one) -- but rather consider it all "automatic instrumentation" and then talk about the different options available.

pellared:
Fixes #2336

## Changes

Clarify "automatic instrumentation" term by removing some fragments that can bring confusion. 

1. Remove "to use the OpenTelemetry APIs" which some may think that library instrumentation is a type of automatic instrumentation
2. Remove "access application code". Personally, I think that e.g. "assembly weaving" (instrumentation done as part of "compilation") is also a kind of automatic instrumentation. For some languages, like .NET, source code is not required, but I imagine some interpreted languages require accessing the code anyway 🤷 I just think that we can remove this part.
3. Change "application code" to "application's source code". To make it more obvious that it is "source code" and e.g. not "bytecode".

### DIFF
--- a/specification/glossary.md
+++ b/specification/glossary.md
@@ -105,7 +105,8 @@ Coding against the OpenTelemetry API such as the [Tracing API](trace/api.md), [M
 
 ### Automatic Instrumentation
 
-Refers to telemetry collection methods that do not require the end-user to write or access application code to use the OpenTelemetry APIs. Methods vary by programming language, and examples include bytecode injection or monkey patching.
+Refers to telemetry collection methods that do not require the end-user to modify application's source code.
+Methods vary by programming language, and examples include bytecode injection or monkey patching.
 
 Synonym: *Auto-instrumentation*.
 


pellared:
@tedsuo @cartermp @chalin @cijothomas PTAL

specification/glossary.md
```diff
@@ -105,7 +105,8 @@ Coding against the OpenTelemetry API such as the [Tracing API](trace/api.md), [M
 
 ### Automatic Instrumentation
 
-Refers to telemetry collection methods that do not require the end-user to write or access application code to use the OpenTelemetry APIs. Methods vary by programming language, and examples include bytecode injection or monkey patching.
+Refers to telemetry collection methods that do not require the end-user to modify application's source code.
+Methods vary by programming language, and examples include bytecode injection or monkey patching.
```
cartermp, line 109: Something I think that might be confusing is that this defines the term in terms of _methods of collection_ but doesn't say anything about how those methods are packaged up (library, agent, something else). If that's not something we want to clarify here, then we'll need to clarify it elsewhere, perhaps not in the spec.

pellared: I think it should be defined in concrete automatic instrumentation implementation. I can imagine lots of possibilities including compiler-hook, profiler, runtime hooks, runtime manipulation, eBPF probe, custom runtimes...

pellared:
PTAL @open-telemetry/instr-wg @open-telemetry/java-instrumentation-approvers  @open-telemetry/dotnet-instrumentation-approvers

yurishkuro:
I don't think we should be removing reference to OTEL APIs, on the contrary, we should be emphasizing it, but clarifying. Specifically, an auto-instrumentation that attaches to the code at runtime and somehow exports OTEL-compatible data, would NOT be a compliant auto-instrumentation in my view, because:
  * it breaks the API/SDK separation envisioned in the OTEL
  * it likely does not allow adding manual instrumentation on top of automatic, because the manual instrumentation has to go through OTEL API.

pellared:
@yurishkuro How about adding:

```
The automatic instrumentation HAS to:

- use (directly or indirectly) the OpenTelemetry APIs,
- allow adding manual instrumentation.
```
?

However, I am not sure if this kind of detail should be contained in the **glossary**. Maybe there should be a dedicated document/spec describing automatic instrumentation (it should not be part of this PR).

reyang (approved):

specification/glossary.md
pellared: I would rather put such ideas in a dedicated document/spec describing automatic instrumentation (it rather should not be part of this PR).

Kielek (approved):

pellared:
I will be AFK for a month. If there will be any blocking feedback then I am OK if someone addresses it on my branch. I allowed edits and access to secrets by maintainers. Of course, this PR can also be closed and someone can open a new one.

_But maybe it is already good enough to be merged_  🤞

mateuszrzeszutek (approved):

yurishkuro (approved):

Issue #2682, opened 2022-07-21
Semantic conventions for Instrumentation Scope Attributes


joaopgrassi:
**What are you trying to achieve?**

Scope attributes were introduced in #2579 with the motivations behind it described in this [OTEP](https://github.com/open-telemetry/oteps/blob/main/text/0201-scope-attributes.md). A first motivation was to have an attribute to hold a [instrumentation scope "short name"](https://github.com/open-telemetry/opentelemetry-specification/pull/2422)

**What did you expect to see?**

Clearly defined semantic conventions around scope attributes so instrumentation/back-ends can rely/do things with it in a consistent manner.

**Additional context.**

I did some experimentation with the build-tools and how we would auto-generate the conventions tables, the same way we do for the rest today. You can see my attempt in this diff: https://github.com/open-telemetry/opentelemetry-specification/compare/main...dynatrace-oss-contrib:opentelemetry-specification:feature/scope_attributes_semconv?expand=1

My idea behind it was:

1. For scope attributes that are common across signals: A `YAML` file in `semantic_conventions`/common/scope.yaml` generating a `.md` file in `specification/common/scope.md`.
2. If there's need for signal-specific scope attributes: A `YAML` file in `semantic_conventions/metrics/scope-metrics.yaml` generating a `.md` file in `specification/metrics/semantic_conventions/scope-metrics.md`

Those would give us: 
- Common attributes:  `scope.*` e.g. `scope.short_name`
- Signal-specific attributes: `scope_metrics.*` e.g. `scope_metrics.my_meter_attribute`

cc @dashpole @tigrannajaryan

tigrannajaryan:
This is part of the Scope initiative. Triaging as accepted.

tigrannajaryan:
Important questions to answer:
- Are Scope attributes considered a completely different namespace from the other attributes? I.e. can I have a Scope attribute `http.host` that has a completely different meaning from the Span attribute `http.host`? 
- Or, are Scope attributes in the same namespace as all other attributes? If they are in the same namespace which of the existing attributes are OK to record on the Scope? All attributes? Some? 
- What does it mean to record an attribute that is defined for a Span at the Scope? Does it mean it applies to all Spans (and is a form of compression) or it applies to the Scope only and the Span can have its own value recorded?

dashpole:
[...]

Issue #2535, opened 2022-05-12
Clarify precedence of attributes when exporting to non-OTLP formats


tigrannajaryan:
The specification does not tell what happens when exporting to a format that is not capable of representing attributes separately for the Resource and for Span/Metric/LogRecord.

The exporter implementations have to flatten to one list of attributes and when the same attribute is present in the Resource and in the Span/Metric/LogRecord the implementation must make a decision about which attribute value takes precedence.

Existing implementations in Collector, for example [Zipkin exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/1c4dd67f7af084c467ce5378115fa151f7276547/pkg/translator/zipkin/zipkinv2/from_translator.go#L198) make this decision on their own (in Zipkin exporter Span attributes have precedence over Resource attributes when converting to Zipkin tags).

I think we need to define this behavior in the specification to make it consistent for all such exporters.

This came up [when discussing](https://github.com/open-telemetry/oteps/pull/201#discussion_r858881588) the precedence of Scope attributes in the OTEP.

pirgeo:
Since scope attributes are now part of the spec, we should probably include them in this effort.

tigrannajaryan:
> Since scope attributes are now part of the spec, we should probably include them in this effort.

I updated the description to also include the Scope attributes.

scheler:
1. If scope attributes describe *only* scope and span attributes describe the span, then the attributes cannot be overridden when flattening.
2. if we want to support overriding attributes, it should not be done only when exporting to non-OTLP destinations but should apply as a general concept for OTLP destinations as well. This will ensure consistency across OTLP and non-OTLP destinations.

mx-psi:
This still feels relevant

PR #2702, opened 2022-07-28
Add the short_name scope attribute

dashpole:
Fixes https://github.com/open-telemetry/opentelemetry-specification/issues/2682

## Changes

Add the `short_name` scope attribute.  Alternatively, it could have been added as `otel.short_name` to distinguish it from non-otel attributes, or `prometheus.namespace` to make it explicitly associated with prefixing prometheus metrics.

@jmacd @tigrannajaryan @jsuereth

### DIFF
--- a/semantic_conventions/scope/scope.yaml
+++ b/semantic_conventions/scope/scope.yaml
@@ -8,5 +8,9 @@ groups:
         type: string
         requirement_level: recommended
         brief: >
-         The single-word name for the instrumentation scope.
-        examples: ['mylibrary']
+         The short name for the instrumentation scope, which should generally be less than 15
+         characters, and generally consist of a single word. It is not required to be globally
+         unique, but should be unique enough to make it very unlikely to collide with other short
+         names.  An example use is as the [namespace](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-naming-and-namespaces)
+         (prefix) for OpenMetrics metric names.
+        examples: ['mylibrary', 'otelhttp']
--- a/semantic_conventions/scope/scope.yaml
+++ b/semantic_conventions/scope/scope.yaml
@@ -1,6 +1,6 @@
 groups:
   - id: scope
-    prefix:
+    prefix: ""
     type: scope
     brief: >
        Scope attributes
--- a/semantic_conventions/scope/scope.yaml
+++ b/semantic_conventions/scope/scope.yaml
@@ -1,6 +1,7 @@
 groups:
   - id: scope
-    prefix: scope
+    prefix:
+    type: scope
     brief: >
        Scope attributes
     attributes:
@@ -12,5 +13,6 @@ groups:
          characters, and generally consist of a single word. It is not required to be globally
          unique, but should be unique enough to make it very unlikely to collide with other short
          names.  An example use is as the [namespace](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-naming-and-namespaces)
-         (prefix) for OpenMetrics metric names.
-        examples: ['mylibrary', 'otelhttp']
+         (prefix) for OpenMetrics metric names. It is recommended for the short_name to be
+         related to the name of the scope.
+        examples: ['mylibrary', 'otelmacaron']
--- a/semantic_conventions/scope/scope.yaml
+++ b/semantic_conventions/scope/scope.yaml
@@ -1,0 +1,12 @@
+groups:
+  - id: scope
+    prefix: scope
+    brief: >
+       Scope attributes
+    attributes:
+      - id: short_name
+        type: string
+        requirement_level: recommended
+        brief: >
+         The single-word name for the instrumentation scope.
+        examples: ['mylibrary']
--- a/specification/scope/README.md
+++ b/specification/scope/README.md
@@ -7,5 +7,5 @@ This document defines standard attributes for the [instrumentation scope](../glo
 <!-- semconv scope -->
 | Attribute  | Type | Description  | Examples  | Requirement Level |
 |---|---|---|---|---|
-| `scope.short_name` | string | The single-word name for the instrumentation scope. | `mylibrary` | Recommended |
+| `scope.short_name` | string | The short name for the instrumentation scope, which should generally be less than 15 characters, and generally consist of a single word. It is not required to be globally unique, but should be unique enough to make it very unlikely to collide with other short names.  An example use is as the [namespace](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-naming-and-namespaces) (prefix) for OpenMetrics metric names. | `mylibrary`; `otelhttp` | Recommended |
 <!-- endsemconv -->
--- a/specification/scope/README.md
+++ b/specification/scope/README.md
@@ -7,5 +7,5 @@ This document defines standard attributes for the [instrumentation scope](../glo
 <!-- semconv scope -->
 | Attribute  | Type | Description  | Examples  | Requirement Level |
 |---|---|---|---|---|
-| `scope.short_name` | string | The short name for the instrumentation scope, which should generally be less than 15 characters, and generally consist of a single word. It is not required to be globally unique, but should be unique enough to make it very unlikely to collide with other short names.  An example use is as the [namespace](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-naming-and-namespaces) (prefix) for OpenMetrics metric names. | `mylibrary`; `otelhttp` | Recommended |
+| `short_name` | string | The short name for the instrumentation scope, which should generally be less than 15 characters, and generally consist of a single word. It is not required to be globally unique, but should be unique enough to make it very unlikely to collide with other short names.  An example use is as the [namespace](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-naming-and-namespaces) (prefix) for OpenMetrics metric names. It is recommended for the short_name to be related to the name of the scope. | `mylibrary`; `otelmacaron` | Recommended |
 <!-- endsemconv -->
--- a/specification/scope/README.md
+++ b/specification/scope/README.md
@@ -1,0 +1,11 @@
+# Scope Semantic Conventions
+
+**Status**: [Experimental](../document-status.md)
+
+This document defines standard attributes for the instrumentation scope.
+
+<!-- semconv scope -->
+| Attribute  | Type | Description  | Examples  | Requirement Level |
+|---|---|---|---|---|
+| `scope.short_name` | string | The single-word name for the instrumentation scope. | `mylibrary` | Recommended |
+<!-- endsemconv -->
--- a/specification/scope/README.md
+++ b/specification/scope/README.md
@@ -2,7 +2,7 @@
 
 **Status**: [Experimental](../document-status.md)
 
-This document defines standard attributes for the instrumentation scope.
+This document defines standard attributes for the [instrumentation scope](../glossary.md#instrumentation-scope).
 
 <!-- semconv scope -->
 | Attribute  | Type | Description  | Examples  | Requirement Level |


semantic_conventions/scope/scope.yaml
```diff
@@ -0,0 +1,12 @@
+groups:
+  - id: scope
+    prefix: scope
```
joaopgrassi, line 3: I dig a bit more and I think we need to introduce a new type here. The default, if empty is `span` as in here: https://github.com/open-telemetry/build-tools/blob/f8714186e10f6bb9fe2f92ada2f3d23760ebed46/semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py#L58

joaopgrassi (commented):
I believe it would make things more streamlined if first tackle the addition of "scope" as a semantic convention concept separately from the Prometheus use-cases/changes. So, a PR only for adding the scope new yaml/md.

dashpole:
> I believe it would make things more streamlined if first tackle the addition of "scope" as a semantic convention concept separately from the Prometheus use-cases/changes. So, a PR only for adding the scope new yaml/md.

Done

semantic_conventions/scope/scope.yaml
dashpole: Does that need to be done before this PR?

tigrannajaryan:
@dashpole you said [here](https://github.com/open-telemetry/opentelemetry-specification/issues/2682#issuecomment-1197355692) that
>I think the answer should be yes (namespaces are different), although we should try and avoid unnecessary collisions where possible to reduce confusion. I believe this is the same answer as the separation between resource vs trace/metric/log attributes.

If the Scope has a different namespace then why are we prefixing the attribute names by `scope.`? That seems unnecessary to me. It would be necessary if we decided that the Scope attribute names are in the same namespace as the Resource attributes and/or Trace or Metric attributes.

I am not sure what is the right approach but it seems like it needs to be one or another and then we need to consistent with that decision, which for now we don't seem to be.

specification/scope/README.md
```diff
@@ -0,0 +1,11 @@
+# Scope Semantic Conventions
+
+**Status**: [Experimental](../document-status.md)
+
+This document defines standard attributes for the instrumentation scope.
```
joaopgrassi, line 5:
```suggestion
This document defines common attributes for the [instrumentation scope](../glossary.md#instrumentation-scope).
```

Or maybe something like 

"This document defines attributes to be used on the [instrumentation scope](../glossary.md#instrumentation-scope) that are common between all signals."

semantic_conventions/scope/scope.yaml
```diff
@@ -0,0 +1,12 @@
...
+    brief: >
+       Scope attributes
+    attributes:
+      - id: short_name
+        type: string
+        requirement_level: recommended
+        brief: >
+         The single-word name for the instrumentation scope.
```
joaopgrassi, line 11: Do we want to specify which characters are allowed in this? If this is going to be appended to metric keys for example I wonder if we should constrain the allowed characters. WDYT?

specification/scope/README.md
```diff
@@ -0,0 +1,11 @@
+# Scope Semantic Conventions
```
joaopgrassi, line 1:
```suggestion
# Instrumentation Scope Semantic Conventions
```

specification/scope/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+**Status**: [Experimental](../document-status.md)
+
+This document defines standard attributes for the instrumentation scope.
+
+<!-- semconv scope -->
+| Attribute  | Type | Description  | Examples  | Requirement Level |
+|---|---|---|---|---|
+| `scope.short_name` | string | The single-word name for the instrumentation scope. | `mylibrary` | Recommended |
```
tigrannajaryan, line 10: Please add an explanation of the semantics. E.g. are there any uniqueness requirements, should it be derived from the regular `Name` or how it is related to `Name` if it is, etc. Are there max length recommendations, what are valid characters to use, etc. Perhaps link to relevant issues/PRs that explain why this is needed.

dashpole:
> If the Scope has a different namespace then why are we prefixing the attribute names by scope.? That seems unnecessary to me. It would be necessary if we decided that the Scope attribute names are in the same namespace as the Resource attributes and/or Trace or Metric attributes.

My thinking was that some scope attributes may only apply to a particular definition of scope, such as `module.path`, or library.something... But I'm also having trouble coming up with examples, so maybe its a stretch.  I'm happy to change it to `short_name` if you think it makes more sense

tigrannajaryan:
> I'm happy to change it to `short_name` if you think it makes more sense

It does look a bit weird. Virtually every other attribute we have in semantic conventions for Resources or Spans or Metrics has some sort of a prefix. We have never defined any attributes in the root namespace although I don't think [the rules](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-naming.md) explicitly prohibit it.

However using a prefix `scope.` in a root namespace what is already asserted to be entirely allocated to the scopes looks weird too. Is there a way to come up with a different meaningful prefix that in the future may contain more than just the `short_name`?

On the other hand the benefit of using `scope.` prefix is that it mitigates the problem described in https://github.com/open-telemetry/opentelemetry-specification/issues/2535. If `scope.` is used as a prefix then in practice it is never going to clash with Resource or Span attribute names, so how we define the precedence will almost not matter.

Sorry, I am probably not helping much. I probably need to make up my mind first :-)

specification/scope/README.md
dashpole: I've added the link.  I prefer the current wording, as these attributes aren't really common across signals.  An instrumentation scope is orthogonal to signals (like resource).

specification/scope/README.md
dashpole: The full semantics I want are described in https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-naming-and-namespaces.  I'll try and capture that here.

semantic_conventions/scope/scope.yaml
dashpole: Do you mean for scope attributes in general, or for the short_name specifically?

semantic_conventions/scope/scope.yaml
joaopgrassi: I approved the other PR. I guess you can go ahead and already use the new type here.

semantic_conventions/scope/scope.yaml
joaopgrassi: For `short_name` specifically. Since it's going to be used in https://github.com/open-telemetry/opentelemetry-specification/pull/2703 (and in potentially other non-OTLP exporters that also do flattening) I wonder if we shouldn't have more constraints around it. Like in https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

joaopgrassi: I saw you added a link to [https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-naming-and-namespaces](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md?rgh-link-date=2022-07-29T20%3A33%3A11Z#metric-naming-and-namespaces). 

I think that helps in general with naming and namespace "topics", but the "what characters are valid" is only below in the [Text format](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#text-format) section in the ABNF def. I wonder if we shouldn't just use something like the Prometheus page uses 

```
It may contain ASCII letters and digits, as well as underscores and colons. It must match the regex [a-zA-Z_:][a-zA-Z0-9_:]*
```
Curious to see what others say about it. I don't know if we have such constraints for other fields in the spec today.

dashpole: We have already defined conversion rules in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md#metric-attributes for translating from what OTel allows to what prometheus allows.  Unless we wanted to make this more specific (e.g. `prometheus.namespace`), we should probably keep this as the normal allowed character set.

joaopgrassi:
>It does look a bit weird. Virtually every other attribute we have in semantic conventions for Resources or Spans or Metrics has some sort of a prefix. We have never defined any attributes in the root namespace although I don't think [the rules](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-naming.md?rgh-link-date=2022-07-29T17%3A54%3A18Z) explicitly prohibit it.

I find strange that we have `otel.scope.name` and then `scope.short_name`. Are we completely opposed to have it under `otel.scope.short_name`? When I read the linked [Attribute naming](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-naming.md#otel-namespace) page above:

>Attribute names that start with otel. are reserved to be defined by OpenTelemetry specification. These are typically used to express OpenTelemetry concepts in formats that don't have a corresponding concept.

It seems `short_name` falls into that well, no? It's like a sibling to `name` and `version` in regards of how important it is. Even so, if `short_name` is going to be used for https://github.com/open-telemetry/opentelemetry-specification/pull/2703. It carries more meaning than an arbitrary attribute on a scope.

Then, other scope attributes don't need any prefix, and can exist by themselves, we just happen to "list" them under the scope semantic conventions bcs it's where they should be applied. And this ties nicely with the next thing:

> On the other hand the benefit of using scope. prefix is that it mitigates the problem described in https://github.com/open-telemetry/opentelemetry-specification/issues/2535

I would think it's better to not have a mixed approach when it comes to solving the [question](https://github.com/open-telemetry/opentelemetry-specification/issues/2535) of how we "flatten" attributes. Like, either all signals have their attributes prefixed or none have. Having a prefix for `scope` and not for `resource` or `span` sounds weird to me, inconsistent.

semantic_conventions/scope/scope.yaml
```diff
@@ -0,0 +1,16 @@
...
+      - id: short_name
+        type: string
+        requirement_level: recommended
+        brief: >
+         The short name for the instrumentation scope, which should generally be less than 15
+         characters, and generally consist of a single word. It is not required to be globally
+         unique, but should be unique enough to make it very unlikely to collide with other short
+         names.  An example use is as the [namespace](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-naming-and-namespaces)
+         (prefix) for OpenMetrics metric names.
+        examples: ['mylibrary', 'otelhttp']
```
jack-berg, line 16: Is `otelhttp` actually a good example? Two http client libraries present on the same application might both use this as the short name and collide. Not sure if that's a problem since both should follow the same semantic conventions. 🤔

dashpole: Do you mean two instances of a client library, or two completely different client libraries?  If the former, that type of collision is fine--metrics will just be aggregated. The latter would indicate that otel has published more than one http client library.  Is that the case that you are aware of?

semantic_conventions/scope/scope.yaml
joaopgrassi: Ah thanks for pointing it out, I think it's all good then :)

semantic_conventions/scope/scope.yaml
jack-berg: Yes the latter, with more than one client library. I think its quite possible: suppose you have dependencies on two client libraries that each are implemented over http and use one of the popular http client libraries in java. One uses okhttp, the other uses apache http. These would each publish `http.server.duration` metrics, but under different instrumentation scope names.

dashpole: In that case I would expect `okhttp` and `apachehttp` as the short names to distinguish the two.  otelhttp is a reference to the golang opentelemetry library for the core golang http library

dashpole:
One topic that came up during today's spec call is the idea of making this attribute's behavior consistent across exporters.  For example, prometheus adds it as a prefix to metrics, but why shouldn't other exporters do the same?  Perhaps other non-OTLP exporters should also add this as a prefix.  Maybe something like `name_prefix` would be clearer that the intention is for this to be a name prefix

semantic_conventions/scope/scope.yaml
jack-berg: Why not something like `gohttp` instead of `otelhttp`?

dashpole: That could also work.  The go SIG chose `otelhttp` as the package name, so that would likely be the easiest for users to understand, but it doesn't particularly matter for this discussion.  The only requirement is for it to be unique within an application, so we don't generally need to worry about cross-language naming conflicts, and the go SIG can choose the short name they prefer for those packages.  I'll change it to [otelmacaron](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/gopkg.in/macaron.v1/otelmacaron) to make this less contentious.

semantic_conventions/scope/scope.yaml
dashpole: used the new type

dashpole: Presubmits are failing.  How do I get a new version of the build-tools repo to run on this PR?

semantic_conventions/scope/scope.yaml
jack-berg: > I'll change it to [otelmacaron](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/gopkg.in/macaron.v1/otelmacaron) to make this less contentious.

Perfect. Sorry to be pedantic but `otelhttp` is just generic enough to potentially cause instrumentation authors to think its some sort of standard for all http instrumentation.

specification/scope/README.md
```diff
@@ -0,0 +1,11 @@
+# Scope Semantic Conventions
+
+**Status**: [Experimental](../document-status.md)
+
+This document defines standard attributes for the [instrumentation scope](../glossary.md#instrumentation-scope).
+
+<!-- semconv scope -->
+| Attribute  | Type | Description  | Examples  | Requirement Level |
+|---|---|---|---|---|
+| `scope.short_name` | string | The short name for the instrumentation scope, which should generally be less than 15 characters, and generally consist of a single word. It is not required to be globally unique, but should be unique enough to make it very unlikely to collide with other short names.  An example use is as the [namespace](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-naming-and-namespaces) (prefix) for OpenMetrics metric names. | `mylibrary`; `otelhttp` | Recommended |
```
jack-berg, line 10: Not a blocking comment, but it seems likely that `scope.short_name` will make its way into scope attributes for traces and logs as well. I can't think of any reason why this would be a problem - just want to confirm that everyone understands `scope.short_name` to be a signal agnostic attribute.

jack-berg (approved):

dashpole:
I've opted to update the PR to use the prefix-less `short_name` instead of `scope.short_name` based on discussions above.

tigrannajaryan:
Regarding using `scope.` or `otel.` as a prefix: even if we use it for some attributes (like `scope.short_name`) we must still allow some other attributes that don't have the prefix. An example use case is the Events and Logs API, where we propose to have a Scope attribute named `event.domain`. I believe `event.domain` is a good name for an attribute and there is no need to prefix it with `scope.` or `otel.`.

So, at the very least, IMO, we cannot have a hard rule that says `scope.` or `otel.` must be always a prefix for Scope attributes.

joaopgrassi:
> I've opted to update the PR to use the prefix-less short_name instead of scope.short_name based on discussions above.

@dashpole does the tooling works in this case, to auto generate the table? (without a prefix)

dashpole:
> @dashpole does the tooling works in this case, to auto generate the table? (without a prefix)

I had to add one commit to the scope build-tools PR to allow an empty prefix, but otherwise it works.

specification/scope/README.md
jmacd: Is this attribute "special" in some way, compared with others? In the companion #2703 an example is given with two scope attributes, `scope.short_name` and `library_mascot`. 

If I receive an OTLP export (for any signal) with two scopes that contain otherwise identically-named logs/spans/metrics, then:

- the scope.short_name is implicit qualifier for the named logs/spans/metrics, making them distinct and unrelated; metrics belonging to the two scopes cannot internally break the single-writer rule (despite sharing same attributes).
- the library_mascot is just like a resource attribute, has no special properties; metrics belonging to the two scopes would break the single-writer rule, if they were to share the same attributes.

Not sure if my interpretation is correct or not, would like this section to clarify.

dashpole: > just like a resource attribute, has no special properties; metrics belonging to the two scopes would break the single-writer rule, if they were to share the same attributes.

This doesn't seem correct for resource attributes.  Reporting CPU usage (assume the metric has no attributes) is possible for two different resources (e.g. two different containers) without breaking the single-writer rule, right?

My take is that:

1. Fundamentally, the scope (full) name+ version is the implicit qualifier which implies that metrics within it distinct and unrelated.  I believe this is currently the status quo.
2. All scope attributes are descriptive (not identifying) of the scope name + version
3. short_name is a shortened identifier which is likely, but not guaranteed, to uniquely identify the scope name + version within a program.

jmacd: OK, thank you. I prefer your interpretation.
In data-model.md, there's a paragraph that can use a slight extension:

```
Metric streams are grouped into individual `Metric` objects,
identified by:

- The originating `Resource` attributes
- The instrumentation `Scope` (e.g., instrumentation library name, version)
- The metric stream's `name`
```
Maybe change "(e.g., instrumentation library name, version)" to "(i.e., instrumentation scope name, version, and attributes)".

semantic_conventions/scope/scope.yaml
dashpole: We will follow https://github.com/open-telemetry/build-tools/pull/114#issuecomment-1209002674 to update this.

specification/scope/README.md
dashpole: I was actually thinking that scope attributes would be non-identifying: "All scope attributes are descriptive (not identifying) of the scope name + version", as that would reduce the burden on non-otel exporters to REQUIRE them to be attached to all telemetry.  Are there use-cases where we want attributes to be identifying?

bogdandrutu: I don't think we want to have them, since we anyway don't use them to construct different meters.

Aneurysm9:
As discussed in the spec SIG today I wonder whether it would be beneficial to define a requirement that scope attributes be prefixed with `otel.scope` when processed by non-OTel components.  This would align with `otel.scope.name` and `otel.scope.version` and provide a clear indication that an attribute was a scope attribute and not an attribute on a signal item.

carlosalberto:
Is it to bad to suggest that _only_ root-level attributes _get_ a suffix (`short_name` becomes `scope.short_name` or `otel.scope.short_name`), as they are de facto considered meta attributes?

carlosalberto:
Approving, but as mentioned in the Spec call too, let's allow some days for some days to iron out final details.

specification/scope/README.md
jmacd: > I don't think we want to have them, since we anyway don't use them to construct different meters.

I expect varying scope attributes will produce different meters, seems to be the intention behind #2579, specifically https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#get-a-meter 

specifically "[since 1.13.0] attributes (optional): Specifies the instrumentation scope attributes to associate with emitted telemetry." and "Implementations MUST return different Meter instances when called repeatedly with different values of parameters."

jmacd:
(I could not attend the Spec SIG yesterday and the recording is not yet available.)

From the discussion above, looks like we're still debating what to call this attribute and whether it should have "otel.scope" as a prefix.

@dashpole asked two weeks ago:

> Maybe something like `name_prefix` would be clearer that the intention is for this to be a name prefix

I suggest that `namespace` best describes the thing we're talking about. The Prometheus exporter would be specified to construct metric names by stitching the `namespace`, `_`, and the metric name together. This namespace concept that we are describing exactly matches OpenMetrics, which is nice.

I think `otel.scope.` should remain the suggested prefix for instrumentation scope name and version concepts, because they are OpenTelemetry concepts. I would not automatically apply a `otel.scope.` prefix to any of the scope attributes. (Tangent: My guess is that a standard prefix is intended to make it easier to flatten the attributes into a list. Question: [why flatten in the first place, when OpenMetrics specifically describes the use of join?](https://github.com/open-telemetry/opentelemetry-specification/pull/2736)?)

dashpole:
> Why flatten in the first place, when OpenMetrics specifically describes the use of join?

I don't think we should flatten for Prometheus/OpenMetrics.  I've proposed a join-able info metric in https://github.com/open-telemetry/opentelemetry-specification/pull/2703.  I think the [guidance](https://github.com/open-telemetry/opentelemetry-specification/blob/8ed127a403c59e1baa5203c6cdf472ea21df0332/specification/common/mapping-to-non-otlp.md#instrumentation-scope-attributes) is for other non-OTLP exporters.

A scope is already somewhat of a namespace, and i'm hoping to introduce something thats closely tied to the scope name to avoid creating a new concept.  I also worry an un-prefixed namespace would collide with metric/span/logRecord attributes often.

dashpole:
Based on https://github.com/open-telemetry/opentelemetry-specification/pull/2703#issuecomment-1231892392, this isn't required anymore.

Issue #2493, opened 2022-04-15
Record instrumentation_scope attribute in Prometheus export


anuraaga:
Currently, instrumentation_scope is not used in the documentation for converting OTLP metrics to Prometheus. However, because it is part of identity for the OTLP model, it should be added as an attribute. Otherwise, the OTel SDK can record two metric streams with the same name and attributes, but different instrumentation scope, and the prometheus exporter will output them as identical streams causing a failure to scrape.

anuraaga:
@open-telemetry/wg-prometheus

arminru:
Related to #1906.

anuraaga:
Ah this is mostly a dupe of #1906 so feel free to close this if it makes sense. We need to prioritize it though, while we don't intend to publish Prometheus exporter as stable next month, with metrics going stable we expect a lot of bug reports for this since it's easy to run into.

jack-berg:
#2703 should close this.

PR #2703, opened 2022-07-29
Add Instrumentation Scope and Version as labels in Prometheus

dashpole:
Fixes: #2493

Related: https://github.com/open-telemetry/opentelemetry-specification/issues/1906

This is a second attempt at https://github.com/open-telemetry/opentelemetry-specification/pull/2422.

~~No longer depends on https://github.com/open-telemetry/opentelemetry-specification/pull/2702~~

### Background: Naming Collisions

OpenTelemetry encourages the use of semantic conventions to make metric naming similar across instrumentation.  For example, if I have two http client libraries in my application, they would each produce a metric named `http.client.duration`, but with different meters (e.g. [otelmux](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/0dd27453a1ce8e433cb632e175a27f28ee83998d/instrumentation/github.com/gorilla/mux/otelmux) vs [otelhttp](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/0dd27453a1ce8e433cb632e175a27f28ee83998d/instrumentation/net/http/otelhttp)).  A prometheus exporter which receives both of these metrics would not be able to serve both of those histograms.  This would occur anytime a user uses two libraries which produces the same category (e.g. http, database, rpc, etc) of metrics, or if the two libraries just happen to use the same name for a metric.  Depending on the language, it may fail to create the Prometheus exporter, or may fail to send some, or all metrics if the same labels keys and values are present in both.

### Desired User Experience

As a user, I can use a Prometheus exporter with OpenTelemetry without experiencing strange errors/behavior due to naming collisions, and without having to apply transformations to metric names to work around these, except in rare cases.

As a user, I can easily add scope attributes to my metrics in Prometheus by joining with an info-style metric.  This is a common pattern in Prometheus: https://grafana.com/blog/2021/08/04/how-to-use-promql-joins-for-more-effective-queries-of-prometheus-metrics-at-scale/.

### Design

Add `opentelemetry_scope_name` and `opentelemetry_scope_version` as labels to all metrics.  This ensures that if two libraries produce the same metric points, they don't collide because the scope name/version labels will differ.

Those labels also serve as "join keys" to be able to add scope attributes to Prometheus metrics.  This is accomplished by introducing an `opentelemetry_scope_info` metric containing the same `opentelemetry_scope_name` and `opentelemetry_scope_version` labels, but also including scope attributes.  This also enables the collector's Prometheus receiver to reconstruct the original Instrumentation Scope when receiving the metrics.

@open-telemetry/wg-prometheus
@jmacd @tigrannajaryan @jsuereth

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ release.
 
 ### Compatibility
 
+- Specify how Prometheus exporters and receivers handle instrumentation scope.
+  ([#2703](https://github.com/open-telemetry/opentelemetry-specification/pull/2703)).
+
 ### OpenTelemetry Protocol
 
 ### SDK Configuration
--- a/specification/metrics/data-model.md
+++ b/specification/metrics/data-model.md
@@ -1424,8 +1424,8 @@ scope_metrics:
         - value: 1
 ```
 
-Metrics which are not found to be associated with an instrumentation scope are
-all placed within an empty instrumentation scope, and do not have any labels
+Metrics which are not found to be associated with an instrumentation scope MUST
+all be placed within an empty instrumentation scope, and MUST not have any labels
 removed.
 
 #### Resource Attributes
--- a/specification/metrics/data-model.md
+++ b/specification/metrics/data-model.md
@@ -1498,7 +1498,11 @@ following the rules described in the [`Metric Attributes`](#metric-attributes)
 section below.
 
 Prometheus exporters MUST add the scope name as the `otel_scope_name` label and
-the scope version as the `otel_scope_version` label on all metric points.
+the scope version as the `otel_scope_version` label on all metric points by
+default.
+
+Prometheus exporters SHOULD provide a configuration option to disable the
+`otel_scope_info` metric and `otel_scope_` labels.
 
 #### Gauges
 
--- a/specification/metrics/data-model.md
+++ b/specification/metrics/data-model.md
@@ -66,9 +66,11 @@ linkTitle: Data Model
     + [Dropped Types](#dropped-types)
     + [Start Time](#start-time)
     + [Exemplars](#exemplars-1)
+    + [Instrumentation Scope](#instrumentation-scope)
     + [Resource Attributes](#resource-attributes)
   * [OTLP Metric points to Prometheus](#otlp-metric-points-to-prometheus)
     + [Metric Metadata](#metric-metadata-1)
+    + [Instrumentation Scope](#instrumentation-scope-1)
     + [Gauges](#gauges-1)
     + [Sums](#sums-1)
     + [Histograms](#histograms-1)
@@ -1385,6 +1387,48 @@ retrieved from the `trace_id` and `span_id` label keys, respectively.  All
 labels not used for the trace and span ids MUST be added to the OpenTelemetry
 exemplar as attributes.
 
+#### Instrumentation Scope
+
+Each `opentelemetry_scope_info` metric point present in a batch of metrics
+SHOULD be dropped from the incoming scrape, and converted to an instrumentation
+scope. The `opentelemetry_scope_name` and `opentelemetry_scope_version` labels,
+if present, MUST be converted to the Name and Version of the Instrumentation
+Scope. Additional labels MUST be added as scope attributes, with keys and
+values unaltered. Other metrics in the batch which have
+`opentelemetry_scope_name` and `opentelemetry_scope_version`
+labels that match an instrumentation scope MUST be placed within the matching
+instrumentation scope, and MUST remove those labels. For example, the
+OpenMetrics metrics:
+
+```
+# TYPE opentelemetry_scope_info info
+opentelemetry_scope_info{opentelemetry_scope_name="go.opentelemetry.io.contrib.instrumentation.net.http.otelhttp",opentelemetry_scope_version="v0.24.0",library_mascot="bear"} 1
+# TYPE otelhttp_http_server_duration counter
+otelhttp_http_server_duration{opentelemetry_scope_name="go.opentelemetry.io.contrib.instrumentation.net.http.otelhttp",opentelemetry_scope_version="v0.24.0"...} 1
+```
+
+becomes:
+
+```yaml
+# within a resource_metrics
+scope_metrics:
+  scope:
+    name: go.opentelemetry.io.contrib.instrumentation.net.http.otelhttp
+    version: v0.24.0
+    attributes:
+      library_mascot: bear
+  metrics:
+  - name: http_server_duration
+    data:
+      sum:
+        data_points:
+        - value: 1
+```
+
+Metrics which are not found to be associated with an instrumentation scope are
+all placed within an empty instrumentation scope, and do not have any labels
+removed.
+
 #### Resource Attributes
 
 When scraping a Prometheus endpoint, resource attributes MUST be added to the
@@ -1444,6 +1488,19 @@ The description of an OTLP metrics point MUST be added as
 The data point type of an OTLP metric MUST be added as
 [OpenMetrics TYPE metadata](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metricfamily).
 It also dictates type-specific conversion rules listed below.
+
+#### Instrumentation Scope
+
+Prometheus exporters SHOULD generate an [Info](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#info)-typed
+metric named `opentelemetry_scope_info`. If present, Instrumentation
+Scope `name` and `version` MUST be added as `opentelemetry_scope_name` and
+`opentelemetry_scope_version` labels. Scope attributes MUST also be added as
+labels following the rules described in the
+[`Metric Attributes`](#metric-attributes) section below.
+
+Prometheus exporters MUST add the scope name as the `opentelemetry_scope_name`
+label and SHOULD add the scope version as the `opentelemetry_scope_version` label
+on all metrics points.
 
 #### Gauges
 
--- a/specification/metrics/data-model.md
+++ b/specification/metrics/data-model.md
@@ -1391,20 +1391,19 @@ exemplar as attributes.
 
 Each `opentelemetry_scope_info` metric point present in a batch of metrics
 SHOULD be dropped from the incoming scrape, and converted to an instrumentation
-scope. The `opentelemetry_scope_name` and `opentelemetry_scope_version` labels,
-if present, MUST be converted to the Name and Version of the Instrumentation
-Scope. Additional labels MUST be added as scope attributes, with keys and
-values unaltered. Other metrics in the batch which have
-`opentelemetry_scope_name` and `opentelemetry_scope_version`
+scope. The `otel_scope_name` and `otel_scope_version` labels, if present, MUST
+be converted to the Name and Version of the Instrumentation Scope. Additional
+labels MUST be added as scope attributes, with keys and values unaltered. Other
+metrics in the batch which have `otel_scope_name` and `otel_scope_version`
 labels that match an instrumentation scope MUST be placed within the matching
 instrumentation scope, and MUST remove those labels. For example, the
 OpenMetrics metrics:
 
 ```
 # TYPE opentelemetry_scope_info info
-opentelemetry_scope_info{opentelemetry_scope_name="go.opentelemetry.io.contrib.instrumentation.net.http.otelhttp",opentelemetry_scope_version="v0.24.0",library_mascot="bear"} 1
-# TYPE otelhttp_http_server_duration counter
-otelhttp_http_server_duration{opentelemetry_scope_name="go.opentelemetry.io.contrib.instrumentation.net.http.otelhttp",opentelemetry_scope_version="v0.24.0"...} 1
+opentelemetry_scope_info{otel_scope_name="go.opentelemetry.io.contrib.instrumentation.net.http.otelhttp",otel_scope_version="v0.24.0",library_mascot="bear"} 1
+# TYPE http_server_duration counter
+http_server_duration{otel_scope_name="go.opentelemetry.io.contrib.instrumentation.net.http.otelhttp",otel_scope_version="v0.24.0"...} 1
 ```
 
 becomes:
@@ -1492,15 +1491,15 @@ It also dictates type-specific conversion rules listed below.
 #### Instrumentation Scope
 
 Prometheus exporters SHOULD generate an [Info](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#info)-typed
-metric named `opentelemetry_scope_info`. If present, Instrumentation
-Scope `name` and `version` MUST be added as `opentelemetry_scope_name` and
-`opentelemetry_scope_version` labels. Scope attributes MUST also be added as
-labels following the rules described in the
-[`Metric Attributes`](#metric-attributes) section below.
+metric named `opentelemetry_scope_info`. If present, Instrumentation Scope
+`name` and `version` MUST be added as `otel_scope_name` and
+`otel_scope_version` labels. Scope attributes MUST also be added as labels
+following the rules described in the [`Metric Attributes`](#metric-attributes)
+section below.
 
-Prometheus exporters MUST add the scope name as the `opentelemetry_scope_name`
-label and SHOULD add the scope version as the `opentelemetry_scope_version` label
-on all metrics points.
+Prometheus exporters MUST add the scope name as the `otel_scope_name` label and
+SHOULD add the scope version as the `otel_scope_version` label on all metrics
+points.
 
 #### Gauges
 
--- a/specification/metrics/data-model.md
+++ b/specification/metrics/data-model.md
@@ -1389,7 +1389,7 @@ exemplar as attributes.
 
 #### Instrumentation Scope
 
-Each `opentelemetry_scope_info` metric point present in a batch of metrics
+Each `otel_scope_info` metric point present in a batch of metrics
 SHOULD be dropped from the incoming scrape, and converted to an instrumentation
 scope. The `otel_scope_name` and `otel_scope_version` labels, if present, MUST
 be converted to the Name and Version of the Instrumentation Scope. Additional
@@ -1400,8 +1400,8 @@ instrumentation scope, and MUST remove those labels. For example, the
 OpenMetrics metrics:
 
 ```
-# TYPE opentelemetry_scope_info info
-opentelemetry_scope_info{otel_scope_name="go.opentelemetry.io.contrib.instrumentation.net.http.otelhttp",otel_scope_version="v0.24.0",library_mascot="bear"} 1
+# TYPE otel_scope_info info
+otel_scope_info{otel_scope_name="go.opentelemetry.io.contrib.instrumentation.net.http.otelhttp",otel_scope_version="v0.24.0",library_mascot="bear"} 1
 # TYPE http_server_duration counter
 http_server_duration{otel_scope_name="go.opentelemetry.io.contrib.instrumentation.net.http.otelhttp",otel_scope_version="v0.24.0"...} 1
 ```
@@ -1491,15 +1491,14 @@ It also dictates type-specific conversion rules listed below.
 #### Instrumentation Scope
 
 Prometheus exporters SHOULD generate an [Info](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#info)-typed
-metric named `opentelemetry_scope_info`. If present, Instrumentation Scope
+metric named `otel_scope_info`. If present, Instrumentation Scope
 `name` and `version` MUST be added as `otel_scope_name` and
 `otel_scope_version` labels. Scope attributes MUST also be added as labels
 following the rules described in the [`Metric Attributes`](#metric-attributes)
 section below.
 
 Prometheus exporters MUST add the scope name as the `otel_scope_name` label and
-SHOULD add the scope version as the `otel_scope_version` label on all metrics
-points.
+the scope version as the `otel_scope_version` label on all metric points.
 
 #### Gauges
 


specification/metrics/data-model.md
```diff
@@ -1326,6 +1371,25 @@ The data point type of an OTLP metric MUST be added as
...
 It also dictates type-specific conversion rules listed below.
 
+#### Instrumentation Scope
+
+OpenMetrics prevents naming collisions for metrics from different libraries by
+adding a [metric namespace](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-naming-and-namespaces),
+which is generally a single-word prefix, to metrics. The
+[`scope.short_name` scope attribute](../scope/README.md#scope-semantic-conventions),
+serves the same purpose. `scope.short_name`, if present, MUST be attached as a
+prefix to all metrics from an Instrumentation Scope.
```
aabmass, line 1381: nit, this doesn't mention the prefix should be followed by an underscore `<scope.short_name>_`

aabmass (approved):

specification/metrics/data-model.md
```diff
@@ -1273,6 +1275,49 @@ retrieved from the `trace_id` and `span_id` label keys, respectively.  All
...
+added to the Instrumentation Scope as the value for the `scope.short_name`
+attribute. Additional labels MUST be added as scope attributes, with keys and
+values unaltered. Other metrics in the batch which have a prefix equal to the
+prefix of the `<prefix>_otel_scope_info` metric have the matching prefix
+removed, and are placed within that Instrumentation Scope.  For example, the
+Prometheus metrics:
+
+```
+# TYPE otelhttp_otel_scope_info gauge
+otelhttp_otel_scope_info{name=go.opentelemetry.io.contrib.instrumentation.net.http.otelhttp,version=v0.24.0,library_mascot=bear} 1
```
dashpole, line 1293: needs quotes around label values

dashpole, line 1293: suggestion from the meeting: write out opentelemetry fully.

dashpole:
A comment from the prometheus wg: We should consider splitting `build_info` out of `target_info`.  There is a question about consistency across languages, and what the structure/overlap of those conventions are with otel's conventions.

specification/metrics/data-model.md
dashpole: done

dashpole: done

specification/metrics/data-model.md
dashpole: done

specification/metrics/data-model.md
```diff
@@ -1337,6 +1382,25 @@ The data point type of an OTLP metric MUST be added as
...
+
+OpenMetrics prevents naming collisions for metrics from different libraries by
+adding a [metric namespace](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-naming-and-namespaces),
+which is generally a single-word prefix, to metrics. The
+[`scope.short_name` scope attribute](../scope/README.md#scope-semantic-conventions),
+serves the same purpose. `<scope.short_name>_`, if present, MUST be attached as a
+prefix to all metrics from an Instrumentation Scope.
+
+For each Instrumentation Scope with a `scope.short_name` attribute, Prometheus
+exporters SHOULD generate an info-typed metric named
```
tigrannajaryan, line 1395: What's an `info-typed metric`? Is it an `Info` metric as defined by OpenMetrics?

specification/metrics/data-model.md
```diff
@@ -1337,6 +1382,25 @@ The data point type of an OTLP metric MUST be added as
...
+`<scope.short_name>_opentelemetry_scope_info`. If multiple Instrumentation Scopes have
```
tigrannajaryan, line 1396: >multiple Instrumentation Scopes

What does multiple mean here? Does this try to prevent conflicting Scope definitions from emitting bad Prometheus data? Is this about multiple Scopes that have the same short_name but different name/version? In what context should this be evaluated? Within one OTLP message? Multiple in separately delivered OTLP messages?

specification/metrics/data-model.md
```diff
@@ -1284,6 +1286,49 @@ retrieved from the `trace_id` and `span_id` label keys, respectively.  All
...
+Name and Version of the Instrumentation Scope. The prefix of the metric MUST be
+added to the Instrumentation Scope as the value for the `scope.short_name`
+attribute. Additional labels MUST be added as scope attributes, with keys and
+values unaltered. Other metrics in the batch which have a prefix equal to the
+prefix of the `<prefix>_opentelemetry_scope_info` metric have the matching prefix
+removed, and are placed within that Instrumentation Scope.  For example, the
+Prometheus metrics:
+
+```
+# TYPE otelhttp_opentelemetry_scope_info gauge
```
tigrannajaryan, line 1303: Shouldn't this be?
```suggestion
# TYPE otelhttp_opentelemetry_scope_info info
```

specification/metrics/data-model.md
dashpole: Its a bit confusing.  Info is a [naming convention](https://prometheus.io/docs/practices/naming/#metric-names) for a gauge in prometheus, but is a [type in OpenMetrics](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#info).

specification/metrics/data-model.md
dashpole: > Does this try to prevent conflicting Scope definitions from emitting bad Prometheus data?
> Is this about multiple Scopes that have the same short_name but different name/version?

Yes, and yes.  Prometheus does not allow a single metric to have points with different sets of labels.  Since two scopes may not have the same scope attribute keys, the two produced `<scope.short_name>_opentelemetry_scope_info` metrics would not be valid prometheus.

As an alternative, we could not preserve scope attributes using this mechanism (that was the previous proposal).  We would just have an `opentelemetry_scope_info` metric with one metric point for each scope, with name, version, and short_name labels.

> In what context should this be evaluated? Within one OTLP message? Multiple in separately delivered OTLP messages?

Fundamentally, this should be evaluated within a single scrape of the Prometheus endpoint.  The current SDK spec for [pull exporters](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#pull-metric-exporter) doesn't specify the format metrics are provided to pull exporters in, but i'll assume its a [batch](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#exportbatch) of metrics, as defined for the push exporter, which corresponds to one OTLP message.

specification/metrics/data-model.md
dashpole: Updated to include a reference to the OpenMetrics info type definition

specification/metrics/data-model.md
dashpole: Updated to:
"If multiple Instrumentation Scopes **within the batch of metrics being scraped** have the same`scope.short_name`,"

dashpole:
The two ideas from the spec sig today were:

1. Use the last word of the instrumentation scope name (and also, maybe allow overriding in the exporter w/ a mapping)
2. Change guidance for the scope name to be shorter and readable.  Deal with scope uniqueness separately.

The preference at the meeting seemed to be for the first option, so i'm going to update this PR to propose that.

dashpole:
I've updated the description and proposal to propose using the last word of the instrumentation scope.

tigrannajaryan:
> 1. Use the last word of the instrumentation scope name (and also, maybe allow overriding in the exporter w/ a mapping)

I like this. If we allow overriding it in the exporter we have a complete solution. Name clashes are going to be rare and for that rare case requiring overriding in the exporter is fine. It can even be done via an env variable specific to prometheus exporter so that no code changes are necessary. We may want to have some diagnostics to detect the clashes and show in the logs or otherwise surface the problem.

specification/metrics/data-model.md
```diff
@@ -1434,6 +1476,29 @@ The data point type of an OTLP metric MUST be added as
...
+OpenMetrics prevents naming collisions for metrics from different libraries by
+adding a [metric namespace](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-naming-and-namespaces),
+which is generally a single-word prefix, to metrics. The Instrumentation Scope
+serves a similar purpose in OpenTelemetry, but is too long to be practical as a
+prefix to metrics. Prometheus exporters MUST add a prefix to all metrics within
+an instrumentation scope based on the instrumentation scope name. It is
+recommended to use the last word of the Instrumentation Scope name.  For
+example, a scope with the name
+`go.opentelemetry.io.contrib.instrumentation.net.http.otelhttp` would result in
+`otelhttp_` being added as a prefix to metrics within that scope.
```
arminru, line 1490: While I really appreciate simple solutions, I think this one might not be ideal in many cases.

If I have two scopes (fully qualified class names), say
```
com.example.FooSomething.Impl
io.opentelemetry.BarOtherthing.Impl
```
both would result in the same prefix `Impl_`, which is likely not desirable.

Do you think an example like this is too far-fetched or would it be realistic that the rightmost term is often a very generic suffix or one that easily collides?

dyladan: I agree. i don't think I'm convinced by the statement "Name clashes are going to be rare" made above

tigrannajaryan: >i don't think I'm convinced by the statement "Name clashes are going to be rare" made above

Is there a way that we can figure out how rare we should expect this to be? Perhaps scan all our existing instrumentation names to see how unique the prefixes are?

dyladan: > > i don't think I'm convinced by the statement "Name clashes are going to be rare" made above
> 
> Is there a way that we can figure out how rare we should expect this to be? Perhaps scan all our existing instrumentation names to see how unique the prefixes are?

Maybe instrumentation we provide but we can't possibly scan all of our user code. I think @arminru provided several very reasonable examples where this is not going to work.

joaopgrassi: In .NET is very common to have classes holding [extension methods](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods). In such cases, it's also very common for the class name to end with `Extensions`, as you can see in the official docs there (`MyExtensions`). It's kinda of a convention at this point that everyone relies. 

I did a quick search on grep.app, and you can see how many of these are out there, specially in the .NET runtime repo. I also believe that taking the "last word" is not a very good idea and a lot of conflicts will happen:

https://grep.app/search?q=class%5C%20.%2AExtensions%24&regexp=true&filter[lang][0]=C%23

MrAlias: For what it is worth, this is a recommendation. I think it is a good starting place for instrumentation libraries. And if they find they are colliding and users are complaining they can update their prefix to be more specific (e.g. last two words), right?

specification/metrics/data-model.md
```diff
@@ -1434,6 +1476,29 @@ The data point type of an OTLP metric MUST be added as
...
 
+#### Instrumentation Scope
+
+OpenMetrics prevents naming collisions for metrics from different libraries by
+adding a [metric namespace](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-naming-and-namespaces),
+which is generally a single-word prefix, to metrics. The Instrumentation Scope
+serves a similar purpose in OpenTelemetry, but is too long to be practical as a
+prefix to metrics. Prometheus exporters MUST add a prefix to all metrics within
+an instrumentation scope based on the instrumentation scope name. It is
+recommended to use the last word of the Instrumentation Scope name.  For
```
MrAlias, line 1487:
```suggestion
RECOMMENDED to use the last word of the Instrumentation Scope name.  For
```

specification/metrics/data-model.md
jack-berg: > And if they find they are colliding and users are complaining they can update their prefix to be more specific (e.g. last two words), right?

IMO instrumentation shouldn't have to be aware of the constraints of a particular reader / exporter. If instrumentation follows the general guidelines of using the API, it should work with any of the readers / exporters provided out of the box, including prometheus. 

I like the idea of having a general rule for how the prefix is derived, which works in most cases. But I also like supplementing the prometheus reader / exporter with an easy way to override the prefix when collisions do occur.

Do folks have any better solutions than using the last word? Maybe the last two words?

specification/metrics/data-model.md
```diff
@@ -1434,6 +1476,29 @@ The data point type of an OTLP metric MUST be added as
 [OpenMetrics TYPE metadata](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metricfamily).
 It also dictates type-specific conversion rules listed below.
 
+#### Instrumentation Scope
+
+OpenMetrics prevents naming collisions for metrics from different libraries by
+adding a [metric namespace](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-naming-and-namespaces),
+which is generally a single-word prefix, to metrics. The Instrumentation Scope
```
pirgeo, line 1483: Do I understand correctly that adding a prefix to the `info` object is enough for OpenMetrics compatible receivers to connect the labels on the `info` to all metrics that have the same prefix?

dashpole: It can be done using a join in promql (group_left).  See https://grafana.com/blog/2021/08/04/how-to-use-promql-joins-for-more-effective-queries-of-prometheus-metrics-at-scale/ for how info metrics can be used to supplement existing metrics.  The user would need to do a join for each metric they want to add the additional labels to

specification/metrics/data-model.md
dyladan: I'm just not sure what "last word" even means in my language (JS) where there is no accepted FQ class name syntax

tigrannajaryan:
>Depending on the language, it may fail to create the Prometheus exporter, or may fail to send some, or all metrics.

@dashpole I have realized that I have no idea why this is a problem, why it may fail to create exporter or fail to send the metrics. Can you please clarify this?

tigrannajaryan:
>A prometheus exporter which receives both of these metrics would not be able to serve both of those histograms.

Is this a problem only for histograms or other metric types as well?

dashpole:
> I have realized that I have no idea why this is a problem, why it may fail to create exporter or fail to send the metrics. Can you please clarify this?

@tigrannajaryan Prometheus itself will do some [validation on registration](https://github.com/prometheus/client_golang/blob/5b7e8b2e6716df0ceda9df81feb15910c7efa150/prometheus/collector.go#L34) ([golang implementation](https://github.com/prometheus/client_golang/blob/5b7e8b2e6716df0ceda9df81feb15910c7efa150/prometheus/registry.go#L288)), and will do [additional validation](https://github.com/prometheus/client_golang/blob/5b7e8b2e6716df0ceda9df81feb15910c7efa150/prometheus/registry.go#L910) at collection time to ensure uniqueness of metric points.  I'm mostly familiar with the go implementation, but I suspect other languages do similar validation.  OpenMetrics [requires](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metricfamily) a unique labelset per-metric-name, and a [unique name](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#name) within metrics on an endpoint.  Prometheus has similar requirements for [name uniqueness](https://github.com/prometheus/docs/blob/main/content/docs/instrumenting/exposition_formats.md#comments-help-text-and-type-information) and [label-set uniqueness](https://github.com/prometheus/docs/blob/main/content/docs/instrumenting/exposition_formats.md#grouping-and-sorting).

> Is this a problem only for histograms or other metric types as well?

It is a problem for all metric types.

specification/metrics/data-model.md
dashpole: Thanks for the feedback @arminru, @dyladan and @joaopgrassi.  I think adding a default in this way would likely end up doing more harm than good.  I raised this for discussion at the Spec sig meeting this morning, and proposed using naming conventions, rather than the "last word".  I'll update the PR shortly to reflect the new proposal.

specification/metrics/data-model.md
```diff
@@ -1434,6 +1476,29 @@ The data point type of an OTLP metric MUST be added as
...
+recommended to use the last word of the Instrumentation Scope name.  For
+example, a scope with the name
+`go.opentelemetry.io.contrib.instrumentation.net.http.otelhttp` would result in
+`otelhttp_` being added as a prefix to metrics within that scope.
+
+For each Instrumentation Scope, Prometheus exporters SHOULD generate an
+[Info](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#info)-typed
+metric named `<prefix>_opentelemetry_scope_info` using the same prefix used
+above. If multiple Instrumentation Scopes within the batch of metrics being
+scraped would have the same prefix, do not generate either metric. If present, Instrumentation
```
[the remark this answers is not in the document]
dashpole: No longer applicable

specification/metrics/data-model.md
dashpole: no longer applicable

specification/metrics/data-model.md
dashpole: I've updated the proposal.

dashpole:
I've updated the proposal to change the naming guidance for metrics instead of adding any conversion logic.  I've also removed the `opentelemetry_scope_info` metric from this PR, and will propose that separately.

ahayworth (changes requested):
I feel (very strongly) that this kind of guidance should be scoped in a section called "Special Notes for Prometheus Exporters", or something to that effect. For example, the [OpenTelemetry to Jaeger](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/jaeger.md) docs might be a good source of inspiration.

I don't think that we should push this guidance on the _entire_ OpenTelemetry metrics ecosystem based on the limitations of one family of exporters. Even if Prometheus is a very common TSDB, we should not presume that it is the only one out there or that all metrics systems operate with precisely the same constraints (either presumption can be trivially disproved). The fact that Prometheus client libraries choke on this scenario doesn't mean that the entire standard needs to conform to the specifics of its implementation.

The problems that arose from encoding information into metric names is part of why [metrics2.0](http://metrics20.org/) came about in the first place. That isn't to say one should _never_ encode information into a metric's name, but only that we should be very careful about doing so, when absolutely needed.

Again - this guidance is fine, and probably even necessary. But it should be scoped as part of "additional guidance for Prometheus-like exporters" or something to that effect, rather than imposing _SHOULD_-level guidance on all of OpenTelemetry.

dashpole:
@tigrannajaryan It would be problematic if they had the same labels.  E.g. 
```
http_client_duration_sum{http_url="/foo"} 1
http_client_duration_sum{http_url="/bar"} 2
http_client_duration_sum{http_url="/bar"} 3
```

Would be invalid.

I also realized after the spec meeting that I was wrong about prometheus disallowing different label keys on a single metric.  They initially implemented that requirement in prometheus, but then reversed course and removed the requirement after OpenMetrics decided not to include it: https://github.com/prometheus/client_golang/pull/417

That means the "infeasible alternative" above: `Add the namespace as a label to all Prometheus metrics instead of a prefix` is actually a viable option, and one that would be less imposing on the instrumentation ecosystem (to @ahayworth's feedback).  The only downside of that approach would be handling potential conflicting descriptions/units from collisions, but that seems like an acceptable tradeoff.

Sorry for the frequent changes.  I'm hopeful this will be the last round of significant changes.

tigrannajaryan:
>It would be problematic if they had the same labels

Only if the have the exact same timestamp, right? If the timestamps are different then that's OK, that's just 2 different datapoints. I am not sure what the chances of this happening are. I guess depends on timestamp precision and whether the 2 libraries happen to be installed in such a way that they intercept the same http call in this example. If this does happen then perhaps that's an acceptable situation (i.e. the first to record the datapoint wins)? I assume this won't result in crashes or anything, simply the duplicates will be ignored, which I think is an acceptable limitation.

>That means the "infeasible alternative" above: Add the namespace as a label to all Prometheus metrics instead of a prefix is actually a viable option, and one that would be less imposing on the instrumentation ecosystem

Yes. So, essentially the output would look like this:

```
http_client_duration_sum{http_url="/foo" scope=otelmux} 1
http_client_duration_sum{http_url="/bar" scope=otelmux} 2
http_client_duration_sum{http_url="/bar" scope=otelhttp} 3
```

We can also use the full Scope name here, I don't think there is a need to shorten it.

>The only downside of that approach would be handling potential conflicting descriptions/units from collisions, but that seems like an acceptable tradeoff.

This should not happen if libraries follow the semantic conventions. 

One more thing that may happen is that libraries make different choices for what labelsets to use (if semantic convention allows for variation). This should be fine as well, right? If I understand it correctly in Prometheus it is valid to do this:

```
http_client_duration_sum{http_url="/foo" http_method=put scope=otelmux} 1
http_client_duration_sum{http_url="/bar" http_method=get scope=otelmux} 2
http_client_duration_sum{http_url="/bar" scope=otelhttp} 3
```

dashpole:
> Only if they have the exact same timestamp, right? 

No.  Prometheus/OpenMetrics metric points are required to have a [unique labelset](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric) within the MetricFamily (i.e. metric name + label set must be unique).  Prometheus also recommends against adding explicit timestamps unless serving (very) old metrics.  This is avoided by the introduction of an additional label.

> We can also use the full Scope name here, I don't think there is a need to shorten it.

Correct.

> One more thing that may happen is that libraries make different choices for what labelsets to use (if semantic convention allows for variation). This should be fine as well, right?

Yes.  Metrics [SHOULD](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric) (not MUST) have the same label keys.  I think this approach abides by that statement.

tigrannajaryan:
> > We can also use the full Scope name here, I don't think there is a need to shorten it.
> 
> Correct.
> 
> > One more thing that may happen is that libraries make different choices for what labelsets to use (if semantic convention allows for variation). This should be fine as well, right?
> 
> Yes. Metrics [SHOULD](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric) (not MUST) have the same label keys: `Metrics with the same name for a given MetricFamily SHOULD have the same set of label names in their LabelSet.` I think this approach abides by that statement.

So, if this is the case then we don't need to do any prefixing and don't need a short name or similar, right? We add the scope name as a label, we require all libraries to properly honour semantic conventions such the description, units, data point types are the same regardless of what library emits it and this should be exportable to Prometheus. 

Does this solve the problem this PR tries to address or there is more to it?

I wonder if there are any downsides to adding the scope name as a label? Supposedly it is a low cardinality field, so should be fine. Users may not care to see query results broken down by libraries, so they may have to always aggregate to eliminate the scope name in their queries. I don't know how easy is this to do in PromQL, it may or may not be a problem.

dashpole:
> Does this solve the problem this PR tries to address or there is more to it?

This solves it.  Nothing more to it.

> I wonder if there are any downsides to adding the scope name as a label? Supposedly it is a low cardinality field, so should be fine. Users may not care to see query results broken down by libraries, so they may have to always aggregate to eliminate the scope name in their queries. I don't know how easy is this to do in PromQL, it may or may not be a problem.

Aggregating away labels is easy to do in promql.  From [these examples](https://prometheus.io/docs/prometheus/latest/querying/examples/): `http_requests_total{job="apiserver", handler="/api/comments"}` would aggregate away labels other than job and handler.

As for cardinality, it should maintain the overall cardinality of the OTel metrics, since those are already separated by scope.

Downsides I can think of are:

* Increased size of the (text) scrape result means higher memory usage for both the application and scraper.
* Users likely won't want to often break down metrics by instrumentation scope.  The extra labels may be seen as a nuisance.
* Metric descriptions can still (theoretically) conflict.
* We are intentionally not following Prometheus/OpenMetrics [guidance](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-naming-and-namespaces) around metric naming: `Metric names SHOULD indicate what piece of code they come from`.

tigrannajaryan:
> * We are intentionally not following Prometheus/OpenMetrics [guidance](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-naming-and-namespaces) around metric naming: `Metric names SHOULD indicate what piece of code they come from`.

This is probably out of topic, bit I am not sure this is a good recommendation. The ability to aggregate http request time across libraries seems very useful to me. If I want to know the response time of a particular destination, I do not necessarily care which source is making request (what application or what library). Recommending that metrics are different for each source makes aggregation impossible/difficult.

This guidance also seems to be contradicting what Prometheus recommends to make a label:
```
as a rule of thumb, either the sum() or the avg() over all dimensions of a given metric should be meaningful 
(though not necessarily useful). If it is not meaningful, split the data up into multiple metrics. For example, 
having the capacity of various queues in one metric is good
```
This seems to be one of the cases when aggregation is meaningful.

dashpole:
> This is probably out of topic, bit I am not sure this is a good recommendation. The ability to aggregate http request time across libraries seems very useful to me.

Prometheus doesn't have semantic conventions to align instrumentation in that way.  Prometheus assumes users often would name things the same even when they are different things.  I think we would be making a reasonable decision to not add prefixes given our conventions.  I agree with the utility of being able to aggregate across sources.

The prometheus guidance on the total being meaningful just tells you when you should **split** a metric, not when you should combine.  It is intended to prevent someone from adding, for example, a memory metric with `user`, `system`, and `total` dimensions, as the sum across labels wouldn't be meaningful.  From the [OM Spec](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-names-versus-labels): `There are situations in which both using multiple Metrics within a MetricFamily or multiple MetricFamilies seem to make sense.`.  It then gives an example of success vs failure being a case where separate metrics is better than a single one with labels.

joaopgrassi:
Hi @dashpole 
Giving the latest (great) discussions and the last update you did, I think a rename on the title of the PR is in place? Right now it's not aligned with the chosen approach.

specification/metrics/data-model.md
```diff
@@ -1434,6 +1478,19 @@ The data point type of an OTLP metric MUST be added as
 [OpenMetrics TYPE metadata](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metricfamily).
 It also dictates type-specific conversion rules listed below.
 
+#### Instrumentation Scope
+
+Prometheus exporters SHOULD generate an [Info](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#info)-typed
+metric named `opentelemetry_scope_info`. If present, Instrumentation
+Scope `name` and `version` MUST be added as `opentelemetry_scope_name` and
```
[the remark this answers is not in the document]
dashpole: The best we could do would be `otel_scope_name` and `otel_scope_version` because prom doesn't support `.` characters.  Do you think that would satisfy the existing spec, or should I still include the "takes precedence" language in the non-otlp spec?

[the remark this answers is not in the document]
tigrannajaryan: >The best we could do would be otel_scope_name and otel_scope_version because prom doesn't support . characters.

I think this is good enough and matches the existing rules of how to convert unsupported characters to Prometheus.

tigrannajaryan (approved):

specification/metrics/data-model.md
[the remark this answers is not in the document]
joaopgrassi: Yeah I'm also ok with that 👍

specification/metrics/data-model.md
```diff
@@ -1385,6 +1387,47 @@ retrieved from the `trace_id` and `span_id` label keys, respectively.  All
 labels not used for the trace and span ids MUST be added to the OpenTelemetry
 exemplar as attributes.
 
+#### Instrumentation Scope
+
+Each `opentelemetry_scope_info` metric point present in a batch of metrics
```
joaopgrassi, line 1392: I guess the info metric can still have `opentelemetry_`? Not sure what is best, keep it like that = clear what it is vs use `otel` so it's consistent with the labels but could be "strange" if one is not aware of what it means. But again the labels are already using `otel` so.. 🤔

specification/metrics/data-model.md
```diff
@@ -1445,6 +1488,19 @@ The data point type of an OTLP metric MUST be added as
...
+
+Prometheus exporters SHOULD generate an [Info](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#info)-typed
+metric named `opentelemetry_scope_info`. If present, Instrumentation Scope
+`name` and `version` MUST be added as `otel_scope_name` and
+`otel_scope_version` labels. Scope attributes MUST also be added as labels
+following the rules described in the [`Metric Attributes`](#metric-attributes)
+section below.
+
+Prometheus exporters MUST add the scope name as the `otel_scope_name` label and
+SHOULD add the scope version as the `otel_scope_version` label on all metrics
```
joaopgrassi, line 1501: When converting from Prometheus to OTLP it says:

>[...] metrics in the batch which have `otel_scope_name` and `otel_scope_version`
labels that match an instrumentation scope MUST be placed within the matching
instrumentation scope.

Here we say exporters MUST add the scope name but it uses SHOULD for version. Was there a reason for this? Sorry if it was already explained somewhere else but I couldn't find it.

So I wonder if we should change it so version is also a must? Just thinking if implementations will be more "complicated" because they have to check, do I have the name and version, or just the name? For ex, there are metrics with name+version populated and then another metric with just name. Does it go together with the others? Or do I create a new scope with version "0" to put the one without version? How does this affect the case when a same scope name has 2 versions?

dashpole: Thinking though it some more, I do think version needs to be required.  I had thought that some users might prefer a lighter-weight approach that didn't allow joining metrics with scope info as easily, but it would have collision issues if someone ran two versions of the same library in their app somehow.

specification/metrics/data-model.md
tigrannajaryan: First hit for `otel` on Google is OpenTelemetry so I think it is fine to use the abbreviation.

dashpole: updated to otel_scope_info

specification/metrics/data-model.md
dashpole: updated to must

MrAlias (approved):

joaopgrassi (approved):
Looks good! Last thing: Not sure if we need a changelog entry for this case?

tigrannajaryan:
> Looks good! Last thing: Not sure if we need a changelog entry for this case?

I think we do.

dashpole:
Added a changelog

specification/metrics/data-model.md
```diff
@@ -1385,6 +1387,47 @@ retrieved from the `trace_id` and `span_id` label keys, respectively.  All
...
+  metrics:
+  - name: http_server_duration
+    data:
+      sum:
+        data_points:
+        - value: 1
+```
+
+Metrics which are not found to be associated with an instrumentation scope are
+all placed within an empty instrumentation scope, and do not have any labels
```
carlosalberto, line 1428: MUST?

carlosalberto (approved):

specification/metrics/data-model.md
dashpole: yes.  It now reads:

```
Metrics which are not found to be associated with an instrumentation scope MUST
all be placed within an empty instrumentation scope, and MUST not have any labels
```

jack-berg (approved):

jmacd:
@dashpole I anticipate users might react to the new metric attributes in their Prometheus data and come to request an option to disable these attributes, in exchange for the risk of exporter errors caused by duplicates. Do you see any potential problems with offering such an option?

dashpole:
I don't see a problem with that, and think it is a good idea.  I added that to this change.

jmacd (approved):

carlosalberto:
@ahayworth Hey, could you take a look again? You had requested changes on this PR, but it has changed ever since ;)

dashpole:
I believe @ahayworth's [feedback](https://github.com/open-telemetry/opentelemetry-specification/pull/2703#pullrequestreview-1106528437), which was that we should not recommend adding a prefix to all metrics, has been addressed.

I'm happy with the current state, and would like to see this merged unless others have feedback.

jmacd (approved):

tigrannajaryan:
@open-telemetry/specs-metrics-approvers @open-telemetry/technical-committee I will merge this tomorrow unless I see objections.

ahayworth:
@carlosalberto Thank you for poking me: I agree with @dashpole that my feedback has indeed been addressed. ❤️

ahayworth (approved):

reyang (approved):

CHANGELOG.md
```diff
@@ -46,6 +46,8 @@ release.
   ([#2770](https://github.com/open-telemetry/opentelemetry-specification/pull/2770)).
 - Fix OpenMetrics valid label keys, and specify prometheus conversion for metric name.
   ([#2788](https://github.com/open-telemetry/opentelemetry-specification/pull/2788))
+- Specify how Prometheus exporters and receivers handle instrumentation scope.
```
carlosalberto, line 49: Update CHANGELOG needed. Otherwise we are ready to go!

dashpole: done

PR #2705, opened 2022-08-02
Add process.threads

atoulme:
## Changes

Adds a new metric, `process.threads`, to the process metrics available.

Related issues https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/12482


specification/metrics/semantic_conventions/process-metrics.md
```diff
@@ -43,6 +43,7 @@ Below is a table of Process metric instruments.
...
 | `process.network.io.receive`      | Counter                                           | By    | Network bytes received.                                                                                                             |                                                                                                                                                                                                 |
 | `process.network.io.transmit`     | Counter                                           | By    | Network bytes transmitted.                                                                                                          |                                                                                                                                                                                                 |
+| `process.threads.count`           | UpDownCounter                                     | By   | Process threads count.                                                                                                             |
```
cijothomas, line 46: the unit is incorrect.

atoulme: Oh, I had assumed by was per process. OK. I guess I could go with using `none` ?

atoulme: Found the standard - using {threads}

dmitryax (approved):
Please update https://github.com/open-telemetry/opentelemetry-specification/blob/main/CHANGELOG.md as well

specification/metrics/semantic_conventions/process-metrics.md
```diff
@@ -31,18 +31,19 @@ metrics](runtime-environment-metrics.md).
...
+| `process.network.io.transmit`     | Counter                                           | By         | Network bytes transmitted.                                                                                                          |                                                                                                                                                                                                 |
+| `process.threads.count`           | UpDownCounter                                     | {threads}  | Process threads count.                                                                                                              |                                                                                                                                                                                                 |
```
arminru, line 46: Per our naming guidelines this should be `process.threads` (without the `.count` suffix) unless we expect more attributes to be added under `process.threads.*` and thus forming a namespace:
https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/metrics/semantic_conventions#pluralization

Are there any other related attributes used in the collector?

atoulme: Not at this time. We could report thread time stats, but that's rather heavy. I will go with your suggestion and rename.

atoulme:
Changelog updated, thanks!

specification/metrics/semantic_conventions/process-metrics.md
tigrannajaryan: If we ever intend to add any other thread related metric in the future then we should keep the `.count` suffix.

specification/metrics/semantic_conventions/process-metrics.md
```diff
@@ -31,18 +31,19 @@ metrics](runtime-environment-metrics.md).
...
+| `process.network.io.transmit`     | Counter                                           | By         | Network bytes transmitted.                                                                                                          |                                                                                                                                                                                                 |
+| `process.threads`           | UpDownCounter                                     | {threads}  | Process threads count.                                                                                                              |                                                                                                                                                                                                 |
```
ThomsonTan, line 46: nit: align `UpDownCounter` with the other rows in the second column.

specification/metrics/semantic_conventions/process-metrics.md
atoulme: I missed this comment. I am not sure. We can wait a bit and discuss in a meeting.

reyang (approved):

jack-berg (approved):

arminru (approved):

atoulme:
Here is a bit I'll present tomorrow at the spec meeting:
I believe process.threads is a finite metric, and will not be used as a namespace. An other namespace, under thread.*, should be created if we ever standardize data under a thread. This is because listening to thread activity would typically be done via a different type of monitor than a monitor for processes. For threads, we would use instrumentation in-process, such as jmx, to listen to thread activity.

carlosalberto (approved):

jmacd (changes requested):
I feel that this semantic convention belongs in the `process.runtime.*` section. In Go, for example, there is no "thread" concept and we already have a runtime-metrics count of goroutines.

The suffix `.count` is problematic because of the way it will translate into Prometheus/OpenMetrics. I would suggest that `process.runtime.threads` is a fine name, however--you will know it's a count because of the point kind.

bogdandrutu:
@jmacd are you proposing `process.runtime.thread.count`?

reyang:
> I feel that this semantic convention belongs in the `process.runtime.*` section. In Go, for example, there is no "thread" concept and we already have a runtime-metrics count of goroutines.

I think the fact that Golang and Node.js don't have `thread` is a good reason why this shouldn't be part of `process.runtime.*`. I feel threads is an operating system concept. Although there are indeed operating systems that don't have threads, but it seems pretty common to park it under OS and process.

reyang:
> @jmacd are you proposing `process.runtime.thread.count`?

`process.runtime.thread.count` might have a different semantic. I'll take .NET as an example, `process.thread.count` could mean "how many threads does this process have (from the operating system's point of view)", while "process.runtime.thread.count` could mean "how many [managed threads](https://docs.microsoft.com/en-us/dotnet/standard/threading/managed-threading-basics) does the .NET runtime inside this process have" (and there could be a more complex situation where multiple instances of runtime coexist, although I've never seen that in reality except for some SQL stored procedure hosting).

atoulme:
I agree with @reyang - as said on Slack:
> the process.runtime namespace does not allow to set a threads entry in it, since it is supposed to process.runtime.{environment} as per spec: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/runtime-environment-metrics.md
Second, the number of threads is the number of threads exposed at the OS level, not the level of threads or goroutines as experienced inside the program. That's the difference say between looking at Java threads and the number of threads in `top -H -p <pid>`.

atoulme:
@jmacd and anyone - any insights or follow up?

carlosalberto:
Let's discuss it on today's call (with whoever can make it) - hopefully we can merge this (either with minor renaming or not) today ;)

bogdandrutu:
> I feel that this semantic convention belongs in the process.runtime.* section. In Go, for example, there is no "thread" concept and we already have a runtime-metrics count of goroutines.

Inside the go VM indeed cannot read threads, from outside you can actually read the number of OS threads for the go process.

atoulme:
What is the next step here?

carlosalberto:
@jmacd Do you have a strong instance against going with `process.threads`? This was briefly discussed in the past Spec call, and the Golang case was deemed as a very-specific case compared to the rest of the languages, when it comes to this item, so we could deem it as a special case (Python also has coroutines, but it _also_ has threads - that is the closest IIRC).

jmacd (approved):

PR #2706, opened 2022-08-02
Add additional process metrics to the metrics semantic conventions

### LINKED ISSUE #2708: Add additional process metrics to the metrics semantic conventions

I would like to add the following metrics to the metrics semantic conventions:
- `process.open_file_descriptors`: Count of file descriptors a process has open
- `process.voluntary_ctx_switches`: Count of times a process was voluntarily context switched
- `process.involuntary_ctx_switches`: Count of times a process was involuntarily context switched

**Additional context.**

Related issues https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/12482

For the context switch metrics, I named them with the `ctx` abbreviation for "context" since that appears to be standard with vendors and in Linux, where it is abbreviated as `ctxt`. Additionally, things such as "input/output" are abbreviated to "io" in the process metrics semantic conventions. However, if it would be preferred, I think the full name of `context` would also be fine for the metric names.

I have opened a PR for this change here: https://github.com/open-telemetry/opentelemetry-specification/pull/2706

evan-bradley:
## Changes

Fixes #2708

Add the following process metrics to the semantic conventions:
- `process.context_switches`
- `process.open_file_descriptors`

Related issues https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/12482

This will need to be rebased on https://github.com/open-telemetry/opentelemetry-specification/pull/2705 before it can be merged.


specification/metrics/semantic_conventions/process-metrics.md
```diff
@@ -31,18 +31,21 @@ metrics](runtime-environment-metrics.md).
...
+| `process.network.io.receive`       | Counter                                            | By                 | Network bytes received.                                                                                                             |                                                                                                                                                                                                 |
+| `process.network.io.transmit`      | Counter                                            | By                 | Network bytes transmitted.                                                                                                          |                                                                                                                                                                                                 |
+| `process.open_file_descriptors`    | UpDownCounter                                      | {file descriptors} | Number of file descriptors in use by the process.                                                                                   |
```
bogdandrutu, line 46: If anything related to this in filesystem https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/system-metrics.md#systemfilesystem---filesystem-metrics we should be consistent.

specification/metrics/semantic_conventions/process-metrics.md
```diff
@@ -31,18 +31,21 @@ metrics](runtime-environment-metrics.md).
...
+| `process.voluntary_ctx_switches`   | Counter                                            | {context switches} | Number of times the process has been voluntarily context switched.                                                                  |
+| `process.involuntary_ctx_switches` | Counter                                            | {context switches} | Number of times the process has been involuntarily context switched.                                                                |
```
bogdandrutu, line 48: Are we ok with the "ctx" abbreviation? Should we use "context" ?

specification/metrics/semantic_conventions/process-metrics.md
evan-bradley: Since file descriptors are just an abstraction and may not necessarily point to anything on the filesystem, I don't think we should namespace this under `filesystem`. Additionally, none of the attributes apply here since this is just a number. Additional metadata would have to be retrieved on a per-file descriptor basis.

specification/metrics/semantic_conventions/process-metrics.md
evan-bradley: I originally used "ctx" since that seemed to be the standard in Linux and in vendors that collect this metric, but if there is any uncertainty, I think "context" is a better choice. I've changed it to "context".

specification/metrics/semantic_conventions/process-metrics.md
```diff
@@ -31,18 +31,21 @@ metrics](runtime-environment-metrics.md).
...
+| `process.network.io.transmit`          | Counter                                            | By                 | Network bytes transmitted.                                                                                                          |                                                                                                                                                                                                 |
+| `process.open_file_descriptors`        | UpDownCounter                                      | {file descriptors} | Number of file descriptors in use by the process.                                                                                   |
+| `process.voluntary_context_switches`   | Counter                                            | {context switches} | Number of times the process has been voluntarily context switched.                                                                  |
```
joaopgrassi, line 47: I'm not an specialist in units standards/UCUM, but I wonder if having a space in `{context switches}` is OK. [Reading the UCUM](https://ucum.org/ucum.html#section-Style) spec I found:

> White space is not recognized in a a unit term and should generally not occur. UCUM implementations may flag whitespace as an error rather than ignore it. Whitespace is not used as a separator of otherwise ambiguous parts of a unit term. 

Then I also saw there's underscore mention in §13

> §13 underscore  1 When in print a unit would have a subscript, an underscore (‘_’) is used to separate the subscript from the stem of the unit symbol.  2 The subscript is part of the unit atom.  3 subscripts are used to disambiguate the two units with the same name but different meanings. 

We don't have a subscript here, but maybe it's still OK to have `{context_switches}`?

specification/metrics/semantic_conventions/process-metrics.md
```diff
@@ -31,18 +31,21 @@ metrics](runtime-environment-metrics.md).
...
-| `process.network.io.transmit`     | Counter                                           | By    | Network bytes transmitted.                                                                                                          |                                                                                                                                                                                                 |
+| Name                                   | Instrument Type ([\*](README.md#instrument-types)) | Unit               | Description                                                                                                                         | Labels                                                                                                                                                                                          |
```
joaopgrassi, line 34:
```suggestion
| Name                                   | Instrument Type ([\*](README.md#instrument-types)) | Unit ([UCUM](README.md#instrument-units))               | Description                                                                                                                         | Labels                                                                                                                                                                                          |
```

I know it's not part of your change, but just trying to make the page a bit more consistent with the others. Hopefully soon we will be able to generate these automatically.

specification/metrics/semantic_conventions/process-metrics.md
```diff
@@ -31,18 +31,21 @@ metrics](runtime-environment-metrics.md).
...
+| `process.open_file_descriptors`        | UpDownCounter                                      | {file descriptors} | Number of file descriptors in use by the process.                                                                                   |
+| `process.voluntary_context_switches`   | Counter                                            | {context switches} | Number of times the process has been voluntarily context switched.                                                                  |
+| `process.involuntary_context_switches` | Counter                                            | {context switches} | Number of times the process has been involuntarily context switched.                                                                |
```
joaopgrassi, line 48:
```suggestion
| `process.involuntary_context_switches` | Counter                                            | {context switches} | Number of times the process switched context involuntarily.                                                                |
```

Maybe?

specification/metrics/semantic_conventions/process-metrics.md
joaopgrassi, line 47:
```suggestion
| `process.voluntary_context_switches`   | Counter                                            | {context switches} | Number of times the process switched context voluntarily.                                                                  |
```

Maybe?

CHANGELOG.md
```diff
@@ -24,6 +24,8 @@ release.
   ([#2619](https://github.com/open-telemetry/opentelemetry-specification/pull/2619)).
 - Clarify async instrument callback identity.
   ([#2538](https://github.com/open-telemetry/opentelemetry-specification/pull/2538)).
+- Add `process.voluntary_context_switches`, `process.involuntary_context_switches`,
+  and `process.open_file_descriptors`, to the metrics semantic conventions.
```
joaopgrassi, line 28:
```suggestion
  and `process.open_file_descriptors`, to the metrics semantic conventions ([#2706](https://github.com/open-telemetry/opentelemetry-specification/pull/2706)).
```
Let's make sure the PR number is part of the changelog entry. I tried adding it via a suggestion but I'm not sure if it looks right.

specification/metrics/semantic_conventions/process-metrics.md
arminru, line 48: We could consider using
```
process.context_switches.voluntary
process.context_switches.involuntary
```
to group them and to make them easier to discover in a backend - but either should be fine.

joaopgrassi (commented):
Unrelated with the PR change, but I noticed the page is using "Labels". Should we change everywhere to use "Attributes" instead?

specification/metrics/semantic_conventions/process-metrics.md
arminru: We should also change the "Labels" column to align with the others:
```
| Name                                       | Description                                              | Unit    | Unit ([UCUM](README.md#instrument-units)) | Instrument Type ([*](README.md#instrument-types)) | Value Type | Attribute Key | Attribute Values      |
```
This should better go into a separate PR as a follow-up, however.

specification/metrics/semantic_conventions/process-metrics.md
evan-bradley: I had considered this, but it seemed unconventional. If there is consensus around this, I also think it would make sense.

specification/metrics/semantic_conventions/process-metrics.md
joaopgrassi: I also wonder, since this is basically counting the # of times a process switched context, couldn't we use the default unit `1`? As in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/README.md#instrument-units

evan-bradley: I was wondering the same, but couldn't find any consensus around this elsewhere in the specification. Even within the same document, the Kafka semantic conventions use both a space and an underscore:

[Example with a space](https://github.com/open-telemetry/opentelemetry-specification/blob/ca593258953811eb33197976b8e565299b874853/specification/metrics/semantic_conventions/instrumentation/kafka.md?plain=1#L40)

[Example with an underscore](https://github.com/open-telemetry/opentelemetry-specification/blob/ca593258953811eb33197976b8e565299b874853/specification/metrics/semantic_conventions/instrumentation/kafka.md?plain=1#L65)

My interpretation is that the proper way according to UCUM is to have it be `{switches_context}`, since the unit is more or less "switches" and in this case they are "context switches", but that doesn't read well.

It would also seem that we could use an apostrophe, accrdoing to §15: 

> §15 apostrophe       1 The apostrophe (‘'’) is used to separate words or abbreviated words in a multi-word unit symbol.

They use this in a similar way to how we are using "context switches" later on:

> 


name | kind of quantity | print | c/s
-- | -- | -- | --
kilogram of wet tissue | mass | kg(wet tissue) | kg{wet'tis}

That also reads a little strange to me, though should also be correct according to UCUM.

specification/metrics/semantic_conventions/process-metrics.md
evan-bradley: I've noticed the metrics semantic conventions documents have quite a few inconsistencies between them, but decided it would be best to leave this one mostly as-is. Is there a document that can be used as a template for how these columns should be titled? I would be glad to open a PR to standardize everything.

CHANGELOG.md
evan-bradley: Good call, I will add this.

specification/metrics/semantic_conventions/process-metrics.md
evan-bradley: The reason I phrased it like this is because usually a kernel will context switch a process, I don't believe the process will perform context switches on most modern operating systems (though it can voluntarily signal it is okay to be context-switched). If you think this wording is unclear, I don't see a reason to be pedantic about the mechanism.

specification/metrics/semantic_conventions/process-metrics.md
evan-bradley: See my [other comment](https://github.com/open-telemetry/opentelemetry-specification/pull/2706#discussion_r938910149); I don't believe processes ordinarily will perform context switching themselves, but am open to rewording this for clarity.

specification/metrics/semantic_conventions/process-metrics.md
joaopgrassi: Fair enough, let's not get distracted and tackle this in a separated PR. I think we can at least add the UCUM link though.

specification/metrics/semantic_conventions/process-metrics.md
joaopgrassi: I read the UCUM [specification](https://ucum.org/ucum.html#section-Examples-for-some-Non-Units.) more and found this:

>Users are free to use curly braces expressions ([§12](https://ucum.org/ucum.html#para-12)) if they think it is important to use symbols rather than the default unit 1.  3 Curly braces expressions are equivalent to the unit 1

So, either `1` or `{context_switches}` or even `{count}` are the same in the end I think. It seems to me the metric name is already clear so we don't need any additional meaning on the unit of it. WDYT?

specification/metrics/semantic_conventions/process-metrics.md
joaopgrassi: No reason, I just thought it could be simplified a bit but I'm also fine as is.

arminru: It could make sense to keep track of the total (either by not setting the attribute if specified there, or by introducing an additional `process.context_switches.total` metric) if it cannot always be distinguished by instrumentation whether it was voluntary or involuntary. I'm not versed enough on this topic to tell if that's ever the case, however.

evan-bradley: @arminru Typically operating systems always report voluntary and involuntary context switches separately, so distinguishing them should not be an issue.

evan-bradley: I think that reporting these as separate metrics makes sense since they will likely be tracked independently, and this seems to be in line with the recent change in guidance for metrics semantic conventions that @joaopgrassi brought up. I'll change the metric names to the following:

```
process.context_switches.voluntary
process.context_switches.involuntary
```

specification/metrics/semantic_conventions/process-metrics.md
evan-bradley: I agree, the metric name is descriptive enough. I will change the unit to `{count}`.

CHANGELOG.md
```diff
@@ -24,6 +24,9 @@ release.
...
```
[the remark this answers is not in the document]
evan-bradley: Good catch, I will change that.

joaopgrassi (approved):

CHANGELOG.md
```diff
@@ -24,6 +24,9 @@ release.
   ([#2619](https://github.com/open-telemetry/opentelemetry-specification/pull/2619)).
 - Clarify async instrument callback identity.
   ([#2538](https://github.com/open-telemetry/opentelemetry-specification/pull/2538)).
+- Add `process.context_switches.voluntary`, `process.context_switches.involuntary`,
```
arminru, line 27: test

specification/metrics/semantic_conventions/process-metrics.md
arminru: I think `{context_switches}` and `{file_descriptors}` would provide more value for telemetry consumers that handle those properly and, for example, render it in the UI. `{count}` is not wrong, however, and we can iterate over this before calling this part of the spec stable.

arminru (approved):

reyang (approved):

specification/metrics/semantic_conventions/process-metrics.md
arminru: I created https://github.com/open-telemetry/opentelemetry-specification/issues/2719 so we don't forget about adding guidance and aligning the conventions before going GA with them.

carlosalberto (approved):

specification/metrics/semantic_conventions/process-metrics.md
```diff
@@ -31,18 +31,21 @@ metrics](runtime-environment-metrics.md).
...
+| `process.open_file_descriptors`        | UpDownCounter                                      | {count}                                   | Number of file descriptors in use by the process.                                                                                   |
+| `process.context_switches.voluntary`   | Counter                                            | {count}                                   | Number of times the process has been voluntarily context switched.                                                                  |
+| `process.context_switches.involuntary` | Counter                                            | {count}                                   | Number of times the process has been involuntarily context switched.                                                                |
```
bogdandrutu, line 48: Does it make sense to have "voluntary/involuntary" as a label, since the sum of them is total?

arminru: see https://github.com/open-telemetry/opentelemetry-specification/pull/2706#discussion_r938876419 on this topic

jmacd (approved):

specification/metrics/semantic_conventions/process-metrics.md
carlosalberto: Right, this makes sense given the precedent with the 'direction' metrics change. Any opinion on that @bogdandrutu ?

bogdandrutu: That is not what I understand. The direction was split because the total did not make sense (the arguments were that sum over the direction does not get total usage, since there were other type of usages as well), is the case here as well?

bogdandrutu: See the https://github.com/open-telemetry/opentelemetry-specification/issues/2589 which explains why that did not make sense, is the case here as well?

specification/metrics/semantic_conventions/process-metrics.md
bogdandrutu: I think the direction example should be better understand and not give as an example without explaining why is similar. That issue had a motivation that the total contains more than just the 2 labels, so was not correct. If this is not the case here we should have 1 metric with 1 attribute. See motivation here https://github.com/open-telemetry/opentelemetry-specification/issues/2589

carlosalberto (changes requested):
Blocking this till #2726 is resolved (we will merge it with the proper changes after that, etc).

evan-bradley:
I've changed this to follow the direction taken in https://github.com/open-telemetry/opentelemetry-specification/pull/2748. I'd like to continue with using a label for now, and later change the approach if necessary after https://github.com/open-telemetry/opentelemetry-specification/issues/2726 is resolved.

astencel-sumo:
@carlosalberto I believe this is unblocked now, is that right?

@evan-bradley is there anything more needed here? (other than refreshing the PR)

carlosalberto (approved):

cijothomas (approved):

Issue #1175, opened 2020-11-02
OpenCensus Compatibility for GA


andrewhsu:
_This is a high-level tracking issue to track items across several github repos._

Languages:
- [ ] Java
- [ ] JS
- [ ] Python
- [ ] Go
- [ ] C# .NET
- [ ] Erlang
- [ ] Ruby

jsuereth:
@andrewhsu would you mind assigning this to me?

Additionally we should expand the description to include a Specification of what compatibility looks like.

dashpole:
I have three changes that should complete the spec-portion of OpenCensus compatibility:

* https://github.com/open-telemetry/opentelemetry-specification/pull/2730: OpenCensus migration plan. This describes how metric and trace bridges should be used to migrate from OpenCensus to OpenTelemetry
* https://github.com/open-telemetry/opentelemetry-specification/pull/2722: Add a "MetricProducer" interface, which can be passed to a MetricReader as a source of metric data.
* https://github.com/open-telemetry/opentelemetry-specification/pull/2732: Add the OpenCensus metric bridge specification, which describes how to implement the MetricProducer interface using metrics from OpenCensus.

dashpole:
I'm working on a golang implementation which follows the proposed spec: https://github.com/open-telemetry/opentelemetry-go/pull/3093

aabmass:
David's PRs were all merged and I think we are looking good here. I am going to open a PR to mark the compatibility doc as stable and close this issue.

aabmass:
From the original description

> * [ ]  C# .NET
> * [ ]  Erlang
> * [ ]  Ruby

I think its safe to say these are out of scope?

PR #2722, opened 2022-08-10
Add MetricProducer to allow sdk.MeterProviders to incorporate metric data from third-party sources

dashpole:

Related to https://github.com/open-telemetry/opentelemetry-specification/pull/2730

Required for https://github.com/open-telemetry/opentelemetry-specification/pull/2732

## Changes

Add a MetricProducer interface to the metrics SDK. It is an optional argument when creating a MetricReader meant to support non-SDK sources of metric data.

Metric data from a MetricProducer is not subject to the MeterProviders or MetricReaders configuration for instruments.  This means Views on MeterProviders, and default Aggregations and Aggregation Temporalities configured on MetricReaders are not applied to data from MetricProducers.

This is needed to support an OpenCensus metric bridge, which is [proposed separately](https://github.com/open-telemetry/opentelemetry-specification/pull/2732).  It could also be used to support other bridges as well, such as a Prometheus bridge.

### Discussion from the spec sig on 8/30:

@dashpole presented high level design options:

1. Bridge incorporated into MeterProvider
1. Bridge separate from MeterProvider, and works with a MetricReader
1. Bridge separate from MeterProvider and MetricReader, works with MetricExporter

3 was eliminated because it won't work easily with pull-based exporters.  The overall sentiment was to stick with option 1, because then resource and exporter configuration for the MeterProvider would be shared.  See https://github.com/open-telemetry/opentelemetry-specification/pull/2722#issuecomment-1231888284 for more details.

### DIFF
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -797,7 +797,7 @@ to (T<sub>n+1</sub>, T<sub>n+2</sub>] - **ONLY** for this particular
 
 The SDK MUST NOT allow a `MetricReader` instance to be registered on more than
 one `MeterProvider` instance. The SDK MUST NOT allow a `MetricReader`
-configured with a `MetricProducer` to be registered with a `MeterProvider`.
+configured with a non-SDK `MetricProducer` to be registered with `MeterProvider`.
 
 ```text
 +-----------------+            +--------------+
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -1100,7 +1100,7 @@ modeled to interact with other components in the SDK:
 
 `MetricProducer` defines the interface which bridges to third-party metric
 sources MUST implement so they can be plugged into an OpenTelemetry
-[MetricReader](#metricreader) as a source of aggregated metric data.
+[MeterProvider](#meterprovider) as a source of aggregated metric data.
 
 ```text
 +-----------------+            +--------------+
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -776,10 +776,11 @@ used with pull-based metrics collection.  A common sub-class of
 `MetricReader`, the periodic exporting `MetricReader` SHOULD be provided
 to be used typically with push-based metrics collection.
 
-The `MetricReader` MUST ensure that data points from OTel instruments are
-output in the configured aggregation temporality for each instrument kind. For
-synchronous instruments being output with Cumulative temporality, this
-means converting [Delta to Cumulative](supplementary-guidelines.md#synchronous-example-cumulative-aggregation-temporality)
+The `MetricReader` MUST ensure that data points from OTel
+[instruments](./api.md#instrument) are output in the configured aggregation
+temporality for each instrument kind. For synchronous instruments being output
+with Cumulative temporality, this means converting
+[Delta to Cumulative](supplementary-guidelines.md#synchronous-example-cumulative-aggregation-temporality)
 aggregation temporality.  For asynchronous instruments being output
 with Delta temporality, this means converting [Cumulative to
 Delta](supplementary-guidelines.md#asynchronous-example-delta-temporality) aggregation
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -86,6 +86,12 @@ Configuration (i.e., [MetricExporters](#metricexporter),
 provide a way to configure all options that are implemented by the SDK. This
 MAY be done at the time of MeterProvider creation if appropriate.
 
+If a meter is created which produces an
+[`InstrumentationScope`](../glossary.md#instrumentation-scope), which matches
+the InstrumentationScope of a [MetricProducer](#metricproducer), or if multiple
+[MetricProducers](#metricproducer) have the same InstrumenatationScope the SDK
+SHOULD emit a warning.
+
 The `MeterProvider` MAY provide methods to update the configuration. If
 configuration is updated (e.g., adding a `MetricReader`), the updated
 configuration MUST also apply to all already returned `Meters` (i.e. it MUST NOT
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -37,6 +37,7 @@ linkTitle: SDK
   * [Exemplar defaults](#exemplar-defaults)
 - [MetricReader](#metricreader)
   * [MetricReader operations](#metricreader-operations)
+    + [Register(MetricProducer)](#registermetricproducer)
     + [Collect](#collect)
     + [Shutdown](#shutdown-1)
   * [Periodic exporting MetricReader](#periodic-exporting-metricreader)
@@ -755,7 +756,9 @@ measurements using the equivalent of the following naive algorithm:
 common configurable aspects of the OpenTelemetry Metrics SDK and
 determines the following capabilities:
 
-* Collecting metrics from the SDK or a [MetricProducer](#metricproducer) on demand.
+* Registering a [MetricProducer](#metricproducer)
+* Collecting metrics from the registered [MetricProducer](#metricproducer) on
+  demand.
 * Handling the [ForceFlush](#forceflush) and [Shutdown](#shutdown) signals from
   the SDK.
 
@@ -763,7 +766,6 @@ To construct a `MetricReader` when setting up an SDK, the caller
 SHOULD provide at least the following:
 
 * The `exporter` to use, which is a `MetricExporter` instance.
-* A [MetricProducer](#metricproducer) (optional) to collect metrics from instead of collecting from the SDK.
 * The default output `aggregation` (optional), a function of instrument kind.  If not configured, the [default aggregation](#default-aggregation) SHOULD be used.
 * The default output `temporality` (optional), a function of instrument kind.  If not configured, the Cumulative temporality SHOULD be used.
 
@@ -797,7 +799,7 @@ to (T<sub>n+1</sub>, T<sub>n+2</sub>] - **ONLY** for this particular
 
 The SDK MUST NOT allow a `MetricReader` instance to be registered on more than
 one `MeterProvider` instance. The SDK MUST NOT allow a `MetricReader`
-configured with a non-SDK `MetricProducer` to be registered with `MeterProvider`.
+registered with a non-SDK `MetricProducer` to be registered with `MeterProvider`.
 
 ```text
 +-----------------+            +--------------+
@@ -820,6 +822,18 @@ idiomatic approach, for example, as `OnForceFlush` and `OnShutdown` callback
 functions.
 
 ### MetricReader operations
+
+#### Register(MetricProducer)
+
+Register causes the MetricReader to use the provided
+[MetricProducer](#metricproducer) for as a source of aggregated metric data in
+subsequent invokations of Collect. It MUST NOT allow more than one
+[MetricProducer](#metricproducer) or [MeterProvider](#meterprovider) instance
+to be registered with the [MetricReader](#metricreader).
+
+If the [MeterProvider](#meterprovider) is an instance of of
+[MetricProducer](#metricproducer), this MAY be used to register the
+MeterProvider.
 
 #### Collect
 
@@ -1060,10 +1074,11 @@ scraper, and `ForceFlush` would not make sense.
 Implementors MAY choose the best idiomatic design for their language. For
 example, they could generalize the [Push Metric Exporter
 interface](#push-metric-exporter) design and use that for consistency, they
-could model the pull exporter as [MetricReader](#metricreader), or they could
-design a completely different pull exporter interface. If the pull exporter is
-modeled as MetricReader, implementors MAY name the MetricExporter interface as
-PushMetricExporter to prevent naming confusion.
+could model the pull exporter as [MetricReader](#metricreader) or
+[MetricProducer](#metricproducer), or they could design a completely different
+pull exporter interface. If the pull exporter is modeled as MetricReader,
+implementors MAY name the MetricExporter interface as PushMetricExporter to
+prevent naming confusion.
 
 The following diagram gives some examples on how `Pull Metric Exporter` can be
 modeled to interact with other components in the SDK:
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -66,7 +66,8 @@ linkTitle: SDK
 
 A `MeterProvider` MUST provide a way to allow a [Resource](../resource/sdk.md) to
 be specified. If a `Resource` is specified, it SHOULD be associated with all the
-metrics produced by any `Meter` from the `MeterProvider`. The [tracing SDK
+metrics produced by any `Meter` from the `MeterProvider` or produced by any
+`MetricProducer`. The [tracing SDK
 specification](../trace/sdk.md#additional-span-interfaces) has provided some
 suggestions regarding how to implement this efficiently.
 
@@ -140,7 +141,9 @@ instances.
 ### View
 
 A `View` provides SDK users with the flexibility to customize the metrics that
-are output by the SDK. Here are some examples when a `View` might be needed:
+are output by the SDK from Measurements made by OpenTelemetry
+[Instruments](./api.md#instrument). Here are some examples when a `View` might
+be needed:
 
 * Customize which [Instruments](./api.md#instrument) are to be
   processed/ignored. For example, an [instrumented
@@ -823,7 +826,8 @@ functions.
 #### Collect
 
 Collects the metrics from the SDK and its [MetricProducers](#metricproducer).
-If there are [asynchronous SDK Instruments](./api.md#asynchronous-instrument-api)
+It retrieves a batch of metric points from MetricProducers by invoking
+`Produce()`. If there are [asynchronous SDK Instruments](./api.md#asynchronous-instrument-api)
 involved, their callback functions will be triggered.
 
 `Collect` SHOULD provide a way to let the caller know whether it succeeded,
@@ -835,6 +839,11 @@ and a failed reasons list to the caller.
 SDK](../overview.md#sdk) authors MAY choose to add parameters (e.g. callback,
 filter, timeout). [OpenTelemetry SDK](../overview.md#sdk) authors MAY choose the
 return value type, or do not return anything.
+
+The `MetricReader` MUST ensure the batch of metric points from
+`MetricProducers` are associated with the `MetricProducer`'s
+`InstrumentationScope()`, and with the `MeterProvider`'s configured resource if
+it is possible for those to conflict.
 
 Note: it is expected that the `MetricReader.Collect` implementations will be
 provided by the SDK, so it is RECOMMENDED to prevent the user from accidentally
@@ -906,8 +915,12 @@ primarily a simple telemetry data encoder and transmitter.
 Metric Exporter has access to the [aggregated metrics
 data](./data-model.md#timeseries-model).  Metric Exporters SHOULD
 report an error condition for data output by the `MetricReader` with
-unsupported Aggregation or Aggregation Temporality, as this condition
-can be corrected by a change of `MetricReader` configuration.
+unsupported Aggregation or Aggregation Temporality. For data originating from
+OpenTelemetry Instruments, this condition can be corrected by a change of
+`MetricReader` configuration. For metric data from a `MetricProducer`, it is an
+indication to the user that metric data is being dropped. The presence of data
+with an unsupported Aggregation or Aggregation Temporality SHOULD NOT prevent
+the exporter from sending other metric data.
 
 There could be multiple [Push Metric Exporters](#push-metric-exporter) or [Pull
 Metric Exporters](#pull-metric-exporter) or even a mixture of both configured at
@@ -1119,10 +1132,12 @@ A `MetricProducer` MUST support the following functions:
 #### Produce() metrics
 
 `Produce` provides metrics from the MetricProducer to the caller. `Produce`
-MUST return a batch of [Metrics](./data-model.md#metric-points), which MUST NOT
-include resource or scope information. `Produce` does not have any required
-parameters, however, [OpenTelemetry SDK](../overview.md#sdk) authors MAY choose
-to add parameters (e.g. timeout).
+MUST return a batch of [Metrics](./data-model.md#metric-points). If the batch
+of Metrics includes an Instrumentation Scope, it MUST be the same as the
+InstrumentationScope returned by InstrumentationScope(). The batch of metrics
+SHOULD NOT include resource. `Produce` does not have any required parameters,
+however, [OpenTelemetry SDK](../overview.md#sdk) authors MAY choose to add
+parameters (e.g. timeout).
 
 `Produce` SHOULD provide a way to let the caller know whether it succeeded,
 failed or timed out. When the `Produce` operation fails, the `MetricProducer`
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -1124,14 +1124,15 @@ A `MetricProducer` MUST support the following functions:
 
 #### Produce() batch
 
+`Produce` provides metrics from the MetricProducer to the caller. `Produce`
+MUST return a batch of [Metric points](./data-model.md#metric-points).
+`Produce` does not have any required parameters, however, [OpenTelemetry
+SDK](../overview.md#sdk) authors MAY choose to add parameters (e.g. timeout).
+
 `Produce` SHOULD provide a way to let the caller know whether it succeeded,
 failed or timed out. When the `Produce` operation fails, the `MetricProducer`
 MAY return successfully collected results and a failed reasons list to the
 caller.
-
-`Produce` does not have any required parameters, however, [OpenTelemetry
-SDK](../overview.md#sdk) authors MAY choose to add parameters (e.g. timeout).
-`Produce` MUST return a batch of [Metric points](./data-model.md#metric-points).
 
 ## Defaults and configuration
 
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -37,7 +37,6 @@ linkTitle: SDK
   * [Exemplar defaults](#exemplar-defaults)
 - [MetricReader](#metricreader)
   * [MetricReader operations](#metricreader-operations)
-    + [Register(MetricProducer)](#registermetricproducer)
     + [Collect](#collect)
     + [Shutdown](#shutdown-1)
   * [Periodic exporting MetricReader](#periodic-exporting-metricreader)
@@ -50,7 +49,8 @@ linkTitle: SDK
   * [Pull Metric Exporter](#pull-metric-exporter)
 - [MetricProducer](#metricproducer)
   * [Interface Definition](#interface-definition-1)
-    + [Produce() batch](#produce-batch)
+    + [Produce() metrics](#produce-metrics)
+    + [InstrumentationScope() InstrumentationScope](#instrumentationscope-instrumentationscope)
 - [Defaults and configuration](#defaults-and-configuration)
 - [Numerical limits handling](#numerical-limits-handling)
 - [Compatibility requirements](#compatibility-requirements)
@@ -80,10 +80,10 @@ an [`InstrumentationScope`](../glossary.md#instrumentation-scope) instance which
 is stored on the created `Meter`.
 
 Configuration (i.e., [MetricExporters](#metricexporter),
-[MetricReaders](#metricreader) and [Views](#view)) MUST be managed solely by the
-`MeterProvider` and the SDK MUST provide a way to configure all options that are
-implemented by the SDK. This MAY be done at the time of MeterProvider creation
-if appropriate.
+[MetricReaders](#metricreader), [MetricProducers](#metricproducer) and
+[Views](#view)) MUST be managed solely by the `MeterProvider` and the SDK MUST
+provide a way to configure all options that are implemented by the SDK. This
+MAY be done at the time of MeterProvider creation if appropriate.
 
 The `MeterProvider` MAY provide methods to update the configuration. If
 configuration is updated (e.g., adding a `MetricReader`), the updated
@@ -756,9 +756,7 @@ measurements using the equivalent of the following naive algorithm:
 common configurable aspects of the OpenTelemetry Metrics SDK and
 determines the following capabilities:
 
-* Registering a [MetricProducer](#metricproducer)
-* Collecting metrics from the registered [MetricProducer](#metricproducer) on
-  demand.
+* Collecting metrics from the SDK on demand.
 * Handling the [ForceFlush](#forceflush) and [Shutdown](#shutdown) signals from
   the SDK.
 
@@ -775,8 +773,8 @@ used with pull-based metrics collection.  A common sub-class of
 `MetricReader`, the periodic exporting `MetricReader` SHOULD be provided
 to be used typically with push-based metrics collection.
 
-The `MetricReader` MUST ensure that data points from the SDK are output in the
-configured aggregation temporality for each instrument kind.  For
+The `MetricReader` MUST ensure that data points from OTel instruments are
+output in the configured aggregation temporality for each instrument kind. For
 synchronous instruments being output with Cumulative temporality, this
 means converting [Delta to Cumulative](supplementary-guidelines.md#synchronous-example-cumulative-aggregation-temporality)
 aggregation temporality.  For asynchronous instruments being output
@@ -784,7 +782,7 @@ with Delta temporality, this means converting [Cumulative to
 Delta](supplementary-guidelines.md#asynchronous-example-delta-temporality) aggregation
 temporality.
 
-The `MetricReader` is not required to ensure data points from a non-SDK
+The `MetricReader` is not required to ensure data points from a
 [MetricProducer](#metricproducer) are output in the configured aggregation
 temporality.
 
@@ -798,8 +796,7 @@ to (T<sub>n+1</sub>, T<sub>n+2</sub>] - **ONLY** for this particular
 `MetricReader` instance.
 
 The SDK MUST NOT allow a `MetricReader` instance to be registered on more than
-one `MeterProvider` instance. The SDK MUST NOT allow a `MetricReader`
-registered with a non-SDK `MetricProducer` to be registered with `MeterProvider`.
+one `MeterProvider` instance.
 
 ```text
 +-----------------+            +--------------+
@@ -823,22 +820,10 @@ functions.
 
 ### MetricReader operations
 
-#### Register(MetricProducer)
-
-Register causes the MetricReader to use the provided
-[MetricProducer](#metricproducer) for as a source of aggregated metric data in
-subsequent invocations of Collect. It MUST NOT allow more than one
-[MetricProducer](#metricproducer) or [MeterProvider](#meterprovider) instance
-to be registered with the [MetricReader](#metricreader).
-
-If the [MeterProvider](#meterprovider) is an instance of a
-[MetricProducer](#metricproducer), this MAY be used to register the
-MeterProvider.
-
 #### Collect
 
-Collects the metrics from the SDK or the [MetricProducer](#metricproducer). If
-there are [asynchronous SDK Instruments](./api.md#asynchronous-instrument-api)
+Collects the metrics from the SDK and its [MetricProducers](#metricproducer).
+If there are [asynchronous SDK Instruments](./api.md#asynchronous-instrument-api)
 involved, their callback functions will be triggered.
 
 `Collect` SHOULD provide a way to let the caller know whether it succeeded,
@@ -1074,11 +1059,10 @@ scraper, and `ForceFlush` would not make sense.
 Implementors MAY choose the best idiomatic design for their language. For
 example, they could generalize the [Push Metric Exporter
 interface](#push-metric-exporter) design and use that for consistency, they
-could model the pull exporter as [MetricReader](#metricreader) or
-[MetricProducer](#metricproducer), or they could design a completely different
-pull exporter interface. If the pull exporter is modeled as MetricReader,
-implementors MAY name the MetricExporter interface as PushMetricExporter to
-prevent naming confusion.
+could model the pull exporter as [MetricReader](#metricreader), or they could
+design a completely different pull exporter interface. If the pull exporter is
+modeled as MetricReader, implementors MAY name the MetricExporter interface as
+PushMetricExporter to prevent naming confusion.
 
 The following diagram gives some examples on how `Pull Metric Exporter` can be
 modeled to interact with other components in the SDK:
@@ -1116,38 +1100,41 @@ modeled to interact with other components in the SDK:
 
 `MetricProducer` defines the interface which bridges to third-party metric
 sources MUST implement so they can be plugged into an OpenTelemetry
-[MetricReader](#metricreader) as a source of aggregated metric data. The SDK's
-in-memory state MAY implement the `MetricProducer` interface for convenience.
+[MetricReader](#metricreader) as a source of aggregated metric data.
 
 ```text
 +-----------------+            +--------------+
 |                 | Metrics... |              |
-| In-memory state +------------> MetricReader |
+| MeterProvider   +------------> MetricReader |
 |                 |            |              |
-+-----------------+            +--------------+
-
-+-----------------+            +--------------+
-|                 | Metrics... |              |
-| MetricProducer  +------------> MetricReader |
-|                 |            |              |
-+-----------------+            +--------------+
++---------------^^+            +--------------+
+|  MetricProducer |
++-----------------+
 ```
 
 ### Interface Definition
 
 A `MetricProducer` MUST support the following functions:
 
-#### Produce() batch
+#### Produce() metrics
 
 `Produce` provides metrics from the MetricProducer to the caller. `Produce`
-MUST return a batch of [Metric points](./data-model.md#metric-points).
-`Produce` does not have any required parameters, however, [OpenTelemetry
-SDK](../overview.md#sdk) authors MAY choose to add parameters (e.g. timeout).
+MUST return a batch of [Metrics](./data-model.md#metric-points), which MUST NOT
+include resource or scope information. `Produce` does not have any required
+parameters, however, [OpenTelemetry SDK](../overview.md#sdk) authors MAY choose
+to add parameters (e.g. timeout).
 
 `Produce` SHOULD provide a way to let the caller know whether it succeeded,
 failed or timed out. When the `Produce` operation fails, the `MetricProducer`
 MAY return successfully collected results and a failed reasons list to the
 caller.
+
+#### InstrumentationScope() InstrumentationScope
+
+`InstrumentationScope` provides the
+[InstrumentationScope](../glossary.md#instrumentation-scope) associated with
+the MetricProducer. A MetricProducer instance MUST return the same
+instrumentation scope for each invocation of InstrumentationScope().
 
 ## Defaults and configuration
 
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -47,6 +47,9 @@ linkTitle: SDK
       - [ForceFlush()](#forceflush)
       - [Shutdown()](#shutdown)
   * [Pull Metric Exporter](#pull-metric-exporter)
+- [MetricProducer](#metricproducer)
+  * [Interface Definition](#interface-definition-1)
+    + [Produce() batch](#produce-batch)
 - [Defaults and configuration](#defaults-and-configuration)
 - [Numerical limits handling](#numerical-limits-handling)
 - [Compatibility requirements](#compatibility-requirements)
@@ -752,7 +755,7 @@ measurements using the equivalent of the following naive algorithm:
 common configurable aspects of the OpenTelemetry Metrics SDK and
 determines the following capabilities:
 
-* Collecting metrics from the SDK on demand.
+* Collecting metrics from the SDK or a [MetricProducer](#metricproducer) on demand.
 * Handling the [ForceFlush](#forceflush) and [Shutdown](#shutdown) signals from
   the SDK.
 
@@ -760,6 +763,7 @@ To construct a `MetricReader` when setting up an SDK, the caller
 SHOULD provide at least the following:
 
 * The `exporter` to use, which is a `MetricExporter` instance.
+* A [MetricProducer](#metricproducer) (optional) to collect metrics from instead of collecting from the SDK.
 * The default output `aggregation` (optional), a function of instrument kind.  If not configured, the [default aggregation](#default-aggregation) SHOULD be used.
 * The default output `temporality` (optional), a function of instrument kind.  If not configured, the Cumulative temporality SHOULD be used.
 
@@ -769,13 +773,17 @@ used with pull-based metrics collection.  A common sub-class of
 `MetricReader`, the periodic exporting `MetricReader` SHOULD be provided
 to be used typically with push-based metrics collection.
 
-The `MetricReader` MUST ensure that data points are output in the
+The `MetricReader` MUST ensure that data points from the SDK are output in the
 configured aggregation temporality for each instrument kind.  For
 synchronous instruments being output with Cumulative temporality, this
 means converting [Delta to Cumulative](supplementary-guidelines.md#synchronous-example-cumulative-aggregation-temporality)
 aggregation temporality.  For asynchronous instruments being output
 with Delta temporality, this means converting [Cumulative to
 Delta](supplementary-guidelines.md#asynchronous-example-delta-temporality) aggregation
+temporality.
+
+The `MetricReader` is not required to ensure data points from a non-SDK
+[MetricProducer](#metricproducer) are output in the configured aggregation
 temporality.
 
 The SDK MUST support multiple `MetricReader` instances to be registered on the
@@ -788,7 +796,8 @@ to (T<sub>n+1</sub>, T<sub>n+2</sub>] - **ONLY** for this particular
 `MetricReader` instance.
 
 The SDK MUST NOT allow a `MetricReader` instance to be registered on more than
-one `MeterProvider` instance.
+one `MeterProvider` instance. The SDK MUST NOT allow a `MetricReader`
+configured with a `MetricProducer` to be registered with a `MeterProvider`.
 
 ```text
 +-----------------+            +--------------+
@@ -814,9 +823,9 @@ functions.
 
 #### Collect
 
-Collects the metrics from the SDK. If there are [asynchronous
-Instruments](./api.md#asynchronous-instrument-api) involved, their callback
-functions will be triggered.
+Collects the metrics from the SDK or the [MetricProducer](#metricproducer). If
+there are [asynchronous SDK Instruments](./api.md#asynchronous-instrument-api)
+involved, their callback functions will be triggered.
 
 `Collect` SHOULD provide a way to let the caller know whether it succeeded,
 failed or timed out. When the `Collect` operation fails or times out on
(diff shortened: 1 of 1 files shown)


specification/metrics/sdk.md
```diff
@@ -1085,6 +1093,44 @@ modeled to interact with other components in the SDK:
...
+`MetricProducer` defines the interface which bridges to third-party metric
+sources MUST implement so they can be plugged into an OpenTelemetry
+[MetricReader](#metricreader) as a source of aggregated metric data. The SDK's
+in-memory state MAY implement the `MetricProducer` interface for convenience.
+
+  ```text
+  +-----------------+ Metrics... +-----------------------------+
+  | In-memory state ------------>                              |
+  +-----------------+            | Exporting MetricReader      |
+  | MetricProducer A------------>                              |
```
dashpole, line 1109: Note: Metric readers should only have one producer.

dashpole: fixed

specification/metrics/sdk.md
```diff
@@ -1085,6 +1094,44 @@ modeled to interact with other components in the SDK:
...
   ```
 
+## MetricProducer
+
+
+`MetricProducer` defines the interface which bridges to third-party metric
+sources MUST implement so they can be plugged into an OpenTelemetry
+[MetricReader](#metricreader) as a source of aggregated metric data. The SDK's
+in-memory state MAY implement the `MetricProducer` interface for convenience.
```
pirgeo, line 1104: Above it says that a MetricProducer must not be allowed to be registered with a MeterProvider, but I think if the in-memory state is modeled as MetricProducer that would happen, right?

specification/metrics/sdk.md
```diff
@@ -1085,6 +1094,44 @@ modeled to interact with other components in the SDK:
...
+| MetricProducer  +------------> MetricReader |
+|                 |            |              |
++-----------------+            +--------------+
+```
+
+### Interface Definition
+
+A `MetricProducer` MUST support the following functions:
+
+#### Produce() batch
```
pirgeo, line 1124: Could you add a short sentence about what this method does?

specification/metrics/sdk.md
```diff
@@ -1085,6 +1094,44 @@ modeled to interact with other components in the SDK:
                                  +-----------------------------+
   ```
 
+## MetricProducer
```
pirgeo, line 1097: Some info about aggregation temporalities would be nice. I assume it depends on the aggregation provided by whatever source data is coming from, but will there be a way to set the preferred output temporality (potentially including conversion (e.g. cumulative to delta))? Some SDKs already have the conversion mechanisms built in, so I assume that should be possible.

pirgeo (commented):
It seems to me that OpenCensus is currently the only use case (but please correct me if I am wrong). I am not familiar with OpenCensus, but I know from Micrometer that there are definitely Metrics APIs that allow plugging a different SDK. I wonder if it is possible to wrap the OpenTelemetry instruments in the OpenCensus instruments so that their usage doesn't change, but they would use the OTel SDK nonetheless.

specification/metrics/sdk.md
dashpole: Only MetricReaders, not MetricProducers, are registered with a MeterProvider.  I assume thats what you meant to say :)

I clarified the above to say: "The SDK MUST NOT allow a `MetricReader`
configured with a **non-SDK** `MetricProducer` to be registered with `MeterProvider`."  The intention is for a MetricReader to have a _single_ source of metric data: either a non-SDK MetricProducer OR a MeterProvider.

specification/metrics/sdk.md
dashpole: Added: "`Produce` provides metrics from the MetricProducer to the caller."

specification/metrics/sdk.md
pirgeo: For clarification: this scenario is fine:
* SDK implements the in-memory state as `MetricProducer` 
* MetricReader is configured with that in-memory state (it accepts a MetricProducer as its source)
* MetricReader is registered with a MeterProvider
* MeterProvider provides meters (obviously 😅) that write to the in-memory state, which is in turn exported by the MetricReader

This scenario is **not** fine:
* Create a MetricProducer (e.g. for OpenMetrics bridge)
* MetricReader is configured with OpenMetrics MetricsProducer
* MetricReader is also registered with a MeterProvider

Registering a MetricReader with the MetricsProducer to a MeterProvider does not make sense, since then there are two "sources" of data for one Reader.

Did I get that right?

specification/metrics/sdk.md
```diff
@@ -751,14 +754,15 @@ measurements using the equivalent of the following naive algorithm:
...
-* Collecting metrics from the SDK on demand.
+* Collecting metrics from the SDK or a [MetricProducer](#metricproducer) on demand.
 * Handling the [ForceFlush](#forceflush) and [Shutdown](#shutdown) signals from
   the SDK.
 
 To construct a `MetricReader` when setting up an SDK, the caller
 SHOULD provide at least the following:
 
 * The `exporter` to use, which is a `MetricExporter` instance.
+* A [MetricProducer](#metricproducer) (optional) to collect metrics from instead of collecting from the SDK.
```
MrAlias, line 765: This adds a bidirectional registration of the `Reader`: a `Reader` is registered with a `MeterProvider` during its creation, but a `MetricProducer` is registered with a `Reader` during its. It makes sense based on the reasoning of how information will be sourced, but does not match the control flow direction (`ForceFlush` and `Shutdown`) of the current `MeterProvider` setup. I recommend we try to stay consistent with the registration direction of the `MeterProvider`.

Currently, in Go, we have a [`register`](https://github.com/open-telemetry/opentelemetry-go/blob/3013ed18f87a1c2fc29400312136668fd6fd28f1/sdk/metric/reader.go#L56-L59) method on a `Reader` (I think this is the layout in a few other languages as well). Can we formalize this reader method here? Specifically it would be a method that accepted a `MetricProducer` and only allows one registration call.

This additional method would allow other `MetricProducer` implementations, like the OC bridge, to accept a `Reader` similar to a `MeterProvider`. Because of this direction of registration, the OC bridge like types could then also call the `ForceFlush` and `Shutdown` methods and act similar to a `MeterProvider`.

This would also remove potential confusion of an end-user: it may not be obvious to them on setup that the optional `MetricProducer` doesn't need to be passed to a `Reader` if that `Reader` is going to be passed to a `MeterProvider`. However, it will add potential confusion if a user wonders if they should be calling this method of the `Reader`.

I don't see a way to avoid the latter confusion without reverting the registration mechanism of the existing `MeterProvider` and unifying on the currently proposed approach. Something we cannot do with our stability guarantees unfortunately.

specification/metrics/sdk.md
dashpole: Yes, that is correct.  One thing to point out is that these two steps are actually doing the same thing:

> * MetricReader is configured with that in-memory state (it accepts a MetricProducer as its source)
> * MetricReader is registered with a MeterProvider

That is because registering a MetricReader with a MeterProvider makes the SDK the source of metrics for the MetricReader.  The spec doesn't appear to forbid registering a MetricReader multiple times with the same MeterProvider (so the scenario is allowed).  But for the SDK, the following is still the normal way to configure things:

* MetricReader is registered with a MeterProvider
* MeterProvider provides meters (obviously 😅) that write to the in-memory state, which is in turn exported by the MetricReader

specification/metrics/sdk.md
dashpole: This definitely needs more thought.  Right now, the phrasing of the spec is that it is the MetricReader's responsibility to ensure the aggregation temporality for instruments is respected.  One issue with this, though, is that it is a function of instrument kind, which is an OTel API concept, rather than a data-model/SDK concept.  That is a problem given the current design because a MetricProducer provides the MetricReader with data that is **already aggregated**, and doesn't preserve which instrument was used to collect the data.  I think the options we have are:

1. Bridges have to incorporate the MetricReader's `func temporality(instrument kind)` into their output by defining a mapping from their instruments to OTel's instrument kinds, and reimplementing cumulative-to-delta or delta-to-cumulative.
2. Require only the SDK to respect the specified output temporality.  This is what i've currently specified (in the MetricReader section).

specification/metrics/sdk.md
dashpole: Done.

dashpole:
> I wonder if it is possible to wrap the OpenTelemetry instruments in the OpenCensus instruments so that their usage doesn't change, but they would use the OTel SDK nonetheless.

Unfortunately, I don't think that is feasible for OpenCensus.  The [instrumentation API](https://opencensus.io/stats/measure/) for OpenCensus is quite different from OTel's.  You just call `Record(number)`, and then decide what instrument to use in the OpenCensus View. The other challenge is that you can't substitute the SDK in OpenCensus with a different implementation (e.g. OTel).

Although I haven't prototyped it out, I believe this model (bridge implements MetricProducer) would also work well for a prometheus bridge.  In Go, I can [Gather](https://github.com/prometheus/client_golang/blob/c7aa2a5b843527449adb99ad113fe14ed15e4eb0/prometheus/registry.go#L409) metrics from a prometheus registry (~SDK), and could convert to an OpenTelemetry batch of metrics relatively easily.

specification/metrics/sdk.md
MrAlias: Missing push?

jmacd:
As I read this PR, a MetricProducer could be bound to just one MetricExporter. Does the MetricProducer user lose access to multi-exporter support?

Thinking about how I'd like to see a bridge to OpenCensus (or Prometheus), I'd like to see a single SDK (with multiple Meters) co-exist with MetricProducers so that a single SDK configuration could include bridged-metrics AND there would be just one export call across the whole SDK+bridges, one gRPC connection, etc.

To me, this suggests integrating a different sort of Producer than the one specified here, which essentially looks like a replacement of the SDK. Instead, can we replace a Meter with a producer for a single Scope? In addition to bridging metrics from OpenCensus and Prometheus, this sort of back-door allows users to work around all the current limitations in OpenTelemetry API. For example, asynchronous histogram does not exist? Just create a MetricProducer and emit some histogram points (e.g., I might use a single-scope MetricProducer to output https://pkg.go.dev/runtime/metrics ([issue](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/2624)).

specification/metrics/sdk.md
dashpole: yep, pushed

dashpole:
> As I read this PR, a MetricProducer could be bound to just one MetricExporter. Does the MetricProducer user lose access to multi-exporter support?

That wasn't my intention.  A MetricReader can have a single MetricProducer, and a single MetricExporter.  But, like a MeterProvider can have multiple MetricReaders, a MetricProducer can be registered with many MetricReaders, each with an exporter.  Let me know if I can make that clearer.

I'll chew on the other feedback and see what I can come up with.

specification/metrics/sdk.md
pirgeo: Having a way of converting directly on the client side would be very nice in my opinion. I feel like one of the aims of the OpenTelemetry project is to open upgrade paths (e.g. coming from OpenCensus but wanting to try out different monitoring solutions). By allowing users to convert their existing instrumentation to whatever OTLP is capable of, the bar for trying out something new is lower.

specification/metrics/sdk.md
```diff
@@ -811,11 +822,23 @@ functions.
 
 ### MetricReader operations
 
+#### Register(MetricProducer)
+
+Register causes the MetricReader to use the provided
+[MetricProducer](#metricproducer) for as a source of aggregated metric data in
+subsequent invokations of Collect. It MUST NOT allow more than one
```
MrAlias, line 829:
```suggestion
subsequent invocations of Collect. It MUST NOT allow more than one
```

specification/metrics/sdk.md
```diff
@@ -811,11 +822,23 @@ functions.
...
+[MetricProducer](#metricproducer) or [MeterProvider](#meterprovider) instance
+to be registered with the [MetricReader](#metricreader).
+
+If the [MeterProvider](#meterprovider) is an instance of of
```
MrAlias, line 833:
```suggestion
If the [MeterProvider](#meterprovider) is an instance of a
```

MrAlias (approved):

dashpole:
@jmacd re: https://github.com/open-telemetry/opentelemetry-specification/pull/2722#issuecomment-1218591597

I think it should be possible to achieve your first ask.  Instead of registering a MetricProducer with a MetricReader, we could pass a MetricProducer to a MeterProvider.  The MeterProvider would add the bridge's metrics to metrics from OTel instruments' metrics, and provide a single batch of metrics to exporters.  This would have the added benefit of simplifying the user experience, since they don't need to create multiple readers (one for the SDK, one for the bridge).

For the second section, I think the fundamental ask is to reduce the scope of the interface from a full batch of metrics (i.e. one ResourceMetrics in Go) to metrics for a single scope (i.e. []Metrics in Go).  Doing that would also simplify the user experience, since the Resource used by the MeterProvider will be automatically used for bridged metrics, rather than needing to be supplied separately.

I've made those updates in the latest commit.

dashpole:
A PoC of this in Go: https://github.com/open-telemetry/opentelemetry-go/pull/3100.

specification/metrics/sdk.md
```diff
@@ -1085,6 +1093,48 @@ modeled to interact with other components in the SDK:
...
+
+### Interface Definition
+
+A `MetricProducer` MUST support the following functions:
+
+#### Produce() metrics
+
+`Produce` provides metrics from the MetricProducer to the caller. `Produce`
+MUST return a batch of [Metrics](./data-model.md#metric-points), which MUST NOT
+include resource or scope information. `Produce` does not have any required
```
jack-berg, line 1122: > which MUST NOT include resource or scope information

This is challenging for java, and I suspect other implementations. We define a `MetricData` interface which represents the in memory representation of the otel data model. `MetricData` includes accessors for [resource and scope information](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/MetricData.java#L34-L42). 

I get why the metric producer shouldn't include the resource, since that's an SDK concern, but the scope seems like a concern of the metric producer.

specification/metrics/sdk.md
```diff
@@ -1085,6 +1093,48 @@ modeled to interact with other components in the SDK:
                                  +-----------------------------+
   ```
 
+## MetricProducer
+
+
+`MetricProducer` defines the interface which bridges to third-party metric
+sources MUST implement so they can be plugged into an OpenTelemetry
+[MetricReader](#metricreader) as a source of aggregated metric data.
```
jack-berg, line 1102: > so they can be plugged into an OpenTelemetry [MetricReader](#metricreader)

I don't see any description on how metric producers are wired into / registered with metric readers / meter providers.

specification/metrics/sdk.md
jack-berg: MetricProducers will be strange in general since in general they'll produce metrics that conflict with meter provider configuration. Examples include exemplars, all the things configurable with views, and temporality. 

If we really do want this MetricProducer concept, perhaps it's better to just say that they operate outside SDK in general. Trying to do anything else will require quite complex reinventing of mechanisms that already exist in SDK internals. And maybe its worth somehow indicating to metric readers / metric exporters that the data doesn't originate from opentelemetry instruments to help debug and avoid confusion.

specification/metrics/sdk.md
dashpole: Sorry, this was meant to say MeterProvider instead of MetricReader.  I've updated it.

It is listed as one of pieces of SDK config the in the "Meter Creation" section:

> Configuration (i.e., MetricExporters, MetricReaders, **MetricProducers** and Views) MUST be managed solely by the `MeterProvider` and the SDK MUST provide a way to configure all options that are implemented by the SDK. This MAY be done at the time of MeterProvider creation if appropriate.

specification/metrics/sdk.md
dashpole: Got it.  I thought it would be simpler to provide scope information as a separate function, but I can switch it to be included in Produce() instead.

It seems like excluding resource would still be an issue for java, though, right?

specification/metrics/sdk.md
dashpole: The motivation for moving the interface into the SDK is https://github.com/open-telemetry/opentelemetry-specification/pull/2722#issuecomment-1218591597.  What do you mean by metrics conflicting?  They will be under a different InstrumentationScope, which should prevent conflicts.  We could probably use a scope attribute to indicate that the metrics came from a MetricProducer if we wanted.

jack-berg: Conflict is probably the wrong word. The metrics will disagree / disobey the configuration of the meter provider. Some examples besides temporality:
- where views can do things like change the name, description, aggregation, and attributes of instruments. It would be pretty challenging to have these transformations applied to metrics provided by a 3rd party producer.
- meter providers have (still experimental for now) the ability to configure the exemplar filter. Does that same concept make sense for exemplars of metrics from 3rd party producers?

specification/metrics/sdk.md
jack-berg: > It seems like excluding resource would still be an issue for java, though, right?

Yes. Perhaps there can be some mechanism for the producer to retrieve the resource from the meter provider? Or perhaps 3rd party producers should just provide an empty resource with the expectation it is overwritten by the meter provider.

specification/metrics/sdk.md
dashpole: I agree it would be very difficult to make metrics from MetricProducers obey MeterProvider configuration, and I can see how it would be confusing for a MeterProvider configured with a MetricProducer to produce some metrics which don't abide by the configuration of the MeterProvider. I think our options are:

1. (original proposal) MetricProducer and MeterProvider are completely separate.  A MetricReader gets aggregated metrics from one of the two. Downsides:
    * Users need to set up extra MetricReaders (one for MetricProducer per-exporter, one for MeterProvider per-exporter).
    * Metrics from MetricProducer and MeterProvider are sent in separate batches to exporters.
    * MetricProducers need to separately be configured with a resource.
1. (current proposal) MeterProvider takes MetricProducers as optional config. Metrics from MetricProducers are appended to metrics produced by OTel instrumentation during collection. Other MeterProvider config (aggregation, view, etc) only applies to metrics from OTel instrumentation, and not to metrics from MetricProducers. Downsides:
    * Confusing that configuration doesn't apply to metrics from MetricProducers
1. (new idea) MetricProducer and MeterProvider are completely separate.  A MetricReader gets aggregated metrics from **multiple** MetricProducers or MeterProviders.  MetricReader combines sources of metrics if the resource is shared (which should generally be the case). Downsides:
    * Additional complexity for MetricReaders to re-aggregate metrics.
    * MetricProducers need to separately be configured with a resource.

cc @jmacd

jack-berg: I prefer option 3, which necessitates formalizing the contract for registering metric producers with metric readers. 

> Additional complexity for MetricReaders to re-aggregate metrics.

Not necessarily. We can say that different metric producers registered with the same metric reader are expected to have zero overlap of scope metrics. If conflicts do occur, then we're in an unspecified error state.

specification/metrics/sdk.md
```diff
@@ -1085,6 +1092,42 @@ modeled to interact with other components in the SDK:
...
+
+### Interface Definition
+
+A `MetricProducer` MUST support the following functions:
+
+#### Produce() metrics
+
+`Produce` provides metrics from the MetricProducer to the caller. `Produce`
+MUST return a batch of [Metrics](./data-model.md#metric-points) with an
+[InstrumentationScope](../glossary.md#instrumentation-scope). `Produce` MUST
```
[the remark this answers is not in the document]
dashpole: The answer to this will depend on the outcome of https://github.com/open-telemetry/opentelemetry-specification/pull/2722#discussion_r950508783 above.

specification/metrics/sdk.md
jsuereth: SO the concern here is that not ALL configuration of a MeterProvider would apply to the additional provider of metrics?

This new API is actually meant to bridge existing Metric APIs (particularly OpenCensus).   In this case, we need to bridge the API *not* at the raw metric event layer, because it's incompatible. Specifically, the notion that we *could* share views, aggregation + exemplars in this instance is impossible because we cannot instrument the raw events in a compatible way.

However, from a user perspective, they'd like to use the same *destination* configuration as otel.  This means MetricReader/MetricExporter configuration.

Given what I'm reading here, I feel like the *new* proposals are complicated the metric system even further than what's in the PR.   Specifically we expect this type of bridge to be used as a temporary workaround to account for metric systems that are in existence until such a time as they migrate to something more OpenTelemetry friendly.  E.g. dropwizard metrics, OpenCensus metrics and *possibly* prometheus client libraries (due to exemplar API incompatibility).

I'd like to suggest we just do the following:

1. Allow MetricProducer to shared MetricReader/MetricExporter configuration w/ MeterProvider in some fashion (ideally no new config env variables or awkward syntax).
2. Prevent unusable metrics from reaching MetricReader/MetricExporter via this hook.  E.g. a  MetricExporter that can't handle cumulative metrics would just not receive cumulatives (and possible log/record an error for user diagnosis).  This helps account for the possible fragility of metric reader/exporters metric handling in OTEL given current config.
3. Denote that MetricProducer is an advanced use case and add caveats all around its definition on limitations of config applicability.


TL;DR: I think the experience of using a MetricReader bridge should be *similar* to configuring two different metric systems to expose prometheus endpoints and then creating a new endpoint that returns the union of metrics from both.  However, for OTEL it would be far more flexible and useful for users.

jsuereth (approved):

specification/metrics/sdk.md
```diff
@@ -768,15 +771,19 @@ used with pull-based metrics collection.  A common sub-class of
...
+The `MetricReader` MUST ensure that data points from OTel instruments are
+output in the configured aggregation temporality for each instrument kind. For
 synchronous instruments being output with Cumulative temporality, this
 means converting [Delta to Cumulative](supplementary-guidelines.md#synchronous-example-cumulative-aggregation-temporality)
 aggregation temporality.  For asynchronous instruments being output
 with Delta temporality, this means converting [Cumulative to
 Delta](supplementary-guidelines.md#asynchronous-example-delta-temporality) aggregation
 temporality.
 
+The `MetricReader` is not required to ensure data points from a
```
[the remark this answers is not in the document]
dashpole: The problem is that aggregation temporality is configured as a function of [instrument](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument) kind, which is an OTel API construct, rather than as a function of [point kind](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md#point-kinds).  So I don't think we actually know what to do with a `sum` datapoint, as we don't know the instrument that recorded it.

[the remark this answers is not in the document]
jack-berg: Good point @dashpole - instrument kind is erased by the time we make it to the data model.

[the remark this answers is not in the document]
jack-berg: It feels more appropriate to this qualification in the bulleted list up on line 766. 

Also, similar to aggregation temporality, the metric reader should be exempt from ensuring that data points follow the default aggregation. Although this might be necessary to specify, since the the default aggregation is not a guarantee.

specification/metrics/sdk.md
pirgeo: So the user that already has an OpenCensus setup would:
* Set up the OTel MetricReader/MetricExporter/MeterProvider to point to their backend, including temporality preference
* Set up the MetricProducer to bridge the OpenCensus metrics to OpenTelemetry
* Plug the MetricProducer into the MeterProvider
* Run the app
  * Metrics that are compatible with the OTel exporter setup will be sent via the OTel exporter
  * Metrics that are incompatible will lead to warnings and will be dropped

Is that right?

If it is, we might end up in a state where some of the metrics are bridged and exported, and some are not. Are we okay with this? It might be a good starting point for migrating metrics that aren't working to another API (e.g, using OTel). Alternatively, sending metrics in parallel from both systems (i.e., OpenCensus directly and via the bridge) could lead to the same metric being reported multiple times.

specification/metrics/sdk.md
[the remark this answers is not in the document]
jsuereth: I understand it's not 1:1 and instrument kind doens't exist here.  As jack calls out, we need to call out SOMEWHERE one of two things:

1. MetricExporters/MetricReaders need to handle all types of points due to this bridge.
2. SDK behavior for Readers/Exporters that do not want to handle all possible aggregations coming from OpenCensus.

dashpole:
Notes from the discussion at the Spec sig today:

I presented high level design options:
    1. Bridge incorporated into MeterProvider
    1. Bridge separate from MeterProvider, and works with a MetricReader
    1. Bridge separate from MeterProvider and MetricReader, works with MetricExporter

3 was eliminated because it won't work easily with pull-based exporters.  The overall sentiment was to stick with option 1.  But we need to call out a few additional things:

* View configuration for the MeterProvider MUST NOT be applied to metric data provided by bridges.
* Resource MUST be provided by the MeterProvider, and any resource information provided by the bridge MUST be overridden.
* Default aggregation and default temporality for MetricReaders MUST NOT be applied to metrics from bridges.
* MetricExporters MUST handle (e.g. by dropping) metrics which don't match their supported aggregations or temporalities.
 
There was an idea that metric readers _could_ take the metric data from bridges and apply the aggregation and temporality for the instrument kind it _likely_ came from (if I understood it correctly).  But we decided that the idea needed prototyping to show feasibility.

specification/metrics/sdk.md
dashpole: Added:

If the batch of Metrics includes an Instrumentation Scope, it MUST be the same as the InstrumentationScope returned by InstrumentationScope(). The batch of metrics SHOULD NOT include resource. 

And:

The `MetricReader` MUST ensure the batch of metric points from
`MetricProducers` are associated with the `MetricProducer`'s
`InstrumentationScope()`, and with the `MeterProvider`'s configured resource if
it is possible for those to conflict.

specification/metrics/sdk.md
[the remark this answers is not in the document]
dashpole: I've opted for option 1:

Metric Exporters SHOULD
report an error condition for data output by the `MetricReader` with
unsupported Aggregation or Aggregation Temporality. For data originating from
OpenTelemetry Instruments, this condition can be corrected by a change of
`MetricReader` configuration. For metric data from a `MetricProducer`, it is an
indication to the user that metric data is being dropped. The presence of data
with an unsupported Aggregation or Aggregation Temporality SHOULD NOT prevent
the exporter from sending other metric data.

jack-berg:
Thanks for that summary @dashpole. One comment on this point:

> MetricExporters MUST handle (e.g. by dropping) metrics which don't match their supported aggregations or temporalities.

I don't think this line is necessary. The aggregation selection of a reader / exporter is just the default exporter - it can always be overridden via views in which case exporters are expected to handle it. And I don't believe its possible to implement dropping temporalities that don't match: Temporality is selected by instrument type. I may choose delta temporality for async counters, and cumulative for sync counters. When an exporter sees a cumulative monotonic sum, it knows it came from a counter, but doesn't know if it came from an async counter or a sync counter. The type of original instrument matters because the exporter may request different temporalities. 

Agree with option 1 with the other additions you've described 👍

dashpole:
@jack-berg I think we are on the same page, and I agree it isn't much of a change from the status quo. PTAL at the changes on line 918.  The spec currently indicates that changing the default aggregation temporality per-instrument can prevent unsupported temporalities from reaching exporters. This just clarifies that it won't work for data from MetricProducers, and recommends gracefully handling data unsupported by the exporter.

To clarify, I meant that exporters should drop unsupported temporality **data** (e.g. if my backend can only accept delta sums, and it gets a cumulative sum), not that anything should be accepted/rejected based on instrument.

specification/metrics/sdk.md
```diff
@@ -898,8 +915,12 @@ primarily a simple telemetry data encoder and transmitter.
 Metric Exporter has access to the [aggregated metrics
 data](./data-model.md#timeseries-model).  Metric Exporters SHOULD
 report an error condition for data output by the `MetricReader` with
-unsupported Aggregation or Aggregation Temporality, as this condition
-can be corrected by a change of `MetricReader` configuration.
+unsupported Aggregation or Aggregation Temporality. For data originating from
+OpenTelemetry Instruments, this condition can be corrected by a change of
+`MetricReader` configuration. For metric data from a `MetricProducer`, it is an
+indication to the user that metric data is being dropped. The presence of data
```
pirgeo, line 921: So, from a user perspective:
* I have an existing OpenCensus instrumentation
* I add the OpenTelemetry bridge and configure my Exporter to send to my backend
* I look at the data in my backend
  * Some of the data shows up
  * Some of the data does not show up, because my exporter only supports certain Aggregation/AggregationTemporality pairings

At this point, I have to decide what to do with the metrics that didn't show up in my backend. Do I have to migrate them to the OpenTelemetry API right away? Do I use another API to export them? Do I run the old and new instrumentation in parallel? 

What I am trying to say is: It seems very strange to me that a bridge might drop half or even all of my data. I would expect all of the data to _go somewhere_. 

I am not trying to block this work, I am just trying to understand it, so: Am I missing something?

specification/metrics/sdk.md
```diff
@@ -769,15 +776,19 @@ used with pull-based metrics collection.  A common sub-class of
...
+The `MetricReader` MUST ensure that data points from OTel instruments are
+output in the configured aggregation temporality for each instrument kind. For
 synchronous instruments being output with Cumulative temporality, this
 means converting [Delta to Cumulative](supplementary-guidelines.md#synchronous-example-cumulative-aggregation-temporality)
 aggregation temporality.  For asynchronous instruments being output
 with Delta temporality, this means converting [Cumulative to
 Delta](supplementary-guidelines.md#asynchronous-example-delta-temporality) aggregation
 temporality.
 
+The `MetricReader` is not required to ensure data points from a
```
pirgeo, line 788: I am playing devil's advocate here, but this feels like a potential loophole in the spec. "MetricReader must ensure" in the previous paragraph, but in the next paragraph it says "if you just implement MetricProducer you are free to do whatever you want". It kinda opens the door to everyone who builds a bridge to just push whatever they want to OTel and then let someone else deal with the consequences. I guess what I am suggesting is to put the burden of making sure the data is compatible on the person building the bridge implementation because they are probably familiar with both the "source" and the "target".

specification/metrics/sdk.md
dashpole: If you are already using OpenCensus, you are presumably starting with an exporter which is compatible with OC's Aggregation/AggregationTemporality pairings.  The goal is to enable such cases to switch from OC + OC exporter to OTel + equivalent exporter.  That enables them to migrate instrumentation over time (as described in the [migration plan](https://github.com/open-telemetry/opentelemetry-specification/pull/2730/files#diff-af102fb60abe17183934e49ad7cf16f8fc621b397d5e8bcd0bda6086320d8054R36)).  Then they can switch exporters, or make use of other OTel-specific features.

We should still decide what happens when someone tries to use the bridge with an exporter that isn't compatible with OC's datamodel, but that isn't the primary use-case.

specification/metrics/sdk.md
dashpole: > feels like a potential loophole in the spec

I don't read it that way.  The first statement applies to OTel instruments. This paragraph just explicitly states what is already implicit in the statement above, which is that the config doesn't cover non-OTel instruments.

> burden of making sure the data is compatible

Even with the existing spec, exporters need to handle "whatever" being pushed to them, as users can configure default temporality and use views to override aggregations.  There is no burden on the SDK today to provide compatible data, and temporality/aggregation are just one of many potential incompatibilities exporters may need to handle.

I could see it being potentially useful for bridges to offer configuration of temporality, but spec-ing it as a requirement for all bridges is probably going too far.  I prefer keeping such configuration out of the SDK itself, as bridge-based config seems like a less-intrusive way to address the issue.

pirgeo: > Even with the existing spec, exporters need to handle "whatever" being pushed to them

This was not my understanding. Especially for "launchers/exporters" by vendors, it is feasible that they offer a convenience method that returns a fully configured MetricReader with the MetricExporter already set up. In this scenario, there is no way that data in a "wrong" AggregationTemporality gets to the exporter since the MetricReader is set up to export data in the AggregationTemporality that the exporter expects. As far as I am aware, Views cannot be used to change AggregationTemporality, so we also don't need to worry about that. The exporter needs to know how to deal with all _Aggregations_, yes, but not with all _AggregationTemporalities_. Is there wording in the spec that all exporters need to be able to deal with all AggreationTemporalities? 

I am probably overthinking this and seeing all the bad things people could do with this, but I am worried that a lot of time was spent making the MetricReader/MetricExporter connection watertight, and now we put a hole into it here.

specification/metrics/sdk.md
```diff
@@ -769,15 +776,19 @@ used with pull-based metrics collection.  A common sub-class of
 `MetricReader`, the periodic exporting `MetricReader` SHOULD be provided
 to be used typically with push-based metrics collection.
 
-The `MetricReader` MUST ensure that data points are output in the
-configured aggregation temporality for each instrument kind.  For
+The `MetricReader` MUST ensure that data points from OTel instruments are
```
reyang, line 779: Minor, consider adding a link to the Instruments in the API spec.

specification/metrics/sdk.md
```diff
@@ -898,8 +915,12 @@ primarily a simple telemetry data encoder and transmitter.
 Metric Exporter has access to the [aggregated metrics
 data](./data-model.md#timeseries-model).  Metric Exporters SHOULD
 report an error condition for data output by the `MetricReader` with
-unsupported Aggregation or Aggregation Temporality, as this condition
-can be corrected by a change of `MetricReader` configuration.
+unsupported Aggregation or Aggregation Temporality. For data originating from
+OpenTelemetry Instruments, this condition can be corrected by a change of
+`MetricReader` configuration. For metric data from a `MetricProducer`, it is an
+indication to the user that metric data is being dropped. The presence of data
+with an unsupported Aggregation or Aggregation Temporality SHOULD NOT prevent
```
reyang, line 922: SHOULD NOT or MUST NOT?

reyang (approved):
LGTM.

specification/metrics/sdk.md
dashpole: I put SHOULD NOT to make this a backwards-compatible change (there may be exporters which today don't send anything if they get a batch of metrics with an unsupported Aggregation Temporality).

reyang: Makes sense, thanks!

specification/metrics/sdk.md
dashpole: > Is there wording in the spec that all exporters need to be able to deal with all AggreationTemporalities?

Yes. In the [MetricExporter spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#metricexporter), it says: `Metric Exporters SHOULD report an error condition for data output by the MetricReader with unsupported Aggregation or Aggregation Temporality, as this condition can be corrected by a change of MetricReader configuration.`

You make a good point about vendor launchers being able to avoid such errors in the current spec.  Requiring MetricProducers to offer configuration of AggregationTemporality wouldn't work, though, as users can still use a misconfigured MetricProducer with a launcher. To avoid impacting the current launcher pattern, we would need to add additional configuration on the MetricReader for determining the aggregation temporality of metrics output from MetricProducers.  I don't love that we would have two different ways of configuring the output aggregation temporality: One that applies to OTel instruments, and another that applies to aggregated metric data...

WDYT?

pirgeo: > I don't love that we would have two different ways of configuring the output aggregation temporality: One that applies to OTel instruments, and another that applies to aggregated metric data...

Yeah, I agree. Especially if we won't do any conversions anyway. I think that the bridge is not the primary use case for the Reader/Exporter, so adding API surface for something that we want users to replace does not seem wise.

I must have missed the spec wording. The concept of supported temporalities (which was dropped a while back IIRC) would be really helpful here. Anyway, I don't think my comment here is blocking. 

I am trying to look at this from a user perspective. We might be sending users with little experience in OTel/AggregationTemporalities and an exporter that doesn't precisely tell them what they are doing wrong on a wild goose chase to figure out their metrics. We probably also don't even want them to have to think about AggregationTemporality, if we can help it. I guess I would have liked a more "fool-proof" migration path. I assume, however, that you are right with your assumption that if the backend supported the metrics before, it will probably support the bridged metrics as well. So 
from that perspective, all my worries are potentially unwarranted anyway. Do you think it would make sense to put a warning somewhere (maybe in the migration plan doc), just so users that do run into problems have a pointer to where they could start looking if things do go wrong?

Also: are there any other use-cases, apart from bridges, that could make use of the MetricProducer?

dashpole: The only other use-case i've seen suggested for this is as a way to work around limitations of the OTel api: https://github.com/open-telemetry/opentelemetry-specification/pull/2722#issuecomment-1218591597.

I'll add a warning to the migration guide (https://github.com/open-telemetry/opentelemetry-specification/pull/2730).

specification/metrics/sdk.md
dashpole: done

specification/metrics/sdk.md
pirgeo: I get it, but I don't like the idea of folks working around the API. What is the point of having a clearly defined and restricted API in the first place if you can just circumvent it? Anyway, I don't know if there is any (easy) way of prohibiting users to do it.

pirgeo:
Thanks to @dashpole for taking the time and walking me through all the different facets of this PR. I just wanted to point out clearly what I discovered in <https://github.com/open-telemetry/opentelemetry-specification/pull/2722#discussion_r965555967>, since it wasn't clear to me when I first read this. **Merging this PR will give everyone the possibility to circumvent the API** if they wish to do so. There is a number of scenarios in which this might be desired (such as bridges), but it might also lead to nonsense data points if the person creating the data points does not know exactly what they are doing. I don't see any way of prohibiting this use, but it feels like we should somehow discourage use of the MetricProducer unless necessary. It seems like there is a lot of potential confusion waiting to happen.

specification/metrics/sdk.md
```diff
@@ -769,15 +776,20 @@ used with pull-based metrics collection.  A common sub-class of
 `MetricReader`, the periodic exporting `MetricReader` SHOULD be provided
 to be used typically with push-based metrics collection.
 
-The `MetricReader` MUST ensure that data points are output in the
-configured aggregation temporality for each instrument kind.  For
-synchronous instruments being output with Cumulative temporality, this
-means converting [Delta to Cumulative](supplementary-guidelines.md#synchronous-example-cumulative-aggregation-temporality)
+The `MetricReader` MUST ensure that data points from OTel
```
jack-berg, line 779:
```suggestion
The `MetricReader` MUST ensure that data points from OpenTelemetry
```

specification/metrics/sdk.md
```diff
@@ -1086,6 +1108,50 @@ modeled to interact with other components in the SDK:
...
+failed or timed out. When the `Produce` operation fails, the `MetricProducer`
+MAY return successfully collected results and a failed reasons list to the
+caller.
+
+#### InstrumentationScope() InstrumentationScope
+
+`InstrumentationScope` provides the
+[InstrumentationScope](../glossary.md#instrumentation-scope) associated with
+the MetricProducer. A MetricProducer instance MUST return the same
+instrumentation scope for each invocation of InstrumentationScope().
```
jack-berg, line 1153: So all metrics from a producer share the same instrumentation scope. On the surface that seems overly restrictive since it would prevent bridging any metric system that has a similar notion of instrumentation scope as opentelemetry. What do we gain from this restriction? 

Not a blocking comment.

jack-berg (approved):

specification/metrics/sdk.md
dashpole: The primary motivation is to reduce instrumentation scope collisions (https://github.com/open-telemetry/opentelemetry-specification/pull/2722#discussion_r952966258 is related).  If we decide it is OK to pass through duplicate instrumentation scopes, and make exporters deal with it, then it seems fine to allow multiple instrumentation scopes from each bridge.

The way the spec is currently written, duplicate instrumentation scopes can't ever happen, because only a single meter with a given name + attributes can exist at a point in time.  My hope was that making a bridge have a single instrumentation scope would make it easier to reason about and handle collisions.

dashpole: We decided at the spec sig meeting today that it is acceptable to pass through duplicate instrumentation scopes.  The metric reader SHOULD emit a warning about the duplication, with steps for users to resolve the duplication.

specification/metrics/sdk.md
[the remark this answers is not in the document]
dashpole: We discussed this at the spec sig today.  We chose to allow passing through duplicate instrumentation scopes with a warning.

specification/metrics/sdk.md
dashpole: Added: 

`If a meter is created which produces an
[InstrumentationScope](../glossary.md#instrumentation-scope), which matches
the InstrumentationScope of a [MetricProducer](#metricproducer), or if multiple
[MetricProducers](#metricproducer) have the same InstrumenatationScope the SDK
SHOULD emit a warning.`

jmacd (approved):

specification/metrics/sdk.md
```diff
@@ -76,10 +81,16 @@ an [`InstrumentationScope`](../glossary.md#instrumentation-scope) instance which
...
-if appropriate.
+[MetricReaders](#metricreader), [MetricProducers](#metricproducer) and
+[Views](#view)) MUST be managed solely by the `MeterProvider` and the SDK MUST
+provide a way to configure all options that are implemented by the SDK. This
+MAY be done at the time of MeterProvider creation if appropriate.
+
+If a meter is created which produces an
+[`InstrumentationScope`](../glossary.md#instrumentation-scope), which matches
+the InstrumentationScope of a [MetricProducer](#metricproducer), or if multiple
+[MetricProducers](#metricproducer) have the same InstrumenatationScope the SDK
```
ahayworth, line 92:
```suggestion
[MetricProducers](#metricproducer) have the same InstrumentationScope the SDK
```

dashpole:
Thanks @bogdandrutu.  I updated the title.  I also pasted the notes from the 8/30 discussion in the PR description where we discussed the high-level design.

bogdandrutu:
My vote is to support Option 2:
1. Hard to explain that views are not there. We clearly separate that views are only for sdk.MeterProvider. 
2. `exporter configuration for the MeterProvider would be shared.`
    * This is true for the option 2 as well, since the MetricReader may be `shared`, isn't it?
3. `MeterProvider` will implement the same interface as any other producer.
4. Indeed user must configure Resource maybe twice, but that is easier in my opinion compare with explaining that the resource will be overwritten. (see "Resource MUST be provided by the MeterProvider, and any resource information provided by the bridge MUST be overridden.")

dashpole:
> My vote is to support Option 2:
> 
> 1. Hard to explain that views are not there. We clearly separate that views are only for sdk.MeterProvider.

Agreed.  That was one of the downsides discussed.  But given that we already have to explain that aggregation/ aggregation temporality configuration doesn't apply, it seems reasonable.

> 2. `exporter configuration for the MeterProvider would be shared.`
>    * This is true for the option 2 as well, since the MetricReader may be `shared`, isn't it?

You are correct that you can share the metric readers still with option 2.  It just means you have to pass the metric readers (and resource) as options to both the sdk and to bridges separately.

> 3. `MeterProvider` will implement the same interface as any other producer.

This is still possible with the current design, but I considered it an implementation detail.

> 4. Indeed user must configure Resource maybe twice, but that is easier in my opinion compare with explaining that the resource will be overwritten. (see "Resource MUST be provided by the MeterProvider, and any resource information provided by the bridge MUST be overridden.")

This may only apply to java (see https://github.com/open-telemetry/opentelemetry-specification/pull/2722#discussion_r950223056), but I haven't checked other languages.  The benefit of this is that you always end up grouping the metrics together under a single resource, which was important, if I'm remembering the discussion correctly.

specification/metrics/sdk.md
```diff
@@ -828,6 +847,11 @@ SDK](../overview.md#sdk) authors MAY choose to add parameters (e.g. callback,
 filter, timeout). [OpenTelemetry SDK](../overview.md#sdk) authors MAY choose the
 return value type, or do not return anything.
 
+The `MetricReader` MUST ensure the batch of metric points from
+`MetricProducers` are associated with the `MetricProducer`'s
+`InstrumentationScope()`, and with the `MeterProvider`'s configured resource if
+it is possible for those to conflict.
```
jmacd, line 853: What does "if it is possible for those to conflict." mean?

dashpole: I meant it to mean:  "If a batch of metric points from a MetricProducer can conflict with the MeterProvider's resource".  This would be possible if a "batch of metric points" includes resource information.  That wouldn't be the case in Go, but would be in Java given their current definition of "batch of metric points".

I could remove "if it is possible for those to conflict", or I could change it to "...configured resource.  If a resource or scope is already present in a batch of metric points from a MetricProducer, that MUST be overridden with the MetricProducer's Instrumentation Scope and the MeterProvider's resource".

jmacd: Thanks. I see this was also discussed in https://github.com/open-telemetry/opentelemetry-specification/pull/2722#discussion_r950223056. Now that I understand the Java SDK has Resource in its data object, I understand this point, which helps explain @bogdandrutu's remarks in https://github.com/open-telemetry/opentelemetry-specification/pull/2722#issuecomment-1251386971.

jmacd:
It appears there is lingering debate over how to require that Resources are-or-are-not-definitely-the-same between the bridged metrics and those from other Meters (and the same concern for Scopes, practically speaking). [Here's what is written in the Resource SDK spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#resource-sdk)

```
When used with distributed tracing, a resource can be associated 
with the [TracerProvider](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#tracerprovider) 
when the TracerProvider is created. That association cannot be changed later. 
When associated with a TracerProvider, all Spans produced by any Tracer 
from the provider MUST be associated with this Resource.

Analogous to distributed tracing, when used with metrics,
a resource can be associated with a `MeterProvider`.
When associated with a [`MeterProvider`](../metrics/api.md#meterprovider),
```

It's that "analogous" MUST we are defending here. It appears the stable Java SDK has a loophole, and now we have wrinkles in the specification because of it. I recommend @dashpole remove all the wrinkle about Resources "if it is possible for those to conflict", leave the specification stating that each producer has an instrumentation scope, and let the Java SDK document its loophole (i.e., it should document that users are required to use the same resource twice or else be out-of-spec).

I expect that to make the Producer interface the same as the internal representation of a Meter has; I expect that means the Producer will be invoked to produce a batch of Metric objects just as one Meter would do; therefore I expect the MetricProducer to produce the same data objects that any other Meter would, because the Scope and Resource are already determined at that level in the data hierarchy.

@bogdandrutu can you clarify if whether you have any reservations not covered above?

jmacd:
> MetricProducer to produce the same data objects that any other Meter would, because the Scope and Resource are already determined at that level in the data hierarchy.

p.s., I'm not opposed to allowing MetricProducers to return more than one scope. A MetricProducer returning multiple Scopes equals multiple single-Scope MetricProducers, IMO, what's important to me is that a MetricProducer doesn't require the MetricReader to sort data points by Scope to successfully export the data.

jmacd:
Would @ahayworth @pirgeo @MadVikingGod and @bogdandrutu in particular please read through the above dialog and see if you agree to approve the PR in its current form? The four of you have commented without approving.

Recall that this is absolutely a barrier to finishing the OTel-Go OpenCensus bridge, which is at least partlyt responsible for stalling development on Collector observability improvements. Thanks.

ahayworth:
> Would @ahayworth [...] please read through the above dialog and see if you agree to approve the PR in its current form? The four of you have commented without approving.

I have no strong feelings on the PR; I happened to notice a typo previously (which was the only thing in my comment). ❤️

specification/metrics/sdk.md
```diff
@@ -76,10 +81,16 @@ an [`InstrumentationScope`](../glossary.md#instrumentation-scope) instance which
...
+[MetricReaders](#metricreader), [MetricProducers](#metricproducer) and
+[Views](#view)) MUST be managed solely by the `MeterProvider` and the SDK MUST
+provide a way to configure all options that are implemented by the SDK. This
+MAY be done at the time of MeterProvider creation if appropriate.
+
+If a meter is created which produces an
+[`InstrumentationScope`](../glossary.md#instrumentation-scope), which matches
+the InstrumentationScope of a [MetricProducer](#metricproducer), or if multiple
+[MetricProducers](#metricproducer) have the same InstrumentationScope the SDK
+SHOULD emit a warning.
```
MadVikingGod, line 93: This seems like 2 separate warnings/errors. 

1. If an instrument is created, after applying views, such that it's scope conflicts with a MetricProducer.  If we create a meter that conflicts but never create an instrument that does, because of scope rename or not creating any instruments, is there an warning?
2. When registering MetricProducers if they were to collide then there should be an error/warning.  This should be a new norm forming sentence.

MadVikingGod (approved):

MadVikingGod:
I'm ok with the current PR, as I think it offers us a way to short-term migrate from OC to otel without the re-instrument everything step.  I think this will be implementable as written.

I still have a preference for approach 2, for two reasons.
1. I think it will be simpler to implement in go by creating a wrapped reader. 
2. I think it makes it easier to explain what is happening in the following contrived example. If you have 2x readers, and 1x MetricProducer that produces delta metrics, neither reader would get the entire stream of data from the producer.  As it is written now the MetricProducer will be shared by the MeterProvider among all Readers.  In approach 2 the user would have to explicitly reuse the producer between 2 or more readers.  

Because of 2 I would also recommend that we have some warning on any producers we create that can make delta metrics.   They shouldn't be used with multiple readers.

specification/metrics/sdk.md
jmacd: I agree this is two separate statements, but it doesn't say anything about per-instrument warnings. I believe we should emit warnings for the conflicting scopes, but allow the instruments to be correctly registered into each. IOW the presence of MetricProducers could produce scope warnings, not instrument warnings. The only way to get instrument warnings should be to have conflicts within a scope (IMO).

jmacd:
> I would also recommend that we have some warning on any producers we create that can make delta metrics.

I would be happy to add that MetricProducers MUST NOT use delta temporality, "for they are expected to be stateless" with respect to the MetricReader.

jmacd:
@dashpole I had been confused (and I suspect others have bene) by the option-numbering scheme mentioned in the PR description, especially considering the numbered-list appearing in (and later quoted) https://github.com/open-telemetry/opentelemetry-specification/pull/2722#issuecomment-1251386971.

@MadVikingGod's points are well taken in arguing for the bridge to be implemented as a wrapped MetricReader. Implementing the bridge inside the Reader is both simpler and allows the correct use of Delta Temporality. The downside of that approach is that warnings will not be realized until the first time they are read. This is completely fine, in my opinion.

@dashpole I recommend closing this PR, revising the contents of this PR, _and opening a new one where we do not list numbered alternatives_.

dashpole:
> I would be happy to add that MetricProducers MUST NOT use delta temporality, "for they are expected to be stateless" with respect to the MetricReader.

It seems like metric reader could still be stateless even with a stateless bridge/reader/exporter if both the bridge and exporter prefer delta.  If there was a client that natively produced delta temporality metrics, I don't think we'd want to disallow bridging that to compatible exporters, right?

dashpole:
I opened https://github.com/open-telemetry/opentelemetry-specification/pull/2838, which removes "if it is possible for those to conflict" language, splits the conflicting-scope warning sentence into two, and removes the numbered options from the description.

PR #2736, opened 2022-08-17
Proposal for flattening attributes from OTLP messages

### LINKED ISSUE #2535: Clarify precedence of attributes when exporting to non-OTLP formats

pirgeo:
Fixes #2535

## Changes

Adds a new document, attribute-precedence.md, which is intended to give guidance on how to flatten out attributes when transforming OTLP to a flat set of attributes with unique keys (e.g. Prometheus/OpenMetrics labels).
This document is intended as a supplementary guideline.

Note that the guidelines are valid for semantic convention attributes as well as for custom attributes, where producers can put whatever they want. These custom attributes might overwrite other custom attributes or semantic convention attributes.

Related issues #2535


specification/common/attribute-precedence.md
```diff
@@ -0,0 +1,177 @@
...
+Below, the precedence for each of the signals is spelled out explicitly.
+Only spans, metric data points and log records are considered.
+
+`A > B` denotes that the attribute on `A` will overwrite the attribute on `B`
+if the keys clash.
+
+### Traces
+
+```
+Span.attributes > ScopeSpans.scope.attributes > ResourceSpans.resource.attributes
```
yurishkuro, line 73: What is a ScopeSpan?

pirgeo: ScopeSpans is the wrapper combining multiple spans and a Scope: https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto#L64

yurishkuro: I see. If I were implementing a flattening scheme, I would keep ALL attributes, e.g. by adding distinguishable prefix for each category, instead of letting them override each other.

jmacd: @yurishkuro yes. Another way of putting this: Why flatten in the first place?

The fact that OpenMetrics specifies a "flat" set of attributes does not mean we should flatten resource and scope-level attributes. OpenMetrics specifies how to join resource attributes using the `target_into` metric, and @dashpole has https://github.com/open-telemetry/opentelemetry-specification/pull/2703 in progress, specifying how to join scope attributes using an opentelemetry_scope_info metric.

specification/common/attribute-precedence.md
```diff
@@ -0,0 +1,177 @@
...
+attribute3: resource-attribute-3      # from the resource, not overwritten
+attribute4: scope-attribute-4         # from the scope, not overwritten
+
+# data point 2
+service.name: my-service              # from the resource
+attribute1: data-point-2-attribute-1  # overwrites attribute1 on resource & scope
+attribute2: scope-attribute-2         # overwrites attribute2 on resource
+attribute3: resource-attribute-3      # from the resource, not overwritten
+attribute4: data-point-2-attribute-4  # overwrites attribute4 from the scope
+```
```
scheler, line 170: Asking just for my understanding. Can you list a real life example of an attribute that could be present both at the resource and at the level of a span/metric/log_record?

specification/common/attribute-precedence.md
yurishkuro: > Another way of putting this: Why flatten in the first place?

No, I don't have an issue with flattening, it may be necessary due to the limitations of a target telemetry platform. But it does not mean that flattening should be a lossy transformation, which is what you're proposing.

pirgeo: The big argument against adding a prefix is that it changes the attribute key. Users can not query for the attribute they added, since the name changed. This impacts semantic conventions as well: Should all semantic conventions be prefixed? I feel like we would need to do that to stay consistent.

> But it does not mean that flattening should be a lossy transformation, which is what you're proposing.

That is true, but this is the idea behind this proposal: If you want to, you _can_ overwrite attributes. If you don't want to overwrite attributes, you can rename them (e.g., add a prefix explicitly). In either case, you will get the attributes that you defined, and don't have to go looking for the renamed version.

specification/common/attribute-precedence.md
pirgeo: One thing that comes to mind is a `responsible_team`. In the resource attribute, you would set the department that is responsible for the respective product, and you can overwrite it on Spans/Metrics if the code producing them is maintained by a specific team within that department. That way, the resource would be the "fallback" but you can be more specific if you want to.

scheler: Thanks, I used to think we will never have same attribute in both the resource and the signals since they serve a different purpose, but looks like there can be a few valid cases.

I want to question and understand your example though. The team responsible for running my service pod could be devops (specified at resource level), but the team responsible for the application is an engineering/dev team (specified in the span). Don't you want to capture both?

pirgeo: I agree, that they should never overlap, but since its possible to put whatever you want on the attributes it might happen.

For the example, I thought about it more like the team that developed the code. The team running the code is separate in my opinion. `responsible_team` might not be the best name for the attribute, maybe separate attributes like `dev_team` and `devops_team` would be better if you want to distinguish them.

I was thinking of an attribute value of `responsible_team=dev` (or `dev_team=dev`) on the resource for all the code that was written before instrumentation was added. All code written (and instrumented) later will add the actual team as an attribute on the span/metric/log (e.g. `responsible_team=java-service-team`/`dev_team=java-service-team`).

pirgeo: @scheler does this answer your question?

scheler: @pirgeo since there are no examples from the existing set of attribute semantic conventions that overlap between resource and signal, it would good to add some text to the specification on how to deal with it if we see valid use-cases in future. Maybe as part of https://github.com/open-telemetry/opentelemetry-specification/issues/2753.

specification/common/attribute-precedence.md
jack-berg: > If you want to, you can overwrite attributes.

This is a good point. I buy the logic that conflicts are unlikely to be common, and thus that this proposal is going to work reasonably well in most situations. If a conflict does arise, the user has an escape hatch with the ability to wrap the exporter with logic that adds a prefix to the key in conflict.

specification/common/attribute-precedence.md
```diff
@@ -0,0 +1,177 @@
...
+that exporters should follow when translating from the hierarchical OTLP format
+to non-hierarchical formats.
+
+A mapping is required when flattening out attributes from the structured OTLP
+format, which has attributes at different levels (e.g., Resource attributes,
+InstrumentationScope attributes, attributes on Spans/Metrics/Logs) to a
+non-hierarchical representation (e.g., Prometheus/OpenMetrics labels).
+In the case of OpenMetrics, the set of labels is flat and must have unique
+labels only
+(<https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#labelset>).
```
jack-berg, line 37: As @jmacd points out [here](https://github.com/open-telemetry/opentelemetry-specification/pull/2736/files#r948355142), there is ongoing work to solve this in a different way specifically for prometheus/openmetrics in #2703. It would be better to use zipkin as an example here.

joaopgrassi:
Not stale

PR #2743, opened 2022-08-19
Clarify the HTTP client `http.retry_count` spec

### LINKED ISSUE #2737: HTTP retry spec clarification needed

I started working on implementing the HTTP retry spec in the Java HTTP client instrumentations, and I ran into several questions that I'd like to clarify.

* According to [this example](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-redirects-examples) redirects don't add the `http.retry_count` attribute, is that right?
* What about authorization scenarios? E.g. the server returns 401, so the HTTP client automatically adds a new `Authorization` header and retries the request. Should this scenario follow the same behavior as redirects?
* Does the retry counter reset if e.g. redirect requests are retried? E.g.
   ```
   GET / - 500 (CLIENT, trace=t1, span=s1)
    |
    --- server (SERVER, trace=t1, span=s2)

   GET / - 302 (CLIENT, trace=t2, span=s1, http.retry_count=1)
    |
    --- server (SERVER, trace=t2, span=s2)

   GET /hello - 500 (CLIENT, trace=t3, span=s1)
    |
    --- server (SERVER, trace=t3, span=s2)

   GET /hello - 200 (CLIENT, trace=t4, span=s1, http.retry_count=1)
    |
    --- server (SERVER, trace=t4, span=s2)
   ```
* I've skimmed through several HTTP clients sources, and some of them use the same code for all kind of retries (redirects, auth, actual retries). Would it be correct to assume that status codes >=400 (or 500? depending on the auth scenario I guess) should be treated as retries? Request finished with 3xx (and lower) would not get the retry count attribute.

**Additional context.**

https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5722

mateuszrzeszutek:
Fixes #2737

## Changes

Clarifies when the `http.retry_count` attribute should be added to the HTTP client span (every time the HTTP request gets re-sent, no matter what the cause is).

I also added a grouping "Examples" header for all examples, similar to other (messaging, DB) convention docs.


semantic_conventions/trace/http.yaml
```diff
@@ -117,6 +117,10 @@ groups:
         type: int
         brief: >
           The ordinal number of request re-sending attempt.
+        note: >
+          The retry counter SHOULD be updated each time an HTTP request gets re-sent by the client, regardless of what
+          was the cause of a retry (e.g. redirection, authorization failure, 503 Server Unavailable, network issues,
```
Oberon00, line 122:
```suggestion
          was the cause of the re-sending (e.g. redirection, authorization failure, 503 Server Unavailable, network issues,
```

semantic_conventions/trace/http.yaml
```diff
@@ -117,6 +117,10 @@ groups:
         type: int
         brief: >
           The ordinal number of request re-sending attempt.
```
Oberon00, line 119:
```suggestion
          The ordinal number of request re-sending attempt (for any reason, including redirects).
```

I think the name "retry_count" is not optimal and I would at least call this aspect out in the brief.

specification/trace/semantic_conventions/http.md
```diff
@@ -161,10 +165,10 @@ Retries and redirects cause more than one physical HTTP request to be sent.
 A CLIENT span SHOULD be created for each one of these physical requests.
 No span is created corresponding to the "logical" (encompassing) request.
 
-For retries, `http.retry_count` attribute SHOULD be added to each retry span
-with the value that reflects the ordinal number of request retry attempt.
+For each re-sent HTTP request, `http.retry_count` attribute SHOULD be added to each retry span
```
Oberon00, line 168:
```suggestion
For each time a HTTP request is re-sent, the `http.retry_count` attribute SHOULD be added to each retry span
```

specification/trace/semantic_conventions/http.md
```diff
@@ -161,10 +165,10 @@ Retries and redirects cause more than one physical HTTP request to be sent.
...
+with the value that reflects the ordinal number of request re-send attempt.
```
Oberon00, line 169:
```suggestion
set to the ordinal number of the request re-send attempt.
```

specification/trace/semantic_conventions/http.md
```diff
@@ -355,7 +361,35 @@ GET / - 200 (CLIENT, trace=t3, span=s1, http.retry_count=2)
...
+```
+
+Example of redirects with no trace started upfront:
+
+```
+GET / - 401 (CLIENT, trace=t1, span=s1)
+ |
+ --- server (SERVER, trace=t1, span=s2)
+
+GET /hello - 200 (CLIENT, trace=t2, span=s1, http.retry_count=1)
```
Oberon00, line 387: Ugh, that this causes another trace to be started is ugly, but a logical consequence of how things are defined. So I think the example is correct, but maybe this hints at room for improvement (in _some_ way, e.g. adding a mechanism to starting multiple root spans in the with the same trace ID; changing the parent/child relationships of the HTTP spans like all becoming children of the first attempt; etc.)

specification/trace/semantic_conventions/http.md
```diff
@@ -355,7 +361,35 @@ GET / - 200 (CLIENT, trace=t3, span=s1, http.retry_count=2)
  --- server (SERVER, trace=t3, span=s1)
 ```
 
-## HTTP redirects examples
+### HTTP client authorization retry examples
+
+Example of redirects in the presence of a trace started by an inbound request:
```
Oberon00, line 366:
```suggestion
Example of retries in the presence of a trace started by an inbound request:
```

specification/trace/semantic_conventions/http.md
```diff
@@ -355,7 +361,35 @@ GET / - 200 (CLIENT, trace=t3, span=s1, http.retry_count=2)
...
+  -- GET / - 401 (CLIENT, trace=t1, span=s2)
+  |   |
+  |   --- server (SERVER, trace=t1, span=s3)
+  |
+  -- GET /hello - 200 (CLIENT, trace=t1, span=s4, http.retry_count=1)
+      |
+      --- server (SERVER, trace=t1, span=s5)
+```
+
+Example of redirects with no trace started upfront:
```
Oberon00, line 380:
```suggestion
Example of retries with no trace started upfront:
```

Oberon00 (approved):

specification/trace/semantic_conventions/http.md
joaopgrassi: Sorry if this was already discussed, but I wonder why do we use `retry` in the key name, but in the text `re-send`? 🤔. Like, why we don't say: `For each time a HTTP request is retried, the..`?

semantic_conventions/trace/http.yaml
```diff
@@ -117,6 +117,10 @@ groups:
...
+        note: >
+          The retry counter SHOULD be updated each time an HTTP request gets re-sent by the client, regardless of what
```
joaopgrassi, line 121:
```suggestion
          The retry count SHOULD be updated each time an HTTP request gets re-sent by the client, regardless of what
```
Not sure what is best, counter as in the "counter variable" or the "count concept". I feel `counter` is already giving impl details in the spec.

specification/trace/semantic_conventions/http.md
joaopgrassi: Another suggestion, would be dropping "For each":

```suggestion
Each time an HTTP request is re-sent, the `http.retry_count` attribute SHOULD be added to each retry span
```

specification/trace/semantic_conventions/http.md
joaopgrassi: if going with the suggestion, don't forget an `and` before `set` :)

specification/trace/semantic_conventions/http.md
```diff
@@ -161,10 +165,10 @@ Retries and redirects cause more than one physical HTTP request to be sent.
 A CLIENT span SHOULD be created for each one of these physical requests.
 No span is created corresponding to the "logical" (encompassing) request.
 
-For retries, `http.retry_count` attribute SHOULD be added to each retry span
-with the value that reflects the ordinal number of request retry attempt.
+For each time a HTTP request is re-sent, the `http.retry_count` attribute SHOULD be added to each retry span
+set to the ordinal number of the request re-send attempt.
```
joaopgrassi, line 169:
```suggestion
and set to the ordinal number of the request re-send attempt.
```

semantic_conventions/trace/http.yaml
mateuszrzeszutek: Hm, should we rename it to `resend_count` then? "Resend" might be a little less intuitive than "retry", but on the other hand it doesn't introduce any new concept - it relies on the HTTP RFC terminology

joaopgrassi (commented):
Thanks for the extra examples and for better formatting them :) 

I just have a question on naming. The attribute is `retry` but in the text we use the word `re-send/sent`. Not sure if it's a problem but I wonder if we shouldn't be consistent?

specification/trace/semantic_conventions/http.md
mateuszrzeszutek: > hanging the parent/child relationships of the HTTP spans like all becoming children of the first attempt

I think it could be a little confusing, like for example the first SERVER span would be on the same level as all the resent CLIENT spans.

How about using span links though? Spans emitted for resent HTTP requests could have a link to the original CLIENT span (in fact, the `retry_count` could even be an attribute of the link, not the span itself).

semantic_conventions/trace/http.yaml
mateuszrzeszutek: True; and if the attribute name uses "count" then I think it's only reasonable to use "count" everywhere. Will change, thanks!

mateuszrzeszutek:
> I just have a question on naming. The attribute is `retry` but in the text we use the word `re-send/sent`. Not sure if it's a problem but I wonder if we shouldn't be consistent?

I'm in favor of this idea; initially I didn't want to change too much of the spec, just clarify things, but if more people than just me think it's a valid point I think we could rename it to `resend_count`.
(I also mentioned this in https://github.com/open-telemetry/opentelemetry-specification/pull/2743#discussion_r951400886)

semantic_conventions/trace/http.yaml
joaopgrassi: I'm more inclined in retry, resend somehow reads odd to me. My comment around inconsistency was more to always use retry everywhere.

Also, I'm not entirely sure what is the process If we decide to rename the attribute. The conventions are experimental still, but I _think_ we will need to introduce a schema change in that case?

Oberon00: > My comment around inconsistency was more to always use retry everywhere

That would be consistent and easy to understand, but it would be wrong 😉

joaopgrassi: Ah got it, now it makes sense and thanks for the background on it. If we don't want to rename the existing, there could be another just for the "non error" cases like redirect but that's probably an overkill, so maybe renaming it's the way to go. Then I guess we will do need a schema change?

Oberon00: If we had two counters that would be very confusing and open many questions IMHO, the first obvious one is that of categorizing. Retries and network errors are clear, but what about re-sends after an authentication challenge for example? Error or not?
The second one, do you reset the other counter to zero if you increment the first? No matter what you do, you don't get an unambiguous order of the requests from the two counters anymore. You would need a third counter.
So I would definitely say we should stay with a single counter, no matter what it is called.

mateuszrzeszutek: > Then I guess we will do need a schema change?

Yes, I suppose so.

I'll add a commit renaming this attribute to `resend_count`, let's see how it looks like.

mateuszrzeszutek:
@open-telemetry/specs-approvers @open-telemetry/technical-committee please take a look at this PR

cc @lmolkova (sorry, I completely forgot to tag you before 🙈)

trask (approved):

lmolkova (approved):
LGTM, thank you @mateuszrzeszutek!

specification/trace/semantic_conventions/http.md
```diff
@@ -346,16 +352,44 @@ GET / - 500 (CLIENT, trace=t1, span=s1)
...
 
-## HTTP redirects examples
+### HTTP client authorization retry examples
+
+Example of retries in the presence of a trace started by an inbound request:
+
+```
+request (SERVER, trace=t1, span=s1)
+  |
+  -- GET / - 401 (CLIENT, trace=t1, span=s2)
```
joaopgrassi, line 371: Shouldn't the initial call here also be `/hello`? Or am I missing something? Also in the examples in `Example of retries with no trace started upfront`.

mateuszrzeszutek: Good catch, thanks! I'll fix these.

specification/trace/semantic_conventions/http.md
```diff
@@ -128,23 +130,25 @@ before any HTTP-redirects that may happen when executing the request.
...
+| `http.resend_count` | int | The ordinal number of request resending attempt (for any reason, including redirects). [2] | `3` | Recommended: if and only if request was retried. |
+| [`net.peer.name`](span-general.md) | string | Host identifier of the ["URI origin"](https://www.rfc-editor.org/rfc/rfc9110.html#name-uri-origin) HTTP request is sent to. [3] | `example.com` | Required |
+| [`net.peer.port`](span-general.md) | int | Port identifier of the ["URI origin"](https://www.rfc-editor.org/rfc/rfc9110.html#name-uri-origin) HTTP request is sent to. [4] | `80`; `8080`; `443` | Conditionally Required: [5] |
 
 **[1]:** `http.url` MUST NOT contain credentials passed via URL in form of `https://username:password@www.example.com/`. In such case the attribute's value should be `https://www.example.com/`.
 
-**[2]:** Determined by using the first of the following that applies
+**[2]:** The resend count SHOULD be updated each time an HTTP request gets resent by the client, regardless of what was the cause of the resending (e.g. redirection, authorization failure, 503 Server Unavailable, network issues, or any other).
```
reyang, line 139: What is considered as a "redirect"? Could a browser get an HTTP redirect response which points to FTP?

joaopgrassi:
Sorry, forgot to hit send in one comment: You probably need to add an entry to the changelog for this as well.

specification/trace/semantic_conventions/http.md
Oberon00: That's not really relevant. If there would be such a thing as a redirect to FTP, you would not resend the HTTP request, so the point is moot. This spec part has the purpose of _not_ needing to care whether it's a redirect or not.

mateuszrzeszutek: Also, I think it's up to the HTTP spec to define what a "redirect" or a "resend" is -- we're just using these terms here in the same meaning as in https://www.rfc-editor.org/rfc/rfc9110

reyang: I'm very confused, this PR suggests that "resend count SHOULD be updated each time **an HTTP request gets resent**", what does "an HTTP request gets resent" mean? "An HTTP request" = same HTTP request (same verb, URI, headers, payload) or something else (e.g. as long as it is a redirect, it is considered as "an HTTP request")?

reyang: > That's not really relevant. If there would be such a thing as a redirect to FTP, you would not resend the HTTP request (you might establish an FTP connection), so the point is moot.

The proposal seems to imply that redirects can only happen from HTTP to HTTP, and that's why I asked for clarification. Do we know that for all the other protocols that HTTP can redirect to, there is guarantee that they won't redirect back to HTTP? (or do we want to say if the redirect points to something that is not HTTP, that's undefined behavior which is not intended to be covered by the spec?)

Oberon00: > The proposal seems to imply that redirects can only happen from HTTP to HTTP

That's not what this implies. This attribute is only set on spans tracing HTTP requests, so it does not matter to this spec if you get redirected to something else, because then you get outside the scope of this spec. Maybe the confusion comes from the word "updated"? Because actually you should not update the attribute on the span, but just make sure that the next span your instrumentation produces for the same logical request hast the attribute incremented.

Oberon00: > what does "an HTTP request gets resent" mean?

As @mateuszrzeszutek mentioned, this term is also used in the RFC (https://www.rfc-editor.org/rfc/rfc9110#name-redirection-3xx). It's also not clearly & generally defined there and I think that's OK. APIs/libraries and consequently instrumentations will have different possibilities to track what a "resend" is.
If you asked me, I would say: A request is resent when an HTTP client library sends more than one HTTP request to satisfy the same API call. This may happen due to following redirects (that's the context where the RFC uses it), "retries" (which are not clearly defined either), authentication challenges, and maybe other reasons I'm not aware of.

joaopgrassi (approved):

specification/trace/semantic_conventions/http.md
reyang: > If you asked me, I would say: A request is resent when an HTTP client library sends more than one HTTP request to satisfy the same API call. This may happen due to following redirects (that's the context where the RFC uses it), "retries" (which are not clearly defined either), authentication challenges, and maybe other reasons I'm not aware of.

This makes a lot of sense and I think it's a good addition to the spec since it clarifies a lot.

reyang: > > Do we know that for all the other protocols that HTTP can redirect to, there is guarantee that they won't redirect back to HTTP?
> 
> No, we but I don't think we need to specify this, it seems to be an highly hypothetical edge case. (AFAIK, the concept of "redirect" does not exist in many protocols other than HTTP, in the first place). You could also argue this for the previous spec "what if after an HTTP failure, the retry happens by downloading the URL from FTP?". I think these questions are not relevant in practice.

I don't know if it's rare of not (not an expert in protocols), here is my concern (and the reason why I asked these questions):
1. Do we have domain experts here?
2. When we don't have a clear picture/understanding, do we try to be clear+defensive (e.g. callout that "if the redirect/retry/etc. points to something that is not HTTP, it is out of the scope of this spec) or leave it vague (and assume that we won't run into the problems)?

reyang (approved):
LGTM.

schemas/1.13.0
```diff
@@ -9,6 +9,10 @@ versions:
             attribute_map:
               net.peer.ip: net.sock.peer.addr
               net.host.ip: net.sock.host.addr
+        # https://github.com/open-telemetry/opentelemetry-specification/pull/2743
```
carlosalberto, line 12: Move to the (new) 1.14 schema ;)

mateuszrzeszutek: Done :+1:
Thanks!

mateuszrzeszutek:
Ping @open-telemetry/technical-committee : is this PR ready to be merged? Is there anything else that I can help with?

cijothomas (approved):

specification/trace/semantic_conventions/http.md
reyang: The last question is not a blocker for this PR, marking it as resolved for now to unblock the progress.

Issue #2711, opened 2022-08-04
Specify how to stop and restart metrics reporting


jmacd:
**What are you trying to achieve?**

Trying to address  
https://github.com/open-telemetry/opentelemetry-js/issues/2997
and
https://github.com/open-telemetry/opentelemetry-specification/issues/1891

There are situations when a metrics SDK wants to stop reporting data for a particular instrument and attribute set. This comes about differently for asynchronous/synchronous instruments, depending on cardinality choice. 

In every combination of sync/async and delta/cumulative, we find the situation may arise. We find that to safely stop reporting metrics requires attention to what information is lost to the consumer, especially where it may lead to inaccurate rate calculations.

For example:

- synchronous/cumulative: cumulative implies long-term memory use, so need to stop reporting when too much is too much; the question is what start time to use when restarting, and what it says about the rate in the interim period
- synchronous/delta: this case is always safe, SDKs are free to stop reporting an instrument/attribute pair when it has not been used during a collection cycle; this leaves a gap, considered normal for delta temporality
- asynchronous/cumulative: in this case, it is safe to stop reporting the instrument/attribute pair--the user is free to simply not observe the attributes. This will leave a gap in the record, it's is not considered good practice for cumulative reporting
- asynchronous/delta: in this case, it is safe to stop reporting the instrument/attribute pair, but restarting the same instrument/attribute pair is complicated _for the same reason as synchronous/cumulative_.

What did you expect to see?

In the 8/3 Prometheus-WG SIG meeting this was discussed. An idea to use the NO_DATA_PRESENT staleness marker as a way to communicate to the consumer. There appears to be some benefit to issuing NO_DATA_PRESENT data points for a period of time before being allowed to forget the value and erase it from memory.

Informally, I think we expect to see that in case the same instrument/attributes pair is re-used immediately, the new start time assigned will be no earlier than the last NO_DATA_PRESENT data point that was written. Ideally, the new start time assigned will be no later than the previous collection timestamp..

jack-berg:
> asynchronous/delta: in this case, it is safe to stop reporting the instrument/attribute pair, but restarting the same instrument/attribute pair is complicated for the same reason as synchronous/cumulative.

[...]

PR #2747, opened 2022-08-23
[enh] Make max_accumulations configurable and add `remove` method for meters.

tjiuming:
Make max_accumulations configurable and add `remove` method for meters.

Related PR: https://github.com/open-telemetry/opentelemetry-java/pull/4647

### DIFF
--- a/specification/metrics/api.md
+++ b/specification/metrics/api.md
@@ -641,7 +641,7 @@ def pf_callback(result):
     result.Observe(37741921, ("pid", 4),   ("bitness", 64))
     result.Observe(10465,    ("pid", 880), ("bitness", 32))
 
-meter.create_observable_counter(name="PF", description="process page faults", pf_callback)
+meter.create_observable_counter(name="PF", description="process page faults", max_accumulations=1000, pf_callback)
 ```
 
 ```csharp
@@ -656,7 +656,7 @@ interface IAtomicClock
 
 IAtomicClock clock = AtomicClock.Connect();
 
-meter.CreateObservableCounter<UInt64>("caesium_oscillates", () => clock.GetCaesiumOscillates());
+meter.CreateObservableCounter<UInt64>("caesium_oscillates", 1000, () => clock.GetCaesiumOscillates());
 ```
 
 #### Asynchronous Counter operations
@@ -677,7 +677,7 @@ class Device:
 
     def __init__(self, meter, x):
         self.x = x
-        counter = meter.create_observable_counter(name="usage", description="count of items used")
+        self.counter = meter.create_observable_counter(name="usage", description="count of items used", max_accumulations=1000)
         self.cb = counter.register_callback(self.counter_callback)
 
     def counter_callback(self, result):
@@ -685,6 +685,9 @@ class Device:
 
     def read_counter(self):
         return 100  # ...
+    
+    def remove(self):
+        self.counter.remove({'x', self.x})
 
     def stop(self):
         self.cb.unregister()
@@ -856,6 +859,7 @@ def cpu_frequency_callback():
 meter.create_observable_gauge(
     name="cpu.frequency",
     description="the real-time CPU clock speed",
+    max_accumulations=1000,
     callback=cpu_frequency_callback,
     unit="GHz",
     value_type=float)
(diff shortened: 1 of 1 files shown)


specification/metrics/api.md
```diff
@@ -300,6 +301,13 @@ instrument. It MUST be treated as an opaque string from the API and SDK.
 * It MUST support at least 1023 characters. [OpenTelemetry
   API](../overview.md#api) authors MAY decide if they want to support more.
 
+#### Instrument max_accumulations
+
+The `max_accumulations` is an optional free-form integer provided by the author of the
```
cijothomas, line 306: what is "accumulation"? This seems like an implementation detail from Java?

tjiuming: like Counter.Child in Prometheus. Currently, OTEL SDK limits the maximum number of Child to 2000, it should be configurable

cijothomas: Dont see anything like "Counter.Child" in OTel specs. Most likely this is something very java specific.

tjiuming: So, in Otel specs, the number of child in unlimited? `accumulation`'s limits is just a java spec?

how about `remove` method? should it be added to Otel spec?

tjiuming: @cijothomas sorry, I've read the https://github.com/open-telemetry/opentelemetry-java/pull/4647 's comments carefully, `max_accumulations` is truly a java spec, please ignore about it.

besides, I wonder that if `remove` method should be added to Otel spec?

jack-berg: Accumulation is a term internal to java's implementation. The max accumulations refers to the number of distinct attribute sets / timeseries that are allowed in memory for a particular instrument at any time. If exceeded, we log a warning and drop the measurement. 

The topic of metrics SDK memory limits was punted on for the initial stable release of the metrics spec, and in java (and I believe dotnet) we opted for defensive approach with a fixed limit until the spec clears things up.

cijothomas: Thanks for clarifying! Yes .NET also uses its own mechanism, called `SetMaxMetricPointsPerMetricStream`.
 https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/docs/metrics/customizing-the-sdk#changing-maximum-metricpoints-per-metricstream

Would be good to have it in spec. (allowing sufficient flexibility so that we wont break existing API in .NET)

tjiuming: @jack-berg @cijothomas If I understand you correctly, it's good to add `max_accumulations` and `remove` to spec right?

jmacd:
I believe this is the same as https://github.com/open-telemetry/opentelemetry-specification/issues/1891

jmacd:
OpenTelemetry has resisted adding a `remove()` or `delete()` method; I personally do not feel it is necessary. We are able to reset start timestamps and eject streams from memory, we just need to clearly write the specification.

tjiuming:
> OpenTelemetry has resisted adding a `remove()` or `delete()` method; I personally do not feel it is necessary. We are able to reset start timestamps and eject streams from memory, we just need to clearly write the specification.

if there are a meter named `broker_published_message_count` to record each topic's incoming messages, we often code like this:
```
Counter counter = ...
counter.add({namespace = 'xxx', topic = 'xxx'}, 1)
```
after the topic deleted, I believe we need to discard the accumulation directly, not through other indirect ways:
```
counter.remove({namespace = 'xxx', topic = 'xxx'})
```

specification/metrics/api.md
cijothomas: I dont know if there is a need to have `remove`.
I definitely like spec to have the equivalent of `max_accumulations`. (need a diff. name, as accumulation is java specific.) .NET named it `SetMaxMetricPointsperMetricStream`

jmacd:
@tjiuming thanks for explaining. We should discuss these two problems separately. For the limit question, please use #1891. 

Although I understand the precedent set in Prometheus and implied meaning in OpenMetrics when data is removed from a metrics pipeline, I would like OpenTelemetry to ensure correctness. In the example you gave, it is difficult to know when it is safe to remove the metric; we might for example specify that data must be reported before it can be removed, which leads to fairly difficult questions. After you remove the data, how should the data be safely removed from memory?  For this question, please see #2711. 

Lastly, I wonder if you've considered using delta temporality to work around this problem? The intention behind delta temporality is that any synchronous-instrument-timeseries that you stop using can be erased from memory after it is exported. If you configured your SDK to export to an OTel collector using delta temporality, the collector could be configured to address this problem. The problem still has to be solved (i.e., there's a limit on memory) but at least it's outside your SDK.

PR #205, opened 2019-08-07
Add SDK span processors spec

pavolloffay:
Resolves #155 
Resolves #54 

`Block when the queue is full processor` will be added in a separate PR.

Signed-off-by: Pavol Loffay <ploffay@redhat.com>

PR #2750, opened 2022-08-25
Make messaging context propagation requirements explicit

### LINKED ISSUE #2749: Integrate OTEP-205 (messaging context propagation requirements) into the spec

Integrate the [OTEP-205 Context propagation requirements for messaging semantic conventions](https://github.com/open-telemetry/oteps/blob/main/text/trace/0205-messaging-semantic-conventions-context-propagation.md) into the spec.

**Additional context.**

As stated in the OTEP, the existing messaging semantic conventions for tracing define trace examples that implicitly require the context to be propagated via the message. The OTEP's goal was to make the context propagation requirement clear and explicit so instrumentation authors have more guidance when working on messaging instrumentation.

joaopgrassi:
Fixes #2749

## Changes

Related [OTEP #205](https://github.com/open-telemetry/oteps/blob/main/text/trace/0205-messaging-semantic-conventions-context-propagation.md)

This PR integrates the [Context propagation](https://github.com/open-telemetry/oteps/blob/main/text/trace/0205-messaging-semantic-conventions-context-propagation.md#context-propagation) section from the OTEP into the trace messaging semantic conventions. 

I did some slight modification on the text, just to give more context on the "problem" and why this change is necessary. The core concept of the OTEP is still maintained of course.


blumamir (approved):

lmolkova (approved):
Left a small comment, but looks good overall!

specification/trace/semantic_conventions/messaging.md
```diff
@@ -80,6 +81,46 @@ Often such destinations are unnamed or have an auto-generated name.
...
+
+The message creation context is created by the producer and should be propagated
+to the consumer(s).
+
+Consumer traces cannot be directly correlated to producer traces if the message
+creation context is not attached and propagated with the message.
+
+A producer SHOULD attach a message creation context to each message.
+The message creation context SHOULD be attached in such a way that it is
+not possible to be changed by intermediaries.
```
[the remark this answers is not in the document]
joaopgrassi: I see your point. What about this:

```
A producer SHOULD attach a message creation context to each message.
The message creation context SHOULD be attached to the most appropriate section on the message,
(e.g. annotations, properties) so that it flows unchanged to consumers.
Intermediaries SHOULD NOT modify the creation context on the message.
```

specification/trace/semantic_conventions/messaging.md
```diff
@@ -80,6 +81,46 @@ Often such destinations are unnamed or have an auto-generated name.
...
+that all components are instrumented and propagate context accordingly.
+For example, an application using a messaging system offered via
+software as a service (SaaS) that does not propagate context will lead to incomplete traces.
+
+To be able to correlate consumer traces with producer traces without requiring
+intermediary instrumentation, the context needs to be propagated on a
+*per-message* basis instead of on a *per-request* basis. This allows all components
+to have access to the same per-message context information, making it possible
+to correlate all stages involved in processing a message with the message's
+creation.
```
dpauls, line 101: I'm not sure *exactly* what is being said here. I'm reading it possibly two ways:
1. A request can contain multiple messages (i.e. a batch), and the context should be per-message, not per-batch.
2. Intermediaries may modify request headers, so the creation context shouldn't go in request headers associated with the message, it should go in the message itself, where it is immutable to intermediaries.

I think the intent is the latter. If that's the case, my suggestion is changing this paragraph to something like:
> To be able to correlate consumer traces with producer traces, it's important the message creation context such that it cannot be modified by intermediaries. If, for example, the context is placed in request headers associated with the message, uninstrumented intermediaries or intermediaries not aware of the propagation mechanism will not remove or modify the context. To achieve this, the context should be propagated within the message where it is immutable by brokers.

specification/trace/semantic_conventions/messaging.md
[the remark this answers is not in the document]
dpauls: I prefer the original text. Given that the SHOULD directive is being use here, if there are scenarios in a messaging system where it's not possible to associate immutable annotations or headers to the message payload, then the SHOULD directive is violated, and justifiably so.

dpauls (changes requested):

specification/trace/semantic_conventions/messaging.md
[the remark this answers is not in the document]
joaopgrassi: Yeah that's what I think as well but I can see how others might be confused. 

>The message creation context SHOULD be attached in such a way that it is
not possible to be changed by intermediaries.

If the context can't be attached in such way, then the SHOULD is implicitly applying to intermediaries, meaning they should not change it.

specification/trace/semantic_conventions/messaging.md
joaopgrassi: It's not saying neither I think :D. 

The only purpose of this statement is to compliment the "problem" introduced in the first two paragraphs. In those paragraphs, I mention a message can go through many "hops" and traditional context propagation cannot really be used, since we can't rely on all components to be instrumented. 

This paragraph gives the "solution" - meaning that to be able to have a meaningful trace, the context has to be propagated via the message, and not using traditional context propagation mechanisms (http headers for ex) that are "request based". That's where the "per-message" vs "per-request" basis comes from. 

It's pretty much the same as in the original OTEP, under "Motivation": https://github.com/open-telemetry/oteps/blob/main/text/trace/0205-messaging-semantic-conventions-context-propagation.md#motivation

specification/trace/semantic_conventions/messaging.md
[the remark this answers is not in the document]
joaopgrassi: @lmolkova @dpauls what you think about this? I understand both your sides, but I think I tend to agree with Liudmila's concerns of "no physical limitations and brokers can change it" . The proposal I send above makes this explicit, while also making it clear the context needs to add to a section on a message so that it "flows" to consumers. Let me know if you have other suggestions so we can resolve this.

specification/trace/semantic_conventions/messaging.md
joaopgrassi: @dpauls did my reasoning help in clarify anything? I think it still makes sense to have it, (maybe with some modification) to indicate that "traditional" context propagation is not sufficient and that we need to use the message as a way of propagating context. 

Let me know if you have other suggestions to make the text clearer.

specification/trace/semantic_conventions/messaging.md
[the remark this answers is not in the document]
lmolkova: Would a little bit more relaxed version of this work? My only concern is with the `possible`  word

>  A producer SHOULD attach a message creation context to each message.
The message creation context SHOULD be attached in such a way that intermediaries don't modify it.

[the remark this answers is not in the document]
lmolkova: >  if there are scenarios in a messaging system where it's not possible to associate immutable annotations or headers to the message payload, then the SHOULD directive is violated, and justifiably so.

You're right, but I'm afraid somebody can read it in a way 'if there are no immutable annotations, I won't attach a context at all'. What I want people to read: do your best effort to propagate the context unchanged to the consumers.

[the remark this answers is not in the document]
joaopgrassi: Removing `possible` I think doesn't help much. 

> do your best effort to propagate the context unchanged to the consumers

There's two actors in here. Producers are on one side (need to attach the context to a "good" place in the message) and intermediaries that should not alter it. 

The current text only addresses producers. It puts the burden on them and in a way that we _now_ it's not always possible to achieve (not all messaging systems offer a immutable place to put the context).

That's why tried to introduce two clear sentences, one addressing each "actor". Like above, 

```
A producer SHOULD attach a message creation context to each message.
The message creation context SHOULD be attached to the most appropriate section on the message,
(e.g. annotations, properties) so that it reaches consumers with its original value.
Intermediaries SHOULD NOT modify the creation context on the message.
```

This gives a bit more guidance for producers on how they should attach the context and how it should arrive on consumers, and makes it clear that other parties involved should not modify it.

specification/trace/semantic_conventions/messaging.md
```diff
@@ -80,6 +81,46 @@ Often such destinations are unnamed or have an auto-generated name.
...
+all components must propagate context down the chain.
+
+Due to the complex nature of messaging systems, it cannot be assumed
+that all components are instrumented and propagate context accordingly.
+For example, an application using a messaging system offered via
+software as a service (SaaS) that does not propagate context will lead to incomplete traces.
+
+To be able to correlate consumer traces with producer traces without requiring
+intermediary instrumentation, the context needs to be propagated on a
+*per-message* basis instead of on a *per-request* basis. This allows all components
```
brunobat, line 98: One clarification here...
The spec now mentions: 
`A message is an envelope with a potentially empty payload. This envelope may offer the possibility to convey additional metadata, often in key/value form.`
It seems to assume the message metadata should be used for context propagation, meaning the individual  messaging attributes related with tracing should be placed there.
Is this correct?
If so, it seems that context propagation should already happen in a per-message basis. 
I think it would be important to clarify the metadata's role.

specification/trace/semantic_conventions/messaging.md
```diff
@@ -80,6 +81,46 @@ Often such destinations are unnamed or have an auto-generated name.
 
 Given these definitions, the remainder of this section describes the semantic conventions for Spans describing interactions with messaging systems.
 
+### Context propagation
+
+A message may traverse many different components and layers in one or more intermediaries
```
pyohannes, line 86: We should add the term "intermediary" to the definitions section of this document. Otherwise, it might not be clear for some readers what is meant by it.

specification/trace/semantic_conventions/messaging.md
pyohannes: > The page seems to assume the message metadata should be used for context propagation, meaning the individual messaging attributes related with tracing should be placed there.

Message metadata is in most cases be the most natural choice for propagating per-message context. However, at this point we don't want to be too specific, but rather leave it to future protocol-specific recommendations (AMQP, MQTT, HTTP) of how to propagate the creation context.

pyohannes (approved):
One small request, otherwise this looks good to me.

specification/trace/semantic_conventions/messaging.md
brunobat: thanks @pyohannes, that makes sense to me. Can this be explicitly mentioned?

joaopgrassi: Hi @brunobat !

>The page seems to assume the message metadata should be used for context propagation, meaning the individual messaging attributes related with tracing should be placed there.

That portion in the document is very generic and doesn't give any mention about context propagation for tracing. So I feel it's not addressing anything nor clarifying. By just looking at that I don't feel it's clear that a context should be attached to metadata.

Moreover, in a thread above, I suggested changing the portion "A producer SHOULD attach a message creation context to each message..." to:

```
A producer SHOULD attach a message creation context to each message.
The message creation context SHOULD be attached to the most appropriate section on the message,
(e.g. annotations, properties) so that it reaches consumers with its original value.
Intermediaries SHOULD NOT modify the creation context on the message.
```

I feel this gives more "hints" on how/where the context _should_ be placed, without explicitly saying where, since we can't tell exactly where as Johannes explained.

>thanks @pyohannes, that makes sense to me. Can this be explicitly mentioned?

This is already clarified below, where it says: 

>This document does not specify the exact mechanisms on how the creation context is attached/extracted to/from messages. Future versions of these conventions will give clear recommendations, following industry standards including, but not limited to [Trace Context: AMQP protocol](https://w3c.github.io/trace-context-amqp/) and [Trace Context: MQTT protocol](https://w3c.github.io/trace-context-mqtt/) once those standards reach a stable state.

Do you think that's not addressing it, or not clear enough? If so, I'm open for more suggestions on how to make it clearer :).

specification/trace/semantic_conventions/messaging.md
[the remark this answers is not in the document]
joaopgrassi: @lmolkova @pyohannes , @brunobat , @dpauls @blumamir , pinging to see if you agree we should use the proposal change above. If the majority agrees, I will change it.

specification/trace/semantic_conventions/messaging.md
joaopgrassi: Good point. I will add it, just need to properly "research" a bit to get a good enough terminology wording😅

brunobat (approved):

specification/trace/semantic_conventions/messaging.md
pyohannes: This could be used: https://github.com/open-telemetry/oteps/blob/main/text/trace/0173-messaging-semantic-conventions.md#intermediary.

Which is based on the [CloudEvents definition](https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#intermediary).

specification/trace/semantic_conventions/messaging.md
joaopgrassi: Ok, marking this as resolved them. There's another thread above about changing the text, so will continue there.

specification/trace/semantic_conventions/messaging.md
dpauls: Ok, I understand, but as worded I think it can be a bit hard to understand. I think the description of the message-creation context, and the requirements about how producers should attach a context, which should not be modifiable by intermediaries. covers all of this. Do we need to say anything here? Or just explain that the solution to the problem is the message creation context, described below?

specification/trace/semantic_conventions/messaging.md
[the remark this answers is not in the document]
dpauls: In the meeting today, my understanding of the concern is that *if* there is no way in a messaging system to attach the creation context such that intermediaries cannot modify it, we don't want the interpretation be that no message creation context should be attached at all. However, I feel the original wording has two distinct requiremnts that cover this:
* A producer SHOULD attach a message creation context to each message.
* The message creation context SHOULD be attached in such a way that it is
not possible to be changed by intermediaries.

If the second clause cannot be fulfilled, it does not indicate the first clause is not applicable. This seems like the simplest, clearest way of expressing our intent, without imposing requirements on intermediaries.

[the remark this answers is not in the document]
pyohannes: I'd prefer this normative section to be very concise and not make any _requirements_ for intermediaries (as intermediaries will most likely not feel compelled to comply to this document). 

I wonder what you think about a slightly changed wording:
> A producer SHOULD attach a message creation context to each message.
If possible, the message creation context SHOULD be attached in such a way that it cannot be changed by intermediaries.

This should make it clearer that the second requirement depends on the first. This could be elaborated further in the non-normative text, giving an example of using read-only message attributes to propagate context.

specification/trace/semantic_conventions/messaging.md
joaopgrassi: I noticed that we didn't have terminology for `producers` and `consumers`. I borrowed the text for them. I also made it clear a section for `Message` inside "definitions".

specification/trace/semantic_conventions/messaging.md
joaopgrassi: @dpauls Hum, reading this I'm a bit confused.

> Such tracing would necessarily evolve the trace context associated with the messaging system's transport layer

Can you elaborate on it? What did you mean with "evolve" here?

Anyway I agree the current text I have is too much. I tried taking parts of yours and came up with this:

```
Messaging systems themselves may trace messages as the messages travels from
producers to consumers. Such tracing would cover the transport layer but would
not help in correlating producers with consumers. To be able to directly
correlate producers with consumers, another context that is propagated with
the message is required.
```
What do you think?

specification/trace/semantic_conventions/messaging.md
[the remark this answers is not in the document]
joaopgrassi: Ok, I went with @pyohannes suggestion above. Let me know what you think. [bca1e39](https://github.com/open-telemetry/opentelemetry-specification/pull/2750/commits/bca1e39ffa00a662d52dc4b9388be5900567c183)

specification/trace/semantic_conventions/messaging.md
joaopgrassi: I went with it, let me know what you think! [bca1e39](https://github.com/open-telemetry/opentelemetry-specification/pull/2750/commits/bca1e39ffa00a662d52dc4b9388be5900567c183)

dpauls: > What did you mean with "evolve" here?

That's an excellent question, it was a poor choice of words, that's for sure. That's what I get for trying to write while keeping an ear open in a meeting.

I think I was trying to say:
> Such tracing would necessarily modify the trace context associated with the messaging system's transport layer.

Your suggested text is good though.

specification/trace/semantic_conventions/messaging.md
```diff
@@ -43,6 +50,27 @@ With Apache Kafka, the physical broker a message is written to depends on the nu
 
 Messages can be delivered to 0, 1, or multiple consumers depending on the dispatching semantic of the protocol.
 
+### Producer
+
+The "producer" is a specific instance, process or device that creates and
+publishes a message. "Publishing" is the process of sending a message or batch
```
dpauls, line 56: This is a minor point, but here we use the term of a "batch" without defining it. It is somewhat explained in the examples section, but I think it deserves a definition, possibly right after message. Or perhaps changing "batch" to "a batch of messages" is sufficient.

If we want to define the term batch, something like this could be aded after the "Message" section, before the "Producer" section.

> Batch
A batch is a term used to describe a collection of messages that are sent, received, or processed as a single operation.

dpauls (approved):
Overall, I'm happy with these changes. I had one small suggestion regarding the term "batch", which would be nice to address, but I am willing to approve even if we don't define the term.

specification/trace/semantic_conventions/messaging.md
joaopgrassi: Good point. I think we can properly define what a batch is in a follow up (or maybe part of the several iterations we have coming). For now, I used your suggestion of `batch of messages`. I think that's clearer. Thanks for pointing it out! [b97911d](https://github.com/open-telemetry/opentelemetry-specification/pull/2750/commits/b97911d45553dc8254fcea8cc5c1880efbfb7e65)

specification/trace/semantic_conventions/messaging.md
joaopgrassi: Alright, Will close this then! thanks for the suggestions.

specification/trace/semantic_conventions/messaging.md
```diff
@@ -80,6 +108,38 @@ Often such destinations are unnamed or have an auto-generated name.
...
+
+A message may traverse many different components and layers in one or more intermediaries
+when it is propagated from the producer to the consumer(s). To be able to correlate
+consumer traces with producer traces using the existing context propagation mechanisms,
+all components must propagate context down the chain.
+
+Messaging systems themselves may trace messages as the messages travels from
+producers to consumers. Such tracing would cover the transport layer but would
+not help in correlating producers with consumers. To be able to directly
+correlate producers with consumers, another context that is propagated with
```
brunobat, line 121: I imagine intermediaries should be able to correlate the message to the producer.

carlosalberto (approved):

specification/trace/semantic_conventions/messaging.md
joaopgrassi: Yeah it would _if_ the intermediaries are instrumented, which I think it's not the case today? We discussed this in the last SIG meeting and we decided to avoid for now trying to specify anything related to intermediaries. That's why I think it's better to not be specific about this here.

pyohannes (approved):

CHANGELOG.md
```diff
@@ -33,6 +33,9 @@ release.
   metrics semantic conventions
   ([#2706](https://github.com/open-telemetry/opentelemetry-specification/pull/2706))
 
+- Make context propagation requirements explicit for messaging semantic conventions
```
reyang, line 36:
```suggestion
- Make context propagation requirements explicit for messaging semantic conventions
```

reyang (approved):
LGTM.

specification/trace/semantic_conventions/messaging.md
```diff
@@ -43,6 +50,27 @@ With Apache Kafka, the physical broker a message is written to depends on the nu
...
+
+### Consumer
+
+A "consumer" receives the message and acts upon it. It uses the context and
+data to execute some logic, which might lead to the occurrence of new events.
+
+The consumer receives, processes, and settles a message. "Receiving" is the
+process of obtaining a message from the intermediary, "processing" is the
+process of acting on the information a message contains, "settling" is the
+process of notifying an intermediary that a message was processed successfully.
```
ppatierno, line 67: Is "settling" really the notification about the message to be processed?
In general, the intermediary is only interested to know that the message was delivered to the consumer so on "receiving" it should be settled.
Processing the message is part of the "contract" between producer and consumer, so that by using a request-reply pattern and a "reply-to" address/queue/topic, the consumer can notify (via a new message) to the producer that the processing happened.

ppatierno (commented):
Sorry to be late on this, but I left a comment.

PR #2754, opened 2022-08-26
Clarify that recording status refers to IsRecording and not Set Status

bogdandrutu:
Small clarification for the specs about what operations can be called after span finishes. There are other improvements that can be made here such as using right capitalized word for may, but that is a different concern/PR.

### DIFF
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -436,7 +436,7 @@ Links SHOULD preserve the order in which they're set.
 ### Span operations
 
 With the exception of the function to retrieve the `Span`'s `SpanContext` and
-the `IsRecording` status, none of the below may be called after the `Span` is
+`IsRecording`, none of the below may be called after the `Span` is
 finished.
 
 #### Get Context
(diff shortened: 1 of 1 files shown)


specification/trace/api.md
```diff
@@ -436,7 +436,8 @@ Links SHOULD preserve the order in which they're set.
 ### Span operations
 
 With the exception of the function to retrieve the `Span`'s `SpanContext` and
-recording status, none of the below may be called after the `Span` is finished.
```
bogdandrutu, line 439: when I read this(I thought about `Set Status` section) I was confused, so trying to remove that confusion from the text.

specification/trace/api.md
```diff
@@ -436,7 +436,8 @@ Links SHOULD preserve the order in which they're set.
...
+the `IsRecording` status, none of the below may be called after the `Span` is
```
Oberon00, line 439:
```suggestion
`IsRecording`, none of the below may be called after the `Span` is
```

Maybe we should just drop the word "status" here completely.

Oberon00 (approved):

carlosalberto (approved):

PR #2755, opened 2022-08-29
Definition of boolean environment variables on the SDK

### LINKED ISSUE #2729: Definition of Boolean for Environment Variables
Some environment variables need the definition of a Boolean type.
Example: OTEL_SDK_DISABLED
This discussion started in this PR: #2679 
Because this is an important enough subject, we decided to spinoff the definition into this new issue. 
The original discussion was proposing the following definition: 

### Boolean value
Any value that represents a boolean MUST interpret the case-insensitive string `"true"`, meaning `"True"` or `"TRUE"` are also accepted as `"true"`.
An SDK MAY extend this definition and define additional values that are interpreted as true.
Any value not explicitly defined here or by the SDK as a true value, including unset and empty values, MUST be interpreted as false.

Also, a recommendation on the naming of environment variable attributes should also be agreed in order to align boolean property names.

brunobat:
Signed-off-by: brunobat <brunobat@gmail.com>

Fixes #2729 
Present PR is based on the issue's discussion.

## Changes

Adds the definition of a boolean environment variables to the SDK.

## Related issues

[Make OTEL_EXPERIMENTAL_SDK_ENABLED stable](https://github.com/open-telemetry/opentelemetry-specification/issues/2667)


specification/sdk-environment-variables.md
```diff
@@ -14,6 +14,13 @@ The SDK MUST interpret an empty value of an environment variable the same way as
 
 **Status**: [Stable](document-status.md)
 
+### Boolean value
+
+Any value that represents a boolean MUST only interpret the string `"true"` as meaning true.
+An SDK MAY NOT extend this definition and define additional values that are interpreted as true.
```
carlosalberto, line 20: Any reason to not make this a MUST?

specification/sdk-environment-variables.md
```diff
@@ -14,6 +14,13 @@ The SDK MUST interpret an empty value of an environment variable the same way as
 
 **Status**: [Stable](document-status.md)
 
+### Boolean value
+
+Any value that represents a boolean MUST only interpret the string `"true"` as meaning true.
```
jack-berg, line 19: I'd prefer to see this be case insensitive, which is the [idiomatic way to strings to boolean](https://docs.oracle.com/javase/7/docs/api/java/lang/Boolean.html#parseBoolean(java.lang.String)) in java.

brunobat: yeah... That was the initial proposal, however concerns were raised from other languages.

specification/sdk-environment-variables.md
brunobat: Just because different languages might have other practices. If no one objects I will add this change.

specification/sdk-environment-variables.md
```diff
@@ -14,6 +14,13 @@ The SDK MUST interpret an empty value of an environment variable the same way as
...
+An SDK MAY NOT extend this definition and define additional values that are interpreted as true.
+Any value not explicitly defined here or by the SDK as a true value, including unset and empty values, MUST be interpreted as false.
```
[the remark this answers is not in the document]
brunobat: Do we have a current practice on how to log these things?... I don't want to open a new can of worms.

specification/sdk-environment-variables.md
jack-berg: Are you referring to this [comment](https://github.com/open-telemetry/opentelemetry-specification/issues/2729#issuecomment-1222584006), or something else?

It wasn't clear to me from that discussion that there was disagreement with case insensitivity. @dyladan can you clarify?

specification/sdk-environment-variables.md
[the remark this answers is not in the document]
carlosalberto: In other types we have some text mentioning SDKS should report a warning, e.g. Enum section. Can use the same.

specification/sdk-environment-variables.md
jack-berg: @dyladan confirmed in the 8/30 Spec SIG that he's not opposed to case insensitivity.

Aneurysm9, line 19:
```suggestion
Any value that represents a boolean MUST interpret only the string `"true"` as meaning true.
```

dyladan: I don't have a particular opinion on case insensitivity but it should be consistent across all languages. I would suggest we do not use "the idiom in language X is Y" as an argument here since I would strongly recommend we implement the exact same behavior in all SDK languages. A much stronger argument would be "the idiom in MOST OTel supported languages is X"

specification/sdk-environment-variables.md
Aneurysm9: I think this needs to be `MUST NOT` otherwise we will have inconsistency between languages and surprising behavior for users.  Alternately, this sentence can be stricken entirely as the previous sentence precludes treating any other values as `true`.

brunobat: MUST NOT will be used.

specification/sdk-environment-variables.md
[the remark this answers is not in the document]
brunobat: Will add @arminru suggestion

specification/sdk-environment-variables.md
brunobat: I think extending to many values the definition of "true" might add unneeded complexity. It seems to me that case insensitive "true" should be enough. 
Will update the definition and wait for further feedback.

specification/sdk-environment-variables.md
```diff
@@ -14,6 +14,14 @@ The SDK MUST interpret an empty value of an environment variable the same way as
 
 **Status**: [Stable](document-status.md)
 
+### Boolean value
+
+Any value that represents a boolean MUST interpret only the case-insensitive string "true", meaning "True" or "TRUE" are also accepted as "true".
```
Aneurysm9, line 19:
```suggestion
Any value that represents a boolean MUST interpret only the case-insensitive string "true", meaning "True" or "TRUE" are also accepted, as "true".
```

specification/sdk-environment-variables.md
```diff
@@ -14,6 +14,14 @@ The SDK MUST interpret an empty value of an environment variable the same way as
...
+An SDK MUST NOT extend this definition and define additional values that are interpreted as true.
+Any value not explicitly defined here or by the SDK as a true value, including unset and empty values, MUST be interpreted as false.
```
Aneurysm9, line 21:
```suggestion
Any value not explicitly defined here as a true value, including unset and empty values, MUST be interpreted as false.
```

SDKs are prohibited from defining additional truthy values, no need to include them here.

jack-berg (approved):

arminru (approved):

marcalff (changes requested):

specification/sdk-environment-variables.md
```diff
@@ -14,6 +14,14 @@ The SDK MUST interpret an empty value of an environment variable the same way as
...
+If any value other than a true value, `"false"`, empty, or unset is used, a warning SHOULD be logged to inform users about the fallback to false being applied.
+All boolean environment variables SHOULD be named and defined such that false is the expected safe default behavior.
+
```
marcalff, line 24: I have to voice an objection here.

Le't consider a use case and see how it would work:

Assume a feature written to support both IPV4 and IPV6, at a time where IPV4 is mainstream and IPV6 not yet adopted. Desired default values are to enable IPV4 and disable IPV6 by default.

With this rule, environment variables are to be named:
- `IPV4_DISABLED`
- `IPV6_ENABLED`

This creates confusion in environment variable names from the introduction of the feature, for no apparent benefit.

When later the IPV6 feature is more mainstream, and the default is changed to have both IPV4 and IPV6 enabled, the environment variables naming convention are still different, causing confusion forever.

I would suggest to base the name of an environment variable only on what a feature **IS**, so it stays immutable.

The default value of an environment variable depends on how the feature **IS USED**, and is not immutable.

Coupling the name with the value is not a good thing, in my opinion.

Please reconsider.

Regards,
-- Marc

marcalff (changes requested):

specification/sdk-environment-variables.md
```diff
@@ -14,6 +14,14 @@ The SDK MUST interpret an empty value of an environment variable the same way as
 
 **Status**: [Stable](document-status.md)
 
+### Boolean value
+
+Any value that represents a boolean MUST interpret only the case-insensitive string "true", meaning "True" or "TRUE" are also accepted as "true".
+An SDK MUST NOT extend this definition and define additional values that are interpreted as true.
+Any value not explicitly defined here or by the SDK as a true value, including unset and empty values, MUST be interpreted as false.
+If any value other than a true value, `"false"`, empty, or unset is used, a warning SHOULD be logged to inform users about the fallback to false being applied.
```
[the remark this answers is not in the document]
brunobat: We could remove "false", "False", "FALSE" from the warnings. Would that work for you?

specification/sdk-environment-variables.md
brunobat: not sure about that comma.

specification/sdk-environment-variables.md
brunobat: SHOULD is not mandatory and I think allows this particular cases:
IPV4 (default value True)
IPV6 (default value False)
They would be valid property names, even if they are quite ambiguous on what they represent. 
In most cases using Disabled and Enabled sufixes clarifies the property meaning and will not require frequent renaming.

CHANGELOG.md
```diff
@@ -85,6 +85,9 @@ release.
 
 ### SDK Configuration
 
+- Add the definition of a Boolean environment variable
+  ([#2755])(https://github.com/open-telemetry/opentelemetry-specification/pull/2755)
```
reyang, line 89: The format seems wrong:

```suggestion
  ([#2755](https://github.com/open-telemetry/opentelemetry-specification/pull/2755))
```

specification/sdk-environment-variables.md
Aneurysm9: The phrase `meaning "True" or "TRUE" are also accepted` is parenthetical and can be elided without changing the meaning of the containing sentence.  Thus, it should be delimited by commas.

specification/sdk-environment-variables.md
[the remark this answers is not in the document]
brunobat: On the Spec SIG Meeting, removing ""false", "False", "FALSE" from the warnings" was accepted.

reyang (approved):

specification/sdk-environment-variables.md
dyladan: > IPV6_ENABLED can not be renamed to IPV6_DISABLED, because that would break applications

is that true? If the default for version 1 is ipv4 and you have `IPV6_ENABLED=true`, then the default changes in version 2 and you have the same environment, the behavior is unchanged because the old variable name is now just an unknown name and is dropped. Maybe i'm missing something here?

specification/sdk-environment-variables.md
yurishkuro: > Any value that represents a boolean MUST interpret only the case-insensitive string "true"

Please rephrase this. How can a "value" "interpret" anything?

specification/sdk-environment-variables.md
marcalff: @dyladan,

If there is an existing item in a user configuration file:
- export IPV6_ENABLED="XYZ"

looking up for an environment variable of a different name (now IPV6_DISABLED) will ignore the user setting, potentially changing behavior, so renaming causes issues.

My point is that, without renaming, when things change of default value, and with history accumulated after a couple of versions, some environment variables will be named FOO_ENABLED, others BAR_DISABLED, even when they are related and are expected to follow the same rules, without a clear reason.

In other words, as a user I never know if the setting for FOO is FOO_ENABLED or FOO_DISABLED, and setting for similar things (IPV4 and IPV6) can use inconsistent naming, if the rule is followed too literally.

Like ending up with IPV4_DISABLED and IPV6_ENABLED in a world where both IPV4 and IPV6 are enabled.

@brunobat,

I am ok with suffixes like _ENABLED or _DISABLED, this adds clarity.

The point that causes concern is that someone SHOULD pick the suffix depending on the default value, instead of picking a name for what the setting represents.

In other words either
- IPV4_ENABLED (=True by default)
- IPV6_ENABLED (=False by default)
or
- IPV4 (=True by default)
- IPV6 (=False by default)
looks better in my opinion compared to:
- IPV4_DISABLED (=False by default) <-- because the default value SHOULD be false
- IPV6_ENABLED (=False by default)

Now, if this is a "SHOULD" only and not a "MUST", it still leaves freedom for people to pick a proper name, but there is more ...

My understanding of
>   Any value not explicitly defined here or by the SDK as a true value, including unset and empty values, MUST be interpreted as false.

is an other way to say that the default value ("unset") of an environment variable MUST be false.

So, when reading the spec,
- IPV4_ENABLED (=True by default)
- IPV6_ENABLED (=False by default)
is illegal

Maybe I am not understanding this correctly.

Regards,
-- Marc

jack-berg: That's a good point @marcalff. I think that changes the default value is a breaking change and not allowed, but as you point out we may want to change the default value in the next major version and we'd be in a confusing state as the names of env vars would have to change.

IMO this trumps the benefit of naming env vars such that the default is always false, such that an empty / missing value aligns with the default. I've seen plenty of instances where env vars / system properties default to true, so this behavior wouldn't be surprising to me.

I'm switching my stance, but don't have a super strong opinion. I will say that if we continue down the path of naming variables such that the default is always false, we should disallow changing the default value even in major versions to avoid the problems described by @marcalff.

trask: fwiw, I prefer `*enabled` flags over `*disabled` flags (even if that means the default value is true), and we use this pattern for all boolean flags in the otel java agent

brunobat: @MrAlias any comment on this?

specification/sdk-environment-variables.md
brunobat: updated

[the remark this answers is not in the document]
brunobat: Fixed

specification/sdk-environment-variables.md
brunobat: On the Spec SIG meeting we agreed:
Renaming and changing the default value MUST not happen between major versions.
Will update the definition

MrAlias: A boolean value should not be used to define the internet protocol used by an application. It is a poor example of a boolean variable and presents a straw man argument.

I disagree there is going to be a case where OTel defines an environment variable as a boolean value with an intended default and later decides to change the default value. If you have a valid example, I would appreciate hearing it.

If this definition prohibits OTel developers from including boolean environment variables that have defaults change from under users in the future, all the better. It may seem like this is an added burden to OTel developers that they need to think harder about the design of their configuration, but this project is not prioritized solely for OTel developers. Instead, its users need to be considered first.

Removing user confusion is key. User confusion around if they _are_ or _are not_ required to provide an environment variable diminishes when they know that by not providing it it will be assumed false. They do not need to go look up documentation elsewhere about the value, they just know because that is the OTel default.

Similarly, they know these values will always have the default of `false`. If they upgrade their use of OTel they know that default value will not change. It would be a huge frustration for a user to upgrade their system and all of a sudden they now need to set `OTEL_SDK_DISABLED=false` because we allowed the default to change.

_**If**_ there does exist a valid example where a boolean variable needs to have it's default value changed, though I have yet to hear one, it still makes sense from the user perspective to keep static defaults. A new variable would be defined and the old variable deprecated. Old code would still work, maybe with a logged error to encourage moving the new value, and the new configuration would also work. Still the user is able to rely on each defaulting to `false` without further documentation lookup. If you version past a major version and you drop support for the old configuration variable, then you have kept in line with [this projects versioning and stability guarantees](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#sdk-stability). Allowing the default value to change would violate these guarantees.

MrAlias (approved):

specification/sdk-environment-variables.md
marcalff: > I disagree there is going to be a case where OTel defines an environment variable as a boolean value with an intended default and later decides to change the default value. If you have a valid example, I would appreciate hearing it.

I don't have an example in mind, so let's put this to rest.

marcalff (approved):

CHANGELOG.md
```diff
@@ -82,6 +82,9 @@ release.
 
 ### SDK Configuration
 
+- Add the definition of a Boolean environment variable
```
carlosalberto, line 85: Update the location of the entry in the CHANGELOG please ;)

specification/sdk-environment-variables.md
```diff
@@ -14,6 +14,14 @@ The SDK MUST interpret an empty value of an environment variable the same way as
 
 **Status**: [Stable](document-status.md)
 
+### Boolean value
+
+Any value that represents a Boolean MUST be set to true only by the case-insensitive string `"true"`, meaning `"True"` or `"TRUE"` are also accepted, as true.
+An SDK MUST NOT extend this definition and define additional values that are interpreted as true.
+Any value not explicitly defined here as a true value, including unset and empty values, MUST be interpreted as false.
+If any value other than a true value, case-insensitive string `"false"`, empty, or unset is used, a warning SHOULD be logged to inform users about the fallback to false being applied.
+All Boolean environment variables SHOULD be named and defined such that false is the expected safe default behavior. Renaming and changing the default value MUST not happen between major versions.
```
dyladan, line 23: >Renaming and changing the default value MUST not happen between major versions

This is confusingly worded. I think they're trying to say that these changes can only happen when the major version is changed, but it could easily be interpreted to be the opposite.

```suggestion
All Boolean environment variables SHOULD be named and defined such that false is the expected safe default behavior.
Renaming and changing the default value MUST NOT happen without a major version change.
```

dyladan (approved):
LGTM with a minor wording clarification

brunobat:
Seems there are no outstanding issues. Can we merge?

CHANGELOG.md
brunobat: I don't really know where else to place this. Should I create a new section for "OpenTelemetry Environment Variable" that is actually part of the SDK Configuration.

carlosalberto: Unreleased section ;) (we released 1.13 on Monday)

jack-berg (approved):

CHANGELOG.md
brunobat: got it

brunobat:
@jmacd I think this is ready now

Issue #753, opened 2020-07-31
Current "Invalid" SpanContext definition precludes TraceState-only SpanContext


Oberon00:
## Current Definition of IsValid
Currently, we consider a SpanContext invalid if the trace or span ID is invalid (i.e. it is invalid unless both are valid), and this is exposed with an API function. As a result, for example in Java, an invalid span context is not inherited by child spans (a fresh context with new span and trace ID is generated, discarding any TraceState or TraceFlags).

It is also not propagated by any propagators, but that is a decision the propagators make for themselves and could be overriden by custom propagators.

One problem with this is, if you wanted to implement a non-W3C-based extractor that doesn't have a trace+span ID available, you are now forced to invent a trace+span ID on your own in the propagator.

## Data that might be on an extracted SpanContext

1. IsRemote will be true.
2. A sampled flag has to be there (even if none was propagated via the original propagation protocol, the propagator will have to invent one)
3. A trace ID might be there (even if there is no span ID)
4. A span ID might be there (even if there is no trace ID)
5. TraceState entries might be there (which may be read from headers other than W3C `tracestate`)

## Proposed spec changes

> A SpanContext MUST be considered valid unless it has neither a valid trace nor span ID, nor IsRemote=true.

Rationale: The minimal information that can be contained on a remote parent is just the sampled flag, and we may still want to respect that.

This has one implication for the SDK spec I'm aware of:

> The tracing SDK MUST accept any valid SpanContext as per the above definition as parent for a span if requested. However, a child span MUST NOT inherit an invalid parent trace ID. In that case the SDK MUST generate a new trace ID (as it would do if no parent was used). All other properties of the SpanContext are inherited as usual (e.g. IsRemote is made false, TraceState is copied).

We should also clarify (independent of other changes):

> An invalid SpanContext MUST be treated as if no SpanContext was provided in all cases where the option of providing no SpanContext exists (and vice versa).

bogdandrutu:
@Oberon00 there is already a definition of "IsValid", what else do we need?

Oberon00:
Oh, I missed that. But my argument against the current definition (I thought was a Java thing) still stands.

bogdandrutu:
@Oberon00 I think you need to challenge this in w3c specs, see https://www.w3.org/TR/trace-context/#no-traceparent-received 4.2.3

bogdandrutu:
[...]

PR #2763, opened 2022-08-30
Refactor messaging attributes and per-message attributes in batching scenarios

lmolkova:
Messaging instrumentation SIG is working on [spec changes](https://github.com/open-telemetry/oteps/pull/192) and this change is one of the first steps to bring the consensus we reached in SIG to the spec.

This change clarifies that per-message attributes should be set on links when the corresponding span represents a batching operation. It introduced breaking changes (attribute renames)
- `messaging.message_id` to `messaging.message.id`
- `messaging.conversation_id` to `messaging.message.conversation_id`
- `messaging.message_payload_size_bytes` to `messaging.message.payload_size_bytes`
- `messaging.message_payload_compressed_size_bytes` to `messaging.message.payload_compressed_size_bytes`.

Going forward, we'd like to reserve `messaging.message.` namespace for other possible per-message attributes.

It also adds the `messaging.batch_size` attribute which intends to:
- denote a bathing operation
- record number of messages sent/received/processed - since links can be dropped or there can be some other links not describing messages, link count is not a reliable indicator of batch size. Assuming a high-scale scenario, users might want to opt-out from per-message tracing, but then batch_size would still be useful

### DIFF
--- a/semantic_conventions/trace/messaging.yaml
+++ b/semantic_conventions/trace/messaging.yaml
@@ -103,7 +103,7 @@ groups:
         requirement_level:
           recommended: If different than `net.peer.name` and if `net.sock.peer.addr` is set.
       - ref: net.app.protocol.name
-        examples: ['AMQP', 'MQTT']
+        examples: ['amqp', 'mqtt']
       - ref: net.app.protocol.version
     constraints:
       - include: network
@@ -231,7 +231,7 @@ groups:
     brief: >
       Attributes for RabbitMQ
     attributes:
-      - id: routing_key
+      - id: message.routing_key
         type: string
         requirement_level:
           conditionally_required: If not empty.
@@ -267,12 +267,12 @@ groups:
         brief: >
           Client Id for the Consumer or Producer that is handling the message.
         examples: 'client-5'
-      - id: partition
+      - id: message.partition
         type: int
         brief: >
           Partition the message is sent to.
         examples: 2
-      - id: tombstone
+      - id: message.tombstone
         type: boolean
         requirement_level:
           conditionally_required: If value is `true`. When missing, the value is assumed to be `false`.
--- a/semantic_conventions/trace/messaging.yaml
+++ b/semantic_conventions/trace/messaging.yaml
@@ -1,4 +1,40 @@
 groups:
+  - id: messaging.message
+    prefix: messaging
+    type: span
+    brief: 'Semantic convention describing per-message attributes populated on messaging spans or links.'
+    attributes:
+      - id: destination.name
+        type: string
+        brief: 'The message destination name'
+        note: |
+          Destination name SHOULD uniquely identify specific queue, topic, or other entity within broker. If
+          broker does not have such notion, destination name SHOULD uniquely identify broker.
+
+          If destination name is the same for all messages being published, 
+          it MUST be set on corresponding publish, receive, process or other span.
+        examples: ['MyQueue', 'MyTopic']
+      - id: message.id
+        type: string
+        brief: 'A value used by the messaging system as an identifier for the message, represented as a string.'
+        examples: '452a7c7c7c7048c2f887f61572b18fc2'
+      - id: message.conversation_id
+        type: string
+        brief: >
+          The [conversation ID](#conversations) identifying the conversation to which the message belongs,
+          represented as a string. Sometimes called "Correlation ID".
+        examples: 'MyConversationId'
+      - id: message.payload_size_bytes
+        type: int
+        brief: >
+          The (uncompressed) size of the message payload in bytes.
+          Also use this attribute if it is unknown whether the compressed or uncompressed payload size is reported.
+        examples: 2738
+      - id: message.payload_compressed_size_bytes
+        type: int
+        brief: 'The compressed size of the message payload in bytes.'
+        examples: 2048
+
   - id: messaging
     prefix: messaging
     type: span
@@ -11,13 +47,6 @@ groups:
         requirement_level: required
         brief: 'A string identifying the messaging system.'
         examples: ['kafka', 'rabbitmq', 'rocketmq', 'activemq', 'AmazonSQS']
-      - id: destination
-        type: string
-        requirement_level: required
-        brief: >
-          The message destination name. This might be equal to the span name
-          but is required nevertheless.
-        examples: ['MyQueue', 'MyTopic']
       - id: destination_kind
         type:
           allow_custom_values: false
@@ -48,26 +77,31 @@ groups:
         type: string
         brief: 'Connection string.'
         examples: ['tibjmsnaming://localhost:7222', 'https://queue.amazonaws.com/80398EXAMPLE/MyQueue']
-      - id: message_id
-        type: string
-        brief: 'A value used by the messaging system as an identifier for the message, represented as a string.'
-        examples: '452a7c7c7c7048c2f887f61572b18fc2'
-      - id: conversation_id
-        type: string
-        brief: >
-          The [conversation ID](#conversations) identifying the conversation to which the message belongs,
-          represented as a string. Sometimes called "Correlation ID".
-        examples: 'MyConversationId'
-      - id: message_payload_size_bytes
+      - id: batch.size
         type: int
-        brief: >
-          The (uncompressed) size of the message payload in bytes.
-          Also use this attribute if it is unknown whether the compressed or uncompressed payload size is reported.
-        examples: 2738
-      - id: message_payload_compressed_size_bytes
-        type: int
-        brief: 'The compressed size of the message payload in bytes.'
-        examples: 2048
+        brief: The number of messages sent, received, or processed in the scope of the batching operation.
+        requirement_level:
+          conditionally_required: >
+            If available within the messaging system, and only if the span describes operations that operate with message batches.
+        note: >
+            Instrumentation SHOULD always set it on batch `receive` operations regardless
+            of the actual number of received messages (e.g. 0, 1, or more).
+        examples: [0, 1, 2]
+      - ref: messaging.destination.name
+        requirement_level:
+          conditionally_required: if available and if the value applies to all messages in the batch.
+      - ref: messaging.message.id
+        requirement_level:
+          recommended: Only if messaging.batch.size is not set or if the value applies to all messages in the batch.
+      - ref: messaging.message.conversation_id
+        requirement_level:
+          recommended: Only if messaging.batch.size is not set or if the value applies to all messages in the batch.
+      - ref: messaging.message.payload_size_bytes
+        requirement_level:
+          recommended: Only if messaging.batch.size is not set or if the value applies to all messages in the batch.
+      - ref: messaging.message.payload_compressed_size_bytes
+        requirement_level:
+          recommended: Only if messaging.batch.size is not set or if the value applies to all messages in the batch.
       - ref: net.peer.name
         note: >
           This should be the IP/hostname of the broker (or other network-level peer) this specific message is sent to/received from.
@@ -163,11 +197,11 @@ groups:
     brief: >
       Attributes for Apache Kafka
     attributes:
-      - id: message_key
+      - id: message.key
         type: string
         brief: >
           Message keys in Kafka are used for grouping alike messages to ensure they're processed on the same partition.
-          They differ from `messaging.message_id` in that they're not unique.
+          They differ from `messaging.message.id` in that they're not unique.
           If the key is `null`, the attribute MUST NOT be set.
         note: >
           If the key type is not string, it's string representation has to be supplied for the attribute.
@@ -220,28 +254,28 @@ groups:
         brief: >
           The unique identifier for each client.
         examples: 'myhost@8742@s8083jm'
-      - id: delivery_timestamp
+      - id: message.delivery_timestamp
         type: int
         requirement_level:
           conditionally_required: If the message type is delay and delay time level is not specified.
         brief: >
           The timestamp in milliseconds that the delay message is expected to be delivered to consumer.
         examples: 1665987217045
-      - id: delay_time_level
+      - id: message.delay_time_level
         type: int
         requirement_level:
           conditionally_required: If the message type is delay and delivery timestamp is not specified.
         brief: >
           The delay time level for delay message, which determines the message delay time.
         examples: 3
-      - id: message_group
+      - id: message.group
         type: string
         requirement_level:
           conditionally_required: If the message type is FIFO.
         brief: >
           It is essential for FIFO message. Messages that belong to the same message group are always processed one by one within the same consumer group.
         examples: 'myMessageGroup'
-      - id: message_type
+      - id: message.type
         type:
           allow_custom_values: false
           members:
@@ -259,12 +293,12 @@ groups:
               brief: 'Transaction message'
         brief: >
           Type of message.
-      - id: message_tag
+      - id: message.tag
         type: string
         brief: >
           The secondary classifier of message besides topic.
         examples: tagA
-      - id: message_keys
+      - id: message.keys
         type: string[]
         brief: >
           Key(s) of message, another way to mark message besides message id.
--- a/semantic_conventions/trace/messaging.yaml
+++ b/semantic_conventions/trace/messaging.yaml
@@ -71,20 +71,18 @@ groups:
         type: int
         brief: The number of messages sent, received, or processed in the scope of the batching operation.
         requirement_level:
-          conditionally_required: >
-            If available within the messaging system, and only if the span describes operations that operate with message batches.
+          conditionally_required: If the span describes operation on a batch of messages.
         note: >
-            Instrumentation MUST set it on batch `receive` operations regardless
-            of the actual number of received messages (e.g. 0, 1, or more).
-            Instrumentations SHOULD set `messaging.batch.size` on spans representing
-            other operations - when it's not set, it's assumed to be `1`.
+            Instrumentations SHOULD NOT set `messaging.batch.size` on spans that operate with a single message.
+            When client library supports batch and single-message API for the same operation, instrumentations SHOULD
+            use `messaging.batch.size` for batching APIs and SHOULD NOT use it for single-message APIs.
         examples: [0, 1, 2]
       - ref: messaging.message.id
         requirement_level:
           recommended: Only if span represents operation on a single message.
       - ref: messaging.message.conversation_id
         requirement_level:
-          recommended: Only if messaging.batch.size is not set or if the value applies to all messages in the batch.
+          recommended: Only if span represents operation on a single message.
       - ref: messaging.message.payload_size_bytes
         requirement_level:
           recommended: Only if span represents operation on a single message.
--- a/semantic_conventions/trace/messaging.yaml
+++ b/semantic_conventions/trace/messaging.yaml
@@ -11,9 +11,6 @@ groups:
         note: |
           Destination name SHOULD uniquely identify specific queue, topic, or other entity within broker. If
           broker does not have such notion, destination name SHOULD uniquely identify broker.
-
-          If destination name is the same for all messages being published,
-          it MUST be set on corresponding span.
         examples: ['MyQueue', 'MyTopic']
       - id: source.name
         type: string
@@ -21,9 +18,6 @@ groups:
         note: |
           Source name SHOULD uniquely identify specific queue, topic, or other entity within broker. If
           broker does not have such notion, source name SHOULD uniquely identify broker.
-
-          If source name is the same for all messages being received or processed,
-          it MUST be set on corresponding span.
         examples: ['MyQueue', 'MyTopic']
       - id: message.id
         type: string
@@ -58,10 +52,21 @@ groups:
         requirement_level: required
         brief: 'A string identifying the messaging system.'
         examples: ['kafka', 'rabbitmq', 'rocketmq', 'activemq', 'AmazonSQS']
-      - id: url
-        type: string
-        brief: 'Connection string.'
-        examples: ['tibjmsnaming://localhost:7222', 'https://queue.amazonaws.com/80398EXAMPLE/MyQueue']
+      - id: operation
+        type:
+          allow_custom_values: true
+          members:
+            - id: publish
+              value: "publish"
+            - id: receive
+              value: "receive"
+            - id: process
+              value: "process"
+        brief: >
+          A string identifying the kind of messaging operation as defined in the
+          [Operation names](#operation-names) section above.
+        requirement_level:
+          conditionally_required:
       - id: batch.size
         type: int
         brief: The number of messages sent, received, or processed in the scope of the batching operation.
@@ -167,19 +172,6 @@ groups:
     span_kind: consumer
     brief: 'Semantic convention for a consumer of messages received from a messaging system'
     attributes:
-      - id: operation
-        type:
-          allow_custom_values: false
-          members:
-            - id: receive
-              value: "receive"
-            - id: process
-              value: "process"
-        brief: >
-          A string identifying the kind of message consumption as defined in the
-          [Operation names](#operation-names) section above.
-          If the operation is "send", this attribute MUST NOT be set, since the
-          operation can be inferred from the span kind in that case.
       - id: consumer_id
         type: string
         brief: >
--- a/semantic_conventions/trace/messaging.yaml
+++ b/semantic_conventions/trace/messaging.yaml
@@ -1,6 +1,7 @@
 groups:
   - id: messaging.message
     prefix: messaging
+    # todo (https://github.com/open-telemetry/build-tools/issues/123)
     type: span
     brief: 'Semantic convention describing per-message attributes populated on messaging spans or links.'
     attributes:
@@ -11,8 +12,18 @@ groups:
           Destination name SHOULD uniquely identify specific queue, topic, or other entity within broker. If
           broker does not have such notion, destination name SHOULD uniquely identify broker.
 
-          If destination name is the same for all messages being published, 
-          it MUST be set on corresponding publish, receive, process or other span.
+          If destination name is the same for all messages being published,
+          it MUST be set on corresponding span.
+        examples: ['MyQueue', 'MyTopic']
+      - id: source.name
+        type: string
+        brief: 'The message source name'
+        note: |
+          Source name SHOULD uniquely identify specific queue, topic, or other entity within broker. If
+          broker does not have such notion, source name SHOULD uniquely identify broker.
+
+          If source name is the same for all messages being received or processed,
+          it MUST be set on corresponding span.
         examples: ['MyQueue', 'MyTopic']
       - id: message.id
         type: string
@@ -47,32 +58,6 @@ groups:
         requirement_level: required
         brief: 'A string identifying the messaging system.'
         examples: ['kafka', 'rabbitmq', 'rocketmq', 'activemq', 'AmazonSQS']
-      - id: destination_kind
-        type:
-          allow_custom_values: false
-          members:
-            - id: queue
-              value: "queue"
-              brief: "A message sent to a queue"
-            - id: topic
-              value: "topic"
-              brief: "A message sent to a topic"
-        requirement_level:
-          conditionally_required: If the message destination is either a `queue` or `topic`.
-        brief: 'The kind of message destination'
-      - id: temp_destination
-        type: boolean
-        requirement_level:
-          conditionally_required: If value is `true`. When missing, the value is assumed to be `false`.
-        brief: 'A boolean that is true if the message destination is temporary.'
-      - id: protocol
-        type: string
-        brief: 'The name of the transport protocol.'
-        examples: ['AMQP', 'MQTT']
-      - id: protocol_version
-        type: string
-        brief: 'The version of the transport protocol.'
-        examples: '0.9.1'
       - id: url
         type: string
         brief: 'Connection string.'
@@ -87,12 +72,7 @@ groups:
             Instrumentation SHOULD always set it on batch `receive` operations regardless
             of the actual number of received messages (e.g. 0, 1, or more).
         examples: [0, 1, 2]
-      - ref: messaging.destination.name
-        requirement_level:
-          conditionally_required: if available and if the value applies to all messages in the batch.
       - ref: messaging.message.id
-        requirement_level:
-          recommended: Only if messaging.batch.size is not set or if the value applies to all messages in the batch.
       - ref: messaging.message.conversation_id
         requirement_level:
           recommended: Only if messaging.batch.size is not set or if the value applies to all messages in the batch.
@@ -117,7 +97,9 @@ groups:
         tag: connection-level
         requirement_level:
           recommended: If different than `net.peer.name` and if `net.sock.peer.addr` is set.
-
+      - ref: net.app.protocol.name
+        examples: ['AMQP', 'MQTT']
+      - ref: net.app.protocol.version
     constraints:
       - include: network
 
@@ -127,6 +109,47 @@ groups:
     extends: messaging
     span_kind: producer
     brief: 'Semantic convention for producers of messages sent to a messaging systems.'
+    attributes:
+      - ref: messaging.destination.name
+        requirement_level:
+          conditionally_required: If the value applies to all messages in the batch.
+      - id: destination.kind
+        type:
+          allow_custom_values: false
+          members:
+            - id: queue
+              value: "queue"
+              brief: "A message sent to a queue"
+            - id: topic
+              value: "topic"
+              brief: "A message sent to a topic"
+        requirement_level:
+          conditionally_required: If the message destination is either a `queue` or `topic`.
+        brief: 'The kind of message destination'
+      - id: destination.template
+        type: string
+        requirement_level:
+          conditionally_required: >
+            If available. Instrumentations MUST NOT use `messaging.destination.name` as template
+            unless low-cardinality of destination name is guaranteed.
+        brief: Low cardinality field representing messaging destination
+        note: >
+          Destination names could be constructed from templates.
+          An example would be a destination name involving a user name or product id.
+          Although the destination name in this case is of high cardinality,
+          the underlying template is of low cardinality and can be effectively
+          used for grouping and searching spans.
+        examples: ['/customers/{customerId}']
+      - id: destination.temporary
+        type: boolean
+        requirement_level:
+          conditionally_required: If value is `true`. When missing, the value is assumed to be `false`.
+        brief: 'A boolean that is true if the message destination is temporary and might not exist anymore after messages are processed.'
+      - id: destination.anonymous
+        type: boolean
+        requirement_level:
+          conditionally_required: If value is `true`. When missing, the value is assumed to be `false`.
+        brief: 'A boolean that is true if the message destination is anonymous (could be unnamed or have auto-generated name).'
 
   - id: messaging.producer.synchronous
     prefix: messaging
@@ -165,6 +188,40 @@ groups:
           `messaging.kafka.consumer_group`. For brokers, such as RabbitMQ and Artemis, set it to the `client_id`
           of the client consuming the message.
         examples: 'mygroup - client-6'
+      - ref: messaging.source.name
+        requirement_level:
+          conditionally_required: If the value applies to all messages in the batch.
+      - id: source.kind
+        type:
+          allow_custom_values: true
+          members:
+            - id: queue
+              value: "queue"
+              brief: "A message received from a queue"
+            - id: topic
+              value: "topic"
+              brief: "A message received from a topic"
+        requirement_level:
+          conditionally_required: If the message source is either a `queue` or `topic`.
+        brief: 'The kind of message source'
+      - id: source.template
+        type: string
+        brief: 'Low cardinality field representing messaging source'
+        requirement_level:
+          conditionally_required: >
+            If available. Instrumentations MUST NOT use `messaging.source.name` as template
+            unless low-cardinality of source name is guaranteed.
+        examples: ['/customers/{customerId}']
+      - id: source.temporary
+        type: boolean
+        requirement_level:
+          recommended: when supported by messaging system and only if the source is temporary. If missing, assumed to be false.
+        brief: 'denotes that the source is a temporary source and might not exist anymore after messages are processed'
+      - id: source.anonymous
+        type: boolean
+        requirement_level:
+          recommended: when supported by messaging system and only if the source is anonymous. If missing, assumed to be false.
+        brief: 'denotes that the source is an anonymous source (could be unnamed or have auto-generated name)'
 
   - id: messaging.consumer.synchronous
     prefix: messaging
--- a/semantic_conventions/trace/messaging.yaml
+++ b/semantic_conventions/trace/messaging.yaml
@@ -62,11 +62,11 @@ groups:
               value: "receive"
             - id: process
               value: "process"
+        requirement_level: required
         brief: >
           A string identifying the kind of messaging operation as defined in the
           [Operation names](#operation-names) section above.
-        requirement_level:
-          conditionally_required:
+        note: If custom value is used, it MUST known to be of low cardinality.
       - id: batch.size
         type: int
         brief: The number of messages sent, received, or processed in the scope of the batching operation.
--- a/semantic_conventions/trace/messaging.yaml
+++ b/semantic_conventions/trace/messaging.yaml
@@ -74,19 +74,23 @@ groups:
           conditionally_required: >
             If available within the messaging system, and only if the span describes operations that operate with message batches.
         note: >
-            Instrumentation SHOULD always set it on batch `receive` operations regardless
+            Instrumentation MUST set it on batch `receive` operations regardless
             of the actual number of received messages (e.g. 0, 1, or more).
+            Instrumentations SHOULD set `messaging.batch.size` on spans representing
+            other operations - when it's not set, it's assumed to be `1`.
         examples: [0, 1, 2]
       - ref: messaging.message.id
+        requirement_level:
+          recommended: Only if span represents operation on a single message.
       - ref: messaging.message.conversation_id
         requirement_level:
           recommended: Only if messaging.batch.size is not set or if the value applies to all messages in the batch.
       - ref: messaging.message.payload_size_bytes
         requirement_level:
-          recommended: Only if messaging.batch.size is not set or if the value applies to all messages in the batch.
+          recommended: Only if span represents operation on a single message.
       - ref: messaging.message.payload_compressed_size_bytes
         requirement_level:
-          recommended: Only if messaging.batch.size is not set or if the value applies to all messages in the batch.
+          recommended: Only if span represents operation on a single message.
       - ref: net.peer.name
         note: >
           This should be the IP/hostname of the broker (or other network-level peer) this specific message is sent to/received from.
@@ -117,7 +121,7 @@ groups:
     attributes:
       - ref: messaging.destination.name
         requirement_level:
-          conditionally_required: If the value applies to all messages in the batch.
+          conditionally_required: If one message is being published or if the value applies to all messages in the batch.
       - id: destination.kind
         type:
           allow_custom_values: false
@@ -143,7 +147,7 @@ groups:
           An example would be a destination name involving a user name or product id.
           Although the destination name in this case is of high cardinality,
           the underlying template is of low cardinality and can be effectively
-          used for grouping and searching spans.
+          used for grouping and aggregation.
         examples: ['/customers/{customerId}']
       - id: destination.temporary
         type: boolean
@@ -155,7 +159,6 @@ groups:
         requirement_level:
           conditionally_required: If value is `true`. When missing, the value is assumed to be `false`.
         brief: 'A boolean that is true if the message destination is anonymous (could be unnamed or have auto-generated name).'
-
   - id: messaging.producer.synchronous
     prefix: messaging
     type: span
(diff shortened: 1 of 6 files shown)


blumamir (commented):
Thank you @lmolkova

specification/trace/semantic_conventions/messaging.md
```diff
@@ -196,6 +204,22 @@ For batch receiving and processing (see the [Batch receiving](#batch-receiving)
 Even though in that case one might think that the processing span's kind should be `INTERNAL`, that kind MUST NOT be used.
 Instead span kind should be set to either `CONSUMER` or `SERVER` according to the rules defined above.
 
+### Per-message attributes
+
+Note that each or messaging operations (`send`, `receive`, or `process`) can describe a batch of messages. For batch operations per-message attributes cannot be set on
+corresponding span and SHOULD instead be set on link. See [Batch Receiving](#batch-receiving) and [Batch Processing](#batch-processing) for more information on correlation using links.
```
blumamir, line 210: It is implied from the SHOULD wording, but I wonder if it is appropriate to document here that instrumentation is allowed to not create some or all links for reasons of volumes/costs or if the information is not available

For example, `message.id` is not always defined for all systems (redis pub-sub, sockeit.io).
Some attributes can be present but supply no value to the user in some cases - `messaging.message.payload_size_bytes` can be recorded but it can be considered noise for many users.
Maybe some attributes are simply too expensive or complicated to capture.

If instrumentation doesn't need or chooses not to store anything on links attributes, they may be dropped entirely IMO.

blumamir, line 210: Creating a link requires a context to set as the link target. This context might not always be available:
- producer is not instrumented
- no appropriate mechanism to propagate context in a specific system (redis pub-sub, socket.io for example)
- intermediaries and message routing mechanisms which drop the context.

However, using links is the only way to record per message information in batch operations, when might be desired even when context is not available.

Do you think we should also describe how to generate links in the absence of creation context? we discussed using an invalid (zeroed) context for that if I remember correctly. We can also add it in a future PR but I think it should eventually be speced.

blumamir, line 210: I think it is also appropriate to state here that each message should be captured as link only once and not more.
In the absence of clear systematic rules of how to differentiate specific messages by attributes, and the fact that links can be dropped, It will not be possible to tell if a span with `messaging.batch.size` of 2 and two links contains one link per message, or zero link for the first message and two for the seconds.

Constraining the messages to be recorded at most once as links, will allow UIs to derive a meaningful set of unique messages from links with no ambiguaty.

Oberon00: > Do you think we should also describe how to generate links in the absence of creation context?

I don't think links are ideal then, maybe use events instead?

> Do you think we should also describe how to generate links in the absence of creation context? we discussed using an invalid (zeroed) context for that if I remember correctly.

I don't think SDKs (consistently?) support creating invalid links right now (tangentially related #753).

And also: How should this be handled on the sending side? A link (necessarily) always points to a parent, so I don't think it even makes sense semantically. This feels like a very important issue/inconsistency to me. Should span events be used instead?

blumamir: The problem with span events is that they don't capture a context, thus making it impossible to correlate a single message flow when producer and consumer are not in the same trace (or even when they are in the same trace sometimes?)

Links offer both the silo to store per-message attributes, and a way to "jump" from consumer to producer.

Oberon00: I was talking about using events for where you don't have a context you can capture.

Oberon00: I agree, it's not nice at all. And events don't really make sense semantically either, but IMHO more than links on the outgoing side

specification/trace/semantic_conventions/messaging.md
```diff
@@ -196,6 +204,22 @@ For batch receiving and processing (see the [Batch receiving](#batch-receiving)
...
+
+Following attributes apply to links describing each message.
```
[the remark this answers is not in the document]
joaopgrassi: Somewhat related but maybe not: Should we remove the message-specific attributes from the general table and only have them here? Not sure but it might confuse people on why they are duplicated.

specification/trace/semantic_conventions/messaging.md
```diff
@@ -381,7 +405,9 @@ Process C:                              | Span Recv1 | Span Recv2 |
 |-|-|-|-|-|-|
 | Span name | `"Q send"` | `"Q send"` | `"Q receive"` | `"Q receive"` | `"Q process"` |
 | Parent |  |  | Span Prod1 | Span Prod2 |  |
-| Links |  |  |  |  | Span Prod1 + Prod2 |
+| Links |  |  |  |  | [Span Prod1, Span Prod2 ] |
+| Link attributes |  |  |  |  | Span Prod1: `messaging.message.id`: `"a1"`  |
```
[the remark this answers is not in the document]
lmolkova: I'm as surprised as you, but current spec  does not use links on receive spans. The idea it has is:
- Receive  batch has ho links
- Receive process spans are its children and there is process span per child
- Process spans have links to message context

I.e. receive is still correlated to message, but via link on the child.

specification/trace/semantic_conventions/messaging.md
```diff
@@ -196,6 +204,22 @@ For batch receiving and processing (see the [Batch receiving](#batch-receiving)
 Even though in that case one might think that the processing span's kind should be `INTERNAL`, that kind MUST NOT be used.
 Instead span kind should be set to either `CONSUMER` or `SERVER` according to the rules defined above.
 
+### Per-message attributes
+
+Note that each or messaging operations (`send`, `receive`, or `process`) can describe a batch of messages. For batch operations per-message attributes cannot be set on
```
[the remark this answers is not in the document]
lmolkova: Let's discuss it tomorrow. It'd be great to clarify recommendations for the non-batching case.  I think it'd be overkill to use links for rabbit or other single-send systems. Also requiring attributes on links means that this spec will stop working for all backends.

specification/trace/semantic_conventions/messaging.md
lmolkova: all of the message attributes come with `recommended` requirement level which [explicitly](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-requirement-level.md#recommended) mentions they could be disabled.

lmolkova: I want to refrain from changing the current state of affairs in this PR. Current spec already uses links and not events and does not deal with the edge cases (missing context on the message). We'll nee to solve it in scope of all messaging work. but let's not do it in this PR and keep status quo for now.

semantic_conventions/trace/messaging.yaml
```diff
@@ -48,26 +74,29 @@ groups:
...
-          Also use this attribute if it is unknown whether the compressed or uncompressed payload size is reported.
-        examples: 2738
-      - id: message_payload_compressed_size_bytes
-        type: int
-        brief: 'The compressed size of the message payload in bytes.'
-        examples: 2048
+        brief: Number of messages sent, received, or processed in scope of this operation.
+        requirement_level:
+          conditionally_required: >
+            If available and only if the span describes operations that operate with message batches. It SHOULD NOT be set when
```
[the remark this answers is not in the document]
lmolkova: I figured `only if` is enough. `conditionally_required` implies MUST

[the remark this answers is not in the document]
lmolkova: the same `only if` is enough, the second statement is redundunt

specification/trace/semantic_conventions/messaging.md
```diff
@@ -125,6 +126,8 @@ The following operations related to messages are defined for these semantic conv
 
 ## Messaging attributes
 
+Following messaging attributes apply to spans describing `send`, `receive`, and `process`  operations.
```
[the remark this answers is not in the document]
lmolkova: removed the sentence, as it does not clarify anything.

specification/trace/semantic_conventions/messaging.md
lmolkova: I ended up changing this to 

>  For batch operations per-message attributes cannot be set on the corresponding span and MAY instead be set on a link

this way events can also be used. 

Once we finalize the consensus in messaging SIG on span structure, we'll bring it here via different PR.

Oberon00: > Current spec already uses links and not events and does not deal with the edge cases (missing context on the message).

The current spec doesn't use attributes on links, so I always interpreted that as creating links only for the messages you have context.
I don't think an incoming message without context is an edge case, it seems perfectly normal to me.

specification/trace/semantic_conventions/messaging.md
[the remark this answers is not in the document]
joaopgrassi: I will add to the agenda for today's meeting.

specification/trace/semantic_conventions/messaging.md
lmolkova: > The current spec doesn't use attributes on links, so I always interpreted that as creating links only for messages with context.

Right, but the current spec already uses links (and not events). The only thing I want to make with this change is to allow per-message attributes to appear on links.
This change makes current spec limitations more obvious but does not change the status quo.

We'll have another PR on when and how links are populated separately. We can discuss events/links/empty context there

specification/trace/semantic_conventions/messaging.md
```diff
@@ -135,27 +136,40 @@ The following operations related to messages are defined for these semantic conv
...
 
 **[1]:** If the message destination is either a `queue` or `topic`.
 
 **[2]:** If value is `true`. When missing, the value is assumed to be `false`.
 
-**[3]:** This should be the IP/hostname of the broker (or other network-level peer) this specific message is sent to/received from.
+**[3]:** Instrumentation SHOULD always set it on batch `receive` operations regardless of the actual number of received messages (e.g. 0, 1, or more).
 
-**[4]:** If different than `inet` and if any of `net.sock.peer.addr` or `net.sock.host.addr` are set. Consumers of telemetry SHOULD accept both IPv4 and IPv6 formats for the address in `net.sock.peer.addr` if `net.sock.family` is not set. This is to support instrumentations that follow previous versions of this document.
+**[4]:** If available and only if the span describes operations that operate with message batches.
```
dpauls, line 156:
```suggestion
**[4]:** If available within the messaging system, and only if the span describes operations that operate with message batches.
```
The reason for this change is that before it looked like the intent was to say "Available if and only if", except a mistake was made. I think this makes it clearer.

specification/trace/semantic_conventions/messaging.md
```diff
@@ -135,27 +136,40 @@ The following operations related to messages are defined for these semantic conv
...
-| [`net.sock.peer.port`](span-general.md) | int | Remote socket peer port. | `16456` | Recommended: [6] |
+| [`net.sock.peer.name`](span-general.md) | string | Remote socket peer name. | `proxy.example.com` | Recommended: [11] |
+| [`net.sock.peer.port`](span-general.md) | int | Remote socket peer port. | `16456` | Recommended: [12] |
 
 **[1]:** If the message destination is either a `queue` or `topic`.
 
 **[2]:** If value is `true`. When missing, the value is assumed to be `false`.
 
-**[3]:** This should be the IP/hostname of the broker (or other network-level peer) this specific message is sent to/received from.
+**[3]:** Instrumentation SHOULD always set it on batch `receive` operations regardless of the actual number of received messages (e.g. 0, 1, or more).
```
dpauls, line 154: Is this note needed? I feel like note 4 below covers this. Perhaps if the desire is to express "SHOULD", this could be added at the beginning of note 4?

specification/trace/semantic_conventions/messaging.md
```diff
@@ -135,27 +136,40 @@ The following operations related to messages are defined for these semantic conv
...
 
-**[4]:** If different than `inet` and if any of `net.sock.peer.addr` or `net.sock.host.addr` are set. Consumers of telemetry SHOULD accept both IPv4 and IPv6 formats for the address in `net.sock.peer.addr` if `net.sock.family` is not set. This is to support instrumentations that follow previous versions of this document.
+**[4]:** If available and only if the span describes operations that operate with message batches.
 
-**[5]:** If different than `net.peer.name` and if `net.sock.peer.addr` is set.
+**[5]:** only if messaging.batch.size is not set or set to `1`.
```
dpauls, line 158:
```suggestion
**[5]:** Only if messaging.batch.size is not set or if the value applies to all messages in the batch.
```
The purpose for this suggestion is to leave the door open to the possibility all messages in a batch could be part of the same conversation. Regardless of whether any existing messaging systems behave this way, I see no reason to insist the batch size needs to be 1 for this to apply to a batch.

specification/trace/semantic_conventions/messaging.md
```diff
@@ -135,27 +136,40 @@ The following operations related to messages are defined for these semantic conv
...
 
-**[6]:** If defined for the address family and if different than `net.peer.port` and if `net.sock.peer.addr` is set.
+**[6]:** only if messaging.batch.size is not set or set to `1`.
+
+**[7]:** only if messaging.batch.size is not set or set to `1`.
+
+**[8]:** only if messaging.batch.size is not set or set to `1`.
+
```
dpauls, line 165: Can note 5 be reused for all of these?

specification/trace/semantic_conventions/messaging.md
lmolkova: This note describes `receive` specifically, while  note above applies to all possible messaging operations.
Receive is special because you can have 0 messages received and also it's common to request N messages and potentially receive M, so defaulting to 1 does not make sense. 

This is also a recommendation, while `conditionally_required` level on note [4] implies MUST.

specification/trace/semantic_conventions/messaging.md
lmolkova: The source of truth is messaging.yaml file, this notes are auto-generated and I feel it's important to have conditions specified on each attribute.

joaopgrassi:
Not stale

semantic_conventions/trace/messaging.yaml
```diff
@@ -1,4 +1,40 @@
...
+  - id: messaging.message
+    prefix: messaging
+    type: span
+    brief: 'Semantic convention describing per-message attributes populated on messaging spans or links.'
+    attributes:
+      - id: destination.name
+        type: string
+        brief: 'The message destination name'
+        note: |
+          Destination name SHOULD uniquely identify specific queue, topic, or other entity within broker. If
```
joaopgrassi, line 11:
```suggestion
          The destination name SHOULD uniquely identify a queue, topic, or other entity within a broker. If
```

semantic_conventions/trace/messaging.yaml
```diff
@@ -1,4 +1,40 @@
...
+          broker does not have such notion, destination name SHOULD uniquely identify broker.
```
joaopgrassi, line 12:
```suggestion
          the broker does not have such notion, the destination name SHOULD uniquely identify the broker.
```

semantic_conventions/trace/messaging.yaml
```diff
@@ -1,4 +1,40 @@
...
+
+          If destination name is the same for all messages being published,
```
joaopgrassi, line 14:
```suggestion
          If the destination name is the same for all messages being published, 
```

semantic_conventions/trace/messaging.yaml
```diff
@@ -1,4 +1,40 @@
...
+    attributes:
+      - id: destination.name
+        type: string
+        brief: 'The message destination name'
+        note: |
+          Destination name SHOULD uniquely identify specific queue, topic, or other entity within broker. If
+          broker does not have such notion, destination name SHOULD uniquely identify broker.
+
+          If destination name is the same for all messages being published, 
+          it MUST be set on corresponding publish, receive, process or other span.
```
joaopgrassi, line 15: Do we need to have `or other span` here? If so, maybe something like: 

```suggestion
          it MUST be set on the corresponding publish, receive and process span. It MAY also be set on other spans related to any of aforementioned operations.
```

semantic_conventions/trace/messaging.yaml
```diff
@@ -48,26 +77,31 @@ groups:
...
-        examples: 2738
-      - id: message_payload_compressed_size_bytes
+      - id: batch.size
         type: int
-        brief: 'The compressed size of the message payload in bytes.'
-        examples: 2048
+        brief: The number of messages sent, received, or processed in the scope of the batching operation.
+        requirement_level:
+          conditionally_required: >
+            If available within the messaging system, and only if the span describes operations that operate with message batches.
```
joaopgrassi, line 85:
```suggestion
            If available within the messaging system, and only if the span describes operations that involve a batch of messages.
```
Maybe? I mostly wanted to avoid:  "operations that operate"

specification/trace/semantic_conventions/messaging.md
```diff
@@ -169,6 +191,8 @@ Additionally `net.peer.port` from the [network attributes][] is recommended.
 Furthermore, it is strongly recommended to add the [`net.transport`][] attribute and follow its guidelines, especially for in-process queueing systems (like [Hangfire][], for example).
 These attributes should be set to the broker to which the message is sent/from which it is received.
 
+Note that attributes in `messaging.message` namespace describe message. It's recommended to use `messaging.{system}.message` namespace for specific messaging system attributes that describe individual messages.
```
joaopgrassi, line 194: >namespace describe message

I'm a bit confused with this sentence. Should it also say?

>namespace describe an individual message.

specification/trace/semantic_conventions/messaging.md
```diff
@@ -196,6 +220,28 @@ For batch receiving and processing (see the [Batch receiving](#batch-receiving)
 Even though in that case one might think that the processing span's kind should be `INTERNAL`, that kind MUST NOT be used.
 Instead span kind should be set to either `CONSUMER` or `SERVER` according to the rules defined above.
 
+### Per-message attributes
+
+All messaging operations (`send`, `receive`, `process`, or other not covered by this specification) can describe a batch of messages. For batch operations per-message attributes cannot be set on the corresponding span and MAY instead be set on a link. See [Batch Receiving](#batch-receiving) and [Batch Processing](#batch-processing) for more information on correlation using links.
+
+Following attributes apply to links describing each message.
```
joaopgrassi, line 227:
```suggestion
The following attributes apply to links describing each message on a batch.
```
We want to say the following attributes apply to links when working with a batch right? Or regardless? I think we didn't reach a decision if we will _always_ record message-specific attributes on a link, regardless if it's a batch or not. Maybe we should pin this down, so we know how to phrase things here.

joaopgrassi:
Not stale

specification/trace/semantic_conventions/messaging.md
```diff
@@ -390,4 +438,5 @@ Process C:                              | Span Recv1 | Span Recv2 |
 | `messaging.destination` | `"Q"` | `"Q"` | `"Q"` | `"Q"` | `"Q"` |
```
pyohannes, line 438: We probably also need to fix attribute names throughout the examples (like changing this line to `messaging.destination.name`).

specification/trace/semantic_conventions/messaging.md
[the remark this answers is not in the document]
pyohannes: The idea to put message specific attributes on links works for the new span structure we worked out in the messaging workgroup, however, it will not fit with some of the examples in this spec (as there are cases like this one, where there are no links to hold attributes).

specification/trace/semantic_conventions/messaging.md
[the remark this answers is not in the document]
lmolkova: discussed on SIG meeting: don't be prescriptive for now, clarify links in future PRs along with link/event choice

specification/trace/semantic_conventions/messaging.md
lmolkova: Discussed at SIG meeting:
- `messaging.{system}.message.attr` is preferable to keep all system-specific attr in the same namespace
- `{system}` should match exactly `messaging.system`
- we should also describe namespace pattern for messaging.destination and source.
- Perhaps we should standardize it across specs in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-naming.md

specification/trace/semantic_conventions/messaging.md
lmolkova: We discussed attributes on span vs links on Thu meeting and decided to not be prescriptive in this PR just yet. So I rephrased it, LMK what you think

semantic_conventions/trace/messaging.yaml
lmolkova: removed the sentence since we decided to address attribute placement (on spans or links) in a separate PR.

specification/trace/semantic_conventions/messaging.md
lmolkova: * allow jms and amqp and specific system on the span
* messaging.system vs messaging.{system}.message needs further discussion, but keep it open for now.

pyohannes: Sorry, I missed the discussion about multiple messaging system combinations last week. My take on this:
* AMQP qualifies more than a protocol than a system. So, you might have a `messaging.system` set to `kafka`, and `net.app.protocol.name` set to `amqp`.
* There's the case when AMQP is used without any messaging system on top. I think at one point there should be semantic conventions specific to AMQP that (probably) prescribe to dedicated spans for AMQP operations and can be combined with spans prescribed by these conventions. AMQP semantic conventions would model the transport layer.
* I think if messaging systems are used on top of each other (e. g. JMS and RabbitMQ) each of those will create spans independently. This causes other problems (@lmolkova treats them in https://github.com/open-telemetry/oteps/pull/172), however, I think it's the only reasonable way to model this reliable.

specification/trace/semantic_conventions/messaging.md
```diff
@@ -189,64 +192,130 @@ The following operations related to messages are defined for these semantic conv
...
+**[3]:** If available. Instrumentations MUST NOT use `messaging.destination.name` as template unless low-cardinality of destination name is guaranteed.
+
+**[4]:** If value is `true`. When missing, the value is assumed to be `false`.
+
+**[5]:** If value is `true`. When missing, the value is assumed to be `false`.
+
+**[6]:** Destination name SHOULD uniquely identify specific queue, topic, or other entity within broker. If
+broker does not have such notion, destination name SHOULD uniquely identify broker.
+
+**[7]:** If the value applies to all messages in the batch.
```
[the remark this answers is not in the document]
blumamir: I know it might be a bit too specific, but for a batch of messages the value can be the same by protocol (reading a batch of messages from a single queue), or it can be that we read a few arbitrary messages and they happen to have the same destination at a single invocation.
I think this attribute is required for the first case (same destination by definition), and SHOULD NOT be set for the second case.
WDYT?

specification/trace/semantic_conventions/messaging.md
```diff
@@ -189,64 +192,130 @@ The following operations related to messages are defined for these semantic conv
...
+contains attributes that describe the logical entity messages are published to, and `messaging.source` describes
+logical entity messages are received from.
+
 [network attributes]: span-general.md#general-network-connection-attributes
 [`net.transport`]: span-general.md#network-transport-attributes
 [Hangfire]: https://www.hangfire.io/
 
+### Producer attributes
+
+For message producers, the following additional attributes may be set:
```
blumamir, line 252: Here we say "**may** be set", but it's a bit confusing, as the required level for these attributes is not MAY.
They are not optional, they should be set if available.
Maybe replace the word may with something else?

specification/trace/semantic_conventions/messaging.md
```diff
@@ -189,64 +192,130 @@ The following operations related to messages are defined for these semantic conv
...
+
 For message consumers, the following additional attributes may be set:
 
 <!-- semconv messaging.consumer -->
 | Attribute  | Type | Description  | Examples  | Requirement Level |
 |---|---|---|---|---|
-| `messaging.operation` | string | A string identifying the kind of message consumption as defined in the [Operation names](#operation-names) section above. If the operation is "send", this attribute MUST NOT be set, since the operation can be inferred from the span kind in that case. | `receive` | Recommended |
 | `messaging.consumer_id` | string | The identifier for the consumer receiving a message. For Kafka, set it to `{messaging.kafka.consumer_group} - {messaging.kafka.client_id}`, if both are present, or only `messaging.kafka.consumer_group`. For brokers, such as RabbitMQ and Artemis, set it to the `client_id` of the client consuming the message. | `mygroup - client-6` | Recommended |
+| `messaging.source.kind` | string | The kind of message source | `queue` | Conditionally Required: [1] |
+| `messaging.source.template` | string | Low cardinality field representing messaging source | `/customers/{customerId}` | Conditionally Required: [2] |
```
blumamir, line 295: for destination, we have this text describing what it means:

> Destination names could be constructed from templates. An example would be a destination name involving a user name or product id. Although the destination name in this case is of high cardinality, the underlying template is of low cardinality and can be effectively used for grouping and searching spans.

Maybe add it here so a user that landed on this section has the context as well? or extract to a common section and link both to the description?

specification/trace/semantic_conventions/messaging.md
```diff
@@ -189,64 +192,130 @@ The following operations related to messages are defined for these semantic conv
...
+| `messaging.source.temporary` | boolean | denotes that the source is a temporary source and might not exist anymore after messages are processed |  | Recommended: [3] |
+| `messaging.source.anonymous` | boolean | denotes that the source is an anonymous source (could be unnamed or have auto-generated name) |  | Recommended: [4] |
```
blumamir, line 297: for destination we wrote:

> A boolean that is true if the message destination is temporary and might not exist anymore after messages are processed.

maybe we should align both descriptions for consistency?

specification/trace/semantic_conventions/messaging.md
```diff
@@ -256,8 +325,37 @@ For batch receiving and processing (see the [Batch receiving](#batch-receiving)
 Even though in that case one might think that the processing span's kind should be `INTERNAL`, that kind MUST NOT be used.
 Instead span kind should be set to either `CONSUMER` or `SERVER` according to the rules defined above.
 
+### Per-message attributes
+
+All messaging operations (`publish`, `receive`, `process`, or other not covered by this specification) can describe a batch of messages.
+
+The following attributes describe individual messages. For batch operations per-message attributes cannot be set on the corresponding unless they are the same for all batch and MAY instead be set on a link. See [Batch Receiving](#batch-receiving) and [Batch Processing](#batch-processing) for more information on correlation using links.
```
blumamir, line 332:
```suggestion
The following attributes describe individual messages. For batch operations per-message attributes cannot be set on the corresponding span attributes unless they are the same for all messages in the batch and MAY instead be set on a link. See [Batch Receiving](#batch-receiving) and [Batch Processing](#batch-processing) for more information on correlation using links.
```

specification/trace/semantic_conventions/messaging.md
```diff
@@ -256,8 +325,37 @@ For batch receiving and processing (see the [Batch receiving](#batch-receiving)
...
+| `messaging.message.conversation_id` | string | The [conversation ID](#conversations) identifying the conversation to which the message belongs, represented as a string. Sometimes called "Correlation ID". | `MyConversationId` | Recommended |
+| `messaging.message.payload_size_bytes` | int | The (uncompressed) size of the message payload in bytes. Also use this attribute if it is unknown whether the compressed or uncompressed payload size is reported. | `2738` | Recommended |
+| `messaging.message.payload_compressed_size_bytes` | int | The compressed size of the message payload in bytes. | `2048` | Recommended |
+
+**[1]:** Destination name SHOULD uniquely identify specific queue, topic, or other entity within broker. If
+broker does not have such notion, destination name SHOULD uniquely identify broker.
+
+**[2]:** Source name SHOULD uniquely identify specific queue, topic, or other entity within broker. If
+broker does not have such notion, source name SHOULD uniquely identify broker.
+<!-- endsemconv -->
```
blumamir, line 349: Do we have to repeat the table here? 
Seems a bit confusing, as these attributes are now specified twice in the document.

Also, some attributes never makes sense in a batch or need clarification on their meaning for batch (`messaging.message.id`, `messaging.message.payload_size_bytes`, `messaging.message.payload_compressed_size_bytes`)

specification/trace/semantic_conventions/messaging.md
```diff
@@ -256,8 +325,37 @@ For batch receiving and processing (see the [Batch receiving](#batch-receiving)
...
+
 ### Attributes specific to certain messaging systems
 
+All attributes that are specific for a messaging system SHOULD be populated in `messaging.{system}` namespace. Message-, destination-, or source-specific attributes SHOULD be populated under corresponding namespace:
```
blumamir, line 353: What is the meaning of the hyphen? `Message-`, `destination-`?

specification/trace/semantic_conventions/messaging.md
```diff
@@ -256,8 +325,37 @@ For batch receiving and processing (see the [Batch receiving](#batch-receiving)
...
+
+* Message-specific under `messaging.{system}.message`
+* Destination-specific under `messaging.{system}.destination`
+* source-specific under `messaging.{system}.source`
```
blumamir, line 357: These sentences are not consistent with the capital first letter.
Since they are found in attribute names in lower case, I suggest making them all lowercase.

specification/trace/semantic_conventions/messaging.md
```diff
@@ -276,7 +374,7 @@ For Apache Kafka, the following additional attributes are defined:
 <!-- semconv messaging.kafka -->
 | Attribute  | Type | Description  | Examples  | Requirement Level |
 |---|---|---|---|---|
-| `messaging.kafka.message_key` | string | Message keys in Kafka are used for grouping alike messages to ensure they're processed on the same partition. They differ from `messaging.message_id` in that they're not unique. If the key is `null`, the attribute MUST NOT be set. [1] | `myKey` | Recommended |
+| `messaging.kafka.message.key` | string | Message keys in Kafka are used for grouping alike messages to ensure they're processed on the same partition. They differ from `messaging.message.id` in that they're not unique. If the key is `null`, the attribute MUST NOT be set. [1] | `myKey` | Recommended |
 | `messaging.kafka.consumer_group` | string | Name of the Kafka Consumer Group that is handling the message. Only applies to consumers, not producers. | `my-group` | Recommended |
```
blumamir, line 378: Should this be changed to `messaging.kafka.source.consumer_group`?

specification/trace/semantic_conventions/messaging.md
```diff
@@ -276,7 +374,7 @@ For Apache Kafka, the following additional attributes are defined:
...
 | `messaging.kafka.client_id` | string | Client Id for the Consumer or Producer that is handling the message. | `client-5` | Recommended |
 | `messaging.kafka.partition` | int | Partition the message is sent to. | `2` | Recommended |
```
blumamir, line 380: I wonder if the partition should be hosted under a namespace 🤔 (`source`, `destination`)

specification/trace/semantic_conventions/messaging.md
```diff
@@ -189,64 +192,130 @@ The following operations related to messages are defined for these semantic conv
...
+| `messaging.batch.size` | int | The number of messages sent, received, or processed in the scope of the batching operation. [1] | `0`; `1`; `2` | Conditionally Required: [2] |
+| `messaging.message.conversation_id` | string | The [conversation ID](#conversations) identifying the conversation to which the message belongs, represented as a string. Sometimes called "Correlation ID". | `MyConversationId` | Recommended: [3] |
+| `messaging.message.id` | string | A value used by the messaging system as an identifier for the message, represented as a string. | `452a7c7c7c7048c2f887f61572b18fc2` | Recommended |
+| `messaging.message.payload_compressed_size_bytes` | int | The compressed size of the message payload in bytes. | `2048` | Recommended: [4] |
+| `messaging.message.payload_size_bytes` | int | The (uncompressed) size of the message payload in bytes. Also use this attribute if it is unknown whether the compressed or uncompressed payload size is reported. | `2738` | Recommended: [5] |
+| [`net.app.protocol.name`](span-general.md) | string | Application layer protocol used. The value SHOULD be normalized to lowercase. | `AMQP`; `MQTT` | Recommended |
```
ppatierno, line 201: The value should be normalized to lowercase but then the examples are uppercase. Should we use lowercase in the examples as well? To reinforce that the should be that way.

specification/trace/semantic_conventions/messaging.md
```diff
@@ -256,8 +325,37 @@ For batch receiving and processing (see the [Batch receiving](#batch-receiving)
...
 
+All attributes that are specific for a messaging system SHOULD be populated in `messaging.{system}` namespace. Message-, destination-, or source-specific attributes SHOULD be populated under corresponding namespace:
+
+* Message-specific under `messaging.{system}.message`
+* Destination-specific under `messaging.{system}.destination`
+* source-specific under `messaging.{system}.source`
+
 #### RabbitMQ
 
 In RabbitMQ, the destination is defined by an *exchange* and a *routing key*.
```
pyohannes, line 361: In the line below, `messaging.destination` should be changed to `messaging.destination.name`.

specification/trace/semantic_conventions/messaging.md
pyohannes:
```suggestion
All attributes that are specific for a messaging system SHOULD be populated in `messaging.{system}` namespace. Attributes that are specific to the message, the message destination, or the message source SHOULD be populated under corresponding namespace:
```

specification/trace/semantic_conventions/messaging.md
```diff
@@ -189,64 +192,130 @@ The following operations related to messages are defined for these semantic conv
...
-| [`net.peer.name`](span-general.md) | string | Logical remote hostname, see note below. [3] | `example.com` | Conditionally Required: If available. |
-| [`net.sock.family`](span-general.md) | string | Protocol [address family](https://man7.org/linux/man-pages/man7/address_families.7.html) which is used for communication. | `inet6`; `bluetooth` | Conditionally Required: [4] |
+| `messaging.operation` | string | A string identifying the kind of messaging operation as defined in the [Operation names](#operation-names) section above. | `publish` | Required |
+| `messaging.batch.size` | int | The number of messages sent, received, or processed in the scope of the batching operation. [1] | `0`; `1`; `2` | Conditionally Required: [2] |
+| `messaging.message.conversation_id` | string | The [conversation ID](#conversations) identifying the conversation to which the message belongs, represented as a string. Sometimes called "Correlation ID". | `MyConversationId` | Recommended: [3] |
+| `messaging.message.id` | string | A value used by the messaging system as an identifier for the message, represented as a string. | `452a7c7c7c7048c2f887f61572b18fc2` | Recommended |
```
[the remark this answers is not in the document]
lmolkova: I think we have to keep yml definition full and correct. We need to explain which per-message attributes are common (and can appear on every span), which of them apply only to publish/receive/etc.

specification/trace/semantic_conventions/messaging.md
lmolkova: moved both to `messaging.kafka.message` namespace and also did the same for `messaging.rabbitmq.message.routing_key` as it's message-specific

specification/trace/semantic_conventions/messaging.md
```diff
@@ -189,64 +192,130 @@ The following operations related to messages are defined for these semantic conv
...
+| `publish` | publish |
+| `receive` | receive |
+| `process` | process |
 <!-- endsemconv -->
 
 Additionally `net.peer.port` from the [network attributes][] is recommended.
 Furthermore, it is strongly recommended to add the [`net.transport`][] attribute and follow its guidelines, especially for in-process queueing systems (like [Hangfire][], for example).
 These attributes should be set to the broker to which the message is sent/from which it is received.
 
+Note that attributes in `messaging.message` namespace describe an individual message, `messaging.destination` namespace
```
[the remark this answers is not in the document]
lmolkova: do you mean that `messaging.message.conversation_id` can be used to describe the batch? I believe with your notes above, it's the only attribute in `message` namespace that can be shared between different messages? 

I guess the point here that `message` attributes describe message properties, but as an optimization, message properties can be populated once on the corresponding span. Do you think I should change the description so it's more clear?

specification/trace/semantic_conventions/messaging.md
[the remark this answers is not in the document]
lmolkova: > I know it might be a bit too specific, but for a batch of messages the value can be the same by protocol (reading a batch of messages from a single queue), or it can be that we read a few arbitrary messages and they happen to have the same destination at a single invocation.
I think this attribute is required for the first case (same destination by definition), and SHOULD NOT be set for the second case.

why SHOULD NOT? What if I'm sending 100 messages to a single destination on broker that supports per-message destination - do I pay more for this attribute repeated 100 times?  Probably not. I think instrumentation should do its best effort to avoid redundancy.

specification/trace/semantic_conventions/messaging.md
lmolkova: If we want yaml to be accurate and strict, we need to specify per-message attributes for each span. We also need to specify attributes that can go on links.

specification/trace/semantic_conventions/messaging.md
lmolkova: consumer group is a group of consumers, it does no describe source

dpauls (changes requested):

specification/trace/semantic_conventions/messaging.md
```diff
@@ -189,64 +192,130 @@ The following operations related to messages are defined for these semantic conv
...
+| [`net.sock.family`](span-general.md) | string | Protocol [address family](https://man7.org/linux/man-pages/man7/address_families.7.html) which is used for communication. | `inet6`; `bluetooth` | Conditionally Required: [8] |
 | [`net.sock.peer.addr`](span-general.md) | string | Remote socket peer address: IPv4 or IPv6 for internet protocols, path for local communication, [etc](https://man7.org/linux/man-pages/man7/address_families.7.html). | `127.0.0.1`; `/tmp/mysql.sock` | Recommended |
-| [`net.sock.peer.name`](span-general.md) | string | Remote socket peer name. | `proxy.example.com` | Recommended: [5] |
-| [`net.sock.peer.port`](span-general.md) | int | Remote socket peer port. | `16456` | Recommended: [6] |
+| [`net.sock.peer.name`](span-general.md) | string | Remote socket peer name. | `proxy.example.com` | Recommended: [9] |
+| [`net.sock.peer.port`](span-general.md) | int | Remote socket peer port. | `16456` | Recommended: [10] |
 
-**[1]:** If the message destination is either a `queue` or `topic`.
+**[1]:** Instrumentation SHOULD always set it on batch `receive` operations regardless of the actual number of received messages (e.g. 0, 1, or more).
```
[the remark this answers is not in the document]
lmolkova: After discussion on SIG meeting, changed to `Instrumentations SHOULD NOT set `messaging.batch.size` on spans that operate with a single message. When client library supports batch and single-message API for the same operation, instrumentations SHOULD use `messaging.batch.size` for batching APIs and SHOULD NOT use it for single-message APIs.`

specification/trace/semantic_conventions/messaging.md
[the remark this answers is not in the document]
lmolkova: per-message attribute table is removed

semantic_conventions/trace/messaging.yaml
```diff
@@ -1,73 +1,96 @@
...
-        examples: ['tibjmsnaming://localhost:7222', 'https://queue.amazonaws.com/80398EXAMPLE/MyQueue']
-      - id: message_id
+        brief: 'The message source name'
+        note: |
+          Source name SHOULD uniquely identify specific queue, topic, or other entity within broker. If
+          broker does not have such notion, source name SHOULD uniquely identify broker.
+        examples: ['MyQueue', 'MyTopic']
+      - id: message.id
         type: string
         brief: 'A value used by the messaging system as an identifier for the message, represented as a string.'
```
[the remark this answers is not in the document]
lmolkova: JMS message id is mapped to AMQP message id (https://www.oasis-open.org/committees/download.php/60574/amqp-bindmap-jms-v1.0-wd09.pdf), also JMS documentation describes it as `provider-assigned value`

https://docs.oracle.com/javaee/7/api/javax/jms/Message.html#setJMSMessageID-java.lang.String-

```
java.lang.String getJMSMessageID()

The JMSMessageID header field contains a value that uniquely identifies each message sent by a provider.

When a message is sent, JMSMessageID can be ignored. When the send or publish method returns, it contains a provider-assigned value.

A JMSMessageID is a String value that should function as a unique key for identifying messages in a historical repository. The exact scope of uniqueness is provider-defined. It should at least cover all messages for a specific installation of a provider, where an installation is some connected set of message routers.

All JMSMessageID values must start with the prefix 'ID:'. Uniqueness of message ID values across different providers is not required.

Since message IDs take some effort to create and increase a message's size, some JMS providers may be able to optimize message overhead if they are given a hint that the message ID is not used by an application. By calling the MessageProducer.setDisableMessageID method, a JMS client enables this potential optimization for all messages sent by that message producer. If the JMS provider accepts this hint, these messages must have the message ID set to null; if the provider ignores the hint, the message ID must be set to its normal unique value.

```

```
void setJMSMessageID([String](http://docs.oracle.com/javase/7/docs/api/java/lang/String.html?is-external=true) id)

This method is for use by JMS providers only to set this field when a message is sent. This message cannot be used by clients to configure the message ID. This method is public to allow a JMS provider to set this field when sending a message whose implementation is not its own.
```

If I understand JMS terminology correctly, the provider is a bridge between JMS and a specific client library, i.e. application does not generate Id and the provider guarantees uniqueness if the broker supports/needs it.

It seems that JMS does not intend message id to be a separate concept and providers are expected to map it to something the broker supports.

@dpauls do you believe that we can still have multiple message Ids on the same message? For the edge cases, it should be fine to have `messaging.specific-system.message.id`.

blumamir (approved):
Thank you so much @lmolkova for putting this up and for the great discussions.
All my previous comments are resolved.

I added a few minor non-blocking suggestions.

specification/trace/semantic_conventions/messaging.md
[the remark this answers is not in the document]
blumamir: This is no longer relevant after latest changes. You can resolve it (I cannot resolve it myself)

blumamir:
Also, I support moving this PR out of draft and asking for reviews from the community so we can progress towards merging this.

specification/trace/semantic_conventions/messaging.md
```diff
@@ -276,11 +365,11 @@ For Apache Kafka, the following additional attributes are defined:
...
+| `messaging.kafka.message.key` | string | Message keys in Kafka are used for grouping alike messages to ensure they're processed on the same partition. They differ from `messaging.message.id` in that they're not unique. If the key is `null`, the attribute MUST NOT be set. [1] | `myKey` | Recommended |
 | `messaging.kafka.consumer_group` | string | Name of the Kafka Consumer Group that is handling the message. Only applies to consumers, not producers. | `my-group` | Recommended |
 | `messaging.kafka.client_id` | string | Client Id for the Consumer or Producer that is handling the message. | `client-5` | Recommended |
-| `messaging.kafka.partition` | int | Partition the message is sent to. | `2` | Recommended |
-| `messaging.kafka.tombstone` | boolean | A boolean that is true if the message is a tombstone. |  | Conditionally Required: [2] |
+| `messaging.kafka.message.partition` | int | Partition the message is sent to. | `2` | Recommended |
```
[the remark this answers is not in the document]
pyohannes: I think it's right to have the partition as part of the `message` namespace. In Kafka, you can have message from different partitions in a single batch (see [here](https://stackoverflow.com/questions/64423092/consuming-kafka-batch-for-multiple-partitions)). This can't be modelled if the partition is part of the `source` or `destination` namespace.

pyohannes (approved):
I think this is ready for some wider discussion. We should open an issue, and if it's accepted, open a new clean PR based on this draft.

semantic_conventions/trace/messaging.yaml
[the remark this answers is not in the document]
lmolkova: we discussed with @dpauls that we might need more research in this space, but the discussion is not blocking this PR.
I added a new draft item to [project](https://github.com/orgs/open-telemetry/projects/20) linking to this discussion so we don't forget about it and going to resolve this thread.

specification/trace/semantic_conventions/messaging.md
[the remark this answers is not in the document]
lmolkova: agreed with @pyohannes  - partition is message property ([ProducerRecord](https://kafka.apache.org/24/javadoc/index.html?org/apache/kafka/clients/producer/ProducerRecord.html), [ConsumerRecord](https://kafka.apache.org/24/javadoc/index.html?org/apache/kafka/clients/consumer/ConsumerRecord.html))

specification/trace/semantic_conventions/messaging.md
```diff
@@ -256,17 +329,33 @@ For batch receiving and processing (see the [Batch receiving](#batch-receiving)
...
 
 In RabbitMQ, the destination is defined by an *exchange* and a *routing key*.
-`messaging.destination` MUST be set to the name of the exchange. This will be an empty string if the default exchange is used.
+`messaging.destination.name` MUST be set to the name of the exchange. This will be an empty string if the default exchange is used.
 
 <!-- semconv messaging.rabbitmq -->
 | Attribute  | Type | Description  | Examples  | Requirement Level |
 |---|---|---|---|---|
-| `messaging.rabbitmq.routing_key` | string | RabbitMQ message routing key. | `myKey` | Conditionally Required: If not empty. |
+| `messaging.rabbitmq.message.routing_key` | string | RabbitMQ message routing key. | `myKey` | Conditionally Required: If not empty. |
```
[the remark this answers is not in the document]
lmolkova: Per Rabbit docs, [routing key is a message property](https://www.rabbitmq.com/tutorials/amqp-concepts.html#messages). It's [also available](https://www.rabbitmq.com/consumers.html#message-properties) on consumers

lmolkova:
Thanks everyone for the review! Per discussion at the last SIG meeting, I'm going to close this one and open a new PR (to simplify final review process).

PR #2766, opened 2022-08-31
Explicitly allow fail-fast for invalid enum values

pellared:
## Why

For the auto-instrumentation scenario, we find that crashing the application in case of an invalid configuration is better. The reason is that it gives the users immediate feedback and forces them to fix the issue. Otherwise, it is likely that will find it when it is too late. For example, they want to analyze the telemetry when something strange is going on with the application; it would occur that someone typed `jeager` instead of `jaeger` for `OTEL_TRACES_EXPORTER`, and because of that, they will miss spans in their backend.

This is our preference in .NET Auto-Instrumentation. Reference: 
- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/design.md#error-handling
- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/952

## What

Added `or cause the application to fail on initialization` possibility when the user provides an enum value the SDK does not recognize.

AFAIK this is how the .NET SDK is working. See: https://github.com/open-telemetry/opentelemetry-dotnet/blob/a37198c6d0f1814f8f5a995413649436657a2e2b/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptions.cs#L78

## Changes

Make the enum value description compatible with the basic error handling principle: https://github.com/open-telemetry/opentelemetry-specification/blob/1fd2db2ad56668959a2dc99551b76d0ff2fc1e9e/specification/error-handling.md?plain=1#L18

I think it is a backward compatible change as the old behavior is still acceptable per spec.

FYI @open-telemetry/dotnet-instrumentation-approvers


pellared:
PTAL @open-telemetry/dotnet-maintainers

pellared:
The `markdown-link-check` [build failures](https://github.com/open-telemetry/opentelemetry-specification/runs/8111863693?check_suite_focus=true) are not related to this PR.

nrcventura:
I think this ability to fail fast is particularly important for the .NET auto-instrumentation project, because it is the best way for us to provide something to our users to let them know that the configuration is bad and needs to be fixed. Alternative options would likely lead to potentially unexpected behavior, such as no telemetry data being available. With the .net auto-instrumentation project developers may not even have access to the warnings that could be generated from an invalid enum, depending on how their how responsibilities are separated (separate observability team vs direct control of observability).

Kielek (approved):

reyang (changes requested):
I don't understand how would this proposal work - imagine the environment variable is used by some exporter, and that exporter is enabled at a later stage (e.g. the application is using some dynamic setting to allow exporters to be loaded and initialized), the wrong environment variable will cause the application to crash.

pellared:
@reyang PTAL https://github.com/open-telemetry/opentelemetry-specification/pull/2766/commits/8b28ee8d3dbd4daa3b91e2db44a718a2015d1b10

I made it explict that it cannot fail later at runtime; stolen from: https://github.com/open-telemetry/opentelemetry-specification/blob/1fd2db2ad56668959a2dc99551b76d0ff2fc1e9e/specification/error-handling.md?plain=1#L18

specification/sdk-environment-variables.md
```diff
@@ -21,7 +21,9 @@ If an SDK chooses to support an integer-valued environment variable, it SHOULD s
 ### Enum value
 
 For variables which accept a known value out of a set, i.e., an enum value, SDK implementations MAY support additional values not listed here.
-For variables accepting an enum value, if the user provides a value the SDK does not recognize, the SDK MUST generate a warning and gracefully ignore the setting.
+For variables accepting an enum value, if the user provides a value the SDK does not recognize,
+the SDK MUST generate a warning and gracefully ignore the setting,
+or _fail fast_ and cause the application to fail, but MUST NOT cause the application to fail later at run time.
```
cijothomas, line 26:
```suggestion
or _fail fast_ and cause the application to fail at startup, but MUST NOT cause the application to fail later at run time.
```

reyang (approved):

cijothomas (approved):

pellared:
PTAL @open-telemetry/technical-committee @open-telemetry/specs-approvers

specification/sdk-environment-variables.md
```diff
@@ -21,7 +21,9 @@ If an SDK chooses to support an integer-valued environment variable, it SHOULD s
 ### Enum value
 
 For variables which accept a known value out of a set, i.e., an enum value, SDK implementations MAY support additional values not listed here.
-For variables accepting an enum value, if the user provides a value the SDK does not recognize, the SDK MUST generate a warning and gracefully ignore the setting.
+For variables accepting an enum value, if the user provides a value the SDK does not recognize,
+the SDK MUST generate a warning and gracefully ignore the setting,
+or _fail fast_ and cause the application to fail at startup, but MUST NOT cause the application to fail later at run time.
```
scheler, line 26: What is the difference between `SDK implementations MAY support additional values not listed here` and `a value the SDK does not recognize`?  I think it may be better to explicitly write about each case of `allow_custom_values`.

Also, I suggest moving `NOT cause the application to fail later` closer to `gracefully ignore the setting`.

scheler (changes requested):

specification/sdk-environment-variables.md
pellared: > What is the difference between `SDK implementations MAY support additional values not listed here` and a` value the SDK does not recognize`?

The first sentence says that SDK can support more values than the spec defines. The second is about handling a value not supported by the SDK.

> I suggest moving `NOT cause the application to fail` later closer `to gracefully ignore the setting`.

Can you please propose a suggestion? Take notice that  `NOT cause the application to fail` is just to put more stress that the instrumentation should never cause a crash at runtime (after initialization). This was added to address https://github.com/open-telemetry/opentelemetry-specification/pull/2766#pullrequestreview-1095114941

scheler: How about this:
```
For variables accepting an enum value, if the user provides a value the SDK does not recognize,
the SDK SHOULD either _fail fast_ and cause the application to fail at startup or generate a warning and gracefully ignore the setting but not cause the application to fail later at runtime.
```

Note the use of SHOULD - I do not know if MUST is really required but I do not have a strong opinion about it.

pellared: TBH it is less clear for me 😄 But I appreciate your feedback 👍

pellared:
> it is not clear to me if this is an allowed breaking change or no

I think it is a backward compatible change as the old functionality is still acceptable per spec. Personally, I see it more as clarification (or even fix) as this meets the basic error handling guidelines.

>  The document is marked as "Mixed" and there is no status in the "Enum value" section. For "Mixed" document do we implicitly assume unmarked sections to be "Stable" or "Experimental"?

This is a subject for a seperate PR.

pellared:
I updated the description to describe some background.

tigrannajaryan:
> The whole "Special configuration types" section is marked as "Stable". Then I would say that "Enum value" section is "Stable" as well.

I think this means this PR makes a breaking change to a "Stable" section of the spec. An application that previously would not fail can now fail after updating to a newer SDK that implements this change. That is not allowed.

Now, there is one argument that can be made to still allow this change: .NET SDK already works as this PR describes - it fails fast. Question: what do other language SDKs do currently? Do they fail fast or ignore the invalid values?

If all/most other SDKs also fail fast then I think we can say that the spec has a bug and what is written does not match the reality.

pellared:
> An application that previously would not fail can now fail after updating to a newer SDK that implements this change.

That would be indeed a breaking change but in the SDK 😉 But I do not have a strong opinion here.

carlosalberto:
As discussed in yesterday's Spec call, one alternative is to fail fast _when_ using a debug mode flag (maybe set through an environment variable). See #2817 

Would this help _at all_ for your considered case @pellared ?

specification/sdk-environment-variables.md
```diff
@@ -21,7 +21,9 @@ If an SDK chooses to support an integer-valued environment variable, it SHOULD s
 ### Enum value
 
 For variables which accept a known value out of a set, i.e., an enum value, SDK implementations MAY support additional values not listed here.
-For variables accepting an enum value, if the user provides a value the SDK does not recognize, the SDK MUST generate a warning and gracefully ignore the setting.
+For variables accepting an enum value, if the user provides a value the SDK does not recognize,
```
yurishkuro, line 24: I am not sure why this behavior is only under Enum values, shouldn't there be a common error behavior for all env values? What if integer var is set to `blah`?

specification/sdk-environment-variables.md
yurishkuro, line 26: I don't think this change is necessary - https://github.com/open-telemetry/opentelemetry-specification/issues/2817#issuecomment-1257098051

specification/sdk-environment-variables.md
pellared: I was thinking about it as well, but I want to avoid scope creep in this PR. Even right now it is not clear how to proceed

PR #2780, opened 2022-09-09
Introduce "Scope full name" for backwards compatibility

### LINKED ISSUE #2762: Scope attributes as part of identity is a breaking change
During today's SIG meeting, @tigrannajaryan point me to  https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#get-a-meter.

This is a breaking change, since after this change is applied as a receiver of telemetry I can receive now multiple scopes with same name/version which was forbidden before.

jmacd:
Fixes #2762

## Changes

This requires producers to include Scope Attributes in the Scope "full name", which uses the same encoding as HTTP `User-Agent` to include multiple key-values with optional comments. This addresses the backwards-compatibility problem with Scope Attributes by ensuring that the full name uniquely (and redundantly) specifies the scope attributes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -94,7 +94,7 @@ release.
 - Introduce Instrumentation Scope Attributes
   ([#2579](https://github.com/open-telemetry/opentelemetry-specification/pull/2579))
 - Introduce Instrumentation Scope Full name requirements
-  ([#2780])(https://github.com/open-telemetry/opentelemetry-specification/pull/2780).
+  ([#2780](https://github.com/open-telemetry/opentelemetry-specification/pull/2780)).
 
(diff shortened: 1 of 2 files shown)


jmacd:
Can you explain the "breakage"? I don't see any breakage.

Oberon00:
I meant "breakage" as in "breaking change". I.e., instead of making a breaking change to the spec, adding any new attribute to a scope is a breaking change. And it is also harder to group scopes by their actual name now.

tigrannajaryan:
@jmacd if you agree with https://github.com/open-telemetry/opentelemetry-specification/pull/2789 then we can close this.

PR #2788, opened 2022-09-12
Fix OpenMetrics valid label keys, and specify prometheus conversion for metric name.

dashpole:
## Changes

This fixes an incorrect statement about the allowed characters in prometheus label keys found here: https://github.com/open-telemetry/opentelemetry-go/pull/3135#discussion_r968823675.

During the process, I also realized we don't specify conversion rules for the metric name, including the rules for sanitizing the metric name when converting from OTLP to Prom.

Source: https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#abnf

In that spec, `metricname-initial-char` allows `:`, but `label-name-initial-char` does not.


MrAlias (approved):

specification/metrics/data-model.md
```diff
@@ -1488,7 +1499,7 @@ The following OTLP data points MUST be dropped:
...
+OpenTelemetry Metric Attributes MUST be converted to [Prometheus labels](https://Prometheus.io/docs/concepts/data_model/#metric-names-and-labels).  String Attribute values are converted directly to Metric Attributes, and non-string Attribute values MUST be converted to string attributes following the [attribute specification](../common/README.md#attribute).  Prometheus metric label keys are required to match the following regex: `[a-zA-Z_]([a-zA-Z0-9_])*`.  Metrics from OpenTelemetry with unsupported Attribute names MUST replace invalid characters with the `_` character. This may cause ambiguity in scenarios where multiple similar-named attributes share invalid characters at the same location.  In such unlikely cases, if multiple key-value pairs are converted to have the same Prometheus key, the values MUST be concatenated together, separated by `;`, and ordered by the lexicographical order of the original keys.
```
damemi, line 1502: labels with a double-underscore prefix `__` are reserved in prometheus's model, should we mention that? how are they handled? (since the regex technically allows it)

dashpole: Added: Multiple consecutive `_` characters MUST be replaced with a single `_`
character."

damemi (approved):

dashpole:
cc @open-telemetry/wg-prometheus

bogdandrutu (approved):

jmacd:
@dashpole related question, maybe not answered in this PR: 
What is prescribed for empty label keys? 
The [specification] merely requires valid Unicode sequences. I think I would prefer it said non-empty sequences, but I'm not sure what SDKs should do in this case. What do you think?

dashpole:
For Prometheus/OpenMetrics, empty label keys [aren't allowed](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#abnf). We should drop + warn for an empty label.  For exporters written using prometheus client libraries, that is already done by the client's validation (e.g. [prometheus/client_golang](https://github.com/prometheus/common/blob/main/model/labels.go#L103)).  It might be good to include a catchall statement somewhere to the effect of: "exporters should produce valid prometheus/openmetrics metrics, or MUST drop the metric and emit a warning".

reyang (approved):

jmacd (approved):

PR #2789, opened 2022-09-12
Remove scope attributes form the Scope identity

bogdandrutu:
Fixes https://github.com/open-telemetry/opentelemetry-specification/issues/2762

The identity part can be changed later if the community agrees to accept this change, but to be safe for the moment revert this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -95,6 +95,8 @@ release.
 
 - Introduce Instrumentation Scope Attributes
   ([#2579](https://github.com/open-telemetry/opentelemetry-specification/pull/2579))
+  - Define Instrumentation Scope Attributes as non identifiers
+    ([#2789](https://github.com/open-telemetry/opentelemetry-specification/pull/2789))
 
 ## v1.12.0 (2022-06-10)
 
--- a/specification/logs/api.md
+++ b/specification/logs/api.md
@@ -99,11 +99,15 @@ SHOULD be true by default.
 - `attributes` (optional): Specifies the instrumentation scope attributes to
 associate with emitted telemetry.
 
-Implementations MUST return different `Logger` instances when called repeatedly
-with different values of parameters. Note that always returning a new `Logger`
-instance is a valid implementation. The only exception to this rule is the no-op
-`Logger`: implementations MAY return the same instance regardless of parameter
-values.
+Loggers are identified by `name`, `version`, and `schema_url` fields.  When more
+than one `Logger` of the same `name`, `version`, and `schema_url` is created, it
+is unspecified whether or under which conditions the same or different `Logger`
+instances are returned. It is a user error to create Loggers with different
+attributes but the same identity.
+
+The term *identical* applied to Loggers describes instances where all
+identifying fields are equal. The term *distinct* applied to Loggers describes
+instances where at least one identifying field has a different value.
 
 Implementations MUST NOT require users to repeatedly obtain a Logger again with
 the same name+version+schema_url+event_domain+include_trace_context+attributes
--- a/specification/metrics/api.md
+++ b/specification/metrics/api.md
@@ -147,20 +147,15 @@ This API MUST accept the following parameters:
 * [since 1.13.0] `attributes` (optional): Specifies the instrumentation scope attributes
     to associate with emitted telemetry.
 
-Meters are identified by all of these parameters.
+Meters are identified by `name`, `version`, and `schema_url` fields.  When more
+than one `Meter` of the same `name`, `version`, and `schema_url` is created, it
+is unspecified whether or under which conditions the same or different `Meter`
+instances are returned. It is a user error to create Meters with different
+attributes but the same identity.
 
-Implementations MUST return different `Meter` instances when called repeatedly
-with different values of parameters. Note that always returning a new `Meter` instance
-is a valid implementation. The only exception to this rule is the no-op `Meter`:
-implementations MAY return the same instance regardless of parameter values.
-
-It is unspecified whether or under which conditions the same or different
-`Meter` instances are returned from this function when the same
-(name,version,schema_url,attributes) parameters are used.
-
-The term *identical* applied to Meters describes instances where all identifying fields
-are equal. The term *distinct* applied to Meters describes instances where at
-least one identifying field has a different value.
+The term *identical* applied to Meters describes instances where all identifying
+fields are equal. The term *distinct* applied to Meters describes instances where
+at least one identifying field has a different value.
 
 Implementations MUST NOT require users to repeatedly obtain a `Meter` with
 the same identity to pick up configuration changes. This can be
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -131,23 +131,28 @@ This API MUST accept the following parameters:
 - [since 1.13.0] `attributes` (optional): Specifies the instrumentation scope attributes
   to associate with emitted telemetry.
 
-Implementations MUST return different `Tracer` instances when called repeatedly
-with different values of parameters. Note that always returning a new `Tracer` instance
-is a valid implementation. The only exception to this rule is the no-op `Tracer`:
-implementations MAY return the same instance regardless of parameter values.
+Tracers are identified by `name`, `version`, and `schema_url` fields.  When more
+than one `Tracer` of the same `name`, `version`, and `schema_url` is created, it
+is unspecified whether or under which conditions the same or different `Tracer`
+instances are returned. It is a user error to create Tracers with different
+attributes but the same identity.
+
+The term *identical* applied to Tracers describes instances where all
+identifying fields are equal. The term *distinct* applied to Tracers describes
+instances where at least one identifying field has a different value.
 
 Implementations MUST NOT require users to repeatedly obtain a `Tracer` again
-with the same name+version+schema_url+attributes to pick up configuration changes.
-This can be achieved either by allowing to work with an outdated configuration or
-by ensuring that new configuration applies also to previously returned `Tracer`s.
+with the same identity to pick up configuration changes. This can be
+achieved either by allowing to work with an outdated configuration or by
+ensuring that new configuration applies also to previously returned `Tracer`s.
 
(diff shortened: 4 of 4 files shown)


reyang (approved):

reyang:
Consider adding a changelog entry.

specification/metrics/api.md
```diff
@@ -147,20 +147,14 @@ This API MUST accept the following parameters:
...
-`Meter` instances are returned from this function when the same
-(name,version,schema_url,attributes) parameters are used.
-
-The term *identical* applied to Meters describes instances where all identifying fields
-are equal. The term *distinct* applied to Meters describes instances where at
-least one identifying field has a different value.
+Meters are identified by name, version, and schema_url fields.  When more than one
+Meter of the same `name`, `version`, and `schema_url` is created, it
+is unspecified whether or under which conditions the same or different
+`Meter` instances are returned.
```
Oberon00, line 153: Can we spell out more clearly (as a note, since technically it's already clear), that as a consequence, it is a user error to create scopes with different attributes but the same identity?

I don't think I like that consequence, but it may be the best way to move forward for now, and we can find a better solution later.

bogdandrutu: Done.

specification/logs/api.md
```diff
@@ -99,11 +99,15 @@ SHOULD be true by default.
...
 
-Implementations MUST return different `Logger` instances when called repeatedly
-with different values of parameters. Note that always returning a new `Logger`
-instance is a valid implementation. The only exception to this rule is the no-op
-`Logger`: implementations MAY return the same instance regardless of parameter
-values.
+Loggers are identified by `name`, `version`, and `schema_url` fields.  When more
+than one `Logger` of the same `name`, `version`, and `schema_url` is created, it
+is unspecified whether or under which conditions the same or different `Logger`
+instances are returned. As a consequence, it is a user error to create Meters
```
tigrannajaryan, line 105:
```suggestion
instances are returned. It is a user error to create Loggers
```

specification/metrics/api.md
```diff
@@ -147,20 +147,15 @@ This API MUST accept the following parameters:
...
-`Meter` instances are returned from this function when the same
-(name,version,schema_url,attributes) parameters are used.
-
-The term *identical* applied to Meters describes instances where all identifying fields
-are equal. The term *distinct* applied to Meters describes instances where at
-least one identifying field has a different value.
+Meters are identified by `name`, `version`, and `schema_url` fields.  When more
+than one `Meter` of the same `name`, `version`, and `schema_url` is created, it
+is unspecified whether or under which conditions the same or different `Meter`
+instances are returned. As a consequence, it is a user error to create a `Meter`
```
tigrannajaryan, line 153:
```suggestion
instances are returned. It is a user error to create a `Meter`
```

jmacd (approved):

specification/logs/api.md
bogdandrutu: Done.

MrAlias:
If scopes have attributes, but these attributes are not identifying, how will we resolve two providers with the same identity (same name, version, etc.) but different attributes? Does this mean OTel implementations should not add attributes to scopes?

bogdandrutu:
> If scopes have attributes, but these attributes are not identifying, how will we resolve two providers with the same identity (same name, version, etc.) but different attributes? 

Providers? If you are talking about providers then they have different resource.

Meters/Tracers/Loggers? Then I think, since is clearly stated that this is not valid and undefined, an valid implementation may return the same instance as the first call and ignore attributes from the second call (may also report an error via the error reporting mechanism implemented, log, etc.), another approach is to return a no-op implementation second time and report an error. 

> Does this mean OTel implementations should not add attributes to scopes?

You should, most likely I miss something because I don't understand where this question is coming from.

MrAlias:
> > If scopes have attributes, but these attributes are not identifying, how will we resolve two providers with the same identity (same name, version, etc.) but different attributes?
> 
> Providers? If you are talking about providers then they have different resource.

:facepalm:, yeah, sorry I wrote that in error. I did mean Meters/Tracers/Loggers.

> Meters/Tracers/Loggers? Then I think, since is clearly stated that this is not valid and undefined, an valid implementation may return the same instance as the first call and ignore attributes from the second call (may also report an error via the error reporting mechanism implemented, log, etc.), another approach is to return a no-op implementation second time and report an error.

Hmm, that doesn't seem ideal. That could have some implementations returning the original silently while others completely fail to record things.

tigrannajaryan:
> Hmm, that doesn't seem ideal. That could have some implementations returning the original silently while others completely fail to record things.

I think that's the point of undefined behavior. Users should not do this. If they do then then bad things can happen (including crashing, although I suppose we won't do it). 

This is just for now. We can refine this behavior and specify what should actually happen. But because we are not yet sure what is the desirable behavior we are making it undefined in the spec and explicitly say that it is a user error.

tigrannajaryan (approved):

jmacd:
@bogdandrutu I agree with the undefined behavior, however of following options you listed for a valid implementation:

1.
> may return the same instance as the first call and ignore attributes from the second call

That sounds OK. There is no change of identity by definition and so equivalent telemetry data is produced. Users of the SDK that makes this choice will likely press for work on the specification to enable identifying scope attributes.

2.
> another approach is to return a no-op implementation second time and report an error.

That sounds NOT OK. You can tell the user there was a problem, but to stop reporting telemetry is a grave mistake.

Here's another behavior I believe is OK:

3. Return the intended scope with different attributes than others prior, therefore produce "conflicting" scopes. This is in line with existing behavior specified for SDKs, namely to pass through Metric data for all instrument definitions regardless of conflicts. Users of the SDK that makes this choice will likely press their vendors to recognize (and submit PRs to accomplish) the identifying nature of scope attributes without confirmation from the specification.

MrAlias:
> > Hmm, that doesn't seem ideal. That could have some implementations returning the original silently while others completely fail to record things.
> 
> I think that's the point of undefined behavior. Users should not do this. If they do then then bad things can happen (including crashing, although I suppose we won't do it).
> 
> This is just for now. We can refine this behavior and specify what should actually happen. But because we are not yet sure what is the desirable behavior we are making it undefined in the spec and explicitly say that it is a user error.

Gotcha. But this leads me back to my original question of if an implementation should even add attributes to a scope at this point. If they aren't exported, cause undefined behavior if defined at creation of a Meter/Tracer/Logger, and make the scope an invalid map key to use for identity, why add them?

tigrannajaryan:
> Gotcha. But this leads me back to my original question of if an implementation should even add attributes to a scope at this point. If they aren't exported, cause undefined behavior if defined at creation of a Meter/Tracer/Logger, and make the scope an invalid map key to use for identity, why add them?

They only cause problem if you obtain Meter/Tracer/Logger with a different set of attributes, which is outlawed by this PR. If used correctly (always use the same attributes for the same Scope identity) then it will/should work fine. It should export without problems.

MrAlias:
> 3. Return the intended scope with different attributes than others prior, therefore produce "conflicting" scopes. This is in line with existing behavior specified for SDKs, namely to pass through Metric data for all instrument definitions regardless of conflicts. Users of the SDK that makes this choice will likely press their vendors to recognize (and submit PRs to accomplish) the identifying nature of scope attributes without confirmation from the specification.

If this is a valid implementation, it seems fine to me. @bogdandrutu is this behavior intended to be allowed?

CHANGELOG.md
```diff
@@ -95,6 +95,8 @@ release.
 
 - Introduce Instrumentation Scope Attributes
   ([#2579](https://github.com/open-telemetry/opentelemetry-specification/pull/2579))
+  - Define Instrumentation Scope Attributes as non identifiers
+    ([#2789](https://github.com/open-telemetry/opentelemetry-specification/pull/2789))
```
jack-berg, line 99:
```suggestion
- Define Instrumentation Scope Attributes as non identifiers
   ([#2789](https://github.com/open-telemetry/opentelemetry-specification/pull/2789))
```

jack-berg (approved):

CHANGELOG.md
bogdandrutu: Intentionally had this as a subset of the other. why changing?

jack-berg: Didn't realize that! Please disregard.

bogdandrutu:
>> Return the intended scope with different attributes than others prior, therefore produce "conflicting" scopes. This is in line with existing behavior specified for SDKs, namely to pass through Metric data for all instrument definitions regardless of conflicts. Users of the SDK that makes this choice will likely press their vendors to recognize (and submit PRs to accomplish) the identifying nature of scope attributes without confirmation from the specification.

> If this is a valid implementation, it seems fine to me. @bogdandrutu is this behavior intended to be allowed?

I would say that for the moment any implementation that does not make an assumption or needs the attributes to be identifiers is ok. I prefer to avoid the "conflicting" scopes proposal, since if that is available and start to be used will become part of your stability guarantees and you will not be able to change if the specification changes in a way to define that behavior, which is OK to do for the specification when UB is defined.

You should explicitly mark that as undefined behavior and tell users that can produce unspecified behaviors.

MrAlias:
> > > Return the intended scope with different attributes than others prior, therefore produce "conflicting" scopes. This is in line with existing behavior specified for SDKs, namely to pass through Metric data for all instrument definitions regardless of conflicts. Users of the SDK that makes this choice will likely press their vendors to recognize (and submit PRs to accomplish) the identifying nature of scope attributes without confirmation from the specification.
> 
> > If this is a valid implementation, it seems fine to me. @bogdandrutu is this behavior intended to be allowed?
> 
> I would say that for the moment any implementation that does not make an assumption or needs the attributes to be identifiers is ok. I prefer to avoid the "conflicting" scopes proposal, since if that is available and start to be used will become part of your stability guarantees and you will not be able to change if the specification changes in a way to define that behavior, which is OK to do for the specification when UB is defined.
> 
> You should explicitly mark that as undefined behavior and tell users that can produce unspecified behaviors.

gotcha :+1:

MrAlias (approved):

PR #2794, opened 2022-09-13
Clarify that Scope attributes are not part of its identity

tigrannajaryan:
Resolves https://github.com/open-telemetry/opentelemetry-specification/issues/2762

This change says that Scope attributes are NOT part of the Tracer/Meter/Logger identity.

The consequence of this change is the following:

- It is no longer possible to obtain a Logger with different sets of attributes from the same instrumentation library (using the same name, version,schema_url). It may be desirable to make this possible in the future, but right now it is impossible since the attributes are not part of the Scope's identity.

- Likewise, it is no longer possible to obtain a Tracer and Meter with different sets of attributes from the same instrumentation library (using the same name, version,schema_url). This is likely the only desirable way for Tracer and Meter.

This is an alternate to draft https://github.com/open-telemetry/opentelemetry-specification/pull/2780

### DIFF
--- a/specification/logs/api.md
+++ b/specification/logs/api.md
@@ -100,13 +100,13 @@ SHOULD be true by default.
 associate with emitted telemetry.
 
 Implementations MUST return different `Logger` instances when called repeatedly
-with different values of parameters. Note that always returning a new `Logger`
+with different values of (name,version,schema_url) parameters. Note that always returning a new `Logger`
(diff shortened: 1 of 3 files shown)


bogdandrutu:
Duplicate of https://github.com/open-telemetry/opentelemetry-specification/pull/2789

tigrannajaryan:
Closed in favour of https://github.com/open-telemetry/opentelemetry-specification/pull/2789

PR #2807, opened 2022-09-15
Fix nits: incorrect unit for CPU utilization

### LINKED ISSUE #2802: The CPU utilization metrics seem to have wrong unit
It seems to be talking about ratio (which is a number in range `[0, 1.0]`), while the unit `s` seems wrong. https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/process-metrics.md#process

image

reyang:
Fixes #2802


cijothomas (approved):

tigrannajaryan (approved):
Unrelated to the change: I think we need to add examples to metric semantic conventions just like we have already for traces. This would help expose mistakes like this.

bogdandrutu (approved):

PR #2825, opened 2022-09-23
Add `process.start_time` resource attribute to semantic conventions

### LINKED ISSUE #1273: Metrics Start-time resource semantic convention

There has been some discussion about an Uptime metric. For example, the OpenTelemetry-Go `runtime` instrumentation includes one:

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/d1534b84593e617bff9a848454a992a7af49385c/instrumentation/runtime/runtime.go#L122

There is a related request for an `up` metric, meaning something like "was able to produce metrics" in #1078. The uptime metric is different and can be used for monitoring process longevity, for example. There is a question of whether we should standardize a semantic-conventional metric name for uptime.

However, note that when we know the process start time, we are able to deduce the uptime provided we know that a process was up. Logically, a combination of the `up` metric and a `process.start_time` resource combine so that we can synthesize an `process.uptime` metric.

I've encountered a reason to prefer the use of a `start_time resource and an `up` metric as opposed to an `process.uptime` metric, stated as follows.

An `UpDownSumObserver` instrument writes an OTLP Non-Monontonic Cumulative Sum data point, there is a well-defined conversion to Gauge in systems such as Prometheus that do not recognize Non-Monotonic Cumulatives. An `UpDownCounter` instrument writes an OTLP Non-Monotonic Delta Sum data point for the Stateless export configuration, but it is converted to a Cumulative in the default configuration. As long as the state that we maintain in an SDK for Delta-to-Cumulative conversion is never reset, there is no difference to the consumer of an OTLP Non-Monotonic Cumulative Sum (OTLP-NMCS) data point whether it was originally an `UpDownSumObserver` or an `UpDownCounter`. 

If we move the Delta-to-Cumulative conversion out of the process (e.g., into a sidecar), then there may be a difference between an OTLP-NMCS that was reset and one that was never reset. We could use the start-time resource to detect this difference.  This feels significant because ultimately, if the user is going to view a Cumulative Sum as its current, total value, then we should know whether it's the cumulative from the beginning of the process or cumulative from an arbitrary reset point. In a user-interface for a OTLP-NMCS timeseries, I would consider a generating an error to say that for Non-Monotonic Sums that have been reset you should only use Rate views, not Total views.

Concretely speaking the proposed semantic convention would be named `process.start_time` and would be documented [here](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/resource/semantic_conventions/process.md#process).

astencel-sumo:
Fixes #1273

## Changes

Adds a new resource attribute `process.start_time` to the semantic conventions.

Here's a related pull request that adds `process.uptime` and `system.uptime` metrics: https://github.com/open-telemetry/opentelemetry-specification/pull/2824.

I also wanted to add a `system.start_time` resource attribute, but I'm not sure where to put it - there's no `system` namespace in the resource attributes at the moment; the closest I could find is the [`os`](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.13.0/specification/resource/semantic_conventions/os.md) namespace. Should I add an `os.start_time` attribute?

### DIFF
--- a/specification/resource/semantic_conventions/process.md
+++ b/specification/resource/semantic_conventions/process.md
@@ -35,6 +35,7 @@
 | `process.command_line` | string | The full command used to launch the process as a single string representing the full command. On Windows, can be set to the result of `GetCommandLineW`. Do not set this if you have to assemble it just for monitoring; use `process.command_args` instead. | `C:\cmd\otecol --config="my directory\config.yaml"` | Conditionally Required: See alternative attributes below. |
 | `process.command_args` | string[] | All the command arguments (including the command/executable itself) as received by the process. On Linux-based systems (and some other Unixoid systems supporting procfs), can be set according to the list of null-delimited strings extracted from `proc/[pid]/cmdline`. For libc-based executables, this would be the full argv vector passed to `main`. | `[cmd/otecol, --config=config.yaml]` | Conditionally Required: See alternative attributes below. |
 | `process.owner` | string | The username of the user that owns the process. | `root` | Recommended |
+| `process.start_time` | int | The start time of the process in UTC, expressed as number of seconds since the Unix epoch. | `1663931438` | Recommended |
 
 **Additional attribute requirements:** At least one of the following sets of attributes is required:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,8 @@ release.
   ([#2874](https://github.com/open-telemetry/opentelemetry-specification/pull/2874))
 - Add `process.paging.faults` metric to semantic conventions
   ([#2827](https://github.com/open-telemetry/opentelemetry-specification/pull/2827))
+- Add `process.start_time` resource attribute to semantic conventions
+  ([#2825](https://github.com/open-telemetry/opentelemetry-specification/pull/2825))
 
 ### Compatibility
 
--- a/semantic_conventions/resource/process.yaml
+++ b/semantic_conventions/resource/process.yaml
@@ -68,6 +68,12 @@ groups:
         brief: >
           The username of the user that owns the process.
         examples: 'root'
+      - id: start_time
+        type: int
+        brief: >
+          The start time of the process in UTC,
+          expressed as number of seconds since the Unix epoch.
+        examples: [1663931438]
     constraints:
       - any_of:
           - process.executable.name


specification/resource/semantic_conventions/process.md
```diff
@@ -35,6 +35,7 @@
 | `process.command_line` | string | The full command used to launch the process as a single string representing the full command. On Windows, can be set to the result of `GetCommandLineW`. Do not set this if you have to assemble it just for monitoring; use `process.command_args` instead. | `C:\cmd\otecol --config="my directory\config.yaml"` | Conditionally Required: See alternative attributes below. |
 | `process.command_args` | string[] | All the command arguments (including the command/executable itself) as received by the process. On Linux-based systems (and some other Unixoid systems supporting procfs), can be set according to the list of null-delimited strings extracted from `proc/[pid]/cmdline`. For libc-based executables, this would be the full argv vector passed to `main`. | `[cmd/otecol, --config=config.yaml]` | Conditionally Required: See alternative attributes below. |
 | `process.owner` | string | The username of the user that owns the process. | `root` | Recommended |
+| `process.start_time` | int | The start time of the process in UTC, expressed as number of seconds since the Unix epoch. | `1663931438` |
```
reyang, line 38: Is second good enough or a higher precision is desired?

astencel-sumo: Perhaps the ECS's [process.start](https://www.elastic.co/guide/en/ecs/8.4/ecs-process.html#field-process-start) is the better option here?

reyang: Related to https://github.com/open-telemetry/oteps/pull/199#issuecomment-1256380583.

jsuereth (changes requested):
Semantic conventions on Resource Attributes that can affect metric identity should be held off for now.

Additionally, I'd really like to understand the use case for including `start_time` in the resource better.  I don't think the justification in the PR is enough.  Can you outline the use case / expected use of this compared to metric start times, and an actual metric recording process start time?

-

astencel-sumo:
> Additionally, I'd really like to understand the use case for including `start_time` in the resource better. I don't think the justification in the PR is enough. Can you outline the use case / expected use of this compared to metric start times, and an actual metric recording process start time?

There are two reasons for proposing the process start time as an attribute of the process resource:

1. I believe it's part of the identity of the process. The process ID, parent ID, executable path, parameters are good enough in most cases to uniquely identify a process, but if you imagine a process creating many many child processes one after another (each lasting for a short time), I assume it's possible for the process ID counter to wrap around (the max is [~4 million](https://unix.stackexchange.com/questions/16883/what-is-the-maximum-value-of-the-process-id)) and start reusing process IDs, right? Note that for scenarios like this, resolution higher than a second is probably better, as Reiley noted [above](https://github.com/open-telemetry/opentelemetry-specification/pull/2825#discussion_r978762386).

2. I considered adding a metric that contains the start time of the process, e.g. `process.start_time`. But this metric would be a constant by definition and I'm not sure it's a good candidate for a metric if its value never changes? I mean the value would only change when an actual new process with the same resource attributes started reporting a new value of `process.start_time`, but that's really a different resource, which gets us back to point 1 :slightly_smiling_face:. I'm not against such metric, but I thought the resource attribute is more fitting here.

Does adding this resource attribute cause cardinality explosion? I don't think so. It causes a "serial" split of timeseries rather than a "parallel" one (I wonder if this description makes sense).

As a side note, please note the [comments](https://github.com/open-telemetry/opentelemetry-specification/issues/1273#issuecomment-1257942046) on this PR's issue about the Elastic Common Schema's `process.start` field that contains the RFC-3339 timestamp. Dmitrii likes the `start_time` name better, but I think it's good to stick to the ECS whenever possible. What do you think about this Josh?

astencel-sumo:
@jsuereth any update on this?

astencel-sumo:
Rebased the PR.

astencel-sumo:
@jsuereth can you please take a look at my [comment](https://github.com/open-telemetry/opentelemetry-specification/pull/2825#issuecomment-1268078956) addressing your comment above?

jsuereth:
> There are two reasons for proposing the process start time as an attribute of the process resource:
>
> 1. I believe it's part of the identity of the process. The process ID, parent ID, executable path, parameters are good enough in most cases to uniquely identify a process, but if you imagine a process creating many many child processes one after another (each lasting for a short time), I assume it's possible for the process ID counter to wrap around (the max is [~4 million](https://unix.stackexchange.com/questions/16883/what-is-the-maximum-value-of-the-process-id)) and start reusing process IDs, right? Note that for scenarios like this, resolution higher than a second is probably better, as Reiley noted https://github.com/open-telemetry/opentelemetry-specification/pull/2825#discussion_r978762386.

Process identity isn't in a vacuum.  All telemetry from OTEL includes a timestamp when it was created, and metrics specifically include start/now timestamps.  Additionally, process isn't the only identity we'll attach.  We'll include VM, pod, container, etc. instance metadata.

Is pulling process start time something we want all SDKs doing via an external mechanism? I realize you're taking the process-metric view, but this also impacts SDKs. E.g. [Java's ProcessResource](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/resources/library/src/main/java/io/opentelemetry/instrumentation/resources/ProcessResource.java).   Do you have a way to do this for all languages?

At a minimum, I think process.start_time should be "optional" not "recommended".


> 2. I considered adding a metric that contains the start time of the process, e.g. process.start_time. But this metric would be a constant by definition and I'm not sure it's a good candidate for a metric if its value never changes? I mean the value would only change when an actual new process with the same resource attributes started reporting a new value of process.start_time, but that's really a different resource, which gets us back to point 1 slightly_smiling_face. I'm not against such metric, but I thought the resource attribute is more fitting here.

Almost every metric produced by a metric SDK will (effectively) include a start_time that is likely the process start time.  I'm not suggesting you create a new metric I'm asking about the (inherent) duplication that will likely occur here.

More importantly, it's *likely* that most SDKs just assume a start timestamp of the first timestamp they grab on startup.  Is this ok to use for process start_time?  Do we expect consistency for SDK-reported metrics and process.start_time to align with the metric start_time?

> Does adding this resource attribute cause cardinality explosion? I don't think so. It causes a "serial" split of timeseries rather than a "parallel" one (I wonder if this description makes sense).

Yes. I can understand this argument.  However, even if it does a serial split, it can still cause load on TSDBs in other ways.  This why, even for serial splits, OpenMetrics still encourages being frugal and minimal with labels.  My main question here is whether or not process start time adds enough value to justify its weight?



> As a side note, please note the https://github.com/open-telemetry/opentelemetry-specification/issues/1273#issuecomment-1257942046 on this PR's issue about the Elastic Common Schema's process.start field that contains the RFC-3339 timestamp. Dmitrii likes the start_time name better, but I think it's good to stick to the ECS whenever possible. What do you think about this Josh?

I think we should aim for compat with ECS when there's no clear advantage one way or another.

Issue #2795, opened 2022-09-13
[otlp] Decide if original and lowerCamelCase JSON keys should be supported


tigrannajaryan:
The decision that we need to make is whether both the original proto field name and the camelCase version of the field name should be acceptable. It is acceptable by [Prototobuf JSON definition](https://developers.google.com/protocol-buffers/docs/proto3#json). Do we see any need to deviate from this?

tigrannajaryan:
It is not entirely clear what the definition of JSON-friendliness is. They keys used by OTLP are valid in JSON. 

Perhaps one decision that we need to make is whether both the original proto field name and the camelCase version of the field name should be acceptable. It is acceptable by [Prototobuf JSON definition](https://developers.google.com/protocol-buffers/docs/proto3#json). Do we see any need to deviate from this?

tigrannajaryan:
cc @mwear @dyladan @bogdandrutu @jmacd @Aneurysm9 in case you have any thoughts about this.

dyladan:
I think the default behavior of the protobuf client is to change keys to lowerCamelCase and is likely what the JS exporter is currently doing. I don't see any reason not to accept both though since it is default proto behavior.

tigrannajaryan:
I agree with @dyladan that we should just keep the default Protobuf behavior for exporters and also keep the ability for receivers to accept both.

@open-telemetry/specs-approvers any opinions on this?

dyladan:
I don't see any reason to restrict the exporters to this behavior. I would just say receivers should expect both and what exporters do you what makes sense for their users expectations.

tigrannajaryan:
> I don't see any reason to restrict the exporters to this behavior. I would just say receivers should expect both and let exporters do what makes sense for their users expectations.

Yes, that's essentially what will happen if we say receivers must accept both.

PR #2829, opened 2022-09-23
Clarify that lowerCamelCase field names MUST be used for OTLP/JSON

tigrannajaryan:
Resolves https://github.com/open-telemetry/opentelemetry-specification/issues/2795

This is a breaking change for OTLP/JSON and is allowed because OTLP/JSON is not yet Stable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,8 @@ release.
 
 ### OpenTelemetry Protocol
 
+- Clarify that lowerCamelCase field names MUST be used for OTLP/JSON
+  ([#2829](https://github.com/open-telemetry/opentelemetry-specification/pull/2829))
 - Add user agent to OTLP exporter specification
   ([#2684](https://github.com/open-telemetry/opentelemetry-specification/pull/2684))
 - Prohibit usage of enum value name strings in OTLP/JSON
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@ release.
 
 ### OpenTelemetry Protocol
 
+- Clarify that lowerCamelCase field names MUST be used for OTLP/JSON
+  ([#2829](https://github.com/open-telemetry/opentelemetry-specification/pull/2829))
+
 ### SDK Configuration
 
 ### Telemetry Schemas
@@ -81,8 +84,6 @@ release.
 
 ### OpenTelemetry Protocol
 
-- Clarify that lowerCamelCase field names MUST be used for OTLP/JSON
-  ([#2829](https://github.com/open-telemetry/opentelemetry-specification/pull/2829))
 - Add user agent to OTLP exporter specification
   ([#2684](https://github.com/open-telemetry/opentelemetry-specification/pull/2684))
 - Prohibit usage of enum value name strings in OTLP/JSON
--- a/specification/protocol/otlp.md
+++ b/specification/protocol/otlp.md
@@ -424,6 +424,13 @@ for mapping between Protobuf and JSON, with the following deviations from that m
   This aligns with the behavior of the Binary Protobuf unmarshaler and ensures that adding
   new fields to OTLP messages does not break existing receivers.
 
+- The keys of JSON objects are field names converted to lowerCamelCase. Original
+  field names are not valid to use a keys of JSON objects.
+  For example this is a valid JSON representation of a Resource:
+  `{ "attributes": {...}, "droppedAttributesCount": 123 }`, and this is NOT a valid
+  representation:
+  `{ "attributes": {...}, "dropped_attributes_count": 123 }`.
+
 Note that according to [Protobuf specs](
 https://developers.google.com/protocol-buffers/docs/proto3#json) 64-bit integer
 numbers in JSON-encoded payloads are encoded as decimal strings, and either


specification/protocol/otlp.md
```diff
@@ -420,9 +420,14 @@ for mapping between Protobuf and JSON, with the following deviations from that m
...
 Note that according to [Protobuf specs](
-https://developers.google.com/protocol-buffers/docs/proto3#json) 64-bit integer
-numbers in JSON-encoded payloads are encoded as decimal strings, and either
-numbers or strings are accepted when decoding.
+https://developers.google.com/protocol-buffers/docs/proto3#json):
+
+- 64-bit integer numbers in JSON-encoded payloads are encoded as decimal strings, and
+  either numbers or strings are accepted when decoding.
+
+- Encoders are allowed to produce field names either equal to the original field names
```
yurishkuro, line 428: This reads as if the choice of the field name can vary from encoder to encoder. That's not what the proto spec says:

> Message field names are mapped to lowerCamelCase and become JSON object keys. If the json_name field option is specified, the specified value will be used as the key instead. 

There's no encoder-choice in that wording: `json_name` takes precedent, otherwise its lowerCamelCase.

But the parsers, for some (probably bad historical) reasons are also required to recognized original field names from proto files, even though encoders are not allowed to do that.

yurishkuro: I think we can be more strict and only allow/require two options, never the original field names.

tigrannajaryan: > This reads as if the choice of the field name can vary from encoder to encoder.

Yes, this is exactly the case. (Some?) Protobuf encoders have an option that allows to choose whether the original names or lowerCamelCase is used when encoding. This option applies to all fields in the encoded messages and sub-messages. See for example `OrigName` boolean setting in Go's Marshaler: https://pkg.go.dev/github.com/golang/protobuf@v1.5.2/jsonpb#Marshaler

The `json_name` option you refer to is a per-field decorator that allows to specify a custom name just for a particular field. When `json_name` decorator is specified for the field then it will be used instead of lowerCamelCase (when lowerCamelCase _encoder_ option is chosen). In OTLP we don't use `json_name` option anywhere, so this is not relevant for OTLP.

tigrannajaryan: > I think we can be more strict and only allow/require two options, never the original field names.

I open to considering this. In discussion https://github.com/open-telemetry/opentelemetry-specification/issues/2795 @dyladan preferred the flexibility of being able to use either option in the exporters.

dyladan: I don't feel strongly but is there any value in restricting?

tigrannajaryan: > I don't feel strongly but is there any value in restricting?

If we restrict to only one of the options then receiver implementations can be simpler. We already support both [in the Collector](https://github.com/open-telemetry/opentelemetry-collector/blob/65ea4f0d82645a4b0e1a1acfe039cffe26c0a346/pdata/plog/internal/plogjson/json.go#L94) and it is not a big deal, so I don't this is a high cost to pay if we believe it is a useful convenience for senders.

yurishkuro: The value in restricting is the stability of the wire format. Since it's JSON, we must assume it may be consumed by tools that are not Protobuf-based, so allowing multiple field names is going to be just painful for consumers.

The way I read the Protobuf spec:
  * the spec provides NO variability in naming for producers, it's completely deterministic: if override is given via `json_name` then use that, otherwise use lowerCamelCase. So given a proto file, the field names are fixed.
  * those Go marshallers that @tigrannajaryan mentioned are apparently non-compliant (when OrigName option is enabled), and that's probably why the fallback to original name is called out for receivers. Since I assume most Proto libraries are actually able of being compliant for producers (bonkus options notwithstanding), there seems to be little benefit for allowing that behavior in OTLP JSON.

tigrannajaryan: I don't think Go marshaler is non-compliant. The [spec](https://developers.google.com/protocol-buffers/docs/proto3#json) says that such an option may be provided by implementation:

>... By default proto3 JSON printer should convert the field name to lowerCamelCase and use that as the JSON name. **An implementation may provide an option to use proto field name as the JSON name instead.** Proto3 JSON parsers are required to accept both the converted lowerCamelCase name and the proto field name.

That said, I don't mind restricting to one of the options (i.e. provide no choice in OTLP). The current default behavior I believe is what we should choose in that case, so we will support lowerCamelCase only.

Aneurysm9: I think that we need to accept either representation.  The proto3 spec says that producers may make this choice and we cannot control what producers give us, so Postel's Law tells us that we need to support both.

Aneurysm9 (approved):

specification/protocol/otlp.md
tigrannajaryan: @yurishkuro we discussed this in the Spec SIG meeting and didn't hear any other arguments in favour of only allowing one of the names.

It seems like we have the following cons and pros for support both key names:
Pros
- Matches default behavior of Protobuf JSON implementations. Don't need to carve out yet another exception, which we believe we should avoid unless there are strong reasons to do so.
- Existing implementation may already be using both options. Disallowing this would be a breaking change. Although it is explicitly allowed breaking change (because OTLP/JSON is not Stable) it is still a pain that if possible we would like to avoid causing

Cons
- Receiver implementations must correctly support parsing both key names. If you use an existing library you get this for free, but if you implement manually then it is extra work (although not difficult to do - we already do it in the Collector).
- Support 2 names seems like unnecessary complication since it doesn't allow anything that is also not possible with the more limited approach where only one of the names is allowed. Nevertheless, it is already a design choice made by Protobuf. Our deviation from this design choice is also a complication, it doesn't come from free (we need to document it and change Collector implementation).

I don't know which of these arguments are stronger.

Maybe there are other arguments that can help to make a decision.

yurishkuro: I disagree with the pros.

> Pro: Matches default behavior of Protobuf JSON implementations.

No, it does not. Default behavior is not to use original name. Proto spec says (https://developers.google.com/protocol-buffers/docs/proto3#json_options):

> *Use proto field name instead of lowerCamelCase name:* By default proto3 JSON printer should convert the field name to lowerCamelCase and use that as the JSON name.

The ability to send original name is not a default behavior, it's an optional behavior that OTEL exporters would need to explicitly opt-in into, which they do not have to do (and it's even worse if they make it an option for end-user to set).

> Pro: Existing implementation may already be using both options. Disallowing this would be a breaking change.

I expect it should be easy to check if any existing implementations actually opt-in into that non-default behavior (how many of them are even using JSON format, probably just JS in the browser?) Don't know about breaking change for previously unregulated behavior - in the absolute sense, probably, but if existing SDKs don't use that then it seems reasonable to restrict.

> Nevertheless, it is already a design choice made by Protobuf.

Don't know how much it is a design choice rather than legacy fallback. OTEL, on the other hand, (a) does not have legacy requirements, and (b) does not need to be bound by Proto decisions. We reuse Proto's JSON spec for convenience, but if it doesn't cost us anything to define a clean JSON format, then we don't need to carve out exceptions to the format the way Proto had to for legacy reasons.

Let me suggest an alternative way to resolve it: if someone is objecting to stricter specification, can they provide specific evidence why dual-naming support is absolutely required?  I think the burden should be on these niche use cases to prove that they deserve a complication of the protocol.

Aneurysm9: > The ability to send original name is not a default behavior, it's an optional behavior that OTEL exporters would need to explicitly opt-in into, which they do not have to do (and it's even worse if they make it an option for end-user to set).

This seems to assume that we're the only implementers of OTLP exporters.  That's not a safe assumption.  Someone coming along and attempting to implement OTLP independently may make that choice.  This is again a case of Postel's Law telling us that we should be prepared to accept data in this manner even if we would never choose to emit it.

>We reuse Proto's JSON spec for convenience, but if it doesn't cost us anything to define a clean JSON format, then we don't need to carve out exceptions to the format the way Proto had to for legacy reasons.

I think this would be a case of carving out a different set of exceptions.  Protobuf has already made those exceptions and now they are the _status quo_ to be excepted from.  If we're not intending to use Protobuf then we're not bound by the decisions Protobuf has made, but we do so intend and thus are so bound.

yurishkuro: > Someone coming along and attempting to implement OTLP independently may make that choice.

Not if we state explicitly in the spec that the choice is not allowed.

> Protobuf has already made those exceptions and now they are the status quo to be excepted from.

Like I said, they did it for legacy reasons. We do not have these legacy requirements that we need to support.

Why should we make our ecosystem _harder to use_ just because Protobuf had to make that trade-off? The whole point of JSON format is that it does not require Protobuf libraries to use the data. It's not just us accepting the data, I can see tons of tools in the future, like Jupyter notebooks in Python, or various UIs, that could be using OTEL JSON for reading the data. We're going to burden all those potential users with this dual-naming requirement for no good reason.

yurishkuro (approved):

jmacd (approved):

dyladan (approved):

carlosalberto (approved):

jsuereth (approved):

tigrannajaryan:
Let's give this another 24 hours and then merge if there are no objections.

CHANGELOG.md
```diff
@@ -78,6 +78,8 @@ release.
 
 ### OpenTelemetry Protocol
 
+- Clarify that lowerCamelCase field names MUST be used for OTLP/JSON
```
carlosalberto, line 81: Oops, this changelog entry is outdated (we already released 1.14, etc)

tigrannajaryan: Fixed.

tigrannajaryan:
Has enough approvals. No objections for several days. Merging.

PR #2831, opened 2022-09-23
Link to RPC in Metrics Semantic Conventions

oppegard:
## Changes

I'm proposing this change because of the poor discoverability of RPC metrics on the generated website: https://opentelemetry.io/docs/reference/specification/metrics/semantic_conventions/

Should the RPC file also be renamed to `rpc-metrics.md` to match the others?


MrAlias (approved):

joaopgrassi (approved):
+1 for renaming it to `rpc-metrics.md` for the sake of consistency. We could also consider opening an issue on the website repo so the lack of discoverability is fixed there as well :)

reyang (approved):

oppegard:
I renamed `rpc.md` to `rpc-metrics.md`. If we don't have agreement, let me know and I can revert the change.

PR #2835, opened 2022-09-26
Contributing Geo fields from ECS

### LINKED ISSUE #2834: Add Geo fields from Elastic Common Schema

Some observability use cases require localization of resources. Examples are:

- Observability of mobile devices: Geographically localizing the mobile devices provides valuable insights into how mobile apps perform depending on their location. 
- For applications that are geographically distributed across multiple regions and data centers, the geo location can provide useful insights as well.

Having an additional `geo` namespace on the resource attributes serves use cases as outlined above and allows to filter and group data by geo locations. 

As part of the [OTEP to support the Elastic Common Schema (ECS)](https://github.com/open-telemetry/oteps/pull/199), we propose to start with adopting the rich set of the ECS geo fields as a new namespace in the resource attributes. 

**Additional context.**

This is related to this OTEP: https://github.com/open-telemetry/oteps/pull/199

AlexanderWert:
Fixes #2834

## Changes
Adds `geo` namespace from the Elastic Common Schema to the resource attributes to allow filtering and grouping of observability data by geo location.  

Related OTEP: https://github.com/open-telemetry/oteps/pull/199

### DIFF
--- a/specification/resource/semantic_conventions/geo.md
+++ b/specification/resource/semantic_conventions/geo.md
@@ -1,0 +1,38 @@
+# Geo
+
+**Status**: [Experimental](../../document-status.md)
+
+**type:** `geo`
+
+**Description**: Geo fields can carry data about a specific location related to a resource or event. This geolocation information can be derived from techniques such as Geo IP, or be user-supplied.
+
+<!-- semconv geo -->
+| Attribute  | Type | Description  | Examples  | Requirement Level |
+|---|---|---|---|---|
+| `geo.city_name` | string | City name. | `Montreal`; `Berlin` | Recommended |
+| `geo.continent_code` | string | Two-letter code representing continent’s name. | `AF` | Recommended |
+| `geo.continent_name` | string | Name of the continent. | `North America`; `Europe` | Recommended |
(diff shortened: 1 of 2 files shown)


specification/resource/semantic_conventions/geo.md
```diff
@@ -0,0 +1,38 @@
...
+| Value  | Description |
+|---|---|
+| `AF` | Africa |
+| `AN` | Antarctica |
+| `AS` | Asia |
+| `EU` | Europe |
+| `NA` | North America |
+| `OC` | Oceania |
+| `SA` | South America |
+<!-- endsemconv -->
```
scheler, line 38: 1. For mobile devices, the geo attributes will be changing as the user moves around and they cannot be updated on the `Resource` currently. This PR will have to wait until https://github.com/open-telemetry/oteps/pull/208 is resolved or there's another mechanism introduced that allows changing attributes on a Resource.
2. For apps running in data centers (and mobile apps), I believe the customers will have to run a Geo mapping service to map the IP to the geo coordinates. If for some reason this service is unavailable, there should be a way to pass the original source IP in the `Resource` so that the telemetry service can use that for the geo mapping. Can you include support for that in this PR or add a reference to it if it's already there?

PR #2732, opened 2022-08-16
Add OpenCensus metric bridge specification

dashpole:
Part of https://github.com/open-telemetry/opentelemetry-specification/issues/1175

Related to https://github.com/open-telemetry/opentelemetry-specification/pull/2730

Depends on https://github.com/open-telemetry/opentelemetry-specification/pull/2722.  Do not merge until after that PR.

## Changes

Specify that OpenCensus metric bridges implement the MetricProducer interface, and describe non-trivial transformations involved.

This follows the path taken by the [java opencensus-shim](https://github.com/open-telemetry/opentelemetry-java/blob/main/opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/metrics/OpenCensusMetricProducer.java#L27).

PR #2838, opened 2022-09-26
Add MetricProducer as a source of external metric data

dashpole:

Related to https://github.com/open-telemetry/opentelemetry-specification/pull/2730

Required for https://github.com/open-telemetry/opentelemetry-specification/pull/2732

Supersedes https://github.com/open-telemetry/opentelemetry-specification/pull/2722.  The numbering scheme for alternatives caused some confusion, and it was [requested](https://github.com/open-telemetry/opentelemetry-specification/pull/2722#issuecomment-1258424607) that I open a new PR.

## Changes

Add a MetricProducer interface to the metrics SDK. It is an optional argument when creating a MetricReader meant to support non-SDK sources of metric data.

Metric data from a MetricProducer is not subject to the MeterProviders or MetricReaders configuration for instruments.  This means Views on MeterProviders, and default Aggregations and Aggregation Temporalities configured on MetricReaders are not applied to data from MetricProducers.

This is needed to support an OpenCensus metric bridge, which is [proposed separately](https://github.com/open-telemetry/opentelemetry-specification/pull/2732).  It could also be used to support other bridges as well, such as a Prometheus bridge.

### DIFF
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -47,6 +47,10 @@ linkTitle: SDK
       - [ForceFlush()](#forceflush)
       - [Shutdown()](#shutdown)
   * [Pull Metric Exporter](#pull-metric-exporter)
+- [MetricProducer](#metricproducer)
+  * [Interface Definition](#interface-definition-1)
+    + [Produce() metrics](#produce-metrics)
+    + [InstrumentationScope() InstrumentationScope](#instrumentationscope-instrumentationscope)
 - [Defaults and configuration](#defaults-and-configuration)
 - [Numerical limits handling](#numerical-limits-handling)
 - [Compatibility requirements](#compatibility-requirements)
@@ -62,7 +66,8 @@ linkTitle: SDK
 
 A `MeterProvider` MUST provide a way to allow a [Resource](../resource/sdk.md) to
 be specified. If a `Resource` is specified, it SHOULD be associated with all the
-metrics produced by any `Meter` from the `MeterProvider`. The [tracing SDK
+metrics produced by any `Meter` from the `MeterProvider` or produced by any
+`MetricProducer`. The [tracing SDK
 specification](../trace/sdk.md#additional-span-interfaces) has provided some
 suggestions regarding how to implement this efficiently.
 
@@ -76,10 +81,16 @@ an [`InstrumentationScope`](../glossary.md#instrumentation-scope) instance which
 is stored on the created `Meter`.
 
 Configuration (i.e., [MetricExporters](#metricexporter),
-[MetricReaders](#metricreader) and [Views](#view)) MUST be managed solely by the
-`MeterProvider` and the SDK MUST provide a way to configure all options that are
-implemented by the SDK. This MAY be done at the time of MeterProvider creation
-if appropriate.
+[MetricReaders](#metricreader), [MetricProducers](#metricproducer) and
+[Views](#view)) MUST be managed solely by the `MeterProvider` and the SDK MUST
+provide a way to configure all options that are implemented by the SDK. This
+MAY be done at the time of MeterProvider creation if appropriate.
+
+If multiple [MetricProducers](#metricproducer) have the same
+InstrumenatationScope the SDK SHOULD emit a warning. If a meter is created
+which produces an [`InstrumentationScope`](../glossary.md#instrumentation-scope)
+identical to the InstrumentationScope of a [MetricProducer](#metricproducer),
+the SDK should emit a warning.
 
 The `MeterProvider` MAY provide methods to update the configuration. If
 configuration is updated (e.g., adding a `MetricReader`), the updated
@@ -136,7 +147,9 @@ instances.
 ### View
 
 A `View` provides SDK users with the flexibility to customize the metrics that
-are output by the SDK. Here are some examples when a `View` might be needed:
+are output by the SDK from Measurements made by OpenTelemetry
+[Instruments](./api.md#instrument). Here are some examples when a `View` might
+be needed:
 
 * Customize which [Instruments](./api.md#instrument) are to be
   processed/ignored. For example, an [instrumented
@@ -769,13 +782,18 @@ used with pull-based metrics collection.  A common sub-class of
 `MetricReader`, the periodic exporting `MetricReader` SHOULD be provided
 to be used typically with push-based metrics collection.
 
-The `MetricReader` MUST ensure that data points are output in the
-configured aggregation temporality for each instrument kind.  For
-synchronous instruments being output with Cumulative temporality, this
-means converting [Delta to Cumulative](supplementary-guidelines.md#synchronous-example-cumulative-aggregation-temporality)
+The `MetricReader` MUST ensure that data points from OpenTelemetry
+[instruments](./api.md#instrument) are output in the configured aggregation
+temporality for each instrument kind. For synchronous instruments being output
+with Cumulative temporality, this means converting
+[Delta to Cumulative](supplementary-guidelines.md#synchronous-example-cumulative-aggregation-temporality)
 aggregation temporality.  For asynchronous instruments being output
 with Delta temporality, this means converting [Cumulative to
 Delta](supplementary-guidelines.md#asynchronous-example-delta-temporality) aggregation
+temporality.
+
+The `MetricReader` is not required to ensure data points from a
+[MetricProducer](#metricproducer) are output in the configured aggregation
 temporality.
 
 The SDK MUST support multiple `MetricReader` instances to be registered on the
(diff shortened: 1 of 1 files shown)


dashpole:
cc @jsuereth @bogdandrutu @jmacd @reyang @pirgeo @MadVikingGod @MrAlias @jack-berg

jmacd:
Thank you @dashpole. I think resetting the review state here could help us improve this feature. I would like us now to also consider the "ALTERNATIVE" proposal known as option number two in #2722. 

I re-read this PR and think with small changes we could have the alternative approach (which I now favor). Whereas you wrote:

```text
+-----------------+            +--------------+
|                 | Metrics... |              |
| MeterProvider   +------------> MetricReader |
|                 |            |              |
+---------------^^+            +--------------+
|  MetricProducer |
+-----------------+
```

The alternative looks like one (or more) MetricProducers in a pipeline between the Reader and the Exporter. 

```text
+-----------------+                 +--------------+              +----------------+
|                 | Internal data   |              | Export data  |                | To exporter...
| MeterProvider   +-----------------+ MetricReader +--------------+ MetricProducer +---------------->
|                 |                 |              |              |                |
+-----------------+                 +--------------+              +----------------+
```

Most of the PR doesn't have to change much. MetricReaders no longer care, what's left is a question about trade-offs.

- In the alternative proposal, MetricProducers have to be configured for each Reader instead of once per SDK: SDKs could perhaps alleviate this with a helper to make adding producers to all readers easy.
- In the alternative proposal, errors are deferred until the first time metrics are read by the exporter: this seems acceptable.

@dashpole what do you think?

bogdandrutu:
@jmacd you mentioned that

```
When used with distributed tracing, a resource can be associated 
with the [TracerProvider](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#tracerprovider) 
when the TracerProvider is created. That association cannot be changed later. 
When associated with a TracerProvider, all Spans produced by any Tracer 
from the provider MUST be associated with this Resource.

Analogous to distributed tracing, when used with metrics,
a resource can be associated with a `MeterProvider`.
When associated with a [`MeterProvider`](../metrics/api.md#meterprovider),
```

> It's that "analogous" MUST we are defending here. It appears the stable Java SDK has a loophole, and now we have wrinkles in the specification because of it. I recommend @dashpole remove all the wrinkle about Resources "if it is possible for those to conflict", leave the specification stating that each producer has an instrumentation scope, and let the Java SDK document its loophole (i.e., it should document that users are required to use the same resource twice or else be out-of-spec).

I don't think the specification says that resource MUST only be associated the providers. Out of all the concepts so far that is true, and the statement you linked tells how that association must be implemented, but I don't see anywhere where it is forbidden to have other sources of telemetry with their own "Resource".

Because of that I still believe the version where MetricProvider implementing the Reader is ok.

MrAlias (approved):

jmacd:
> Because of that I still believe the version where MetricProvider implementing the Reader is ok.

Not sure what you're trying to say. I feel that this is a new topic. I quoted part of the Resource SDK specification. Above the part I quoted, 

```
A [Resource](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/overview.md#resources) is an immutable representation of the entity producing telemetry as [Attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#attribute).
```

If we allow the Bridge to configure its own Resource, there's a good chance the bridge will appear as a separate Resource meaning a different entity. I believe the under-specified spirit of the Resource SDK Specification is that we should not encourage a bridge implementation that uses different Resource attributes than the OTel MeterProvider it is associated with.

dashpole:
I think its slightly confusing for a MetricProducer to receive metrics from a MetricReader.  Would this:

```
+-----------------+                 +--------------+            
|                 | Internal data   |              | To exporter... 
| MeterProvider   +-----------------+ MetricReader +------------->
|                 |                 |              |
+-----------------+                 +------|-------+              
                                  +--------+-------+
                                  |                | 
                                  + MetricProducer +
                                  |                |
                                  +----------------+
```

instead of this:
```
+-----------------+                 +--------------+              +----------------+
|                 | Internal data   |              | Export data  |                | To exporter...
| MeterProvider   +-----------------+ MetricReader +--------------+ MetricProducer +---------------->
|                 |                 |              |              |                |
+-----------------+                 +--------------+              +----------------+
```
 be a reasonable approach?

reyang:
> ```
> +-----------------+                 +--------------+              +----------------+
> |                 | Internal data   |              | Export data  |                | To exporter...
> | MeterProvider   +-----------------+ MetricReader +--------------+ MetricProducer +---------------->
> |                 |                 |              |              |                |
> +-----------------+                 +--------------+              +----------------+
> ```

I think there is value of having this component and it should be called "MetricProcessor". Calling it "MetricProducer" will be problematic if later we want to introduce MetricProcessor (and we'll struggle with the ordering of these components in the pipeline).

jsuereth (approved):

reyang:
> I think its slightly confusing for a MetricProducer to receive metrics from a MetricReader. Would this:
> 
> ```
> +-----------------+                 +--------------+            
> |                 | Internal data   |              | To exporter... 
> | MeterProvider   +-----------------+ MetricReader +------------->
> |                 |                 |              |
> +-----------------+                 +------|-------+              
>                                   +--------+-------+
>                                   |                | 
>                                   + MetricProducer +
>                                   |                |
>                                   +----------------+
> ```

How about this?

```text
+-----------------+                 +--------------+            
|                 | Internal data   |              | To exporter... 
| MeterProvider   +-----------------> MetricReader +------------->
|                 |                 |              |
+-----------------+                 +------^-------+              
                                           |
+-----------------+                        |
|                 | Bridged data           |
| MetricWriter    +------------------------+
|                 |
+-----------------+
```

dyladan:
> How about this?
> 
> ```
> +-----------------+                 +--------------+            
> |                 | Internal data   |              | To exporter... 
> | MeterProvider   +-----------------> MetricReader +------------->
> |                 |                 |              |
> +-----------------+                 +------^-------+              
>                                            |
> +-----------------+                        |
> |                 | Bridged data           |
> | MetricWriter    +------------------------+
> |                 |
> +-----------------+
> ```
> 
> And here is my reason:
> 
> * `MetricProducer` sounds like a broad term, it can also mean a combination of "MeterProvider" and "MetricReader".
> * `MetricWriter` makes it clear that this component is intended for some external systems that handle aggregation already, and use this interface to write/feed the pre-aggregated metrics to the reader (which limits the control of several settings, such as aggregation temporality, view, histogram buckets).

As long as we're discussing naming `MetricWriter` sounds like something that persists data to a disk or something like that. Maybe `MetricBridge` since that's what it's actually being used for?

reyang:
> As long as we're discussing naming `MetricWriter` sounds like something that persists data to a disk or something like that. Maybe `MetricBridge` since that's what it's actually being used for?

+1 👍

dashpole:
Overall, i'd like to settle on Integrating with the MetricReader vs MeterProvider question before we dive into other details.  I'm happy to revisit naming, interfaces, or other aspects after that.

I would summarize the tradeoffs as:

Bridge integrates with MeterProvider (:tada:):
* [Simpler configuration UX](https://github.com/open-telemetry/opentelemetry-specification/pull/2722#discussion_r957513196): Passing the bridge to the MeterProvider is all that needs to be done. No configuration is required for/with each metric reader.

Bridge integrates with MetricReader (:rocket:):
* [More logically consistent](https://github.com/open-telemetry/opentelemetry-specification/pull/2722#issuecomment-1251386971): It is clear that View configuration of the MeterProvider does not apply to Bridges.  It isn't possible for us to make View configuration apply to Bridges because it is based on Instrument, rather than Aggregation.
* [Less confusion for unsupported Temporalities if using multiple Readers](https://github.com/open-telemetry/opentelemetry-specification/pull/2722#issuecomment-1258360580):  When using multiple readers, users might be confused if some readers don't support the temporality of metrics from the bridge, while others do.  This integration more explicitly links exporters and bridges.

Orthogonal-issues that I don't believe impact the decision above:
* It is possible to share resource regardless of where this integration happens.  We will also have to tackle the issue with how java represents resource regardless of which approach we take.
* It is possible to share interfaces with the SDK to simplify implementation regardless of where the integration happens.  I don't believe either implementation will be significantly more complex than the other.
* Depending on the interface we choose for the Bridge, is possible to detect duplicate scopes between bridges/SDK on initialization of meters/producers, or at collection time, or not at all.  This also isn't impacted by our choice of where we integrate.

Feel free to propose benefits/drawbacks of the approaches that i've missed.  I'd like to take the temperature of the room with :tada: or :rocket: reactions to this comment.

reyang:
+1 on "Bridge integrates with MetricReader" - I think most users won't directly interact with the bridge, I would imagine that OpenCensus (or Micrometer, whatever) Bridge to be packaged under some distro/family-pack, so folks can use it easily during the migration, without having to understand all the implementation details.

dashpole:
Yeah, I'm still working on it.  But I don't have any updates to share right now.

dyladan:
I just mean it should be reopened. Closed PRs tend to be lost

PR #2841, opened 2022-09-27
semantic conventions: add structured stacktrace to exception attributes

### LINKED ISSUE #2839: Structured Stacktrace in Span Attributes/Logs

A backwards compatible way of having stacktraces in spans and logs be structured data instead of strings. This would allow backends to parse and display usable (like links to the line in a file) information without having to know how to parse a specific languages stacktrace -- and not leave those of us using less popular languages on the outside looking in on these features :)

**Additional context.**

I'd open this for discussion as a PR but the semantic convention generation tool does not appear to accept a type array-of-maps that is needed for the attribute value. I don't want to spend my time on adding this to the tool if this idea is shot down.

The idea is to add an additional field to the exception semantic convention `exception.structured_stacktrace` with type `map<string, any>[]` and specific keys in the map representing the parts of a stack frame. Example:

```
[{"function_name": "com.example.GenerateTrace.methodB", 
  "filename": "GenerateTrace.java", 
  "line_number": 13}, 
 {"function_name": "com.example.GenerateTrace.methodA", 
  "filename": "GenerateTrace.java", 
  "line_number": 9}, 
 {"function_name": "com.example.GenerateTrace.main", 
  "filename": "GenerateTrace.java", 
  "line_number": 5}]'
```

Other optional keys would also be defined like `column_number`, [see the OpenCensus StackFrame proto](https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/trace/v1/trace.proto#L348). 

This example is the same as the one used currently in the [semantic convention for exceptions](https://github.com/open-telemetry/opentelemetry-specification/blob/0ff1c36708af773f1ed8b978946959471ffa6234/specification/trace/semantic_conventions/exceptions.md).

tsloughter:
Fixes #2839

## Changes

Add structured stacktrace to exception attributes. The stacktrace is broken up to 4 attributes: `exception.structured_stacktrace.function_names`,  `exception.structured_stacktrace.filenames`,  `exception.structured_stacktrace.line_numbers`, `exception.structured_stacktrace.column_numbers`

This may be premature to open but I thought a PR showing the suggested changes and open for comment would help discussion.

There are other fields that may be worth including as well that were in the stacktrace proto of OpenCensus https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/trace/v1/trace.proto#L348

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,11 @@ release.
 - Add `process.context_switches`, and `process.open_file_descriptors`, to the
   metrics semantic conventions
   ([#2706](https://github.com/open-telemetry/opentelemetry-specification/pull/2706))
+- Add structured stacktrace to exception attributes. The stacktrace is broken up
+  to 4 attributes: `exception.structured_stacktrace.function_names`,
+  `exception.structured_stacktrace.filenames`,
+  `exception.structured_stacktrace.line_numbers`,
+  `exception.structured_stacktrace.column_numbers` ([#2841](https://github.com/open-telemetry/opentelemetry-specification/pull/2841))
 
 ### Compatibility
 
--- a/semantic_conventions/trace/exception.yaml
+++ b/semantic_conventions/trace/exception.yaml
@@ -26,6 +26,30 @@ groups:
         at com.example.GenerateTrace.methodB(GenerateTrace.java:13)\n
         at com.example.GenerateTrace.methodA(GenerateTrace.java:9)\n
         at com.example.GenerateTrace.main(GenerateTrace.java:5)'
+      - id: structured_stacktrace.function_names
+        type: string[]
+        brief: >
+          The fully-qualified names that uniquely identify the function or
+          method that is active in the frame.
+        examples: ["com.example.GenerateTrace.methodB", "com.example.GenerateTrace.methodA",
+        "com.example.GenerateTrace.main"]
+        note: |-
+          The `structured_stacktrace` attribute takes precedence over `stacktrace` when both appear. Each array must start with the top of the stack.
+      - id: structured_stacktrace.filenames
+        type: string[]
+        brief: >
+          The source file names where each function call appears.
+        examples: ["GenerateTrace.java", "GenerateTrace.java", "GenerateTrace.java"]
+      - id: structured_stacktrace.line_numbers
+        type: int[]
+        brief: >
+          The line number in file where each function call appears.
+        examples: [13, 9, 5]
+      - id: structured_stacktrace.column_numbers
+        type: int[]
+        brief: >
+          The column number, if available, in each file where the function call appears.
+        examples: [8, 4, 4]
       - id: escaped
         type: boolean
         brief: >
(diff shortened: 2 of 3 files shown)


tigrannajaryan:
I remember structured vs plaintext was discussed at length when the exception semantic conventions were first introduced. I would advise to try to find the participants of that past discussion and ask their opinion. The discussion should have left some trail in the PRs, spec SIG meeting notes, Slack, etc.

CHANGELOG.md
```diff
@@ -29,6 +29,11 @@ release.
 - Add `process.context_switches`, and `process.open_file_descriptors`, to the
   metrics semantic conventions
   ([#2706](https://github.com/open-telemetry/opentelemetry-specification/pull/2706))
+- Add structured stacktrace to exception attributes. The stacktrace is broken up
+  to 4 attributes: `exception.structured_stacktrace.function_names`,
```
tigrannajaryan, line 33: In the past few months I think we have seen multiple use cases where we want attributes with map values. I would argue that if such value are widely needed then we need to lift the restriction and allow map values. Vendors who don't support map values can flatten the data in their exporters.

tsloughter: Agreed. I've wanted to also find the discussion/notes around why this was left out. Was it simply that some vendors don't support it?

tigrannajaryan: Here is the issue: https://github.com/open-telemetry/opentelemetry-specification/issues/376

tsloughter: Ah, that issue. Thanks, I saw you include this PR in the comments already.

jmacd: I support map-valued attributes (and specifying how to flatten them). I see how we could use semantic conventions for conveying source location in stacktraces, but this seems to me a very expensive option. Looking at other fields the OpenCensus stacktrace message contained, I find this mechanism to avoid repetitive stacktraces:

```
  // The hash ID is used to conserve network bandwidth for duplicate
  // stack traces within a single trace.
  //
  // Often multiple spans will have identical stack traces.
  // The first occurrence of a stack trace should contain both
  // `stack_frames` and a value in `stack_trace_hash_id`.
  //
  // Subsequent spans within the same request can refer
  // to that stack trace by setting only `stack_trace_hash_id`.
  //
  // TODO: describe how to deal with the case where stack_trace_hash_id is
  // zero because it was not set.
  uint64 stack_trace_hash_id = 2;
```

And this is where most of my concerns here lie -- a structured stack is a more-expensive way to encode stacktraces than the simple string, unless you have a way to avoid repetition.

tsloughter: Right. We could do similar to the OpenCensus method of not repeating parts of stacktraces through the use of a hash id. But it is also still an optional field that doesn't need to be filled in for those languages that the vendors can parse. Instead of asking vendors to parse languages outside a certain subset we only ask that they fallback on a single structure all the unsupported languages will use.

It is a bit of a pain to say which languages need to be sending the structured version, so some may want to make it a user defined option, except for major players like Java, .net and Go who can pretty much guarantee support.

PR #2844, opened 2022-09-28
Add `process.memory.utilization` metric to semantic conventions

### LINKED ISSUE #2811: Add `process.memory.utilization` metric

I want to add a `process.memory.utilization` metric to the semantic conventions. The value would be the amount of physical memory used by the process (`process.memory.usage` according to the convention) divided by the total amount of memory available on the host.

**Additional context.**

Here's a related proposal to add this metric in `hostmetrics` receiver: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/14084.

A similar metric already exists in the `system` namespace: `system.memory.utilization`. It has the `state` attribute, which of course the `process.memory.utilization` metric would not have.

I'm a bit worried that the total memory available on the host does not have a name in the convention, but that would be a separate issue.

astencel-sumo:
Fixes #2811

## Changes

Adds a new metric `process.memory.utilization` to the semantic conventions.

## Context


specification/metrics/semantic_conventions/process-metrics.md
```diff
@@ -36,6 +36,7 @@ Below is a table of Process metric instruments.
...
 | `process.memory.usage`                 | UpDownCounter                                      | By        | The amount of physical memory in use.                                                                                               |                                                                                                                                                                                                 |
+| `process.memory.utilization`           | Gauge                                              | 1         | Percentage of total physical memory on the system that is used by the process.                                                      |                                                                                                                                                                                                 |
```
bogdandrutu, line 39: any label/attribute for this metric?

astencel-sumo: Hm no, I don't think so. Looking at the [`gopsutil` implementation](https://github.com/shirou/gopsutil/blob/v3.22.9/process/process.go#L318-L336), a process's memory utilization is just the amount of physical memory that the process uses (RSS) divided by the total amount of RAM available on the system.

I know that `gopsutil` is not necessarily a determinant of what can or should be implemented, but I think this single value is what's most interesting to the user.

jsuereth: If we wanted to provide other usage numbers, how would this scale to account for them?

Also, should this be a derived metric from `process.memory.usage` and another metric outlining total host memory?

astencel-sumo: @bogdandrutu @jsuereth does this make sense to you? Any further comments?

astencel-sumo: @jsuereth @bogdandrutu Please take a look.

jsuereth: The issue in my mind is still that folks look at SemConv as a "minimum" list of requirements, not a set of options.

1. If this is indeed just an optional thing that could be filled out, add a comment about how OTEL prefers raw usage numbers and this is just an option for interacting with legacy / non-OTEL-first systems.
2. I don't think your view on SemConv being "possible things" vs. "requirements list" is shared by all who read this.  That's a higher-level discussion we can/should have.   I actually think the later is more common mindset than the former.

So, I understand your argument about how this is a possible thing for legacy systems, but I'd like to see that in the PR itself, as a comment or some other denotation for when to choose/use utilization vs. other mechanism.

astencel-sumo: Oh, I guess I had some assumptions that are not necessarily true. Thanks for pointing this out Josh. This probably means I would need to change the way I think about semantic conventions.

Going back to this specific metric and PR, there already exist `system.memory.usage` and `system.memory.utilization` metrics in the semantic conventions. I was hoping that adding `process.memory.utilization` to the convention would be one of the easier things to add.

Issue #2726, opened 2022-08-12
Mistake while applying the logic in #2617 from #2589: process.network.io and system.network.* should not have been split


bogdandrutu:
The issue https://github.com/open-telemetry/opentelemetry-specification/issues/2589 explains why the direction for `disk` is not the right thing, which makes total sense.

But we got that blindly and apply for every "attribute" called "direction" without discussing if that makes sense or not, lucky I thing that is not yet released, and hopefully not implemented by users. As an example the logic in #2589 does not apply for `network` metrics, why did we deprecate that? It makes total sense to know the total `network` bytes.

PS: If you have an argument why I did not say anything, is because in June/July I was PTO, so ...

tigrannajaryan:
@bogdandrutu just to clarify, are you saying https://github.com/open-telemetry/opentelemetry-specification/pull/2617  should be reverted? That PR split `process.disk.io` and `process.network.io` by eliminating the `direction` attribute. Do you think both of this metric splitting are wrong?

jmacd:
In the discussion surrounding this issue, we debated the usefulness and meaningfulness of either having or not having a `direction` attribute vs having or not having independent metrics. The reason I supported this "change of direction" 😁 is that we have a strong precedent in existing instrumentation favoring separate metrics. 

Migration into OTel from all of the old metrics technologies, where precedent favors separate metrics, was looking difficult. We concluded that the meaningfulness was outweighed by a strong precedent and low utility. If the question boils down to "how often do you find yourself plotting the sum of directions?", the answer was "not often". If the question boils down to "how much value is there in binding the two directions together in one metric?" the answer was "not much".

Since the PR in question merged, I found another potential reason to oppose the "change of direction".  If we're going to review the arguments, here's another one. In [OTEP 176 Columnar Encoding for OTLP](https://github.com/open-telemetry/oteps/pull/171) specifically [this section](https://github.com/open-telemetry/oteps/pull/171/files#diff-8b2a4811d420878fd569924408084d8731e0915142a33cc1bfeb1242441247c1R462), @lquerel observes that in a multi-variate metrics system, _it makes a lot of sense to maintain the connection that is being lost in #2617_. In the example used in the OTEP, it's the `state` attribute that is considered significant. 

> For each of these states, the metrics share the same attributes, timestamp, ... Taken individually, these metrics don't make much sense. Knowing the free memory without knowing the used memory or the total memory is not very informative.

[...]

PR #2845, opened 2022-09-28
Add `process.disk.operations` metric to semantic conventions

### LINKED ISSUE #2812: Add `process.disk.operations` metric

I want to add a `process.disk.operations` metric to the semantic conventions, akin to the existing `system.disk.operations` metric. The metric should have a `direction` attribute with the value of `read` or `write`.

The description of the metric is "The number of disk operations performed by the process". 

**Additional context.**

Here's a related proposal to add this metric in `hostmetrics` receiver: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/14084.

astencel-sumo:
Fixes #2812

## Changes

Adds a new metric `process.disk.operations` to the semantic conventions.

## Context


jsuereth (approved):

astencel-sumo:
What needs to happen for this pull request to be merged? (Other than me updating the branch again)

jack-berg (approved):

specification/metrics/semantic_conventions/process-metrics.md
```diff
@@ -38,6 +38,7 @@ Below is a table of Process metric instruments.
...
 | `process.disk.io`               | Counter                                            | By        | Disk bytes transferred.                                                                                                             | `direction` SHOULD be one of: `read`, `write`                                                                                                                                                   |
+| `process.disk.operations`       | Counter                                            | {operations} | Number of disk operations performed by the process                                                                  | `direction` SHOULD be one of: `read`, `write`                                                                                                                                                                   |
```
reyang, line 41: What about "control" (e.g. seek, S.M.A.R.T. signal)?

reyang (changes requested):
https://github.com/open-telemetry/opentelemetry-specification/pull/2845/files#r1009798522

specification/metrics/semantic_conventions/process-metrics.md
portertech: I like how this is inline with `process.disk.io`, w/ direction being either `read` or `write`. I suspect SMART capabilities would need their own metric(s). I don't how to address seek.

reyang: There are lots and back-and-forth, e.g. #2617.

astencel-sumo: @reyang I do recognize that this has been a much-discussed topic:

- the "direction" attribute was removed in favor of separate metric names https://github.com/open-telemetry/opentelemetry-specification/issues/2589
- the change was then reverted https://github.com/open-telemetry/opentelemetry-specification/issues/2726

The current state is that metrics very similar to the proposed `process.disk.operations` metric exist in the semantic conventions in the same shape as the proposed metric (with the "direction" attribute of either "read" or "write"):
- [system.disk.io](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.14.0/specification/metrics/semantic_conventions/system-metrics.md#systemdisk---disk-controller-metrics) has a [system.disk.operations](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.14.0/specification/metrics/semantic_conventions/system-metrics.md#systemdisk---disk-controller-metrics) counterpart
- [process.disk.io](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.14.0/specification/metrics/semantic_conventions/process-metrics.md#process) does not have such counterpart

This proposal adds the `process.disk.operations` metric that is related to `process.disk.io` in the same was that the `system.disk.operations` is related to `system.disk.io` metric. I believe this is not a controversial change and there's no reason to not add this metric in the current shape.

If you want the discuss the shape of these metrics, I'd say it's better to open a separate issue.

reyang: > I believe this is not a controversial change and there's no reason to not add this metric in the current shape.

I have different understanding - we already know the existing ones are wrong and will take breaking changes before they can be marked stable, it'll be better to avoid adding more things that require a fix.

PR #2846, opened 2022-09-28
Add `process.signals_pending` metric to semantic conventions

### LINKED ISSUE #2826: Add `process.signals_pending` metric to semantic conventions

I want to add a `process.signals_pending` metric to the semantic conventions.

**Additional context.**

If I understand correctly, this metric is useful to see if/when a process stopped accepting signals, which can be useful in investigating issues.

Telegraf exposes this metric in its [procstat](https://github.com/influxdata/telegraf/blob/v1.24.1/plugins/inputs/procstat/README.md) input plugin. Adding a semantic convention for this metric in OT should make migration from Telegraf to OpenTelemetry easier.

Related proposal to add this metric in the `hostmetrics` receiver: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/14084.

astencel-sumo:
Fixes #2826 

## Changes

Adds a new metric `process.signals_pending` to the semantic conventions.

## Context


specification/metrics/semantic_conventions/process-metrics.md
```diff
@@ -42,6 +42,7 @@ Below is a table of Process metric instruments.
...
 | `process.context_switches`             | Counter                                            | {count}   | Number of times the process has been context switched.                                                                              | `type` SHOULD be one of: `involuntary`, `voluntary`                                                                                                                                             |
+| `process.signals_pending`              | UpDownCounter                                      | {signals} | Number of pending signals for the process.                                                                                          |                                                                                                                                                                                                 |
```
jsuereth, line 45: Signals are generally POSIX only.  Can you call out how this applies to non-POSIX systems (should they adapt signal-like things here?

Specifically:

- Is this *required* for POSIX systems or just recommended
- What does this mean for non-POSIX systems (or it shouldn't be filled out on those)

astencel-sumo: Yes, signals are a POSIX-only concept. I assume non-POSIX systems won't report this metric. I wouldn't say the metric is either required or recommended for any system (POSIX or not).

As I understand it, the OT specification only establishes a name for a specific information (a metric, an attribute, etc.), but does not specify weather a piece of information is required or recommended to be delivered by a system.

arminru, line 45:
```suggestion
| `process.signals_pending`              | UpDownCounter                                      | {signals} | Number of pending POSIX signals for the process.                                                                                    |                                                                                                                                                                                                 |
```
I think it would be helpful to explicitly call them "POSIX signals" here to distinguish them from "telemetry signals" which are used a lot in this domain.

astencel-sumo: Agree, updated the wording.

astencel-sumo: @jsuereth does this answer your questions? Do you think it makes sense?

jsuereth (approved):

jack-berg (approved):

specification/metrics/semantic_conventions/process-metrics.md
```diff
@@ -43,6 +43,7 @@ Below is a table of Process metric instruments.
...
 | `process.paging.faults`         | Counter                                            | {faults}  | Number of page faults the process has made.                                                                                         | `type`, if specified, SHOULD be one of: `major` (for major, or hard, page faults), `minor` (for minor, or soft, page faults).                                                                   |
+| `process.signals_pending`       | UpDownCounter                                      | {signals} | Number of pending POSIX signals for the process.                                                                                    |                                                                                                                                                                                                 |
```
reyang, line 46: What about non-POSIX systems? Should they use this same metrics or they should invent a different name?

reyang (changes requested):
https://github.com/open-telemetry/opentelemetry-specification/pull/2846/files#r1009799891

specification/metrics/semantic_conventions/process-metrics.md
portertech: I wouldn't expect non-POSIX systems to use the same metric (and I don't know what they would use). That said, if there is a valid use-case down the road, we can change the description to accommodate. We are considering description changes to be a stable change https://docs.google.com/document/d/1Nvcf1wio7nDUVcrXxVUN_f8MNmcs0OzVAZLvlth1lYY/edit#heading=h.e71xzqqz0b5t

reyang: > I wouldn't expect non-POSIX systems to use the same metric (and I don't know what they would use).

Would it make sense to put it as `process.posix.signals_pending`?

portertech: @jsuereth WDYT?

astencel-sumo: > > I wouldn't expect non-POSIX systems to use the same metric (and I don't know what they would use).
> 
> Would it make sense to put it as `process.posix.signals_pending`?

I think I understand where this comes from. As described in the metrics semantic conventions' [General Guidelines](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.14.0/specification/metrics/semantic_conventions#general-guidelines):

> Associated metrics SHOULD be nested together in a hierarchy based on their usage. Define a top-level hierarchy for common metric categories: for OS metrics, like CPU and network; for app runtimes, like GC internals. Libraries and frameworks should nest their metrics into a hierarchy as well. This aids in discovery and adhoc comparison. This allows a user to find similar metrics given a certain metric.

and further below:

> Semantic ambiguity SHOULD be avoided. Use prefixed metric names in cases where similar metrics have significantly different implementations across the breadth of all existing metrics. For example, every garbage collected runtime has slightly different strategies and measures. Using a single set of metric names for GC, not divided by the runtime, could create dissimilar comparisons and confusion for end users. (For example, prefer process.runtime.java.gc* over process.runtime.gc.*.) Measures of many operating system metrics are similarly ambiguous.

If I understand correctly, your point Reiley is that if a non-POSIX system has a different understanding of what a "signal" is, the metric `process.signals_pending` might have a different meaning, thus supporting the division into a `process.posix.signals_pending` metric a a `process.something-else-that-is-not-posix.signals_pending` or similar metric. This is similar to the `process.runtime.java.gc*` example in the documentation.

I have two concerns with this:
1. The `.posix.` namespace would be based on a feature of a system, not on a specific component, like the Java runtime, which in my view makes it less discoverable and harder to understand by users.
2. Most of the systems we run (definitely in the server world, given Linux dominance) are POSIX, which makes it the default setting. I'd be more in favor to give a separate namespace to a non-POSIX-specific concept, while keeping the POSIX-specific concepts - at least the most popular ones - in the default namespace.

reyang: > I have two concerns with this:
> 
> 1. The `.posix.` namespace would be based on a feature of a system, not on a specific component, like the Java runtime, which in my view makes it less discoverable and harder to understand by users.
> 2. Most of the systems we run (definitely in the server world, given Linux dominance) are POSIX, which makes it the default setting. I'd be more in favor to give a separate namespace to a non-POSIX-specific concept, while keeping the POSIX-specific concepts - at least the most popular ones - in the default namespace.

I think this requires more discussion, it's not a specific to this PR.

astencel-sumo: Good point. Unfortunately I wasn't able to join today's Specification SIG to put this under discussion.

Issue #2784, opened 2022-09-09
Add BatchLogRecordProcessor environment variable configuration


alanwest:
Much like the [batch span processor config](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#batch-span-processor). Need similar settings for the batch log record processor.

Open questions:

1. Naming: `OTEL_BLRP_*` or `OTEL_BLP_*`?
2. What should the default values be? I suspect the same defaults we use for the batch span processor are fine.

PR #2861, opened 2022-10-10
Add log attribute limit configuration

### LINKED ISSUE #2860: Clarify that attribute limit configuration also applies to LogRecords
The specification _kind of_ implies that attributes limits also apply to logs, but much of the spec is very span-centric. Add details of log-specific attribute limit configuration matching the trace specification.

alanwest:
Fixes #2860

Adds log attribute limit configuration.

These new environment variables bring more consistency between spans and logs. Therefore, I believe it should be exempt from the [moratorium on new env vars](https://github.com/open-telemetry/opentelemetry-specification/issues/2891#issuecomment-1289241503).


specification/logs/sdk.md
```diff
@@ -130,6 +131,36 @@ information
 that was added to the `LogRecord` (as with
 [ReadableLogRecord](#readablelogrecord)).
 
+## LogRecord Limits
+
+`LogRecord` attributes MUST adhere to the [common rules of attribute limits](../common/README.md#attribute-limits).
```
djaglowski, line 136: Do we need rules that address how to handle nested attributes? For example, how many attributes is this?

```json
"attributes": {
  "foo": "bar",
  "one_or_many": {
    "nested": {
      "a": "b",
      "c": "d"
    },
    "also_nested": "e"
  }
}
```

alanwest: Good question. I had previously assumed that the attributes for log records followed the [common definition of attributes](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/common#attribute)

Though, the [log data model spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#type-any) _does_ seem to suggest your example is a valid log record attribute. It even seems to suggest that resource attributes for a log record could be nested... this raises my eyebrows a bit 🤨.

Is this really intended? I'll bring to the log SIG this week.

alanwest: I knew that the OTLP data model supported nested attribute, but I guess I was not aware that the spec's log data model supported them. Also just discovered https://github.com/open-telemetry/opentelemetry-specification/pull/2620.

Here's what I propose. This PR addresses all the supported types that log records have in common with traces, metrics, etc. I think separate from this PR, rules of attribute limits should be defined for nested attributes. I can imagine other configuration settings that may uniquely be useful for nested attributes like a depth limit.

MSNev: I've also been attempting to clarify this and the eariler iterations of this PR #2620 had wording around having a "maximum" depth and I think this one #2888 is also related

djaglowski: > Here's what I propose. This PR addresses all the supported types that log records have in common with traces, metrics, etc. I think separate from this PR, rules of attribute limits should be defined for nested attributes. I can imagine other configuration settings that may uniquely be useful for nested attributes like a depth limit.

No objections from me.

specification/sdk-environment-variables.md
```diff
@@ -137,6 +137,17 @@ See the SDK [Span Limits](trace/sdk.md#span-limits) section for the definition o
...
+## LogRecord Limits
+
+**Status**: [Experimental](document-status.md)
+
+See the SDK [LogRecord Limits](logs/sdk.md#logrecord-limits) section for the definition of the limits.
+
+| Name                                        | Description                                | Default | Notes |
+| ------------------------------------------- | -------------------------------------------| ------- | ----- |
+| OTEL_LOGRECORD_ATTRIBUTE_VALUE_LENGTH_LIMIT | Maximum allowed attribute value size       |         | Empty value is treated as infinity. Non-integer and negative values are invalid. |
+| OTEL_LOGRECORD_ATTRIBUTE_COUNT_LIMIT        | Maximum allowed log record attribute count | 128     |       |
```
tigrannajaryan, line 149: I think we use LOGS elsewhere, not LOGRECORD, e.g. OTEL_EXPORTER_OTLP_LOGS_INSECURE

alanwest: I've struggled with knowing when we should use the term `Log` vs. `LogRecord` 😆.

I don't have a strong opinion either way. The span one is singular not plural i.e., `OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT` so if we want to change it, I'd suggest `OTEL_LOG_ATTRIBUTE_COUNT_LIMIT`.

I raised a similar question when introducing the batch processor environment variables #2784. We landed on using the prefix `OTEL_BLRP_*` instead of `OTEL_BLP_*`. For example `OTEL_BLRP_SCHEDULE_DELAY`. Should we consider changing these to use `BLP`?

jack-berg: I've been operating using the "Log" or "Logs" to refer to the log signal, and "LogRecord" to refer to individual log records.

So [OTEL_LOGS_EXPORTER](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#exporter-selection) is okay because its selecting the exporter for the log signal. Note that similarly the env var for traces is `OTEL_TRACES_EXPORTER` rather than `OTEL_SPANS_EXPORTER`, presumably in reference to the trace signal rather than individual span records.

`OTEL_BLRP_*` is aligned with the `OTEL_BSP_*` env vars, which reference span records not the trace signal.

And similarly, `OTEL_LOGRECORD_ATTRIBUTE_*` (or perhaps `OTEL_LOG_RECORD_ATTRIBUTE_`) would be aligned with `OTEL_SPAN_ATTRIBUTE_*` because we're referencing configuration for records, not for the signal.

jack-berg (approved):

specification/sdk-environment-variables.md
tigrannajaryan: Discussed in Log SIG and most opinions are that LOGRECORD is OK in the env var name.

specification/logs/sdk.md
```diff
@@ -130,6 +131,36 @@ information
...
+```
+
+**Configurable parameters:**
+
+* [all common options applicable to attributes](../common/README.md#configurable-parameters)
+
+There SHOULD be a log emitted to indicate to the user that an attribute was
+discarded due to such a limit. To prevent excessive logging, the log
+should be emitted at most once per log record (i.e., not per discarded
+attribute.)
```
arminru, line 162: It might be a bit confusing that we're talking about two entirely different kinds of logs or log streams here. One could interpret "emit a log" as "create/export a log record".

How about something like this to clarify it?

```suggestion
There SHOULD be a message printed in the SDK's log to indicate to the user
that an attribute was discarded due to such a limit.
To prevent excessive logging, the message should be printed at most once per
`LogRecord` (i.e., not per discarded attribute.)
```

This slightly diverges from what we have in the trace spec but has the meaning, so I think it should be fine, but we can also update the trace spec accordingly.

arminru (approved):

tigrannajaryan (approved):

specification/trace/sdk.md
```diff
@@ -377,23 +377,25 @@ Following configuration properties should be available when creating the sampler
...
-which then SHOULD be called `SpanLimits`. Implementations MAY provide additional
-configuration such as `AttributePerEventCountLimit` and `AttributePerLinkCountLimit`.
+The options MAY be bundled in a class, which then SHOULD be called `SpanLimits`.
+Implementations MAY provide additional configuration such as
+`AttributePerEventCountLimit` and `AttributePerLinkCountLimit`.
 
 ```java
 public final class SpanLimits {
-  SpanLimits(int attributeCountLimit, int linkCountLimit, int eventCountLimit);
+  SpanLimits(int attributeCountLimit, int attributeValueLengthLimit, int linkCountLimit, int eventCountLimit);
```
tigrannajaryan, line 393: I just noticed this change. This should be considered a spec bug, right? Not a breaking change.

jack-berg: Shouldn't be a breaking change because 1. This is just example code. 2. This parameter is one of the [common options](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/common#configurable-parameters) mentioned [here](https://github.com/open-telemetry/opentelemetry-specification/blame/main/specification/trace/sdk.md#L409). 

Separately, we might consider removing this constructor and switching this to be an interface because the both the current version of the constructor  on main and this version are missing argument (i.e. there are 4 arguments but 6 methods). Additionally, this code doesn't compile because the syntax for the constructor and methods is wrong. How about:
```
public interface SpanLimits {
  int getAttributeCountLimit();
  int getAttributeValueLengthLimit();
  int getAttributeCountPerEventLimit();
  int getAttributeCountPerLinkLimit();
  int getEventCountLimit();
  int getLinkCountLimit();
}
```

alanwest: Apologies, thought I had pushed up this change. Good to go now.

tigrannajaryan:
This includes changes to trace sdk file. Either fix the the PR description to explain why it is in one PR with the log limits or split into 2 separate PRs if unrelated.

tigrannajaryan:

alanwest (commented):
> This includes changes to trace sdk file. Either fix the the PR description to explain why it is in one PR with the log limits or split into 2 separate PRs if unrelated.

I'm fine splitting the changes to the trace SDK into a separate PR. Though, the corresponding sections of log and trace SDK should mirror each other. I think it makes sense to make the changes in trace and log specs in parallel to eliminate the possibility of continued divergence.

Below is my thought process for each of the changes I've made to the trace SDK.

specification/trace/sdk.md
```diff
@@ -414,7 +414,8 @@ public final class SpanLimits {
 
 There SHOULD be a log emitted to indicate to the user that an attribute, event,
 or link was discarded due to such a limit. To prevent excessive logging, the log
-should not be emitted once per span, or per discarded attribute, event, or links.
+should be emitted at most once per span (i.e., not per discarded attribute,
+event, or link).
```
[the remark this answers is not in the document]
tigrannajaryan: I agree it is was confusing. What you suggest is more clear, but I am not sure it is exactly what we want to do. I think this should go through a separate PR with its own changelog entry. I am worried that we will merge this with log changes and clarifications to traces will get lost in the noise.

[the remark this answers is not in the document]
alanwest: I've reverted all changes to `trace/sdk.md`. For now I also adopted this confusing statement about logging in the log SDK, as well. I will open a new issue for addressing this.

jack-berg (approved):

tigrannajaryan (approved):

specification/logs/sdk.md
```diff
@@ -127,6 +128,34 @@ information
...
+}
+```
+
+**Configurable parameters:**
+
+* [all common options applicable to attributes](../common/README.md#configurable-parameters)
+
+There SHOULD be a message printed in the SDK's log to indicate to the user
+that an attribute was discarded due to such a limit.
+To prevent excessive logging, the message should not be printed once per
```
MSNev, line 156: This wording seems backwards, and reads as a suggestion, perhaps re-phrase slightly.

There SHOULD be a message printed in the SDK's log to indicate that an attribute was discarded due to such limit. This message MUST NOT be repeated for every `LogRecord` or discarded attribute.

alanwest: > This wording seems backwards, and reads as a suggestion, perhaps re-phrase slightly.

I completely agree. We have a nearly identical statement in the trace SDK spec

https://github.com/open-telemetry/opentelemetry-specification/blob/094fe9f34bc2fdb9797c4437c6062dcf61760de0/specification/trace/sdk.md?plain=1#L415-L417

Previously, I had changed the wording both here and in the trace SDK spec to make it more clear, but per this conversation https://github.com/open-telemetry/opentelemetry-specification/pull/2861#discussion_r1073937079, I simply adopted the abysmal wording of the trace spec and will address both together in a separate PR.

specification/trace/sdk.md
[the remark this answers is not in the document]
alanwest: #3147

tigrannajaryan:
Has enough approvals, 2 days passed since last change, merging.

PR #2878, opened 2022-10-13
Clarify that Scope is defined at build time

tigrannajaryan:
I have seen several time a confusion around the nature of the scope and whether scope attributes can change at runtime.

The purpose of this PR is to make sure we all agree the scope is a build time concept, or if we disagree then explicitly specify what else the scope can denote and how its attribute can change at runtime.

### DIFF
--- a/specification/glossary.md
+++ b/specification/glossary.md
@@ -170,6 +170,12 @@ fully qualified class name).
 The instrumentation scope is used to obtain a
 [Tracer or Meter](#tracer-name--meter-name).
 
+The instrumentation scope may have zero or more additional attributes that provide
+additional information about the scope. For example for a scope that specifies an
+instrumentation library an additional attribute may be recorded to denote the URL of the
+repository URL the library's source code is stored. Since the scope is a build-time
+concept the attributes of the scope cannot change at runtime.
+
 ### Tracer Name / Meter Name
 
 This refers to the `name` and (optional) `version` arguments specified when


carlosalberto:
(Although maybe "application initialization time" is more accurate than build time, etc)

tigrannajaryan:
> (Although maybe "application initialization time" is more accurate than build time, etc)

I am actually going explicitly for "build time" since Scope is defined as "A logical unit of the application code". A unit of code is fully defined at build time. "application initialization time" would be different and would imply something like a `process.id` or `thread.id` is a valid Scope attribute, but I believe it is not and I want to prohibit it by this clarification.

Oberon00 (approved):
While technically I think the scope is implemented as a runtime entity in most languages, I think *conceptually* I agree it is clearly a build-time concept indeed.

specification/glossary.md
```diff
@@ -170,6 +170,12 @@ fully qualified class name).
 The instrumentation scope is used to obtain a
 [Tracer or Meter](#tracer-name--meter-name).
 
+The instrumentation scope may have zero or more additional attributes that provide
+additional information about the scope. For example for a scope that specifies an
+instrumentation library an additional attribute may be recorded to denote the URL of the
+repository URL the library's source code is stored. Since the scope is a build-time
+concept the attributes of the scope cannot change at runtime.
```
arminru, line 177: @tigrannajaryan 

> I am actually going explicitly for "build time" since Scope is defined as "A logical unit of the application code". A unit of code is fully defined at build time. "application initialization time" would be different and would imply something like a `process.id` or `thread.id` is a valid Scope attribute, but I believe it is not and I want to prohibit it by this clarification.

The glossary is not really a normative part of the spec, right? It might be necessary to also clarify this in the respective API spec sections.

tigrannajaryan: >The glossary is not really a normative part of the spec, right?

I think it is. Why not? In fact, I think it contains some rather important definitions. If we don't think it is normative we should definitely duplicate them elsewhere. But why would we do that? If it is not clear I would suggest to make it explicit that the definitions in the glossary are normative.

jack-berg (approved):

ahayworth:
> Since the scope is a build-time concept the attributes of the scope cannot change at runtime.

I think it's worth calling out that there are many languages that do not have a concept of "build time" in this way. For example, there is no equivalent concept of "build time" in Ruby; "application initialization" is the closest analogue we have. I understand the goals of the PR, but I do think we should find a way to define it that makes allowances for interpreted languages.

tigrannajaryan:
> > Since the scope is a build-time concept the attributes of the scope cannot change at runtime.
> 
> I think it's worth calling out that there are many languages that do not have a concept of "build time" in this way. For example, there is no equivalent concept of "build time" in Ruby; 

It is not about "building" in the "compiling" sense, it is about "building" in the "developer writing code" sense. This is applicable to any language. If it is not clear we can make this more explicit. 

>"application initialization" is the closest analogue we have.

No, that's exactly the opposite of what I suggest. "application initialization" happens at runtime.

ahayworth:
> It is not about "building" in the "compiling" sense, it is about "building" in the "developer writing code" sense. This is applicable to any language. If it is not clear we can make this more explicit.

> No, that's exactly the opposite of what I suggest. "application initialization" happens at runtime.

Using "build-time" to mean "the time at which an application developer writes code" is strange to me. Typically when I see "build-time" I equate it to "compile-time" (or an equivalent linking or packaging step, depending on the ecosystem in question). I like your suggestion of making that more explicit - that avoids a lot of ambiguity. And would have prevented my comments in the first place. 😄

specification/glossary.md
Oberon00: I don't think it is really normative. It's a glossary. And you also don't use MUST/SHOULD/MAY anywhere. Which in this case is probably fine since this is a broad guideline for how people should think about the scope.

reyang (approved):

specification/glossary.md
tigrannajaryan: Merging as is for now. We can decide in a future PR which parts of the glossary we need to move to the normative parts of the spec.

tigrannajaryan:
Has enough approvals. Merging. Further refinements can be done in a future PR.

Issue #2856, opened 2022-10-05
Clarify that the JSON log example is the spec is NOT the OTLP JSON format


wdengw:
**What are you trying to achieve?**
Find out what is the right JSON format. 

What did you expect to see?
The right format to follow

**Additional context.**
OTel log format document and its java implementation has a mismatch.  In the document,  [JSON log example](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#example-log-records), resource attributes are added directly under Resource tag.
```
{
  ...
  "Resource": {
    "service.name": "donut_shop",
    "service.version": "2.0.0",
    "k8s.pod.uid": "1138528c-c36e-11e9-a1a7-42010a800198",
  },
  ...
} 
```
In OpenTelemetry-java the [implementation](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java), Resource class holds Attributes, and resource attributes are added into Attributes.  Will the output be
```
{
  ...
  "Resource": {
    "Attributes": {
      "service.name": "donut_shop",
      "service.version": "2.0.0",
      "k8s.pod.uid": "1138528c-c36e-11e9-a1a7-42010a800198",
    },
  },
  ...
} 
```

Which is the correct one to follow?

Add any other context about the problem here. If you followed an existing documentation, please share the link to it.

reyang:
@wdengw during the triage meeting, we consider this as a Java implementation issue, we need your help to follow up with the Java SIG. Meanwhile, we agree that the spec should have some clarification that the current JSON example is not the OTLP JSON format, we'll repurpose this issue and make updates to the spec. Assigning to @tigrannajaryan.

jack-berg:
The java implementation of OTLP JSON is correct. This [test](https://github.com/open-telemetry/opentelemetry-java/blob/main/exporters/logging-otlp/src/test/java/io/opentelemetry/exporter/logging/otlp/OtlpJsonLoggingLogRecordExporterTest.java#L96-L160) demonstrates the OTLP JSON encoded of a complete payload. The referenced example in [Example Log Records](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#example-log-records) is NOT valid OTLP JSON format; however, it also doesn't claim to be OTLP JSON so I fail to see the problem.

wdengw:
[...]

PR #2885, opened 2022-10-17
Delete misleading JSON log example

tigrannajaryan:
Instead, link to the existing well-defined JSON file format.

Resolves https://github.com/open-telemetry/opentelemetry-specification/issues/2856


reyang (approved):

jack-berg (approved):

scheler:
@tigrannajaryan I think we should remove `Resource` and `InstrumentationScope` as fields in a LogRecord, see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#log-and-event-record-definition  - maybe you had Resource in LogRecords in the early versions and that's why the examples logRecord had Resource in it.

djaglowski (approved):

jsuereth (approved):

Issue #2869, opened 2022-10-11
Handle conflicting metric descriptions and types in prometheus exporter 


jack-berg:
With #2703 instrumentation scope name and version are included as attributes in the prometheus exporter, which helps to accommodate OpenTelemetry metric's ability to have instruments with the same name under different scopes. 

However, these instruments can still have conflicting types and descriptions, which present problems in prometheus scrapers as discussed in [open-telemetry/opentelemetry-java#4834](https://github.com/open-telemetry/opentelemetry-java/issues/4834).

The following scenario illustrates the problem:
- Scope name: `scope1`, instrument name: `my_instrument`, instrument type: `counter`, description: `description1`
- Scope name: `scope2`, instrument name: `my_instrument`, instrument type: `gauge`, description: `description2`

As of #2703, this will be serialized in the prometheus exporter as follows:
```
# HELP my_instrument description1
# TYPE my_instrument counter
my_instrument{otel_scope_name="scope1"} 1
# HELP my_instrument description2
# TYPE my_instrument gauge
my_instrument{otel_scope_name="scope2"} 1
```

One possible solution is to detect conflicting descriptions / types in the prometheus exporter and log an error and / or skip the offending metric. Views can be used to change the name of one of the offending instruments to avoid the conflict.

dashpole:
+1

OpenMetrics MetricFamily names (i.e. what type, description, unit, are associated with) [MUST be unique](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metricset) within the metrics served on an endpoint:

> Each MetricFamily name MUST be unique.

reyang:
Here is what I'm thinking:
1. This is not a unique problem in the exporter, we also have the same challenge when the Collector is trying to convert OTLP to Prometheus, and this is something @dashpole was trying to cover [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md#otlp-metric-points-to-prometheus)
2. For exporter, one approach is to make sure `SDK -> PrometheusExporter -> Prometheus` and `SDK -> OTLP Exporter -> Collector -> PromethuesExporter -> Prometheus` will end up having the same result, which sounds nice but I don't feel it is necessary.
[...]

PR #2890, opened 2022-10-18
Prevent duplicate Prometheus description comments

dashpole:
Fixes https://github.com/open-telemetry/opentelemetry-specification/issues/2869

Follow-up to https://github.com/open-telemetry/opentelemetry-specification/pull/2703.  That PR prevents collisions between data points, but not collisions between the descriptions of the metrics.

## OpenMetrics requirements

From https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metricfamily:
> A MetricFamily MUST have a name, HELP, TYPE, and UNIT metadata.
> MetricFamily names are a string and MUST be unique within a MetricSet.

Basically, for a given metric name, we MUST have exactly one HELP, TYPE, and UNIT comment.

## Changes

This is a bit light on details, but I can add more if people think that would be helpful.

Basically, this is saying that exporters need to deduplicate Prometheus "descriptions" when sending.  For TYPE conflicts, some metric has to be dropped, as the data sent might not be valid (e.g. you can't send counter information with a histogram TYPE).  For UNIT and HELP comments, it is up to the exporter how to merge, but the exporter MUST NOT keep both comments.

### Example approach

* Keep a map of metric name -> Description when constructing a batch of metrics for a scrape.
* Each time a new metric (with Description) is added, look up the name in the map:
    * If it isn't found, add it to the map
    * If it is found, and the desc is the same, do nothing
    * If it is found, and has conflicting type information, drop the metric + log a warning
    * If it is found, and has conflicting help or unit, use the stored help + unit instead of the provided one + log a warning.

### DIFF
--- a/specification/metrics/data-model.md
+++ b/specification/metrics/data-model.md
@@ -1469,9 +1469,12 @@ in keys).
 #### Metric Metadata
 
 Prometheus SDK exporters MUST NOT allow duplicate UNIT, HELP, or TYPE
-comments for the same metric name to be present on the Prometheus endpoint.
-Exporters MUST drop metrics to prevent conflicting TYPE comments, but
-SHOULD NOT drop metrics as a result of conflicting UNIT or HELP comments.
+comments for the same metric name to be returned in a single scrape of the
+Prometheus endpoint. Exporters MUST drop entire metrics to prevent conflicting
+TYPE comments, but SHOULD NOT drop metric points as a result of conflicting
+UNIT or HELP comments. Instead, all but one of the conflicting UNIT and HELP
+comments (but not metric points) SHOULD be dropped. If dropping a comment or
+metric points, the exporter SHOULD warn the user through error logging.
 
 The Name of an OTLP metric MUST be added as the
 [OpenMetrics MetricFamily Name](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metricfamily),
--- a/specification/metrics/data-model.md
+++ b/specification/metrics/data-model.md
@@ -1468,6 +1468,11 @@ in keys).
 
 #### Metric Metadata
 
+Prometheus SDK exporters MUST NOT allow duplicate UNIT, HELP, or TYPE
+comments for the same metric name to be present on the Prometheus endpoint.
+Exporters MUST drop metrics to prevent conflicting TYPE comments, but
+SHOULD NOT drop metrics as a result of conflicting UNIT or HELP comments.
+
 The Name of an OTLP metric MUST be added as the
 [OpenMetrics MetricFamily Name](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metricfamily),
 with unit and type suffixes added as described below. The metric name is
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ release.
 
 ### Metrics
 
+- Handle duplicate description comments during Prometheus conversion.
+  ([#2890](https://github.com/open-telemetry/opentelemetry-specification/pull/2890))
+
 ### Logs
 
 ### Resource


dashpole:
cc @jack-berg @reyang @open-telemetry/wg-prometheus

specification/metrics/data-model.md
```diff
@@ -1468,6 +1468,11 @@ in keys).
 
 #### Metric Metadata
 
+Prometheus SDK exporters MUST NOT allow duplicate UNIT, HELP, or TYPE
+comments for the same metric name to be present on the Prometheus endpoint.
+Exporters should drop entire metrics to prevent conflicting TYPE comments, but
```
MrAlias, line 1473:
```suggestion
Exporters SHOULD drop entire metrics to prevent conflicting TYPE comments, but
```

specification/metrics/data-model.md
```diff
@@ -1468,6 +1468,11 @@ in keys).
...
+SHOULD NOT drop metrics as a result of conflicting UNIT or HELP comments.
```
MrAlias, line 1474: Do we want prescriptive guidelines on what to do with the conflicting comments?

```suggestion
SHOULD NOT drop metrics as a result of conflicting UNIT or HELP comments. Instead,
the conflicting UNIT and/or HELP comment(s) SHOULD be dropped.
```

dashpole: It probably makes sense to drop conflicting unit comments.  I was thinking an exporter might decide to merge conflicting help comments

MrAlias: > It probably makes sense to drop conflicting unit comments. I was thinking an exporter might decide to merge conflicting help comments

SGTM :+1:

specification/metrics/data-model.md
```diff
@@ -1468,6 +1468,11 @@ in keys).
 
 #### Metric Metadata
 
+Prometheus SDK exporters MUST NOT allow duplicate UNIT, HELP, or TYPE
+comments for the same metric name to be present on the Prometheus endpoint.
+Exporters SHOULD drop entire metrics to prevent conflicting TYPE comments, but
+SHOULD NOT drop metrics as a result of conflicting UNIT or HELP comments.
+
```
jack-berg, line 1475: I think we should encourage SDK exporters to log a warning when these conditions are detected, and encourage use of the view API to resolve the conflict by renaming the offending metrics.

specification/metrics/data-model.md
```diff
@@ -1468,6 +1468,11 @@ in keys).
 
 #### Metric Metadata
 
+Prometheus SDK exporters MUST NOT allow duplicate UNIT, HELP, or TYPE
+comments for the same metric name to be present on the Prometheus endpoint.
```
reyang, line 1472: It will be important to clarify what does "present on the Prometheus endpoint" actually mean:
1) during the entire lifecycle of that endpoint, which could span across multiple HTTP scraping requests.
2) within a single scraping request

specification/metrics/data-model.md
reyang: +1

specification/metrics/data-model.md
dashpole: It now reads:

```
Exporters MUST drop entire metrics to prevent conflicting
TYPE comments, but SHOULD NOT drop metric points as a result of conflicting
UNIT or HELP comments. Instead, all but one of the conflicting UNIT and HELP
comments (but not metric points) SHOULD be dropped.
```

specification/metrics/data-model.md
dashpole: Added: `If dropping a comment or metric points, the exporter SHOULD warn the user through error logging.`

specification/metrics/data-model.md
dashpole: Changed to "to be returned in a single scrape of the Prometheus endpoint".

specification/metrics/data-model.md
Aneurysm9: Where/when should such warnings be emitted?  It feels like it would be best to do so when the offending instrument is created, but I don't think that's known as it depends on exporter and view configuration as well.  Emitting a warning every time a metric is dropped could get very verbose very quickly.

jack-berg (approved):

MrAlias (approved):

specification/metrics/data-model.md
jack-berg: I was planning on implementing the log in the prometheus exporter the first time the conflict is detected, and skipping on subsequent detections.

reyang: I'm a bit on the fence here, the "SHOULD warn the user through error logging" parts sounds like something for the SDK specification, not the data model specification.

I'll still approve this PR given the good intention + better clarity + the entire "Prometheus Compatibility" section is marked as Experimental.

reyang (approved):
LGTM although I'm a bit on the fence https://github.com/open-telemetry/opentelemetry-specification/pull/2890/files#r998590686.

specification/metrics/data-model.md
dashpole: I've been thinking this entire section should be moved under compatibility, TBH, rather than data model.

Aneurysm9 (approved):

specification/metrics/data-model.md
jsuereth: Agree on moving this to a new section, I'd just make sure to link from here.  I think originally this started as just a data model mapping specification and is becoming a mix between data model + exporter behavior.

reyang:
@dashpole would you update the changelog?

dashpole:
> @dashpole would you update the changelog?

Done

specification/metrics/data-model.md
reyang: > Agree on moving this to a new section, I'd just make sure to link from here. I think originally this started as just a data model mapping specification and is becoming a mix between data model + exporter behavior.

This can be a separate PR.

PR #2919, opened 2022-11-04
Rename built-in ExemplarFilters

cijothomas:
Renames the built-in ExemplarFilters to AlwaysOn, AlwaysOff, TraceBased. (inspired from the built-in samplers in trace spec)

Minor clarification about ExemplarFilters, that they only make measurements eligible for being Exemplars.

Fix links:
Currently, the set of built-in ExemplarFilters are not documented in the SDK page, but user is taken to a [link](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#defaults-and-configuration), which takes to yet another [link](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#metrics-sdk-configuration), which lists the available ExemplarFilters. This PR adds the built-in filter details in the SDK page itself, instead of only being documented in the environment variable page.


specification/metrics/sdk.md
```diff
@@ -658,8 +658,28 @@ This interface SHOULD have access to:
...
+Sampled here simply makes the measurement eligible for being included as an
+exemplar. `ExemplarReservoir` makes the final decision if a measurement becomes
+an exemplar.
+
+### Built-in ExemplarFilters
+
+OpenTelemetry supports a number of built-in exemplar filters to choose from.
+The default is `SampleWithTrace`.
+
+#### AlwaysSample
```
reyang, line 670: Given we're already using the term "sample", might be a good idea to align the terminology with traces https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#built-in-samplers.

Built-in ExemplarFilters:
* AlwaysOn
* AlwaysOff
* TraceBased

specification/sdk-environment-variables.md
```diff
@@ -266,9 +266,9 @@ Known values for `OTEL_LOGS_EXPORTER` are:
 
 Known values for `OTEL_METRICS_EXEMPLAR_FILTER` are:
 
-- `"none"`: No measurements are eligible for exemplar sampling.
-- `"all"`: All measurements are eligible for exemplar sampling.
-- `"with_sampled_trace"`: Only allow measurements with a sampled parent span in context.
+- `"none"`: [NeverSample](./metrics/sdk.md#neversample)
```
reyang, line 269: Similar to https://github.com/open-telemetry/opentelemetry-specification/pull/2919/files#r1014317441, "none" could be misleading ("none" filter sounds to me like "I just removed the air filter in my furnace", which is the same as "we're not going to filter out anything, all the particles will passthrough").

I would suggest "always_on" and "always_off".

specification/metrics/sdk.md
cijothomas: Yes agree! My preference is also to steal/borrow/inspire-from the Tracing Sampler Terminology :)

reyang (approved):
LGTM with a changelog entry.

specification/sdk-environment-variables.md
```diff
@@ -266,9 +266,9 @@ Known values for `OTEL_LOGS_EXPORTER` are:
 
 Known values for `OTEL_METRICS_EXEMPLAR_FILTER` are:
 
-- `"none"`: No measurements are eligible for exemplar sampling.
-- `"all"`: All measurements are eligible for exemplar sampling.
-- `"with_sampled_trace"`: Only allow measurements with a sampled parent span in context.
+- `"always_on"`: [AlwaysOn](./metrics/sdk.md#alwayson)
```
jsuereth, line 269: The only complaint I have (and why I deviated initially) is that an always on "filter" can still lead to no exemplars, depending on the reservoir chosen.

@jack-berg Is this a breaking change for Java metrics?

jsuereth (commented):
I'm ok aiming for consistency, but two questions raised -

1. Will this be confusing to users
2. Should we be doing renames during feature-freeze?  This causes churn / could break an SDK that has the feature implemented.  Just want to check impact before making that change.

specification/sdk-environment-variables.md
jack-berg: Its only a breaking change for the unstable [autoconfigure](https://github.com/open-telemetry/opentelemetry-java/tree/main/sdk-extensions/autoconfigure#exemplars) module which interpets environment variables. So not a breaking change.

cijothomas: _Sampled here simply makes the measurement eligible for being included as an
exemplar. `ExemplarReservoir` makes the final decision if a measurement becomes
an exemplar._

^ This wording was added to the ExemplarFilter to avoid that confusion, that ExemplarFilter is only making a measurement "eligible".

(I see it like an Election where a person is a candidate if x/y/z conditions meet (Filter) Whether they get elected depends on the voting...(Reservoir))

reyang: > The only complaint I have (and why I deviated initially) is that an always on "filter" can still lead to no exemplars, depending on the reservoir chosen.

I feel this is the same as trace/span samplers - even with always-on samplers, one can filter out all the spans in a processor and end up with nothing?

cijothomas: ^ yes. In the case of traces, one has to do something (like setup filter processor etc) to face this situation. For metric exemplar case, the default Reservoirs will be doing it w/o user doing any additional action. Hence I added the extra clarification note.

specification/metrics/sdk.md
```diff
@@ -658,8 +662,28 @@ This interface SHOULD have access to:
...
+#### AlwaysOn
+
+An ExemplarFilter which makes all measurements eligible for being an Exemplar.
+
+#### AlwaysOff
+
+An ExemplarFilter which makes no measurements eligible for being an Exemplar.
+Using this ExemplarFilter is as good as disabling Exemplar feature.
+
+#### SampleWithTrace
```
jmacd, line 683: I suggest "WhenSampled" would be a more direct explanation of this mode, also would make it grammatically consistent with AlwaysOn and AlwaysOff.

jsuereth: `WhenTraceSampled` ?

Somehow it should denote the interaction with Trace.

specification/sdk-environment-variables.md
jsuereth: Thanks for the updated wording.  This LGTM

jsuereth (approved):

jack-berg (approved):
I like the suggest to rename `SampledWithTrace` to `TraceBased` for consistency

specification/metrics/sdk.md
reyang:
```suggestion
#### TraceBased
```

specification/sdk-environment-variables.md
```diff
@@ -266,9 +266,9 @@ Known values for `OTEL_LOGS_EXPORTER` are:
...
+- `"always_off"`: [AlwaysOff](./metrics/sdk.md#alwaysoff)
+- `"sample_with_trace"`: [SampleWithTrace](./metrics/sdk.md#samplewithtrace)
```
reyang, line 271:
```suggestion
- `"trace_based"`: [TraceBased](./metrics/sdk.md#tracebased)
```

carlosalberto:
Reopening as it seems this was briefly forgotten, but has enough approvals to be merged. @cijothomas please consider addressing the latest bits so we can merge this ;)

specification/metrics/sdk.md
cijothomas: Changing to TraceBased based on multiple suggestions.

specification/metrics/sdk.md
```diff
@@ -658,8 +662,28 @@ This interface SHOULD have access to:
   [Span](../trace/api.md#span).
 - A `timestamp` that best represents when the measurement was taken.
 
-See [Defaults and Configuration](#defaults-and-configuration) for built-in
-filters.
+Sampled here simply makes the measurement eligible for being included as an
```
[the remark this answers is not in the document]
cijothomas: thanks.

cijothomas:
@reyang @jack-berg @jsuereth Please do a final check. "TraceBased" suggestion is done, and also added changelog.

jmacd (approved):

PR #2888, opened 2022-10-17
Support maps and heterogeneous arrays as attribute values

tigrannajaryan:
Resolves https://github.com/open-telemetry/opentelemetry-specification/issues/376

Use cases where this is necessary or useful:

1. Specify more than one resource in the telemetry: https://github.com/open-telemetry/opentelemetry-specification/issues/579
2. Data coming from external source, e.g. AWS Metadata: https://github.com/open-telemetry/opentelemetry-specification/pull/596#issuecomment-628983680 or https://github.com/open-telemetry/opentelemetry-specification/issues/376#issuecomment-1227501082
3. Capturing HTTP headers: https://github.com/open-telemetry/opentelemetry-specification/issues/376#issuecomment-908493520
4. Structured stack traces: https://github.com/open-telemetry/opentelemetry-specification/pull/2841
5. Payloads as attributes: https://github.com/open-telemetry/oteps/pull/219#discussion_r971065645

This is a draft PR to see what the change looks like.

If this PR is merged it will be nice to follow it up with:
- A standard way of flattening maps and nested objects when converting from OTLP to formats that don't support maps/nested objects.
- Recommendations for semantic conventions to use/not use complex objects.

PR #2926, opened 2022-11-07
Define the shape of a event by encapsulating the payload in event.data.

MSNev:
## Changes

This PR extends the event specification to define the "shape" of an event by encapsulating the payload (content) of the event in the `event.data` attribute.

This defines how events with a payload should be defined, this enables support for downstream systems to handle both known known domain-specific events (identified by `event.domain` and `event.name`) and any other arbitrary new or vender specific events.

Downstream systems can safely assume that when a Log record contains the `event.domain`, `event.name` and `event.data` that the Log record represents an event and can then perform any validation or storage (routing) changes that the vender may require.

This provides support not only for events being defined by the RUM Sig, but also includes support for the mapping of [CloudEvents](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md), which can be defined by a later PR.
Possible CloudEvents mapping Example:
- event.domain: "cloudevents"
- event.name: "xxxx"
- event.data: (The cloudEvents `data`)
With Cloud Events Context Attributes (Required and optional) being mapped to top level OpenTelemetry Attributes of the same name

There was a discussion around using the "Body" of a log message as the transport for the payload instead of defining an `event.data`, however, while the Logs is intended to be the primary transport mechanism events the definition of an event is not limited to Logs only. This allows for an event to be transmitted as a span event (which doesn't have a "Body") and also supports direct logical mapping for [CloudEvents](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md).


specification/logs/semantic_conventions/events.md
```diff
@@ -7,11 +7,23 @@ by `LogRecord`s. All standalone Events have a name and a domain. The Event domai
...
+provide additional context about the `event`, whether an Attribute is required
+or optional SHOULD be defined by the domain-specific schema definition of the
+`event`.
+
 <!-- semconv event -->
 | Attribute  | Type | Description  | Examples  | Requirement Level |
 |---|---|---|---|---|
 | `event.name` | string | The name identifies the event. | `click`; `exception` | Required |
 | `event.domain` | string | The domain identifies the context in which an event happened. An event name is unique only within a domain. [1] | `browser` | Required |
+| `event.data` | [any](../data-model.md#type-any) \| [map<string, any>](../data-model.md#type-mapstring-any) | The domain-specific `payload` of the `event` which provides details about the occurrence of the named event. | `{ connectStart: 100, connectEnd: 103, domainLookupStart: 80, domainLookupEnd: 90 }` | Optional |
```
ramthi, line 26: why 'any' type at all?  I thought data should always be a map!

tigrannajaryan, line 26: Would it not be more natural to stored this in the LogRecord `Body` field? Do we expect `Body` to contain something else instead or be empty?

MSNev: This was primarily to provide potential support for Cloud Events, where the `data` may be binary encoded. So not necessarily for RUM events but for support of other domain's that may wish to have something different.

MSNev: `Body` technically could be used, but the primary (expected) use case (from a RUM sig perspective) would be to have this as a `map<string, any>`. So for backends to be able to "process" the value then we would (or should) define an optional attribute to provide the ability to identify the encoding type (ie. something like https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#datacontenttype)

tigrannajaryan: `Body` is of type `any` that can also be a `map<string, any>`, so not sure what the problem is.

tigrannajaryan: @MSNev can you summarize in the PR description what we discussed in the SIG meeting regarding Body vs event.data? It will useful to have the reasoning not to use the Body in the PR for posterity.

specification/logs/semantic_conventions/events.md
```diff
@@ -7,11 +7,23 @@ by `LogRecord`s. All standalone Events have a name and a domain. The Event domai
 is a namespace for event names and is used as a mechanism to avoid conflicts of
 event names.
 
+Events MAY include domain-specific information about the occurrence (`payload`)
+of the `event`, when present this information MUST be included in the `event.data`
+attribute the values included SHOULD conform to the domain-specific defined schema
+for the identified `event`. This is the primary container for details about the
+`event`.
+
+Events MAY also include one or more additional Attributes which can be used to
```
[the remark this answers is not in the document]
scheler: The additional attributes can only describe event metadata though, that is, it should be unfit to put those attributes in event.data.

ronyis:
I think that log events should also support representing the payload of application requests. For example, in order for a user to collect the response body of an outgoing HTTP request, saving it as part of a span is problematic since the body could be read after the span representing the request is closed (many times it's also read asynchronously).

Using a `map<string, any>` is not generic enough for this example, using `any` is preferable.

I'm also in favor of using the log `Body` instead of an attribute. For the case of using a Span Event, I think we should be able to support an `any` type attribute as well (similar to as proposed in https://github.com/open-telemetry/opentelemetry-specification/pull/2888).

scheler:
not stale

scheler:
@tigrannajaryan I used the term control attributes during the SIG meeting today. CloudEvents uses the term `context attributes` for the attributes that are  separate from event `data` and represent context metadata. Some examples of these attributes are - 

1. id - unique instance id
2. schema uri - schema for the payload content in `event.data`
3. export priority - used to determine if the event must exported right away skipping batch export

Note none of these are in the spec currently, but we anticipate their need in the future. and they clearly are separate from the event payload. Also, for the purpose of schema validation against the schema uri it will be helpful to isolate the payload.

We could introduce these control/context attributes in future as the need arises, and this PR paves the way for it.

MSNev:
Not stale

MSNev:
not stale

MSNev:
Related https://github.com/open-telemetry/opentelemetry-specification/issues/2994

jack-berg:
One of the main criticisms of complex attribute types from the Java SIG is the ambiguity in when to use flattened attributes vs complex attribute alternative. (IMO, this has been resolved after [this addition](https://github.com/open-telemetry/opentelemetry-specification/pull/2888#issuecomment-1424709307)) The critique is that having additional API surface area without clear guidance on when to use it is a bad user experience.

I see that same idea popping up here: If I'm a user calling the Event API to produce custom events, when / why should I include data in `event.data` vs in top level attributes?

scheler:
@MSNev @jack-berg 

I thought about this before but was not sure how to restrict the usage of `other_attributes`. We could try and model this the same way as CloudEvents API, which doesn't accept a generic collection for top level attributes (they are called context attributes in CE) but only have explicit setters for each of the pre-defined context attributes mentioned in the spec.

Here's an example usage of the CE java api - 

```java
final CloudEvent event = CloudEventBuilder.v1()
    .withId("000")
    .withType("example.demo")
    .withSource(URI.create("http://example.com"))
    .withData("text/plain","Hello world!".getBytes("UTF-8"))
    .build();
```

What do you think of taking a similar approach in the OTel Events API?  The top level attributes so far are only `event.name` and `event.domain` and both of them are already set through other means in the API. So, I suggest we remove the `other_attributes` parameter and add setters each top level attribute we add the OTel Events specification.

MSNev:
> I see that same idea popping up here: If I'm a user calling the Event API to produce custom events, when / why should I include data in event.data vs in top level attributes?

Ok, I believe I understand the concern. I'll add some wording to attempt to clarify that the `event.data` is ONLY for the properties associated for the specific event (domain/name combination) and how user Custom events should follow the same pattern so that backends that treat the `event.data` as the core data for any event received.

MSNev:
> So, I suggest we remove the other_attributes parameter and add setters for each top level attribute we add to the OTel Events specification in future.

With the separation of the Log / Event API, we need to change the Event API, so that it's more lightweight and doesn't require the creation of an "Event" that is passed to the Logging API (whatever it's name ends up being). 

Instead the event API should be lightweight and only contain some simply `emit` style function(s) that take either a pre-created Event instance or a single function that takes the required arguments, languages should not be required to implement the API as a builder pattern.

So something like
```ts
class OTelEventAPI {
    /**
     * Construct and emit a named event for the domain associated with the current instance.
     * @param name - The name of the event to emit, will be sent as in `event.name` 
     * @param data - The detail / payload of the event, the attributes should only contain the named attributes for the 
     * semantic conventions (definition) of the event identified by the domain / name.
     * @additional - Optional additional attributes that the application may pass 
     */
    public emit(name: string, data: Attributes, additional?: Attributes);

    /**
    * Emit the pre-constructed event to the domain associated with the current instance
    * @param theEvent - The pre-created event
    */
    public emit(theEvent: OTelEvent);
}
```

I know that there are lots of outstanding questions for the above like
- How do you construct the event
- What about the event domain, what if you create an event for the wrong domain and send it. Simplistically, this will NOT be blocked. But if we combine the domain/name as proposed by https://github.com/open-telemetry/opentelemetry-specification/issues/2994 this problem no longer exists.
- When should the additional be used (will attempt to clarify in this PR as above)
- How will the conventions for a named domain/name event be validated -- it won't (on the client), but backends (may) perform validation of events based on the published (not done yet) OpenTelemetry supported / defined domain/name combinations. 
- User/Application Custom Events, they can be passed through this API without any issues and backends can choose what to do with an "unknown" `Event` (identified as the `domain` / `name` / `data` combination) definition. ie. Validate based on their own definition, store as a named event with the event detail taken from `event.data` or send to some other user table.
- This seems incomplete -- yes it is, its for general discussion on the "Shape" of the Event not an attempt to identify what the entire API surface might look like. That discussion should occur as part of another PR to update the existing [Event API spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/event-api.md) etc.

specification/logs/semantic_conventions/events.md
```diff
@@ -42,4 +53,11 @@ unrelated events.
 | `browser` | Events from browser apps |
 | `device` | Events from mobile apps |
 | `k8s` | Events from Kubernetes |
-<!-- endsemconv -->
+
+`event.data` contains the domain-specific information about the occurrence of the
```
MSNev, line 57: TODO (MSNev): Add more details around how the content of `event.data` will following defined semantic conventions based on the define event "name" (currently a domain/name combination) and how this will work for custom (application / user (non-OpenTelemetry defined)) events

jack-berg:
The problems that this solves are too abstract for me support at this time. In particular, it's not clear how downstream validation of `event.data` would work and why including the event payload as top level attributes prevents this. There's a chance of collision, but is this a theoretical problem or one we've seen in practice? From a validation standpoint, a receiver of events in possession of a schema should be able to validate the known fields according to the schema, ignoring unrecognized fields. It's free to drop or accept unrecognized fields needed. This appears to be true whether or not the payload is encapsulated in a single field.

Based off discussions in the 2/22/23 log sig, this change seems important to the client instrumentation working group. If this inclusion is in fact blocking progress, can a demo be given which shows concrete situations where including event data as top level attributes fails?

MSNev:
> In particular, it's not clear how downstream validation of event.data would work

I have attempted to not restrict or define how (or if) downstream validation would occur as some vendors will and some wont. And others may choose to provide validation as a development (non-production) client side validation tool.

> and why including the event payload as top level attributes prevents this.

Simplistically, It doesn't. If someone wants to construct a "Custom" event which only uses top-level attributes they could, as part of this specification definition I'm saying they SHOULDN'T because of the chance of collisions and to simplify how back-end systems can identify, process and store "Events". (Which can also be viewed as a specific Structured Log records). There are additional reasons why specifying that event detail is passed in a single attribute (Which I'll include below -- I explicitly left out of the spec detail). This is also the one of the primary reasons for choosing to send "events" via Logs as it IS the only signal that supports nested attributes. Otherwise, we could have just continued to hack zero-length Span's with span Event (which has additional issues)

>There's a chance of collision, but is this a theoretical problem or one we've seen in practice?

Because many of the RUM events are defined by scraping and passing data based on existing W3C standards (at least for browser events), these names are fixed and as they evolve additional names WILL (and have) be added. The only way to avoid the collisions would be to prefix EVERY field which has performance, memory and payload considerations (all of which are major issues for client runtimes, whether a browser, mobile device, IOT or embedded into "SMART Tv's". (Yes, I've supported this)

## Why use a single attribute ##

Without encapsulating the event payload into a single nested attribute, the following (or combination) would need to occur, whether to avoid collisions or not.
- prefix every "value" with a common prefix (eg `event.<name / data>.xxxx`) this will cause an explosion with the payload size, especially when the event data is large like [W3C resource timing fields](https://developer.mozilla.org/en-US/docs/Web/Performance/Navigation_and_resource_timings) (these are used for page load as well Ajax requests and will include the "resources" (scripts, css, images) loaded as part of page load)
- Each "new" attribute name should be "added" to semantic conventions to avoid using the same name for different things / semantics (or be prefixed), this will slow down the event definition even further than it already is and would be unnecessary when the fields are only required for a single event. We have a specific Goal is to AVOID defining individual attribute names, we explicitly want to define the  semantic conventions (collection of value / names) at the "Event" level only.
- Generally RUM systems can have MANY (10s, 100s) of specific events, and it's not just a "Span" / "Log", each event has a specific meaning for monitor real user interactions with a system. And depending on the application (or observability vendor) there may have many/many/many custom events (beyond what we will define as part of the standard OpenTelemetry Events) and will need to be supported.

I've mentioned payload size before, but let me restate here, during the "Page/Application unload" (user navigates to another page, closes the browser, switches tabs and on mobile devices turns the device (screen) off - forcing the app to hibernation (and potentially never wake up)), so at this point any SDK should attempt to send all observability content (as it may not have any persistence capability or even a chance to re-hydrate and send (eg. classic examples of this are crashes).

During this "unload" state modern browsers have a restricted set of API's to allow the "delivery" of any network requests, both of these have a MAXIMUM payload of 64kb (of total outstanding (unsent) requests -- you can't just send multiple 64kb payloads you can send 1 total or several which add up to 64kb), and while there is currently an experimental compression API (it's not supported on Firefox) and it appears that it can't be supported by the primary sending API (because we can't include headers informing the receiver that the HTTP request is compressed (gzipped)), likewise on unsupported browsers including the additional client side code (which increases the page load time) also doesn't help, because you still can't add the additional header telling the receiver that the payload is compressed....
-

MSNev:
Not stale

MSNev:
Waiting for outcome of https://github.com/open-telemetry/opentelemetry-specification/issues/3406

Issue #2798, opened 2022-09-13
Add examples to OTLP/JSON


tigrannajaryan:
OTLP/JSON is a human readable format. It would be nice to have some example payloads in the OTLP specification to demonstrate what it can look like. Examples sometimes are very useful when you want to understand the structure of the payload.

See for example a user asking for it: https://github.com/open-telemetry/opentelemetry-specification/issues/2773#issuecomment-1245573473

abitrolly:
Zipkin also shows how it would look like https://zipkin.io/pages/data_model.html

sebastien-rosset:
How are timestamps supposed to be serialized when exporting OpenTelemetry metrics in JSON format? Is it nanoseconds since epoch or RFC 3339? There seems to be conflicting guidelines in the spec:
1. [OpenTelemetry file exporter](https://opentelemetry.io/docs/reference/specification/protocol/file-exporter/) provides examples where timestamps are serialized as the number of nanoseconds since epoch, e.g. "endTimeUnixNano":"1581452773000000789" - The file is https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/file-exporter.md
3. In the [JSON protobuf encoding](https://opentelemetry.io/docs/reference/specification/protocol/otlp/#json-protobuf-encoding), there is a link to https://protobuf.dev/programming-guides/proto3/#json, which states timestamps should be encoded as RFC 3339, e.g., "1972-01-01T10:00:20.021Z"

tigrannajaryan:
> How are timestamps supposed to be serialized when exporting OpenTelemetry metrics in JSON format? Is it nanoseconds since epoch or RFC 3339?

It is not RFC 3339. See the time_unix_nano field definition [here](https://github.com/open-telemetry/opentelemetry-proto/blob/eb214e0ceab7a595a5e690dc725b54442705204f/opentelemetry/proto/metrics/v1/metrics.proto#L354):
```
  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January
  // 1970.
```

> There seems to be conflicting guidelines in the spec:
> 
[...]

Issue #2773, opened 2022-09-06
Spec format for spans in build logs and bash script output


abitrolly:
**What are you trying to achieve?**

Given a build log https://gitlab.com/gitlab-org/gitlab-docs/-/jobs/2983094851 I want to quickly see what took the most time in build process. I can modify GitLab code to render spans, but I need a uniform way to report span timings in output stream so that tools can parse this in a universal way.

What did you expect to see?

A human readable and machine readable one liners to output span info.

One possibility is breadcrumb syntax.
```
I <span name> start |
| <span name> end | <time>s |
```
And an example.
```
| minifying js start |
| minifying js end | 144.18s |
```

**Additional context.**

Add any other context about the problem here. If you followed an existing documentation, please share the link to it.

Could not find any diagrams, and/or examples of spans in any representation format.

abitrolly:
@arminru you are the expert in this repo, so I don't know where it belong. I just came here with the problem that I need to extract telemetry from build logs. This belongs to specification, because without the spec it is hard to persuade people to format the messages in a suitable way for a wider range of tools, rather than my personal script.

arminru:
@abitrolly So you're not looking for a particular format to be used by the logging/console/stdout exporters (already) built into most OpenTelemetry SDK (e.g. the `ConsoleSpanExporter` for OTel JS) and the OpenTelemetry Collector but rather for a general format to be suggested which third party tools could follow?

abitrolly:
@arminru if there is a particular standardized format that I can reuse, I am all interested in that. But I could not find anything here.

abitrolly:
> (e.g. the `ConsoleSpanExporter` for OTel JS)

I see that it is a class https://open-telemetry.github.io/opentelemetry-js/classes/_opentelemetry_sdk_trace_base.ConsoleSpanExporter.html and not a format, so no examples, "prints spans", but no stated goal that the spans are actually machine readable. So I don't see it is a suitable lead.

arminru:
@abitrolly Indeed. I think all of the logging/console/stdout exporters in OTel are intended to be human readable for development and debugging and not intended for machine-readability since there are plenty of machine-readable export formats available (first and foremost OTLP). There is a [file exporter in the OTel collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/fileexporter) which uses the OTLP/JSON format but this does not guarantee stable field names either.
[...]

PR #2930, opened 2022-11-08
Declare OTLP/JSON Stable

tigrannajaryan:
See also https://github.com/open-telemetry/opentelemetry-proto/pull/436


carlosalberto (approved):

tigrannajaryan:
We need a lot of eyes on this before we merge this.

reyang (approved):

jack-berg (approved):

Aneurysm9 (approved):

jsuereth (approved):

bogdandrutu (commented):
Would like to clarify the behavior for invalid (all 0) trace/span IDs. In other words we have 3 cases for trace/logs/metrics exemplar:
* "traceId" key is not present
* "traceId" key is present and has an empty value
* "traceId" key is present and value is all 0s

Can we consider these 3 cases the same? Is there a difference between them, if yes, what is the semantic?

jmacd (approved):

jmacd:
I asked @bogdandrutu because I didn't understand immediately how https://github.com/open-telemetry/opentelemetry-specification/pull/2930#pullrequestreview-1177620347 was supposed to block stabilizing OTLP/JSON.

There are more cases than the three listed above. We expect that TraceIDs should be exactly 32 hex digits and SpanIDs should be exactly 16 hex digits. This follows from the conditions for a "valid" TraceID and SpanID here: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#spancontext

We already have language about rejecting invalid data, which makes me think no more specification is needed. We should reject IDs with all zeros, whether it is because they are empty or missing, because valid identifiers require at least one non-zero byte.

tigrannajaryan:
> We already have language about rejecting invalid data, which makes me think no more specification is needed. We should reject IDs with all zeros, whether it is because they are empty or missing, because valid identifiers require at least one non-zero byte.

This is true for Spans but not for LogRecords, where traceid/spanid are optional fields.

jmacd:
> This is true for Spans but not for LogRecords, where traceid/spanid are optional fields.

Bogdan's remark was clear about presence and absence, though https://github.com/open-telemetry/opentelemetry-specification/pull/2930#pullrequestreview-1177620347.

If the "traceId" is absent, then the field is absent, and if the "traceId" is present then the field is present. We only expect valid data when a field is present.

bogdandrutu:
> If the "traceId" is absent, then the field is absent, and if the "traceId" is present then the field is present. We only expect valid data when a field is present.

Does that mean absent is equivalent with present and invalid?

Aneurysm9:
> Does that mean absent is equivalent with present and invalid?

If the field is required, then I would say yes.  If the field is not require, then no.

codeboten (approved):

carlosalberto:
Hey @tigrannajaryan 

Adding a small note after (briefly) discussing this: the question regarding `trace_id` being missing/empty/all-zeroes is specially important as we are [deviating](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#json-protobuf-encoding) from the standard protbuf/JSON handling for `trace_id`, hence having this super clear would be important for components such as the Collector.

As discussed: we should at least mention some of the mentioned cases (missing/empty `trace_id`) is undefined behavior.

cc @bogdandrutu @jmacd

abitrolly:
> We need a lot of eyes on this before we merge this.

I can not read the spec. I need to read examples, but there are no examples as long as https://github.com/open-telemetry/opentelemetry-specification/issues/2798 is open.

abitrolly:
Another concern is that spec doesn't cover how to deal with incomplete and missing spans, which should occur quite often in case of errors and exceptions. Basically in cases where the proper trace would be most useful.

https://github.com/open-telemetry/opentelemetry-specification/issues/2773#issuecomment-1245686563

tigrannajaryan:
Depends on https://github.com/open-telemetry/opentelemetry-specification/issues/3040

tigrannajaryan:
> Would like to clarify the behavior for invalid (all 0) trace/span IDs. In other words we have 3 cases for trace/logs/metrics exemplar:
> 
> * "traceId" key is not present
> * "traceId" key is present and has an empty value
> * "traceId" key is present and value is all 0s
> 
> Can we consider these 3 cases the same? Is there a difference between them, if yes, what is the semantic?

Here is a PR to clarify this: https://github.com/open-telemetry/opentelemetry-proto/pull/442

tigrannajaryan:
Not stale. Blocked by https://github.com/open-telemetry/opentelemetry-proto/pull/442

jonatan-ivanov (approved):

bogdandrutu (approved):

djaglowski (approved):

jmacd (approved):

carlosalberto:
@tigrannajaryan the last blocking issue has been resolved - I think it's time to merge it (and release it as part of the February release).

abitrolly:
IWithout making end of span optional you will have to break the stable implementation, or people will resort to hacks for debugging spans in real-time.

carlosalberto:
@abitrolly Probably a comment for a different PR/issue? Please follow up otherwise with a new issue ;)

carlosalberto:
Merging as 1.18.0  was released.

jonatan-ivanov:
Is there any roadblock in front of marking the whole [OTLP spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md) stable (right now it is mixed but its components are stable)?

PR #2937, opened 2022-11-10
 add convention attribute k8s cluster version

### LINKED ISSUE #2913: add convention attribute for k8s cluster version

It would be nice to support cluster versions. Cluster upgrades can cause problems with software that communicates with this api. But I'am missing a convention entry for k8s cluster version equal to the cluster name (`conventions.AttributeK8SClusterName`).

An entry like `conventions.AttributeK8SClusterVersion = "k8s.cluster.version"`.

**Additional context.**

This is related to the [OpenShift metadata provider](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/15694) implementation in otelcol-contrib.

frzifus:
Signed-off-by: Benedikt Bongartz <bongartz@klimlive.de>

Fixes #2913


semantic_conventions/resource/k8s.yaml
```diff
@@ -10,6 +10,12 @@ groups:
         brief: >
           The name of the cluster.
         examples: ['opentelemetry-cluster']
+      - id: version
+        type: string
+        brief: >
+          The version of the cluster.
```
arminru, line 16:
```suggestion
          The version of the Kubernetes implementation of the cluster.
```
or
```suggestion
          The Kubernetes version of the of the cluster.
```
How about this to make clear this is about k8s and not the version of the installed app?

semantic_conventions/resource/k8s.yaml
```diff
@@ -10,6 +10,12 @@ groups:
...
+        examples: ['v1.21.11']
+
```
arminru, line 18: You could add a `note:` saying how this value can be retrieved and/or reference the Kubernetes docs for it.

semantic_conventions/resource/k8s.yaml
```diff
@@ -10,6 +10,12 @@ groups:
         brief: >
           The name of the cluster.
         examples: ['opentelemetry-cluster']
+      - id: version
```
arminru, line 13: @jsuereth @reyang @jmacd Do you have any concerns about adding this (potentially high-cardinality over time) attribute to the resources?

semantic_conventions/resource/k8s.yaml
joaopgrassi: If you add it as a `note:` like in here https://github.com/frzifus/opentelemetry-specification/blob/add_convention_attribute_k8s_cluster_version/semantic_conventions/trace/faas.yaml#LL13C11-L13C34 then it will show `[1]` as in here https://github.com/frzifus/opentelemetry-specification/blob/add_convention_attribute_k8s_cluster_version/specification/trace/semantic_conventions/faas.md#general-attributes

You will also need to re-generate the tool to update the markdown.

frzifus: let me check that :+1:

frzifus: mh.. i made some changes. But after calling `make` again it failed with the following:

```bash
docker run --rm -v /home/frzifus/git/go/observability/opentelemetry-specification/semantic_conventions:/source -v /home/frzifus/git/go/observability/opentelemetry-specification/specification:/spec \
	otel/semconvgen:0.14.0 -f /source markdown -md /spec --md-check
usage: gen-semconv [-h] [--debug]
                   [--only {span,resource,event,metric,units,scope}]
                   [--yaml-root folder] [--exclude EXCLUDE]
                   [YAML_FILE [YAML_FILE ...]] {code,markdown} ...
gen-semconv: error: No semantic convention model found!
make: *** [Makefile:83: table-check] Fehler 2
```

Any idea whats wrong?

frzifus: ok - seems its caused by `selinux`. https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label

semantic_conventions/resource/k8s.yaml
frzifus: mh.. At least from the scenarios ive in mind, it doesnt change to often. Depending on your retention period you may have 2-3 different versions in your storage.

frzifus: any update here?

MrAlias (approved):

specification/resource/semantic_conventions/k8s.md
```diff
@@ -25,6 +25,10 @@ Kubernetes object, but "name" is usually more user friendly so can be also set.
 | Attribute  | Type | Description  | Examples  | Requirement Level |
 |---|---|---|---|---|
 | `k8s.cluster.name` | string | The name of the cluster. | `opentelemetry-cluster` | Recommended |
+| `k8s.cluster.version` | string | The Kubernetes version of the of the cluster. [1] | `v1.21.11` | Recommended |
```
tigrannajaryan, line 28: Adding a recommended attribute to the Resource potentially is a breaking change for exporters that turn Resource attributes into metric dimensions.
@jsuereth WDYT?

frzifus:
can someone re-open this pr? seems ive no permission :x: 

cc @open-telemetry/specs-approvers

tigrannajaryan (changes requested):
I am temporarily blocking this to make sure we don't accidentally merge until my comment about this being a breaking change is addressed.

specification/resource/semantic_conventions/k8s.md
tigrannajaryan: @jsuereth when you have a moment it would be great to have your opinion on this.

PR #2940, opened 2022-11-11
Move event.domain from InstrumentationScope to LogRecord

### LINKED ISSUE #2939: Move event.domain from InstrumentScope to LogRecord
In the 11/9 Log SIG we discussed whether it was appropriate for `event.domain` to be defined as an attribute in InstrumentationScope, or is more appropriate an attribute on LogRecord.

We reasoned that LogRecord is more appropriate as follows:

- Because InstrumentationScope is a [build-time concept](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/glossary.md#instrumentation-scope), its not appropriate to InstrumentationScopes with the same name / version / schema, but different attributes.
- Therefore, it wouldn't be appropriate to have a single InstrumentationScope emit events in more than one `event.domain`.
- This means that emitting events in different domains requires you to emit those events in different InstrumentationScopes.
- We did not intend `event.domain` to impact decisions about how broad or narrow a particular InstrumentationScope is. Perhaps the InstrumentationScope represents an entire application. It should be perfectly valid to emit events in different domains.

Conclusion: Move `event.domain` to an attribute on LogRecord.

jack-berg:
Resolves #2939.

### DIFF
--- a/specification/logs/api.md
+++ b/specification/logs/api.md
@@ -92,7 +92,8 @@ the scope has a version (e.g. a library version). Example value: 1.0.0.
 - `schema_url` (optional): Specifies the Schema URL that should be recorded in
 the emitted telemetry.
 - `event_domain` (optional): Specifies the domain for the Events emitted, which
-should be added as `event.domain` attribute of the instrumentation scope.
+  MUST be added as an attribute with the key `event.domain`
+  to [emitted Events](#emit-event).
 - `include_trace_context` (optional): Specifies whether the Trace Context should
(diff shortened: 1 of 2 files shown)


tigrannajaryan (approved):

jsuereth (approved):

djaglowski (approved):

jack-berg:
@open-telemetry/specs-logs-approvers PTAL.

reyang (approved):

scheler (approved):

Issue #2936, opened 2022-11-10
Make sure it is very clear we are not building a Logging API


tigrannajaryan:
We are only building a logging "backend" which is supposed to be used via appenders by existing logging libraries. Additionally, we may provide helpers to build "events" that make easier to build LogRecords with specific semantic convention. Such helpers will also call the logging "backend" APIs.

We do not intend to have user-facing logging API in the spec.

tigrannajaryan:
@open-telemetry/specs-approvers FYI.

tigrannajaryan:
@open-telemetry/specs-logs-approvers FYI.

reyang:
+1

PR #2941, opened 2022-11-11
Split out Event API from Log API

### LINKED ISSUE #2917: Proposal for pure syntactic sugar event API
Proposal to change the event API to pure syntactic sugar over the log API.

Currently, the event API shows up in two places in the log API:
1. [LoggerProvider - Get A Logger](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/api.md#get-a-logger): `event_domain` is an optional field, required if you intend to use the resulting `Logger` to emit events.
2. [Logger - Emit Event](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/api.md#emit-logrecord). Emit a named event with the `event_domain` specified when initializing the `Logger`. 

Alternatively, the event API could just be a wrapper around `Logger`:
```
// Get a LoggerProvider
LoggerProvider loggerProvider = GlobalLoggerProvider.get();

// Create a EventLogger, which is configured with a Logger and a domain to be added as event.domain
// The EventLogger will delegate to the Logger when building / emitting events
EventLogger eventLogger = EventLogger.create(loggerProvider.get("logger-name"), "my-event-domain");

// Build / emit an event
eventLogger.eventBuilder("my-event-name")
    .setAttribute(AttributeKey.stringKey("k1"), "v1")
    .setAttribute(AttributeKey.stringKey("k2"), "v2")
    .emit();
// Emits event with attributes {"event.domain": "my-event-domain", "event.name": "my-event-name", "k1": "v1", "k2": "v2"}
```
Sample code [here](https://github.com/jack-berg/opentelemetry-java/blob/syntactic-sugar-event-api/api/logs/src/test/java/io/opentelemetry/api/events/EventLoggerTest.java#L10-L25). Sample source for [EventLogger](https://github.com/jack-berg/opentelemetry-java/blob/syntactic-sugar-event-api/api/logs/src/main/java/io/opentelemetry/api/events/EventLogger.java). 

In this design, `EventLogger` has a concrete implementation rather than being an interface. As a result, there's no ability to have alternative implementations. Can still provide alternative implementations for `Logger`, which `EventLogger` delegates to.

Advantages include:
- The log API is decoupled from the event API. The Event API depends on log API, not the other way around. This makes it somewhat easier to work towards [stabilizing logs independently of events](https://github.com/open-telemetry/opentelemetry-specification/issues/2911). (However, I do think its possible to have mixed stability with the logs API where [Emit LogRecord](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/api.md#emit-logrecord) is marked as stable while [Emit Event](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/api.md#emit-event) is still experimental. Its up to language sigs to handle this how they see fit.)
- The API better reflects the data model. Its clear that events are just a particular type of log. You can look the source for `EventLogger` and see plainly that it just delegates to `Logger` and sets some attributes. 

I'm not sure there's any existing places where OpenTelemetry provides syntactic sugar like this. The need for such a thing may be unique to the log signal, since logs are a really raw / low level signal upon which other things may be build (i.e. events, and potentially profiles). Its worth thinking about how additional pseudo-signals built on top of logs might manifest if this pattern is adopted and extended to other use cases.

jack-berg:
Resolves #2917.

Depends on #2940.


specification/logs/event-api.md
```diff
@@ -0,0 +1,73 @@
...
+
+</details>
+
+The Event API offers convenience methods
+for [emitting LogRecords](./api.md#emit-logrecord) that conform
+to the [semantic conventions for Events](./semantic_conventions/events.md).
+
+Unlike the Trace, Metric and Log APIs, the Event API does not support
+alternative implementations. Instead, the Event API is a concrete implementation
+which delegates to the [Log API](./api.md).
```
tigrannajaryan, line 27: Do we have to assert this in the spec? Does it matter from spec perpective?

tigrannajaryan (approved):
I am going to give it this a quick approval since I think it is important to do the splitting and then work separately to refine the Log part and the Event part.

Particularly for the Log part I want us to address https://github.com/open-telemetry/opentelemetry-specification/issues/2936 quickly since we are causing confusion with the way the spec is written now.

djaglowski (approved):

specification/logs/event-api.md
jack-berg: No not necessary. My thinking was that this would help language SIGs better understand how to implement. 

But I'll remove since its always easy to add more information if implementers report confusion.

specification/logs/event-api.md
```diff
@@ -0,0 +1,73 @@
...
+
+This function MAY be named `logEvent`.
+
+**Parameters:**
+
+* `event_name` - the Event name. This argument MUST be recorded as a `LogRecord`
+  attribute with the key `event.name`. Care MUST be taken by the implementation
+  to not override or delete this attribute while the Event is emitted to
+  preserve its identity.
+* `logRecord` - the [LogRecord](./api.md#logrecord) representing the Event.
```
scheler, line 58: The second parameter here could just be `attributes`. The `LogRecord` data model has fields that we prefer Events to not set, for eg., Body. This function should internally create a LogRecord object using the event_name, event_domain and attributes provided and emit that.

jack-berg: I agree with the general point you're making, but would prefer to address that in a separate issue. Currently this PR is essentially a refactor, with the structure of the API changing but without changes in available functionality. Probably best to keep it that way separately debate whether to limit the Event API fields to be a subset LogRecord.

scheler (approved):

reyang (approved):

tigrannajaryan:
@jack-berg please rebase.

Aneurysm9 (approved):

alanwest (approved):

specification/logs/event-api.md
```diff
@@ -0,0 +1,69 @@
...
+emitting `Events` as `LogRecords`.
+
+### EventLogger Operations
+
+The `EventLogger` MUST provide functions to:
+
+#### Create EventLogger
+
+New `EventLogger` instances are created though a constructor or factory method
+on `EventLogger`.
```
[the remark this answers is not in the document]
tigrannajaryan: We have enough approvals. I'll merge and this can be refined in a follow up PR.

PR #2943, opened 2022-11-11
Make http.status_code metric attribute an int

### LINKED ISSUE #2855: Semantic convention type mismatch for http.status_code

I am trying to understand if it is intentional that the type as described by the semantic conventions for `http.status_code` is different for Metrics and Traces. 

- Metrics defines it as a [string](https://github.com/open-telemetry/opentelemetry-specification/blame/ce32f948d2f6a78d95fd87f0d945513ee2a0d0aa/specification/metrics/semantic_conventions/http-metrics.md#L51) (although ambiguous because the example value `200` does not use quotes (`"`))
- Traces defines it as [int](https://github.com/open-telemetry/opentelemetry-specification/blame/d1c24937aad93b8d91c3b677976ccb8f432df5b0/specification/trace/semantic_conventions/http.md#L65)

What did you expect to see?

I expect these to be the same type for each of Logs/Metrics/Traces and that all SDKs follow the same convention. For example, in the dotnet SDK, Metrics use `string` and Traces use `int`.

**Additional context.**

- My assumption is that this could be an artifact of when Metrics only support attributes of type string.
- Metrics are handled differently by different SDKs (might be related to the ambiguity mentioned above)
    - Python uses an [int](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/32d7ff4082120c87b5d14dfd5e753010a78f1290/instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py#L332)
    - Dotnet uses a [string](https://github.com/open-telemetry/opentelemetry-dotnet/blob/8800b2ffdbdc013390e54a0994288111f3e95a54/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs#L84)
- I am currently working around this by using the [attributes proccessor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/attributesprocessor) with the `convert` function.

alanwest:
Fixes #2855

The [trace](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#common-attributes) semantic conventions have a nice `Type` column. The [metric](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md#attributes) conventions does not. It'd be nice if the structure of these tables matched. I'd be happy follow up with with this, but for now I wanted to make the smallest possible change to make sure we agree `http.status_code` should be an integer attribute.


jack-berg (approved):

trask (approved):

yurishkuro:
because metrics are sensitive to cardinality, I've seen instrumentations using strings like 4xx, 5xx for status code.

svrnm:
> because metrics are sensitive to cardinality, I've seen instrumentations using strings like 4xx, 5xx for status code.

I am wondering if there should be 2 attributes to cover that, i.e. `http.status_code` and maybe `http.status_code_group`, or is there a standardised concept for such a kind of metric aggregation?

reyang (approved):

cijothomas (approved):

svrnm:
> Interesting. I think it would be ideal if the `http.status_code` had the same semantics between traces and metrics. Maybe a separate attribute `http.status_code_class` could be proposed for capturing `2xx`, `4xx`, etc?

status_code_class is better, since it's the official name (see https://datatracker.ietf.org/doc/html/rfc9110#section-15), I am wondering if it should be "1xx", "2xx", etc or "Informational", "Successful", etc?

alanwest:
> status_code_class is better, since it's the official name (see https://datatracker.ietf.org/doc/html/rfc9110#section-15), I am wondering if it should be "1xx", "2xx", etc or "Informational", "Successful", etc?

Good question. Opened #2947 for follow up.

reyang:
> @jsuereth I saw this only today, and couldn't join the semconv wg meeting, but shouldn't this be seen as a breaking change?
> 
> * Existing queries using the string "200" for ex, or any string-based operation like `startsWith`, `contains` would break if this is changed to int. For ex, this [prometheus query](https://prometheus.io/docs/prometheus/latest/querying/examples/)

No, for Prometheus Exporter, the integer `200` label should be converted to `"200"`, which means from consumption side it'll be the same.

joaopgrassi:
>No, for Prometheus Exporter, the integer 200 label should be converted to "200", which means from consumption side it'll be the same.

Right, Prometheus was just one example, but did we (or should we) consider metric back-ends that don't support non-string attribute values? If users of such back-ends are already using instrumentation libraries that sends this as strings and rely on the attribute being a string for their dashboards/alerts, by moving forward with this, we are essentially breaking them or am I missing something?

reyang:
> @reyang right, Prometheus was just one example, but did we (or should we) consider metric back-ends that don't support non-string attribute values? If users of such back-ends are already using instrumentation libraries that sends this as strings and rely on the attribute being a string for their dashboards/alerts, by moving forward with this, we are essentially breaking them or am I missing something?

Changing type from string to int is a breaking change IMHO (@jsuereth FYI). We shouldn't encourage such change once the semantic convention is Stable. For this particular case, I would highly encourage/support aligning the types at Experimental stage.

reyang:
🤷‍♂️

image

PR #2946, opened 2022-11-14
Add RabbitMQ message headers to semantic conventions

### LINKED ISSUE #2895: Add RabbitMQ `message.header` attributes to semantic conventions

I want to add RabbitMQ message header attributes in consumer spans to the semantic conventions. The value of adding the message headers is to help users get more information from the traces which could allow for easier troubleshooting.
Since headers may contain sensitive data, we should require an explicit configuration of which headers are to be captured.

I propose the following naming for the attributes:

`messaging.rabbitmq.message.header.<key>`

samimusallam:
Fixes #2895

## Changes

- Add semantic conventions for RabbitMQ message headers.


carlosalberto:
@samimusallam Your PR is empty (only Changelog was updated).

samimusallam:
> @samimusallam Your PR is empty (only Changelog was updated).

fixed :) it seems one of the lint commands I ran removed my changes without noticing

specification/trace/semantic_conventions/messaging.md
```diff
@@ -263,11 +263,14 @@ Instead span kind should be set to either `CONSUMER` or `SERVER` according to th
 In RabbitMQ, the destination is defined by an *exchange* and a *routing key*.
 `messaging.destination` MUST be set to the name of the exchange. This will be an empty string if the default exchange is used.
 
-<!-- semconv messaging.rabbitmq -->
```
lmolkova, line 266: let's preserve attribute auto-generation - you can add headers as an additional section after the auto-generated table (maybe even as a continuation of it.

specification/trace/semantic_conventions/messaging.md
```diff
@@ -263,11 +263,14 @@ Instead span kind should be set to either `CONSUMER` or `SERVER` according to th
...
 | Attribute  | Type | Description  | Examples  | Requirement Level |
 |---|---|---|---|---|
 | `messaging.rabbitmq.routing_key` | string | RabbitMQ message routing key. | `myKey` | Conditionally Required: If not empty. |
-<!-- endsemconv -->
+| `messaging.rabbitmq.message.header.<key>` | string[] | RabbitMQ message headers. `<key>` being the normalized message header name (lowercase, with - characters replaced by _), the value being the header values. [1] [2] | `messaging.rabbitmq.message.header.my_header=["my_header_value"]` | Recommended |
```
lmolkova, line 269: can you please add a link to rabbitmq documentation here and specify what's considered to be a header. E.g. is content-type a header or only optional properties are headers?

specification/trace/semantic_conventions/messaging.md
```diff
@@ -263,11 +263,14 @@ Instead span kind should be set to either `CONSUMER` or `SERVER` according to th
...
+
+**[1]:** Instrumentations SHOULD require an explicit configuration of which message headers are to be captured. Including all headers can be a security risk - explicit configuration helps avoid leaking sensitive information.
+
+**[2]:** The attribute value MUST consist of either multiple header values as an array of strings or a single-item array containing a possibly comma-concatenated string, depending on the way the library provides access to message headers.
```
lmolkova, line 273: does rabbitmq support array headers?

lmolkova (changes requested):
@samimusallam can you please take a look at https://github.com/open-telemetry/opentelemetry-specification/pull/2763?

Messaging SIG is introducing `messaging.message` namespace.  Could you consider an option to put rabbitmq custom headers to `messaing.rabbitmq.message` namespace instead ?

specification/trace/semantic_conventions/messaging.md
samimusallam: It does in the NodeJS client, not sure about other languages but should be similar

specification/trace/semantic_conventions/messaging.md
samimusallam: Sounds good! I'll add the headers section as a continuation below the auto-generated table

samimusallam:
> @samimusallam can you please take a look at #2763?
> 
> Messaging SIG is introducing `messaging.message` namespace. Could you consider an option to put rabbitmq custom headers to `messaing.rabbitmq.message` namespace instead ?

I've added the headers to `messaging.rabbitmq.message.header.<key>` as agreed to [here](https://github.com/open-telemetry/opentelemetry-specification/issues/2895#issuecomment-1302384172) are you referring to using `messaging.rabbitmq.message.<key>` instead?

specification/trace/semantic_conventions/messaging.md
samimusallam: The `content-type` property as an example exists outside the headers field. There are a couple more properties that are defined outside the headers field which I think could also be relevant to be captured in spans.
As written in the [docs](https://www.rabbitmq.com/publishers.html#message-properties) the headers field contains an arbitrary map so it's not something specific.

semantic_conventions/trace/messaging.yaml
```diff
@@ -141,21 +141,6 @@ groups:
       Semantic convention for servers that consume messages received from messaging systems
       and always send back replies directed to the producers of these messages.
 
-  - id: messaging.rabbitmq
```
jsuereth, line 144: Why was this removed?

samimusallam: As mentioned in [this comment](https://github.com/open-telemetry/opentelemetry-specification/pull/2946#issuecomment-1315804525) I've had some issues with the table generator as I couldn't escape the `<` and `>` characters for the headers attribute. So I took the same approach that was done in the HTTP semantic conventions which is to manually write the table.

specification/trace/semantic_conventions/messaging.md
samimusallam: I couldn’t manually add the headers below the generated table, it doesn't expand to the same table.
 Does it make sense to create a new table below the generated table just for the attribute headers?

osherv (approved):

samimusallam:

Not stale, waiting on reviewers

carlosalberto:
Removed the stale label - it looks like a few people are still on holidays, but hopefully soon a final cycle of reviews comes.

samimusallam:
@open-telemetry/specs-approvers @open-telemetry/specs-trace-approvers Please take a look at this

specification/trace/semantic_conventions/messaging.md
```diff
@@ -371,6 +371,16 @@ In RabbitMQ, the destination is defined by an *exchange* and a *routing key*.
 | `messaging.rabbitmq.destination.routing_key` | string | RabbitMQ message routing key. | `myKey` | Conditionally Required: If not empty. |
 <!-- endsemconv -->
 
+RabbitMQ message headers:
+
+| Attribute                                 | Type     | Description                                                                                                                                                         | Examples                                                          | Requirement Level |
+|-------------------------------------------|----------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------|-------------------|
+| `messaging.rabbitmq.message.header.<key>` | string[] | RabbitMQ message headers. `<key>` being the normalized message header name (lowercase, with - characters replaced by _), the value being the header values. [1] [2] | `messaging.rabbitmq.message.header.my_header=["my_header_value"]` | Recommended       |
```
lmolkova, line 378: I believe this should be `optional`. Optional means that it's not populated by default, but users can opt-in. For example, instrumentation library may provide configuration for users to enable it. 

`Recommended` would mean that all instrumentation libraries SHOULD populate all headers by default. Since number and content of the headers is not known, they should not be populated by default.

samimusallam: Oh I miss understood `Recommended`. Will update it to `optional`.

samimusallam:
Can this PR be reopened? It went stale for too long

specification/trace/semantic_conventions/messaging.md
lmolkova: Thank you for sharing the docs! would you mind adding them to the md file too so instrumentation authors know precisely what they refer to?

samimusallam: Added :)

specification/trace/semantic_conventions/messaging.md
```diff
@@ -371,6 +371,18 @@ In RabbitMQ, the destination is defined by an *exchange* and a *routing key*.
...
+| `messaging.rabbitmq.message.header.<key>` | string[] | RabbitMQ message headers [1]. `<key>` being the normalized message header name (lowercase, with - characters replaced by _), the value being the header values. [2] [3] | `messaging.rabbitmq.message.header.my_header=["my_header_value"]` | Optional          |
+
+**[1]:** Refer to [the RabbitMQ documentation](https://www.rabbitmq.com/publishers.html#message-properties) for more details on message headers.
+
+**[2]:** Instrumentations SHOULD require an explicit configuration of which message headers are to be captured. Including all headers can be a security risk - explicit configuration helps avoid leaking sensitive information.
+
+**[3]:** The attribute value MUST consist of either multiple header values as an array of strings or a single-item array containing a possibly comma-concatenated string, depending on the way the library provides access to message headers.
```
lmolkova, line 384: The type mentioned above is `string[]` and this note suggests it can be either an array or a single string. 

Given that it's an optional attribute anyway, it may be reasonable to ask instrumentations to split comma-separated-string into array.

samimusallam: Please let me know if the updated note is clear

samimusallam:
Not stale, waiting for review @open-telemetry/technical-committee, @open-telemetry/specs-approvers, @open-telemetry/specs-trace-approvers and @open-telemetry/instr-wg

semantic_conventions/trace/messaging.yaml
Oberon00: Hi, you can't remove the entry here, this will lead to breaking changes for all languages that use the YAML files to generate code.

samimusallam: Hi, this is outdated. This is not removed anymore.

samimusallam: Yes exactly, the `<>` are not supported yet so I had to write it manually :)

specification/trace/semantic_conventions/messaging.md
Oberon00: Why split by comma? The comma rule is something that applies to HTTP, but why would it apply to messaging headers? Does any messaging system or RabbitMQ specifically support multiple values for the same header, or does an array maybe not make sense here? OTOH, are non-string values supported as headers?

specification/trace/semantic_conventions/messaging.md
```diff
@@ -371,6 +371,16 @@ In RabbitMQ, the destination is defined by an *exchange* and a *routing key*.
 | `messaging.rabbitmq.destination.routing_key` | string | RabbitMQ message routing key. | `myKey` | Conditionally Required: If not empty. |
 <!-- endsemconv -->
 
+RabbitMQ message headers:
+
+| Attribute                                 | Type     | Description                                                                                                                                                         | Examples                                                          | Requirement Level |
+|-------------------------------------------|----------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------|-------------------|
+| `messaging.rabbitmq.message.header.<key>` | string[] | RabbitMQ message headers [1]. `<key>` being the normalized message header name (lowercase, with - characters replaced by _), the value being the header values. [2] | `messaging.rabbitmq.message.header.my_header=["my_header_value"]` | Optional          |
```
Oberon00, line 378: I think message headers are supported in most messaging systems and the definition does also not contain anything RabbitMQ-specific. Should we make it a general `messaging.message.header.<key>` attribute instead?

CC @lmolkova @pyohannes

One thing I'm wondering about though is case sensitivity: Here, messaging systems might behave differently, probably case-insensitive systems should apply the transformation described here (IIUC, it is copied from HTTP), while others should report the key as-is.

PR #2950, opened 2022-11-15
Add connect to rpc.yaml

### LINKED ISSUE #2949: Add connect to trace rpc semantic conventions

Add [connect](http://connect.build) rpc framework to semantic conventions

**Additional context.**
https://github.com/open-telemetry/opentelemetry-specification/pull/2950

joshcarp:
Fixes #2949 

## Changes

Adds connect as a new rpc convention in trace/rpc.yaml


yurishkuro:
-1

imo `rpc.connect` is a super confusing naming. I suggest disambiguating it in some way, like `buf-connect`

joshcarp:
I have changed `connect` to `buf_connect`

semantic_conventions/trace/rpc.yaml
```diff
@@ -210,3 +213,47 @@ groups:
       - id: uncompressed_size
         type: int
         brief: "Uncompressed size of the message in bytes."
+
+  - id: rpc.buf_connect
+    prefix: rpc.buf_connect
```
jsuereth, line 218: Please make sure to add a section in the Markdown where this is documented.


Specifically: `<!-- semconv rpc.buf_connect -->` should show up somewhere in the Markdown.

semantic_conventions/trace/rpc.yaml
```diff
@@ -210,3 +213,47 @@ groups:
...
+    type: span
+    extends: rpc
+    brief: 'Tech-specific attributes for Connect.'
+    attributes:
+      - id: error_code
```
jsuereth, line 223: You may want to (in the soon-to-be-added Connect specific markdown) outline how the existing error_code / span status attributes should interact with connect.

joshcarp:
Hey, jumping back on this. 
Markdown is edited and I've changed it to the correct buf connect specifications.

semantic_conventions/trace/rpc.yaml
```diff
@@ -210,3 +213,48 @@ groups:
...
+            - id: unimplemented
+              value: unimplemented
+            - id: internal
+              value: internal
+            - id: unavailable
+              value: unavailable
+            - id: data_loss
+              value: data_loss
+            - id: unauthenticated
+              value: unauthenticated
```
jmacd, line 257: can you say why this section looks different than the gRPC section above which also includes a "brief" attribute? This also looks unnecessarily duplicated -- can we instead refer to the gRPC section above?

joshcarp: It is unnecessarily duplicated but the generator needs to have  both an key/id and a value. They're similar to grpc but different because all connect error codes are strings, not ints

joshcarp:
Hello, still want to get this PR merged. 
Don't seem to have permissions to reopen the PR. 

gRPC and connect are different protocols but share a lot of similarities; error codes are one of their slight differences 
- gRPC uses ints as error codes which have a corresponding error status
- connect uses strings as error codes and the error status/description is therefore just the error code.

There is already an implementation that will use this of this over at https://github.com/bufbuild/connect-opentelemetry-go too if it helps

PR #2951, opened 2022-11-15
Define MetricProducer as a third-party provider of metric data to MetricReaders

dashpole:
Supersedes #2838 and #2722.

Issue: https://github.com/open-telemetry/opentelemetry-specification/issues/1175

This is needed to support an OpenCensus metric bridge, which is [will be proposed separately](https://github.com/open-telemetry/opentelemetry-specification/pull/2732). It could also be used to support other bridges as well, such as a Prometheus bridge.

## Changes

Define the MetricProducer interface, with one method: Produce().  MetricProducers SHOULD have a single InstrumentationScope, and can be configured with a Resource.

Add a RegisterProducer(metricProducer) function to MetricReader.

MetricProducers SHOULD allow configuration of temporality, but this isn't enforced via the metric reader.  This is because the reader's temporality is configured as a function of OpenTelemetry instrument.

### Changes from the previous iteration

* **Integrate with the MetricReader instead of MeterProvider** (Feedback from https://github.com/open-telemetry/opentelemetry-specification/pull/2838#issuecomment-1259910518)
* This no longer attempts to prevent duplicate Instrumentation Scopes, or limits a bridge to a single scope (it is still recommended to use a single scope).
* It now recommends that bridges allow configuration of the output temporality.

### Things that are left up to implementations

* How the MetricReader handles multiple metric batches is NOT specified.  SDKs are free to pass each metric batch to the exporter individually, or it may attempt to merge batches if, for example, the resource was the same between the batches.  This also implies that any merge logic (e.g. handling duplicate instrumentation scopes) is also not specified, and is left up to the implementation.

@jsuereth @bogdandrutu @jmacd @reyang @pirgeo @MadVikingGod @MrAlias @jack-berg

### DIFF
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -757,7 +757,7 @@ measurements using the equivalent of the following naive algorithm:
 common configurable aspects of the OpenTelemetry Metrics SDK and
 determines the following capabilities:
 
-* Registering a [MetricBridge](#metricbridge)
+* Registering [MetricBridge](#metricbridge)(s)
 * Collecting metrics from the SDK and any registered
   [MetricBridges](#metricbridge) on demand.
 * Handling the [ForceFlush](#forceflush) and [Shutdown](#shutdown) signals from
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -831,7 +831,7 @@ functions.
 RegisterProducer causes the MetricReader to use the provided
 [MetricProducer](#metricproducer) as a source of aggregated metric data in
 subsequent invocations of Collect. RegisterProducer is expected to be called
-during initialization, but MAY be invoked later.  Multiple registrations
+during initialization, but MAY be invoked later. Multiple registrations
 of the same MetricProducer MAY result in duplicate metric data being collected.
 
 If the [MeterProvider](#meterprovider) is an instance of
--- a/specification/metrics/sdk.md
+++ b/specification/metrics/sdk.md
@@ -37,7 +37,7 @@ linkTitle: SDK
   * [Exemplar defaults](#exemplar-defaults)
 - [MetricReader](#metricreader)
   * [MetricReader operations](#metricreader-operations)
-    + [RegisterBridge(metricBridge)](#registerbridgemetricbridge)
+    + [RegisterProducer(metricProducer)](#registerproducermetricproducer)
     + [Collect](#collect)
     + [Shutdown](#shutdown-1)
   * [Periodic exporting MetricReader](#periodic-exporting-metricreader)
@@ -48,7 +48,7 @@ linkTitle: SDK
       - [ForceFlush()](#forceflush)
       - [Shutdown()](#shutdown)
   * [Pull Metric Exporter](#pull-metric-exporter)
-- [MetricBridge](#metricbridge)
+- [MetricProducer](#metricproducer)
   * [Interface Definition](#interface-definition-1)
     + [Produce() batch](#produce-batch)
 - [Defaults and configuration](#defaults-and-configuration)
@@ -757,9 +757,9 @@ measurements using the equivalent of the following naive algorithm:
 common configurable aspects of the OpenTelemetry Metrics SDK and
 determines the following capabilities:
 
-* Registering [MetricBridge](#metricbridge)(s)
+* Registering [MetricProducer](#metricproducer)(s)
 * Collecting metrics from the SDK and any registered
-  [MetricBridges](#metricbridge) on demand.
+  [MetricProducers](#metricproducer) on demand.
 * Handling the [ForceFlush](#forceflush) and [Shutdown](#shutdown) signals from
   the SDK.
 
@@ -786,7 +786,7 @@ Delta](supplementary-guidelines.md#asynchronous-example-delta-temporality) aggre
 temporality.
 
 The `MetricReader` is not required to ensure data points from a non-SDK
-[MetricBridge](#metricbridge) are output in the configured aggregation
+[MetricProducer](#metricproducer) are output in the configured aggregation
 temporality, as these data points are not collected using OpenTelemetry
 instruments.
 
@@ -824,23 +824,23 @@ functions.
 
 ### MetricReader operations
 
-#### RegisterBridge(metricBridge)
+#### RegisterProducer(metricProducer)
 
-RegisterBridge causes the MetricReader to use the provided
-[MetricBridge](#metricbridge) as a source of aggregated metric data in
-subsequent invocations of Collect. RegisterBridge is expected to be called
+RegisterProducer causes the MetricReader to use the provided
+[MetricProducer](#metricproducer) as a source of aggregated metric data in
+subsequent invocations of Collect. RegisterProducer is expected to be called
 during initialization, but MAY be invoked later.  Multiple registrations
-of the same metricBridge MAY result in duplicate metric data being collected.
+of the same MetricProducer MAY result in duplicate metric data being collected.
 
 If the [MeterProvider](#meterprovider) is an instance of
-[MetricBridge](#metricbridge), this MAY be used to register the
+[MetricProducer](#metricproducer), this MAY be used to register the
 MeterProvider, but MUST NOT allow multiple [MeterProviders](#meterprovider)
 to be registered with the same MetricReader.
 
 #### Collect
 
 Collects the metrics from the SDK and any registered
-[MetricBridges](#metricbridge). If there are
+[MetricProducers](#metricproducer). If there are
 [asynchronous SDK Instruments](./api.md#asynchronous-instrument-api) involved,
 their callback functions will be triggered.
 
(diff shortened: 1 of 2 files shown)


specification/metrics/sdk.md
```diff
@@ -813,11 +824,23 @@ functions.
 
 ### MetricReader operations
 
+#### Register(MetricBridge)
+
+Register causes the MetricReader to use the provided
+[MetricBridge](#metricbridge) for as a source of aggregated metric data in
+subsequent invokations of Collect.
+
+If the [MeterProvider](#meterprovider) is an instance of of
```
reyang, line 833:
```suggestion
If the [MeterProvider](#meterprovider) is an instance of
```

specification/metrics/sdk.md
```diff
@@ -1087,6 +1110,58 @@ modeled to interact with other components in the SDK:
...
+```
+
+### Interface Definition
+
+A `MetricBridge` MUST support the following functions:
+
+#### Produce() batch
+
+`Produce` provides metrics from the MetricBridge to the caller. `Produce`
+MUST return a batch of [Metric points](./data-model.md#metric-points).
```
reyang, line 1151: Is this batch expected to be the same format as `Export(batch)`?

specification/metrics/sdk.md
```diff
@@ -813,11 +824,23 @@ functions.
 
 ### MetricReader operations
 
+#### Register(MetricBridge)
```
reyang, line 827: Few cents:
1. use lowercase parameter, similar to `Export(batch)` - e.g. `Register(metricBridge)` or `Register(bridge)`
2. consider a more specific function name (`Register` sounds too generic to me).
3. maybe consider a different verb, `Register` seems to suggest that this function is idempotent (e.g. if one calls Register twice with the same input, it'll have the same effect as a single call).
4. clarify whether it is expected to be called at any time during the lifecycle of MetricReader, or only called during the initialization.

specification/metrics/sdk.md
```diff
@@ -1087,6 +1110,58 @@ modeled to interact with other components in the SDK:
...
   ```
 
+## MetricBridge
+
+
+`MetricBridge` defines the interface which bridges to third-party metric
+sources MUST implement so they can be plugged into an OpenTelemetry
+[MetricReader](#metricreader) as a source of aggregated metric data. The SDK's
+in-memory state MAY implement the `MetricBridge` interface for convenience.
```
reyang, line 1120: Would this lead us to `MetricProduer` name? 😃 (I personally would vote for `MetricProducer` because it has a clear semantic that the data flows in one direction, `bridge` sounds like a duplex thing to me although we already used the term in other places...)

reyang (approved):
Overall, I'm supportive - this is a critical part which makes it possible to converge OpenCensus.
I've left several non-blocking comments considering the added part is Experimental.

specification/metrics/sdk.md
```diff
@@ -753,7 +757,9 @@ measurements using the equivalent of the following naive algorithm:
 common configurable aspects of the OpenTelemetry Metrics SDK and
 determines the following capabilities:
 
-* Collecting metrics from the SDK on demand.
+* Registering a [MetricBridge](#metricbridge)
```
jack-berg, line 760:
```suggestion
* Registering [MetricBridge](#metricbridge)(s)
```

specification/metrics/sdk.md
```diff
@@ -1087,6 +1110,58 @@ modeled to interact with other components in the SDK:
...
+### Interface Definition
+
+A `MetricBridge` MUST support the following functions:
+
+#### Produce() batch
+
+`Produce` provides metrics from the MetricBridge to the caller. `Produce`
+MUST return a batch of [Metric points](./data-model.md#metric-points).
+`Produce` does not have any required parameters, however, [OpenTelemetry
+SDK](../overview.md#sdk) authors MAY choose to add parameters (e.g. timeout).
```
jack-berg, line 1153: Timeout being an optional parameter is suspect. Seems to leave the door open for future parameters but it would be inconvenient to add parameters later.

specification/metrics/sdk.md
```diff
@@ -1087,6 +1110,58 @@ modeled to interact with other components in the SDK:
...
+
+`Produce` SHOULD provide a way to let the caller know whether it succeeded,
+failed or timed out. When the `Produce` operation fails, the `MetricBridge`
+MAY return successfully collected results and a failed reasons list to the
+caller.
```
jack-berg, line 1158: Produce returns a batch of metric points. Is it expected to wrap that batch in some sort of promise / future? Would it be acceptable for `Produce` to indicate failure with an exception / error?

jack-berg (approved):
Non blocking comments!

specification/metrics/sdk.md
dashpole: This mirrors the language used for Collect from MetricReader: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#collect.  This is meant to be basically the same as Collect() but from an external source.

specification/metrics/sdk.md
dashpole: I don't think it necessarily has to be, but I would expect most languages to implement it that way.  I've tried to keep language consistent ("batch of Metric points") to imply that the two are at least very similar.

dashpole:
> I've left several non-blocking comments considering the added part is Experimental.

Question i'm not entire sure of: I've marked the MetricBridge as experimental, but the Register() function is part of the MetricReader, which is stable.  Does that make sense?  Or would both essentially need to be stable for any stable languages to implement this?

specification/metrics/sdk.md
dashpole: (leaving this comment for others to provide opinions)

specification/metrics/sdk.md
dashpole: (same comment as above about using identical language to collect).

specification/metrics/sdk.md
dashpole: Changed Register(MetricBridge) -> RegisterBridge(metricBridge).

Clarified behavior with multiple calls, and when it is expected to be called:
```
RegisterBridge is expected to be called during initialization, but MAY be invoked later.  Multiple registrations of the same metricBridge MAY result in duplicate metric data being collected.
```

I thought about it, but don't have any ideas that are better than Register.  Any suggestions?

jmacd: I'm happy with `RegisterBridge`.

specification/metrics/sdk.md
jmacd: +1

reyang:
> > I've left several non-blocking comments considering the added part is Experimental.
> 
> Question i'm not entire sure of: I've marked the MetricBridge as experimental, but the Register() function is part of the MetricReader, which is stable. Does that make sense? Or would both essentially need to be stable for any stable languages to implement this?

I feel it is clear to me. Maybe adding a note in the MetricReader.Register() section saying it is Experimental?

MadVikingGod (approved):

specification/metrics/sdk.md
```diff
@@ -1087,6 +1112,58 @@ modeled to interact with other components in the SDK:
...
+## MetricBridge
+
+**Status**: [Experimental](../document-status.md)
+
+`MetricBridge` defines the interface which bridges to third-party metric
+sources MUST implement so they can be plugged into an OpenTelemetry
+[MetricReader](#metricreader) as a source of aggregated metric data. The SDK's
+in-memory state MAY implement the `MetricBridge` interface for convenience.
+
+`MetricBridge` implementations SHOULD accept configuration for the
```
pirgeo, line 1124: Am I reading this right that it is possible to pass configuration, and the MetricBridge _might_ do conversions if it is able to (e.g. there is a utility lib), but there is no requirement to do any conversions? In the following case: 
* User requests deltas
* `MetricBridge` has cumulative values and is not equipped to convert them
* `MetricBridge` therefore just passes the unconverted data through

Would it make sense to specify a warning log at this point, or is this self-explanatory?
Also, would it fall to the user to include the necessary libraries, if they exist?

dashpole: The intent is that MetricBridges themselves SHOULD offer that configuration. If a bridge doesn't offer such configuration, then users won't be able to request it, and there wouldn't be anything to emit a warning log about.

SDKs can make it more likely that bridges will choose to offer that config by making it easier to implement.

specification/metrics/sdk.md
dashpole: It is now RegisterProducer, given the other change from MetricBridge -> MetricProducer.

dashpole:
> I feel it is clear to me. Maybe adding a note in the MetricReader.Register() section saying it is Experimental?

Done.

specification/metrics/sdk.md
pirgeo: Got it 👍

specification/metrics/sdk.md
```diff
@@ -813,11 +824,27 @@ functions.
...
 ### MetricReader operations
 
+#### RegisterProducer(metricProducer)
+
+**Status**: [Experimental](../document-status.md)
+
+RegisterProducer causes the MetricReader to use the provided
+[MetricProducer](#metricproducer) as a source of aggregated metric data in
+subsequent invocations of Collect. RegisterProducer is expected to be called
+during initialization, but MAY be invoked later.  Multiple registrations
```
reyang, line 834: nit:

```suggestion
during initialization, but MAY be invoked later. Multiple registrations
```

reyang (approved):
LGTM assuming the changelog will be updated.

dashpole:
> LGTM assuming the changelog will be updated.

Added a changelog

MrAlias (approved):

PR #2959, opened 2022-11-17
Limit configuration envvars: unset = empty.

Oberon00:
Fixes #(no issue)

## Changes

Hopefully editorial only:

The spec currently already says in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#parsing-empty-value:

> The SDK MUST interpret an empty value of an environment variable the same way as when the variable is unset.

But then in the limit specification it used to say

> Empty value is treated as infinity.

This would contradict the above spec, but the empty value is also the default, so the default is infinity, and if it's not set the default is used, so in the end you could interpret this with unset = empty.

However, if you did not see the spec about empty=unset above, there was one interpretation possible where setting the envvar to an empty value would overwrite another value explicitly set via a different configuration source (there is no different source specified at the moment though). This interpretation is now no longer possible.


Oberon00:
I'm not sure if I should add a changelog for this, WDYT?

specification/sdk-environment-variables.md
```diff
@@ -130,7 +130,7 @@ See the SDK [Span Limits](trace/sdk.md#span-limits) section for the definition o
 
 | Name                                   | Description                                    | Default | Notes |
 | -------------------------------------- | ---------------------------------------------- | ------- | ----- |
-| OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT | Maximum allowed attribute value size           |         | Empty value is treated as infinity. Non-integer and negative values are invalid. |
+| OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT | Maximum allowed attribute value size           | no limit |       |
```
jack-berg, line 133: The `Notes` column can be deleted now

jack-berg (approved):

specification/sdk-environment-variables.md
```diff
@@ -119,7 +119,7 @@ See the SDK [Attribute Limits](common/README.md#attribute-limits) section for th
 
 | Name                              | Description                          | Default | Notes |
 | --------------------------------- | ------------------------------------ | ------- | ----- |
-| OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT | Maximum allowed attribute value size |         | Empty value is treated as infinity. Non-integer and negative values are invalid. |
+| OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT | Maximum allowed attribute value size | no limit|       |
```
reyang, line 122: Would there be a concern if we remove the "negative values are invalid"? E.g. with the proposed change, one could interpret it as "if the environment variable is set to `-1`, the limit will be `-1` (which means nothing will be taken)".

specification/sdk-environment-variables.md
Oberon00: I thought the tables should be kept consistent, but I'm personally fine with removing the note column wherever possible.

specification/sdk-environment-variables.md
Oberon00: This is an important but separate point. I created PR #2963 to fix that (and issue #2962)

arminru (approved):
LGTM as long as the dropped note will be covered by #2963.

Issue #2923, opened 2022-11-07
How to create an "up" metric


tomas-mota:
Sorry if this is out of place here, I have tried to ask in the slack but did not get any satisfiable answer.

**What are you trying to achieve?**
I want to create an up metric for my services.
When using bare prometheus, due to the pull mechanism, it is easy to know that if the scrape failed, the service is down. With otel, since you set a metric-expiration (which is nice to have for other metrics), the service might no longer be sending the up metric but it is still reported as being there with a value of 1, so we won't know that the service is down until it expires.

Is there some way to achieve this that I'm overlooking? Something I considered as a workaround is having an uptime metric, or current time metric, that we would expect to keep increasing if the service is up.

tomas-mota:
Forgot to mention, this is using the [prometheus exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/prometheusexporter)

dashpole:
The prometheus up metric is basically a healthcheck-as-a-metric.  You could achieve almost the same behavior by having something external to the application periodically health-check the application, and report a metric based on the result.  It won't be tied to your other metrics in the same way the prometheus up metric is, meaning this metric being in the healthy state doesn't imply your other metrics have been successfully collected in the way it does for prometheus.  But it does give you an external health signal.  The [httpcheck](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/httpcheckreceiver) receiver in the collector might be able to fulfil that role.

Alternatively, you could introduce a metric in the application whose value is the current time, and alert when the value of the metric is older than a threshold.

jsuereth:
If you're using a prometheus exporter + pull-based system, then an `up` metric should be created BY prometheus and makes a lot of sense.

`up` metrics aren't really the same in a push-based world.  Typically, you'd have a heartbeat metric, rather than an `up`, and you need to interact with them subtly different.   `up` means the endpoint was a live and CPU responded and such.  A heartbeat, you need to expect it every N seconds, and delays/shifts can signal problems, but with a lot of noise.

Generally, I think the notion of "health" is important, but I wouldn't use the exact same solution for pull-based health + push-based health.

I know @jmacd has done a lot of thinking here.  Josh, let us know if you think this idea has legs or if we should go a different direction.

tomas-mota:
[...]

Issue #2288, opened 2022-01-25
Support independent, per-Reader Metric Views configuration 


jmacd:
**What are you trying to achieve?**

Having implemented a mostly-complete Views implementation in https://github.com/open-telemetry/opentelemetry-go/pull/2525, I have the following feature request. 

Make it possible to configure Views on a per-Reader basis. 

Why? It seems like a shame to specify a powerful way to configure what gets exported from a metrics SDK and not let each Reader configure independent Views.  Without a change of the specification, the user who wants to configure independent Views will be required to make a "Multi-MeterProvider" that are configured with one Reader/Views pair each. I view this as a poor outcome because the SDK will lose any opportunity to share work among the two readers, particularly for synchronous instruments.

I make this request after implementing the draft PR linked above, where it looks like only a slight change. The way that PR is structured, there's almost no difference between supporting independent views because an "intermediate" aggregation is computed for synchronous instruments. No matter how many Views are attached to the same synchronous instrument, only distinct aggregations need to be computed on the fly. It means that if you configure 1 Counter instrument and 2 different Views over different attribute sets (with different output names, of course), the runtime would track 1 intermediate aggregation (having Delta temporality) and then output into each View on collection. The SDK just knows that it has a number of destinations per intermediate aggregation; it doesn't matter how many readers there are, at least the way I've structured it there.

**Example**

The current specification has an example like this:

```
# metrics from X and Y (reported as Foo and Bar) will be exported
meter_provider
    .add_view("X")
    .add_view("Foo", instrument_name="Y")
    .add_view(
        "Bar",
        instrument_name="Y",
        aggregation=HistogramAggregation(buckets=[5.0, 10.0, 25.0, 50.0, 100.0]))
    .add_metric_reader(PeriodicExportingMetricReader(ConsoleExporter()))
```

With the proposed change, we could write examples like this:

```
# Instrument Foo is renamed differently for the two exporters 
meter_provider
    .add_metric_reader(PullExportingMetricReader(PrometheusExporter())
        .add_view("FooRenamedTheOtherWay", instrument_name="Foo")),
    .add_metric_reader(PeriodicExportingMetricReader(OTLPExporter())
        .add_view("FooRenamedOneWay", instrument_name="Foo"))
```

We could configure ExponentialHistograms for OTLP while configuring ExplicitBucketHistograms for Prometheus, for example, or limit the set of attributes differently for these two exporters.

jsuereth:
[...]

Issue #1655, opened 2021-04-27
Determine how to report dropped metrics


jsuereth:
In #1618, the question of who to handle "dropped" metrics within OTLP was asked.   In that PR, a naive algorithm for Delta => Cumulative sum conversion was proposed that was intended to be lightweight (in memory consumption) and does NOT allow for out-of-order (or delayed) delta sum points.

This raises the question of how you would report "dropped" metrics within OTLP for this scenario.   Do we want to provide some known (semantic) convention?  Or should this be a first class attribute (like dropped_attribute_count on spans)?

For now, tagging this as DataModel + Semantic convention until we've had a chance to discuss the best way to handle this.

jsuereth:
# Context (in note form)

- Tracing [allows dropping data when over some limit](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#span-limits).  This is reflected beside every location that carries an attribute can report [dropped attributes count](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto#L174).
- Metrics API/SDK (currently) does not specify any mechanism to drop/limit the amount of data collected and reported.  Instead it focuses on "aggregation" of data.
  - For [Histogram](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#histogram), the data model [requires reporting the total number of measurements in the aggregation in a field called "count"](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L532).
  - For Sums (specifically synchronous instrument sums), there is no way to report the number of measurements during a reporting interval that lead to a value.
    - For synchronous Sum, it's possible no values were reported during an interval
    - For asynchronous sum it's possible an "exception" occurred during the fetch of the value.
  - Gauges are (currently) async only, meaning there should only be one Gauge per-reporting interval.  It is possible there is an "exception" during processing of a gauge.
  - The View / Processor API is still being fleshed out.  While a user will be able to drop metrics via this library, it's unclear what will be visible and how to proceed at this moment.
- From a "Collector" Perspective, the following a sources of dropped metric data:
  - The [Memory Limiter](https://github.com/open-telemetry/opentelemetry-collector/tree/main/processor/memorylimiter) processor can be used to drop incoming metrics when the collector itself has exceeded memory threshold limits
[...]

PR #2960, opened 2022-11-18
Specify MeterProvider configurable cardinality limits

### LINKED ISSUE #1891: Metrics SDK: define standard circuit breakers for uncollected, infinite cardinality recordings

jmacd:
Fixes #1891.

**EDIT: Updated to specify limit at the MeterProvider level, not the Reader level. Updated to use a hard limit configurable per view.**

## Changes

Adds MeterProvider optional support for a maximum cardinality limit.

The recommended default is 2000, based on this comment by https://github.com/open-telemetry/opentelemetry-specification/issues/1891#issuecomment-967672167 @jack-berg. 

~The Prometheus-WG SIG discussed this on Nov 9, 2022 and reached this recommended solution to the problem outlined in #1891. The consequence of exceeding these limits is in line with the current Prometheus server behavior, which drops targets that misbehave. The discussed was summarized here: https://github.com/open-telemetry/opentelemetry-specification/issues/1891#issuecomment-1309089908~


specification/metrics/sdk.md
```diff
@@ -877,6 +879,54 @@ from `MetricReader` and start a background task which calls the inherited
...
 
+### MetricReader configurable limits
+
+**Status**: [Experimental](../document-status.md)
+
+MetricReader SHOULD support being configured with limits to be applied
+to metrics produced from a single Instrumentation Library Scope.  This
+mechanism supports protecting metrics pipelines from excessive data
+production, which is usually caused by a problem known as "explosition
+of cardinality", in which the number of timeseries produced by the SDK
```
reyang, line 890: Considering folks who might not have clear understanding, we've mentioned "cardinality explosion" in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/supplementary-guidelines.md#synchronous-example-cumulative-aggregation-temporality. Maybe we can have a dedicated section in the supplementary guideline doc and link from here, WDYT? (I can send a follow-up PR if folks like the idea).

CHANGELOG.md
```diff
@@ -13,6 +13,11 @@ release.
 
 ### Metrics
 
+- Recommended limits applied to MetricReaders by default to protect
```
reyang, line 16: Do we want it on the reader level or the provider level?
Imagine a provider serving multiple readers, the final memory usage will be determined by the "most demanding reader".

MrAlias: Agreed. It does not seem appropriate to define this at the reader level. The trace span limits are defined at the `TracerProvider` level, the view which is used to address these limits is defined at the `MeterProvider` level, and if an environment variable is added later it will have to deal with the the many-to-one complexity.

specification/metrics/sdk.md
```diff
@@ -877,6 +879,54 @@ from `MetricReader` and start a background task which calls the inherited
...
+provided the application stops using excessive cardinality.  However,
+for Readers that have Synchronous instruments configured with
+Cumulative aggregation temporality, once a limit is reached the
+library cannot recover and the SDK SHOULD disable it for the remaining
+lifetime of the SDK.
+
+The configurable limits are:
+
+1. The number of timeseries per instrument, with recommended default value 2000.
+2. The number of timeseries per instrumentation library scope, with recommended default value 2000.
```
jack-berg, line 923: What's the use case for limiting the timeseries for a particular scope? I feel like most of the time users will opt to limit the number of timeseries per instrument because its easier to reason about. It will be easier to implement instrument limits than scope limits since scope limits will require that each instrument not only be aware of its own cardinality but also the other instruments in the scope.

CHANGELOG.md
jack-berg: > the view which is used to address these limits is defined at the MeterProvider level

Reader level view configuration has been requested in #2288. 

In my opinion the SDK doc implies an implementation where metric storage is separately managed for each reader, and therefore it feels natural to make the configuration of the metric storage a reader level concern. Configuring limits at the provider level might be easier to understand for casual users, but I suspect implementations may end up passing that shared configuration to each metric reader's storage similarly to how it would look if each metric storage was configurable. 

Something to consider: If we make it configurable at the provider level will we inevitably get requests to make it configurable at the reader level? If so it probably makes sense to get ahead of it.

MrAlias: Is configuration of limits per-reader the 80% of use-cases or 20%? Are we solving for the edge-case to the determent of the general-case here?

I imagine running multiple reader setups is the uncommon case and the 80% of uses will request to be able to apply their limit configuration the same place they apply their universal views and at the same level they apply their trace limits.

I think adding per-reader limits makes sense, but similar to views, I don't think that is where implementations should start.

reyang: > This setting is not necessarily about placing a limit total memory usage, it can also be used for protecting the downstream pipeline.

I feel MetricReader configuration would not be able to prevent cardinality explosion on the downstream pipeline - the pipeline is going to receive metric streams from indefinite number of clients, although each individual configuration on the SDK level can certainly help to prevent the application from running out-of-memory.

specification/metrics/sdk.md
reyang: I guess similar to tiered throttling - to prevent a certain scope going wild and eating up all the quota?

CHANGELOG.md
jmacd: Thanks. Like #2288, I think it's fine to begin with a provider-level configuration. I will update this PR.

specification/metrics/sdk.md
jmacd: I was thinking in terms of mental effort for the person configuring these limits. If the goal is to limit the cost, then it's easier to reason about per-library limits than per-instrument limits, since otherwise the cost depends on the number of instruments (which could change). I would argue against provider-level limits because the failure mode for a provider-level limit-exceeded is not very good and there's not much the user can do.

specification/metrics/sdk.md
jmacd: Sounds good. I added a link from this passage to the section in supplementary guidelines where the term appear for now.

specification/metrics/sdk.md
```diff
@@ -586,6 +593,57 @@ Attributes which belong to Metrics are exempt from the
...
+Note that limits expressed here refer to exported data for a single
+collection cycle.  The number of timeseries referred to in each of
+these limits equals the number of distinct data points produced over
+one interval.
+
+Under some circumstances, depending on the configured aggregation
+temporality, the MeterProvider may be able to recover after these
+limits are reached.  In particular, for Synchronous instruments
+configured with Delta aggregation temporality, the MeterProvider
+SHOULD remove timeseries from memory when they become stale, making it
```
pirgeo, line 629: Do we have a definition for stale somewhere?

specification/metrics/sdk.md
```diff
@@ -586,6 +593,57 @@ Attributes which belong to Metrics are exempt from the
...
+temporality, once a limit is reached the MeterProvider SHOULD disable
+it for the remaining lifetime of the MeterProvider.
+
+The configurable limits are:
+
+1. The number of timeseries per instrument, with recommended default value 2000.
+2. The number of timeseries per instrumentation library scope, with recommended default value 2000.
+
+For both of these limits, the limit MAY be unset to indicate an
+unlimited configuration and the zero value is considered invalid.
```
pirgeo, line 643: Would it make sense to treat 0 (or everything less than or equal to 0) as unset as well?

specification/metrics/sdk.md
```diff
@@ -586,6 +593,57 @@ Attributes which belong to Metrics are exempt from the
...
+ta
+production, which is usually caused by a problem known as ["explosition
+of cardinality"](supplementary-guidelines.md#synchronous-example-cumulative-aggregation-temporality),
+in which the number of timeseries produced by application code exceeds a
+reasonable number.  These limits help protect the overall system from
+individual sources of excessive metrics instrumentation.
+
+When the limits are reached by an the individual instrumentation
+library scope, the entire batch of metrics for the scope MUST be
+dropped and an error reported to the user indicating:
```
MrAlias, line 612: I interpret this as saying if limits are reached than no data will be exported for that instrumentation scope that collection cycle. Is this the intention?

If so, it seems like we could fail "more gracefully" by dropping the data that exceeded the limit and exporting the previously recorded data (while still emitting an error).

specification/metrics/sdk.md
jack-berg: > Iit's easier to reason about per-library limits than per-instrument limits

We're in opinion territory, but I think per-instrument limits are easier to reason about. 

In any case, we need to weigh the benefit of scope level limits against the cost of implementation. I'd prefer to not have to coordinate with a lock across instruments.

> otherwise the cost depends on the number of instruments (which could change).

Maybe in some cases but most of the time the number of instruments will be constant. If you're dynamically creating instruments I imagine most of the time the variable which is changing is better suited as an attribute.

specification/metrics/sdk.md
jack-berg: > If so, it seems like we could fail "more gracefully" by dropping the data that exceeded the limit and exporting the previously recorded data (while still emitting an error).

In some cases that would be useful, but in other cases that would be harmful as the continued flow of some data might reduce suspicion that something is wrong. If I don't have reason to suspect something's wrong, I don't check the logs and don't see the error.

reyang: One approach I've seen in Microsoft - have a "catch-all" stream, for example, if we have the following streams:
* x="a", y="a", sum=3
* x="a", y="b", sum=4
* x="b", y="a", sum=5
* x="b", y="b", sum=6
* x="c", y="a", sum=10
* x="c", y="b", sum=10
* x="c", y="c", sum=10

We can cap it at 5 and report:
* x="a", y="a", sum=3
* x="a", y="b", sum=4
* x="b", y="a", sum=5
* x="b", y="b", sum=6
* x="DIM_CAP", y="DIM_CAP", sum=30

reyang: I feel "MUST" is too strong, maybe start with "SHOULD" so later we can adjust the behavior (e.g. making it configurable).

jack-berg: That's a nice trick!

specification/metrics/sdk.md
reyang: A stepping back question - why do we want to have unlimited configuration?

specification/metrics/sdk.md
```diff
@@ -586,6 +593,57 @@ Attributes which belong to Metrics are exempt from the
...
 
+## Cardinality limits
+
+**Status**: [Experimental](../document-status.md)
+
+MeterProviders SHOULD support being configured with limits to be applied
+to metrics produced from a single Instrumentation Library Scope.  This
+mechanism supports protecting metrics pipelines from excessive da
+ta
+production, which is usually caused by a problem known as ["explosition
```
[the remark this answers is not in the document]
reyang: I think explosion is the intended word.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/supplementary-guidelines.md#synchronous-example-cumulative-aggregation-temporality

_Imagine if we have a long running service and we collect metrics with 7 attributes and each attribute can have 30 different values. We might eventually end up having to remember the complete set of all 21,870,000,000 permutations! This cardinality explosion is a well-known challenge in the metrics space._

specification/metrics/sdk.md
pirgeo: Good point. I guess if you don't know what you can expect, but you regularly exceed the 2000 TS limit? You could set it to 0 and run it for a while to see if it stabilized somewhere? Not advocating for it, I think having sensible defaults is good.

specification/metrics/sdk.md
jmacd: @reyang 

> One approach I've seen in Microsoft - have a "catch-all" stream, for example, if we have the following streams

A similar (better, IMO) approach would use unbiased weighted sampling to select 5 of the 7 streams (in your example), meaning to adjust the sum-or-rate-value of sampled timeseries. This way, averaged over time, the expected value of every timeseries will be correct. This is not the time or place to try and formalize this notion, but [it really can be done (e.g., here's a library that does it)](https://github.com/lightstep/varopt) and this is likely when I would configure unlimited cardinality for the MeterProvider and MetricReader. In my exporter, I would apply the downsampling from over-limit cardinality to stay inside of a configured limit.

reyang: @jmacd thanks for sharing! One downside I can see is the lack of high-fidelity total sum (e.g. with the "DIM_CAP" example above, we have high confidence that the total sum is `3+4+5+6+30`).

It seems based on discussions here we should give some flexibility instead of forcing the data to be dropped?

specification/metrics/sdk.md
reyang: > You could set it to 0 and run it for a while to see if it stabilized somewhere?

If the goal is to get some rough idea about the cardinality, maybe setting something really large (e.g. 10 million)?

jmacd: Agreed. I'm happy to remove the "unlimited" configuration as an option and recommend very large limits.

specification/metrics/sdk.md
dpk83: Dropping the data silently will negatively impact service developers. This leads to the drop of data result in firing incidents and the engineers have no clue what's going wrong and spend hours looking into wrong direction before they can finally realize it's drop of data. There needs to be a way to actively signal the consumers of the SDK to know when the data is bring dropped due to hitting cardinality limit.

There definitely needs to be some hook or some sort of mechanism to provide this indication reliably to the customers.

dpk83: We have already run into similar problems where if the data is not dropped the memory grows unbounded and if it's dropped silently then the metrics are wrong indicating problems with the service while the actual issue is in the telemetry.

jmacd: @dpk83 What would you suggest the SDK can do when the limit is reached for an instrument? It will produce an error and go through standard error handlers, I think.

For my vendor, I am encouraging the use of "lowmemory" temporality which allows delta temporality instruments to be dropped when they are not in use, however as @pirgeo points out, we haven't really defined what "stale" means.

specification/metrics/sdk.md
jmacd: I would like us to have a definition somewhere that helps explain how delta temporality allows SDKs to drop memory, but I realize that this PR probably isn't the place to do it. I believe we should add something in the data model to say when a point can be omitted by an exporter, then by extension if the SDK internal state allows removing memory that would produce a point that can be omitted by an exporter, it is allowed to do so.

specification/metrics/sdk.md
jmacd: This is done, thanks.

specification/metrics/sdk.md
```diff
@@ -586,6 +593,57 @@ Attributes which belong to Metrics are exempt from the
...
+of cardinality"](supplementary-guidelines.md#synchronous-example-cumulative-aggregation-temporality),
+in which the number of timeseries produced by application code exceeds a
+reasonable number.  These limits help protect the overall system from
```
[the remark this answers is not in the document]
jmacd: I now think it would be best to avoid the new term, and simply say what we mean. I've written:

```
MeterProviders SHOULD support being configured with limits to be
applied to metrics produced from a single metric instrument.  This
mechanism supports protecting metrics pipelines from excessive data
production in cases when the number of timeseries produced by
application code grows large, which can happen due to several factors.

Whether because of single attributes having many distinct values or
because of combinatorial expansion among many attributes, these limits
help protect the overall system from individual sources of excessive
metrics instrumentation.
```

jack-berg (approved):

specification/metrics/sdk.md
dpk83: > @dpk83 What would you suggest the SDK can do when the limit is reached for an instrument? It will produce an error and go through standard error handlers, I think.

SDK can emit a different metric when events are dropped/throttled. Developers can setup alerts on the metrics to monitor whenever any metrics are throttled/dropped, so it's not lost silently. 
Another mechanism is to give the developers a hook to get a callback when metrics are dropped. Developers can then themselves emit another metrics/log to monitor these cases.

jmacd: @dpk83 I support this notion, but I do not think that concern should block this PR.  If a user would prefer the risk of running out of memory to dropping metrics, they just need to configure a limit that is so large they would OOM before metrics are dropped.

I think we as a community are interested in someone picking up this project, to prototype and propose the specification for how to count dropped metrics/spans/logs. We have done some preparatory work in this regard, with the introduction of `PartialSuccess` fields.

jmacd: @dpk83 See https://github.com/open-telemetry/opentelemetry-specification/issues/1655

dpk83: @jmacd  I don't intend to block this PR, as the topic is related so I wanted to make sure that this is considered and there is an item to address the problem to provide the right experience.

> If a user would prefer the risk of running out of memory to dropping metrics, they just need to configure a limit that is so large they would OOM before metrics are dropped.

Regarding the above, the main issue here is the fact that a lot of times users end up adding a high cardinality dimension without either being aware of the fact that the dimension is high cardinality or being aware of the impact of this on memory. In this case the users don't really want the memory to grow unbounded but when such a thing happens they need to be able to catch/spot such problems so that they can take appropriate action and fix it. 

With the silent dropping of the metrics when they go beyond the cardinality we are just going to make the life of these users difficult. Providing a mechanism to easily catch such issues and get an indication quickly is very valuable. Think of this just like a throttling case for any scenarios, if the events are throttled the effect of throttling will show up in a different form to the user and without any throttling metrics user will end up spending a lot of time investigating all parts which could result in the issue.

jmacd:
Probably worth discussing in the maintainers call next week, to see if all existing SDKs are comfortable adding this limit.

specification/metrics/sdk.md
```diff
@@ -590,6 +598,48 @@ Attributes which belong to Metrics are exempt from the
...
+Under some circumstances, depending on the configured aggregation
+temporality, the MeterProvider may be able to recover after these
+limits are reached.
+
+### Specific cardinality limits
+
+There is only one configurable limit: the number of timeseries per
+instrument, with recommended default value 2000.
+
+SDKs SHOULD NOT support unlimited cardinality.  There is no unset
```
MrAlias, line 639: Is there a reason to not make this a requirement? When would it be valid for an implementation to support unlimited cardinality?

MrAlias, line 639: If an SDK doesn't support cardnality limits, it will support unlimited cardinality. Given cardnality limits are recommendations, this statement currently only states the same thing as "MeterProviders SHOULD support being configured with limits". I think the intent is the following though:

```suggestion
SDKs SHOULD NOT support the configuration of unlimited limits.  There is no unset
```

specification/metrics/sdk.md
```diff
@@ -590,6 +598,48 @@ Attributes which belong to Metrics are exempt from the
...
+  offending instrument to limit cardinality or otherwise correct the problem.
+
+Note that limit expressed here refers to exported data for a single
+instrument and collection interval.  The number of timeseries referred
+to here corresponds with the number of distinct data points produced
+by the instrument over one collection interval.
+
+Under some circumstances, depending on the configured aggregation
+temporality, the MeterProvider may be able to recover after these
+limits are reached.
```
MrAlias, line 632: This seems to contradict the fact that "metrics for the instrument MUST be dropped and an error reported to the user". Should the drop behavior be recommended and the exception called out explicitly that it shouldn't be dropped when the MeterProvider can recover?

specification/metrics/sdk.md
```diff
@@ -590,6 +598,48 @@ Attributes which belong to Metrics are exempt from the
...
+production in cases when the number of timeseries produced by
+application code grows large, which can happen due to several factors.
+
+Whether because of single attributes having many distinct values or
+because of combinatorial expansion among many attributes, these limits
+help protect the overall system from individual sources of excessive
+metrics instrumentation.
+
+When the limit is reached by an the individual metric instrument, the
+entire batch of metrics for the instrument MUST be dropped and an
```
MrAlias, line 617: I thought the resolution of [this comment](https://github.com/open-telemetry/opentelemetry-specification/pull/2960#discussion_r1035295900) meant there would be a more nuanced truncation strategy here. Like all additional timeseries being flattened into a single one?

Did I miss something?

specification/metrics/sdk.md
jmacd: I marked this conversation resolved because I do not believe the benefit of some kind of graceful degradation when the cardinality limit is reached is worth the amount of code and complexity that will go into it.

I also think we have different definitions of "silent". The user would see an error log message when the limit is reached. Besides, and orthogonal to this issue, [users should be counting dropped metric data points](https://github.com/open-telemetry/opentelemetry-specification/issues/1655). I think our time and energy will be better spent (1) discussing how to count dropped metric data points, spans and logs generally, (2) discussing how to make SDK logs more actionable/monitorable so that users do not experience any kind of silent telemetry outage.

(Besides, if we are prioritizing how to focus our attention on the metrics specification, I believe we have agreed that https://github.com/open-telemetry/opentelemetry-specification/issues/2923 is more important.)

specification/metrics/sdk.md
jmacd: Thanks. I followed up in https://github.com/open-telemetry/opentelemetry-specification/pull/2960#discussion_r1063803281. I don't think the effort and complexity of a more nuanced strategy is worthwhile.

specification/metrics/sdk.md
jmacd: I was referring to a potential implementation detail, but apparently the detail is not clear after my edits. If an SDK is using delta temporality and uses certain implementation techniques (which I'd like to keep out of scope), then the cardinality limit can be evaluated on a per-export basis; the loss of metrics is temporary in this case, whereas with cumulative temporality it seems fruitless to try to recover.

specification/metrics/sdk.md
jmacd: An existing stable SDK that supports unlimited cardinality might object to this as a harder requirement. I don't object.

reyang (approved):

specification/metrics/sdk.md
MrAlias: Notes from the is morning's spec SIG meeting:

> Consolidating all instrument measurement values into a single "OVERFLOW" attribute when cardinality exceeds the configured limit seems to have consensus on the approach here.

jmacd:
I plan to revise this PR with the fixes discussed in a Spec SIG several weeks ago.
https://github.com/open-telemetry/opentelemetry-specification/pull/2960#discussion_r1072749795

reyang (approved):

jmacd:
I posted an update to this PR based on feedback received in the January 6 Specification SIG meeting. I will close all the open conversations as I think this resolves them.  Feel free to continue discussing.

I did not complete a prototype of this, however I studied the matter in relation to the [Lightstep Metrics SDK](https://github.com/lightstep/otel-launcher-go/tree/main/lightstep/sdk/metric) which I maintain with several optimizations that Lightstep internally requested. That SDK takes a fairly different approach to memory management than the OTel-Go community SDK and I wanted to be sure that the optimizations we have made would not be compromised by the proposed cardinality limit feature.

These limits have been requested as a kind of "circuit breaker", in the sense that they protect the application and its connected services from itself.  When an application is behaving badly, it is essential to prevent buildup of memory everywhere in the SDK by implementing limits.  However, I do not want to take an additional lock or make an additional allocation for this purpose, if possible.

I considered whether to implement the limits at the point where attribute sets are first processed (i.e., at the metric events) as well as the point where they are assembled into a complete export (i.e., during the collection cycle).  While adding these limits inside the Views mechanism could be powerful and allow Views to correct cardinality problems, implementing limits late in the pipeline would not automatically protect the application from memory buildup early in the pipeline.

This leads to the idea of implementing limits as soon as possible in the pipeline.  Here, I found myself wanting an inexact implemention.  Considering synchronous instruments in particular, I propose to have the overflow take effect at a point in time before or during the collection cycle in which the limit is first exceeded.

This means (for synchronous instruments) the I would specify that an SDK can allow more than the limit number of timeseries to be read by a Reader, depending on concurrency and implementation details.  It is not a hard limit.  Like a circuit breaker, there is a delay betwen when the limit is exhausted and the change of SDK behavior.

My goal is that when an overflow condition is happening for synchronous instruments, the SDK can "simply" replace the event's attribute set with the overflow attribute set and then proceed on the normal code path.  That is, when overflow is happening, your attributes may be replaced with overflow attributes. 

As a result, when the limit is reached, I do not want any guarantee that metric attribute sets are uniquely mapped into one of the pre-overflow-existing timeseries or into the overflow timeseries (again, because concurrency and synchronization).  I suggest that SDKs should make an effort to ensure that existing timeseries are used when they were pre-existing (i.e., try to re-use timeseries from before the overflow, don't require it).

The Lightstep Metrics SDK is engineered so that when Delta aggregation temporality is used, stale timeseries can be removed from memory by the SDK automatically.  This makes a cardinality limit more difficult to reach--growth has to take place within one or two collection cycles (depending on concurrency).  The SDK should not be required to release stale entries (by some definition) before calculating whether a limit is reached. This scenario involving delta temporality is also why I do not want to require that events are uniquely mapped into a pre-existing timeseries or the overflow timeseries. In cases where there are two Delta temporality readers in use, transient overflow conditions may happen such that the longer-period exporter sees a mixture of overflow and non-overflow conditions, I just don't want to force some kind of correctness onto this situation.

Asynchronous instruments could have substantially different requirements specified.  Because asynchronous callback execution is per Reader, it is easy to make the limit hard and exact.  However, I think a cumulative temporality exporter for asynchronous instruments should not be required to use additional memory to implement the limit. This because there is no way to identify "pre-overflow-existing" timeseries without additional memory that cumulative asynchronous instruments are meant to avoid, so when limits are reached for asynchronous instruments I would anticipate an arbitrary subset of timeseries being exported.

PTAL. I don't really like the way this is specified, would like help. @MrAlias @reyang @dashpole especially.

specification/metrics/sdk.md
jmacd: PTAL latest changes.

specification/metrics/sdk.md
jmacd: PTAL latest changes.

specification/metrics/sdk.md
jmacd: PTAL latest changes.

specification/metrics/sdk.md
jmacd: PTAL latest changes.

jack-berg:
> Being able to correct cardinality problems via Views seems essential, and while this is a step forward, this feels like a big omission.

The view tool to correct cardinality problems is attribute filter. This cardinality limit is a circuit breaker blunt instrument. I think its appropriate to make it configurable at the meter provider level (at least initially) because I don't think its something we want users to lean on in normal operations. I.e. if you see `overflow=true` in your data, something has gone wrong. Go fix it with a view that filters attributes to reduce cardinality. 

> That doesn't seem like it would help cumulative exporters, though, unless i'm misunderstanding.

I think your example is incorrect. Given recordings to the following attributes for collection windows:

1. A, B, C
2. D, E, F
3. A, C, E

And a cardinality limit of 3, I'd expect to see exports for the collection windows as follows:

1. A, B, C
2. A, B, C, O
3. A, B, C, O

When temporality is cumulative, the SDK maintains state for all series its seen since start. If a measurements are reported for a new series and the cardinality limit is exceeded, those measurements go to the series `overflow=true`.

dashpole:
> I think your example is incorrect.

I think you are probably correct, and your example is how I would want it to work.  I was misinterpreting how it would work with cumulatives.

> Go fix it with a view that filters attributes to reduce cardinality.

Thats exactly what I want to be able to do, so maybe this is another misunderstanding on my part. My interpretation of "[this won't] allow Views to correct cardinality problems" and "implementing limits late in the pipeline would not automatically protect the application from memory buildup early in the pipeline" was that the limit is applied to the cardinality of **pre-view-filtered** attribute sets, rather than the **post-view-filtered** attribute sets.

As a user, If i'm using an instrumented library that adds an attribute which causes me problems, i'd like to be able to fix it by adding a view which filters out the problematic attribute.  If I added the view, but still get overflow, i'd be pretty disappointed.

jmacd:
@dashpole thanks for probing this question. I did explore the idea of setting cardinality limits at the level of an individual view clause or in the SDK-wide default for each instrument. 

It is appealing to add per-view-clause limits that could be considered hard limits on the output size. I briefly considered carrying this out -- I would put the new optional cardinality limit alongside the Aggregator settings (e.g., histogram size) so that on a per-view basis we have a way to control this. This raises configuration questions, but they're beside the matter here (e.g., do you also configure per-MetricReader defaults based on instrument kind?). 

We could imagine propagating the maximum limit of the configured Views for an instrument forward in the pipeline(s) so that the first piece of code that handles metric events knows something about the downstream limits, which leads the the question -- then what? What is required of the SDK when by Reader/View configuration there are both filters and limits in place? I have concerns here that we are going to add difficult requirements to the SDKs. To apply a filter is a relatively expensive operation, and we don't have widely-available "bound instrument" APIs to amortize the cost of attribute set handling. The SDK will either have to apply every filter to every event so that each Reader/View can independently manage the limits, or it will take on new synchronization, or it will have to have separate limits for the "front" part of the SDK, especially the part that deals with synchronous instruments.

Yes, I think we can add per-View and per-Reader limits, AND think the "circuit-breaker" limits we are talking about in this PR should be considered a separate safety mechanism that applies to a different part of the SDK at the front of the pipeline(s). I would recommend that SDKs configure a circuit-breaker limit at cardinality greater than the maximum cardinality of any single view. (As in an electrical circuit, the circuit breaker is generally larger than the rated load on the circuit.)

I did end up prototyping this: https://github.com/lightstep/otel-launcher-go/pull/385

The organization of the Lightstep metrics SDK has a "front" part that caches a `uint64` fingerprint of the original attribute set for synchronous instruments in a transient map that can be emptied after the attribute set is no longer in use. In that SDK, the circuit-breaker is to protect against short-term growth of that map in the Update() code path. Per-view limits would be orthogonol, treated in the Collect() code path. Does that make sense?

dashpole:
@jmacd Thanks!

To be clear, i'm less concerned about how a user configures the breaker or how granular it is, and more concerned with whether it is before or after filtering.

Here is a quick prototype I threw together for what I had in mind: https://github.com/dashpole/opentelemetry-go/pull/1/files.  I believe this should protect against short-term growth of the Map in the Update() path for Sums, and be checked immediately after the attribute set is filtered (which is the first thing done on the Update() path).

A few thoughts:

* When thinking about cardinality explosion, i'm primarily concerned with the growth of internal data structures in the SDK eating an unbounded amount of memory.  Regardless of limits we impose here, the user or instrumentation library is still going to do O(cardinality) work to construct each attribute set.  Its possible calling Filter is still too expensive to be done in cases where we really need a circuit breaker, but It would be nice to see that demonstrated.
* Placing the limits on data structures within the SDK that are per-attribute-set made this fairly trivial to implement in Go, as I just looked for where attribute.Set was used as a key.  (waves hands) It feels like that approach might work in other languages as well.  One difficulty with adding a pre-filter attribute set limit is that many SDKs (at least Go) don't actually have any data structures based on pre-filtered attribute sets.  They would presumably need to add a new data structure to track the sets seen before filters are applied.
* One other question: Wouldn't Delta and Cumulative Readers want different overflow behavior?  The cumulative one wants to limit to a stable set of attributes over time, otherwise the cumulative cardinality of will still explode.  The delta one presumably doesn't want to track the attributes from collection-to-collection, as that is a benefit of using delta.  Would performing overflow checking once-per-instrument mean that delta readers would still pay the attribute tracking cost?

jmacd:
> Placing the limits on data structures within the SDK that are per-attribute-set made this fairly trivial to implement in Go, 

I see a recent change https://github.com/open-telemetry/opentelemetry-go/pull/3695 simplifies the matter. 

I would agree that per-view configurable cardinality limits make sense -- especially if we can first agree that SDKs are allowed to implement a pipeline with intermediate stages that are able to overflow by design.  The design prior to https://github.com/open-telemetry/opentelemetry-go/pull/3695 used an intermediate data structure, but only as an optimization. I would recommend that implementations use the MaxTimeseries limit to configure pruning of data structures that can be pruned.

Generalizing, the `MaxTimeseries` limit is meant as as a threshold above which performance AND correctness may degrade. In a hypthetical filter mapping optimization, the intermediate structure can be pruned but it will cause performance loss under high cardinality. In the Lightstep metrics SDK examplke, the intermediate structure _could_ be flushed before overflow to push the problem down to the aggregators, but I would not prefer this approach. I am looking for us to agree that any pipeline component that needs to overflow by configuration and design SHOULD use a specified attribute set as the overflow indicator.

I would also like to survey the other SDKs (we're using Go implementations exclusively in this discussion). How do you approach intermediate data structures used to lower the cost of attribute handling, for example?

specification/metrics/sdk.md
```diff
@@ -597,6 +605,83 @@ Attributes which belong to Metrics are exempt from the
...
+
+In this version of the specification, the SDK's `MaxTimeseries`
+setting limits all instruments and all instrumentation libraries
+equally; future versions of this specification may support finer-grain
+configuration of cardinality limits.
+
+Implementations details determine how concurrent API events will be
+handled.  To enforce the `MaxTimeseries` limit:
+
+- An instrument with no configured views SHOULD NOT implement
```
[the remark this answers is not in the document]
MrAlias: Yeah, I'm a bit confused here as well. The default aggregations will still be used if no view is provided by the user, right? Wouldn't we want to have limits in this situation as well?

specification/metrics/sdk.md
```diff
@@ -597,6 +605,83 @@ Attributes which belong to Metrics are exempt from the
...
+  cardinality limits.
+- An instrument with one or more configured views SHOULD detect that
+  the limit was reached before or during the next collection event.
+- An instrument is NOT REQUIRED to be remove stale timeseries (e.g.,
```
MrAlias, line 644:
```suggestion
- An instrument is NOT REQUIRED to remove stale timeseries (e.g.,
```

specification/metrics/sdk.md
```diff
@@ -597,6 +605,83 @@ Attributes which belong to Metrics are exempt from the
...
+  exclude zero-valued sums from consideration) before evaluating the
+  cardinality limit.
+
+When the limit is reached by an the individual metric instrument,
+subsequent new timeseries created for that instrument will have their
+complete attribute set replaced by a single-attribute named `overflow`
+with value `true`.
```
MrAlias, line 651:
```suggestion
with the boolean value `true`.
```

specification/metrics/sdk.md
```diff
@@ -597,6 +605,83 @@ Attributes which belong to Metrics are exempt from the
...
+  cardinality limits.
+- An instrument with one or more configured views SHOULD detect that
+  the limit was reached before or during the next collection event.
+- An instrument is NOT REQUIRED to be remove stale timeseries (e.g.,
+  exclude zero-valued sums from consideration) before evaluating the
+  cardinality limit.
+
+When the limit is reached by an the individual metric instrument,
+subsequent new timeseries created for that instrument will have their
+complete attribute set replaced by a single-attribute named `overflow`
```
MrAlias, line 650: Should this use the `otel.*` namespace?

Maybe?

```suggestion
complete attribute set replaced by a single-attribute named `otel.metric.instrument.overflow`
```

specification/metrics/sdk.md
```diff
@@ -597,6 +605,83 @@ Attributes which belong to Metrics are exempt from the
...
+
+The implementation is permitted to evaluate the cardinality limit at a
+time when it is convenient.  Implementations are NOT REQUIRED to
+exactly meet the cardinality limit when it is reached, because that
+implies an unwanted degree of synchronization.  Implementations MAY
+exceed the limit due to the effects of concurrency.  Implementations
+MAY defer the overflow test until the nexts collection cycle.
+
+When cardinality limits are reached, the implementation is REQUIRED
+eventually to stop creating new timeseries for the instrument.
```
MrAlias, line 661: This is a pretty vague requirement.

If we don't have hard criteria for when this should happen we should probably just make this a recommendations to the SDK authors to do this as soon as they can to save memory.

Or maybe the requirement is just that the algorithm used needs to ensure there is never a situation where an infinite amount of memory would be used before the limit is applied?

specification/metrics/sdk.md
```diff
@@ -597,6 +605,83 @@ Attributes which belong to Metrics are exempt from the
...
+The SDK SHOULD try to maintain correctness for timeseries that existed
+before the overflow event.  However, if this feature requires
+substantial additional memory the SDK SHOULD prefer to export an
+arbitrary subset of timeseries when the limit is reached.
+
+### Specific cardinality limits
+
+The SDKs that support cardinality limits MUST support the configuration and enforcement of the following limits.
+
+- The number of timeseries per instrument, with a RECOMMENDED default value 2000.
```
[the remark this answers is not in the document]
MrAlias: I think from the above section this is per collection cycle, right?

specification/metrics/sdk.md
```diff
@@ -597,6 +605,83 @@ Attributes which belong to Metrics are exempt from the
...
+
+All SDKs MUST NOT support the configuration or enforcement of any limit not listed above.
```
MrAlias, line 679: @jack-berg I thought Java already has other limits, right? Or am I miss-remembering?

dashpole:
> I am looking for us to agree that any pipeline component that needs to overflow by configuration and design SHOULD use a specified attribute set as the overflow indicator.

I think the specified attribute set is a reasonable approach, although I would also support simply dropping the observations.  (side question: is the overflow attribute set actually useful for gauges?) Regardless, I would want have a self-observability metric giving me a counter of overflow events as well, broken down by metric name so I can write an alert for when this occurs.

> How do you approach intermediate data structures used to lower the cost of attribute handling, for example?

I would specify it without respect to the part of the SDK in which the data structures exists, if possible.  Something like "The SDK should not keep track of more than `MaxTimeseries` unique attribute sets for any instrument in any data structure in the SDK" (with the usual caveats about precision and synchronization).

I would (personally) separately specify something like "the SDK should avoid any caching based on attribute sets until after filters have been applied so that filters can be used to alleviate cardinality problems".  I think this can be considered separately from the topic of SDK cardinality limits, as long as this proposal doesn't require cardinality limits to be enforced prior to filtering.

specification/metrics/sdk.md
MrAlias, line 650: This means that when the limit is hit for an instrument the only data that will be sent to an export pipeline is one attribute ("overflow", `true`) and the summed value of all measurements, right?

specification/metrics/sdk.md
jack-berg: No we currently just have a hardcoded limit of 2000 per instrument.

specification/metrics/sdk.md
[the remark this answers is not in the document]
MadVikingGod:
```suggestion
- An instrument that is selected by no views SHOULD NOT implement
  cardinality limits.
```
I was confused by this as well.  Are we trying to say that if the instrument is not selected by a view, and thus won't export, does not need to implement this?  Assuming that it will be a drop aggregation and not store any of these values anyways.

specification/metrics/sdk.md
```diff
@@ -597,6 +605,83 @@ Attributes which belong to Metrics are exempt from the
 time. Attribute truncation or deletion could affect identity of metric time
 series and the topic requires further analysis.
 
+## Cardinality limits
+
+**Status**: [Experimental](../document-status.md)
+
+MeterProviders SHOULD support being configured with a limit applied to
```
MadVikingGod, line 612: Is this something that should be at the MeterProvider when it looks like it's setting a per-Instrument or per-Aggreagtion limit?

specification/metrics/sdk.md
```diff
@@ -597,6 +605,83 @@ Attributes which belong to Metrics are exempt from the
...
+
+- An instrument with no configured views SHOULD NOT implement
+  cardinality limits.
+- An instrument with one or more configured views SHOULD detect that
+  the limit was reached before or during the next collection event.
+- An instrument is NOT REQUIRED to be remove stale timeseries (e.g.,
+  exclude zero-valued sums from consideration) before evaluating the
+  cardinality limit.
+
+When the limit is reached by an the individual metric instrument,
```
MadVikingGod, line 648: How does this interact with filters in a view?  Should the limit be counted before any attributes are removed or after?  If I have two instruments that filter to the same time series how is the limit calculated?

jmacd:
Thank you @MrAlias @MadVikingGod @dashpole @jack-berg.

Your feedback convinced me to totally change my approach to this PR, and I've rewritten it from scratch after making a new prototype. My prototype is in https://github.com/lightstep/otel-launcher-go/pull/385.

In changing my approach to this PR, I realized that there is a second cardinality limit that makes sense for the Lightstep metrics SDK that may not make sense for every SDK. The Lightstep SDK has an intermediate data structure which holds aggregate values between collection events; in my prototype, I used a separate limit to avoid memory blowup in the intermediate state. For synchronous instruments, this can be seen as memory limit -- it could be derived from the combined views in effect (roughly).

I'm not sure an instrument-level limit makes sense for asynchronous instruments, since any state is built per collection. That's why I specified that the first observations should be preferred -- ideally there is no intermediate state and observations pass straight to the aggregators, with each view setting an independent cardinality limit.

specification/metrics/sdk.md
jmacd: I replaced this older approach with a new one in which Views configure limits. PTAL.

specification/metrics/sdk.md
jmacd: In my prototype w/ the LS metrics SDK ([as detailed below](https://github.com/open-telemetry/opentelemetry-specification/pull/2960#issuecomment-1454203830)), I used an instrument-level limit and a View-level limit. The instrument-level limit is only needed because there's an intermediate data structure where an independent limit makes sense. I excluded this part from this PR, which only proposes (now) a view-level cardinality limit so that views can correct overflow problems.

specification/metrics/sdk.md
jmacd: I replaced this older approach with a new one in which Views configure limits. PTAL.

specification/metrics/sdk.md
jmacd: I replaced this older approach with a new one in which Views configure limits. PTAL.

specification/metrics/sdk.md
jmacd: May I propose `otel.metric.overflow=true`? I'm afraid the extra `instrument` is a little too verbose.

specification/metrics/sdk.md
jmacd: Fixed.

specification/metrics/sdk.md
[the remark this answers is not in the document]
jmacd: I replaced this older approach with a new one in which Views configure limits. PTAL.

specification/metrics/sdk.md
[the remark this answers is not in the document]
jmacd: Yes, I clarified this point. PTAL.

specification/metrics/sdk.md
```diff
@@ -581,6 +590,50 @@ execution.
...
+
+View configuration SHOULD support applying per-view cardinality linits.
+
+The cardinality limit is taken as an exact, hard limit on the number
+of data points that can be written per collection.  Each view MUST NOT
+output more than the configured cardinality limit number of data ponts
+per period.
+
+An overflow attribute is defined, named `otel.metric.overflow` having
+(boolean) value `true`, used to report a synthetic aggregation of the
```
pirgeo, line 608: `otel.metric.overflow` would be `true` until the process restarts, right? There is no way it is ever set to `false`, or am I missing something?

specification/metrics/sdk.md
```diff
@@ -581,6 +590,50 @@ execution.
...
+metric events that could not be independently aggregated because of
+the limit.  The overflow attribute set MUST be included in the limit
+calculation, thus the maximum number of distinct, non-overflow
+attributes is one less than the limit.
```
pirgeo, line 612: We only need to add 1 here since adding the `otel.metric.overflow=true` does not change the number of metric streams that are open. In other words, adding the same attribute to all streams does not make the cardinality explode, correct?

pirgeo: Or would, as soon as the overflow is detected, all further data collected on the metrics stream that has no attributes **but** the otel.metric.overflow attribute?

specification/metrics/sdk.md
```diff
@@ -581,6 +590,50 @@ execution.
...
+The RECOMMENDED default cardinality limit is 2000.
+
+#### Synchronous instrument cardinality limits
+
+Views of synchronous instruments with cumulative aggregation
+temporality MUST continue export the all attribute sets that were
+observed prior to the beginning of overflow.  Metric events
+corresponding with attribute sets that were not observed prior to the
+overflow will be reflected in a single data point described by (only)
+the overflow attribute.
```
pirgeo, line 623: Will the single data point contain the sum of all the other metrics streams that we were unable to keep in memory since the cardinality limit was reached?

specification/metrics/sdk.md
```diff
@@ -581,6 +590,50 @@ execution.
...
+
+Views of synchronous instruments with delta aggregation temporality
+can choose an arbitrary subset of attributes to output.  Regardless of
```
pirgeo, line 626: An arbitrary subset of attribute sets, that is, correct? An attribute set would represent one metrics stream in memory.

pirgeo (commented):
A few clarifying questions

specification/metrics/sdk.md
jack-berg: > Or would, as soon as the overflow is detected, all further data collected on the metrics stream that has no attributes but the otel.metric.overflow attribute?

This.

specification/metrics/sdk.md
jack-berg: If temporality is delta, all the series are cleared on collection and the `otel.metric.overflow=true` series goes away with the rest of them. But yet, `otel.metric.overflow` should never be `false`.

specification/metrics/sdk.md
jack-berg: Yes.

specification/metrics/sdk.md
```diff
@@ -234,6 +237,12 @@ are the inputs:
     `exemplar_reservoir` (optional) to use for storing exemplars. This should be
     a factory or callback similar to aggregation which allows different
     reservoirs to be chosen by the aggregation.
+  * **Status**: [Experimental](../document-status.md) - the
+    `cardinality_limit` (optional) associated with the view.  This
+    should be an positive integer to be taken as a hard limit on the
+    number of data points that will be emitted during a single
+    collection by a single instrument.  See [cardinality limits](#cardinality-limits),
+    below.
```
jack-berg, line 245: You've removed the meterprovider level configuration in favor of only a view level option. While I think its ok to have a view level cardinality limit option, it shouldn't be the only way because views have sharp edges that make them somewhat surprising to use:

Suppose I want to change the default cardinality limit for all instruments.. I register a view that selects all instruments and sets a cardinality limit of let's say 100. Separately, I want to change the histogram buckets boundaries for an instrument, so I register a view which selects `http.server.duration` and sets the bucket boundaries to be my new values, let's say `[1, 2, 3, 4, 5]`. 

A user is might expect to have `http.server.duration` output with a cardinality limit of 100, and bucket boundaries of `[1, 2, 3, 4, 5]`, but in reality they'll get two metrics:
- One with name `http.server.duration`, a cardinality limit of 100, and the default bucket boundaries of `[0, 5, 10, ...]`
- Another with name `http.server.duration`, the default cardinality limit of 2000, and bucket boundaries of `[1, 2, 3, 4, 5]`. 

This somewhat surprising behavior falls out from the fact that we don't try to do any merging when multiple views match an instrument. Users must be careful that each instrument has only one matching view or get confusing duplicated metrics.

This is a good enough reason to add a meterprovider level configuration option.

specification/metrics/sdk.md
pirgeo: Ah, yeah, that makes sense, I didn't think about that. Thanks for clarifying!

specification/metrics/sdk.md
```diff
@@ -581,6 +590,50 @@ execution.
...
+
+The cardinality limit is taken as an exact, hard limit on the number
+of data points that can be written per collection.  Each view MUST NOT
+output more than the configured cardinality limit number of data ponts
+per period.
+
+An overflow attribute is defined, named `otel.metric.overflow` having
+(boolean) value `true`, used to report a synthetic aggregation of the
+metric events that could not be independently aggregated because of
+the limit.  The overflow attribute set MUST be included in the limit
```
joaopgrassi, line 610: I read this multiple times and might be just me being stupid but I lack a clear, normative statement of what SDKs should do. I feel that the text is implicitly telling that SDKs need to put all data into a new stream that contains ONLY the overflow attribute but I wonder if it wouldn't be best to "split" the explanation of the problem and WHAT the SDK should do. Something like:

>An overflow attribute is defined, named `otel.metric.overflow`. The overflow attribute is used to report a synthetic >aggregation of the metric events that could not be independently aggregated because of the limit. 

> The overflow attribute set MUST be included in the limit calculation, thus the maximum number of distinct, non-overflow attributes is one less than the limit.

> When the cardinality limit is reached, the SDK MUST create a new metric stream containing the `otel.metric.overflow`. The attribute MUST have a value of `true`

specification/metrics/sdk.md
jmacd: I'm interested in breaking out a separate issue to discuss the problem brought up by your example above.

The way I was thinking about it, the View mechanism could configure the cardinality limit, but like with other defaults, there would be a per-instrument-kind default specified at the `MetricReader` level. I've added one bullet to the constructor for `MetricReader` to reflect this position.

specification/metrics/sdk.md
jmacd: I clarified this topic, hopefully it's better now.

specification/metrics/sdk.md
jmacd: Reworded:
```
An overflow attribute set is defined, containing a single attribute
`otel.metric.overflow` having (boolean) value `true`, which is used to
report a synthetic aggregation of the metric events that could not be
independently aggregated because of the limit.

The SDK MUST create an Aggregator with the overflow attribute set
prior to reaching the cardinality limit and use it to aggregate events
for which the correct Aggregator could not be created.  The maximum
number of distinct, non-overflow attributes is one less than the
limit, as a result.
```
How's that?

specification/metrics/sdk.md
jmacd: I reworded this, PTAL.

specification/metrics/sdk.md
jmacd: This sentence was misleading. I've reworded the section to read:
```
Views of synchronous instruments with delta aggregation temporality
MAY choose an arbitrary subset of attributes to output.

Regardless of aggregation temporality, the SDK MUST ensure that every
metric event is reflected in exactly one Aggregator, which is either
an Aggregator associated with the correct attribute set or an
aggregator associated with the overflow attribute set.

Events MUST NOT be double-counted or dropped during overflow an
overflow.
```

specification/metrics/sdk.md
```diff
@@ -853,6 +915,7 @@ SHOULD provide at least the following:
 * The `exporter` to use, which is a `MetricExporter` instance.
 * The default output `aggregation` (optional), a function of instrument kind.  If not configured, the [default aggregation](#default-aggregation) SHOULD be used.
 * The default output `temporality` (optional), a function of instrument kind.  If not configured, the Cumulative temporality SHOULD be used.
+* The default aggregation cardinality limit to use, a function of instrument kind.  If not configured, a default value of 2000 SHOULD be used.
```
MadVikingGod, line 918: Isn't the limit view-scoped?  Why would we need the instrument kind?

MadVikingGod (approved):
Observation: Because this is view scoped, some synchronizing between different aggregators will be necessary.  There will need to be some way of tracking how many unique data points across the view exist to understand if an aggregator can add new ones. I don't know if this is a good or bad decision, but this is the first requirement requiring this sync on the write side (other requirements make you have a sync on the collect).

How do the Synchronous and Asynchronous instruments interact?  I ask because we won't know how many the Asynchronous will produce until collect, but we will know the synchronous as the program runs.  Some example:
We set a limit of 10 in one view.  We have some synchronous instruments that will produce 7 data points; we will know this before collect.  We also have asynchronous instruments that will produce 5 data points; this will be known during collect.  Should we drop the last 2 asynchronous data points?
Again, we have a limit of 10.  We have a synchronous that, before the first collect, produces 5, and an async that also produces 5.  After the first collect the synchronous instruments will produce 7.  Do 2 synchronous data points get dropped?

specification/metrics/sdk.md
jmacd: I think we're having some kind of terminology mix-up. I don't intend for there to be interaction between instruments. When I refer to a single "view" I mean the one or more individual aggregations that results from evaluating all the View inputs. 

[This terminology is not precise in the specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#view), in any case I would suggest we add the term "aggregation" to the proposed limit, as in `aggregation_cardinality_limit` to indicate that the limit is per instrument-output. I want it to be clear that there's no synchronization required between instruments or between aggregations assigned to the same instrument.

pirgeo (commented):
I think to me it is not entirely clear that the limit will always be applied **per metric name** (what you call aggregation, I think). 

Also: I am not really sure how to configure it. Can I say "all my metric aggregations should have a maximum of X metric streams each" or can I say "The metric with name X should be allowed to have Y metric streams, while all others should only have a maximum of X metric streams"? My understanding of Views is not as extended as I would like, but I think both of these should be possible?

specification/metrics/sdk.md
joaopgrassi: Looks great!

jack-berg (approved):
This seems good and is experimental so leaves room further improvements if issues are discovered during implementation. Let's get this merged.

srikanthccv:
This had a couple of approvals and no blocking comments. @jmacd, do you plan to re-open this?

specification/metrics/sdk.md
```diff
@@ -582,6 +591,62 @@ execution.
 The implementation MUST complete the execution of all callbacks for a
 given instrument before starting a subsequent round of collection.
 
+### Cardinality limits
```
tylerbenson, line 594: Maybe I'm misunderstanding, but it seems like this cardinality limit only protects the metric pipeline that the data is being sent to, but does not help prevent memory abuse due to cardinality problems on the process collecting the metrics in the first place.  This is unfortunate as that is usually something a user cares strongly about. Users should be given the tools to put cardinality limits in place for each instrument, in addition to being able to configure default limits for all instruments.

tylerbenson: Is the cardinality limit on ingest/collection in the process, or only on export? If the limit is on the in-memory state, then it works for both.  If the limit is only for export, then the in-memory representation can still grow unconstrained.

tylerbenson: Ok, sounds good.  Thanks for clarifying.

jmacd:
Addressing @tylerbenson's https://github.com/open-telemetry/opentelemetry-specification/pull/2960#discussion_r1170565775
https://github.com/open-telemetry/opentelemetry-specification/pull/3216

Having spoken with @tylerbenson and @jack-berg, it appears that this specification will indeed satisfy the need to protect the application from memory abuse. What @tylerbenson would like to see is an Advice mechanism, following the example in https://github.com/open-telemetry/opentelemetry-specification/pull/3216, so that instruments can be declared with optional API-level advice indicating the desired cardinality limit. Basically, we think it makes sense to add the limit as a View configuration because Views are the foundation of the metrics SDK's functionality. At the same time, we recognize that Views are not likely how these limits will be set in most cases. Therefore this PR also specifies a Reader-level default, and we expect in a future PR to extend this to the API via Advice, so that developers can declare cardinality limits in code. I suggest we accept this PR in its current form, we can file an issue to follow it with Advice support.

specification/metrics/sdk.md
jmacd: https://github.com/open-telemetry/opentelemetry-specification/pull/2960#issuecomment-1515244723

specification/metrics/sdk.md
reyang: > This is a good enough reason to add a meterprovider level configuration option.

+1

carlosalberto (approved):

carlosalberto:
@open-telemetry/specs-approvers Please take a look at the latest iteration. This PR has been discussed/standing for a long time now, and would be great to have it adjusted/merged soon.

specification/metrics/sdk.md
```diff
@@ -862,6 +927,7 @@ SHOULD provide at least the following:
 * The `exporter` to use, which is a `MetricExporter` instance.
 * The default output `aggregation` (optional), a function of instrument kind.  If not configured, the [default aggregation](#default-aggregation) SHOULD be used.
 * The default output `temporality` (optional), a function of instrument kind.  If not configured, the Cumulative temporality SHOULD be used.
+* The default aggregation cardinality limit to use, a function of instrument kind.  If not configured, a default value of 2000 SHOULD be used.
```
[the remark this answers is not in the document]
jmacd: I was trying to follow the existing configuration structure, to simplify the task for SDK authors. Since there are already per-kind defaults, adding a new per-kind default made sense to me. OTOH, I see your point and I would be glad to specify an SDK-wide default without reference to instrument kind.

pirgeo (approved):

specification/metrics/sdk.md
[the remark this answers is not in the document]
alanwest: I don't have a strong opinion that a per-kind default should not exist, so I guess an SDK-wide default could be additive to what you have here in this PR.

I was more considering this from an end user usability perspective. Could we say "SDKs MAY also offer a way to set a global default irrespective of instrument kind. Any limit applied to instrument kind overrides the global default."?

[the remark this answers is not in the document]
jack-berg: > specify an SDK-wide default without reference to instrument kind.

For the java implementation I would probably start with this and add the reader / view level options once users ask for it.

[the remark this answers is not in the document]
jmacd: @jack-berg does the "a function of instrument kind" part bother you, or does it make sense? I was thinking about how histograms are more expensive to send than other instruments, to a different limit for histograms than for other points could be useful.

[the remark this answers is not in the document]
jack-berg: Although I don't think it will be very popular, I don't think it bothers me. Just because a user has the option of changing the cardinality limit as a function of instrument kind doesn't mean they have to. I don't think it changes to the complexity of the implementation much to offer it.

specification/metrics/sdk.md
```diff
@@ -235,6 +238,12 @@ are the inputs:
     `exemplar_reservoir` (optional) to use for storing exemplars. This should be
     a factory or callback similar to aggregation which allows different
     reservoirs to be chosen by the aggregation.
+  * **Status**: [Experimental](../document-status.md) - the
+    `aggregation_cardinality_limit` (optional) associated with the view.  This
+    should be an positive integer to be taken as a hard limit on the
```
alanwest, line 243:
```suggestion
    should be a positive integer to be taken as a hard limit on the
```

alanwest (approved):

lzchen (approved):

carlosalberto:
@jmacd Should we include this in the May release?

carlosalberto:
Talked to @jmacd, we are ready to roll this in. Thanks a lot for the iteration (and patience)!

PR #2961, opened 2022-11-18
Specify a low-memory temporality preference for the OTLP metrics exporter

### LINKED ISSUE #2738: Minimum-memory-usage temporality preference for OTLP exporter

In https://github.com/open-telemetry/opentelemetry-java/issues/4680, users are looking for an easy way to configure the minimum-memory settings. This means choosing delta temporality for synchronous instruments and cumulative temporality for asynchronous instruments.

I propose the name "stateless". [Lightstep implements this setting in its OTel launcher](https://github.com/lightstep/otel-launcher-go/#metrics-temporality-settings), because this configuration works well for Lightstep. [This update applies to the OTLP exporter specification.
](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk_exporters/otlp.md).

**Additional context.**

This was discussed in #2404.

jmacd:
Fixes #2738

## Changes

Adds a low-memory aggregation temporality preference configuration for the OTLP metrics exporter, "lowmemory" which configures the OTLP exporter to use delta temporality for synchronous Counter and Histogram instruments.

Related issue: #1891

Note that a parallel change https://github.com/open-telemetry/opentelemetry-specification/pull/2960 specifies limits and states that SDKs SHOULD remove timeseries from memory when they are stale and Delta aggregation temporality is configured, which they will be able to do in conjunction with the "lowmemory" preference here.

Note that this could benefit users even when metrics data is collected by Prometheus servers. The [OTel Collector's `prometheusexporter` supports delta-to-cumulative translation](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/cf8c031b2374b54592ee2d4bd757e2e44c8311a5/exporter/prometheusexporter/accumulator.go#L213), making it possible with a "lowmemory" OTLP exporter configuration to shift the memory requirement for cumulative timeseries calculation from the client to a collector.


specification/metrics/sdk_exporters/otlp.md
```diff
@@ -49,6 +49,14 @@ The recognized (case-insensitive) values for `OTEL_EXPORTER_OTLP_METRICS_TEMPORA
 * `cumulative`: Choose cumulative aggregation temporality for all instrument kinds.
 * `delta`: Choose Delta aggregation temporality for Counter, Asynchronous Counter and Histogram instrument kinds, choose
   Cumulative aggregation for UpDownCounter and Asynchronous UpDownCounter instrument kinds.
+* `lowmemory`: This configuration uses Delta aggregation temporality for Synchronous Counter and Histogram and uses Cumulative aggregation temporality for Asynchronous Counter, UpDownCounter and Asynchronous UpDownCounter instrument kinds.
```
reyang, line 52: Given Ln47 mentioned "case-insensitive", would it be a good idea to use "LowMemory" for better readability (and capitalize "Cumulative"/"Delta" for consistency)?

reyang (approved):

carlosalberto (approved):

specification/metrics/sdk_exporters/otlp.md
jack-berg, line 52: Why cumulative temporality for synchronous up down counter?

jmacd: The reasoning is the same that led the "delta" preference to produce cumulative UpDownCounter.

Discussed here: https://github.com/open-telemetry/opentelemetry-specification/issues/2738#issuecomment-1309107016

The reason I used "lowmemory" instead of "minmemory" or "stateless" (as I had originally proposed) is that there exists a reasonable configuration with _all_ synchronous instruments configured for delta temporality, it's just that that makes it difficult to convert UpDownCounter to Gauge data points for systems that do not support non-monotonic deltas.

jack-berg: Makes sense. Useful to distinguish "lowmemory" from "minmemory".

jack-berg, line 52:
```suggestion
* `lowmemory`: This configuration uses Delta aggregation temporality for Synchronous Counter and Histogram and uses Cumulative aggregation temporality for Synchronous UpDownCounter, Asynchronous Counter, and Asynchronous UpDownCounter instrument kinds.
```

jack-berg (approved):

specification/metrics/sdk_exporters/otlp.md
jmacd: Done.

jmacd: Another reason can be stated: the OTel collector's Prom-Exporter support for delta-to-cumulative translation does not apply to non-monotonic counters.

PR #2963, opened 2022-11-18
Specify how to deal with invalid numeric envvars

### LINKED ISSUE #2962: Invalid numeric configuration environment variable value handling is unspecified
While all other environment variable value types have a specification for how to deal with invalid/unparseable values, the section for numeric values does not.

See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md

This is what most other value types have:

>If (conditions for being invalid) the SDK MUST generate a warning, gracefully ignore the setting and use the default value if it is defined.

This originally came up in https://github.com/open-telemetry/opentelemetry-specification/pull/2959#discussion_r1025760510

Oberon00:
Fixes #2962

## Changes

Specify handling of invalid and out-of-range numeric configuration environment variable values.


specification/sdk-environment-variables.md
```diff
@@ -27,6 +27,16 @@ Renaming or changing the default value MUST NOT happen without a major version u
 
 If an SDK chooses to support an integer-valued environment variable, it SHOULD support nonnegative values between 0 and 2³¹ − 1 (inclusive). Individual SDKs MAY choose to support a larger range of values.
 
+For variables accepting an numeric value, if the user provides a value the SDK cannot parse,
+or which is outside the valid range for the configuragion item SDK SHOULD
+generate a warning and gracefully ignore the setting. In particular, SDKs
+SHOULD NOT assign a custom interpretation e.g. to negative values if a negative
+value does not naturally apply to a configuration, but treat it like any other
+invalid value.
```
arminru, line 35: I think some examples would be helpful here ("e.g., for attribute length limits").

specification/sdk-environment-variables.md
```diff
@@ -27,6 +27,16 @@ Renaming or changing the default value MUST NOT happen without a major version u
 
 If an SDK chooses to support an integer-valued environment variable, it SHOULD support nonnegative values between 0 and 2³¹ − 1 (inclusive). Individual SDKs MAY choose to support a larger range of values.
 
+For variables accepting an numeric value, if the user provides a value the SDK cannot parse,
+or which is outside the valid range for the configuragion item SDK SHOULD
```
arminru, line 31:
```suggestion
or which is outside the valid range for the configuration item, the SDK SHOULD
```

specification/sdk-environment-variables.md
```diff
@@ -27,6 +27,16 @@ Renaming or changing the default value MUST NOT happen without a major version u
...
+generate a warning and gracefully ignore the setting. In particular, SDKs
```
arminru, line 32:
```suggestion
generate a warning and gracefully ignore the setting, i.e., treat them as not set.
In particular, SDKs
```

specification/sdk-environment-variables.md
Oberon00:
```suggestion
invalid value (for example, a value specifying a buffer size must naturally be non-negative. Treating a negative value as "buffer everything" would be an example of such a discouraged custom interpretation. Instead the default buffer size should be used. Note that this could make a difference even if the custom interpretation is identical with the default value, because it might reset a value set from other configuration sources with the default.)
```

specification/sdk-environment-variables.md
```diff
@@ -27,6 +27,17 @@ Renaming or changing the default value MUST NOT happen without a major version u
 
 If an SDK chooses to support an integer-valued environment variable, it SHOULD support nonnegative values between 0 and 2³¹ − 1 (inclusive). Individual SDKs MAY choose to support a larger range of values.
 
+For variables accepting an numeric value, if the user provides a value the SDK cannot parse,
+or which is outside the valid range for the configuration item, the SDK SHOULD
+generate a warning and gracefully ignore the setting, i.e., treat them as not set.
+In particular, SDKs
+SHOULD NOT assign a custom interpretation e.g. to negative values if a negative
+value does not naturally apply to a configuration, but treat it like any other
```
Oberon00, line 35: I guess this is important to note. I don't want to forbid the specification from specifying certain interpretations for such values, I just want to stress that if it doesn't, then implementations should not make something up.

```suggestion
value does not naturally apply to a configuration and does not have an explicitly specified meaning, but treat it like any other
```

specification/sdk-environment-variables.md
arminru: Sounds good to me.

CHANGELOG.md
```diff
@@ -181,6 +181,8 @@ release.
   ([2679](https://github.com/open-telemetry/opentelemetry-specification/pull/2679))
 - Add the definition of a Boolean environment variable
   ([#2755](https://github.com/open-telemetry/opentelemetry-specification/pull/2755)).
+- Specify handling of invalid numeric environment variables
+  ([#2963](https://github.com/open-telemetry/opentelemetry-specification/pull/2963))
```
arminru, line 185: This slipped into the 1.15.0 section but should go into Unreleased instead.

specification/sdk-environment-variables.md
```diff
@@ -27,6 +27,21 @@ Renaming or changing the default value MUST NOT happen without a major version u
...
+SHOULD NOT assign a custom interpretation e.g. to negative values if a negative
+value does not naturally apply to a configuration and does not have an explicitly specified meaning, but treat it like any other
+invalid value (for example, a value specifying a buffer size must naturally be non-negative.
+Treating a negative value as "buffer everything" would be an example of such a discouraged custom interpretation.
+Instead the default buffer size should be used.)
+Note that this could make a difference even if the custom interpretation is identical with the default value,
+because it might reset a value set from other configuration sources with the default.
+(Note: This paragraph was added after stabilization and the requirements are
+thus qualified as "SHOULD" to allow SDKs to avoid breaking changes. For new
+implementations, these should be treated as MUST-requirements.)
```
arminru, line 43:
```suggestion
> The following paragraph was added after stabilization and the requirements are
thus qualified as "SHOULD" to allow SDKs to avoid breaking changes.
For new
implementations, these should be treated as MUST requirements.

For variables accepting a numeric value, if the user provides a value the SDK cannot parse,
or which is outside the valid range for the configuration item, the SDK SHOULD
generate a warning and gracefully ignore the setting, i.e., treat them as not set.
In particular, SDKs
SHOULD NOT assign a custom interpretation e.g. to negative values if a negative
value does not naturally apply to a configuration and does not have an explicitly specified meaning, but treat it like any other
invalid value.

For example, a value specifying a buffer size must naturally be non-negative.
Treating a negative value as "buffer everything" would be an example of such a discouraged custom interpretation.
Instead the default buffer size should be used.

Note that this could make a difference even if the custom interpretation is identical with the default value,
because it might reset a value set from other configuration sources with the default.
```

Just some formatting + typo nits.

Before:

For variables accepting an numeric value, if the user provides a value the SDK cannot parse, or which is outside the valid range for the configuration item, the SDK SHOULD generate a warning and gracefully ignore the setting, i.e., treat them as not set. In particular, SDKs SHOULD NOT assign a custom interpretation e.g. to negative values if a negative value does not naturally apply to a configuration and does not have an explicitly specified meaning, but treat it like any other invalid value (for example, a value specifying a buffer size must naturally be non-negative. Treating a negative value as "buffer everything" would be an example of such a discouraged custom interpretation. Instead the default buffer size should be used.) Note that this could make a difference even if the custom interpretation is identical with the default value, because it might reset a value set from other configuration sources with the default. (Note: This paragraph was added after stabilization and the requirements are thus qualified as "SHOULD" to allow SDKs to avoid breaking changes. For new implementations, these should be treated as MUST-requirements.)

-----

After:

> **Note**: The following paragraph was added after stabilization and the requirements are thus qualified as "SHOULD" to allow SDKs to avoid breaking changes. For new implementations, these should be treated as MUST requirements.

For variables accepting a numeric value, if the user provides a value the SDK cannot parse, or which is outside the valid range for the configuration item, the SDK SHOULD generate a warning and gracefully ignore the setting, i.e., treat them as not set. In particular, SDKs SHOULD NOT assign a custom interpretation e.g. to negative values if a negative value does not naturally apply to a configuration and does not have an explicitly specified meaning, but treat it like any other invalid value.

For example, a value specifying a buffer size must naturally be non-negative. Treating a negative value as "buffer everything" would be an example of such a discouraged custom interpretation. Instead the default buffer size should be used.

Note that this could make a difference even if the custom interpretation is identical with the default value, because it might reset a value set from other configuration sources with the default.

arminru (approved):

carlosalberto (approved):

carlosalberto:
Overall good - I'm _really_ hoping nobody uses -1 and similar values as "unlimited" or similar ;)

reyang (approved):

PR #2968, opened 2022-11-18
Clarify specification compliance

reyang:
Follow up https://github.com/open-telemetry/opentelemetry-go/issues/3399#issuecomment-1304056426 and https://github.com/open-telemetry/opentelemetry-go/issues/3399#issuecomment-1319009509.

Notes: Given the broad impact, we will leave this PR open for Dec. 2022 and Jan. 2023 to collect more perspectives.


jack-berg (approved):

jsuereth (approved):

specification/README.md
```diff
@@ -50,12 +50,18 @@ The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
-the [specification][].
+An implementation of the [specification][] is compliant if and only if it meets
+all the following conditions:
+
+* It MUST satisfy all the "MUST", "MUST NOT", "REQUIRED", "SHALL", and "SHALL
+NOT" requirements defined in the [specification][].
+* It MUST NOT provide additional features or functionalities that are not
+  covered by the [specification][], except for any of the following situations:
+  * Such feature/functionality is very specific to a programming language, runtime
+    or framework.
```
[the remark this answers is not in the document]
carlosalberto: I'd even be more specific and mention which version from (1.16) - I think we mentioned something like this in the past for other features?

carlosalberto:
Also, let's mention this in the next Maintainers & Spec calls, so everybody is aware of this ;)

carlosalberto (approved):

specification/README.md
```diff
@@ -50,12 +50,20 @@ The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
+
+* It MUST satisfy all the "MUST", "MUST NOT", "REQUIRED", "SHALL", and "SHALL
+NOT" requirements defined in the [specification][].
+* It MUST NOT provide additional features or functionalities that are not
+  covered by the [specification][], except for any of the following situations:
+  * Such feature/functionality is very specific to a programming language,
+    runtime or framework, and not expected to interfere or overlap with any
+    feature that might (likely) be added to the cross-language specification in
+    future.
+  * Anything that is retrospecitive, for example, a feature/functionality which
```
ahayworth, line 64:
```suggestion
  * Anything that is retrospective, for example, a feature/functionality which
```

specification/README.md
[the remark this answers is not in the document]
reyang: Thanks for the suggestions! I've made the updates.

ahayworth (commented):
This feels too broad to me - I would even go so far to say it's an overreaction.

I would feel much more comfortable if the specification instead carved out an exception for "experimental" work, so that we can encourage language SIGs to try new things to bring back to the broader community. Such experimental work should be clearly delineated as not part of the official SDK and subject to change / breakage at any point. We can put very stern warnings on such "experimental" extensions.

The language here is _very_ broad, and I'm worried that it'll have a significant chilling effect on experimentation by language groups. Oftentimes it's useful to have a working implementation of things proposed to spec groups, and we should not discount the useful benefits of having it easily installable alongside official SDKs. With sufficient warnings and segmentation, this can be safe and effective.

The proposal here will create more problems than it solves, because any SDK that attempts to still do experimental work now faces a very fuzzy test for compliance. I don't foresee this providing a lot of clarity to the group as a whole. I believe we should _not_ adopt this, and instead the Go SDK should simply not do what they were proposing to do. Does it really need to go farther?

dyladan:
I agree with ahayworth. By the current wording it is actually disallowed to do prototypes of unspecified features, experimental semantic conventions, etc. In the meeting before the holiday we discussed that the API may be very strict but the SDK should be allowed to do some experimentation.

This also disallows the types of "convenience API wrappers" which I know @tedsuo has been pushing for for a long time, but others also find quite valuable.

dyladan (changes requested):
Requesting changes to indicate that I am not happy with the existing wording. I am not a spec approver so this is non-binding, but I am a maintainer. For details see my above comment.

specification/README.md
```diff
@@ -50,12 +50,20 @@ The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
-NOT" requirements defined in the [specification][]. Conversely, an
-implementation of the [specification][] is compliant if it satisfies all the
-"MUST", "MUST NOT", "REQUIRED", "SHALL", and "SHALL NOT" requirements defined in
-the [specification][].
+An implementation of the [specification][] is compliant if and only if it meets
+all the following conditions:
+
+* It MUST satisfy all the "MUST", "MUST NOT", "REQUIRED", "SHALL", and "SHALL
+NOT" requirements defined in the [specification][].
+* It MUST NOT provide additional features or functionalities that are not
```
tedsuo, line 58: I recommend that this section only apply to the instrumentation API.

ahayworth: > instrumentation API
By this you mean "the API" vs "the SDK", yes? "Instrumentation API" is a new term to me, and I want to make sure I understand what you mean by it. 😄

specification/README.md
```diff
@@ -50,12 +50,20 @@ The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
+  covered by the [specification][], except for any of the following situations:
+  * Such feature/functionality is very specific to a programming language,
+    runtime or framework, and not expected to interfere or overlap with any
+    feature that might (likely) be added to the cross-language specification in
+    future.
```
cartermp, line 63: It'd be good to understand what's meant by "very specific to a programming language, runtime or framework".

Two examples that comes to mind is how Java and Python both support annotations/decorators to make span creation nicer and how .NET and Python can tie ending a span to some notion of a scope with language constructs. Is each just "very specific to [the] programming language", or are these patterns to discuss and potentially add to the spec, since they're features available to multiple languages?

tsloughter: Personally I'd like to see those things proposed to the spec but in some new "convenience" (there was/is already a SIG and `#otel-convenience` channel on Slack) spec. Because not every language has decorators but those that do will all run into the same design decisions and it can help both users and developers.

trask: > Java and Python both support annotations/decorators to make span creation nicer

in Java at least, this functionality is not part of the OpenTelemetry API or SDK artifacts (instead they are provided by extensions), so my understanding is that this PR doesn't apply to them

MikeGoldsmith:
I also think that the wording is too limiting on experimentation and developing of features outside of the spec. There are many, many things that could potentially be added to the spec in the distant future so saying they cannot be added until the spec acknowledges them could be years apart. 

As others have said, there should be strict guidelines on how the API looks and what it does, but SDK features for convenience should be allowed. Otherwise we're going to have bare SDKs and promote everything else be some sort of extension which I think will lead to community confusion and frustration.

specification/README.md
cartermp: A concern I have with this is that as we grow and get less innovators/early-adopters and more "normal" developers who are used to nice abstractions atop building blocks, it will be a must-have for opentelemetry to be successful here. If the answer is to pull in a bunch of extensions for each language, then that makes "extensions" a _de facto_ API in practice for a large developer population. I don't think that would be a good outcome.

jmacd (approved):

specification/README.md
reyang: > It'd be good to understand what's meant by "very specific to a programming language, runtime or framework".
> 
> Two examples that comes to mind is how Java and Python both support annotations/decorators to make span creation nicer and how .NET and Python can tie ending a span to some notion of a scope with language constructs. Is each just "very specific to [the] programming language", or are these patterns to discuss and potentially add to the spec, since they're features available to multiple languages?

My suggestion is to not over specify things, instead we trust the judgement from maintainers. (Consider an example where a particular language has such specific feature at certain point, then later the feature got replicated/stolen by other languages - I don't think we want to cover all these cases in the spec)

cartermp: I think I'm comfortable with the current wording.

reyang:
@ahayworth @arminru @cartermp @dyladan @tedsuo @trask @tsloughter the PR has been updated to only enforce API (not SDK or anything else), PTAL.

cartermp (approved):
Looks good to me.

specification/README.md
```diff
@@ -50,12 +50,22 @@ The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
+  does not apply to an SDK implementation), it MUST NOT provide additional
+  features or functionalities that are not covered by the [specification][],
+  except for any of the following situations:
+  * Such feature/functionality is very specific to a programming language,
+    runtime or framework, and not expected to interfere or overlap with any
+    feature that might (likely) be added to the cross-language specification in
+    future.
+  * Anything that is retrospective, for example, a feature/functionality which
+    has been released as a stable/GA offering before this compliance rule was
+    formalized (release 1.16.0).
```
arminru, line 68:
```suggestion
    formalized (release 1.19.0).
```

arminru (approved):

trask (approved):

reyang:
@dyladan you've blocked the PR, could you take a look at the updated version and see if that removes the concern you had?

@tedsuo @tsloughter does the PR look good to you after the updates?

dyladan:
I think this is an improvement but still doesn't allow for convenience APIs. Is that intentional? One example that is already available in several languages is `with_span`.

To be clear, I don't think that's a blocker but I want to make sure it was the intended outcome. In order to provide these features in the future I guess a "sugared API" module would need to be released separately that is specifically not a compliant API implementation?

reyang:
> I think this is an improvement but still doesn't allow for convenience APIs. Is that intentional? One example that is already available in several languages is `with_span`.

My thinking is that the convenience APIs might be scenario based and probably fit for separate components that users can optionally choose to use. If a particular programming language has some established pattern across majority of the scenarios, then it is already a "language specific" feature to me. I can give one example which is the "Dependency Injection" in .NET - there are folks who like the DI pattern and there are folks who do not, for ASP.NET Core applications it seems DI is the recommended path.

Regarding `with_span`, I actually consider this as "language specific", I treat "language specific" as the opposite of "language agnostic". `with_span` is useful for a subset of programming languages which leverage static scope or some syntactic sugar to make the code leaner.

dyladan:
> > I think this is an improvement but still doesn't allow for convenience APIs. Is that intentional? One example that is already available in several languages is `with_span`.
> 
> 
> 
> My thinking is that the convenience APIs might be scenario based and probably fit for separate components that users can optionally choose to use. If a particular programming language has some established pattern across majority of the scenarios, then it is already a "language specific" feature to me. I can give one example which is the "Dependency Injection" in .NET - there are folks who like the DI pattern and there are folks who do not, for ASP.NET Core applications it seems DI is the recommended path.
> 
> 
> 
> Regarding `with_span`, I actually consider this as "language specific", I treat "language specific" as the opposite of "language agnostic". `with_span` is useful for a subset of programming languages which leverage static scope or some syntactic sugar to make the code leaner.

IMO that is a dangerously broad definition of language specific and renders the rest of the text essentially meaningless.

dyladan:
> > > I think this is an improvement but still doesn't allow for convenience APIs. Is that intentional? One example that is already available in several languages is `with_span`.
> > 
> > 
> > My thinking is that the convenience APIs might be scenario based and probably fit for separate components that users can optionally choose to use. If a particular programming language has some established pattern across majority of the scenarios, then it is already a "language specific" feature to me. I can give one example which is the "Dependency Injection" in .NET - there are folks who like the DI pattern and there are folks who do not, for ASP.NET Core applications it seems DI is the recommended path.
> > Regarding `with_span`, I actually consider this as "language specific", I treat "language specific" as the opposite of "language agnostic". `with_span` is useful for a subset of programming languages which leverage static scope or some syntactic sugar to make the code leaner.
> 
> IMO that is a dangerously broad definition of language specific and renders the rest of the text essentially meaningless.

@reyang do you not agree?

reyang:
> @reyang do you not agree?

I disagree with the "dangerously broad definition" part, I personally feel it is _okay_ as an intentional ambiguity.
I agree that people could have different interpretations of what "language specific" means, I don't think we're in position to formally define what is "language specific".

dyladan:
If we're going to allow APIs to add helper methods specific to their langauge we should state that specifically. Also, what happens if the otel API adds a method with the same name in the future?

PR #2969, opened 2022-11-21
Define data ownership for LogRecordProcessor

### LINKED ISSUE #2955: Resolve lingering question of LogRecord mutability in LogRecordProcessor
#2681 made it possible for `LogRecordProcessor` to mutate LogRecords in their `onEmit` implementations. Mutating LogRecords is critical for use cases like enriching with baggage, redacting sensitive information, and more. 

After merging, a [conversation](https://github.com/open-telemetry/opentelemetry-specification/pull/2681#discussion_r931463244) took place regarding potential performance impact of allowing mutation due to locking requirements.

**Option 1: Do Nothing**

All `LogRecordProcessor`s are allowed to mutate LogRecords. SDKs have to deal with this by coming up with appropriate locking strategies. 

It may be possible to still have lock-free high performance if none of the registered `LogRecordProcessors` perform mutations. Consider a strategy where mutations are tracked using a lock free CAS (compare and set) counter like java's `AtomicInteger`. If no mutation ever occurs, then `BatchLogProcessor` can transform to an immutable [ReadableLogRecord](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md#readablelogrecord) version needed to call `LogRecordExporter#export` without taking a lock. If mutations do occur, a lock would have to be obtained.

**Option 2: Force `LogRecordProcessor` implementations to indicate whether they mutate**

For example, by adding method to `LogRecordProcessor` called `boolean isMutating()`. 

This makes it somewhat easier to avoid locks in configurations where no `LogRecordProcessors` perform mutations. 

If an implementation trying to mutate the `LogRecord` when `isMutating() == false`, reject the mutation and log a warning.

As noted [here](https://github.com/open-telemetry/opentelemetry-specification/pull/2681#discussion_r955102502), this is the strategy used by collector processors.

**Option 3: Split out start and end phases of emitting a LogRecord**

[SpanProcessor](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#span-processor) is split into two phases: 1. `onStart` runs when the span is starting, and has the ability to mutate data. 2. `onEnd` runs when the span has ended and _DOES NOT_ have the ability to mutate the data.

The Log API could similarly be split into multiple phases of starging and log record and emitting it. `LogRecordProcessor` could then be split into multiple phases as well, with `onStart` allowed to mutate while the `LogRecord` sent to `onEnd` is immutable.

This produces awkward ergonomics for the API since people don't think of emitting a log in multiple phases. Additionally, we'd have to answer questions like what types of data can / can not be set in the start phase, and what data _must_ be set in the start phase.

Not sure of any advantages of this approach above Option 2. 

Let's talk about this and come to some conclusion so we can move forward with confidence.

jack-berg:
Resolves #2955.

States that `LogRecordProcessor#onEmit` can modify `logRecord` until it returns. If a reference to `logRecord` is held after `onEmit` returns, only reads are permitted.

If a processor wants to do asynchronous modification it needs to first make a copy is responsible for calling any downstream processors / exporters that expect to see the changes. Described more [here](https://github.com/open-telemetry/opentelemetry-specification/pull/2969#discussion_r1035284423).


specification/logs/sdk.md
```diff
@@ -186,6 +186,15 @@ therefore it SHOULD NOT block or throw exceptions.
 
 **Returns:** `Void`
 
+**Data Ownership**:
+
+Implementations own `logRecord` from the time `OnEmit` is called until the
```
reyang, line 191: What does "Implementations" mean? Is it "implementations of LogRecordProcessor"?

jack-berg: Yes. Can change "Implementations" to "LogRecordProcessors" if preferable.

specification/logs/sdk.md
```diff
@@ -186,6 +186,15 @@ therefore it SHOULD NOT block or throw exceptions.
...
+implementation returns. Implementations are allowed to freely modify the data
+while they own it. If an implementation wishes to modify the data but does not
```
reyang, line 193: I guess the intention is to give LogRecordProcessors rights to modify the logRecord data while its OnEmit is being called? If that's the only reason, maybe the wording "ownership" can be avoided, e.g. "A LogRecordProcessor has the rights to make modifications from the time `OnEmit` is called until the time `OnEmit` returns".

The reason we might want to avoid touching "ownership":
1. After the LogRecordProcessor.OnEmit returned, who owns the logReord now?
2. What does "own" actually mean? (e.g. responsible for invoking the destructor/Dispose, cleaning up the memory, maintaining the ref count, etc.)
3. What if the LogRecordProcessor.OnEmit invoked another processor's OnEmit? Do they co-own the record now, or the ownership is transferred to the underlying processor?

specification/logs/sdk.md
```diff
@@ -186,6 +186,15 @@ therefore it SHOULD NOT block or throw exceptions.
...
+want to incur the cost synchronously, it can use various techniques to make the
+change asynchronously. For example, by asynchronously copying `logRecord` before
+modifying it.
```
tigrannajaryan, line 196: I don't understand how this works. If a processor makes a copy, then returns the control to the caller, what difference does it make if it asynchronously modifies the copy? Isn't the caller going to use the original anyway?

jack-berg: The idea is if a processor wants to make asynchronous modifications they won't be seen by other processors or the caller.

tigrannajaryan: What's the purpose of doing such modifications if noone else is going to see them?

jack-berg: Perhaps the processor has internal logic that calls other processors (or an exporter) after the modification is made.

The tradeoff for asynchronous modification is that the processor has to manage its own control flow.

tsloughter: Right, the modified version is what gets emitted, so it is seen, just not by the other processors the caller may be making calls to.

tigrannajaryan: If the processor calls another processor (or exporter) that's chaining, isn't it? I still fail to understand what is it that we are stating. Are we saying use chaining model if you need to modify the data asynchronously?

If we are not ready to say this explicitly in the spec then I would argue that we should not be talking about async processing at all since it raises more questions than it answers (it leaves me wondering what happens with the modified copy of the data).

jack-berg: > If the processor calls another processor (or exporter) that's chaining, isn't it? I still fail to understand what is it that we are stating. Are we saying use chaining model if you need to modify the data asynchronously?

Yes. A processor can do whatever it wants internally, including chaining. But the SDK won't give do that for you, or give you tools for it (at least not for the moment).

> If we are not ready to say this explicitly in the spec then I would argue that we should not be talking about async processing at all since it raises more questions than it answers (it leaves me wondering what happens with the modified copy of the data).

I think the spec owes it to the users to explain how async mutation can be accomplished, even if we don't provide explicit tools to make it easy at the moment.

I've created a couple of diagrams to illustrate the difference:

**Synchronous Mutation**

This illustrates an SDK configured with a couple of processors (Processor A and Processor B) which synchronously mutate the emitted LogRecord before its passed to the Batch Processor and queued for export. The mutations of the processors are visible to the processors that follow. The SDK calls each processor synchronously on the same thread the log was emitted on in the order the processors were registered.

sync_mutation

**Asynchronous Mutation**

This illustrates an SDK configured to do async mutation. Processors A and D are each called synchronously on the same thread the log was emitted. Processor A queues up a copy of the LogRecord, and asynchronously calls Processor B, C, and finally the Batch Processor. Processor B and processor C mutate the LogRecord, but their changes are not visible to Processor D since they were made to a copy of the LogRecord. The chaining that Processor A performs in calling processors B and C is an implementation detail of Processor A, and not provided by any constructs in the SDK.

async_mutation

tigrannajaryan: @jack-berg the pictures you have are very clear, thank you.

IMO, it would be better in the text of this document to done one of these:
- Rephrase to something like "processors own the LogRecord only for the durtion of the OnEmit call. Processors cannot hold a reference to the LogRecord and modify it after OnEmit call returns".
- Alternatively, explain fully what async processing means and include the diagrams that you have in this comment thread.

jack-berg: I went with something similar to your first suggestion. I replaced the "cannot hold a reference" phrase with stating that the `LogRecord` is read only after `OnEmit` returns. This was motivated by the java implementation of [BatchLogRecordProcessor](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/logs/src/main/java/io/opentelemetry/sdk/logs/export/BatchLogRecordProcessor.java#L89), which adds the reference to the queue to avoid doing extra work on the calling thread by making a copy. This design is consistent with `BatchSpanProcessor` as well.

If users open report confusion with how to make async modifications, I can add a clarifying footnote which includes these images.

specification/logs/sdk.md
jack-berg: I like this recommendation and have removed "ownership" from the phrasing.

tigrannajaryan (approved):

specification/logs/sdk.md
tigrannajaryan: I like the new wording.

jack-berg:
@open-telemetry/specs-logs-approvers please take a look.

reyang (approved):

PR #2978, opened 2022-11-22
List the machine-id as potential source for a unique host.id

svrnm:
## Changes

This is just a minor change to the description of `host.id`, that suggests the use of the `machine-id` available with some linux systems (if systemd or dbus are used). 

The example value is also updated with an example ID instead of `opentelemetry-test`

For reference: 
* https://www.freedesktop.org/software/systemd/man/machine-id.html
* https://wiki.debian.org/MachineId


semantic_conventions/resource/host.yaml
```diff
@@ -9,7 +9,8 @@ groups:
         type: string
         brief: >
           Unique host ID. For Cloud, this must be the instance_id assigned by the cloud provider.
-        examples: ['opentelemetry-test']
+          For linux systems, the `machine-id` located in `/etc/machine-id` or `/var/lib/dbus/machine-id` may be used.
```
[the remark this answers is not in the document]
svrnm: will fix the nit.

> Would this be the recommendation for Linux containers?

This is for VMs/bare metal or other "non-containers".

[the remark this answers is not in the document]
reyang: > > Would this be the recommendation for Linux containers?
> 
> This is for VMs/bare metal or other "non-containers".

Maybe worth calling it out explicitly.

reyang (approved):

semantic_conventions/resource/host.yaml
```diff
@@ -9,7 +9,8 @@ groups:
         type: string
         brief: >
           Unique host ID. For Cloud, this must be the instance_id assigned by the cloud provider.
-        examples: ['opentelemetry-test']
+          For non-containerised Linux systems, the `machine-id` located in `/etc/machine-id` or `/var/lib/dbus/machine-id` may be used.
```
arminru, line 12:
```suggestion
          For non-containerized Linux systems, the `machine-id` located in `/etc/machine-id` or `/var/lib/dbus/machine-id` may be used.
```
I think we mostly use American English throughout the spec, so this would keep it consistent.

specification/resource/semantic_conventions/host.md
```diff
@@ -9,7 +9,7 @@
 <!-- semconv host -->
 | Attribute  | Type | Description  | Examples  | Requirement Level |
 |---|---|---|---|---|
-| `host.id` | string | Unique host ID. For Cloud, this must be the instance_id assigned by the cloud provider. | `opentelemetry-test` | Recommended |
+| `host.id` | string | Unique host ID. For Cloud, this must be the instance_id assigned by the cloud provider. For non-containerised Linux systems, the `machine-id` located in `/etc/machine-id` or `/var/lib/dbus/machine-id` may be used. | `fdbf79e8af94cb7f9e8df36789187052` | Recommended |
```
arminru, line 12:
```suggestion
| `host.id` | string | Unique host ID. For Cloud, this must be the instance_id assigned by the cloud provider. For non-containerized Linux systems, the `machine-id` located in `/etc/machine-id` or `/var/lib/dbus/machine-id` may be used. | `fdbf79e8af94cb7f9e8df36789187052` | Recommended |
```

arminru (approved):

semantic_conventions/resource/host.yaml
svrnm: Thx

PR #2982, opened 2022-11-23
Add `messaging.kafka.message.offset` attribute

### LINKED ISSUE #2971: Should Kafka offset be used as `messaging.message.id`?

Currently the Go Kafka instrumentations sets the record offset as the value of the `messaging.message_id` attribute: 

```go
func finishProducerSpan(span trace.Span, partition int32, offset int64, err error) {
	span.SetAttributes(
		semconv.MessagingMessageIDKey.String(strconv.FormatInt(offset, 10)),
		semconv.MessagingKafkaPartitionKey.Int64(int64(partition)),
	)
```

I was requested to introduce a similar change to the Java instrumentation, but I'm not sure whether this is valid: an offset is only unique in a partition (which has a dedicated `messaging.kafka.message.partition` attribute), so I don't think that it is 

> A value used by the messaging system as an identifier for the message, represented as a string

as the spec describes this attribute; because you also need the partition id to identify the message, so it's half of an identifier really.

Should the Kafka offset be set as the value of the `messaging.message.id` attribute? Or, alternatively, should we introduce a `messaging.kafka.message.offset` attribute?

**Additional context.**

* https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/github.com/Shopify/sarama/otelsarama/producer.go#L308-L312

mateuszrzeszutek:
Fixes #2971

## Changes

Introduces a new `messaging.kafka.message.offset` attribute that is supposed to store the offset of a Kafka message. I used the naming pattern introduced in #2957, assuming it's probably going to land in the same release as this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ release.
 
 ### Traces
 
+- Add `messaging.kafka.message.offset` attribute.
+  ([#2982](https://github.com/open-telemetry/opentelemetry-specification/pull/2982))
+
 ### Metrics
 
 - Define Experimental MetricProducer as a third-party provider of metric data to MetricReaders.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,6 @@ release.
 
 ### Traces
 
-- Add `messaging.kafka.message.offset` attribute.
-  ([#2982](https://github.com/open-telemetry/opentelemetry-specification/pull/2982))
-
 ### Metrics
 
 - Define Experimental MetricProducer as a third-party provider of metric data to MetricReaders.
@@ -36,6 +33,8 @@ release.
 - Rename `rpc.request.metadata.<key>` and `rpc.response.metadata.<key>` to
   `rpc.grpc.request.metadata.<key>` and `rpc.grpc.response.metadata.<key>`
   ([#2981](https://github.com/open-telemetry/opentelemetry-specification/pull/2981))
+- Add `messaging.kafka.message.offset` attribute.
+  ([#2982](https://github.com/open-telemetry/opentelemetry-specification/pull/2982))
 
 ### Compatibility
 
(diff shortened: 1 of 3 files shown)


semantic_conventions/trace/messaging.yaml
```diff
@@ -189,6 +189,11 @@ groups:
         brief: >
           Partition the message is sent to.
         examples: 2
+      - id: message.offset
```
kenfinnigan, line 192: Should it be `message_offset` to align with other naming of attributes?

pyohannes (approved):
This was discussed in the messaging workgroup (see summary in the related issue).

Having a dedicated offset attribute is preferred to using the offset as message id.

semantic_conventions/trace/messaging.yaml
joaopgrassi, line 192: I guess we are OK in using the "new" namespace here, even though it's only going to be introduced once @lmolkova  PR's is merged https://github.com/open-telemetry/opentelemetry-specification/pull/2957/files?

mateuszrzeszutek: I took this approach to reduce thrash around changing attribute names, assuming that PR is gonna get merged anyway. I don't mind changing it to `messaging.kafka.offset` for the time being, if there's strong conviction to do so.

joaopgrassi (approved):

arminru (approved):

breedx-splk (approved):

carlosalberto (approved):

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ release.
 
 ### Traces
 
+- Add `messaging.kafka.message.offset` attribute.
```
carlosalberto, line 14: This should go into the Semantic Conventions instead btw ;)

mateuszrzeszutek: Oops! fixed, thanks!

PR #2985, opened 2022-11-24
[resource] add openshift platform to all supported cloud providers

### LINKED ISSUE #2970: add openshift platform to all supported cloud providers

Identifying the Openshift platform, which is operated on different cloud providers.

Members of `cloud.platform` like `aws_openshift`, `gcp_openshift`, ... . 

**Additional context.**

frzifus:
Signed-off-by: Benedikt Bongartz <bongartz@klimlive.de>

Fixes #2970

## Additional information

https://aws.amazon.com/de/rosa/
https://www.alibabacloud.com/solutions/redhat-openshift
https://azure.microsoft.com/de-de/products/openshift/
https://console.cloud.google.com/marketplace/browse?q=red%20hat%20openshift&pli=1


CHANGELOG.md
```diff
@@ -23,6 +23,8 @@ release.
   ([#2940](https://github.com/open-telemetry/opentelemetry-specification/pull/2940))
 - Split out Event API from Log API
   ([#2941](https://github.com/open-telemetry/opentelemetry-specification/pull/2941))
+- Extend Cloud Platform Enum with OpenShift entry for all supported cloud providers.
```
joaopgrassi, line 26: Could you please move this entry under the `### Resource` section?

semantic_conventions/resource/cloud.yaml
```diff
@@ -99,6 +105,9 @@ groups:
             - id: azure_app_service
               value: 'azure_app_service'
               brief: Azure App Service
+            - id: azure_openshift
+              value: 'azure_openshift'
+              brief: Red Hat OpenShift on Azure Cloud
```
joaopgrassi, line 110:
```suggestion
              brief: Azure Red Hat OpenShift
```

I'm not sure what we should put on `brief` but it looks like the "product name" to me. I saw for Alibaba you used the exact same name as state on their website, but for the others is a bit different. Should we stay with the "advertised" service/product name on the respective pages? If so, then the others also need to be adjusted.

semantic_conventions/resource/cloud.yaml
```diff
@@ -114,6 +123,9 @@ groups:
             - id: gcp_app_engine
               value: 'gcp_app_engine'
               brief: Google Cloud App Engine (GAE)
+            - id: google_cloud_openshift
+              value: 'google_cloud_openshift'
+              brief: Red Hat OpenShift on Google Cloud
```
joaopgrassi, line 128: I'm not familiar with this, but I noticed on GCP they have multiple "OpenShift options" like k8s engine, container platform, so I was wondering: With only this info here is it clear which one is being used?

joaopgrassi (changes requested):

CHANGELOG.md
frzifus: ups, that was a mistake

semantic_conventions/resource/cloud.yaml
frzifus: Ive here no strong opinion. Let me change that :+1:

semantic_conventions/resource/cloud.yaml
frzifus: This are just different machine images for the same service with more or less installed things. calling it openshift should be fine. The internal Openshift api also only knows gcp.
image

joaopgrassi: Thanks for explaining. Makes sense to me :)

semantic_conventions/resource/cloud.yaml
frzifus: i compared it with the others. Now it should fit :+1:

specification/resource/semantic_conventions/cloud.md
```diff
@@ -38,22 +38,26 @@
...
 | `aws_lambda` | AWS Lambda |
 | `aws_elastic_beanstalk` | AWS Elastic Beanstalk |
 | `aws_app_runner` | AWS App Runner |
+| `aws_openshift` | Red Hat OpenShift on AWS (ROSA) |
 | `azure_vm` | Azure Virtual Machines |
 | `azure_container_instances` | Azure Container Instances |
 | `azure_aks` | Azure Kubernetes Service |
 | `azure_functions` | Azure Functions |
 | `azure_app_service` | Azure App Service |
+| `azure_openshift` | Azure Cloud Red Hat OpenShift |
```
joaopgrassi, line 54: Please re-generate the tables, so the build passes :)

joaopgrassi (approved):
Looks good to the me!

Oberon00:
This seems to be a non-native "Marketplace" feature. I don't really know how these work, is it built on top of other public-facing products of the platform? Do we actually need entries in the enum for that? If we include marketplace offerings, it might make the size of this enumeration potentially endless.

frzifus:
>  I don't really know how these work, is it built on top of other public-facing products of the platform?

Exactly, for example in aws openshift uses ec2 instances as nodes.

> Do we actually need entries in the enum for that?

I would argue - yes. Although openshift uses ec2 instances, they are abstracted away. Which in my view makes openshift the platform. Nothing else happens with eks, which uses ec2 instances and its is also listed as a platform on aws. In both cases its more beneficial to know the underlying platform that potentially caused issues is eks or openshift then ec2.

> If we include marketplace offerings, it might make the size of this enumeration potentially endless.

This depends very much on the definition of a platform. My feeling is that there are not many different platform solutions in the marketplaces.

carlosalberto (approved):

carlosalberto:
Overall LGTM although @Oberon00 raised a good point regarding the potential explosion of values here. Let's try to get more eyes on this, just to stay on the safe side.

joaopgrassi:
One could argue that we already have such things, for ex k8s managed service in cloud providers (Azure AKS). I think adding OpenShift is pretty much the same as those, so if we don't want this then should we keep Azure AKS and the like?

frzifus:
Whats need to move on?

PR #2102, opened 2021-11-05
Define "none" enum value

pellared:
## What

Define "none" as enum value used to set a null object (empty, no-op value).

## Why

1. Having a possibility to "disable" something via enum values when the default settings are "enabling" some feature is critical
1. Enforce consistency

Addresses: https://github.com/open-telemetry/opentelemetry-specification/pull/2045#discussion_r742873983

Related PRs:
- https://github.com/open-telemetry/opentelemetry-specification/pull/2052
- https://github.com/open-telemetry/opentelemetry-specification/pull/1439

PR #3000, opened 2022-12-02
Define "none" numeric value

Kielek:
Fixes N/A

## Changes

Define "none" numeric value.

It will be useful to correctly set `OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT` to no limit. For now, it should use int.maxvalue or similar values.

Related issues for enum handling #2102

### DIFF
--- a/specification/sdk-environment-variables.md
+++ b/specification/sdk-environment-variables.md
@@ -47,6 +47,8 @@ Instead the default buffer size should be used.
 Note that this could make a difference even if the custom interpretation is identical with the default value,
 because it might reset a value set from other configuration sources with the default.
 
+If a null value is acceptable, the numeric value representing it MUST be `"none"`.
+
 ### Enum value
 
 For variables which accept a known value out of a set, i.e., an enum value, SDK implementations MAY support additional values not listed here.


pellared (approved):
👍

Oberon00 (changes requested):
First, I don't think this is specific to numeric values. Booleans, strings, enums, etc., might accept "null" values just as well.
Second, there is no concept of "null value" in this environment specification, so this specification, as written now, has no meaning (or an unclear meaning).
Your goal seems to allow explicitly setting "no limit" in the limit configuration. You should add a more specific spec to that subsection, instead of the generic section here.

(PS: Congrats on creating PR #3000 🥳)

reyang:
> It will be useful to correctly set OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT to no limit. For now, it should use int.maxvalue or similar values.

Why would someone want "no limit"? Related to https://github.com/open-telemetry/opentelemetry-specification/pull/2960#discussion_r1035360661.

reyang (changes requested):
Need to understand why "no limit" is needed.

pellared:
> Need to understand why "no limit" is needed.

I agree that "unbounded limits" is a bad practice from "reliability" standpoint. But I do not see anything in the spec that disallows limitless configurations. Current default for `OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT` and `OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT` is "no limit". Wouldn't changing it be a breaking change? 

If it is possible to have a unlimited configuration then we should allow setting it.

_I have no clear nor strong opinion 😆_

Kielek:
I am closing this issue per last .NET AutoInstrumentation Discussion with @carlosalberto.
For now, there will be no possibility to set no-limit options. It is considered as to unpredictable.

Instead we should consider to change default values for `OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT` and `OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT `. I do not think that is is good practice to have default value which cannot be set by appropriate env. value.

PR #3002, opened 2022-12-02
Define LogRecord batching processor defaults

### LINKED ISSUE #2954: Define defaults for the LogRecord batching processor
The default values for the parameters of the [LogRecord batching processor](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md#batching-processor) are marked TODO.

I propose to adopt the same defaults as the [Span batching processor](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#batching-processor).

maxQueueSize: 2048
scheduledDelayMillis: 5000
exportTimeoutMillis: 30000
maxExportBatchSize: 512

If folks feel defaults should differ between spans and logs, can you suggest any specific investigations that would lead us to a better informed set of defaults?

alanwest:
Fixes #2954

Defines default configuration values for the `LogRecord` batching processor.

The same default values have been adopted from the span batching processor with the exception of `scheduleDelayMillis`. A default value of 1000 milliseconds has been chosen for log rather than the 5000 milliseconds for spans. See https://github.com/open-telemetry/opentelemetry-specification/pull/3002#discussion_r1048520899 for discussion.


reyang (approved):

jack-berg (approved):

specification/logs/sdk.md
```diff
@@ -266,13 +266,13 @@ representations to the configured `LogRecordExporter`.
 
 * `exporter` - the exporter where the `LogRecords` are pushed.
 * `maxQueueSize` - the maximum queue size. After the size is reached logs are
-  dropped. The default value is TODO.
+  dropped. The default value is `2048`.
 * `scheduledDelayMillis` - the delay interval in milliseconds between two
-  consecutive exports. The default value is TODO.
+  consecutive exports. The default value is `5000`.
```
tigrannajaryan, line 271: I think adding 5 second latency to logs is very undesirable. Whatever mistake span/trace definition did, I don't think we are obligated to repeat it.
There are use cases called "live tail" which are close to unusable if you add 5 second latency right at the source, plus whatever latencies exist in the other legs - Collector, network, backend.
IMO, this is the case when we should ignore the consistency argument because it results in bad user experience. I am going to block this temporarily.

tigrannajaryan (changes requested):

specification/logs/sdk.md
alanwest: I mentioned the live-tail use case. Is there a generally accepted duration that is appropriate for live-tailing? Is it the most common use case? If it's not the most common it seems reasonable for a user to set their or defaults - or perhaps even use the simple processor (i.e., zero delay).

tigrannajaryan: So, let's say we have a typical setup with Otel SDK->localhost Collector->Intermediary Collector->Backend. Each of those 4 nodes will likely do some batching. If the defaults were 5 second on each node we are talking about 20 second latency just for batching. I think this is unacceptable for "live" tailing. I would ideally expect a sub-second end-to-end latency.

The purpose of batching is to avoid having thousands or millions of individual export requests with a single tiny log record in the payload, in situations where the source is emitting such log records at high rate (thousands per sec). Batching time of a few  hundred millisecond solves this issue. I don't see what benefits are there for going into the 5 second territory. Perhaps I am missing a user case where such a high delay is important.

If the desire for consistency with traces/metrics is strong I would suggest to file a bug to change the batching defaults there from 5 second to something more reasonable too. I think it can be considered a performance bug and we may be able to change it even though it is in a Stable doc (bug fixes are allowed).

alanwest: > If the desire for consistency with traces/metrics is strong I would suggest to file a bug to change the batching defaults there from 5 second to something more reasonable too.

I do not feel strongly they should be the same, just that we have a understanding of why they're different. I support proposing 5 seconds for traces as a bug based on your arguments here. I also support that they differ if we have wide acceptance of a shorter interval. I will raise this in the spec SIG.

As another data point, I see FluentBit defaults to a 5 second flush interval. https://docs.fluentbit.io/manual/v/1.3/configuration/file#config_section

tedsuo:
So, it's clear that our current defaults – both for the SDK and the Collector – are not great, and we should consider revisiting them. For example, some defaults presume that the SDK is talking to a local collector. Some defaults presume the collector is aggregating data from multiple services. This is due to us adding defaults one at a time, where each default might be reasonable on its own, but making different assumptions from the other defaults. 

My request is that the logging defaults be coherent with the tracing defaults. New users will definitely be confused if they start receiving logs immediately, but don't see their traces for five seconds. So I'm worried about creating a different set of defaults for logging, even if the assumptions behind those new defaults are reasonable on their won.

This is my suggested course of action:
* Please describe the "default deployment" we are actually trying to target.
* I _strongly_ suggest that by default, we assume an SDK is talking to a local collector. This is our suggested deployment pattern for production, and it also works well for new users going through the installation process.
* Pick a set of logging defaults that work well for this scenario, but are not too aggressive or over-optimized.
* Bring the tracing defaults into alignment with the new logging defaults.

If the above course of action is too much work for the logging group to tackle at this time, my request is to match the tracing defaults for now, and defer improving our defaults to a later date. Please do not release a version of the spec where tracing and logging defaults are incoherent with each other.

specification/logs/sdk.md
reyang: > Whatever mistake span/trace definition did, I don't think we are obligated to repeat it.

+1, similar discussion happened around metrics SDK spec, after collecting feedback from multiple vendors and the Prometheus group, [30 seconds were chosen](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#periodic-exporting-metricreader) which is different from the trace exporting behavior.

> There are use cases called "live tail" which are close to unusable if you add 5 second latency right at the source, plus whatever latencies exist in the other legs - Collector, network, backend. IMO, this is the case when we should ignore the consistency argument because it results in bad user experience. I am going to block this temporarily.

-1 on using one specific scenario to drive the decision, anything "live" or "interactive" is on one extreme, the other extreme would be the long latency case (e.g. an IoT device sending data to a satellite), and IMHO the default value should be somewhere between, rather than optimizing for one extreme.

tigrannajaryan:
> This is my suggested course of action:
> 
> * Please describe the "default deployment" we are actually trying to target.
> * I _strongly_ suggest that by default, we assume an SDK is talking to a local collector. This is our suggested deployment pattern for production, and it also works well for new users going through the installation process.
> * Pick a set of logging defaults that work well for this scenario, but are not too aggressive or over-optimized.
> * Bring the tracing defaults into alignment with the new logging defaults.
> 
> If the above course of action is too much work for the logging group to tackle at this time, my request is to match the tracing defaults for now, and defer improving our defaults to a later date. Please do not release a version of the spec where tracing and logging defaults are incoherent with each other.

I like this plan. I think we should go for lower batch delay for both logs and traces (somewhere around 200-1000ms) and use the same value for both. I am personally fine with lowering defaults for the traces and I do not think it is a breaking change, I think it should be allowed.

specification/logs/sdk.md
tigrannajaryan: The situation is not symmetric here and I do not think it is a matter of choosing an "average" value to please everyone.

By choosing a large value we make "live" use cases almost unusable. We make things qualitatively worse and simply make some interesting product features impossible or significantly less valuable for the users.

As opposed to that by choosing a small enough value we perhaps increase the rate of requests to the backend, which does not seem like a catastrophic requirement that is impossible to fullfill or which somehow makes some other interesting use case impossible. It is not a qualitative downside, it is merely a difference in quantity (what request rate the backend needs to handle).

IMO we should go for low enough value to enable "live" use cases and if a particular backend has a problem because of the rate of the requests then the user can change the default (typically can be more easily done in a central Collector that accepts data from many sources and sends to the backend).

alanwest:
> Please describe the "default deployment" we are actually trying to target.

This seems worthy of a separate issue. That is, the "default deployment" described should be more of a top-level thing in the spec making it clear that it provides a guide for the decisions we make across the entire specification.

Regarding @tedsuo's recommendation as it relates to the specific issue of settling on batch processor defaults, the question I have is the order of operations. For example, say we agree that 1000ms is a more satisfying default. Can we do things in this order?

1) Land this PR with a 1000ms default
2) Open a separate PR to change span batch processor also to 1000ms
3) Open a separate issue/PR to describe the "default deployment"

I'd prefer that we are able to continue progress on log stabilization in parallel with furthering the discussion of codifying a notion of a "default deployment".

specification/logs/sdk.md
alanwest: Summarizing some "feelings" heard from SIG meetings

* A default of 200ms is considered too low by enough people that I believe it unlikely we will ever agree to adopt it for traces or logs.
* A default value of 5000ms is considered too high by a some people but not necessarily enough people to declare a consensus that it is too high.
* A default value of 1000ms has been suggested as "good enough" for the specific live-tailing use case. It has also been argued that 1000ms is unlikely to overwhelm a backend wether it be near or far away.

Questions:

1) Does anyone have a strong objection to proposing 1000ms as the default for logs?
2) Is there a strong feeling, unlike for metrics, that logs and traces should be the same in order to make them coherent as suggested here https://github.com/open-telemetry/opentelemetry-specification/pull/3002#issuecomment-1377594018

tigrannajaryan: > Can we do things in this order?
> 
> 1. Land this PR with a 1000ms default
> 2. Open a separate PR to change span batch processor also to 1000ms
> 3. Open a separate issue/PR to describe the "default deployment"

@alanwest I like the plan.

alanwest: @tigrannajaryan I have changed the default to 1000ms.

We still have an opportunity to circle back to this decision prior to marking the log SDK stable. For now, I think merging this helps bring us a step closer to our discussion with the community to declare stabilization.

reyang (approved):

tigrannajaryan (approved):

PR #2997, opened 2022-11-30
Remove markdown link to jsonlines.org

breedx-splk:
I noticed a linter/link check failing on a PR due to jsonlines.org domain not resolving.

PR #3010, opened 2022-12-05
Revert #2997 (add jsonlines link back)

breedx-splk:
#2997 removed the broken markdown link to jsonlines.org. The jsonlines.org site has been moved to Cloudflare and should be more available now, so let's put the link back.


arminru (approved):

carlosalberto (approved):

breedx-splk:
Works on my box. Is it caching?

breedx-splk:
> @breedx-splk I suggest we don't apply this change - in some runs it works, and in some others it doesn't (unless we can verify & discard caching?)

Yeah, super weird. I'm totally fine leaving it out for now. 👍🏻

PR #3014, opened 2022-12-06
Propose new StructuredBody field for logs

djaglowski:
## Motivation

The log data model has already been declared stable, yet an important decision made by the Logs SIG has not been codified in the specification. Specifically, it was the intention of the Logs SIG that `Attributes` should be the appropriate field for representing structured log data. Several proposals to codify this notion in the spec have stalled out.

This proposal attempts to identify an alternative that would provide an explicit home for structured data, without breaking the data model.

Obviously this change may have implications for the SDK, Collector, etc, but I am suggesting that we fully explore this route in case it leads to a broadly acceptable solution.

## Changes

The proposal is to add an optional new field, tentatively called `StructuredBody`. This field would be dedicated to structured log data.

`Attributes` would still be intended for information _about_ the log (e.g. user specified values, or [semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/semantic_conventions/README.md))

`Body` would remain unchanged as well. Notably the indication that "First-party Applications SHOULD use a string message." would still be valid.

`StructuredBody` would also be an alternative to the "data" field proposed in https://github.com/open-telemetry/opentelemetry-specification/pull/2926.

### Related issues
- https://github.com/open-telemetry/opentelemetry-specification/pull/2096
- https://github.com/open-telemetry/opentelemetry-specification/issues/1613
- https://github.com/open-telemetry/opentelemetry-specification/pull/2888
- https://github.com/open-telemetry/opentelemetry-specification/pull/2861#discussion_r993531085
- https://github.com/open-telemetry/opentelemetry-specification/pull/2620
- https://github.com/open-telemetry/opentelemetry-specification/pull/2926
- https://github.com/open-telemetry/opentelemetry-specification/pull/1727
- https://github.com/open-telemetry/opentelemetry-specification/pull/1777
- https://github.com/open-telemetry/opentelemetry-specification/pull/2841
- https://github.com/open-telemetry/opentelemetry-specification/pull/596
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/16495

### Related [OTEP(s)](https://github.com/open-telemetry/oteps) #
- https://github.com/open-telemetry/oteps/pull/97
- https://github.com/open-telemetry/oteps/pull/188

### DIFF
--- a/specification/logs/data-model.md
+++ b/specification/logs/data-model.md
@@ -464,6 +464,15 @@ OpenTelemetry [semantic conventions for Log Attributes](./semantic_conventions/R
 [semantic conventions for Span Attributes](../trace/semantic_conventions/README.md).
 This field is optional.
 
+### Field: `StructuredBody`
+
+Type: `map<string, any>`.
+
+Description: A structured representation of the log record (see the description
+of `any` type above). Can vary for each occurrence of the event coming from the
+same source. This field is optional. SHOULD follow OpenTelemetry [semantic conventions
+for Log Bodies (TODO)](./semantic_conventions/README.md). This field is optional.
+
 #### Errors and Exceptions
 
 Additional information about errors and/or exceptions that are associated with
@@ -946,7 +955,7 @@ Field            | Type               | Description                             
 timestamp        | string             | The time the event described by the log entry occurred. | Timestamp
 resource         | MonitoredResource  | The monitored resource that produced this log entry.    | Resource
 log_name         | string             | The URL-encoded LOG_ID suffix of the log_name field identifies which log stream this entry belongs to. | Attributes["gcp.log_name"]
-json_payload     | google.protobuf.Struct | The log entry payload, represented as a structure that is expressed as a JSON object. | Body
+json_payload     | google.protobuf.Struct | The log entry payload, represented as a structure that is expressed as a JSON object. | StructuredBody
 proto_payload    | google.protobuf.Any | The log entry payload, represented as a protocol buffer. | Body
 text_payload     | string             | The log entry payload, represented as a Unicode string (UTF-8). | Body
 severity         | LogSeverity        | The severity of the log entry.                          | Severity


specification/logs/data-model.md
```diff
@@ -946,7 +955,7 @@ Field            | Type               | Description
 timestamp        | string             | The time the event described by the log entry occurred. | Timestamp
 resource         | MonitoredResource  | The monitored resource that produced this log entry.    | Resource
 log_name         | string             | The URL-encoded LOG_ID suffix of the log_name field identifies which log stream this entry belongs to. | Attributes["gcp.log_name"]
-json_payload     | google.protobuf.Struct | The log entry payload, represented as a structure that is expressed as a JSON object. | Body
+json_payload     | google.protobuf.Struct | The log entry payload, represented as a structure that is expressed as a JSON object. | StructuredBody
```
braydonk, line 958: This does roughly align but may need some more thought from the Google side to line up properly; if there is a `Body` with the message, that message should likely go into `jsonPayload.message`. Not sure what the best way to codify that here would be.

damemi: Also, how does this reconcile with current users who may be parsing structured data to `body`? If `body` is structured, that will still go to `json_payload` too, or do we start flattening anything that's in `body`? This could be an implementation detail

tigrannajaryan:
> We've codified in the spec that `Body` SHOULD be a string (though it can technically be `any`).
> 
> On the other hand, `StructuredBody` would explicitly be structured data, either populated directly by the SDK or a parsed representation of `Body`. It would be similar to what we've been unsuccessfully arguing that `Attributes` should be for logs.

If we think it is necessary to capture structured data I think it is more preferable to remove the limitation that the Body should be a string. I think adding another field complicates the data model and is confusing.

djaglowski:
> If we think it is necessary to capture structured data I think it is more preferable to remove the limitation that the Body should be a string. I think adding another field complicates the data model and is confusing.

I mostly agree, but his would be a breaking change, right? I think what I've proposed avoids that, at least in the data model. 

One additional benefit is that having both `Body` and `StructuredBody` makes it quite easy to both parse and keep the original log. Exporters can then decide what to do if both fields are present.

tigrannajaryan:
> I mostly agree, but his would be a breaking change, right? I think what I've proposed avoids that, at least in the data model.

We uses a SHOULD clause in the data model. I see no problem with adding a list of exceptions to this SHOULD clause and saying that "in these cases you are allowed to use structure data in the Body". We already say "However, a structured body may be necessary to preserve the semantics of some existing log formats". Any number of similar exceptions can be added and it won't be a breaking change.

djaglowski:
This was discussed in the Log SIG today. It was decided that we should add clarification to the log data model's description of the Body, to the effect that structured logs emitted by third-party applications SHOULD use the Body for the structured data.

alexvanboxel:
We're very early adopters of OTel Logs, and we've been capturing JSON logs of our applications and parsing the strings into KeyValueList in the AnyValue (turns out you can decode JSON very in the KVLs). Granted, we have our own custom exporters for Google and DataDog logs that turn them into the appropriate format.

With StructuredBody, it becomes quite confusing... What do I do when we have KVLs, and what if I decided to structure my logs as Protobuf messages with the current body? It's simple: I add it add a Proto AnyType in the bytes field. I would argue that the Protobuf message is more structured than the StructuredBody.

PR #2979, opened 2022-11-22
Add OpenCensus metric bridge specification

dashpole:
Second attempt at https://github.com/open-telemetry/opentelemetry-specification/pull/2732

Part of https://github.com/open-telemetry/opentelemetry-specification/issues/1175

Related to https://github.com/open-telemetry/opentelemetry-specification/pull/2730

## Changes

Specify that OpenCensus metric bridges implement the MetricProducer interface, and describe non-trivial transformations involved.

This follows the path taken by the [java opencensus-shim](https://github.com/open-telemetry/opentelemetry-java/blob/main/opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/metrics/OpenCensusMetricProducer.java#L27).

PR #3015, opened 2022-12-06
Add OpenCensus migration plan

dashpole:

Supersedes https://github.com/open-telemetry/opentelemetry-specification/pull/2730.

Follow-up to https://github.com/open-telemetry/opentelemetry-specification/pull/2979

### Motivation

We have defined what OpenCensus metric and trace bridges, but don't have anywhere with recommendations on how to use them to migrate from OpenCensus to OpenTelemetry.  This adds a description of how users and library authors can migrate to OpenTelemetry.

### Removed

* **OpenCensus TextFormat Propagator**: Not needed, as it is equivalent to the OpenTelemetry W3c TraceContext propagator.
* **B3 Context propagator for OpenCensus**: Not needed, as b3 compatibility is already specified elsewhere, and is not OC-specific.
* **Semantic convention mapping for http**: Bridges should not attempt to translate semantic conventions.  Either way, metrics will need to be migrated.  Translation makes bridges more complex, likely to be wrong, and may not be consistent with the latest OTel semantic conventions.

### Added

* **Migration Guide**: For end-users, operators, and instrumentation library maintainers.  It describes how to use the components proposed below to migrate from OC to OTel.

### Changed

* **BinaryPropagation for gRPC**: Some languages (golang) can implement a BinaryPropagator as a TextMapPropagator.  Other languages cannot do that.  For languages that cannot implement a BinaryPropagator as a TextMapPropagator, **add BinaryPropagation as an option on gRPC instrumentation**.

The alternative to adding OC binary propagation directly to grpc instrumentation would be to introduce a binary propagator interface (similar to TextMapPropagator) in OTel.  Given that this is just for a migration, that likely isn't an appropriately scoped change.

### DIFF
--- a/specification/compatibility/opencensus.md
+++ b/specification/compatibility/opencensus.md
@@ -272,25 +272,3 @@ metric exporters.
 * OpenTelemetry does not currently support context-based attributes (tags).
 * OpenTelemetry does not support OpenCensus' SumOfSquaredDeviation field; this
   is dropped when using the bridge.
-
-## Semantic Convention Mappings
-
-Where possible, the tracing and metrics shims SHOULD provide mappings of labels
-to attributes defined within the OpenTelemetry semantic convetions.
-
-> The principle is to ensure OpenTelemetry exporters, which use these semantic
-> conventions, are likely to export the correct data.
-
-### HTTP Attributes
-
-OpenCensus specifies the following [HTTP Attributes](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md#attributes):
-
-| OpenCensus Name    | OpenTelemetry Name | Comments             |
-| ------------------ | ------------------ |----------------------|
-| `http.host`        | `http.host`        |                      |
-| `http.method`      | `http.method`      |                      |
-| `http.user_agent`  | `http.user_agent`  |                      |
-| `http.status_code` | `http.status_code` |                      |
-| `http.url`         | `http.url`         |                      |
-| `http.path`        | `http.target`      | key-name change only |
-| `http.route`       | N/A                | Pass through ok      |
--- a/specification/compatibility/opencensus.md
+++ b/specification/compatibility/opencensus.md
@@ -2,16 +2,18 @@
 
 **Status**: [Experimental](../document-status.md), Unless otherwise specified.
 
-## Abstract
-
 The OpenTelemetry project aims to provide backwards compatibility with the
 [OpenCensus](https://opencensus.io) project in order to ease migration of
 instrumented codebases.
 
-This functionality will be provided as a bridge layer implementing the
-[OpenCensus API](https://github.com/census-instrumentation/opencensus-specs)
-using the OpenTelemetry API. This layer MUST NOT rely on implementation specific
-details of the OpenTelemetry SDK.
+## Trace Bridge
+
+**Status**: [Experimental, Feature Freeze](../document-status.md)
+
+The trace bridge is provided as a shim layer implementing the
+[OpenCensus Trace API](https://github.com/census-instrumentation/opencensus-specs)
+using the OpenTelemetry Trace API. This layer MUST NOT rely on implementation
+specific details of the OpenTelemetry SDK.
 
 More specifically, the intention is to allow OpenCensus instrumentation to be
 recorded using OpenTelemetry. This Shim Layer MUST NOT publicly expose any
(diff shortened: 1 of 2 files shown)


dashpole:
cc @jack-berg @reyang @jsuereth @jmacd

specification/compatibility/opencensus.md
```diff
@@ -234,3 +272,25 @@ metric exporters.
...
+
+## Semantic Convention Mappings
+
+Where possible, the tracing and metrics shims SHOULD provide mappings of labels
+to attributes defined within the OpenTelemetry semantic convetions.
+
+> The principle is to ensure OpenTelemetry exporters, which use these semantic
+> conventions, are likely to export the correct data.
+
+### HTTP Attributes
```
jack-berg, line 284: FYI these have changed since originally written. The mapping will depend on whether we're talking about clients or servers (e.g. http.host will map to different attributes for client instrumentation than server instrumentation) or talking about traces vs metrics (the http metric semantic conventions aren't 100% consistent with the http trace semantic conventions).

specification/compatibility/opencensus.md
```diff
@@ -2,16 +2,89 @@
...
-using the OpenTelemetry API. This layer MUST NOT rely on implementation specific
-details of the OpenTelemetry SDK.
+## Migration Path
+
+### Breaking Changes when migrating to OpenTelemetry
+
+Migrating from OpenCensus to OpenTelemetry may require breaking changes to the telemetry produced
+because of:
+
+* Different or new semantic conventions for names and attributes (e.g. [`grpc.test.EchoService/Echo`](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/gRPC.md#grpc-trace) vs [`rpc.server.duration`](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.12.0/semantic_conventions/trace/rpc.yaml#L64))
```
jack-berg, line 16: `rpc.server.duration` is a metric name for the RPC metric semantic conventions. Did you mean to link to [this](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/rpc-metrics.md#rpc-server)?

specification/compatibility/opencensus.md
```diff
@@ -2,16 +2,89 @@
...
+
+#### Migrating an Application using Bridges
+
+Starting with an application using entirely OpenCensus instrumention for traces and metrics, it can be migrated by:
+
+1. Migrate the exporters (SDK)
+    1. Install the OpenTelemetry SDK, with an equivalent exporter
+        1. If using an OpenCensus exporter, switch to using an OTLP exporter
+    2. Install equivalent OpenTelemetry resource detectors
+    3. Install OpenTelemetry propagators for OpenCensus' `TextFormat` and `BinaryPropagator` formats.
```
jack-berg, line 44: Which propagators are equivalent to `TextFormat` and `BinaryPropagator`?

dashpole: Good catch.  These are completely different propagators from the current list in the spec.  They don't appear to be well-documented anywhere in OpenCensus.  Here is the OpenCensus Java [TextFormat](https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/trace/propagation/TextFormat.java) and [BinaryFormat](https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/trace/propagation/BinaryFormat.java) propagators.  OTel go has a [binary format propagator](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/propagators/opencensus), but I don't know of any others that have been implemented.

specification/compatibility/opencensus.md
dashpole: I wonder if we should remove the attempt to convert semantic conventions.  It might be better to suggest using OTel views to adapt OTel semconv to the previous OC names...  WDYT about removing this and the above sections?

specification/compatibility/opencensus.md
dashpole: fixed.

specification/compatibility/opencensus.md
dashpole: Opened https://github.com/open-telemetry/opentelemetry-specification/pull/3035

MrAlias (approved):

specification/compatibility/opencensus.md
dashpole: TextFormat turned out to be the same as traceparent (I was convinced it was different, but must have been looking at the wrong thing.)

jack-berg (approved):

jsuereth (approved):

carlosalberto (approved):

carlosalberto:
@dashpole Changelog entry please. Otherwise it looks we are ready to go.

specification/compatibility/opencensus.md
```diff
@@ -2,16 +2,90 @@
...
+### Migration Plans
+
+#### Migrating from the OpenCensus Agent and Protocol
+
+Starting with a deployment of the OC agent, using the OC protocol, migrate by:
+
+1. Deploy the OpenTelemetry collector with OpenCensus and OTLP receivers and equivalent processors and exporters.
+2. **breaking**: For each workload sending the OC protocol, change to sending to the OpenTelemetry collector's OpenCensus receiver.
+3. Remove the deployment of the OC Agent
+4. For each workload, migrate the application from OpenCensus to OpenTelemetry, following the guidance below, and use the OTLP exporter
```
reyang, line 34: nit:

```suggestion
3. Remove the deployment of the OC Agent.
4. For each workload, migrate the application from OpenCensus to OpenTelemetry, following the guidance below, and use the OTLP exporter.
```

reyang (approved):

PR #3017, opened 2022-12-06
Add MaxScale config parameter to ExponentialHistogram

### LINKED ISSUE #2987: Define configurable max scale for exponential histogram aggregation
The exponential histogram aggregation defines a [max scale](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#support-a-minimum-and-maximum-scale) without giving concrete advice on what "reasonable" minimum and maximum scales might be. 

The max scale is particularly important because as I've recently [found](https://github.com/open-telemetry/opentelemetry-java/pull/4989), computing the bucket index with the [the Logarithm Function](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md#all-scales-use-the-logarithm-function) is ~9x slower than the bit manipulation technique. The logarithm function is required for positive scales, while the bit manipulation version can be used at scales less than or equal to zero. 

I suggest we:
1. Make min and max scale configurable. High performance applications may want to avoid the performance hit of indexing using the log function, as well as unnecessary rescaling.
2. Provide guidance on the default min and max scale.
3. Provide guidance on the bounds for min and max scale. Users would not be allowed to configure min and max scale outside of these bounds. 

cc @jmacd @jsuereth

jack-berg:
Resolves #2987.


jmacd (approved):

jack-berg:
@open-telemetry/specs-metrics-approvers PTAL.

reyang (approved):

reyang:
@jack-berg do you want to update changelog?

PR #3018, opened 2022-12-07
Prometheus histogram bucket metric ends in _bucket

dashpole:
I realized when opening https://github.com/open-telemetry/opentelemetry-swift/issues/355 that the current prometheus spec for histograms is incorrect (though almost all implementations of prometheus exporters is correct).

From the [prometheus spec](https://github.com/prometheus/docs/blob/main/content/docs/instrumenting/exposition_formats.md#histograms-and-summaries):

> Each bucket count of a histogram named x is given as a separate sample line with the name x_bucket and a label {le="y"} (where y is the upper bound of the bucket).

From the [OpenMetrics spec](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#histogram-1):

> The MetricPoint's Bucket Values Sample MetricNames MUST have the suffix "_bucket".

This corrects our specification to be in-line with prometheus and OM specs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ release.
 
 ### Compatibility
 
+- Fix Prometheus histogram metric suffixes. Bucket series end in `_bucket`
+  ([#3018](https://github.com/open-telemetry/opentelemetry-specification/pull/3018)).
+
 ### OpenTelemetry Protocol
 
 ### SDK Configuration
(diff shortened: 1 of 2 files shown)


dashpole:
cc @gouthamve @jmacd

gouthamve (approved):
LGTM

dashpole:
cc @open-telemetry/wg-prometheus

jack-berg (approved):

reyang (approved):

dashpole:
I believe this can be merged.

Aneurysm9 (approved):

Issue #2497, opened 2022-04-18
Reconsider modifying metric name with OTLP unit


anuraaga:
The prometheus conversion currently requires metric names to be modified with the OTLP unit

> Additionally, the unit MUST be added as a suffix to the metric name, and SHOULD be converted to [base units](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#units-and-base-units) recommended by OpenMetrics when possible

I think this should be reconsidered, I suspect users don't expect their metric names to not actually be the names, and e.g. when configuring grafana `http_server_duration` seems expected vs `http_server_duration_ms`.

anuraaga:
@open-telemetry/wg-prometheus

dashpole:
The prometheus metric would ideally be named `http_server_duration_seconds`, which is close to the expected naming for that metric in the prometheus ecosystem (`http_request_duration_seconds`, which is used [in the prometheus metric naming examples](https://prometheus.io/docs/practices/naming/#metric-names)).

We could consider making it a "MUST, by default", rather than just a "MUST", to allow users the option of excluding the unit.  According to OpenMetrics, if we add a unit comment (which we really should IMO), we [MUST add the unit as the metric suffix](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metricfamily-metadata) as well.

anuraaga:
Thanks I was looking for the text in OpenMetrics but couldn't find it. It looks like it should be added then, wonder if OTel's default of ms vs seconds will cause issues for users though

dashpole:
Yeah, a metric ending in `_ms` would be strange to prometheus users. Even spelling out `_milliseconds` would be better than `_ms`.  Ideally, exporters would convert metrics in `ms` to `seconds` in their Prometheus exporter.  That is what I intended by `SHOULD be converted to base units recommended by OpenMetrics` in the spec.  Do you think that is feasible?

jack-berg:
> Do you think that is feasible?

We'd have to maintain a table of origin units, target units, and conversion functions. Could probably have a smallish library of common situations that would cover the majority of use cases. Ideally this would be spec'd. 

[...]

Issue #2977, opened 2022-11-22
Semantic Conventions: default to seconds for duration units


gouthamve:
Prometheus and OpenMetrics strongly recommend that the units for measuring durations should be `seconds`. However, the [semantic conventions here](https://opentelemetry.io/docs/reference/specification/metrics/semantic_conventions/http-metrics/) are in `milliseconds`.

This creates a lot of confusion in users who are expecting durations to be seconds and requires an addition ` / 1e3` when doing math with metrics that come from traditional Prometheus sources. For example, today Kubernetes metrics are all in `_seconds` mainly because of the Prometheus conventions. This would be similar for other existing systems as well.

Given its all floats, I think we should try and revist the decision to use `milliseconds` and try to align with Prometheus.

gouthamve:
cc @jmacd @jsuereth

jack-berg:
The [default buckets](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#explicit-bucket-histogram-aggregation) for explicit bucket histogram are aligned to milliseconds for `http.server.duration`. We've previously discussed that changing them is a breaking change and not allowed. Therefore, changing from milliseconds to seconds for duration would render the default buckets useless.

gouthamve:
Hrm, so while converting from OTLP to Prometheus, we can always convert to `_seconds` in the case of explicit bucket histograms. Its simply downscaling the buckets by `/1000`.

But this is not possible with exponential histograms. Can we make it `milliseconds` for exponential histograms?

gouthamve:
This also causes inconsistencies when ingesting metrics from both Prometheus sources and OTel sources into a single db. Half the exponential histograms would be in seconds and the other half would be in milliseconds with no good way to reconcile/aggregate the two.

It is possible convert with explicit bucket histograms, hence initially I didn't mind the unit being milliseconds, but when I realised that its not possible with exponential histograms, I wanted to propose this change.

dashpole:
If we introduced a way for instrumentation to override the default set of buckets (which could still be overridden by views), would that allow individual instrumentation libraries to switch to seconds?

jack-berg:
> If we introduced a way for instrumentation to override the default set of buckets 

So you're imagining using what has been referred to as the "hint API" to perhaps have all `http.server.duration` instrumentation to report in seconds and specify alternative default buckets that are sensible for seconds?

nerdondon:
[...]

PR #3019, opened 2022-12-07
Don't recommend converting to prometheus base units, but do convert to full words

dashpole:
As described in https://github.com/open-telemetry/opentelemetry-specification/issues/2497#issuecomment-1102740872 (and the following discussion), it isn't possible to convert from units to base units (e.g. convert milliseconds to seconds) for all metric types.  This PR withdraws that recommendation.

I opened a few issues for adding units to metric names:

* https://github.com/open-telemetry/opentelemetry-swift/issues/359
* https://github.com/open-telemetry/opentelemetry-rust/issues/927
* https://github.com/open-telemetry/opentelemetry-dotnet/issues/3975
* https://github.com/open-telemetry/opentelemetry-cpp/issues/1840
* https://github.com/open-telemetry/opentelemetry-js/issues/3466

During that process, I didn't want to recommend converting to base units for the reasons above, but I do still think implementations should convert to full words where possible.

cc @gouthamve @jmacd

### DIFF
--- a/specification/compatibility/prometheus_and_openmetrics.md
+++ b/specification/compatibility/prometheus_and_openmetrics.md
@@ -224,9 +224,8 @@ consecutive `_` characters MUST be replaced with a single `_` character.
 The Unit of an OTLP metric point MUST be added as
 [OpenMetrics UNIT metadata](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metricfamily).
 Additionally, the unit MUST be added as a suffix to the metric name, and SHOULD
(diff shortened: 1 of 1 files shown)


gouthamve:
I don't understand why we can't convert from `milliseconds` to `seconds` on fixed bucket histograms. Its easy.

Its not possible in exponential histograms, and we have a pending issue here: https://github.com/open-telemetry/opentelemetry-specification/issues/2977

I would wait for #2977 to resolve before we decide this tbh. 

And even if we keep `milliseconds` in exponential histograms, I might be tempted to keep `seconds` for fixed bucket histograms. I need to think more about this.

dashpole:
Discussed this as the prometheus wg.  We are going to hold off on this while we discuss https://github.com/open-telemetry/opentelemetry-specification/issues/2977

Issue #3006, opened 2022-12-05
Misleading description of IsRecording() in trace API


utezduyar:
I believe the text below from the spec misleading. 

```
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md
IsRecording
Returns true if this Span is recording information like events with the AddEvent operation, attributes using SetAttributes, status with SetStatus, etc.
```

The text makes me think that if any of these APIs (AddEvent, SetAttribute..) are called, the recording status will be TRUE. Seems like `IsRecording` returns True if the Span is sampled and the sample hasn't ended. At least the go implementation tells me that. 

```
// IsRecording returns if this span is being recorded. If this span has ended
// this will return false.
func (s *recordingSpan) IsRecording() bool {
	if s == nil {
		return false
	}
	s.mu.Lock()
	defer s.mu.Unlock()

	return s.endTime.IsZero()
}

```

I believe the current text is valuable but needs to be added in a way that none of these APIs (AddEvent, SetAttribute..) will work (in other terms will be NOOP) after the span is ended. I would be happy to send a patch if I have a point.

Oberon00:
It was meant to say that calling these APIs won't record any information if IsRecording is false, I guess. Wording improvements are welcome.

utezduyar:
Hi my commit has been waiting for "First-time contributors need a maintainer to approve running workflows". Is this something I need to trigger or please let me know if I need to do anything else.

PR #3020, opened 2022-12-07
Update Tracing API: IsRecording wording

utezduyar:
Fixes https://github.com/open-telemetry/opentelemetry-specification/issues/3006

## Changes

Please provide a brief description of the changes here.

For non-trivial changes, follow the [change proposal process](https://github.com/open-telemetry/opentelemetry-specification/blob/main/CONTRIBUTING.md#proposing-a-change) and link to the related issue(s) and/or [OTEP(s)](https://github.com/open-telemetry/oteps), update the [`CHANGELOG.md`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/CHANGELOG.md), and also be sure to update [`spec-compliance-matrix.md`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md) if necessary.

Related issues #

### DIFF
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -454,7 +454,9 @@ The Span interface MUST provide:
 
 #### IsRecording
 
-Returns `true` if this `Span` has been started but not ended.
+Always returns `false` If this is a non-recording `Span` (Sampling
+`Decision` was `DROP` at the time of `Span` creation). Otherwise returns
+`true` if this `Span` has been started but not ended.
 After a `Span` is ended, it usually becomes non-recording and thus
 `IsRecording` SHOULD consequently return `false` for ended Spans.
 Note: Streaming implementations, where it is not known if a span is ended,
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -480,13 +480,6 @@ span's recording is determined independently from the value of this flag
 (typically based on the `sampled` flag of a `TraceFlags` on
 [SpanContext](#spancontext)).
 
-This flag may be `true` despite the entire trace being sampled out. This
-allows to record and process information about the individual Span without
-sending it to the backend. An example of this scenario may be recording and
-processing of all incoming requests for the processing and building of
-SLA/SLO latency charts while sending only a subset - sampled spans - to the
-backend. See also the [sampling section of SDK design](sdk.md#sampling).
-
 Users of the API should only access the `IsRecording` property when
 instrumenting code and never access `SampledFlag` unless used in context
 propagators.
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -462,7 +462,7 @@ Note: Streaming implementations, where it is not known if a span is ended,
 are one expected case where `IsRecording` cannot change after ending a Span.
 
 Recording information like events with the `AddEvent` operation, attributes
-using `SetAttributes`, status with `SetStatus`, etc will not perform when
+using `SetAttributes`, status with `SetStatus`, etc will be discarded when
 `IsRecording` returns `false`.
 
 `IsRecording` SHOULD NOT take any parameters.
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -454,14 +454,15 @@ The Span interface MUST provide:
 
 #### IsRecording
 
-Returns true if this `Span` is recording information like events with the
-`AddEvent` operation, attributes using `SetAttributes`, status with `SetStatus`,
-etc.
-
+Returns `true` if this `Span` has been started but not ended.
 After a `Span` is ended, it usually becomes non-recording and thus
-`IsRecording` SHOULD consequently return false for ended Spans.
+`IsRecording` SHOULD consequently return `false` for ended Spans.
 Note: Streaming implementations, where it is not known if a span is ended,
 are one expected case where `IsRecording` cannot change after ending a Span.
+
+Recording information like events with the `AddEvent` operation, attributes
+using `SetAttributes`, status with `SetStatus`, etc will not perform if
+`IsRecording` returns `false`.
 
 `IsRecording` SHOULD NOT take any parameters.
 
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -467,8 +467,8 @@ processing of all incoming requests for the processing and building of
 SLA/SLO latency charts while sending only a subset - sampled spans - to the
 backend. See also the [sampling section of SDK design](sdk.md#sampling).
 
-After a `Span` is ended, it SHOULD become non-recording and its `IsRecording`
-SHOULD start consequently return `false`. The one known exception to this is
+After a `Span` is ended, it SHOULD become non-recording and `IsRecording`
+SHOULD always return `false`. The one known exception to this is
 streaming implementations of the API that do not keep local state and cannot
 change the value of `IsRecording` after ending the span.
 
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -461,7 +461,7 @@ some form (e.g. in memory). When a `Span` is not recording (`IsRecording` return
 data will not record, making the span effectively a no-op.
 
 This flag may be `true` despite the entire trace not being sampled. This
-allows to record and process information about the individual Span without
+allows information about the individual Span to be recorded and processed without
 sending it to the backend. An example of this scenario may be recording and
 processing of all incoming requests for the processing and building of
 SLA/SLO latency charts while sending only a subset - sampled spans - to the
--- a/specification/trace/api.md
+++ b/specification/trace/api.md
@@ -454,16 +454,23 @@ The Span interface MUST provide:
 
 #### IsRecording
 
-Always returns `false` If this is a non-recording `Span`. Otherwise returns
-`true` if this `Span` has been started but not ended.
-After a `Span` is ended, it usually becomes non-recording and thus
-`IsRecording` SHOULD consequently return `false` for ended Spans.
-Note: Streaming implementations, where it is not known if a span is ended,
-are one expected case where `IsRecording` cannot change after ending a Span.
+A `Span` is recording (`IsRecording` returns `true`) when the data provided to
+it via functions like `SetAttributes`, `AddEvent`, `SetStatus` is captured in
+some form (e.g. in memory). When a `Span` is not recording (`IsRecording` returns
+`false`), all this data is discarded right away. Further attempts to set or add
+data will not record, making the span effectively a no-op.
 
-Recording information like events with the `AddEvent` operation, attributes
-using `SetAttributes`, status with `SetStatus`, etc will be discarded when
-`IsRecording` returns `false`.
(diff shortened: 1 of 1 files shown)


specification/trace/api.md
```diff
@@ -454,15 +454,16 @@ The Span interface MUST provide:
 
 #### IsRecording
 
-Returns true if this `Span` is recording information like events with the
-`AddEvent` operation, attributes using `SetAttributes`, status with `SetStatus`,
-etc.
-
+Returns `true` if this `Span` has been started but not ended.
```
MrAlias, line 457: This statement does not take into account spans that are sampled to be non-recording. A span being started is not the only criteria for a span to be recorded.

utezduyar: Maybe I misinterpret your first sentence but I think sampled but non-recording is against the spec. https://opentelemetry.io/docs/reference/specification/trace/sdk/#shouldsample `The flag combination SampledFlag == true and IsRecording == false could cause gaps in the distributed trace, and because of this OpenTelemetry API MUST NOT allow this combination.`

My patch attempt is by me reverse engineering the golang SDK. Maybe the golang implementation is not complete but otherwise these 2 variants are how the API is implemented.

```
// IsRecording returns if this span is being recorded. If this span has ended
// this will return false.
func (s *recordingSpan) IsRecording() bool {
	if s == nil {
		return false
	}
	s.mu.Lock()
	defer s.mu.Unlock()

	return s.endTime.IsZero()
}
```

And if it is a non-recording span:

```
// IsRecording always returns false.
func (nonRecordingSpan) IsRecording() bool { return false }
```

Maybe something like this accounts the non-recording spans.
`Returns true for recording spans if this Span has been started but not ended. Returns false after the Span is ended or if the Span is a non recording span.`

Let me know what you think please.

arminru: If you start a span but the sampler returns a negative sampling decision, the SDK will return a non-recording span that is not recording and not sampled. You still have a span in a "started" state, so the wording proposed here saying just "started but not ended" will not work, as it does not account for such non-recording spans (the second code snippet you cited).

See also https://opentelemetry.io/docs/reference/specification/trace/sdk/#recording-sampled-reaction-table and https://opentelemetry.io/docs/reference/specification/trace/sdk/#shouldsample ("Return value").

arminru (changes requested):

specification/trace/api.md
utezduyar: I believe I have updated the patch to reflect your comment. I believe it is you who will mark the conversation as "resolved".

arminru: Thanks, Umut!

While this is correct now, it references implementation details from the SDK specification and actually duplicates these statements, making the spec harder to maintain and breaking the distinction made on purpose. The API spec you are editing here defines what IsRecording means to users of the API, while the SDK spec defines what SDKs should do and return based on the sampling decision.
It might be a bit confusing at first but I think we should not mix these two together here.

If there is any clarification that can be made without referencing the SDK but defining better how to _interpret_ the values returned for IsRecording, feel free to propose something.

specification/trace/api.md
```diff
@@ -454,15 +454,18 @@ The Span interface MUST provide:
...
+`Decision` was `DROP` at the time of `Span` creation). Otherwise returns
+`true` if this `Span` has been started but not ended.
 After a `Span` is ended, it usually becomes non-recording and thus
-`IsRecording` SHOULD consequently return false for ended Spans.
+`IsRecording` SHOULD consequently return `false` for ended Spans.
 Note: Streaming implementations, where it is not known if a span is ended,
 are one expected case where `IsRecording` cannot change after ending a Span.
 
+Recording information like events with the `AddEvent` operation, attributes
+using `SetAttributes`, status with `SetStatus`, etc will not perform if
```
jmacd, line 466:
```suggestion
using `SetAttributes`, status with `SetStatus`, etc will not perform when
```

specification/trace/api.md
```diff
@@ -454,15 +454,18 @@ The Span interface MUST provide:
...
-Returns true if this `Span` is recording information like events with the
-`AddEvent` operation, attributes using `SetAttributes`, status with `SetStatus`,
-etc.
-
+Always returns `false` If this is a non-recording `Span` (Sampling
+`Decision` was `DROP` at the time of `Span` creation). Otherwise returns
+`true` if this `Span` has been started but not ended.
 After a `Span` is ended, it usually becomes non-recording and thus
-`IsRecording` SHOULD consequently return false for ended Spans.
+`IsRecording` SHOULD consequently return `false` for ended Spans.
```
jmacd, line 461: (side remark: The appearance of "usually" is ... unusual without a corresponding SHOULD statement. I wonder if we should replace "usually" with "SHOULD" and if not, why not?)

jmacd (approved):

yurishkuro (changes requested):

specification/trace/api.md
utezduyar: English is not my first language but shouldn't SHOULDs be SHALLS? And why upper case? 

I have been quite confused when I read this paragraph. What I got confused is "`it usually becomes non-recording...`" So there is an exception, seems like streaming applications. Then we have this: "`are one expected case where IsRecording cannot change after ending a Span.`" I have not seen this case in go SDK implementation. 

With my limited knowledge, I can agree with what you propose. "`After a Span is ended, it SHOULD become non-recording..`"

specification/trace/api.md
utezduyar: Thanks again. I just removed the the reference to the SDK. @yurishkuro has also suggested the same thing.

specification/trace/api.md
```diff
@@ -454,15 +454,18 @@ The Span interface MUST provide:
 
 #### IsRecording
 
-Returns true if this `Span` is recording information like events with the
-`AddEvent` operation, attributes using `SetAttributes`, status with `SetStatus`,
-etc.
-
+Always returns `false` If this is a non-recording `Span` (Sampling
```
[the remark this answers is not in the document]
utezduyar: I am blank regarding how to address your comment. Any suggestions? 

We do mention about the behavior when recording and sampling flags are set under the section Sampling in the SDK. Maybe the SDK should be made more clear to reflect memory Vs export.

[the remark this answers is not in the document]
Oberon00: IsRecording = false should mean that the data is discarded right away, i.e. nobody will (be able to) read it, not even in memory. It's an effectively no-op span.

yurishkuro:
My proposal to change the original lines 457-464:

A span is _**recording**_ (`IsRecording` returns `true`) when the data provided to it via functions like `SetAttributes`, `AddEvent`, `SetStatus` is captured in some form (e.g., in memory). When a span is not recording (`IsRecording` returns `false`), all this data is discarded right away, making the span effectively a no-op.

{move paragraph from L474 here}

After a `Span` is ended, it SHOULD become non-recording and its `IsRecording` SHOULD start consequently return `false`. The one known exception to this is streaming implementations of the API that do not keep local state and cannot change the value of `IsRecording` after ending the span.

utezduyar:
> 

I have mostly used your text, made couple of minor changes.

specification/trace/api.md
arminru: @utezduyar You can find details on the distinction and the casing in https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification#notation-conventions-and-compliance and the RFCs it refers to.

specification/trace/api.md
```diff
@@ -454,14 +454,23 @@ The Span interface MUST provide:
...
+
+This flag may be `true` despite the entire trace being sampled out. This
+allows to record and process information about the individual Span without
+sending it to the backend. An example of this scenario may be recording and
+processing of all incoming requests for the processing and building of
+SLA/SLO latency charts while sending only a subset - sampled spans - to the
+backend. See also the [sampling section of SDK design](sdk.md#sampling).
+
+After a `Span` is ended, it SHOULD become non-recording and its `IsRecording`
+SHOULD start consequently return `false`. The one known exception to this is
```
arminru, line 471:
```suggestion
SHOULD consequently return `false`. The one known exception to this are
```

utezduyar: This is & are twisting me. Are you sure it is "are"?

specification/trace/api.md
```diff
@@ -454,14 +454,23 @@ The Span interface MUST provide:
...
+it via functions like `SetAttributes`, `AddEvent`, `SetStatus` is captured in
+some form (e.g. in memory). When a `Span` is not recording (`IsRecording` returns
+`false`), all this data is discarded right away. Further attempts to set or add
+data will not record, making the span effectively a no-op.
 
-After a `Span` is ended, it usually becomes non-recording and thus
-`IsRecording` SHOULD consequently return false for ended Spans.
-Note: Streaming implementations, where it is not known if a span is ended,
-are one expected case where `IsRecording` cannot change after ending a Span.
+This flag may be `true` despite the entire trace being sampled out. This
```
yurishkuro, line 463:
```suggestion
This flag may be `true` despite the entire trace not being sampled. This
```

yurishkuro (approved):

specification/trace/api.md
```diff
@@ -454,14 +454,23 @@ The Span interface MUST provide:
...
+some form (e.g. in memory). When a `Span` is not recording (`IsRecording` returns
+`false`), all this data is discarded right away. Further attempts to set or add
+data will not record, making the span effectively a no-op.
 
-After a `Span` is ended, it usually becomes non-recording and thus
-`IsRecording` SHOULD consequently return false for ended Spans.
-Note: Streaming implementations, where it is not known if a span is ended,
-are one expected case where `IsRecording` cannot change after ending a Span.
+This flag may be `true` despite the entire trace not being sampled. This
+allows to record and process information about the individual Span without
```
MrAlias, line 464:
```suggestion
allows information about the individual Span to be recorded and processed without
```

specification/trace/api.md
```diff
@@ -454,14 +454,23 @@ The Span interface MUST provide:
...
+sending it to the backend. An example of this scenario may be recording and
+processing of all incoming requests for the processing and building of
+SLA/SLO latency charts while sending only a subset - sampled spans - to the
+backend. See also the [sampling section of SDK design](sdk.md#sampling).
+
+After a `Span` is ended, it SHOULD become non-recording and its `IsRecording`
+SHOULD start consequently return `false`. The one known exception to this is
```
MrAlias, line 471:
```suggestion
After a `Span` is ended, it SHOULD become non-recording and `IsRecording`
SHOULD always return `false`. The one known exception to this is
```

specification/trace/api.md
MrAlias: "The one known exception" is the singular subject, "is" correct in this case.

MrAlias (approved):

yurishkuro:
@arminru please re-check if you still require changes

arminru (approved):

PR #3023, opened 2022-12-07
Clarify usage of log body for structured logs

djaglowski:
See decision by Log SIG noted in [#3014](https://github.com/open-telemetry/opentelemetry-specification/pull/3014#issuecomment-1341417757)


specification/logs/data-model.md
```diff
@@ -419,9 +419,10 @@ Description: A value containing the body of the log record (see the description
 of `any` type above). Can be for example a human-readable string message
 (including multi-line) describing the event in a free form or it can be a
 structured data composed of arrays and maps of other values. First-party
-Applications SHOULD use a string message. However, a structured body may be
-necessary to preserve the semantics of some existing log formats. Can vary for
-each occurrence of the event coming from the same source. This field is optional.
+Applications SHOULD use a string message. However, a structured body SHOULD be
```
tigrannajaryan, line 422: >However, a structured body SHOULD be used

I am a bit unsure about this phrasing. It seems to mandate using a structure body, without making an exception for the case the source third-party log is not structured.

djaglowski: I can try to reword, but later in the sentence it says "to preserve the semantics of _structured_ logs". It's also a SHOULD, so exceptions may be made.

tigrannajaryan (approved):

specification/logs/data-model.md
```diff
@@ -419,9 +419,10 @@ Description: A value containing the body of the log record (see the description
...
+used to preserve the semantics of structured logs emitted by Third-party
```
MrAlias, line 423:
```suggestion
used to preserve the semantics of structured logs emitted by third-party
```

specification/logs/data-model.md
```diff
@@ -419,9 +419,10 @@ Description: A value containing the body of the log record (see the description
...
+applications. Can vary for each occurrence of the event coming from the same
+source. This field is optional.
```
MrAlias, line 425: "Can vary for each occurrence of the event coming from the same source."

The subject of this sentence is unclear. Can you clarify what can vary here?

djaglowski: I think it is referring to the body. Similar language is used in other descriptions in this file. This sentence is not changed in this PR though.

specification/logs/data-model.md
djaglowski: "Third-party Applications" is defined above as particular term, so my intention here was to reference that term. The same is done earlier in this paragraph for "First-party Applications".

I don't think it's strictly necessary to keep the capitalization. That said, I forgot to capitalize Applications, so I think we should decide one way or the other. I slightly prefer to match the capitalization referenced above.

djaglowski: I've capitalized the A for now, but let me know if you feel strongly about how this is used.

specification/logs/data-model.md
```diff
@@ -419,9 +419,10 @@ Description: A value containing the body of the log record (see the description
 of `any` type above). Can be for example a human-readable string message
 (including multi-line) describing the event in a free form or it can be a
 structured data composed of arrays and maps of other values. First-party
-Applications SHOULD use a string message. However, a structured body may be
-necessary to preserve the semantics of some existing log formats. Can vary for
-each occurrence of the event coming from the same source. This field is optional.
+Applications SHOULD use a string message. However, a structured body SHOULD be
+used to preserve the semantics of structured logs emitted by Third-party
+applications. Can vary for each occurrence of the event coming from the same
```
djaglowski, line 424:
```suggestion
Applications. Can vary for each occurrence of the event coming from the same
```

djaglowski:
@open-telemetry/specs-approvers, please take a look

jack-berg (approved):

jack-berg:
Please add changelog entry. Thanks!

PR #3024, opened 2022-12-07
Eager exporting batch span processor

### LINKED ISSUE #849: Clarify whether BatchSpanProcessor scheduledDelayMillis is between batches or entire processing of queue
Currently we define `scheduledDelayMillis`, `maxQueueSize`, and `maxExportBatchSize`.

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#batching-processor

It's not clear to me whether `scheduledDelayMillis` is supposed to be time in between exports of chunks of size `maxExportBatchSize`, or if it's the delay between processing the queue, when processing we export as many batches as we can, and then wait for the delay amount until the next set of batches. The wording in the spec seems closer to the former, but go and java at least seem to implement as the latter.

dyladan:
Fixes #849

## Changes

Clarify that the BSP should export whenever the queue reaches the batch size or the scheduled delay timer completes, whichever is first.


yurishkuro (approved):

seemk (approved):

arminru (approved):

carlosalberto (approved):

dyladan:
I think this can be merged

specification/trace/sdk.md
```diff
@@ -579,17 +579,23 @@ This is an implementation of the `SpanProcessor` which create batches of finishe
 spans and passes the export-friendly span data representations to the
 configured `SpanExporter`.
 
+The processor SHOULD export a batch when `scheduledDelayMillis` milliseconds
+have elapsed since the processor was created or the previous export completed,
```
blumamir, line 583: In JS [the timer is set when the first span joins the batch](https://github.com/open-telemetry/opentelemetry-js/blob/c93ab9e84edfffbae7e63894f6c960262a8c299e/packages/opentelemetry-sdk-trace-base/src/export/BatchSpanProcessorBase.ts#L123), which makes sense to me. This behavior is not covered by the current wording. It is also not dependent on the previous export, so if previous export takes long time, a new export can start after `scheduledDelayMillis` regardless of the previous export.

I suggest considering adding it as well

jmacd (approved):

specification/trace/sdk.md
MrAlias:
```suggestion
have elapsed since the processor was created or the previous export completed and it has received at lease one span,
```

blumamir: @MrAlias - the text still reads as if the processor has to wait for the previous export to finish before it sets the timer for the next batch. It implies that setting the `scheduledDelayMillis` as soon as we export one batch is incorrect.

In JS, the next timer can start as soon as we **begin** exporting regardless of when the previous batch is complete, which sounds ok but violates the new wording.

seemk: The case where an export is started while another export is in concurrently in flight doesn't seem correct - it would make the max queue size parameter useless as a new export is started right away. The current JS  BSP implementation is flawed anyway, see https://github.com/open-telemetry/opentelemetry-js/pull/3458 for an attempt to fix it.

blumamir: > it would make the max queue size parameter useless as a new export is started right away

Why would a new export start right away?

dyladan: @arminru 3e2ed381e86f6522d340abbdaec944abb5bb9fd6

specification/trace/sdk.md
```diff
@@ -579,17 +579,26 @@ This is an implementation of the `SpanProcessor` which create batches of finishe
...
+or when the queue contains `maxExportBatchSize` or more spans.
+After the processor is created or the previous export completed, a processor
+MAY wait to start the `scheduledDelayMillis` timer until the first span is
```
reyang, line 586: What does "the first span is added to the batch" mean?

For example:

* at Tn, the previous batch export completed.
* at Tn + 5, a new span was added to the batch.
* at Tn + 6, two extra spans were added to the batch.

Assuming the `scheduledDelayMillis` is `10000` (10 seconds), should the next export happen at `Tn + 10` or `Tn + 15`?

jsuereth (approved):

specification/trace/sdk.md
dyladan: This PR allows for both situations (`Tn + 10` and `Tn + 15`) in order to be compatible with as many existing implementations as possible. Can you suggest wording that would be more clear to you?

specification/trace/sdk.md
```diff
@@ -579,17 +579,26 @@ This is an implementation of the `SpanProcessor` which create batches of finishe
 spans and passes the export-friendly span data representations to the
 configured `SpanExporter`.
 
+The processor SHOULD export a batch when `scheduledDelayMillis` milliseconds
```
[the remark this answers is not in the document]
dyladan: I left it undefined in this PR because there are already existing implementations and I didn't want to add new spec that would make them invalid. This PR tries to be as inclusive of existing implementations as possible.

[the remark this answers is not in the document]
reyang: > I left it undefined in this PR ...

Consider calling this out explicitly, the current wording "The processor SHOULD export..." seems to suggest that a batch needs to be exported even if it is empty.

specification/trace/sdk.md
blumamir: I was not aware of this part of the spec: "Export() will never be called concurrently for the same exporter instance."

In this case, this text is ok and the thread can be resolved

specification/trace/sdk.md
```diff
@@ -579,17 +579,26 @@ This is an implementation of the `SpanProcessor` which create batches of finishe
...
+have elapsed since the processor was created or the previous export completed,
+or when the queue contains `maxExportBatchSize` or more spans.
```
blumamir, line 584: > or when the queue contains `maxExportBatchSize` or more spans.

Should also be limited to happen after previous export is completed

specification/trace/sdk.md
alanwest: @blumamir The presence of the max queue size parameter implies that only one export should be in flight at a time. If we allowed for multiple exports at a time, triggered by the max batch size being reached, then the max queue size parameter would be unnecessary.

Though, this one-export-at-a-time behavior could probably be more explicitly stated. It's not great that it's only implied due to the presence of a config parameter. Maybe another paragraph like

> The processor MUST invoke its configured exporter synchronously allowing the export to complete before invoking another export.

specification/trace/sdk.md
dyladan: At least JS doesn't start the timer until the previous export is resolved

specification/trace/sdk.md
reyang, line 584:
```suggestion
or when the queue contains `maxExportBatchSize` or more spans. If the batch is empty (does not contain any item), the processor MAY decide to still export it or simply ignore it.
```

dyladan:
After feedback from @reyang and @blumamir I changed the wording to make it more explicit and hopefully cover more cases. Please take a look and make sure it addresses your concerns.

dyladan:
Dead link not mine

specification/trace/sdk.md
```diff
@@ -579,17 +579,39 @@ This is an implementation of the `SpanProcessor` which create batches of finishe
 spans and passes the export-friendly span data representations to the
 configured `SpanExporter`.
 
+The processor SHOULD export a batch when any of the following happens AND
+there is not an export currently in progress:
+
+- `scheduledDelayMillis` after the processor is constructed OR the first span
+  is received from the span processor
```
reyang, line 586: nit:

```suggestion
  is received from the span processor.
```

specification/trace/sdk.md
```diff
@@ -579,17 +579,39 @@ This is an implementation of the `SpanProcessor` which create batches of finishe
...
+- `scheduledDelayMillis` after the previous export timer ends, OR the previous
+  export completes, OR the first span is added to the queue after the previous
+  export timer ends or previous batch completes.
+- The queue contains `maxExportBatchSize` or more spans.
```
reyang, line 590: This might get trickier - we might need to add "OR when the processor's ForceFlush/Shutdown is being called" (and I can see the wording challenge here).

dyladan: 1cd74b5e28603848d477cfaab522d32f4404990b

reyang (approved):

specification/trace/sdk.md
MSNev: > AND there is not an export currently in progress

Why? While I understand that we want to limit the number of outbound connections, there are exceptions where this will block and cause issues. One seems to be restrictive... I would prefer to see this change to reference a `maxConnections` (defaulting to at least 2) but thats probably another PR.

Additionally, during app/client (page) unload, if an export is already in flight and the page wants to "terminate" this will cause any spans triggered as part of the "unload" to be dropped...

jack-berg (approved):

blumamir (approved):

specification/trace/sdk.md
```diff
@@ -579,17 +579,41 @@ This is an implementation of the `SpanProcessor` which create batches of finishe
...
+- `ForceFlush` is called.
+
+In any of the above cases the processor MUST NOT begin an export if another
+export is in progress. If one of the above events occurs during an ongoing export
+the processor MAY do either of the following:
+
+- Export the next batch immediately when the ongoing export completes. This
+  behavior SHOULD be chosen when the next batch export was triggered by `ForceFlush`.
```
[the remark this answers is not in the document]
dyladan: I disagree. If maxExportBatch size is reached, all spans in the queue are attempted to be exported in one batch. Any spans which are in the queue when that export completes are new spans. Your suggestion would make those export immediately, and possibly trigger many immediate exports all in a row if new spans are arriving all the time.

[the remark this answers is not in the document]
blumamir: But this bullet starts with "Export the next batch immediately when the ongoing export completes", isn't it what we want?

The text is read so that if we have `maxExportBatchSize`, then implementation can choose the second bullet which is to delay export for `scheduledDelayMillis`. That collides with the PR description saying "BSP should export whenever the queue reaches the batch size"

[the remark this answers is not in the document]
dyladan: If you trigger an export by reaching the max batch size, then all spans should be exported in a single batch. When that export ends, the queue should be empty again (or near empty if a couple new spans have arrived). If the queue completely fills during that time, another batch MAY start immediately, but that would be a sign of poor tuning of queues.

specification/trace/sdk.md
```diff
@@ -579,17 +579,41 @@ This is an implementation of the `SpanProcessor` which create batches of finishe
 spans and passes the export-friendly span data representations to the
 configured `SpanExporter`.
 
+The processor SHOULD export a batch when any of the following happens AND
+there is not an export currently in progress:
```
MSNev, line 583: > AND there is not an export currently in progress

Why? While I understand that we want to limit the number of outbound connections, there are exceptions where this will block and cause issues. One seems to be restrictive... I would prefer to see this change to reference a `maxConnections` (defaulting to at least 2) but thats probably another PR.

Additionally, during app/client (page) unload, if an export is already in flight and the page wants to "terminate" this will cause any spans triggered as part of the "unload" to be dropped...

specification/trace/sdk.md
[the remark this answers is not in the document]
blumamir: > When that export ends, the queue should be empty again (or near empty if a couple new spans have arrived)

If it happens that the queue reaches the max batch size while there is an active export, then we want to initiate a new export with this full batch right as the previous one completed, right?

That fulfills the first part "The processor SHOULD export a batch when any of the following happens". Then when reading the second part "If one of the above events occurs during an ongoing export the processor MAY do either of the following".

According to this wording, implementation MAY in this case choose the second option which is "Export the next batch `scheduledDelayMillis` after the ongoing export completes". This doesn't feel like the intention of eager exporting, but it is possible that I am reading this incorrectly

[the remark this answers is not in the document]
MSNev: > When that export ends, the queue should be empty again

I would expect that at the point of committing to send a "batch" (ie. when it's getting packaged / serialized / sent) is when the "waiting" queue is empty (or near if the max payload size is exceeded), not just on receiving a response, otherwise you need to manage a single queue which has a mixture of waiting / sent that has to be processed when the response is received. There, is of course some complications around retries if the request fails as this would be dumping the "failed" spans back into the "waiting" queue (or a "retry" queue)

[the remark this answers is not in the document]
dyladan: I mixed up the previous batch and next batch. Amir was correct here. If there is an export in progress and the queue fills, it should be exported ASAP.

specification/trace/sdk.md
dyladan: This is already specified elsewhere and changing it is out of scope for this PR

>Export() will never be called concurrently for the same exporter instance.

_https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#exportbatch_

MSNev: It also states `Depending on the implementation the result of the export may be returned to the Processor not in the return value of the call to Export() but in a language specific way for signaling completion of an asynchronous task`

So this is not as clear cut as the spec states that the `Export(batch)` function won't be called concurrently, but it's also defined to return `synchronously` with the `ExportResult` of `Success`/ `Failure`. So while in a multi-threaded environment this means that the call to `Export(batch)` is not thread-safe it does not limit that you can't call `Export(batch)` again because the exporter is asynchronously sending the request, it just states that you can't call `Export(batch)` again if any previous `Export(batch)` has not returned (ie. from multiple threads).

In fact is even states ` it does not mean that the task of exporting can not be done concurrently.`...

So having `AND there is not an export currently in progress` is going against the explicit existing specification about having implementation / language specific concurrency.

specification/trace/sdk.md
```diff
@@ -579,17 +579,41 @@ This is an implementation of the `SpanProcessor` which create batches of finishe
...
+  export timer ends or previous batch completes.
+- The queue contains `maxExportBatchSize` or more spans.
+- `ForceFlush` is called.
+
+In any of the above cases the processor MUST NOT begin an export if another
+export is in progress. If one of the above events occurs during an ongoing export
+the processor MAY do either of the following:
+
+- Export the next batch immediately when the ongoing export completes. This
+  behavior SHOULD be chosen when the next batch export was triggered by reaching `maxExportBatchSize` spans in the queue or when`ForceFlush` is called.
```
MrAlias, line 598:
```suggestion
  behavior SHOULD be chosen when the next batch export was triggered by reaching `maxExportBatchSize` spans in the queue or when `ForceFlush` is called.
```

MrAlias (approved):

specification/trace/sdk.md
dyladan: Interesting. Looks like that was changed (somewhat) recently https://github.com/open-telemetry/opentelemetry-specification/pull/2452

In that case would it be better to say "AND Export() has not yet returned" and link to the relevant spec there for additional concurrency details?

MSNev: Yep (including a link to the section so it's not just `has not returned`), or alternatively just drop that portion as "technically" it's already covered by the wording of the Export(batch)

dyladan: [7290f22](https://github.com/open-telemetry/opentelemetry-specification/pull/3024/commits/7290f22231c4abd5ba7ce82f366e552f769b5f5b)

dyladan:
I think this can be merged

reyang:
I'll merge it after the CI finished.

reyang:
The `markdown-link-check` CI failed due to network latency fluctuation...

PR #3029, opened 2022-12-12
Suggest column # as an extra source code attribute

jordigh:
# Changes

In some languages such as Javascript or Haskell, it's common to have multiple lambdas/anonymous functions defined on the same line. 

`function myEval(code){return eval(code)}function handleJSONP(a){return a}(function(){function a(b){var c=myEval(b)}})();`

Without a column number, it's impossible to accurately pinpoint which of the anonymous functions in the same file and the same line we are talking about. A column attribute is necessary to fully specify which unit of code we are interested in.

### DIFF
--- a/specification/trace/semantic_conventions/span-general.md
+++ b/specification/trace/semantic_conventions/span-general.md
@@ -315,4 +315,5 @@ about the span.
 | `code.namespace` | string | The "namespace" within which `code.function` is defined. Usually the qualified class or module name, such that `code.namespace` + some separator + `code.function` form a unique identifier for the code unit. | `com.example.MyHttpService` | Recommended |
 | `code.filepath` | string | The source code file name that identifies the code unit as uniquely as possible (preferably an absolute file path). | `/usr/local/MyApplication/content_root/app/index.php` | Recommended |
 | `code.lineno` | int | The line number in `code.filepath` best representing the operation. It SHOULD point within the code unit named in `code.function`. | `42` | Recommended |
+| `code.column` | int | The column number in `code.filepath` best representing the operation. It SHOULD point within the code unit named in `code.function`. | `16` | Recommended |
 <!-- endsemconv -->
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ release.
 
 ### Semantic Conventions
 
+- Add `code.lineno` source code attribute
+  ([#3029](https://github.com/open-telemetry/opentelemetry-specification/pull/3029))
+
 ### Compatibility
 
 ### OpenTelemetry Protocol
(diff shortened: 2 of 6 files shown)


MrAlias (approved):

yurishkuro (approved):

arminru (commented):
@jordigh Please update the backing YAML files defined in https://github.com/open-telemetry/opentelemetry-specification/tree/main/semantic_conventions as instructed by the README in that folder, this will also make the merge checks pass.

Also, please update the https://github.com/open-telemetry/opentelemetry-specification/blob/main/CHANGELOG.md with your addition.

specification/trace/semantic_conventions/span-general.md
```diff
@@ -315,4 +315,5 @@ about the span.
 | `code.namespace` | string | The "namespace" within which `code.function` is defined. Usually the qualified class or module name, such that `code.namespace` + some separator + `code.function` form a unique identifier for the code unit. | `com.example.MyHttpService` | Recommended |
 | `code.filepath` | string | The source code file name that identifies the code unit as uniquely as possible (preferably an absolute file path). | `/usr/local/MyApplication/content_root/app/index.php` | Recommended |
 | `code.lineno` | int | The line number in `code.filepath` best representing the operation. It SHOULD point within the code unit named in `code.function`. | `42` | Recommended |
+| `code.column` | int | The column number in `code.filepath` best representing the operation. It SHOULD point within the code unit named in `code.function`. | `16` | Recommended |
```
Oberon00, line 318: Does a tab char count as 8 or 1 columns? Or should we leave that undefined?

jordigh: Hm, I believe most tools count tabs as 1 character. Node.js does this. Tab stops are kinda impossible to predict, aren't they? If you have a mix of tabs and spaces, each tab might be less than 8 characters wide depending on where it is relative to the next tab stop.

It seems far better to just leave the value "column number" somewhat vague and particular to each programming language implementation.

svrnm: Vim & VSCode both count tabs based on the set Tab Size

Oberon00:
I don't think you should change anything there. There is only a warning printed which can be ignored.

jordigh:
> I don't think you should change anything there. There is only a warning printed which can be ignored.

It return nonzero and  was showing as a failed build on Github, gotta appease the build gods.

https://github.com/open-telemetry/opentelemetry-specification/actions/runs/3680446681

carlosalberto:
@Oberon00 Any last concern?

Oberon00:
@carlosalberto No concerns with merging the PR, please go ahead!

(but it *is* a bit concerning that this error from the convention generator went undetected. Maybe that build job depends on somebody touching the YAML files and nobody did so since the last convention generator update)

Oberon00 (approved):

Oberon00:
Hi, I think this should be merged instead of going stale

PR #3031, opened 2022-12-13
Reorganize specification related to OTLP file exporter

### LINKED ISSUE #3003: Remove "Built-in exporters" section in Log SDK
The log SDK spec contains the skeleton for a "built-in exporters" section. It only contains things specific to OTLP.

https://github.com/open-telemetry/opentelemetry-specification/blob/150451d4257be3d4673c7fcbf50fca7fb557d346/specification/logs/sdk.md#L369-L385

At this time, the trace SDK does not contain a trace-specific OTLP exporter spec - only a Jaeger and Zipkin exporter spec - see: https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/trace/sdk_exporters. It would seem that the [OTLP exporter spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md) sufficiently covers most things for all signals. Metrics is a small exception as there are [metric-specific OTLP configuration](https://github.com/open-telemetry/opentelemetry-specification/blob/150451d4257be3d4673c7fcbf50fca7fb557d346/specification/metrics/sdk_exporters/otlp.md) options.

It seems there could be a discussion of a greater reorganization of OTLP exporter related specs. For example, maybe centralizing specifics for all signals into the main OTLP exporter spec.

For now, though, I recommend we remove this section from the log SDK spec. We can introduce an `sdk_exporters` directory under the log specification or decide to reorganize things at a later date if necessary.

alanwest:
Fixes #3003 
Contributes to #2911

## Changes

This PR moves and reorganizes parts of the specification relevant to exporting OTLP data to a file.

* Removes the "Build-in exporters" section from the Log SDK. The OTLP exporter has its own specification covering all signals. The part regarding a file exporter is not unique to logs. Exporting OTLP data to a file is also applicable to trace and metric data.
* Moves the file describing OTLP JSON serialization under the `protocol` subdirectory. As discussed in the Logs SIG (12/7/2022), this file provides a foundation for further progress on specifying an OTLP file exporter.


jack-berg (approved):
👍  Java already has exporters that log in OTLP JSON format (e.g. [OtlpJsonLoggingSpanExporter](https://github.com/open-telemetry/opentelemetry-java/blob/main/exporters/logging-otlp/src/main/java/io/opentelemetry/exporter/logging/otlp/OtlpJsonLoggingSpanExporter.java)).

specification/logs/sdk.md
```diff
@@ -365,21 +362,3 @@ return a Failure result.
...
-
-#### OTLP Exporter
-
-Exports to an OTLP network destination via OTLP/gRPC or OTLP/HTTP.
-
-#### OTLP File exporter
-
-Writes to a file or stdout in either OTLP JSON or OTLP Protobuf binary format.
-
-![Logging to File](img/otlp-file.png)
```
tigrannajaryan, line 381: Is this diagram useful? Perhaps we preserve it somewhere?

tigrannajaryan (approved):

specification/logs/sdk.md
alanwest: Good question. Instead of moving it to the OTLP file exporter spec, I removed it because the diagram depicts logging specific components that did not seem applicable inside of a spec describing export of all signals to a file. 

Furthermore, the diagram seemed redundant. I think there are two important concepts it depicts:
1) The relationship of a logging exporter with respect to other log SDK components. This is already depicted [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md#logrecordprocessor).
2) A depiction of log appenders and their interaction with the SDK. This too is described and depicted [here](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/logs#direct-to-collector).

arminru (approved):

Issue #437, opened 2020-02-04
Update Binary format in the Specification


carlosalberto:
As https://github.com/open-telemetry/opentelemetry-specification/pull/426 is temporarily removing Binary format, we need to make sure Binary is re-added and updated properly with the OTEP 66 updates.

Assigning for now to @MikeGoldsmith but we can discuss who can tackle this item going forward.

bogdandrutu:
Added @carlosalberto to the assignees since he promised to add it back :)

carlosalberto:
That works :)

Oberon00:
Can we assign a milestone for this issue? Any release that wants to offer a full OpenTracing or OpenCensus compatibility layer will need this feature.

Oberon00:
Also, we need that feature at Dynatrace and will have to stay on an older version of the Java API until this is fixed.

carlosalberto:
@Oberon00 So the milestone would be 0.5.0 - for which we don't have a date for yet (other than being the next important milestone). Maybe we need to discuss this in the next SIG/community call, to plan what/how happens after beta ;)

andrewhsu:
Talked with @carlosalberto today and based on the history of the intention of binary format (it looks like it was only desired to be temporarily removed), looks like this one is important to get in sooner because it can cause backward breaking change. Therefore, changing priority to p1.

andrewhsu:
from the spec issue triage mtg today, discussed this and moving to after ga

annabokhan:
[Documentation](https://opentelemetry.io/docs/specs/otel/context/api-propagators/) states that the binary propagator support is upcoming, however this thread has been stale for almost 4 years. Is this still planned?

amalgawa:
+1 to above question. It has been 4 years, since no activity. If there are any prior investigations why it was no longer pursued, I can probably help implement it.

svrnm:
@carlosalberto what is the status of this one?

esemeniuc:
+1 need to see this

PR #3035, opened 2022-12-14
Add OpenCensus propagators as allowed in OTel repositories

dashpole:

Addresses https://github.com/open-telemetry/opentelemetry-specification/pull/3015#discussion_r1047817750.  We don't currently list OpenCensus propagators anywhere in our specification, even though they have been implemented (e.g. [go's BinaryFormat propagator](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/propagators/opencensus#opencensus-binary-propagation-format)).

This does not require implementing these propagators, but adds them to the list of propagators allowed in core repositories.  My primary motivation is to document them somewhere in the specification for reference.

cc @jsuereth @bogdandrutu @jack-berg @reyang

### DIFF
--- a/specification/context/api-propagators.md
+++ b/specification/context/api-propagators.md
@@ -346,6 +346,10 @@ as OpenTelemetry extension packages:
 * [OT Trace](https://github.com/opentracing?q=basic&type=&language=). Propagation format
   used by the OpenTracing Basic Tracers. It MUST NOT use `OpenTracing` in the resulting
   propagator name as it is not widely adopted format in the OpenTracing ecosystem.
+* [OpenCensus BinaryFormat](https://github.com/census-instrumentation/opencensus-specs/blob/master/encodings/BinaryEncoding.md#trace-context).
+  Propagation format used by OpenCensus, which describes how to format the span context
+  into the binary format, and does not prescribe a key. It is commonly used with
+  OpenCensus gRPC using the `grpc-trace-bin` propagation key.
 
 Additional `Propagator`s implementing vendor-specific protocols such as AWS
 X-Ray trace header protocol MUST NOT be maintained or distributed as part of


dashpole:
cc @carlosalberto as well

I stumbled on https://github.com/open-telemetry/opentelemetry-specification/issues/437.  It looks like the w3c spec is still a draft.  I would like to propose that we allow re-adding the binary propagator purely as a tool to facilitate migrations from OpenCensus, without tackling the introduction of a standardized w3c binary propagator.

carlosalberto (approved):

carlosalberto:
Approving but we should do a follow up (requirement?) issue to clarify the binary propagation in the related section (for which an OC binary propagator in an additional language should be implemented, as a way to exercise it, etc)

jmacd (approved):

carlosalberto:
cc @bogdandrutu

specification/context/api-propagators.md
```diff
@@ -346,6 +346,13 @@ as OpenTelemetry extension packages:
 * [OT Trace](https://github.com/opentracing?q=basic&type=&language=). Propagation format
   used by the OpenTracing Basic Tracers. It MUST NOT use `OpenTracing` in the resulting
   propagator name as it is not widely adopted format in the OpenTracing ecosystem.
+* [OpenCensus TextFormat](https://www.javadoc.io/doc/io.opencensus/opencensus-api/0.16.1/io/opencensus/trace/propagation/TextFormat.html).
```
MrAlias, line 349: This link is for the Java implementation of the format. Is there a more language agnostic link that can be added here instead that will explain what the format and protocol are?

dashpole: It was the best description I could find.  It doesn't appear to be documented on the opencensus website (I would expect it here: https://opencensus.io/advanced-concepts/context-propagation/).  I'm not sure if the publishing process for the website still works, but we might be able to add it here: https://github.com/census-instrumentation/opencensus-website/blob/5970b0c5598460f0a22ec994826b7a0962352490/content/advanced-concepts/context-propagation.md

dashpole: Actually, it looks like this is just the w3c traceparent propagator:https://github.com/census-instrumentation/opencensus-java/blob/a37129e50172f713f68a58b22027c5147925a526/impl_core/src/main/java/io/opencensus/implcore/trace/propagation/TraceContextFormat.java#L44

I'll remove this item.

MrAlias (approved):

jack-berg (approved):

reyang (approved):

PR #3041, opened 2022-12-15
Mark exponential histogram as stable

### LINKED ISSUE #2832: Stabilize exponential histograms
Exponential histograms have implementations in Java, Dotnet, Go, and Python.

There's still the idea of the [zero threshold](https://github.com/open-telemetry/opentelemetry-specification/pull/2665), which needs to be defined in the proto and likely needs to be configurable in the SDK in some fashion. However, this can be added in a backwards compatible way.

If there are no other issues, we should stabilize the [exponential histogram data model](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md#exponentialhistogram) and [aggregation](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#exponential-bucket-histogram-aggregation).

jack-berg:
Resolves #2832.


specification/metrics/sdk.md
```diff
@@ -352,10 +352,7 @@ The SDK MUST provide the following `Aggregation` to support the
 - [Sum](./sdk.md#sum-aggregation)
 - [Last Value](./sdk.md#last-value-aggregation)
 - [Explicit Bucket Histogram](./sdk.md#explicit-bucket-histogram-aggregation)
-
-The SDK MAY provide the following `Aggregation`:
-
-- [Exponential Bucket Histogram Aggregation](./sdk.md#exponential-bucket-histogram-aggregation)
+- [Exponential Bucket Histogram](./sdk.md#exponential-bucket-histogram-aggregation)
```
jack-berg, line 355: This says SDKs MUST support exponential bucket histograms. Perhaps it should be a "SHOULD" instead of a "MUST"?

carlosalberto (approved):

pirgeo:
Is this pending on https://github.com/open-telemetry/opentelemetry-specification/pull/3017?

jack-berg:
I think #3017 is nice to have, but not blocking since its a backwards compatible addition.

specification/metrics/sdk.md
pirgeo: Do you mean updating line 346 to say SHOULD instead? Then it would be true for all Aggregations, right? That would give the individual languages an out if one of these Aggregations cannot be implemented or would be very awkward to handle.

jack-berg: I mean something like:
```
The SDK MUST provide the following `Aggregation` to support the
[Metric Points](./data-model.md#metric-points) in the
[Metrics Data Model](./data-model.md).

- [Drop](./sdk.md#drop-aggregation)
- [Default](./sdk.md#default-aggregation)
- [Sum](./sdk.md#sum-aggregation)
- [Last Value](./sdk.md#last-value-aggregation)
- [Explicit Bucket Histogram](./sdk.md#explicit-bucket-histogram-aggregation)

The SDK SHOULD provide the following `Aggregation`:

- [Exponential Bucket Histogram](./sdk.md#exponential-bucket-histogram-aggregation)
```
So SDKs could choose not to implement exponential bucket histogram, but must implement all the others.

pirgeo: That makes sense to me

carlosalberto:
Circling back on Jack's last question: should the expo aggregation be a SHOULD or a MUST for each language? 

cc @jsuereth @jmacd @reyang

jmacd (approved):

specification/metrics/sdk.md
jack-berg: FYI, unless there's some pushback I'm inclined to keep this as it currently stands, where SDKs MUST support exponential bucket histograms.

specification/metrics/data-model.md
```diff
@@ -519,7 +519,7 @@ so that worst-case error is within their error tolerance.
 
 ### ExponentialHistogram
 
-**Status**: [Experimental](../document-status.md)
+**Status**: [Stable](../document-status.md)
```
reyang, line 522: What's the plan for preventing the "taint mode"? Do we want to introduce something like zero threshold and infinity threshold, and if yes, do we consider this as a non-blocker (which can be added after Stable release)?

specification/metrics/sdk.md
pirgeo: I don't feel strongly about this, I think either way is fine. The exponential histogram is not required in the default aggregations, and therefore an implementation without it can still be spec compliant. I think using SHOULD might make it easier for languages with a smaller developer base to get to spec compliance more easily (implementing exponential histograms might not be easy for some languages, especially if you are not an expert on the topic). I realize that that is a pretty weak argument to go with SHOULD, and ultimately I feel like it does not make a difference unless I am missing something obvious.

specification/metrics/data-model.md
jack-berg: The data model already [defines](https://github.com/open-telemetry/opentelemetry-specification/issues/2288) a zero threshold field, though it doesn't exist yet in the proto (PR to add it [here](https://github.com/open-telemetry/opentelemetry-proto/pull/441)). 

These are non blocking IMO. There's currently a good default definition for zero threshold. It could be made configurable in a backwards compatible way. Although infinity threshold doesn't yet exist, it can be added in a backwards compatible way with a reasonable default.

MrAlias: Before this gets too far, can we refer to this as "tainted" mode instead of "taint" mode please.

jsuereth:
I'm ok with SHOULD or MUST.  For SDKs that need help implementing exponential histogram, happy to pitch in.

specification/metrics/data-model.md
jmacd: @reyang 

I can't tell whether this is a request to maintain a range limit on histogram measurements, in which case say the SDK would be responsible for returning errors when the data lies outside a configured range. 

OR, 

whether this is a request to reset the data structure when it contains extreme values, in which case say we would configure a "target scale" and every time the data structure's scale falls below the target, the StartTimestamp resets automatically? This would allow you to correctly convey the extreme values and then reset the data structure to continue recording sufficient resolution.

I prefer the second approach. I think both approaches can be done in a backwards compatible way.

reyang: Resolving due to https://github.com/open-telemetry/opentelemetry-specification/pull/3041#discussion_r1065887844.

specification/metrics/data-model.md
```diff
@@ -519,7 +519,7 @@ so that worst-case error is within their error tolerance.
 
 ### ExponentialHistogram
```
reyang, line 520: Do we want to call this Exponential Histogram or something more specific (and reserve the general term for potential future additions)? The actual exponential histogram is a much wider topic, and the thing we're covering here is just a special form which is quite restricted (e.g. "base 2").

jmacd: I am comfortable using the term "Base2ExponentialHistogram" (in specifications) and "base2_exponential_histogram" (in environment variables).

jack-berg: While I don't mind the name `Base2ExponentialHistogram`, I think we should keep it as `ExponentialHistogram`:

- The proto already refers to [ExponentialHistogram](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L182). While message names are allowed to change, [field names cannot](https://github.com/open-telemetry/opentelemetry-proto/blame/main/README.md#L50), and so we would have an inconsistency between field name and message name.
- In java we have a stable [MetricDataType](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/MetricDataType.java) enum which refers to `ExponentialHistogram`. We've kept virtually all the exponential histogram surface area in experimental packages, but added this enum option to allow for experimenting and because the naming `exponential_histogram` was referenced in the stable proto. 
- While a base2 exponential histogram isn't the only type of exponential histogram, it is the first we've added. Thus, it feels reasonable to require that future exponential histograms disambiguate by adding some sort of descriptive prefix. If we reach such a point, we might add language like "ExponentialHistogram is a base 2 histogram aggregation". To me this feels similar to "log" referring to log base 10 unless otherwise specified. The most used version of the thing (or the first) gets to omit the prefix.

reyang: > * The proto already refers to [ExponentialHistogram](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L182). While message names are allowed to change, [field names cannot](https://github.com/open-telemetry/opentelemetry-proto/blame/main/README.md#L50), and so we would have an inconsistency between field name and message name.

Proto types can be augmented to support other types of exponential histograms, which won't be a showstopper.

> While a base2 exponential histogram isn't the only type of exponential histogram, it is the first we've added. Thus, it feels reasonable to require that future exponential histograms disambiguate by adding some sort of descriptive prefix.

It'll be very confusing for a user coming from a different background to spend time reading the docs and realizing that "in order to use base 10 exponential histograms, I have to specify explicit buckets for now rather than using the OpenTelemetry exponential histogram". I guess it'll be helpful to avoid the confusion upfront. One possible way is to add a clear disclaimer saying "hey, this OpenTelemetry Exponential Histogram section is not covering the common sense exponential bucket histograms, ... more explanations ..., you should read the Explicit Bucket Histogram section if you're looking for something more general".

jack-berg: > Proto types can be augmented to support other types of exponential histograms, which won't be a showstopper.

Oh I see, you're thinking that both a base 2 and a future base 10 exponential histogram aggregation could both utilize the same `ExponentialHistogram` protobuf message. If that's the case, then I don't have to worry about my enum name being incorrect and I don't have a preference for the name - would be happy with `Base2ExponentialHistogram` or `ExponentialHistogram`.

reyang: > > Proto types can be augmented to support other types of exponential histograms, which won't be a showstopper.
> 
> Oh I see, you're thinking that both a base 2 and a future base 10 exponential histogram aggregation could both utilize the same `ExponentialHistogram` protobuf message.

Yup.

jack-berg: Updated aggregation name in latest commit.

jack-berg:
I've changed the aggregation name to "Base2 Exponential Bucket Histogram Aggregation" per this [conversation](https://github.com/open-telemetry/opentelemetry-specification/pull/3041#discussion_r1066038209). 

Note that the metric point defined in `data-model.md` is still named [ExponentialHistogram](https://github.com/open-telemetry/opentelemetry-specification/pull/3041/files#diff-6c97e08ad12fe1538de64c06f0cab436e9115287ea35183b88dbd1df77ce0bc5L520), since the same point could be reused with a base 10 exponential histogram aggregation.

specification/metrics/data-model.md
jack-berg: Just realized that we could keep the name of the aggregation `Exponential Bucket Histogram Aggregation`, and accommodate a future base 10 option by making the base a configurable option of a single exponential bucket histogram aggregation. 

This might be a better alternative because it doesn't result in disjointed naming where `data-model.md` references "ExponentialHistogram" and `sdk.md` references "Base2 Exponential Bucket Histogram Aggregation".

CHANGELOG.md
```diff
@@ -21,6 +21,12 @@ release.
 
 - Rename built-in ExemplarFilters to AlwaysOn, AlwaysOff and TraceBased.
   ([#2919](https://github.com/open-telemetry/opentelemetry-specification/pull/2919))
+- Rename exponential histogram aggregation to base 2 exponential histogram
+  aggregation. Rename "OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION"
+  value from "exponential_bucket_histogram" to
+  "base2_exponential_bucket_histogram". Mark exponential histogram data model and
+  base2 exponential histogram aggregation aggregation as stable.
```
MrAlias, line 28:
```suggestion
  base2 exponential histogram aggregation as stable.
```

specification/metrics/data-model.md
jack-berg: Just want to bump this last message @jmacd @reyang.

Essentially, if we want to reuse the ExponentialHistogram data point and proto message for a future possible base 10 variation, wouldn't we also want to reuse the Exponential Bucket Histogram Aggregation, and just add a config option to allow the base to be 10?

reyang: Maybe yes. However, I think the concept such as MaxSize and ZeroThreshold would only exist for "smart histograms" that automatically adjust the scale, and from the programming interface we should make it clear that mathematically it is a special form/subset of exponential histogram, to avoid confusion.

jack-berg: A base 10 histogram could also end up with an adjustable scale.

I'm happy enough renaming the aggregation and addressing this later if needed.

PR #3047, opened 2022-12-19
Add semantic conventions for log record ID

### LINKED ISSUE #597: Decide if we need log record ID field
See discussion here https://github.com/open-telemetry/oteps/pull/97#discussion_r421865344

Log Data model currently does not contain a field that identifies the log record, however having such field may be desirable.

We need to decide if we want such a field and if yes how it will be represented.

This can be added to the Data Model as an optional field, which will ensure we are not breaking the data model.

svrnm:
Fixes #597

## Changes

Please provide a brief description of the changes here.

- Add a section for "generic attributes" to the log semconv
- Add an attribute `log_record.id` making use of ULID as discussed in #597

Some additional notes:
- I kept the PR small, so I left out some other potential attributes, e.g. something for pre-existing ID (like windows event logs) or for storing the used logging/eventing system or even something like a "signature" that might be worth discussing, etc.
- I followed the structure of "generic attributes" from the spans semconv
- I took some of the existing wording from #597 & https://github.com/open-telemetry/oteps/pull/97#discussion_r421865344 to describe the field


djaglowski:
> * Add an attribute `log_record.id` making use of ULID as discussed in [Decide if we need log record ID field #597](https://github.com/open-telemetry/opentelemetry-specification/issues/597)

Should we call this `log.record.id`, to match the pattern used by other log attributes?

svrnm:
> > * Add an attribute `log_record.id` making use of ULID as discussed in [Decide if we need log record ID field #597](https://github.com/open-telemetry/opentelemetry-specification/issues/597)
> 
> Should we call this `log.record.id`, to match the pattern used by other log attributes?

works for me, I just picked `log_record.id` to start with something but I don't have a strong opinion for/against it.

tigrannajaryan:
> > * Add an attribute `log_record.id` making use of ULID as discussed in [Decide if we need log record ID field #597](https://github.com/open-telemetry/opentelemetry-specification/issues/597)
> 
> Should we call this `log.record.id`, to match the pattern used by other log attributes?

I agree. We already have `log.` as a top-level namespace, let's use it. `log.record_id` is another alternate. If we think we need to add anything else about the record then `log.record.id` is a better choice.

semantic_conventions/logs/log-general.yaml
```diff
@@ -0,0 +1,17 @@
+groups:
```
joaopgrassi, line 1: Maybe call the file `general.yaml` since it's already under the logs folder? That would be consistent with the one in the https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/trace/general.yaml `trace` folder

semantic_conventions/logs/log-general.yaml
```diff
@@ -0,0 +1,17 @@
...
+      The attributes described in this section are rather generic. They may be used in any Log Record they apply to.
+    attributes:
+      - id: id
+        type: string
+        requirement_level: recommended
+        brief: >
+          A unique identifier for the Log Record.
+        note: >
+          MUST be an [Universally Unique Lexicographically Sortable Identifier (ULID)](https://github.com/ulid/spec).
+          If a value is provided other log records, with the same value, will be considered to be duplicates and can be removed safely.
```
joaopgrassi, line 14:
```suggestion
          If an id is provided, other log records with the same id will be considered to be duplicates and can be removed safely.
```
Nit: I feel using `id` makes things more clear than just `value`

semantic_conventions/logs/log-general.yaml
```diff
@@ -0,0 +1,17 @@
...
+          This means, that two distinguishable log records MUST have different values.
+          However in the case of duplication it is not guaranteed that duplicated log records have the same value.
```
joaopgrassi, line 16:
```suggestion
          However, in the case of duplication, it is not guaranteed that duplicated log records have the same id.
```

semantic_conventions/logs/log-general.yaml
```diff
@@ -0,0 +1,17 @@
...
+        examples: ["click", "exception"]
```
joaopgrassi, line 17:
```suggestion
        examples: ["01ARZ3NDEKTSV4RRFFQ69G5FAV"]
```
The examples are wrong, right?

specification/logs/semantic_conventions/README.md
```diff
@@ -4,6 +4,7 @@
 
 The following semantic conventions for logs are defined:
 
+* [General](log-general.md): General semantic attributes that may be used in describing Log Records.
```
joaopgrassi, line 7:
```suggestion
* [General](general.md): General semantic attributes that may be used in describing Log Records.
```
If you go ahead with changing the file name

semantic_conventions/logs/log-general.yaml
svrnm: 👍

semantic_conventions/logs/log-general.yaml
svrnm: yeah, I thought I replaced that, not sure what happened here.

specification/logs/semantic_conventions/README.md
svrnm: done

semantic_conventions/logs/log-general.yaml
svrnm: done

jack-berg (approved):

specification/logs/semantic_conventions/general.md
```diff
@@ -0,0 +1,26 @@
...
+<!-- tocstop -->
+
+## General log identification attributes
+
+These attributes may be used for identifying a Log Record.
+
+<!-- semconv log.record -->
+| Attribute  | Type | Description  | Examples  | Requirement Level |
+|---|---|---|---|---|
+| `log.record.id` | string | A unique identifier for the Log Record. [1] | `click`; `exception` | Recommended |
```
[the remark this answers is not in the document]
svrnm: I am not sure about this, to be honest. If the log SDKs take care of it, it only has to be implemented once (wheres there might be multiple appenders per language, right?).

djaglowski (approved):

semantic_conventions/logs/general.yaml
```diff
@@ -0,0 +1,17 @@
+groups:
+  - id: log.record
+    prefix: log.record
```
scheler, line 3: Is there a way we can avoid the `log` prefix here? That way, the attribute can be used for events as well.

specification/logs/semantic_conventions/general.md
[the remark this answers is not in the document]
joaopgrassi: This is a good point. I think it would be a stretch to think appenders would change to include this. Some can be "enhanced" via config files and extra plugins to add more things, but that falls then on the user's lap. 

So I'd agree that it would make sense the SDK populates it if empty, like @jack-berg mentioned.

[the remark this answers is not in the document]
svrnm: should this be included in the document somewhere?

semantic_conventions/logs/general.yaml
svrnm: I am open to it, what would you suggest, just `record.id`?

specification/logs/semantic_conventions/general.md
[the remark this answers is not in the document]
jack-berg: Probably belongs in the [SDK](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md) document somewhere. Not sure there's a subsection it fits into at the moment. I'd think it would be a side affect of the the SDK's implementation of the [Logger#emit](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/api.md#emit-logrecord) method, which would need to occur before any LogProcessor is called.

I'll say that its odd to have the SDK be performing logic which incorporates the semantic conventions. The SDKs do have logic for top level fields (i.e. default span status if none is specified). The only example I can think of where they interact with the semantic conventions is the default resource, which includes a default `service.name`. Also the [Record Exception](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#record-exception) API.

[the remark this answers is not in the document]
svrnm: What would be the alternative to remove that oddness? Have the appender create that ID? Make that ID a field of the LogRecord?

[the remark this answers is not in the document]
svrnm: following up on this, after talking with @scheler: 

Based on the use cases the SDK is a good place for generating that ID. I agree that it might be an odd place, but similar to service.name I see this as a kind of "default" resource (not mandatory, but ideal to have), since it's going to help with identification down the line.

I can add another section here or in the SDK documentation to call that out, if it is the expected behavior.

cc @open-telemetry/specs-logs-approvers

semantic_conventions/logs/general.yaml
svrnm: If I understood you correctly @scheler this can be closed.

specification/logs/semantic_conventions/general.md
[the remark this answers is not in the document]
svrnm: based on the SIG meeting discussion today, the id is optional and can be populated as needed (SDK is a good place with an option, but it could even be a collector processor that populates log records before sending them off to the backend)

svrnm:
Updating with the feedback from todays SIG call:

* Made the attribute optional
* Removed the requirement for ULID

semantic_conventions/logs/general.yaml
scheler: Yes. Now that we have an Event API/SDK separate from Log API/SDK, events can have separate semantic conventions. So, for events we can have `event.id` if required in future.

specification/logs/semantic_conventions/general.md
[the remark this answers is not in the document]
svrnm: @jack-berg can I close this conversation?

reyang:
@jmacd @jsuereth do you see similar need for metrics de-dupe? (I don't, I think metrics de-dupe is already covered, but would want to hear from you)

I feel for traces the trace id + span id combination is good enough for de-dupe, and for metrics the stream identity + resource + time should be good enough.

semantic_conventions/logs/general.yaml
```diff
@@ -0,0 +1,19 @@
...
+    brief: >
+      The attributes described in this section are rather generic. They may be used in any Log Record they apply to.
+    attributes:
+      - id: id
```
reyang, line 7: Do we want to consider the name "uid" (I'm bad at naming, but you get the idea)? The term "id" might lead to something like https://www.myeventlog.com/search/find or https://www.ibm.com/docs/en/iis/11.3?topic=stage-logging-messages-custom-ids

reyang (approved):
LGTM with a non-blocking suggestion.

semantic_conventions/logs/general.yaml
svrnm: so the idea is to encode in the name that this is a unique id and not something that identifies anything else like an eventtype, etc.?

svrnm: @open-telemetry/specs-logs-approvers any preferences here?

djaglowski: `log.record.uid` seems reasonable to me. It's more specific without taking away from discoverability or functionality.

specification/logs/semantic_conventions/general.md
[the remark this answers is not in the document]
svrnm: @jack-berg any thoughts on this?

semantic_conventions/logs/general.yaml
svrnm: updated it to log.recrod.uid accordingly

specification/logs/semantic_conventions/general.md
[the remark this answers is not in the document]
jack-berg: Sorry this got lost in the shuffle. Seems incomplete without specifying if / when the SDK should populate this field, but I don't feel strong enough to block based on that.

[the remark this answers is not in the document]
svrnm: > Sorry this got lost in the shuffle. Seems incomplete without specifying if / when the SDK should populate this field, but I don't feel strong enough to block based on that.

Thanks. I will resolve this conversation for now, but I will follow up with some notes on how the field may be populated.

svrnm:
anything missing to get this merged? TY

djaglowski (approved):

specification/logs/semantic_conventions/general.md
```diff
@@ -0,0 +1,27 @@
...
+## General log identification attributes
+
+These attributes may be used for identifying a Log Record.
+
+<!-- semconv log.record -->
+| Attribute  | Type | Description  | Examples  | Requirement Level |
+|---|---|---|---|---|
+| `log.record.uid` | string | A unique identifier for the Log Record. [1] | `01ARZ3NDEKTSV4RRFFQ69G5FAV` | Optional |
+
+**[1]:** If an id is provided, other log records with the same id will be considered duplicates and can be removed safely. This means, that two distinguishable log records MUST have different values. However, in the case of duplication, it is not guaranteed that duplicated log records have the same id.
```
tigrannajaryan, line 25: >However, in the case of duplication, it is not guaranteed that duplicated log records have the same id.

I am not sure I understand this part. What "case of duplication" is this referring to? Duplication by whom, when and why?

svrnm: This last sentence is here to emphasis that uniqueness is only given in one direction:

The following statement is always true:
For any 2 log records, where the log.record.id is equal, the log records are equal

The following statement is not guaranteed:
For any 2 identical log records, the log.record.id is equal.

So whenever a log line is ingested 2 no matter where you are in the pipeline, the earlier you give it that id, the higher the chance is that you don't have a duplication

jack-berg (approved):
I still think we should define how these can be automatically added by the SDK. Opened #3277 with a suggestion on how we might do so with another built-in LogRecordProcessor. Can add in a followup.

svrnm:
> I still think we should define how these can be automatically added by the SDK. Opened #3277 with a suggestion on how we might do so with another built-in LogRecordProcessor. Can add in a separate PR.

Thank you.

specification/logs/semantic_conventions/general.md
svrnm: @tigrannajaryan do you think this needs to be described more precisely?

semantic_conventions/logs/general.yaml
```diff
@@ -0,0 +1,19 @@
+groups:
+  - id: log.record
+    prefix: log.record
+    brief: >
+      The attributes described in this section are rather generic. They may be used in any Log Record they apply to.
+    attributes:
+      - id: uid
+        type: string
+        requirement_level: optional
```
joaopgrassi, line 9:
```suggestion
        requirement_level: opt_in
```

Since https://github.com/open-telemetry/opentelemetry-specification/pull/3228 was merged, we need to adapt now.

joaopgrassi (approved):

specification/logs/semantic_conventions/general.md
tigrannajaryan: I feel that the current wording is (the particular sentence I quoted) confusing. Perhaps delete it and we can merge as is and if needed more details can be added later (I am not sure how much detail we need here).

svrnm: Yes, the sentence is superfluous, and if it's more confusing than helping, I'll remove it.

svrnm: Removed via https://github.com/open-telemetry/opentelemetry-specification/pull/3047/commits/186207c2675568051c15cf853531131fe93dda23

svrnm:
Anything missing before this can be merged? TY

djaglowski (approved):

tigrannajaryan (approved):

tigrannajaryan:
Thank you for the PR and for your patience @svrnm

PR #3048, opened 2022-12-19
OpenTracing Shim: Add Tracer.Close() operation.

carlosalberto:
Upon cleaning up the existing OT Shim for Java, realized we missed adding `Tracer.Close()`. 

I'd argue we are fine adding this, as:

1. Currently the support for Close() is undefined/no-op (the OpenTelemetry API doesn't expose this, after all)
2.  Among all OpenTracing languages, **only** Java and C++ implement this functionality.

cc @yurishkuro

### DIFF
--- a/specification/compatibility/opentracing.md
+++ b/specification/compatibility/opentracing.md
@@ -12,6 +12,7 @@
   * [Start a new Span](#start-a-new-span)
   * [Inject](#inject)
   * [Extract](#extract)
+  * [Close](#close)
 * [Span Shim and SpanContext Shim relationship](#span-shim-and-spancontext-shim-relationship)
 * [Span Shim](#span-shim)
   * [Get Context](#get-context)
@@ -157,6 +158,9 @@ OpenTracing and OpenTelemetry units.
 
 The API MUST return an OpenTracing `Span`.
 
+For OpenTracing languages implementing [Close](#close), this operation MUST
+return an OpenTracing no-op `Span` if the `Tracer` Shim has been closed.
+
 ### Inject
 
 Parameters:
@@ -204,6 +208,18 @@ return SpanContextShim(extractedSpanContext, extractedBaggage);
 Errors MAY be raised if either the `Format` is not recognized
 or no value could be extracted, depending on the specific OpenTracing Language API
 (e.g. Go and Python do, but Java may not).
+
+## Close
+
+Mark the `Tracer` Shim as closed, effectively preventing creation of new `Span` instances.
+Subsequent calls to [Start a new Span](#start-a-new-span) MUST return a no-op instance.
+
+Closing an already closed instance MUST NOT raise an error.
+
+Observe that not all OpenTracing languages implement this operation.
+
+Note: the underlying OpenTelemetry `Tracer` is not actually closed, as the OpenTelemetry API
+currently does not expose such functionality.
 
 ## Span Shim and SpanContext Shim relationship
 


specification/compatibility/opentracing.md
```diff
@@ -205,6 +209,18 @@ Errors MAY be raised if either the `Format` is not recognized
 or no value could be extracted, depending on the specific OpenTracing Language API
 (e.g. Go and Python do, but Java may not).
 
+## Close
+
+Mark the `Tracer` Shim as closed, effectively preventing creation of new `Span` instances.
+Subsequent calls to [Start a new Span](#start-a-new-span) MUST return a no-op instance.
```
yurishkuro, line 215: If I recall it was primarily used to force-flush any spans that were already finished but not exported (i.e. sitting in a memory buffer). Not starting new live spans was less of a goal (I don't even recall that one, and not sure why it would be important, because if those spans are not finished than they are not that different from no-op).

carlosalberto: That's correct, that operation was added to force-flush (before closing the app). Since we don't the equivalent in the OTel API, the alternative is that this call becomes a no-op from the OT Shim side (that's also fine by me).

carlosalberto: Thanks. Updated the description. Please review.

specification/compatibility/opentracing.md
```diff
@@ -205,6 +206,12 @@ Errors MAY be raised if either the `Format` is not recognized
 or no value could be extracted, depending on the specific OpenTracing Language API
 (e.g. Go and Python do, but Java may not).
 
+## Close
+
+Optional operation. If this operation is implemented for a specific OpenTracing language,
+it MUST flush the underlying `TracerProvider` if such functionality is defined in the
+OpenTelemetry API, and else it MUST be defined as a no-op operation.
```
yurishkuro, line 213:
```suggestion
OpenTelemetry SDK, otherwise it MUST be defined as a no-op operation.
```

carlosalberto: The OpenTracing Shim is expected to NOT depend on the SDK, but on the API ;(

yurishkuro: Then it seems like we're stuck, because OTEL API does not define Close/Flush equivalent.

carlosalberto: Right. So the options are a) We push for `TracerProvider.Close()` (or `Flush()`) or b) Have a no-op for now, waiting for this to be added in the future. Which one do you suggest we go after?

yurishkuro: One other option is to allow lang-specific duck typing, eg if Java tracer provider implements a Closeable then cast and call it.

carlosalberto: @yurishkuro I prefer this last approach. Update the PR, let me know whether the current wording is clear enough.

specification/compatibility/opentracing.md
```diff
@@ -205,6 +206,16 @@ Errors MAY be raised if either the `Format` is not recognized
...
```
yurishkuro, line 213:
```suggestion
```

specification/compatibility/opentracing.md
```diff
@@ -205,6 +206,16 @@ Errors MAY be raised if either the `Format` is not recognized
...
+
+Optional operation. If this operation is implemented for a specific OpenTracing language,
+it MUST close the underlying `TracerProvider` if it implements a closeable interface or method,
+and else it MUST be defined as a no-op operation.
```
yurishkuro, line 217:
```suggestion
it MUST close the underlying `TracerProvider` if it implements a "closeable" interface or method;
otherwise it MUST be defined as a no-op operation.
```

yurishkuro (approved):

tylerbenson (approved):

ocelotl (approved):

specification/compatibility/opentracing.md
```diff
@@ -205,6 +206,12 @@ Errors MAY be raised if either the `Format` is not recognized
 or no value could be extracted, depending on the specific OpenTracing Language API
 (e.g. Go and Python do, but Java may not).
 
+## Close
+
+Optional operation. If this operation is implemented for a specific OpenTracing language,
```
MrAlias, line 211: https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification#notation-conventions-and-compliance

```suggestion
OPTIONAL operation. If this operation is implemented for a specific OpenTracing language,
```

yurishkuro (approved):

specification/compatibility/opentracing.md
```diff
@@ -205,6 +206,12 @@ Errors MAY be raised if either the `Format` is not recognized
 or no value could be extracted, depending on the specific OpenTracing Language API
 (e.g. Go and Python do, but Java may not).
 
+## Close
+
+OPTIONAL operation. If this operation is implemented for a specific OpenTracing language,
+it MUST close the underlying `TracerProvider` if it implements a "closeable" interface or method;
```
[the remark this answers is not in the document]
jmacd:
```suggestion
it MUST close the underlying `Tracer` if it implements a "closeable" interface or method;
```

By my logic, it makes more sense to ask whether the OpenTracing `Tracer` implements a closeable interface, not the `TracerProvider`.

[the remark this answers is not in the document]
carlosalberto: Oh, this is because OTel exposes this functionality on `TracerProvider`, not `Tracer` :(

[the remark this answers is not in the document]
carlosalberto: Hey @jmacd - circling back, there was a [RFC](https://github.com/opentracing/specification/blob/master/rfc/closeable_tracer.md#specification-changes) in OpenTracing explaining this part:

> For Tracers that record Spans for persistance, calling Close should flush the current in-memory collection to the configured persistance store.

So not quite about reference counting, at least in the general purpose.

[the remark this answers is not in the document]
jmacd: Sounds good -- this leaves a question about the behavior of OpenTracing behavior. If I create two Tracers from the shim, then close them one at a time, does this mean the TracerProvider is shutdown twice?  If this sounds problematic (which it does to me), I think we should say that Close() calls `Flush()`, not `Shutdown()`, to protect against double close?

reyang (approved):

specification/compatibility/opentracing.md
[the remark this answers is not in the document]
carlosalberto: So the problem is that `Flush()` is not 'exposed' by any interface (such as `Closeable` in the Java world). So we'd need to either actually do a direct cast (e.g. to `TracerProviderSdk`) or do a full-duck type access (which in C++ and Java would require reflection instead)

I'd instead add a recommendation on NOT creating more than an OpenTracing `Tracer`  (which is exactly the typical use case in OT), as well a catching exceptions when calling `TracerProvider.Close()` from the Shim side, to at least provide a clean operation. What do you think?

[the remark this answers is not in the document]
jmacd: > I'd instead add a recommendation on NOT creating more than an OpenTracing Tracer (which is exactly the typical use case in OT)

Sounds good! Thanks @carlosalberto.

jmacd (approved):

Issue #3038, opened 2022-12-15
expand the declaration of `pool.name` in the spec


haddasbronfman:
In issue #2958 we talked about the lack of pool name in MySQL (JS) which can be used as the value for attribute `pool.name`.
This led us to use another `pool.name` that is built from the data we have about the pool: `host: localhost port: 3306 database: db_test user: root`.

I think we should add it to the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/database-metrics.md#connection-pools) and declare that if the pool has no name in the DB we are working on, then it will be built like the format above.

PR #3050, opened 2022-12-20
Expand pool.name in semantic conventions

haddasbronfman:
according to the discussion here: https://github.com/open-telemetry/opentelemetry-specification/issues/3038 I'm updating the `pool.name attribute`


specification/metrics/semantic_conventions/database-metrics.md
```diff
@@ -58,4 +58,4 @@ Below is a table of the attributes that MUST be included on all connection pool
 
 | Name        | Type   | Description                                                                  | Examples       | Requirement Level |
 |-------------|--------|------------------------------------------------------------------------------|----------------|-------------------|
-| `pool.name` | string | The name of the connection pool; unique within the instrumented application. | `myDataSource` | Required          |
+| `pool.name` | string | The name of the connection pool; unique within the instrumented application. In case the pool has no name in the DB we are working on, then the name should be bui as follows: `host: <host> port: <port> database: <database name> user: <user>`. | `myDataSource`, `host: localhost port: 3306 database: db_test user: root`  | Required          |
```
arminru, line 61:
```suggestion
| `pool.name` | string | The name of the connection pool; unique within the instrumented application. In case the pool has no name in the monitored database, then the name should be as follows: `host: <host> port: <port> database: <database name> user: <user>`. | `myDataSource`, `host: localhost port: 3306 database: db_test user: root`  | Required          |
```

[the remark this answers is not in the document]
haddasbronfman: No problem. I'll change it.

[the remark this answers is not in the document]
haddasbronfman: I added it in another PR: https://github.com/open-telemetry/opentelemetry-specification/pull/3053 sorry for the confusion.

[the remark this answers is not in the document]
arminru: Thank you, but could you please add the change in this PR?
This way it's ensured to stay consistent as it can only merge atomically.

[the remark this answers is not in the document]
haddasbronfman: done

specification/metrics/semantic_conventions/database-metrics.md
```diff
@@ -58,4 +58,4 @@ Below is a table of the attributes that MUST be included on all connection pool
 
 | Name        | Type   | Description                                                                  | Examples       | Requirement Level |
 |-------------|--------|------------------------------------------------------------------------------|----------------|-------------------|
-| `pool.name` | string | The name of the connection pool; unique within the instrumented application. | `myDataSource` | Required          |
+| `pool.name` | string | The name of the connection pool; unique within the instrumented application. In case the pool has no name in the monitored database, then the name should be as follows: `host: '<host>', port: <port>, database: '<database name>', user: '<user>'`. | `myDataSource`, `host: localhost port: 3306 database: db_test user: root`  | Required          |
```
mateuszrzeszutek, line 61: Nit: consider adding the database connection string as a possible option:
```suggestion
| `pool.name` | string | The name of the connection pool; unique within the instrumented application. In case the pool has no name in the monitored database, consider using the connection string (with password removed). Alternatively, you can use the following pattern: `host: '<host>', port: <port>, database: '<database name>', user: '<user>'`. | `myDataSource`, `host: localhost, port: 3306, database: db_test, user: root`  | Required          |
```

The idea behind the `pool.name` attribute was to make it as easy as possible for the owner of the instrumented app to find out which metric measurements are produced by which connection pools in their app; if the connection pool implementation does not name the pools, usually (if available) the connection string is a good option for differentiating between different connection pools (because, if you have >1 DB connection pools in your app, usually they connect to different DBs/schemas/users).

mateuszrzeszutek (approved):

specification/metrics/semantic_conventions/database-metrics.md
tsloughter: Why isn't pool name just optional?

mateuszrzeszutek: You can have multiple connection pools in your application; and if you do not provide the name, how will you distinguish between them? 
Also, almost all instruments used here are up down counters, if you were to provide a measurement without a name it'd break the "additive" property that counters exhibit (measurement without a name != sum of measurements).

haddasbronfman:
Hi, Can someone do a CR please? (I already have one approval from Mateus)
@open-telemetry/specs-approvers or @open-telemetry/specs-metrics-approvers.
Thanks.

haddasbronfman:
Hi Again, can someone do a CR please? (I already have one approval from Mateus)
@open-telemetry/specs-approvers  or @open-telemetry/specs-metrics-approvers .
Thanks.

carlosalberto:
Added to the discussion tomorrow at the Specification - this may have to be reviewed as well by the (new) semantic conventions working group.

specification/metrics/semantic_conventions/database-metrics.md
Aneurysm9: Connection pools aren't required to use a database, though, are they?  I know that they're best practice but it is certainly possible to use, and instrument the use of, a database without connection pooling.

carlosalberto (approved):

carlosalberto:
Overall LGTM - this is not adding a new semantic convention, but suggesting a value for this (specially useful as, right now, it's marked as as required field). We can iterate later to have incremental improvements if/as needed.

@jsuereth This _shouldn't_ impact any of the issues the semantic conventions WG is afraid of, specially as it already exists, but otherwise let me know.

haddasbronfman:
@open-telemetry/arrow-approvers. According to Carlos's comment can someone please do a CR?

haddasbronfman:
@open-telemetry/specs-metrics-approvers Hi again, this PR was already stale and I don't want it to be close. 
Can someone please review it?

specification/metrics/semantic_conventions/database-metrics.md
arminru: The connection string could be an option. We also have it as a dedicated attribute for traces already (and also just recommend to strip credentials there): https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#connection-level-attributes

arminru, line 61:
```suggestion
| `pool.name` | string | The name of the connection pool; unique within the instrumented application. In case the pool has no name in the monitored database, then the name should be as follows: `host: '<host>', port: <port>, database: '<database name>', user: '<user>'`. | `myDataSource`, `host: localhost, port: 3306, database: 'db_test', user: 'root'`  | Required          |
```

haddasbronfman: You mean that instead of "building" the pool name as suggested here, we will use the already existing parameter DB_CONNECTION_STRING to identify the pool name? 

If yes, then I should change the definition of pool.name above to something like that: "...In case the pool has no name in the monitored database, then the name should be the same as 'db.connection_string' defined [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#connection-level-attributes)". agreed?

arminru: I'm not sure if the entire (redacted) connection string is suitable for metrics with regard to cardinality.
@open-telemetry/specs-metrics-approvers any opinions?

arminru: As for omitting `pool.name` in favor of `db.connection_string` in certain cases, this might be problematic for metrics consumers, as they'd have to consider both attributes for querying and aggregating/splitting them.

If we want to go with the connection string fallback, defining it to be set as `pool.name` might be better than introducing a separate `db.connection_string` for metrics - currently we only have it for traces.

haddasbronfman: yes. that was exactly the purpose (at least as I understood it). in case we don't have the name of the pool, we will do: `setAttribure(pool.name, <the values of db.connection_string>)` - meaning that we will set the `pool.name` to be the value of the connection_string.  @mateuszrzeszutek  what do you think?

mateuszrzeszutek: > If we want to go with the connection string fallback, defining it to be set as `pool.name` might be better than introducing a separate `db.connection_string` for metrics - currently we only have it for traces.

👍 
`pool.name` values have to be different in different connection pools; and ideally it should be easy to find a particular pool by its name in the application code. I think that the connection string somewhat fulfills these conditions; if a particular connection pool implementation does not provide a name the connection string should work as a fallback.

haddasbronfman: I agree with you. I committed another change. please review it when you have time. thanks.

specification/metrics/semantic_conventions/database-metrics.md
```diff
@@ -58,4 +58,4 @@ Below is a table of the attributes that MUST be included on all connection pool
 
 | Name        | Type   | Description                                                                  | Examples       | Requirement Level |
 |-------------|--------|------------------------------------------------------------------------------|----------------|-------------------|
-| `pool.name` | string | The name of the connection pool; unique within the instrumented application. | `myDataSource` | Required          |
+| `pool.name` | string | The name of the connection pool; unique within the instrumented application. In case the pool has no name in the monitored database, then the [db.connection_string](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#connection-level-attributes) should be used. | `myDataSource` | Required          |
```
mateuszrzeszutek, line 61: Nit
```suggestion
| `pool.name` | string | The name of the connection pool; unique within the instrumented application. In case the connection pool implementation does not provide a name, then the [db.connection_string](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#connection-level-attributes) should be used. | `myDataSource` | Required          |
```

mateuszrzeszutek (approved):

carlosalberto:
@open-telemetry/specs-approvers please review. There was a last cycle of feedback applied, so this should look better.

haddasbronfman:
Hi, 
it would be great to have approval for this issue from one of you please: 
@jsuereth 
@bogdandrutu 
@jmacd 
@tigrannajaryan 
@arminru 
@reyang

jsuereth (approved):

carlosalberto:
@haddasbronfman Once the markup is fixed (as @jsuereth mentioned), we are ready to merge.

PR #3052, opened 2022-12-21
correct exporter user-agent example

robbkidd:
## Changes

Removes spaces from the example exporter User-Agent string to conform to header field value component definition.

While this section about a User-Agent references [RFC7231](https://datatracker.ietf.org/doc/html/rfc7231#section-5.5.3) for format, the example provided includes spaces in the `product` identifier. The format for `product` is as a `token`, defined in [RFC7230](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.6), which does not include space in the set of valid characters. Including spaces here will throw off most existing user-agent parsing.

Related issues: example added in #2684


arminru:
Thank you!

Please also add an entry in `CHANGELOG.md` so SDK and exporter implementors are made aware of the fix, even if it's just in the example.

arminru (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ release.
 
 ## Unreleased
 
+- Remove spaces from example exporter User-Agent header to conform to RFC7231 & RFC7230.
```
carlosalberto, line 10: I don't think this needs/deserves a CHANGELOG entry (also, this section should be empty, as only the subsections are used 😄 )

carlosalberto:
cc @codeboten

codeboten (approved):

MrAlias (approved):

robbkidd:
I am unclear on what change I should make: remove the changelog entry or move it to a different section?

If a different section, which section? **OpenTelemetry Protocol** because this is about OTLP exporters?

carlosalberto:
@robbkidd Oh, I meant removing the entry from the Changelog basically. On a second though, we want to be explicit to SIGs about updating their user agents, so leets keep keep the entry.

carlosalberto (approved):

PR #3053, opened 2022-12-21
Update CHANGELOG.md

haddasbronfman:
according to #3050


arminru (changes requested):
This should be added in #3050 directly so it's only merged atomically, otherwise the changelog might be inconsistent.

haddasbronfman:
This PR content is now part of this PR: https://github.com/open-telemetry/opentelemetry-specification/pull/3050

PR #3056, opened 2022-12-21
Prometheus Unknown MUST preserve its notion

### LINKED ISSUE #3058: Prometheus Unknown MUST preserve its notion

Preserve the Prometheus Unknown's notion since it would be [convert to Gauge ](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/4eaa9f8f8c89492e4df873bda483ea49da8b194f/receiver/prometheusreceiver/internal/util.go#L111-L112)as discussed in the following two issues: https://github.com/open-telemetry/wg-prometheus/issues/69 and https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/16768

**Additional context.**
* To understand more about Prometheus Type, please [follow this document](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md)
* Understand the conversion between [OTEL and Prometheus](https://github.com/influxdata/influxdb-observability/blob/main/docs/metrics.md#schema-metricsschematelegrafprometheusv1)

khanhntd:
Closes #3058 

## Changes
As Prometheus Receiver currently, all [untyped prometheus metrics are converted to OTEL Gauge](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/4eaa9f8f8c89492e4df873bda483ea49da8b194f/receiver/prometheusreceiver/internal/util.go#L111-L112) since the whole reason was not to drop untyped prometheus metrics. However, for some of the backend (such as CloudWatch) using with awsemf as an exporter/proxy, they [will have limit bound (not supporting +inf or -inf)](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_PutMetricData.html) and combining with some case untyped can be counter, this won't work and will lose metrics

For non-trivial changes, follow the [change proposal process](https://github.com/open-telemetry/opentelemetry-specification/blob/main/CONTRIBUTING.md#proposing-a-change) and link to the related issue(s) and/or [OTEP(s)](https://github.com/open-telemetry/oteps), update the [`CHANGELOG.md`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/CHANGELOG.md), and also be sure to update [`spec-compliance-matrix.md`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md) if necessary.

## Related issues
* https://github.com/open-telemetry/wg-prometheus/issues/69
* https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/16768

## Note
As discussed in the Prometheus WG Group meeting with this [design document](https://docs.google.com/document/d/1uCWJEQ9XjHD-HYFr6Hz5QjpA27Q0yGH0AVON48vQo_k/edit?usp=drive_web&ouid=108605315929149960882), we have agreed to use the attribute to signal the different between gauge and unknown.


specification/compatibility/prometheus_and_openmetrics.md
```diff
@@ -103,6 +103,8 @@ Multiple Prometheus metrics are merged together into a single OTLP Summary:
 
 * The `quantile` label on non-suffixed metrics is used to identify quantile points in summary metrics. Each Prometheus line produces one quantile on the resulting summary.
 * Lines with `_count` and `_sum` suffixes are used to determine the summary's count and sum.
+* If `_count` is not present, the metric MUST be dropped.
```
dashpole, line 106: Can you split this change out into its own PR?

specification/compatibility/prometheus_and_openmetrics.md
```diff
@@ -82,7 +82,7 @@ An [OpenMetrics StateSet](https://github.com/OpenObservability/OpenMetrics/blob/
 
 ### Unknown-typed
 
-A [Prometheus Unknown](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#unknown) MUST be converted to an OTLP Gauge.
+A [Prometheus Unknown](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#unknown) MUST be converted to an OTLP Gauge and MUST preserve its notion (e.g a datapoint attribute specifies unknown-typed)
```
dashpole, line 85: I think we will need to be more prescriptive here.  We should say exactly which attribute we will add.

I also think we should try for the data point flag first before we settle for a datapoint attribute.

specification/compatibility/prometheus_and_openmetrics.md
khanhntd: Yes. Sure thing though !

specification/compatibility/prometheus_and_openmetrics.md
khanhntd: > I think we will need to be more prescriptive here. We should say exactly which attribute we will add.
Will update this.


> I also think we should try for the data point flag first before we settle for a datapoint attribute.
For this, can I ask what is the crossing point between the two since there would be trade off? (e.g performance)

khanhntd: Another question from me: would this attribute can be apply for others Prom Type. It would be good to signal to all exporters the correct type too?

specification/compatibility/prometheus_and_openmetrics.md
```diff
@@ -82,7 +82,7 @@ An [OpenMetrics StateSet](https://github.com/OpenObservability/OpenMetrics/blob/
 
 ### Unknown-typed
 
-A [Prometheus Unknown](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#unknown) MUST be converted to an OTLP Gauge.
+A [Prometheus Unknown](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#unknown) MUST be converted to an OTLP Gauge and MUST preserve its notion (a datapoint attribute `prom_metric_type` specifies [Prometheus unknown-typed](https://github.com/prometheus/prometheus/blob/fa6e05903fd3ce52e374a6e1bf4eb98c9f1f45a7/model/textparse/interface.go#L117))
```
Aneurysm9, line 85:
```suggestion
A [Prometheus Unknown](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#unknown) MUST be converted to an OTLP Gauge and MUST preserve its original type information.  Type preservation MUST be achieved by adding a datapoint attribute `prom_metric_type` with the value of [Prometheus' `MetricTypeUnknown`](https://github.com/prometheus/prometheus/blob/fa6e05903fd3ce52e374a6e1bf4eb98c9f1f45a7/model/textparse/interface.go#L117)).
```

I suggest tightening up the language a bit.  I'm not sure whether we want to make specification dependent on constants defined in Prometheus' code.  Should we define our own mapping or values for this attribute?

khanhntd: Using map would give more varies language for each type of exporter; however, would a general constant that shared between processor/exporter be a good idea though ? I can put this attribute in the internal common constant that can be shared and unified amongst the plugins

Aneurysm9: How collector components deal with such an attribute and its value is an implementation detail for the collector.  Here in the spec we need to define the name of the attribute and the values that are valid.

khanhntd: > How collector components deal with such an attribute and its value is an implementation detail for the collector. Here in the spec we need to define the name of the attribute and the values that are valid.

Understandable. Cause I  thought mapping is defining a set of definition of Prometheus `MetricTypeUnknown` for each of the plugins and using a constant would give a unification constant for all the plugins. So I'm weighing the latter.

specification/compatibility/prometheus_and_openmetrics.md
khanhntd: Hey @dashpole, any feedbacks though !!

specification/compatibility/prometheus_and_openmetrics.md
```diff
@@ -82,7 +82,7 @@ An [OpenMetrics StateSet](https://github.com/OpenObservability/OpenMetrics/blob/
 
 ### Unknown-typed
 
-A [Prometheus Unknown](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#unknown) MUST be converted to an OTLP Gauge.
+A [Prometheus Unknown](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#unknown) MUST be converted to an OTLP Gauge and MUST preserve its original type information. Type preservation MUST be achieved by adding a datapoint attribute `prom_metric_type` with the value of [Prometheus `MetricTypeUnknown`](https://github.com/prometheus/prometheus/blob/fa6e05903fd3ce52e374a6e1bf4eb98c9f1f45a7/model/textparse/interface.go#L117)
```
jmacd, line 85: I wonder if we can overload the `unit` or `description` field to include this information, to avoid placing a metric-level property on every data point.

For example, we could specify a structured description field, one where optional metadata can be added w/o loss of information. As a proof of concept, I did this for the Lightstep metric's SDK to pass hints into the SDK as a JSON blob.  https://github.com/lightstep/otel-launcher-go/tree/main/lightstep/sdk/metric#metric-instrument-hints-api

So for example, the user-provided description string is just one field in the JSON blob. I used this field to place a specifically requested aggregator config, but it could be used in a Prometheus context to report additional qualities such as an unknown aggregation.

khanhntd: Interesting. However, would this request the customers to specifically know the keywords in order to translate to the required fields in JSON blob? If yes, then would it be non-deterministic and result in losing information ?

khanhntd:
Hi @bogdandrutu, could you help me in remove the stale labels since the discussion are still on-going?

specification/compatibility/prometheus_and_openmetrics.md
khanhntd: Another thing: federate metrics or third party metrics are unlikely to have description field. Therefore, reduce the options for us to leverage the hints api

khanhntd:
Still active.

specification/compatibility/prometheus_and_openmetrics.md
```diff
@@ -87,7 +87,7 @@ An [OpenMetrics StateSet](https://github.com/OpenObservability/OpenMetrics/blob/
...
```
dashpole, line 90:
```suggestion
A [Prometheus Unknown](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#unknown) MUST be converted to an OTLP Gauge. It MUST add an additional a datapoint attribute to all datapoints with key `prometheus.type` and value `unknown`, to match the [OpenMetrics type string](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#type).
```

Few fixes:

* No need to talk about preserving type information.  Just talk about what needs to be done (add the attribute)
* Use dots, rather than underscores for the attribute key.
* Omit "metric" in the attribute key, as that is implicit with prometheus (willing to be convinced otherwise here)
* Use prometheus instead of the prom abbreviation.
* Link to the openmetrics spec instead of a go library, and add the literal value instead of the Go type.

dashpole (commented):
Open Question:

Should we use `prometheus.type` or `openmetrics.type`?

dashpole (commented):
I think this is probably the least bad option we've considered.  I'd like to see this make progress.

Can you add the inverse of this to the specification for collector prometheus exporters?  We should explicitly exclude SDK exporters, as this case won't apply.

specification/compatibility/prometheus_and_openmetrics.md
khanhntd: Sorry forgot to comment

khanhntd: > Use prometheus instead of the prom abbreviation.

Agree on this too !!

khanhntd:
```
Should we use prometheus.type or openmetrics.type?
```
Would it be better to use `prometheus.type` since we would also widely support this attribute for StatefulSet and Info even though I agree those two are mostly about attribute's value.

khanhntd:
> Can you add the inverse of this to the specification for collector prometheus exporters? We should explicitly exclude SDK exporters, as this case won't apply.

I'm good with that too. There are:
* `[prometheusexporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/prometheusexporter)` 
* [prometheusremotewriteexporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/prometheusremotewriteexporter)

Are there any inverse for GCP and other exporter we would need to aware of too?

specification/compatibility/prometheus_and_openmetrics.md
khanhntd: > Link to the openmetrics spec instead of a go library, and add the literal value instead of the Go type.

This might be a concern IMO. [A related discussion](https://github.com/open-telemetry/opentelemetry-specification/pull/3056#issuecomment-1414381652)

dashpole:
> Would it be better to use prometheus.type since we would also widely support this attribute for StatefulSet and Info even though I agree those two are mostly about attribute's value.

Note that stateset and info types do not exist in prometheus.  They are openmetrics-specific types, which doesn't support that reasoning.

> Are there any inverse for GCP and other exporter we would need to aware of too?

The phrasing from the [resource attributes section](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#resource-attributes-1) should be fine: "In the Collector's Prometheus pull and push (remote-write) exporters, ...".  It probably belongs in the [OpenTelemetry gauge](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#gauges-1) section.

One thing to consider, is what we should do if data points have different values for `prometheus.type`

khanhntd:
> Note that stateset and info types do not exist in prometheus. They are openmetrics-specific types, which doesn't support that reasoning.

Ahh. I thought you talk about which library should we take the type from (not the constant name)

> The phrasing from the [resource attributes section](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#resource-attributes-1) should be fine: "In the Collector's Prometheus pull and push (remote-write) exporters, ...". It probably belongs in the [OpenTelemetry gauge](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#gauges-1) section.

I agree its only prometheus related plugin. 

> One thing to consider, is what we should do if data points have different values for prometheus.type

You meant when  we will have additional Prom type leverage this attribute?

dashpole:
> You meant when we will have additional Prom type leverage this attribute?

If we get a OTel metric `foo` with two datapoints in the collector's prom exporter: 

* labels [`prometheus.type` = `unknown`]; value 1
* labels [`prometheus.type` = `stateset`]; value 2

My suggestion would be that we leave the metric untouched if `prometheus.type` labels disagree.

khanhntd:
> If we get a OTel metric foo with two datapoints in the collector's prom exporter:

Good point. Another would be: would we also applies the same with receivers when the customers specifies to `relabel` or add a label to override this attribute.

dashpole:
We should probably leave `prometheus.type` labels as-is if it already exists in the receiver.

khanhntd:
> My suggestion would be that we leave the metric untouched if prometheus.type labels disagree.

If we only added `prometheus.type` only when Prometheus unknown, there is a low chance of having another labels with others type such as `Statefulset` even if it's not a receiver IMO (since its only a communication between prometheus plugins together).

dashpole:
Feedback from the SIG meeting:

* Consider adding an Unknown metric type, potentially as a "subtype" of gauge.  It should be collector/protocol/transport-only, and should not impact the API/SDK.
* Please look at other common metric types which may have an notion of unknown metrics (statsd?, other protocols supported by the collector?). What values can other unknown metrics have?  String?  or all float?

khanhntd:
Will updates the information within today !

khanhntd:
Was not able to do it last week since its a hectic week. However, will schedule some time for myself to implement all the asks  from spec meeting.

khanhntd:
Not stale though

khanhntd:
Almost finish the list.

damemi:
bump, any updates on this?

PR #3059, opened 2022-12-21
Add condition with sum and count for Prometheus summaries

khanhntd:
## Changes
Update condition with sum and count for prometheus summaries since summaries [will be dropped if count is not exist ](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/prometheusreceiver/internal/metricfamily.go#L163-L165) and if sum is not exist, then [it should not be set](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/prometheusreceiver/internal/metricfamily.go#L174-L176). It is the [same behavior with histogram](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/prometheusreceiver/internal/metricfamily.go#L99-L137) so [there are not changes in the wording](https://github.com/open-telemetry/opentelemetry-specification/blob/71125db55a0cd58e5982bd8e3243a0ff7a24fb33/specification/compatibility/prometheus_and_openmetrics.md#histograms).


jack-berg (commented):
Seems reasonable.

jack-berg (approved):

specification/compatibility/prometheus_and_openmetrics.md
```diff
@@ -103,6 +103,8 @@ Multiple Prometheus metrics are merged together into a single OTLP Summary:
 
 * The `quantile` label on non-suffixed metrics is used to identify quantile points in summary metrics. Each Prometheus line produces one quantile on the resulting summary.
 * Lines with `_count` and `_sum` suffixes are used to determine the summary's count and sum.
+* If `_count` is not present, the metric MUST be dropped.
+* If `_sum` is not present, the summaries's sum MUST be unset.
```
dashpole, line 107: In OTLP, the [sum of a summary](https://github.com/open-telemetry/opentelemetry-proto/blob/d8729d40f629dba12694b44c4c32c1eab109b00a/opentelemetry/proto/metrics/v1/metrics.proto#L601) is not an optional field, like it [is in a histogram](https://github.com/open-telemetry/opentelemetry-proto/blob/d8729d40f629dba12694b44c4c32c1eab109b00a/opentelemetry/proto/metrics/v1/metrics.proto#L417).

Should this say "the summary's sum MUST be set to zero" instead?

khanhntd: Interesting. Thanks for showing that sum of summary is a required field. In a way, we should also update the behavior for `prometheusreceiver`  since[ its only set](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/prometheusreceiver/internal/metricfamily.go#L174-L176) when there is a SUM.

Aneurysm9: Is setting the value to 0 correct, or should the data point be dropped?

dashpole: The default value is zero, so the code is fine.  I'm mostly just calling out that downstream consumers of the data can't distinguish between unset and set to zero in the same way they can for histograms.

MrAlias (approved):

specification/compatibility/prometheus_and_openmetrics.md
khanhntd: I wonder why the summary is dropped when count does not exist, since count and sum in a way can be recomputed if there is enough data in the quantiles (same applies for histogram). If I am wrong, please correct me. It will agree, losing the metric information overtime since its cumulative

khanhntd:
cc @tigrannajaryan

tigrannajaryan:

specification/compatibility/prometheus_and_openmetrics.md
dashpole: I don't believe that is correct.  Quantiles (e.g. knowing that the 90th percentile is 10) don't tell me anything about the number of events in the population, or the sum of events in the population.

dashpole (approved):

specification/compatibility/prometheus_and_openmetrics.md
khanhntd: I might disagree with that (e.g 10% percentile is 10 while 15% percentile is 15) which shows 10 % population is 10 but also show 5% population (between 10% and 15% percentile is around 15) which can approximately calculate the sum and also the count.

jmacd (approved):

CHANGELOG.md
```diff
@@ -33,6 +33,9 @@ release.
 
 ### Compatibility
 
+- Add condition with sum and count for prometheus summaries
```
reyang, line 36: nit:

```suggestion
- Add condition with sum and count for Prometheus summaries
```

reyang (approved):

CHANGELOG.md
khanhntd: Thanks for catching Yang. I have fixed it.

specification/compatibility/prometheus_and_openmetrics.md
```diff
@@ -103,6 +103,8 @@ Multiple Prometheus metrics are merged together into a single OTLP Summary:
 
 * The `quantile` label on non-suffixed metrics is used to identify quantile points in summary metrics. Each Prometheus line produces one quantile on the resulting summary.
 * Lines with `_count` and `_sum` suffixes are used to determine the summary's count and sum.
+* If `_count` is not present, the metric MUST be dropped.
+* If `_sum` is not present, the summaries's sum MUST be [set to zero.](https://github.com/open-telemetry/opentelemetry-proto/blob/d8729d40f629dba12694b44c4c32c1eab109b00a/opentelemetry/proto/metrics/v1/metrics.proto#L601)
```
Aneurysm9, line 107:
```suggestion
* If `_sum` is not present, the summary's sum MUST be [set to zero.](https://github.com/open-telemetry/opentelemetry-proto/blob/d8729d40f629dba12694b44c4c32c1eab109b00a/opentelemetry/proto/metrics/v1/metrics.proto#L601)
```

Aneurysm9 (approved):

specification/compatibility/prometheus_and_openmetrics.md
khanhntd: Thanks for catching Anthony and I have fixed it in the newest commit !!!

khanhntd:
cc @tigrannajaryan  for re=run the workflow and merging

dashpole (approved):

tigrannajaryan:
@open-telemetry/specs-metrics-approvers this has required number of approvals. I will merge in 2 business days if I don't see objections.

PR #3066, opened 2023-01-03
Clarify prometheus unit conversions

dashpole:
Addresses confusion in this comment https://github.com/open-telemetry/opentelemetry-java/issues/4390#issuecomment-1363304821 by more completely describing the translation steps required for translating between prometheus unit conventions and opentelemetry unit conventions.

This draws on https://prometheus.io/docs/practices/naming/#base-units and https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#units-and-base-units.  It matches the current implementation of the conversion in the collector: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/translator/prometheus/normalize_name.go except that it maintains the current recommendation to convert to base units in Prometheus exporters.

cc @jack-berg @psx95 @jmacd @gouthamve @open-telemetry/wg-prometheus @bertysentry


MrAlias (approved):

carlosalberto (approved):

bertysentry:
Actually the description of the translation of Otel to Prometheus metrics is described here for the translator: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/translator/prometheus

Side note: when will we enable the `pkg.translator.prometheus.NormalizeName` feature gate by default?

bertysentry (approved):

gouthamve (approved):
LGTM

specification/compatibility/prometheus_and_openmetrics.md
```diff
@@ -221,12 +226,16 @@ required to match the regex: `[a-zA-Z_:]([a-zA-Z0-9_:])*`. Invalid characters
...
+The Unit of an OTLP metric point SHOULD be converted to the equivalent unit in Prometheus when possible.  This includes:
+
+* Converting from abbreviations to full words (e.g. "ms" to "milliseconds").
+* Converting to [base units](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#units-and-base-units) (e.g. "milliseconds" to "seconds"), which also requires adjusting the value.
+* Dropping the portions of the Unit within brackets (e.g. {packets}). Brackets MUST NOT be included in the resulting unit. A "count of foo" is considered unitless in Prometheus.
+* Special case: Converting "1" to "ratio".
+* Converting "foo/bar" to "foo_per_bar".
+
+The resulting unit SHOULD be added to the metric as
+[OpenMetrics UNIT metadata](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metricfamily) and as a suffix to the metric name, or the unit MUST be omitted. The unit suffix comes before any type-specific suffixes.
```
fstab, line 238: It could say that the unit should be added as a suffix to the metric name _unless the metric name already contains the unit`.

fstab (changes requested):
Hi :wave:, I came across this PR and added a few comments. Thanks a lot for taking care of this.

specification/compatibility/prometheus_and_openmetrics.md
```diff
@@ -221,12 +226,16 @@ required to match the regex: `[a-zA-Z_:]([a-zA-Z0-9_:])*`. Invalid characters
...
-Additionally, the unit MUST be added as a suffix to the metric name, and SHOULD
-be converted to [base units](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#units-and-base-units)
-recommended by OpenMetrics when possible.  The unit suffix comes before any
-type-specific suffixes.
+The Unit of an OTLP metric point SHOULD be converted to the equivalent unit in Prometheus when possible.  This includes:
+
+* Converting from abbreviations to full words (e.g. "ms" to "milliseconds").
+* Converting to [base units](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#units-and-base-units) (e.g. "milliseconds" to "seconds"), which also requires adjusting the value.
+* Dropping the portions of the Unit within brackets (e.g. {packets}). Brackets MUST NOT be included in the resulting unit. A "count of foo" is considered unitless in Prometheus.
+* Special case: Converting "1" to "ratio".
```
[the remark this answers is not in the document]
dashpole: (note, this is recommending `_ratio`, rather than `_rate`)

In that case, the name would be `system_filesystem_utilization_ratio`, which sounds more reasonable.

specification/compatibility/prometheus_and_openmetrics.md
```diff
@@ -221,12 +226,16 @@ required to match the regex: `[a-zA-Z_:]([a-zA-Z0-9_:])*`. Invalid characters
...
+* Converting "foo/bar" to "foo_per_bar".
```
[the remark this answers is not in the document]
dashpole: I agree the resulting name doesn't sound great.

Our conventions try to encourage users [not to include unit information](https://opentelemetry.io/docs/reference/specification/metrics/semantic_conventions/#units) in the metric name: `Conventional metrics or metrics that have their units included in OpenTelemetry metadata (e.g. metric.WithUnit in Go) SHOULD NOT include the units in the metric name.`, meaning this metric would ideally be named `revenue`, with a unit of `dollars/item` -> `revenue_dollars_per_item`.  But users may choose not to follow that, or may interpret the `per_item` on the metric name as not being "the unit" that the spec recommends against.

For OpenMetrics, we are [required](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#unit) to add the unit suffix to the metric name if we include unit metadata, so we either have to choose a different translation of `dollars/item` in otel, or drop the unit if it includes `/`.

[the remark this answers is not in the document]
gouthamve: I would vote for dropping the unit if it has `/` in it for now. I can't think of valid resultant names in this case.

specification/compatibility/prometheus_and_openmetrics.md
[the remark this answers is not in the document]
gouthamve: Prometheus also recommends `ratio` in these cases: https://prometheus.io/docs/practices/naming/#base-units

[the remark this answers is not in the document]
bertysentry: `_ratio` suffix is correct.

`system.filesystem.utilization` must be converted to `system_filesystem_utilization_ratio` to follow Prometheus naming conventions.

specification/compatibility/prometheus_and_openmetrics.md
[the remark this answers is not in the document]
bertysentry: Including unit `foo/bar` in metric name as `foo_per_bar` is correct and the expected behavior, if we follow Prometheus naming conventions.

@fstab WRT duplicate tokens in the resulting metric name, the [current implementation of OpenTelemetry's Prometheus Exporters](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/ca0bb5627270ef8b039a9543f2b09e637e901c96/pkg/translator/prometheus/normalize_name.go#L152) handles this case nicely by dropping tokens already present. If you have a metric `revenue_per_item` with a unit `dollars/item` you would end up with `revenue_per_item_dollars`, which is probably the best name for this quantity (IMO the unit should be `dollars` and not `dollars/item`).

@gouthamve There are plenty of units in Otel's current semantic conventions that use `/` in their unit and that must be converted to `_per_` metric name to follow Prometheus naming conventions. A few examples are in the [unit test of the Otel-to-Prometheus metric name translator](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/ca0bb5627270ef8b039a9543f2b09e637e901c96/pkg/translator/prometheus/normalize_name_test.go#L140). My opinion is that a proper support for these metrics is important to avoid late breaking changes.

jmacd (approved):
Looks good. Let's make sure we have Prometheus developers' approval before merging.

specification/compatibility/prometheus_and_openmetrics.md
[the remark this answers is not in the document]
gouthamve: The unit test examples make it clear, thanks @bertysentry!

specification/compatibility/prometheus_and_openmetrics.md
```diff
@@ -221,11 +226,18 @@ required to match the regex: `[a-zA-Z_:]([a-zA-Z0-9_:])*`. Invalid characters
...
 
-The Unit of an OTLP metric point MUST be added as
-[OpenMetrics UNIT metadata](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metricfamily).
-Additionally, the unit MUST be added as a suffix to the metric name, and SHOULD
-be converted to [base units](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#units-and-base-units)
-recommended by OpenMetrics when possible.  The unit suffix comes before any
+The Unit of an OTLP metric point SHOULD be converted to the equivalent unit in Prometheus when possible.  This includes:
+
+* Converting from abbreviations to full words (e.g. "ms" to "milliseconds").
+* Converting to [base units](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#units-and-base-units) (e.g. "milliseconds" to "seconds"), which also requires adjusting the value.
```
fstab, line 232: Sorry for being late, I just saw that this is not only about metric names, but also about automatically converting units, like milliseconds to seconds. Automatic conversion will work for static histograms (converting the `le` labels) but not for exponential histograms. My understanding is that the OTel SDKs offer a generic API, and whether or not this is represented as a static histogram or an exponential histogram is a matter of configuring the view (which can be configured outside of the application at deploy time). Switching the view from static histogram to exponential histogram would switch the unit from seconds to milliseconds in Prometheus:
* Switching the view from static histograms to exponential histograms would scale the values up by a factor of 1000 (seconds to milliseconds).
* Switching the view from static histograms to exponential histograms would change the metric name, for example from `my_latencies_seconds` to `my_latencies_milliseconds`.

So switching the view from static histograms to exponential histograms would break dashboards. I'm not sure if this is desired. What do you think?

fstab: I just learned from other Prometheus folks that you'd need to adapt dashboards anyway when switching from static histograms to exponential histograms.

Example query for a static histogram:

```
histogram_quantile(0.9, sum by (le) (rate(my_latencies_bucket[5m])) )
```

Example query for an exponential histogram:

```
histogram_quantile(0.9, sum(rate(my_latencies[5m])))
```

So maybe introducing automatic unit conversion is ok given that you cannot change the view from static histogram to exponential histogram anyways.

fstab: Quick update: I put this on the agenda of the Prometheus Dev summit on 26 January. I want to discuss with the Prometheus community whether there should be an effort to implement a unified query for static and exponential histograms in Prometheus, so that you can switch histogram types without adapting dashboards. I'll post the result here.

jack-berg: More [related discussion](https://github.com/open-telemetry/opentelemetry-specification/issues/2977).

I think we should hold off on this specific portion of the PR, but merge the rest.

One concept that was previously discussed was the idea of a metric processor, which would have the opportunity to process raw measurements prior to aggregation. Unit conversion to meet the requirements of a particular exporter would be a good use case.

dashpole: I agree.  Lets merge the rest without the recommendation to convert to base units, and continue the discussion in #2977.  I've removed this section.

fstab: Update from the Prometheus dev summit: The topic _Common histogram_quantile() function that works with static and native histograms_ didn't get enough votes to make it on the agenda. Based on the low interest and on the feedback in the [doc](https://docs.google.com/document/d/11LC3wJcVk00l8w5P3oLQ-m3Y37iom6INAMEu2ZAGIIE/edit?usp=sharing) I think it's unlikely that we'll get a common `histogram_quantile()` function, i.e. switching the view from static histograms to exponential histograms will break dashboards and alerting rules anyway. So I guess there's no need to keep the names of static histograms and exponential histograms consistent.

specification/compatibility/prometheus_and_openmetrics.md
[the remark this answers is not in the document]
jmacd: Looks resolved.

specification/compatibility/prometheus_and_openmetrics.md
dashpole: Discussed in the prometheus WG.  We will leave out conversion to base units while the seconds vs milliseconds discussion is ongoing.

jmacd: Looks resolved.

